PR #1289, opened 2020-03-19
ORDER BY Resume Optimization

bchong95:
# ORDER BY Resume Optimization

## Description

For cross partition ORDER BY queries we rewrite the query to inject a filter. For example if we left off on partition N with the value `("A", 1)`, then the filters are:

`(> "A") OR (= "A" AND > 1), (> "A") OR (= "A" AND >= 1), (> "A") OR (= "A" AND >= 1)`

The problem is that this filter is not efficient for the backend, since it can not leverage a composite index if any of the prefix filters have an inequality.

The solution is to just filter on `WHERE true` for the partition that we left off on, since we already have a continuation token. Note that for a single partition query we always resume on the partition we left off on meaning this optimization will pay extra dividends there.

In parallel backend will work on pushing the generated filter to the index, since we still need it for the other N - 1 partitions.

PR #3525, opened 2022-10-29
Query: Fixes performance regression on target partition on some ORDER BY queries with continuation

neildsh:
## Description

Fixes performance regression introduced by #1289

The lack of query filters being sent to the target partition can cause excess round trips as the sdk tries to skip over values that have already been seen.

In case of CRM ([ICM 342678670](https://portal.microsofticm.com/imp/v3/incidents/details/342678670/home)) it lead to over 12k round trips. After the fix, there were only 31 round trips.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/OrderBy/OrderByCrossPartitionQueryPipelineStage.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/OrderBy/OrderByCrossPartitionQueryPipelineStage.cs
@@ -1029,7 +1029,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.CrossPartition.OrderBy
 
             // For the target filter we can make an optimization to just return "true",
             // since we already have the backend continuation token to resume with.
-            return (left.ToString(), TrueFilter, right.ToString());
+            return (left.ToString(), target.ToString(), right.ToString());
         }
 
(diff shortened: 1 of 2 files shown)


Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/OrderBy/OrderByCrossPartitionQueryPipelineStage.cs
```diff
@@ -1027,9 +1027,7 @@ private static void AppendToBuilders((StringBuilder leftFilter, StringBuilder ta
                 }
             }
 
-            // For the target filter we can make an optimization to just return "true",
-            // since we already have the backend continuation token to resume with.
-            return (left.ToString(), TrueFilter, right.ToString());
+            return (left.ToString(), target.ToString(), right.ToString());
```
ealsur, line 1030: We can add the originator of the regression on the changelog notes, should the title of the PR be more around the scenario? Would it be Fixes performance regression on ORDER BY queries with continuation ?

sboshra (approved):

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/OrderBy/OrderByCrossPartitionQueryPipelineStage.cs
neildsh: Fixed

PR #3555, opened 2022-11-09
Client Retry Policy: Adds HTTP timeouts with request-level cross-region retry

### LINKED ISSUE #3497: Client Retry Policy: Handle HTTP timeouts with request-level cross-region retry
Reference: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3467

After retries have been exhausted, the `CosmosHttpClientCore` component should throw a 503 with the HTTP timeout as inner exception.

The 503 should reach the ClientRetryPolicy and be handled like 503s are right now: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs#L250

Identify which `HttpTimeoutPolicy` is used for each of the cases:

* Data Plane
* Metadata
* Query Plan
* Address requests

For the cases described in the referenced table (pasting here for easy access), we should convert the last timeout into a 503 with the timeout as an InnerException:

| Type | Read/Write | 
|----------|------------|
| Data plane |  Read |
| Data plane |  Write |
| Metadata |  Read | |
|  Query Plan | Read (it is always read) | 

## Tasks
- [ ] Complete use case scenarios
- [ ] Sequence diagram
- [ ] Presentation and design discussion
- [ ] Define deliverables, delivery date and project milestone

## Use case scenarios

| Data plan in Gateway mode for Reads, Request Timeout |
|----|
| Given a CX wants to read an existing item |
| When I configure my _CosmosClient.ConnectionMode_ for **Gateway** mode (which uses an HTTP protocol) |
|    - And it will send the operational request using _SendHttpHelperAsync_ |
|    - And the retries will throw a 408 (Request Timeout) exception |
|    - And the retries have exhausted to its limits |
| Then the 408 (Request Timeout) exception is converted to a 503 (Service Unavailable) exception |

| Data plan in Gateway mode for Reads, Not Found |
|----|
| Given a CX wants to read an item that does not exist |
| When I configure my _CosmosClient.ConnectionMode_ for **Gateway** mode (which uses an HTTP protocol) |
|    - And it will send the operational request using _SendHttpHelperAsync_ |
| Then the 404 (Not Found) exception is thrown |

| Data plan in Gateway mode for Writes, Request Timeout |
|----|
| Given a CX wants to write a new item (should test all of the following: CREATE, DELETE, REPLACE, UPSERT, PATCH, etc.) |
| When I configure my _CosmosClient.ConnectionMode_ for **Gateway** mode (which uses an HTTP protocol) |
|    - And it will send the operational request using _SendHttpHelperAsync_ |
|    - And the retries will throw a 408 (Request Timeout) exception |
|    - And the retries have exhausted to its limits |
| Then the 408 (Request Timeout) exception is converted to a 503 (Service Unavailable) exception |

| Data plan in Gateway mode for Writes, Conflict |
|----|
| Given a CX wants to create an item that already exist |
| When I configure my _CosmosClient.ConnectionMode_ for **Gateway** mode (which uses an HTTP protocol) |
|    - And it will send the operational request using _SendHttpHelperAsync_ |
| Then the 409 (Not Found) exception is thrown |

| Data plan in Gateway mode for Reads, Request Timeout |
|----|
| Given a CX wants to read an existing item |
| When I configure my _CosmosClient.ConnectionMode_ for **Gateway** mode (which uses an HTTP protocol) |
|    - And it will send the operational request using _SendHttpHelperAsync_ |
|    - And the retries will throw a 408 (Request Timeout) exception |
|    - And the retries have exhausted to its limits |
| Then the 408 (Request Timeout) exception is converted to a 503 (Service Unavailable) exception |

## Clarifying questions
- Do we need to be concerned with sub status codes?
- Just need to clarify that we are not including a solution for Direct Mode.
  - https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3467
- How to exhaust all retries so that it returns a 408 (Request Timeout) exception?
- What is the best means for testing a Metadata Read operation request?
- What kind of exceptions, other than a 408 (Request Timeout), could occur with a Metadata Read?
- What is the best means for testing a Query Plan operation request?
- What kind of exceptions, other than a 408 (Request Timeout), could occur with a Query Plan?
- Should we also be testing this for Change Feed iterator/estimator?

This would mean adapting the HttpTimeoutPolicy interface to know which cases should throw the 503 or which should throw the normal timeout: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/1eaae654845f83faf3283ee3be53d8d2c31e5256/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs#L295

NaluTripician:
# Pull Request Template

## Description

Currently, the SDKs don't factor HTTP timeouts in their cross-regional retry policies. TCP timeouts are covered but HTTP timeouts are only retried locally and then surface to the user.

Because HTTP timeouts do not have any cross-region failover semantics, SDKs could not attempt to recover by rerouting requests to other available regions. Some HTTP operations can be performed on other regions and therefore should be retried.

Here is a list of the different HTTP operations that the SDKs preforms. 

| Type | Read/Write | Multi Master | Can be retried? |
|----------|------------|--------------|-----------------|
| Data plane |  Read | No | Yes |
| Data plane |  Write | No | No |
| Data plane |  Read | Yes | Yes |
| Data plane |  Write | Yes | Yes |
| Metadata |  Read | No | Yes |
| Metadata |  Write | No | No |
| Metadata |  Read | Yes | Yes |
| Metadata |  Write | Yes | No |
|  Query Plan | Read | No | Yes |
|  Query Plan | Read | Yes | Yes |
|  Addresses | Read | No | No |
|  Addresses | Read | Yes | No |

So, out of these operations the ones that are retriable are the following. 

| Type | Read/Write | 
|----------|------------|
| Data plane |  Read |
| Data plane |  Write |
| Metadata |  Read | |
|  Query Plan | Read (it is always read) | 

Looking at the current flow we can see that each HTTP operation is assigned a HttpTimeoutPolicy which dictates its behavior on a timeout. So we need a way to tell the timeout policy if it is one of these types of HTTP operations that can be retried with a 503 status code it should. 

If we follow the same pattern that we have in the Direct mode (retry locally, if retries are exhausted, throw 503), if the HttpTimeoutPolicy converted the HTTP timeout to 503, then it would kickoff the retry on another region. Because each HttpTimeoutPolicy is aware of the type of operation upon initiation, then it can conditionally decide whether to make the timeout a 503 or not if one were to occur.

# HTTP Timeout Policies #

There are 4 different types of HTTP Timeout Policies so we need to look at what they are used for and see if they need to change. 

## NoRetry ##
This policy is used for HTTP requests that should not be retried. Because it should never be retried it does not need to change. 

## ControlPlaneRead ##

Here because all control plane read requests are meta data read requests, we should make sure that all Control Plane Read timeout policies should throw a 503 on timeout. 

## ControlPlaneRetriableHotPath ##

Control Plane Retriable Hot Path policies are used in three different places. 

1. Partition Key Requests - which should not be retried. 
2. Address Requests - which should not be retried. 
2. Query Plan Requests - which should be retried.

## Default ##

All other cases use the Default HTTP Timeout Policy, including all data plane requests and some - but not all - meta data read/write requests. 
So of the 4 timeout policies that exist three need to have the ability to communicate that they should throw 503 errors on timeout. ControlPlaneRead, where it should always do that, and ControlPlaneRetriableHotPath + Default where we need to be flexible on when it can throw a 503 error on timeout.  

### ControlPlaneRead ###
1. Tests
2. GatewyaStoreClient
3. GatewayAccountReader

All of these methods should retry on a timeout. 

### ControlPlaneRetriableHotPath ###
1. Tests
2. GatewayAddressCache
3. HttpTimeoutPolcy `GetTimeoutPolicy` method. 

The tests the cases are mixed what ones should be retried. In the GatewayAddressCache they are address operations and therefore should not be retried. In the `GetTimeoutPolicy` method, only the case for the Query Plan operations should be able to retry. 

### Default ### 
1. Tests
2. HttpTimeoutPolcy `GetTimeoutPolicy` method. 

In the tests cases are mixed on which ones should be retired. In the `GetTimeoutPolicy` method, only meta data reads and data plane operations should be retried. 

So looking at where each of these timeout polices `GetTimeoutPolicy` method is the main method where it is decided which timeout policy is used. Right now it is really only deciding between the Default and ControlPlaneRetriableHotPath. The other policies are specified directly in the methods where the HTTP requests is made.  

# Proposed Solution #
Have two different instances  of each client in the construer with a flag that tells whether it should throw a 503 error on timeout or not. Then in the `GetTimeoutPolicy` method use the document service request to determine where it falls under the umbrella of the scenarios where we want to throw the 503 errors. By setting the flag on assignment of the policy, we can access this information anywhere in the flow of the HTTP gateway request.

So the new flow would be this: 

```mermaid
sequenceDiagram
    participant A as Operation
    participant B as ClientRetryPolicy <br> [Cross region retry]
    participant C as TransportHandler
    participant D as HttpTimeoutPolicy
    participant E as CosmosHttpClientCore
    A->>B: 1. Send HTTP operation
    B->>C: 
    C->>D: Wraps the call around retry layer
    D->>E: 
    E-->>D: OperationCanceledException
    D-->>E: Retries HTTP timeouts on same endpoint
    D-->>D: Repeats until retries exhausted
    D->>C: OperationCanceledException
    D-->>E: ShouldThrow503OnTimeout
    E-->>D: ServiceUnavailableException
    D->>C: ServiceUnavailableException
    C->>B: CosmosException(408)
    C->>B: CosmosException(503)
    B->>A: 
```

## Breaking Change Possibility
Note that changing the exception from a 408 to a 503 could have the possibility of being a breaking change. In order for this not to be a breaking change, one of two things can be done. 

1. Document this change in behavior in the changelogs
2. Make extra code changes to avoid this, however it these changes would be convoluted and mix the line of what each component should be doing. 

## Type of change

## Closing issues

To automatically close an issue: closes #3497


Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -296,6 +297,12 @@ ValueTask<HttpRequestMessage> CreateRequestMessage()
                                     string message =
                                             $"GatewayStoreClient Request Timeout. Start Time UTC:{startDateTimeUtc}; Total Duration:{(DateTime.UtcNow - startDateTimeUtc).TotalMilliseconds} Ms; Request Timeout {requestTimeout.TotalMilliseconds} Ms; Http Client Timeout:{this.httpClient.Timeout.TotalMilliseconds} Ms; Activity id: {System.Diagnostics.Trace.CorrelationManager.ActivityId};";
                                     e.Data.Add("Message", message);
+                                    
+                                    if (timeoutPolicy.ShouldThrow503OnTimeout)
+                                    {
+                                        throw new ServiceUnavailableException(message, e, SubStatusCodes.TransportGenerated503);
```
ealsur, line 303: Use `CosmosExceptionFactory` instead.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -18,21 +18,51 @@ internal abstract class HttpTimeoutPolicy
...
             }
 
+            //Partition Key Requests
             if (documentServiceRequest.ResourceType == ResourceType.PartitionKeyRange)
             {
                 return HttpTimeoutPolicyControlPlaneRetriableHotPath.Instance;
             }
 
+            //Data Plane Read
+            if (!IsMetaData(documentServiceRequest) && documentServiceRequest.IsReadOnlyRequest)
```
ealsur, line 40: nit: Prefix static methods with the class name (`MyClass.MyStaticMethod()`)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -259,6 +260,87 @@ async Task<HttpResponseMessage> sendFunc(HttpRequestMessage request, Cancellatio
...
+                    new ValueTask<HttpRequestMessage>(
+                        result: new HttpRequestMessage(HttpMethod.Get, new Uri("http://localhost"))),
+                        resourceType: ResourceType.Document,
+                        timeoutPolicy: HttpTimeoutPolicyDefault.InstanceShouldThrow503OnTimeout,
+                        clientSideRequestStatistics: new ClientSideRequestStatisticsTraceDatum(DateTime.UtcNow, new TraceSummary()),
+                        cancellationToken: default);
+            }
+            catch (Exception e)
+            {
+                Assert.IsInstanceOfType(e, typeof(ServiceUnavailableException));
```
ealsur, line 291: Can you verify there were retries? For example, check the number of calls to the sendFunc?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayAccountReaderTests.cs
```diff
@@ -61,7 +61,7 @@ public async Task DocumentClient_BuildHttpClientFactory_WithHandler()
                 uri: new Uri("https://localhost"),
                 additionalHeaders: new RequestNameValueCollection(),
                 resourceType: ResourceType.Document,
-                timeoutPolicy: HttpTimeoutPolicyDefault.Instance,
```
ealsur, line 64: Are these changes needed for the test to pass?

ealsur (commented):
Can you do a manual E2E test with an account with 2 regions, client on Gateway mode, created with PreferredRegions (both regions), add a custom MessageHandler and throw the OperationCanceled when the url is for the first region, and do an Item Read and see if the request retried on the other region?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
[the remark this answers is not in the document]
NaluTripician: There is a `!` in front of the `documentServiceRequest.IsReadOnlyRequest` for the write request.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
```diff
@@ -808,7 +808,7 @@ public async Task GatewayStatsDurationTest()
 
             await cosmosHttpClient.SendHttpAsync(() => new ValueTask<HttpRequestMessage>(new HttpRequestMessage(HttpMethod.Get, "http://someuri.com")),
                                                   ResourceType.Document,
-                                                  HttpTimeoutPolicyDefault.Instance,
```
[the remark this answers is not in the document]
NaluTripician: No, but considering that this looks like it is a Data plane read request I thought it might be a good idea to change it to the should throw instance. But if you think it should stay the way it was I can revert the changes.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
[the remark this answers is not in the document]
ealsur: They are using the same policy in both cases, that's what I meant :)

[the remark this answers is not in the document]
NaluTripician: Ahhh I see, yes I think I can merge them into one

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -296,6 +298,20 @@ ValueTask<HttpRequestMessage> CreateRequestMessage()
                                     string message =
                                             $"GatewayStoreClient Request Timeout. Start Time UTC:{startDateTimeUtc}; Total Duration:{(DateTime.UtcNow - startDateTimeUtc).TotalMilliseconds} Ms; Request Timeout {requestTimeout.TotalMilliseconds} Ms; Http Client Timeout:{this.httpClient.Timeout.TotalMilliseconds} Ms; Activity id: {System.Diagnostics.Trace.CorrelationManager.ActivityId};";
                                     e.Data.Add("Message", message);
+                                    
+                                    if (timeoutPolicy.ShouldThrow503OnTimeout)
+                                    {
+                                        CosmosException serviceUnavailableExcpetion = CosmosExceptionFactory.CreateServiceUnavailable(
```
ealsur, line 304: You are throwing the `ServiceUnavailableException` at the end, it should be the `CosmosException`:

```suggestion
                                        throw new CosmosExceptionFactory.CreateServiceUnavailable(
```

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -18,21 +18,45 @@ internal abstract class HttpTimeoutPolicy
...
+            //Meta Data Read
+            if (StaticIsMetaData(documentServiceRequest) && documentServiceRequest.IsReadOnlyRequest)
+            {
+                return HttpTimeoutPolicyDefault.InstanceShouldThrow503OnTimeout;
+            }
+
+            //Default behavior
             return HttpTimeoutPolicyDefault.Instance;
+
+            static bool StaticIsMetaData(DocumentServiceRequest request)
```
ealsur, line 54: IsMetaData is ok as a name

```suggestion
            static bool IsMetaData(DocumentServiceRequest request)
```

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -18,21 +18,45 @@ internal abstract class HttpTimeoutPolicy
...
             }
 
+            //Data Plane Read & Write
+            if (!StaticIsMetaData(documentServiceRequest))
+            {
+                return HttpTimeoutPolicyDefault.InstanceShouldThrow503OnTimeout;
+            }
+
+            //Meta Data Read
+            if (StaticIsMetaData(documentServiceRequest) && documentServiceRequest.IsReadOnlyRequest)
```
ealsur, line 46: Prefix with class name for static methods

```suggestion
            if (HttpTimeoutPolicy.IsMetaData(documentServiceRequest) && documentServiceRequest.IsReadOnlyRequest)
```

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -18,21 +18,45 @@ internal abstract class HttpTimeoutPolicy
...
             }
 
+            //Partition Key Requests
             if (documentServiceRequest.ResourceType == ResourceType.PartitionKeyRange)
             {
                 return HttpTimeoutPolicyControlPlaneRetriableHotPath.Instance;
             }
 
+            //Data Plane Read & Write
+            if (!StaticIsMetaData(documentServiceRequest))
```
ealsur, line 40: Prefix with class name for static methods

```suggestion
            if (!HttpTimeoutPolicy.IsMetaData(documentServiceRequest))
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -16,6 +16,7 @@ namespace Microsoft.Azure.Cosmos.Tests
     using System.Collections.Generic;
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Cosmos.Tracing.TraceData;
+    using FluentAssertions;
```
ealsur, line 19: Is this using needed?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -259,6 +260,113 @@ async Task<HttpResponseMessage> sendFunc(HttpRequestMessage request, Cancellatio
...
+                        timeoutPolicy: HttpTimeoutPolicyDefault.InstanceShouldThrow503OnTimeout,
+                        clientSideRequestStatistics: new ClientSideRequestStatisticsTraceDatum(DateTime.UtcNow, new TraceSummary()),
+                        cancellationToken: default);
+            }
+            catch (Exception e)
+            {
+                Assert.AreEqual(3, count, "Should retry 3 times");
+                Assert.IsInstanceOfType(e, typeof(ServiceUnavailableException));
+            }
+            count = 0;
```
ealsur, line 296: To avoid code duplication you could create a single private method that receives by parameter the resource type, policy, expected retries and expected error type. The method can have the handler definition inside (no need to reset count).

And you just invoke the method on multiple combinations of resource type, policy and expectations.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -296,6 +298,20 @@ ValueTask<HttpRequestMessage> CreateRequestMessage()
...
+                                    {
+                                        throw new CosmosExceptionFactory.CreateServiceUnavailable(
+                                            message: message,
+                                            headers: new Headers()
+                                            {
+                                                ActivityId = System.Diagnostics.Trace.CorrelationManager.ActivityId.ToString()
+                                            },
+                                            innerException: e);
+                                        
+                                        throw new ServiceUnavailableException(message, e, SubStatusCodes.TransportGenerated503);
```
ealsur, line 312: This line is probably generating a build error

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosExceptionFactory.cs
```diff
@@ -283,6 +283,24 @@ internal static class CosmosExceptionFactory
                 innerException);
         }
 
+        internal static CosmosException CreateServiceUnavailable(
+            string message, 
+            Headers headers,
```
ealsur, line 288: Should we receive the substatus code here?

ealsur (commented):
Please update the PR with the results of manual validation

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -16,6 +16,8 @@ namespace Microsoft.Azure.Cosmos.Tests
     using System.Collections.Generic;
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Cosmos.Tracing.TraceData;
+    using System.Drawing.Text;
```
ealsur, line 19: nit: check usings, are these needed?

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Azure.Cosmos
     using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
+    using System.Runtime.CompilerServices;
```
ealsur, line 13: Is this using needed?

ealsur (approved):
Worth bringing it up with the team to confirm calling the behavior change on the changelog is good enough based on Fabian's comments

PR #3562, opened 2022-11-11
AI Integration: Adds a new flag in DistributedTracingOptions

sourabh1007:
## Description
As part of this PR, instead of depending on the _EventSourceListener_ configuration for type of EventLevel it want to listen, making it flag based. 

**Why this change?**
Today AppInsights read only Informational level events and there is no way for customer to change/configure it. So providing a flag at source i.e. cosmos sdk to control it.

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Tracing/EndToEndTraceWriterBaselineTests.cs
```diff
@@ -1272,7 +1271,7 @@ public async Task BulkOperationsAsync()
                 throttleClient.ClientOptions.EnableDistributedTracing = true;
                 throttleClient.ClientOptions.DistributedTracingOptions = new DistributedTracingOptions()
                 {
-                    DiagnosticsLatencyThreshold = TimeSpan.FromMilliseconds(DiagnosticsLatencyThresholdValue)
+                    EnableDiagnosticsTraceForAllEvents = true
```
ealsur, line 1274: How is this related? The previous code was setting a Latency threshold, and this is enabling traces, but there is no threshold now?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/CosmosDbEventSource.cs
```diff
@@ -29,7 +29,7 @@ public static bool IsEnabled(EventLevel level)
         public static void RecordDiagnosticsForRequests(DistributedTracingOptions config,
             OpenTelemetryAttributes response)
         {
-            if (CosmosDbEventSource.IsEnabled(EventLevel.Informational))
+            if (config.EnableDiagnosticsTraceForAllEvents)
```
ealsur, line 32: The flag is called "All" would that mean other EventLevels should also be logged?
What if the flag is fals but the Info level is enabled on the Event Source?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Tracing/EndToEndTraceWriterBaselineTests.cs
sourabh1007: previously I was setting latency threshold very low so that it would capture request diagnostics for all the events. now we have a flag for that. So removed that config and using this flag instead.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/CosmosDbEventSource.cs
sourabh1007: if info level is enables at event listener then, it won't honor it...and will still generate events with warning and error level based on latency threshold Configuration.
I think I should rename to _EnableDiagnosticsTraceForAllRequests_ it will make more sense.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/DistributedTracingOptions.cs
```diff
@@ -28,5 +28,11 @@ internal sealed class DistributedTracingOptions
         /// </summary>
         public TimeSpan? DiagnosticsLatencyThreshold { get; set; }
 
+        /// <summary>
+        /// Set this flag as true if you want to generate (<see cref="System.Diagnostics.Tracing.EventSource"/>) containing request diagnostics string for all the operations.
+        /// If this flag is true then, it won't honour <see cref="DiagnosticsLatencyThreshold"/> value to generate diagnostic traces.
```
ealsur, line 33: Can we add a validation to check that both settings are not set?

ealsur (dismissed):
After https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3563, this now becomes public API contract, please get public API signoff

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/DiagnosticsFilterHelperTest.cs
```diff
@@ -38,7 +38,7 @@ public void CheckReturnFalseOnSuccessAndLowerLatencyThanConfiguredConfig()
 
             DistributedTracingOptions distributedTracingOptions = new DistributedTracingOptions
             {
-                DiagnosticsLatencyThreshold = TimeSpan.FromMilliseconds(20)
+                DiagnosticsLatencyThreshold = this.rootTrace.Duration.Add(TimeSpan.FromSeconds(1))
```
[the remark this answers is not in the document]
sourabh1007: it was flaky test. hence fixing it here itself as it is related and minor change.

ealsur (approved):

ealsur (approved):

PR #3574, opened 2022-11-18
Query: Fixes incorrect FeedResponse.Count when result contains undefined elements

neildsh:
## Description

Do not maintain an independent count on QueryResponse that can go out of sync with the actual resource count.

## Type of change

## Closing issues
Fixes [Incident 349270178](https://portal.microsofticm.com/imp/v3/incidents/details/349270178/home) : Identified bug in Microsoft.Azure.Cosmos 3.31.2 feedResponse.Count


neildsh:
This fix is actually more involved and needs to address the stream api as well

sboshra (approved):
:shipit:

PR #3576, opened 2022-11-18
AI Integration: Adds cosmetic fixes

sourabh1007:
## Description

This PR contains below changes:
1. Coverting StatusCode and SubstatusCode as Int from string.
image

2. Renamed _hashed_machine_id_ to _machine_id_
3. Remove "information not available" message and set null instead.

image

7. On GatewayMode create "Internal" type of Activity otherwise "client" type pf Activity. (Verified by @lmolkova )

## Type of change


Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -92,8 +89,8 @@ public void Record(string key, string value)
                 this.scope.AddAttribute(OpenTelemetryAttributeKeys.NetPeerName, clientContext.Client?.Endpoint?.Host);
 
                 // Client Information
-                this.scope.AddAttribute(OpenTelemetryAttributeKeys.ClientId, clientContext?.Client?.Id ?? OpenTelemetryAttributes.NotAvailable);
-                this.scope.AddAttribute(OpenTelemetryAttributeKeys.UserAgent, clientContext.UserAgent ?? OpenTelemetryAttributes.NotAvailable);
+                this.scope.AddAttribute(OpenTelemetryAttributeKeys.ClientId, clientContext?.Client?.Id);
```
ealsur, line 92: Would this then add the attribute with a `null` value now? Should the attribute be there with null or not be there at all?

sourabh1007: I will confirm it after testing it with appinsight but as far as I remember, if null it won't appear in appinsight.

sourabh1007: in the above screenshot, you can see I don't have any property which has "null" value e.g. db.cosmosdb.request_content_length_bytes, db.cosmosdb.response_content_length_bytes

kirankumarkolli (approved):

kirankumarkolli (approved):

PR #3598, opened 2022-12-07
[Preview] AI integration: Adds IsDistributedTracingEnabled flag as public API to enable/disable this feature

sourabh1007:
## Description

1. Made Distributed tracing flag, public, for preview package.
2. Renamed `DiagnosticsLatencyThreshold `to `LatencyThresholdForDiagnosticEvent`
3. Renamed `EnableDistributedTracing` to `IsDistributedTracingEnabled` and make is always true.
4. Added documentation for all public APIs.
5. Added null check for scopefactory instance

### Why DistributedTracing is always enabled?
There are 2 ways to get these activities i.e AI SDK and Open Telemetry. 
OpenTelemetry users don't need this flag as they generally use Source Name to listen any particular kind of activities and AI SDK users cannot enable/disable these activities from the AI SDK, that's why this flag is there to stop creating activities at source itself.

### Why Open telemetry Support is Experimental?
OTel conventions are experimental, OTel support in Azure SDKs is experimental.

## Type of change

## Perf Impact

MicrosoftTeams-image
MicrosoftTeams-image (1)


Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -992,13 +992,25 @@ public override bool CanConvert(Type objectType)
...
+        /// By default, SDK (Source Name is "Azure.Cosmos.Operation") generates operation level <see cref="System.Diagnostics.Activity"/> for methods calls.
+        /// This flag is there to disable it from source.
+        /// </summary>
+        /// <remarks>Refer https://opentelemetry.io/docs/instrumentation/net/exporters/ to know more about open telemetry exporters</remarks>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+            bool DisableDistributedTracing { get; set; }
```
kirankumarkolli, line 1013: Naming wise, other option is EnableDistributedTracing and default to TRUE. 
Both are logically correct, IMO Enable sounds better :-)

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
```diff
@@ -45,8 +45,13 @@ public class RequestOptions
         /// <summary>
         /// Set Request Level Distributed Tracing Options.
         /// </summary>
-        internal DistributedTracingOptions DistributedTracingOptions { get; set; }
-        
+#if PREVIEW
```
kirankumarkolli, line 48: Is this also recommended by central team? Or our choice?
We can late bind this API if we are still unsure of customization scenarios.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/DistributedTracingOptions.cs
```diff
@@ -23,9 +27,9 @@ internal sealed class DistributedTracingOptions
...
 
         /// <summary>
-        /// Latency Threshold to generate (<see cref="System.Diagnostics.Tracing.EventSource"/>) with Request diagnostics in distributing Tracing.<br></br>
-        /// If it is not set then by default it will generate (<see cref="System.Diagnostics.Tracing.EventSource"/>) for query operation which are taking more than 500 ms and non-query operations taking more than 100 ms.
+        /// SDK generates <see cref="System.Diagnostics.Tracing.EventSource"/> (Event Source Name is "Azure-Cosmos-Operation-Request-Diagnostics") with Request Diagnostics String, If Operation level distributed tracing is not disabled i.e. <see cref="Microsoft.Azure.Cosmos.CosmosClientOptions.DisableDistributedTracing"/>
         /// </summary>
-        public TimeSpan? DiagnosticsLatencyThreshold { get; set; }
+        /// <remarks>If it is not set then, by default, it will generate <see cref="System.Diagnostics.Tracing.EventSource"/> for query operation which are taking more than 500 ms and non-query operations taking more than 100 ms.</remarks>
+        public TimeSpan? LatencyThresholdForRequestDiagnosticEventTrace { get; set; }
```
kirankumarkolli, line 33: LatencyThreshold was used as suffix before, is moving it to Prefix intentional?

sourabh1007: Yes, trying to make it more align to what it is doing. This Latency threshold value is for "Request Diagnostics Trace Event". Previous name can be easily confused with, if we are doing filtering at activity level.

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
sourabh1007: I don't think, Central team has any recommendation on this. @lmolkova can you correct me here?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: I am ok with anything here. @lmolkova can you please confirm if central SDK has any recommendation on this flag?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/DistributedTracingOptions.cs
lmolkova:
```suggestion
        public TimeSpan? LatencyThresholdForDiagnosticEvent { get; set; }
```

or 

```suggestion
        public TimeSpan? LatencyThresholdForRequestDiagnosticEvent { get; set; }
```

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
sourabh1007: JFYI: Right now, it just has one configuration using which customer can control generation of event with request diagnostic string and yes if required in future we can add more such configuration like you suggested above.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
lmolkova: We call it [`IsDistributedTracingEnabled`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/DiagnosticsOptions.cs#L74) and it defaults to true. Would be nice to have consistent naming.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -992,13 +992,23 @@ public override bool CanConvert(Type objectType)
...
-        internal bool EnableDistributedTracing { get; set; }
+        /// <remarks>Refer https://opentelemetry.io/docs/instrumentation/net/exporters/ to know more about open telemetry exporters</remarks>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+            bool IsDistributedTracingEnabled { get; set; }
+#if PREVIEW
+        = true;
```
kirankumarkolli, line 1010: What's the performance impact on team which upgrade preview upgrade with default ON?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,32 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
...
         /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
-        internal CosmosClientBuilder WithDistributingTracing(DistributedTracingOptions options)
+#if PREVIEW
+        public
+#else
+        internal
+#endif 
+            CosmosClientBuilder DisableDistributedTracing()
         {
+            this.clientOptions.IsDistributedTracingEnabled = false;
```
kirankumarkolli, line 437: Conditional on preview?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: In case of Open telemetry/AI Integration support, perf is totally dependent on the listener. As a source, there will be no perf impact at all. Still, I am expecting a new customer to go through the complete software lifecycle, in order to make sure if this new feature is not impacting them in any way.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -992,13 +992,23 @@ public override bool CanConvert(Type objectType)
...
         /// Distributed Tracing Options. <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/>
         /// </summary>
+        /// <remarks> Applicable only when Operation level distributed tracing is not disabled i.e. <see cref="Microsoft.Azure.Cosmos.CosmosClientOptions.IsDistributedTracingEnabled"/></remarks>
         internal DistributedTracingOptions DistributedTracingOptions { get; set; }
 
         /// <summary>
-        /// Gets or sets value indicating whether distributed tracing activities (<see cref="System.Diagnostics.Activity"/>) are going to be created for the SDK methods calls and HTTP calls.
-        /// By default true for Preview package
+        /// By default, SDK (Source Name is "Azure.Cosmos.Operation") generates operation level <see cref="System.Diagnostics.Activity"/> for methods calls.
+        /// This flag is there to disable it from source.
```
ealsur, line 1000: nit: Comment is redundant, the property is called IsXXXEnabled and it's a `bool`, disabled (false) or enabled (true) is implied. What needs to be properly communicated is what is the default.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,32 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
         }
 
         /// <summary>
-        /// If Open Telemetry listener is subscribed for Azure.Cosmos namespace, There are <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/> you can leverage to control it.<br></br>
+        /// Disables Operation level activity generation even if listener is subscribed to Azure.Cosmos.operation source and clears <see cref="DistributedTracingOptions"/> configured for this client instance.
```
ealsur, line 427: For consistency with the other property, source name should be the same?

```suggestion
        /// Disables Operation level activity generation even if listener is subscribed to "Azure.Cosmos.Operation" source.
```

Also, why do we mention that it clears the options? This is the client builder, how does the user set the DistributedTracingOptions and call Disable after?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,32 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
...
+
+            return this;
+        }
+        
+        /// <summary>
+        /// Sets Distributed Tracing Configuration ref. <see cref="DistributedTracingOptions"/>
+        /// </summary>
+        /// <param name="options"><see cref="DistributedTracingOptions"/>.</param>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
+        /// <exception cref="ArgumentException">If Distributed Tracing is disabled</exception>
```
ealsur, line 448: This exception comment is not correct, is it? Unless it is possible for the user to Disable and set the options too after?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,32 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
...
+        /// <remarks>Refer https://opentelemetry.io/docs/instrumentation/net/exporters/ to know more about open telemetry exporters</remarks>
+        internal CosmosClientBuilder EnableDistributedTracingWithOptions(DistributedTracingOptions options = default)
```
ealsur, line 450: Shouldn't this be public for preview?

ealsur: Or the PR is only for the Enabled flag and not the options?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/DistributedTracingOptions.cs
```diff
@@ -23,9 +22,9 @@ internal sealed class DistributedTracingOptions
         internal static readonly TimeSpan DefaultQueryTimeoutThreshold = TimeSpan.FromMilliseconds(500);
 
         /// <summary>
-        /// Latency Threshold to generate (<see cref="System.Diagnostics.Tracing.EventSource"/>) with Request diagnostics in distributing Tracing.<br></br>
-        /// If it is not set then by default it will generate (<see cref="System.Diagnostics.Tracing.EventSource"/>) for query operation which are taking more than 500 ms and non-query operations taking more than 100 ms.
+        /// SDK generates <see cref="System.Diagnostics.Tracing.EventSource"/> (Event Source Name is "Azure-Cosmos-Operation-Request-Diagnostics") with Request Diagnostics String, If Operation level distributed tracing is not disabled i.e. <see cref="Microsoft.Azure.Cosmos.CosmosClientOptions.IsDistributedTracingEnabled"/>
```
ealsur, line 25: Is "Azure-Cosmos-Operation-Request-Diagnostics" still valid as the name?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: Using `<value>` like https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs#L199

ealsur (changes requested):
Please address the public API documentation comments and make sure the PR title follows the pattern

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
sourabh1007: DistributedTracingOptions is internal as of now. No need to mention it is the doc yet.
Right now, this feature in on by default and builder has API to disable it.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
sourabh1007: We decided to go step my step hence exposing the tracing flag right now.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,31 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
         }
 
         /// <summary>
-        /// If Open Telemetry listener is subscribed for Azure.Cosmos namespace, There are <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/> you can leverage to control it.<br></br>
+        /// Disables Operation level activity generation even if listener is subscribed to "Azure.Cosmos.Operation" source.
```
kirankumarkolli, line 427: 'if listener is subscribed to "Azure.Cosmos.Operation" source" 
Is enable dependent on subscription?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,31 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
...
         /// </summary>
-        /// <param name="options">Tracing Options <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/></param>
         /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
-        internal CosmosClientBuilder WithDistributingTracing(DistributedTracingOptions options)
+#if PREVIEW
+        public
+#else
+        internal
+#endif 
+            CosmosClientBuilder DisableDistributedTracing()
```
kirankumarkolli, line 435: For boolean types we are using With<operation(bool). 
Please check WithLimitToEndpoint(bool )

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -992,13 +992,26 @@ public override bool CanConvert(Type objectType)
...
```
ealsur, line 1000: Remove this line.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: Keep in mind, Teams (Microsoft Teams), uses preview packages.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -992,13 +992,26 @@ public override bool CanConvert(Type objectType)
         /// <summary>
         /// Distributed Tracing Options. <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/>
         /// </summary>
+        /// <remarks> Applicable only when Operation level distributed tracing is not disabled i.e. <see cref="Microsoft.Azure.Cosmos.CosmosClientOptions.IsDistributedTracingEnabled"/></remarks>
         internal DistributedTracingOptions DistributedTracingOptions { get; set; }
 
         /// <summary>
-        /// Gets or sets value indicating whether distributed tracing activities (<see cref="System.Diagnostics.Activity"/>) are going to be created for the SDK methods calls and HTTP calls.
-        /// By default true for Preview package
+        /// By default, SDK (Source Name is "Azure.Cosmos.Operation") generates operation level <see cref="System.Diagnostics.Activity"/> for methods calls.
```
ealsur, line 999: Other properties that enable/disable use the "Gets or sets the flag...", what is the default is already explained in the `<value>` tag.

```suggestion
        /// Gets or sets the flag to generate operation level <see cref="System.Diagnostics.Activity"/> for methods calls using the Source Name "Azure.Cosmos.Operation".
```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
sourabh1007: yes, in order to subscribe to the telemetry, customer has to enable is using `isDistributedTracingEnabled` flag (which is by default true) and subscribe to `Azure.Cosmos.Operation` source (AI SDK is already subscribed to this source name).

So, it means, app insight users, really don't have to do anything in order to enable this feature and that's the reason, I added an option to disable it in builder class, in case they don't want to use it.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: If you want to add explanations, use `<remarks>`

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: Teams is another customer for me. If you and @kirankumarkolli are not comfortable with this feature enabled by default. Then we can keep this flag off by default. and let the customer enable it explicitly.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -992,13 +992,25 @@ public override bool CanConvert(Type objectType)
         /// <summary>
         /// Distributed Tracing Options. <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/>
         /// </summary>
+        /// <remarks> Applicable only when Operation level distributed tracing is not disabled i.e. <see cref="Microsoft.Azure.Cosmos.CosmosClientOptions.IsDistributedTracingEnabled"/></remarks>
```
ealsur, line 995: nit: not disabled => enabled for clarity

```suggestion
        /// <remarks> Applicable only when Operation level distributed tracing is enabled through <see cref="Microsoft.Azure.Cosmos.CosmosClientOptions.IsDistributedTracingEnabled"/></remarks>
```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,30 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
         }
 
         /// <summary>
-        /// If Open Telemetry listener is subscribed for Azure.Cosmos namespace, There are <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/> you can leverage to control it.<br></br>
+        /// Disables Operation level activity generation even if listener is subscribed to "Azure.Cosmos.Operation" source.
         /// </summary>
-        /// <param name="options">Tracing Options <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/></param>
+        /// <param name="isEnabled"></param>
```
ealsur, line 429:
```suggestion
        /// <param name="isEnabled">Whether <see cref="CosmosClientOptions.IsDistributedTracingEnabled"/> is enabled.</param>
```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,30 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
         }
 
         /// <summary>
-        /// If Open Telemetry listener is subscribed for Azure.Cosmos namespace, There are <see cref="Microsoft.Azure.Cosmos.DistributedTracingOptions"/> you can leverage to control it.<br></br>
+        /// Disables Operation level activity generation even if listener is subscribed to "Azure.Cosmos.Operation" source.
```
ealsur, line 427: Summary is no longer valid. Maybe something like (feel free to adjust):

```suggestion
        /// Sets whether Distributed Tracing for "Azure.Cosmos.Operation" source is enabled.
```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -424,12 +424,30 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
...
+            return this;
+        }
+        
+        /// <summary>
+        /// Enables Distributed Tracing with a Configuration ref. <see cref="DistributedTracingOptions"/>
+        /// </summary>
+        /// <param name="options"><see cref="DistributedTracingOptions"/>.</param>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>]
+        /// <remarks>Refer https://opentelemetry.io/docs/instrumentation/net/exporters/ to know more about open telemetry exporters</remarks>
+        internal CosmosClientBuilder WithDistributedTracingOptions(DistributedTracingOptions options)
```
ealsur, line 448: nit: This method might probably need to be refactored afterwards when the Options type is public, having 2 with the same name means they can be called twice on the same builder

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -992,13 +992,25 @@ public override bool CanConvert(Type objectType)
...
         internal DistributedTracingOptions DistributedTracingOptions { get; set; }
 
         /// <summary>
-        /// Gets or sets value indicating whether distributed tracing activities (<see cref="System.Diagnostics.Activity"/>) are going to be created for the SDK methods calls and HTTP calls.
-        /// By default true for Preview package
+        /// Gets or sets the flag to generate operation level <see cref="System.Diagnostics.Activity"/> for methods calls using the Source Name "Azure.Cosmos.Operation"..
```
ealsur, line 999:
```suggestion
        /// Gets or sets the flag to generate operation level <see cref="System.Diagnostics.Activity"/> for methods calls using the Source Name "Azure.Cosmos.Operation".
```

ealsur (changes requested):
Few remaining text changes

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
sourabh1007: yes I agree....

ealsur (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Tracing/AssertActivity.cs
```diff
@@ -1,8 +1,10 @@
 ﻿namespace Microsoft.Azure.Cosmos.Tracing
 {
+    using System;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Linq;
+    using Castle.Core.Internal;
```
ealsur, line 7: nit: I don't think you are using Castle.Core?

ealsur (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.json
```diff
@@ -304,6 +304,31 @@
       },
       "NestedTypes": {}
     },
+    "Microsoft.Azure.Cosmos.CosmosClientOptions;System.Object;IsAbstract:False;IsSealed:False;IsInterface:False;IsEnum:False;IsClass:True;IsValueType:False;IsNested:False;IsGenericType:False;IsSerializable:False": {
```
ealsur, line 307: Please get an API approval from @kirankumarkolli

kirankumarkolli (approved):

Issue #3467, opened 2022-09-21
Cross region high availability (Partial Outages)


kirankumarkolli:
Tracking progress based on design and links

- [x] .NET SDK support
- [x] Java SDK support
- [x] Python SDK support
- [x] Node SDK support

ealsur:
## Background

Based on the findings of the recent outage (https://status.azure.com/en-us/status/history/ Azure Cosmos DB - North Europe (Tracking ID 3TPC-DT8))

## What happened

User provided error logs indicate that the outage materialized as HTTP timeouts. The HTTP endpoint was responding to connection attempts but requests were eventually failing with a receive timeouts on the client side.

Currently, SDKs don't factor HTTP timeouts in their cross-regional retry policies. TCP timeouts [are covered](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/docs/SdkDesign.md#direct-mode-retry-layer) but HTTP timeouts surface to the user.

Because HTTP timeouts do not have any cross-region failover semantics, SDKs could not attempt to recover by rerouting requests to other available regions.

## Analysis scope

There are 3 major groups for HTTP requests done by the SDK client:

* Data plane requests: When the SDK is configured in Gateway mode, item operations go through HTTP.
* Metadata requests: Operations for metadata resources (databases, containers, not item operations), regardless of the SDK connection mode, go through HTTP
* TCP Address: For SDKs on Direct mode, the SDK [obtains the TCP addresses](https://learn.microsoft.com/en-us/azure/cosmos-db/sql/sql-sdk-connection-modes#routing) before performing data plane operations using HTTP.
* Query Plan: When [ServiceInterop.DLL cannot be used](https://learn.microsoft.com/en-us/azure/cosmos-db/sql/performance-tips-query-sdk?tabs=v3&pivots=programming-language-csharp#use-local-query-plan-generation) on cross-partition queries, the SDK needs to resolve the Query Plan through HTTP.

All these operations were impacted during the outage with HTTP timeouts.

Based on the [SDK design documentation](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/docs/SdkDesign.md#transport) we can scope the HTTP interactions as the following diagram:

```mermaid
flowchart
    RetryHandler[Cross region retries] 
    RetryHandler[Cross region retries] <--> OtherHandlers[[Other handlers]]
    OtherHandlers <--> TransportClient
    TransportClient <--> GatewayStoreClient
[...]

PR #3607, opened 2022-12-15
[Internal] HttpTimeoutPolicy: Removes Data Plane Writes from being able to be retried

NaluTripician:
# Pull Request Template

## Description

Ref: #3467  [PR #3555](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3555) 

In .NET we cannot differentiate between connection and received timeouts, therefore it is unsafe to allow data plane writes to be retried.

## Type of change


ealsur (approved):
Please update the title to follow the pattern we use. Because this is a fix over a non-released feature, then we probably can use `[Internal]...`

PR #3609, opened 2022-12-17
Add support for decimal type as part of Number64 to handle Queryable LINQ Translation

AmriHS:
## Description
Adding support for decimal type as part of Number64 so it doesn't have to go through double conversion. Double conversion causes decimal precision loss in some cases when translating from LINQ to SQL.
The following are example of decimal with high precision:
- 104.37644171779141
- 0.0000000000000000001

## Type of change

## Closing issues

To automatically close an issue: closes #3609


ealsur:
@adityasa Can you review this proposed change?

Microsoft.Azure.Cosmos/src/Number64.cs
```diff
@@ -103,17 +94,16 @@ public string ToString(IFormatProvider formatProvider)
...
         public string ToString(string format, IFormatProvider formatProvider)
         {
-            string toString;
             if (this.IsDouble)
             {
-                toString = Number64.ToDouble(this).ToString(format, formatProvider);
+                return Number64.ToDouble(this).ToString(format, formatProvider);
             }
-            else
+            else if (this.IsDecimal)
```
adityasa, line 101: >else [](http://example.com/codeflow?start=12&length=4)

Nit, no need for else.

Microsoft.Azure.Cosmos/src/Number64.cs
```diff
@@ -42,49 +42,40 @@ namespace Microsoft.Azure.Cosmos
         /// </summary>
         private readonly long? longValue;
 
+        /// <summary>
+        /// The decimal if the value is a decimal.
+        /// </summary>
+        private readonly decimal? decimalValue;
+        private Number64(decimal value)
```
adityasa, line 49: A newline before would be nice.

Microsoft.Azure.Cosmos/src/Number64.cs
```diff
@@ -202,36 +192,57 @@ public string ToString(string format, IFormatProvider formatProvider)
...
         {
-            long value;
-            if (number64.IsInteger)
+            if (number64.IsDouble)
             {
-                value = number64.longValue.Value;
+                return (long)number64.doubleValue.Value;
             }
-            else
+            else if (number64.IsDecimal)
```
adityasa, line 212: >else [](http://example.com/codeflow?start=12&length=4)

Nit, no need for else.

Microsoft.Azure.Cosmos/src/Number64.cs
```diff
@@ -202,36 +192,57 @@ public string ToString(string format, IFormatProvider formatProvider)
...
         }
 
         public static double ToDouble(Number64 number64)
         {
-            double value;
+            if (number64.IsDecimal)
+            {
+                return (double)number64.decimalValue.Value;
+            }
+            else if (number64.IsInteger)
```
adityasa, line 226: >else [](http://example.com/codeflow?start=12&length=4)

Same as before.

Microsoft.Azure.Cosmos/src/Number64.cs
```diff
@@ -202,36 +192,57 @@ public string ToString(string format, IFormatProvider formatProvider)
...
+
+        public static decimal ToDecimal(Number64 number64)
+        {
             if (number64.IsDouble)
             {
-                value = number64.doubleValue.Value;
+                return (decimal)number64.doubleValue.Value;
             }
-            else
+            else if (number64.IsInteger)
```
adityasa, line 240: >else [](http://example.com/codeflow?start=12&length=4)

Same.

adityasa:
>                 comparison = first.CompareTo(second);

Does this work for Decimal numbers correctly?

---
Refers to: Microsoft.Azure.Cosmos/src/Number64.cs:312 in d3be84f. [](commit_id = d3be84fb51a5f602fc889a98fe56456f6681ca31, deletion_comment = False)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/CosmosLinqJsonConverterTests.cs
```diff
@@ -30,10 +31,52 @@ public void DateTimeKindIsPreservedTest()
...
 
+        [TestMethod]
+        public void DecimalIsPreservedTest()
+        {
+            Dictionary<decimal, string> decimalTranslationDictionary = new Dictionary<decimal, string>()
+            {
+                { 123m, "123" },
+                { 104.37644171779141m, "104.37644171779141" },
+                { 0.00000000000000000012m, "0.00000000000000000012" },
+                { -0.00000000000000000012m, "-0.00000000000000000012" },
```
adityasa, line 42: Can you also add
Decimal.MinValue and MaxValue?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/CosmosLinqJsonConverterTests.cs
```diff
@@ -30,10 +31,52 @@ public void DateTimeKindIsPreservedTest()
...
+            {
+                Expression<Func<TestDocument, bool>> expr = a => a.DecimalValue == key;
+
+                string sql = SqlTranslator.TranslateExpression(expr.Body);
+                Assert.AreEqual($"(a[\"DecimalValue\"] = {decimalTranslationDictionary[key]})", sql);
+            }
+        }
+
+        [TestMethod]
+        public void DecimalMaxValueIsPreservedTest()
```
adityasa, line 55: >DecimalMaxValueIsPreservedTest [](http://example.com/codeflow?start=20&length=30)

I see this as a dedicated test. Is there a specific reason (that I am missing) why it's a test method by itself?

Microsoft.Azure.Cosmos/src/SqlObjects/Visitors/SqlObjectTextSerializer.cs
```diff
@@ -696,6 +696,10 @@ private static unsafe void WriteNumber64(StringBuilder stringBuilder, Number64 v
                     stringBuilder.Append((char)buffer[i]);
                 }
             }
+            else if (value.IsDecimal)
+            {
+                stringBuilder.Append(value.ToString("G", CultureInfo.InvariantCulture));
```
adityasa, line 701: Curious, is this exercised by one of the tests you added?

adityasa:
>         public int CompareTo(Number64 other)

While I admit that we may be missing existing coverage, this may be a good opportunity to add some for this method, whose implementation correctness seems critical.

---
Refers to: Microsoft.Azure.Cosmos/src/Number64.cs:292 in d3be84f. [](commit_id = d3be84fb51a5f602fc889a98fe56456f6681ca31, deletion_comment = False)

adityasa:
Thank you for taking care of this.

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -1387,14 +1387,7 @@ internal static bool TryGetSqlNumberLiteral(object value, out SqlNumberLiteral s
...
             else if (value is decimal decimalValue)
             {
-                if ((decimalValue >= long.MinValue) && (decimalValue <= long.MaxValue) && (decimalValue % 1 == 0))
-                {
-                    sqlNumberLiteral = SqlNumberLiteral.Create(Convert.ToInt64(decimalValue));
-                }
-                else
-                {
-                    sqlNumberLiteral = SqlNumberLiteral.Create(Convert.ToDouble(decimalValue));
-                }
```
adityasa, line 1397: @neildsh - are you aware of any reasons why this was handled specially (in this manner) here, instead of proposed implementation?

AmriHS:
[[](url)](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/814e72eca593e855838434766596e7bf8945e1b7/Microsoft.Azure.Cosmos/src/Json/JsonSerializer.cs#L340)

@adityasa 
I don't think these changes causing any regression on JsonSerializer. However, this referenced code is not right. It is converting decimal to double as part of JSON serialization. If this also need to handle decimal as part of this PR, please let me know as I have the changes locally. Otherwise, I can create another Bug and raise another PR.

adityasa (approved):
:shipit:

neildsh:
This change is trying to stuff a 128 bit value (`decimal`) into a `Number64` which is, as the name indicates, supposed to be 64 bits wide. This change breaks serialization for `Number64` completely. As such, we cannot take this change.

neildsh (changes requested):
This change is trying to stuff a 128 bit value (`decimal`) into a `Number64` which is, as the name indicates, supposed to be 64 bits wide. This change breaks serialization for `Number64` completely. As such, we cannot take this change.

At this point I am not sure if we actually need a type that is 128 bits wide, since the backend query implementation only supports 64 bits right now.

PR #3622, opened 2022-12-27
Query: Adds diagnostic info regarding optimal page size and target range count

neildsh:
## Description

Adds diagnostic info regarding optimal page size and target range count to the trace. From [ICM 355724064](https://portal.microsofticm.com/imp/v3/incidents/details/355724064/home), the claim is that the optimal page size calculated by the sdk is much smaller than what it should be based on the `MaxItemCount` provided by the customer. These diagnostics will help us verify the claim.

## Type of change

- [x] Diagnostics (non-breaking change which adds diagnostic information)


kirankumarkolli:

PR #3623, opened 2023-01-02
[Internal] Client Telemetry: Adds ClientConfig API support

sourabh1007:
## Description
New API (_http://gateway-host/clientconfigs_) had been introduced to get Database Account Client Configuration from Gateway. As part of this PR, Account Client Configuration will be fetched/refreshed with Account Properties in every 5 minutes.

### Just to answer few questions asked below:
**What happens if it fails?**
Right now, it fetches only client telemetry configuration. but still I believe, it should behave similar to Account Properties API. Hence initialization should fail if this API fails.

**What is the availability of this endpoint?** 
Same as any routing gateway API or account API.

**What if we get the account information but the call to get the Client Telemetry config fails? Would it be disabled?**
It is any like any other configuration where we will have a flag which will tell if feature is enabled or not and other information will be client telemetry URL.  Client Telemetry Job will act according to the returned config.

Client Telemetry Integration will be taken care as part of separate PR.
_**Note: Supports 2.14.11 and onwards version of emulator**_

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -6423,22 +6423,6 @@ namespace Microsoft.Azure.Cosmos
             return TaskHelper.InlineIfPossible(() => this.GetDatabaseAccountPrivateAsync(this.ReadEndpoint), this.ResetSessionTokenRetryPolicy.GetRequestPolicy());
         }
 
-        /// <summary>
-        /// Read the <see cref="AccountClientConfiguration"/> from the Azure Cosmos DB service as an asynchronous operation.
-        /// </summary>
-        /// <returns>
-        /// A <see cref="AccountClientConfiguration"/> wrapped in a <see cref="System.Threading.Tasks.Task"/> object.
-        /// </returns>
-        internal Task<AccountClientConfiguration> GetDatabaseAccountClientConfigurationAsync()
-        {
-            return TaskHelper.InlineIfPossible(() => this.GetDatabaseAccountClientConfigurationPrivateAsync(this.ReadEndpoint), this.ResetSessionTokenRetryPolicy.GetRequestPolicy());
-        }
-
-        Task<AccountClientConfiguration> IDocumentClientInternal.GetDatabaseAccountClientConfigurationInternalAsync(Uri serviceEndpoint, CancellationToken cancellationToken)
-        {
-            return this.GetDatabaseAccountClientConfigurationPrivateAsync(serviceEndpoint, cancellationToken);
-        }
-
         private async Task<AccountClientConfiguration> GetDatabaseAccountClientConfigurationPrivateAsync(Uri serviceEndpoint, CancellationToken cancellationToken = default)
         {
             serviceEndpoint = new Uri(serviceEndpoint.OriginalString + "clientconfigs");
@@ -6525,6 +6509,8 @@ namespace Microsoft.Azure.Cosmos
                 {
                     (await this.QueryPartitionProvider).Update(databaseAccount.QueryEngineConfiguration);
                 }
+
+                databaseAccount.ClientConfiguration = await this.GetDatabaseAccountClientConfigurationPrivateAsync(serviceEndpoint, cancellationToken);
 
                 return databaseAccount;
             }
@@ -6773,8 +6759,6 @@ namespace Microsoft.Azure.Cosmos
             this.UseMultipleWriteLocations = this.ConnectionPolicy.UseMultipleWriteLocations && accountProperties.EnableMultipleWriteLocations;
 
             this.GlobalEndpointManager.InitializeAccountPropertiesAndStartBackgroundRefresh(accountProperties);
-            
-            this.GlobalEndpointManager.InitializeClientConfigBackgroundRefresh();
         }
 
         internal void CaptureSessionToken(DocumentServiceRequest request, DocumentServiceResponse response)
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -6773,6 +6773,8 @@ namespace Microsoft.Azure.Cosmos
             this.UseMultipleWriteLocations = this.ConnectionPolicy.UseMultipleWriteLocations && accountProperties.EnableMultipleWriteLocations;
 
             this.GlobalEndpointManager.InitializeAccountPropertiesAndStartBackgroundRefresh(accountProperties);
+            
+            this.GlobalEndpointManager.InitializeClientConfigBackgroundRefresh();
         }
 
         internal void CaptureSessionToken(DocumentServiceRequest request, DocumentServiceResponse response)
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -6761,7 +6761,6 @@ namespace Microsoft.Azure.Cosmos
             await this.accountServiceConfiguration.InitializeAsync();
             AccountProperties accountProperties = this.accountServiceConfiguration.AccountProperties;
             this.UseMultipleWriteLocations = this.ConnectionPolicy.UseMultipleWriteLocations && accountProperties.EnableMultipleWriteLocations;
-
             this.GlobalEndpointManager.InitializeAccountPropertiesAndStartBackgroundRefresh(accountProperties);
         }
 
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -6450,7 +6450,17 @@ namespace Microsoft.Azure.Cosmos
                 return request;
             }
 
-            return await gatewayModel.GetDatabaseAccountClientConfigAsync(CreateRequestMessage);
+            try
+            {
+                return await gatewayModel.GetDatabaseAccountClientConfigAsync(CreateRequestMessage);
+            }
+            catch (Exception ex)
+            {
+                DefaultTrace.TraceError($"Error while fetching client configuration. Sevice Endpoint: {serviceEndpoint}, Exception: {ex.StackTrace}");
+            }
+
+            //In case of any exception, return this as null.
+            return null;
         }
 
         /// <summary>
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -6424,27 +6424,27 @@ namespace Microsoft.Azure.Cosmos
         }
 
         /// <summary>
-        /// Read the <see cref="AccountProperties"/> from the Azure Cosmos DB service as an asynchronous operation.
+        /// Read the <see cref="AccountClientConfiguration"/> from the Azure Cosmos DB service as an asynchronous operation.
         /// </summary>
         /// <returns>
-        /// A <see cref="AccountProperties"/> wrapped in a <see cref="System.Threading.Tasks.Task"/> object.
+        /// A <see cref="AccountClientConfiguration"/> wrapped in a <see cref="System.Threading.Tasks.Task"/> object.
         /// </returns>
-        public Task<string> GetDatabaseAccountClientConfigurationAsync()
+        public Task<AccountClientConfiguration> GetDatabaseAccountClientConfigurationAsync()
         {
             return TaskHelper.InlineIfPossible(() => this.GetDatabaseAccountClientConfigurationPrivateAsync(this.ReadEndpoint), this.ResetSessionTokenRetryPolicy.GetRequestPolicy());
         }
 
-        Task<string> IDocumentClientInternal.GetDatabaseAccountClientConfigurationInternalAsync(Uri serviceEndpoint, CancellationToken cancellationToken)
+        Task<AccountClientConfiguration> IDocumentClientInternal.GetDatabaseAccountClientConfigurationInternalAsync(Uri serviceEndpoint, CancellationToken cancellationToken)
         {
             return this.GetDatabaseAccountClientConfigurationPrivateAsync(serviceEndpoint, cancellationToken);
         }
 
-        private async Task<string> GetDatabaseAccountClientConfigurationPrivateAsync(Uri serviceEndpoint, CancellationToken cancellationToken = default)
+        private async Task<AccountClientConfiguration> GetDatabaseAccountClientConfigurationPrivateAsync(Uri serviceEndpoint, CancellationToken cancellationToken = default)
         {
             serviceEndpoint = new Uri(serviceEndpoint.OriginalString + "clientconfigs");
 
             await this.EnsureValidClientAsync(NoOpTrace.Singleton);
-            if (this.GatewayStoreModel is GatewayStoreModel)
+            if (this.GatewayStoreModel is GatewayStoreModel gatewayModel)
             {
                 async ValueTask<HttpRequestMessage> CreateRequestMessage()
                 {
@@ -6469,18 +6469,7 @@ namespace Microsoft.Azure.Cosmos
                     return request;
                 }
 
-                using (HttpResponseMessage responseMessage = await this.httpClient
-                    .SendHttpAsync(
-                        CreateRequestMessage,
-                        ResourceType.DatabaseAccount,
-                        HttpTimeoutPolicyControlPlaneRead.Instance,
-                        null, cancellationToken))
-                {
-                    using (DocumentServiceResponse documentServiceResponse = await ClientExtensions.ParseResponseAsync(responseMessage))
-                    {
-                        return documentServiceResponse.ResponseBody.ReadAsString();
-                    }
-                }
+                return await gatewayModel.GetDatabaseAccountClientConfigAsync(CreateRequestMessage);
             }
 
             return null;
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -6425,7 +6425,7 @@ namespace Microsoft.Azure.Cosmos
 
         private async Task<AccountClientConfiguration> GetDatabaseAccountClientConfigurationPrivateAsync(GatewayStoreModel gatewayModel, Uri serviceEndpoint, CancellationToken cancellationToken = default)
         {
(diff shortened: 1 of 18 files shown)


Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6414,6 +6414,42 @@ public Task<AccountProperties> GetDatabaseAccountAsync()
             return TaskHelper.InlineIfPossible(() => this.GetDatabaseAccountPrivateAsync(this.ReadEndpoint), this.ResetSessionTokenRetryPolicy.GetRequestPolicy());
         }
 
+        private async Task<AccountClientConfiguration> GetDatabaseAccountClientConfigurationPrivateAsync(Uri serviceEndpoint, CancellationToken cancellationToken = default)
+        {
+            serviceEndpoint = new Uri(serviceEndpoint.OriginalString + "clientconfigs");
+
+            await this.EnsureValidClientAsync(NoOpTrace.Singleton);
+            if (this.GatewayStoreModel is GatewayStoreModel gatewayModel)
```
ealsur, line 6422: nit: The caller is already all these checks, is the duplication adding anything?

Microsoft.Azure.Cosmos/src/Resource/Settings/AccountProperties.cs
```diff
@@ -241,6 +241,11 @@ internal long ProvisionedDocumentStorageInMB
             }
         }
 
+        /// <summary>
+        /// This contains Account Client Configuration such as client telemetry configuration fetched using separate gateway API i.e. /clientconfigs
+        /// </summary>
+        internal AccountClientConfiguration ClientConfiguration { get; set; }
```
ealsur, line 247: Should it have a `[JsonIgnore]`?

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -162,6 +161,33 @@ public virtual async Task<DocumentServiceResponse> ProcessMessageAsync(DocumentS
...
+        
+        internal virtual async Task<AccountClientConfiguration> GetDatabaseAccountClientConfigAsync(Func<ValueTask<HttpRequestMessage>> requestMessage)
+        {
+            AccountClientConfiguration clientConfiguration = new AccountClientConfiguration();
+
+            // Get the ServiceDocumentResource from the gateway.
+            using (HttpResponseMessage responseMessage = await this.gatewayStoreClient.SendHttpAsync(
+                requestMessage,
+                ResourceType.DatabaseAccount,
+                HttpTimeoutPolicyNoRetry.Instance,
```
sourabh1007, line 173: TTN: not retrying

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6414,6 +6414,46 @@ public Task<AccountProperties> GetDatabaseAccountAsync()
...
+                return request;
+            }
+
+            try
+            {
+                return await gatewayModel.GetDatabaseAccountClientConfigAsync(CreateRequestMessage);
+            }
+            catch (Exception ex)
+            {
+                DefaultTrace.TraceError($"Error while fetching client configuration. Sevice Endpoint: {serviceEndpoint}, Exception: {ex.StackTrace}");
```
sourabh1007, line 6450: returning null if this API fails

kirankumarkolli:
GatewayStoreModel is created after the GatewayAccountReader is created (which is also passed to the GlobalEndpointManager). 

How about passing GatewayAccountReader into the GatewayStoreModel?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientCreateAndInitializeTest.cs
```diff
@@ -53,18 +55,67 @@ public async Task Cleanup()
...
+                                }
+                            };
+                            string payload = JsonConvert.SerializeObject(accountClientConfiguration);
+                            ;
+                            result.Content = new StringContent(payload, Encoding.UTF8, "application/json");
+
+                            return Task.FromResult(result);
+                        }
+
+                        httpCallsMade++;
```
ealsur, line 91: Move the increment before the if

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientCreateAndInitializeTest.cs
```diff
@@ -53,18 +55,67 @@ public async Task Cleanup()
...
+                        {
+                            HttpResponseMessage result = new HttpResponseMessage(HttpStatusCode.OK);
+                            string payload = JsonConvert.SerializeObject(new AccountClientConfiguration());
+                            ;
+                            result.Content = new StringContent(payload, Encoding.UTF8, "application/json");
+
+                            return Task.FromResult(result);
+                        }
+
+                        httpCallsMade++;
```
ealsur, line 113: Increment before the if, otherwise it is not reflecting the true number of http calls

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientCreateAndInitializeTest.cs
```diff
@@ -238,7 +300,7 @@ public async Task CreateAndInitializeAsync_WithValidDatabaseAndContainer_ShouldO
                 cosmosClientOptions: cosmosClientOptions);
 
             // Assert.
-            Assert.AreEqual(5, httpCallsMade);
+            Assert.AreEqual(6, httpCallsMade);
```
ealsur, line 303: Why is this increasing? How is the test enabling Client Telemetry?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientCreateAndInitializeTest.cs
```diff
@@ -299,6 +361,7 @@ public async Task CreateAndInitializeAsync_WithGatewayModeEnabled_ShouldNotOpenC
             {
                 RequestCallBack = (request, cancellationToken) =>
                 {
+                    Console.WriteLine(request.RequestUri);
```
ealsur, line 364: Remove

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientCreateAndInitializeTest.cs
```diff
@@ -327,7 +390,7 @@ public async Task CreateAndInitializeAsync_WithGatewayModeEnabled_ShouldNotOpenC
 
             // Assert.
             Assert.IsNotNull(cosmosClient);
-            Assert.AreEqual(1, httpCallsMade);
+            Assert.AreEqual(2, httpCallsMade); // Account Call and Client Config call
```
ealsur, line 393: Why is this increasing? How is the test enabling Client Telemetry?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosDatabaseAccountSettingsTests.cs
```diff
@@ -4,48 +4,90 @@
...
+            Assert.IsTrue(accountProperties.ReadableRegions.Count() > 0);
+            Assert.IsNotNull(accountProperties.WritableRegions);
+            Assert.IsTrue(accountProperties.WritableRegions.Count() > 0);
+
+            Assert.IsNotNull(accountProperties.ClientConfiguration);
+            Assert.IsNotNull(accountProperties.ClientConfiguration.ClientTelemetryConfiguration);
+            Assert.AreEqual(telemetryEndpoint, accountProperties.ClientConfiguration.ClientTelemetryConfiguration.Endpoint);
+            Assert.IsTrue(accountProperties.ClientConfiguration.ClientTelemetryConfiguration.IsEnabled);
+
+            cosmosClient.Dispose();
```
ealsur, line 67: Why not `using`?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosDatabaseAccountSettingsTests.cs
```diff
@@ -4,48 +4,90 @@
...
         }
 
+
         [TestMethod]
-        public async Task GetCosmosDatabaseAccountSettings()
+        public async Task GetCosmosDatabaseAccountSettings_WhenDisabledClientTelemetry()
         {
-            AccountProperties accountProperties = await this.cosmosClient.ReadAccountAsync();
+            CosmosClient cosmosClient = TestCommon.CreateCosmosClient(
```
ealsur, line 74:
```suggestion
            using CosmosClient cosmosClient = TestCommon.CreateCosmosClient(
```

That way if the test fails, the client is disposed anyway. Removing the dispose call

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosDatabaseAccountSettingsTests.cs
```diff
@@ -4,48 +4,90 @@
...
+                useGateway: true);
```
ealsur, line 75: Why Gateway mode? The original TestInitialization was not using it

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientCreateAndInitializeTest.cs
sourabh1007: Http call to get client configuration will always be there. It returns telemetry configuration, does matter if it was already enabled or disabled.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientCreateAndInitializeTest.cs
sourabh1007: it not about enabling client telemetry..its about making http call to get client config from gateway.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosDatabaseAccountSettingsTests.cs
sourabh1007: in this case it doesn't matter but point taken.

sourabh1007:
> out passing GatewayAccountReader into the GatewayStor

We can do a lot of refactoring, but I would like to avoid it doing as part of this PR. maybe I can take it as separate PR.

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
```diff
@@ -37,16 +36,16 @@ internal sealed class GatewayAccountReader
...
         {
             INameValueCollection headers = new RequestNameValueCollection();
             await this.cosmosAuthorization.AddAuthorizationHeaderAsync(
                 headersCollection: headers,
-                serviceEndpoint,
-                HttpConstants.HttpMethods.Get,
-                AuthorizationTokenType.PrimaryMasterKey);
-
+                requestAddress: serviceEndpoint,
+                verb: "GET",
```
ealsur, line 46: Why not keep `HttpConstants.HttpMethods.Get`?

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
```diff
@@ -58,11 +57,11 @@ private async Task<AccountProperties> GetDatabaseAccountAsync(Uri serviceEndpoin
                         additionalHeaders: headers,
                         resourceType: ResourceType.DatabaseAccount,
                         timeoutPolicy: HttpTimeoutPolicyControlPlaneRead.Instance,
-                        clientSideRequestStatistics: stats,
+                        clientSideRequestStatistics: null,
```
ealsur, line 60: Wouldn't this drop the trace?

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
```diff
@@ -37,16 +36,16 @@ internal sealed class GatewayAccountReader
             this.connectionPolicy = connectionPolicy;
             this.cancellationToken = cancellationToken;
         }
-
-        private async Task<AccountProperties> GetDatabaseAccountAsync(Uri serviceEndpoint)
+        
+        private async Task<T> GetDatabaseAccountAsync<T>(Uri serviceEndpoint)
```
ealsur, line 40: What other `T` is there to read the DatabaseAccount that is not `AccountProperties`?

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -162,6 +161,28 @@ public virtual async Task<DocumentServiceResponse> ProcessMessageAsync(DocumentS
...
+        
+        internal virtual async Task<AccountClientConfiguration> GetDatabaseAccountClientConfigAsync(Func<ValueTask<HttpRequestMessage>> requestMessage)
+        {
+            AccountClientConfiguration clientConfiguration = null;
+
+            // Get the ServiceDocumentResource from the gateway.
+            using (HttpResponseMessage responseMessage = await this.gatewayStoreClient.SendHttpAsync(
+                requestMessage: requestMessage,
+                resourceType: ResourceType.DatabaseAccount,
+                timeoutPolicy: HttpTimeoutPolicyNoRetry.Instance,
```
ealsur, line 173: Do we use the same NoRetry policy to get the account information?

ealsur (changes requested):
So this will be one more HTTP call during initialization, what happens if it fails? What is the availability of this endpoint? Is it the same as the account endpoint? What if we get the account information but the call to get the Client Telemetry config fails? Would it be disabled? Can we add these details to the Description?

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
ealsur: What if it fails?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientCreateAndInitializeTest.cs
ealsur: Can this be called out in the Description? What is the explicit endpoints involved? Is there a Design that shows the sequence of endpoints involved?

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
sourabh1007: `AccountClientConfigutaion`, it is similar to account API. just endpoint/url is changed.

ealsur: But then it is not the DatabaseAccount (the method is called GetDatabaseAccountAsync). Why are we reusing the method for a different model?

ealsur:
> What happens if it fails?
> Right now, it fetches only client telemetry configuration. but still I believe, it should behave similar to Account Properties API. > **Hence initialization should fail if this API fails.**

> What if we get the account information but the call to get the Client Telemetry config fails? Would it be disabled?
> It is any like any other configuration where we will have a flag which will tell if feature is enabled or not and other information > will be client telemetry URL. Client Telemetry Job will act according to the returned config.

The question here is how does the SDK react on the scenario where the account information was obtained, but for whatever reason, the Client Telemetry call fails. 

Why would we fail the client initialization if the Client Telemetry configuration fails? The SDK client has all the information needed to initialize, wouldn't this add a new failure point to customers initializing? Could we make it more resilient? What if it fails to initialize and we set it as disabled, and 5 minutes after with the refresh, it gets the right information, would the client telemetry job initialize?

What happens if the first initialization works (we get the information, we enable the Client Telemetry Job) but in one of the 5 min refreshes we fail to get it? Would we disable the Client Telemetry job? Or retain the previous known configuration (that it's enabled)?

Is the new client config endpoint available in all regions? For example, if the account is in West US, we can fetch the account information from `myaccount.documents.azure.com` and from `myaccount-westus.documents.azure.com`, is the same valid for the new API? Or is it just limited to the global endpoint `myaccount.documents.azure.com`?

sourabh1007:
@ealsur Please see my comments below:

**The question here is how does the SDK react on the scenario where the account information was obtained, but for whatever reason, the Client Telemetry call fails.**
_You mean "Account Client Config API" call is failed? If yes, then, first there should not any reason to fail this call only because it is developed and hosted similar to "Database Account Properties" call. Even if it fails, I think it is important for us to know that it is failing and should be fixed or supported by us like any other API. Suppressing failure of this API will increase the problem for us as we will never know that this API is failing at all._

**Why would we fail the client initialization if the Client Telemetry configuration fails? The SDK client has all the information needed to initialize, wouldn't this add a new failure point to customers initializing? Could we make it more resilient? What if it fails to initialize and we set it as disabled, and 5 minutes after with the refresh, it gets the right information, would the client telemetry job initialize?**
_Same as above, there should not any scenario where "only" this API should be failing. If it is happening, we should know.  It is re-checking for the job configuration in every 5 minutes and start and stop the job accordingly._

**What happens if the first initialization works (we get the information, we enable the Client Telemetry Job) but in one of the 5 min refreshes we fail to get it? Would we disable the Client Telemetry job? Or retain the previous known configuration (that it's enabled)?**
_it should mimic the AccountProperties API behavior._

**Is the new client config endpoint available in all regions? For example, if the account is in West US, we can fetch the account information from myaccount.documents.azure.com and from myaccount-westus.documents.azure.com, is the same valid for the new API? Or is it just limited to the global endpoint myaccount.documents.azure.com?**
_It is account level configuration hence shouldn't matter which region call it is. this API will return client configuration for that account._

sourabh1007:
incorporating all changes/feedback in a new PR which is a new design for this feature.

PR #3631, opened 2023-01-05
[DO NOT REVIEW] Test PR To Validate Test Failures For Replica Validation

kundadebdatta:
# Pull Request Template

## Description

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -196,6 +196,39 @@ public bool TryRemove(TKey key)
             return this.values.TryRemove(key, out _);
         }
 
+        /// <summary>
+        /// Refreshes the async non blocking cache on-demand for the given <paramref name="key"/>
+        /// and caches the result for later usage.
+        /// </summary>
+        /// <param name="key">The requested key to be refreshed.</param>
+        /// <param name="singleValueInitFunc">A func delegate to be invoked at a later point of time.</param>
+        public async Task RefreshAsync(
```
ealsur, line 205: nit: Add tests to https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Routing/AsyncCacheNonBlockingTests.cs

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -497,6 +504,26 @@ await this.GetServerAddressesViaGatewayAsync(request, collectionRid, new[] { par
...
+                    // with `Unknown` (initial) status.
+                    PartitionAddressInformation mergedAddresses = this.MergeAddresses(result.Item2, cachedAddresses);
+
+                    foreach (TransportAddressUri address in mergedAddresses.Get(Protocol.Tcp)?.ReplicaTransportAddressUris)
+                    {
+                        // The main purpose for this step is to move address health status from Unhealthy to UnhealthyPending.
+                        address.SetRefreshedIfUnhealthy();
+                    }
+
+                    this.ValidateUnhealthyPendingReplicas(mergedAddresses.Get(Protocol.Tcp)?.ReplicaTransportAddressUris);
```
ealsur, line 522: nit: Because this line and 516 both execute the `Get`, would it be ok to put it on a variable to avoid the double `Get` call?

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -782,6 +809,86 @@ private static string ProtocolString(Protocol protocol)
...
+
+            IEnumerable<TransportAddressUri> addressesNeedToValidation = addresses
+                .Where(address => address
+                    .GetCurrentHealthState()
+                    .GetHealthStatus() == TransportAddressHealthState.HealthStatus.UnhealthyPending);
+
+            if (addressesNeedToValidation.Any())
+            {
+                Task openConnectionsInBackgroundTask = Task.Run(async () => await this.openConnectionsHandler.TryOpenRntbdChannelsAsync(
+                    addresses: addressesNeedToValidation.ToList()));
```
ealsur, line 834: nit: Is the ToList required? it would allocate/materialize a new List?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayAddressCacheTests.cs
```diff
@@ -810,19 +848,442 @@ public async Task OpenConnectionsAsync_WhenSomeAddressResolvingFailsWithExceptio
...
+                serviceIdentity: this.serviceIdentity,
+                forceRefreshPartitionAddresses: true,
+                cancellationToken: CancellationToken.None);
+
+            Assert.AreEqual(4, addressInfo.AllAddresses.Count);
+            Assert.AreEqual(0, addressInfo.AllAddresses.Count(x => x.PhysicalUri == oldAddress));
+            Assert.AreEqual(1, addressInfo.AllAddresses.Count(x => x.PhysicalUri == newAddress));
+
+            // Added a delay for 200 milli seconds for the background open connection task to finish.
+            await Task.Delay(200);
```
ealsur, line 968: Any way to make this more deterministic? Maybe with a ManualResetEvent instead of random time?

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -782,6 +809,86 @@ private static string ProtocolString(Protocol protocol)
...
+        /// new addresses list contains some of the addresses, which are already cached, then reset the health state
+        /// of the new address to that of the cached one. If the the new addresses doesn't contain any of the cached
+        /// addresses, then keep using the health state of the new addresses, which should be `unknown`.
+        /// </summary>
+        /// <param name="newAddresses">A list of <see cref="PartitionAddressInformation"/> containing the latest
+        /// addresses being returned from gateway.</param>
+        /// <param name="cachedAddresses">A list of <see cref="PartitionAddressInformation"/> containing the cached
+        /// addresses from the async non blocking cache.</param>
+        /// <returns>A list of <see cref="PartitionAddressInformation"/> containing the merged addresses.</returns>
+        private PartitionAddressInformation MergeAddresses(
```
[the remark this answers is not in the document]
kundadebdatta: I already have tests written in `GatewayAddressCacheTests` which covers the `MergeAddresses()` in it's flow. However, making it static completely make sense.

Directory.Build.props
```diff
@@ -4,7 +4,7 @@
 		<ClientOfficialVersion>3.31.2</ClientOfficialVersion>
 		<ClientPreviewVersion>3.31.2</ClientPreviewVersion>
 		<ClientPreviewSuffixVersion>preview</ClientPreviewSuffixVersion>
-		<DirectVersion>3.29.4</DirectVersion>
+		<DirectVersion>3.30.0</DirectVersion>
```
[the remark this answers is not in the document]
kundadebdatta: Yes, I will mention the other incoming changes in the description section.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: I already have tests written in `GatewayAddressCacheTests` which covers the `AsyncCacheNonBlocking.RefreshAsync()` in it's flow. However, I will add a separate test in `AsyncCacheNonBlockingTests` to record the code change.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: Yes, at the moment, `TryOpenRntbdChannelsAsync()` (this resides in Direct package) accepts the incoming collection as `List<T>`. However, this could be an item for potential improvement when I get back to the diagnostics work. Will take a note of this, so that this could be addressed later.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayAddressCacheTests.cs
kundadebdatta: This is a great suggestion. Refactored code to use `ManualResetEvent` to make the operation more deterministic in nature. For anyone who wants to learn how  `ManualResetEvent` works, please take a look at [this documentation](https://learn.microsoft.com/en-us/dotnet/api/system.threading.manualresetevent?view=net-7.0) and this [post in stackoverflow](https://stackoverflow.com/questions/56173500/manualresetevent-waitone-and-set-repeating-multiple-times).

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: Thanks. Refactored code to make it more clean.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: Added a dedicated test for this.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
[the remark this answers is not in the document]
kundadebdatta: Marked the method as `static`.

ealsur (dismissed):

ealsur (approved):

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -196,6 +196,39 @@ public bool TryRemove(TKey key)
...
+           TKey key,
```
kirankumarkolli, line 206: Can it be modeled as GetAsync(forceRefresh=true)?

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -84,10 +85,24 @@ internal class GatewayAddressCache : IAddressCache, IDisposable
...
 
+        /// <summary>
+        /// Gets the address information from the gateway and sets them into the async non blocking cache for later lookup.
+        /// Additionally attempts to establish Rntbd connections to the backend replicas based on `shouldOpenRntbdChannels`
+        /// boolean flag.
+        /// </summary>
+        /// <param name="databaseName">A string containing the database name.</param>
+        /// <param name="collection">An instance of <see cref="ContainerProperties"/> containing the collection properties.</param>
+        /// <param name="partitionKeyRangeIdentities">A read only list containing the partition key range identities.</param>
+        /// <param name="shouldOpenRntbdChannels">A boolean flag indicating whether Rntbd connections are required to be
```
kirankumarkolli, line 103: Some context on scenarios where it would be set to false would help.

kundadebdatta: I have the context mentioned on line #179 - #182.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -278,6 +300,27 @@ public void SetOpenConnectionsHandler(IOpenConnectionsHandler openConnectionsHan
                     this.suboptimalServerPartitionTimestamps.TryAdd(partitionKeyRangeIdentity, DateTime.UtcNow);
                 }
 
+                // Refresh the cache on-demand, if there were some address that remained as unhealthy long enough (more than 1 minute)
+                // and need to revalidate its status. The reason it is not dependent on 410 to force refresh the addresses, is being:
+                // When an address is marked as unhealthy, then the address enumerator will deprioritize it and move it back to the
+                // end of the transport uris list. Therefore, it could happen that no request will land on the unhealthy address for
+                // an extended period of time therefore, the chances of 410 (Gone Exception) to trigger the forceRefresh workflow may
+                // not happen for that particular replica.
+                if (addresses
```
kirankumarkolli, line 309: Its iterating over all replica addresses in hot-path right?

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -475,6 +520,27 @@ await this.GetServerAddressesViaGatewayAsync(request, collectionRid, new[] { par
                     throw new PartitionKeyRangeGoneException(errorMessage) { ResourceAddress = collectionRid };
                 }
 
+                if (this.isReplicaAddressValidationEnabled)
```
kirankumarkolli, line 523: && cachedAddresses!= null

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
ealsur: Intellisense would show this text, which for the consumer might be better than having to dig into the code to learn it :)

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
ealsur: Maybe it's worth only doing IF there was actually an address refresh (`if (forceRefreshPartitionAddresses || request.ForceCollectionRoutingMapRefresh)`) ?

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: I do agree that it could ne modeled as `GetAsync(forceRefresh=true)`. However, thinking more on this, if we do that, then the below piece of code will be executed which IMO is not necessary for the refresh operation:

```
                try
                {
                    TValue cachedResult = await initialLazyValue.GetValueAsync();
                    if (forceRefresh == null || !forceRefresh(cachedResult))
                    {
                        return cachedResult;
                    }
                }
                catch (Exception e)
                {
                    // This is needed for scenarios where the initial GetAsync was
                    // called but never awaited.
                    if (initialLazyValue.ShouldRemoveFromCacheThreadSafe())
                    {
                        bool removed = this.TryRemove(key);

                        DefaultTrace.TraceError(
                            "AsyncCacheNonBlocking Failed GetAsync. key: {0}, tryRemoved: {1}, Exception: {2}",
                            key,
                            removed,
                            e);
                    }

                    throw;
                }

```

Basically there is no need to fetch the cached value by doing `await initialLazyValue.GetValueAsync();` and we can avoid this operation. Therefore, I would be more comfortable keeping the `RefreshAsync` code as is.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: Not really. There could still be a replica that remains in `Unhealthy` status for a period of one minute or more, even before or after address refresh. So, the idea here is, we should still force refresh the cache on-demand even if there were no force refresh triggered by the 401 `GoneException`, so that if there are any `Unhealthy` replicas, at least they get a chance to re-validate their status.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: I remember from one of my past conversation with @xinlian12 that in future, we should also consider validating the replicas with `Unknown` statuses as well. So, adding this condition `&& cachedAddresses!= null` will restrict that in future. However, upon further thinking, I felt it should be good to add the check before the condition `address.SetRefreshedIfUnhealthy()`. The reason being: If `cachedAddresses` are `null`, that would mean that the newly returned addresses from gateway would remain in `Unknown`  status and there is no cached state that could transition them into `Unhealthy`. Therefore added the check accordingly.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: Updated the verbiage accordingly.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -196,6 +196,39 @@ public bool TryRemove(TKey key)
...
+            {
+                try
+                {
+                    await initialLazyValue.CreateAndWaitForBackgroundRefreshTaskAsync(singleValueInitFunc);
+                }
+                catch (Exception ex)
+                {
+                    if (initialLazyValue.ShouldRemoveFromCacheThreadSafe())
+                    {
+                        bool removed = this.TryRemove(key);
```
xinlian12, line 219: I think it should be removed only if `(this.removeFromCacheOnBackgroundRefreshException(e))`

with the current logic, the cache will be cleared in case a timeout exception happens, which is not the desired behavior

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
xinlian12: hmm, here is my thought

The `MergeAddress` should always happen, because we always want to track the state of the replica and we always want to de-prioritize the unhealthy replica (which happens in the addressEnumerator)

The `isReplicaAddressValidationEnabled ` only decides whether we want to do open connections in the background, so I think this checking should push down a little bit more

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -760,6 +834,86 @@ private static string ProtocolString(Protocol protocol)
...
+                return newAddresses;
+            }
+
+            PerProtocolPartitionAddressInformation currentAddressInfo = newAddresses.Get(Protocol.Tcp);
+            PerProtocolPartitionAddressInformation cachedAddressInfo = cachedAddresses.Get(Protocol.Tcp);
+            Dictionary<string, TransportAddressUri> cachedAddressDict = new ();
+
+            foreach (TransportAddressUri transportAddressUri in cachedAddressInfo.ReplicaTransportAddressUris)
+            {
+                cachedAddressDict[transportAddressUri.ToString()] = transportAddressUri;
```
xinlian12, line 894: using a dictionary here may cause problem when using Emulator, you might get duplicate addresses, I can be very wrong, please double check

kundadebdatta: Thanks for pointing this out Annie! This is already taken care. I have already encountered the problem before and therefore changed the code to use `cachedAddressDict[transportAddressUri.ToString()] ` instead of  doing `cachedAddressDict.Add(transportAddressUri.ToString())`. The `cachedAddressDict[transportAddressUri.ToString()] ` will ensure the dictionary is overwritten in case as duplicate key is encountered, while the later one throws exception for duplicate keys. So, we should be good here.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: I understand that in Java, we are doing `MergeAddress` outside of the `isReplicaAddressValidationEnabled ` check, which by default uses the merging strategy with or without replica validation enabled. However, after thinking more on this, I thought of putting the merging logic inside the  `isReplicaAddressValidationEnabled ` check. Reason being this:

- If a Cx is opting-out the replica validation feature, then for the Cx, ideally the flow should be exactly how it was before the replica validation change, which is without the merging logic in place.  
- Additionally, setting the transport addresses to `Unhealthy` with a 401 `GoneException` is anyhow going to happen with or without the replica validation enabled. Therefore, when a replica validation feature is disabled, then the `AddressEnumerator` will anyhow prefer a `Healthy` replica over other three statuses. So we are covered there too.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: Thanks for pointing this out @xinlian12 ! I agree with you and made the code changes to reflect the same. Also added an unit test to cover the scenario! Tx!

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
xinlian12: thanks, make sense

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
xinlian12: Discussed offline. 
Basically we want to remain the same behavior as of today if the replica validation is disabled as always calling mergeAddress will take extra computation power etc which can have a latency impact. Discussed offline, @kundadebdatta  will do some perf tests offline to verify the perf impact and decide whether mergeAddress will always be called (so SDK will not lose the replica status, which could be beneficial even when replica validation is disabled)

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -229,6 +252,7 @@ public void SetOpenConnectionsHandler(IOpenConnectionsHandler openConnectionsHan
 
                             return this.GetAddressesForRangeIdAsync(
                                 request,
+                                cachedAddresses: currentCachedValue,
```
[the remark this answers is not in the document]
kundadebdatta: Had an offline conversation with @xinlian12 and agreed this this is an existing behavior and chances of happening this is really thin. I will keep a note of it for any future related issues.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kirankumarkolli: That is true, one extra dict lookup. 
This code path is not hot right so one extra lookup in such scenarios might be reasonable right?

kundadebdatta: I do agree that for our current scenario, this code execution is not in hot path. However, what if tomorrow some other code flow uses it in the hot path. Then that will cost once additional dic lookup, which is unnecessary for the purpose of `RefreshAsync` IMO.

I am completely open to refactor the code to use `GetAsync()` but IMO, the existing code makes the method more purposeful going by it's definition.

To make the code more clean and concise, one thing we could do is to consolidate the below code into one `private` method so that it can be invoked from both `GetAsync()` and `RefreshAsync()`. That way, we can avoid any code duplication.

```
              try
              {
                  await initialLazyValue.CreateAndWaitForBackgroundRefreshTaskAsync(singleValueInitFunc);
              }
              catch (Exception ex)
              {
                  if (this.removeFromCacheOnBackgroundRefreshException(ex))
                  {
                      bool removed = this.TryRemove(key);
                      DefaultTrace.TraceError(
                          "AsyncCacheNonBlocking Failed RefreshAsync. key: {0}, tryRemoved: {1}, Exception: {2}",
                          key,
                          removed,
                          ex);
                  }

                  throw;
              }
```

Let me know your thoughts on this.

kirankumarkolli: +1 for the refractoring

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -760,6 +834,86 @@ private static string ProtocolString(Protocol protocol)
...
+            }
+
+            IEnumerable<TransportAddressUri> addressesNeedToValidation = addresses
+                .Where(address => address
+                    .GetCurrentHealthState()
+                    .GetHealthStatus() == TransportAddressHealthState.HealthStatus.UnhealthyPending);
+
+            if (addressesNeedToValidation.Any())
+            {
+                Task openConnectionsInBackgroundTask = Task.Run(async () => await this.openConnectionsHandler.TryOpenRntbdChannelsAsync(
```
kirankumarkolli, line 858: Do #Tasks needs to be bounded?

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: Refactored the code to reflect the proposed change.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: IMO, it's not required. Reason being - 

- In real world scenario, the chances of validating a large number of unhealthy pending replicas in a hot path are low.
- The connection opening and concurrency control is already taken care in `RntbdOpenConnectionHandler` which is being used here to open the connections.

kirankumarkolli: #tasks ~ #partitions. 

During service upgrades it's possible that significant of them might need upgrade. Will that impact this flow?

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -120,30 +120,11 @@ private static bool RemoveNotFoundFromCacheOnException(Exception e)
...
-                        }
-                    }
-
-                    throw;
-                }
+                return await this.UpdateCacheAndGetValueFromBackgroundTaskAsync(
+                    key: key,
+                    initialValue: initialLazyValue,
+                    callbackDelegate: singleValueInitFunc,
+                    operationName: "GetAsync");
```
kirankumarkolli, line 127: Nit: Prefer nameof(GetAsync)

kirankumarkolli (dismissed):

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kirankumarkolli: This comment can be persured independent of this PR.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: Addressed with Latest Commit.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -196,6 +177,70 @@ public bool TryRemove(TKey key)
...
+           TKey key,
+           Func<TValue, Task<TValue>> singleValueInitFunc)
+        {
+            if (this.values.TryGetValue(key, out AsyncLazyWithRefreshTask<TValue> initialLazyValue))
+            {
+                await this.UpdateCacheAndGetValueFromBackgroundTaskAsync(
+                    key: key,
+                    initialValue: initialLazyValue,
+                    callbackDelegate: singleValueInitFunc,
+                    operationName: "RefreshAsync");
```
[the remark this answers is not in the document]
kundadebdatta: Addressed with Latest Commit.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: Have taken a note and will create a separate task to track this.

kirankumarkolli (approved):

ealsur (approved):

PR #3632, opened 2023-01-05
Query: Fixes handling of CosmosUndefined, CosmosGuid and CosmosBinary in unordered DISTINCT

neildsh:
## Description

DistinctMap.UnorderedDistinctMap was missing code paths for handling some CosmosElement types like CosmosUndefined, CosmosGuid and CosmosBinary. This change adds the missing code paths to prevent `ArgumentOutOfRangeException`s on queries where these elements are encountered.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Distinct/DistinctMap.UnorderedDistinctMap.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Distinct/DistinctMap.UnorderedDistinctMap.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.Distinct
         /// This class does that with the additional optimization that it doesn't store the whole JSON.
         /// Instead this class takes a GUID like hash and store that instead.
         /// </summary>
-        private sealed class UnorderdDistinctMap : DistinctMap
+        private sealed class UnorderedDistinctMap : DistinctMap
         {
             /// <summary>
             /// Length of UInt128 (in bytes).
@@ -98,6 +98,8 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.Distinct
                 public const string StringsLength16Plus = "StringsLength16+";
                 public const string Arrays = "Arrays";
                 public const string Object = "Object";
+                public const string Guids = "Guids";
+                public const string Blobs = "Blobs";
                 public const string SimpleValues = "SimpleValues";
             }
 
@@ -144,11 +146,23 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.Distinct
             private readonly HashSet<UInt128> objects;
 
             /// <summary>
+            /// HashSet for all CosmosGuids seen.
+            /// This set only stores the hash, since we don't want to spend the space for storing large arrays.
+            /// </summary>
+            private readonly HashSet<UInt128> guids;
+
+            /// <summary>
+            /// HashSet for all CosmosBinarys seen.
+            /// This set only stores the hash, since we don't want to spend the space for storing large objects.
+            /// </summary>
+            private readonly HashSet<UInt128> blobs;
+
+            /// <summary>
             /// Stores all the simple values that we don't want to dedicate a hash set for.
             /// </summary>
             private SimpleValues simpleValues;
 
-            private UnorderdDistinctMap(
+            private UnorderedDistinctMap(
                 HashSet<Number64> numbers,
                 HashSet<uint> stringsLength4,
                 HashSet<ulong> stringsLength8,
@@ -156,6 +170,8 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.Distinct
                 HashSet<UInt128> stringsLength16Plus,
                 HashSet<UInt128> arrays,
                 HashSet<UInt128> objects,
+                HashSet<UInt128> guids,
+                HashSet<UInt128> blobs,
                 SimpleValues simpleValues)
             {
                 this.numbers = numbers ?? throw new ArgumentNullException(nameof(numbers));
@@ -165,6 +181,8 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.Distinct
                 this.stringsLength16Plus = stringsLength16Plus ?? throw new ArgumentNullException(nameof(stringsLength16Plus));
                 this.arrays = arrays ?? throw new ArgumentNullException(nameof(arrays));
                 this.objects = objects ?? throw new ArgumentNullException(nameof(objects));
+                this.guids = guids ?? throw new ArgumentNullException(nameof(guids));
+                this.blobs = blobs ?? throw new ArgumentNullException(nameof(blobs));
                 this.simpleValues = simpleValues;
             }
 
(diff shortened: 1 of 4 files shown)


Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Distinct/DistinctMap.UnorderedDistinctMap.cs
```diff
@@ -141,21 +143,35 @@ private static class PropertyNames
             /// HashSet for all object seen.
             /// This set only stores the hash, since we don't want to spend the space for storing large objects.
             /// </summary>
-            private readonly HashSet<UInt128> objects;
+            private readonly HashSet<UInt128> objects;
+
+            /// <summary>
+            /// HashSet for all CosmosGuids seen.
+            /// This set only stores the hash, since we don't want to spend the space for storing large arrays.
```
neildsh, line 150: Fix comment

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Distinct/DistinctMap.UnorderedDistinctMap.cs
```diff
@@ -141,21 +143,35 @@ private static class PropertyNames
...
+
+            /// <summary>
+            /// HashSet for all CosmosGuids seen.
+            /// This set only stores the hash, since we don't want to spend the space for storing large arrays.
+            /// </summary>
+            private readonly HashSet<UInt128> guids;
+
+            /// <summary>
+            /// HashSet for all CosmosBinarys seen.
+            /// This set only stores the hash, since we don't want to spend the space for storing large objects.
```
neildsh, line 156: >objects [](http://example.com/codeflow?start=104&length=7)

Fix comment

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Distinct/DistinctMap.UnorderedDistinctMap.cs
```diff
@@ -72,7 +72,7 @@ private enum SimpleValues
         /// This class does that with the additional optimization that it doesn't store the whole JSON.
         /// Instead this class takes a GUID like hash and store that instead.
         /// </summary>
-        private sealed class UnorderdDistinctMap : DistinctMap
+        private sealed class UnorderedDistinctMap : DistinctMap
```
adityasa, line 75: >UnorderedDistinctMap [](http://example.com/codeflow?start=29&length=20)

Consider implementing the visitor directly on this class as an alternative.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Distinct/DistinctMap.UnorderedDistinctMap.cs
```diff
@@ -179,16 +197,8 @@ public override bool Add(CosmosElement cosmosElement, out UInt128 hash)
...
-                {
-                    CosmosArray cosmosArray => this.AddArrayValue(cosmosArray),
-                    CosmosBoolean cosmosBoolean => this.AddSimpleValue(cosmosBoolean.Value ? SimpleValues.True : SimpleValues.False),
-                    CosmosNull _ => this.AddSimpleValue(SimpleValues.Null),
-                    CosmosNumber cosmosNumber => this.AddNumberValue(cosmosNumber.Value),
-                    CosmosObject cosmosObject => this.AddObjectValue(cosmosObject),
-                    CosmosString cosmosString => this.AddStringValue(cosmosString.Value),
-                    _ => throw new ArgumentOutOfRangeException($"Unexpected {nameof(CosmosElement)}: {cosmosElement}"),
-                };
+                CosmosElementVisitor visitor = new CosmosElementVisitor(this);
```
adityasa, line 200: >this [](http://example.com/codeflow?start=72&length=4)

The circular dependency (on UnorderedDistinctMap class) is a bit awkward (which also prevents CosmosElementVisitor from being a private member, which is otherwise stateless).

Alternatively you can push the set of dictionaries inside the visitor or let UnorderedDistinctMap implement the visitor directly (instead of a private class).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/DistinctQueryPipelineStageTests.cs
```diff
@@ -16,25 +17,202 @@
...
+
+            await RunTests(testCases);
+        }
+
+        private static async Task RunTests(IEnumerable<DistinctQueryPipelineStageTestCase> testCases)
+        {
+            foreach (DistinctQueryPipelineStageTestCase testCase in testCases)
+            {
+                IEnumerator<CosmosElement> enumerator = testCase.Input.GetEnumerator();
+                int pageSize = 0;
```
adityasa, line 181: >pageSize [](http://example.com/codeflow?start=20&length=8)

Nit, pageCount (assuming my understanding is right)?

adityasa: Actually I am not sure, since in that case it seems redundant to pages.Count.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/DistinctQueryPipelineStageTests.cs
```diff
@@ -16,25 +17,202 @@
...
+                {
+                    if(pageSize <= testCase.PageSize)
+                    {
+                        pages[pages.Count - 1].Add(enumerator.Current);
+                        ++pageSize;
+                    }
+                    else
+                    {
+                        pageSize = 0;
+                        pages.Add(new List<CosmosElement>());
```
adityasa, line 193: We probably don't hit this case at the moment, since it seems like we are skipping enumerator.Current and "expected" will likely need to reflect that too?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/CosmosUndefinedQueryTests.cs
```diff
@@ -75,7 +75,10 @@ private static async Task UntypedTests(Container container)
                     expectedCount: 0),
                 MakeUndefinedProjectionTest(
                     query: $"SELECT VALUE AVG(c.{nameof(MixedTypeDocument.MixedTypeField)}) FROM c",
-                    expectedCount: 0),
+                    expectedCount: 0),
+                MakeUndefinedProjectionTest(
+                    query: $"SELECT DISTINCT VALUE SUM(c.{nameof(MixedTypeDocument.MixedTypeField)}) FROM c",
```
adityasa, line 80: >SELECT DISTINCT VALUE SUM(c.{nameof(MixedTypeDocument.MixedTypeField)}) FROM c [](http://example.com/codeflow?start=29&length=78)

Neat.

adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Distinct/DistinctMap.UnorderedDistinctMap.cs
```diff
@@ -179,16 +197,8 @@ public override bool Add(CosmosElement cosmosElement, out UInt128 hash)
...
+                return cosmosElement.Accept(visitor);
```
sboshra, line 201: please cache the visitor instance

sboshra (dismissed):
:shipit:

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/DistinctQueryPipelineStageTests.cs
neildsh: good catch!

kirankumarkolli (dismissed):

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Distinct/DistinctMap.UnorderedDistinctMap.cs
neildsh: This was discussed offline.

Summary:
I was able to address the main point which was to avoid the allocation of the visitor on each `Add` call by making it a private member. However, I dont want to implement the visitor directly in the parent, since that exposes a bunch of public methods.  Pushing the dictionaries into the visitor is not super attractive either since the parent then just becomes a pure shim, and we already have one shim layer. I think having the tiny inner class just as a dispatcher for `CosmosElement` is okay, since it's half a page of pretty obvious code.

PR #3642, opened 2023-01-11
Client Encryption: Adds validation code to check if the Key Vault URI provided in wrap metadata is a valid key identifier.

kr-santosh:
## Description

If Azure keyvault is used as a key store provider and if a user ends up using just the key name instead of complete key identifier, key vault returns the latest version of the key by default. This can result in failures during unwrap/rewrap scenario where if a rotation policy is configured on key vault, it creates a new version of the key and there is no way to set a default version of key.

## Type of change


Directory.Build.props
```diff
@@ -5,9 +5,9 @@
 		<ClientPreviewVersion>3.31.2</ClientPreviewVersion>
 		<ClientPreviewSuffixVersion>preview</ClientPreviewSuffixVersion>
 		<DirectVersion>3.30.0</DirectVersion>
-		<EncryptionOfficialVersion>2.0.0</EncryptionOfficialVersion>
+		<EncryptionOfficialVersion>2.0.1</EncryptionOfficialVersion>
 		<EncryptionPreviewVersion>2.0.0</EncryptionPreviewVersion>
-		<EncryptionPreviewSuffixVersion>preview</EncryptionPreviewSuffixVersion>
+		<EncryptionPreviewSuffixVersion>preview02</EncryptionPreviewSuffixVersion>
```
ealsur, line 10: nit: You don't need to change the suffix if you are already bumping the version?

Microsoft.Azure.Cosmos.Encryption/changelog.md
```diff
@@ -3,6 +3,16 @@ Preview features are treated as a separate branch and will not be included in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### <a name="2.0.1"/> [2.0.1](https://www.nuget.org/packages/Microsoft.Azure.Cosmos.Encryption/2.0.1) - 2023-03-11
+
+#### Added
+- [#3642](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3642) Adds validation code to check if the Key Vault URI provided in wrap metadata is a valid key identifier.
+
+### <a name="2.0.0-previewV02"/> [2.0.0-preview02](https://www.nuget.org/packages/Microsoft.Azure.Cosmos.Encryption/2.0.0-preview02) - 2023-01-11
```
ealsur, line 11: This version would not exist, the version bump would generate 2.0.1 and 2.0.1-preview02

Directory.Build.props
ealsur: You might want to generate 2.0.1 and 2.0.1-preview instead?

Microsoft.Azure.Cosmos.Encryption/src/EncryptionDatabaseExtensions.cs
```diff
@@ -91,6 +91,17 @@ public static class EncryptionDatabaseExtensions
...
             }
 
+            if (string.Equals(encryptionCosmosClient.KeyEncryptionKeyResolverName, KeyEncryptionKeyResolverName.AzureKeyVault))
+            {
+                // https://KEYVAULTNAME.vault.azure.net/keys/KEYNAME/ID
+                string[] keyVaultUriSegments = new Uri(encryptionKeyWrapMetadata.Value).Segments;
+
+                if (keyVaultUriSegments.Length != 4 || !string.Equals(keyVaultUriSegments[1], "keys/", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    throw new ArgumentException($"Invalid key vault uri'{encryptionKeyWrapMetadata.Value}' passed. Pass the complete Azure keyvault key identifier. Please refer to https://aka.ms/CosmosClientEncryption for more details.");
```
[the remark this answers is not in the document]
kr-santosh: Hi @ealsur , We want to restrict, users from passing just the keyname as part of keyvault uri rather than the complete key identifier. The reason is described in the PR description. This affects only the new keys being created and rewrapped. The problems associated with passing just the keyname is described in the PR description.

Directory.Build.props
kr-santosh: my bad. Got confused with previous versioning we had.

Microsoft.Azure.Cosmos.Encryption/src/EncryptionDatabaseExtensions.cs
abhijitpai, line 101: space after uri
URI to capital
keyvault / key vault => Key Vault

In the other exception as well.

Microsoft.Azure.Cosmos.Encryption/src/EncryptionDatabaseExtensions.cs
```diff
@@ -91,6 +91,17 @@ public static class EncryptionDatabaseExtensions
                     + " Please refer to https://aka.ms/CosmosClientEncryption for more details.");
             }
 
+            if (string.Equals(encryptionCosmosClient.KeyEncryptionKeyResolverName, KeyEncryptionKeyResolverName.AzureKeyVault))
+            {
+                // https://KEYVAULTNAME.vault.azure.net/keys/KEYNAME/ID
```
abhijitpai, line 96: keyVersion instead of Id, below as well

abhijitpai (approved):

abhijitpai (approved):

ealsur (approved):

ealsur (approved):

PR #3688, opened 2023-02-03
[Internal] Tests: Refactors emulator CI

### LINKED ISSUE #3698: CI: Emulator tests randomly failing with connectivity issues
Since 1/27 the pipelines have been failing with Connectivity issues, caused by higher CPU usage from Emulator.

ealsur:
# Major findings

## Alive CosmosClients

**There is a large number of clients being created**, client initialization consumes CPU and performs HTTP calls that are not needed. Each client that is not disposed, leaves background Tasks running (like refreshing the account information or the CPU monitor) which account for more HTTP requests to emulator.

image

this PR reduces the number:

image

Tests can inherit from `BaseCosmosClientHelper` that benefit from the boilerplate and cleanup code:

```
[Microsoft.Azure.Cosmos.SDK.EmulatorTests.TestClass]
public sealed class MyClassTest : BaseCosmosClientHelper
{
	[TestInitialize]
	public async Task Initialize()
	{
		await base.TestInitialize();
	}

	[TestCleanup]
	public async Task Cleanup()
	{
		await base.TestCleanup();
	}
}
```

This already does 2 things:

1. Create a `CosmosClient` instance accessible on `this.GetClient()` and a `Database` on `this.database`.
2. Cleanup the client and database after any test execution

This is the cleanest way to make sure all resources are cleaned up afterwards.

If the test, for some reason, needs to create a custom `CosmosClient`, it **should be disposed**, either through a `using` keyword or a manual `client.Dispose()` call.

## Leaking databases

**There is code creating databases and not deleting, this piles up on disk and can eventually generate 503s with out of capacity errors**

If the test creates any Database, it needs to be deleted at the end of the test run to avoid leaking disk resources.

If the test uses `BaseCosmosClientHelper`, and `this.database`, then it already is covered for database leaks.

## Blocking async calls

**Blocking calls raise CPU and cause thread starvation**

Under no scenario our tests should be blocking async calls. Unless there is a specific need (like testing SynchronizationContext), all calls should be fully async.

## Emulator woes

The Cosmos DB emulator on certain occasions gets stuck when initializing, to workaround this scenario, we are shutting it down when it is not coming up after 5 minutes and restarting it. This resolves the stuck scenario.

On top of that, we were not checking the final state of the emulator and starting the Test run even when the state was not "Running" which would generate a lot of failing tests and false investigations. We are now exiting early, before running any Test if the Emulator completely failed to run.

Example of run where the Emulator failed to start and Tests are not triggered:

image

image

## Side effect benefits

Code refactor seems to have positively affected running time of tests.

Snapshot of the Emulator Tests on a run without these changes:

image

Snapshot of the Emulator Tests on this branch:

image

Closes #3698


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Batch/BatchTestBase.cs
```diff
@@ -63,41 +63,40 @@ public class BatchTestBase
...
             InitializeGatewayContainers();
-            InitializeSharedThroughputContainer();
+            await InitializeSharedThroughputContainer();
         }
 
-        private static void InitializeDirectContainers()
+        private static async Task InitializeDirectContainers()
         {
-
+
```
kirankumarkolli, line 75: nit: Remove

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Batch/BatchTestBase.cs
```diff
@@ -168,24 +165,25 @@ private static void InitializeSharedThroughputContainer()
...
 
             if (BatchTestBase.SharedThroughputDatabase != null)
             {
-                BatchTestBase.SharedThroughputDatabase.DeleteStreamAsync().GetAwaiter().GetResult();
+                await BatchTestBase.SharedThroughputDatabase.DeleteStreamAsync();
             }
 
-            BatchTestBase.Client.Dispose();
+            BatchTestBase.Client?.Dispose();
+            BatchTestBase.GatewayClient?.Dispose();
```
kirankumarkolli, line 186: Thanks you.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAadTests.cs
```diff
@@ -29,98 +29,102 @@ public async Task AadMockTest(ConnectionMode connectionMode)
...
-            using (CosmosClient cosmosClient = TestCommon.CreateCosmosClient())
-            {
-                Database database = await cosmosClient.CreateDatabaseAsync(databaseId);
-                Container container = await database.CreateContainerAsync(
-                    containerId,
-                    "/id");
-            }
+            using CosmosClient cosmosClient = TestCommon.CreateCosmosClient();
+            Database database = await cosmosClient.CreateDatabaseAsync(databaseId);
+        Container container = await database.CreateContainerAsync(
```
kirankumarkolli, line 34: nit: Indentation

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAadTests.cs
```diff
@@ -29,98 +29,102 @@ public async Task AadMockTest(ConnectionMode connectionMode)
...
             string databaseId = Guid.NewGuid().ToString();
             string containerId = Guid.NewGuid().ToString();
-            using (CosmosClient cosmosClient = TestCommon.CreateCosmosClient())
-            {
-                Database database = await cosmosClient.CreateDatabaseAsync(databaseId);
-                Container container = await database.CreateContainerAsync(
-                    containerId,
-                    "/id");
-            }
+            using CosmosClient cosmosClient = TestCommon.CreateCosmosClient();
```
kirankumarkolli, line 32: nit: Braces for using helps readability, inside section then gets indented.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosContainerTests.cs
```diff
@@ -1167,62 +1167,68 @@ public async Task ReadReplaceThroughputResponseTests()
...
 
-            Assert.AreEqual(0, toStreamCount, "Custom serializer to stream should not be used for offer operations");
-            Assert.AreEqual(0, fromStreamCount, "Custom serializer from stream should not be used for offer operations");
-            await databaseNoThroughput.DeleteAsync();
+                Assert.AreEqual(0, toStreamCount, "Custom serializer to stream should not be used for offer operations");
+                Assert.AreEqual(0, fromStreamCount, "Custom serializer from stream should not be used for offer operations");
+            }
+            finally
+            {
+                await databaseNoThroughput.DeleteAsync();
```
kirankumarkolli, line 1229: nit: We rely non good coding but times it gets missed. Thoughts on a model where these are tracked and asserted. (OUTSIDE OF SCOPE OF THIS PR THOUGH)

templates/emulator-setup.yml
```diff
@@ -1,30 +1,52 @@
...
+      Write-Host "Downloading Cosmos Emulator - $env:EMULATORMSIURL" -ForegroundColor green 
+      Invoke-WebRequest "$env:EMULATORMSIURL" -OutFile "$env:temp\azure-cosmosdb-emulator.msi"
+      Write-Host "Finished Downloading Cosmos Emulator - $env:temp\azure-cosmosdb-emulator.msi" -ForegroundColor green 
+      dir "$env:temp" 
+      choco install lessmsi
+      choco upgrade lessmsi
+      mkdir "$env:temp\Azure Cosmos DB Emulator"
+      lessmsi x "$env:temp\azure-cosmosdb-emulator.msi" "$env:temp\Azure Cosmos DB Emulator\"
+      Add-MpPreference -ExclusionPath "$env:temp\Azure Cosmos DB Emulator\SourceDir\Azure Cosmos DB Emulator"
+      Add-MpPreference -ExclusionPath "$env:localappdata\CosmosDBEmulator"
```
kirankumarkolli, line 14: Is it recommended? If so we might need to do this for JAVA etc... as well right?

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosContainerTests.cs
ealsur: We already have it, in the form of `BaseCosmosClientHelper`, but refactoring all the test files to follow the same pattern was a lot of work. For future tests, we want to enforce the pattern though during PR review

templates/emulator-setup.yml
ealsur: My understanding is that this was already notified to Wes and the central SDK folks

PR #3713, opened 2023-02-17
CosmosNullReferenceException: Refactors CosmosNullReferenceException to pass along InnerException property on parent NullReferenceException

### LINKED ISSUE #3705: Refactor CosmosNullReferenceException to capture the Inner Exception in Constructor
- Refactor `CosmosNullReferenceException` to capture the inner exception that is thrown internally, so that the root cause for the null reference exception could be troubleshooted.

jeet1995:
## Description

Sets the `InnerException` property on the base `NullReferenceException` when instantiating `CosmosNullReferenceException`. 

## Closing issues

To automatically close an issue: closes #3705


philipthomas-MSFT (dismissed):

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
```diff
@@ -531,6 +531,7 @@ protected virtual void Dispose(bool disposing)
                 catch (NullReferenceException nullRefException) when (!(nullRefException is CosmosNullReferenceException))
                 {
                     CosmosNullReferenceException nullException = new CosmosNullReferenceException(
+                        nullRefException.Message,
```
ealsur, line 534: Why is this needed? The nullRefException object is already being passed down, you can extract the message from it in the constructor of the `CosmosNullReferenceException` class.

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosNullReferenceException.cs
```diff
@@ -24,10 +24,12 @@ internal class CosmosNullReferenceException : NullReferenceException
         /// Create an instance of CosmosNullReferenceException
         /// </summary>
         internal CosmosNullReferenceException(
+            String message,
             NullReferenceException originalException,
-            ITrace trace)
+            ITrace trace) 
+            : base(message ?? throw new ArgumentNullException(nameof(message)), originalException ?? throw new ArgumentNullException(nameof(originalException)))
```
ealsur, line 30:
```suggestion
            : base(originalException?.Message ?? throw new ArgumentNullException(nameof(originalException)), originalException ?? throw new ArgumentNullException(nameof(originalException)))
```

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosNullReferenceException.cs
```diff
@@ -24,10 +24,12 @@ internal class CosmosNullReferenceException : NullReferenceException
         /// Create an instance of CosmosNullReferenceException
         /// </summary>
         internal CosmosNullReferenceException(
+            String message,
```
ealsur, line 27:
```suggestion
```

ealsur: You can use `originalException.Message` to obtain the message without any new extra parameters

ealsur (changes requested):
Thanks for taking care of this, only a few small comments :)

FabianMeiswinkel (dismissed):
LGTM

philipthomas-MSFT (approved):

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosNullReferenceException.cs
```diff
@@ -25,9 +25,10 @@ internal class CosmosNullReferenceException : NullReferenceException
         /// </summary>
         internal CosmosNullReferenceException(
             NullReferenceException originalException,
-            ITrace trace)
+            ITrace trace) 
+            : base(originalException?.Message ?? throw new ArgumentNullException(nameof(originalException)), originalException ?? throw new ArgumentNullException(nameof(originalException)))
```
kirankumarkolli, line 29: Probability of Null passing for exception is very low. And the checks are unnecessary.

If the base type supports null then why not let it flow?

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosNullReferenceException.cs
ealsur: But in our case, it would be unexpected, right? Practically not possible on the existing code flow but if someone in the future modifies the flow and forgets to pass the originating exception, it would be good to catch it?

jeet1995: The base CTOR overloads with `Message` and `InnerException` arguments accept null through the type hierarchy. Throwing in derived class is necessary IMO.

PR #3717, opened 2023-02-20
[Internal] ClientTelemetry: Adds logic to limit payload size to 2 MB

sourabh1007:
## Description

Client Telemetry Service accepts maximum 2 Mb of payload. So as part of this PR, checking if json size is greater than 2 MB then break into chunks during serialization.
Limiting number of `requestInfo `will be implemented as separate PR.
## Design Doc

https://microsoftapc-my.sharepoint.com/:w:/g/personal/sourabhjain_microsoft_com/EfdDbQm27gZDpRGWMli0tO0Be4B_VS_A_GzIXf3ovabvjQ?e=LkmlfP

Updated this design doc with the approach taken as part of this PR.

## Type of change


Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -142,7 +143,7 @@ private void StartObserverTask()
         ///  2. Load VM metedata information (one time at the time of initialization)
         ///  3. Calculate and Send telemetry Information to juno service (never ending task)/// </summary>
         /// <returns>Async Task</returns>
-        private async Task EnrichAndSendAsync()
```
ealsur, line 145: Why is this a ValueTask? On which scenario the operation will return a non-async response?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -164,35 +165,18 @@ private async Task EnrichAndSendAsync()
...
                     ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram> cacheRefreshInfoSnapshot
-                       = Interlocked.Exchange(ref this.cacheRefreshInfoMap, new ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram>());
-
-                    this.clientTelemetryInfo.OperationInfo = ClientTelemetryHelper.ToListWithMetricsInfo(operationInfoSnapshot);
-                    this.clientTelemetryInfo.CacheRefreshInfo = ClientTelemetryHelper.ToListWithMetricsInfo(cacheRefreshInfoSnapshot);
-
-                    await this.SendAsync();
+                        = Interlocked.Exchange(ref this.cacheRefreshInfoMap, new ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram>());
+                    
+                    this.ProcessAndSend(
```
ealsur, line 178: How is this change related to the 2MB limit? This is changing how the telemetry is sent from foreground to a fire-and-forget task

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -202,7 +186,44 @@ private async Task EnrichAndSendAsync()
...
+
+                string json = JsonConvert.SerializeObject(this.clientTelemetryInfo, settings);
 
+                // if JSON is greater than 2 MB
+                if (json.Length > ClientTelemetryOptions.PayloadSizeThreshold)
+                {
+                    foreach (string property in ClientTelemetryOptions.PropertiesContainMetrics)
+                    {
+                        settings.ContractResolver = new IncludePropertyContractResolver(property);
+                        json = JsonConvert.SerializeObject(this.clientTelemetryInfo, settings);
```
ealsur, line 215: Is it possible for 1 of these `property` objects to be greater than 2Mb? What if the whole reason it was 2Mb was because there was 1 property that was larger?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: As processing part is growing over time, it is code refactoring to put all the processing part at one function which will be fire and forget.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: `ValueTask` is kind of Wrapper on Task only, and it has its benefits in terms of memory allocation. Thats I am using `ValueTask` here. Please enlighten me if it won't provide any benefit over `Task`.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: covering this also.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+
+            JsonSerializerSettings settings = ClientTelemetryOptions.JsonSerializerSettings;
+            string json = JsonConvert.SerializeObject(clientTelemetryInfo, settings);
+
+            // If JSON is greater than 2 MB
+            if (json.Length > ClientTelemetryOptions.PayloadSizeThreshold)
+            {
+                foreach (string property in ClientTelemetryOptions.PropertiesContainMetrics)
+                {
+                    clientTelemetryInfo.OperationInfo = null; //clearing everything
```
kirankumarkolli, line 72: Is clearing in every loop iteration needed?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+                    clientTelemetryInfo.CacheRefreshInfo = null;
```
kirankumarkolli, line 73: SystemInfo is repeated with each iteration right is it exacted?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+
+                    if (property == "OperationInfo")
```
kirankumarkolli, line 75: One other alternative is to dynamically keep adding records one by one and see if its reaching the limits. 
Please check out CosmosTraceDiagnostics serialization.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+            clientTelemetryInfo.OperationInfo = ClientTelemetryHelper.ToListWithMetricsInfo(operationInfoSnapshot);
+            clientTelemetryInfo.CacheRefreshInfo = ClientTelemetryHelper.ToListWithMetricsInfo(cacheRefreshInfoSnapshot);
+
+            JsonSerializerSettings settings = ClientTelemetryOptions.JsonSerializerSettings;
+            string json = JsonConvert.SerializeObject(clientTelemetryInfo, settings);
+
+            // If JSON is greater than 2 MB
+            if (json.Length > ClientTelemetryOptions.PayloadSizeThreshold)
+            {
+                foreach (string property in ClientTelemetryOptions.PropertiesContainMetrics)
```
kirankumarkolli, line 70: What's the point of this loop, we can just do section one for OperationInfo and other for CacheRefreshInfo right?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
kirankumarkolli: It will allow combining both OperationInfo and CacheRefreshInfo also into a single reqeust.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryOptions.cs
```diff
@@ -94,6 +94,15 @@ internal static class ClientTelemetryOptions
             MaxDepth = 64, // https://github.com/advisories/GHSA-5crp-9r3c-p9vr
         };
 
+        internal static readonly HashSet<string> PropertiesContainMetrics = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "OperationInfo", "CacheRefreshInfo" };
+
+        internal static int PayloadSizeThreshold = 1024 * 1024 * 2; // 2MB
+        internal static Dictionary<string, int> PropertiesWithPageSize = new Dictionary<string, int>
+        {
+            { "OperationInfo", 1000 },
```
kirankumarkolli, line 102: Where are these limits coming from?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+                {
+                    clientTelemetryInfo.OperationInfo = null; //clearing everything
+                    clientTelemetryInfo.CacheRefreshInfo = null;
+
+                    if (property == "OperationInfo")
+                    {
+                        List<Dictionary<OperationInfo, (LongConcurrentHistogram latency, LongConcurrentHistogram requestcharge)>> pages 
+                            = this.GetPages<OperationInfo, (LongConcurrentHistogram latency, LongConcurrentHistogram requestcharge)>(operationInfoSnapshot, property);
```
kirankumarkolli, line 78: Am under assumption that we are limiting the @pages or entries we publish to start with. 
This needs to be guarded on service and appropriately throttle as well.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+                            endpointUrl,
+                            "POST",
+                            AuthorizationTokenType.PrimaryMasterKey);
+
+                    foreach (string key in headersCollection.AllKeys())
+                    {
+                        request.Headers.Add(key, headersCollection[key]);
+                    }
+
+                    request.Headers.Add(HttpConstants.HttpHeaders.DatabaseAccountName, globalDatabaseAccountName);
```
kirankumarkolli, line 179: I understand that its a older decision. 
Wondering if addressing also needs to change to include account info. 
Address will allow most web services to parameterize and flow context natively (including tracing even in-case of errors before header processing)

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+
+                using HttpResponseMessage response = await this.httpClient.SendHttpAsync(CreateRequestMessage,
+                                                    ResourceType.Telemetry,
+                                                    HttpTimeoutPolicyNoRetry.Instance,
+                                                    null,
+                                                    cancellationToken);
+
+                if (!response.IsSuccessStatusCode)
+                {
+                    DefaultTrace.TraceError("Juno API response not successful. Status Code : {0},  Message : {1}", response.StatusCode, response.ReasonPhrase);
```
kirankumarkolli, line 197: "Juno" is internal name, please exclude.

kirankumarkolli, line 197: Prefer: $"... Status Code : {response.StatusCode},  Message : {response.ReasonPhrase}"

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,214 @@
...
+            this.tokenProvider = tokenProvider ?? throw new ArgumentNullException(nameof(tokenProvider));
+        }
+
+        internal Task ProcessAndSendAsync(
+            ClientTelemetryProperties clientTelemetryInfo, 
+            ConcurrentDictionary<OperationInfo, (LongConcurrentHistogram latency, LongConcurrentHistogram requestcharge)> operationInfoSnapshot,
+            ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram> cacheRefreshInfoSnapshot,
+            CancellationToken cancellationToken)
+        {
+            return Task.Run(async () => await this.GenerateOptimalSizeOfPayloadAndSendAsync(clientTelemetryInfo, operationInfoSnapshot, cacheRefreshInfoSnapshot, cancellationToken), cancellationToken);
```
kirankumarkolli, line 42: Task might fail (missing exception handling and goes un-observed)

kirankumarkolli (changes requested):

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryOptions.cs
sourabh1007: I checked the size of each JSON block and calculated how many JSON blocks it will take to make it 1.5 MB (.5 MB is reserved for other info and future enhancements) and then rounded off to a nearby number.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: yes..you are right.. I will add the number of page limit also

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: I did that but got a comment to remove revert it to this again. @ealsur what do you think?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: it is being tracked in ClientTelemetry class line 185. Yes, I missed to put the loggers there, putting it.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: yes...it is same.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
ealsur: ValueTask is used when an async method can return a cached value without performing an async operation most of the time.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -170,39 +163,43 @@ private async Task EnrichAndSendAsync()
...
 
                     ConcurrentDictionary<RequestInfo, LongConcurrentHistogram> requestInfoSnapshot
                       = Interlocked.Exchange(ref this.requestInfoMap, new ConcurrentDictionary<RequestInfo, LongConcurrentHistogram>());
 
-                    this.clientTelemetryInfo.OperationInfo = ClientTelemetryHelper.ToListWithMetricsInfo(operationInfoSnapshot);
-                    this.clientTelemetryInfo.CacheRefreshInfo = ClientTelemetryHelper.ToListWithMetricsInfo(cacheRefreshInfoSnapshot);
-                    this.clientTelemetryInfo.RequestInfo = ClientTelemetryHelper.ToListWithMetricsInfo(requestInfoSnapshot);
-                    
-                    await this.SendAsync();
+                    _ = this.processor
```
ealsur, line 184: This is a behavior change that is unrelated to the 2Mb size. This is changing the sending of the telemetry data from async in-line, to fire-and-forget. Can you please undo this and treat this as a separate PR with some indication of why is this a beneficial change? What is the performance benefit? How is concurrency handled on `this.numberOfFailures`?

ealsur (changes requested):
Please undo the foreground/background change when sending data and defer that to another PR, focus of this PR is the 2Mb limit, right?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: old code, not applicable now..hence resolving

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: old code, not applicable now..hence resolving. Implemented as you suggested.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryOptions.cs
sourabh1007: old code, not applicable now..hence resolving

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: it will be covered as separate PR.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: as discussed in last catch up with tools federation team. we will revisit it if anything comes up in security review.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: As part of 2Mb payload size change, I wrote a custom serialization logic for client telemetry payload which will break the payload and will send those chunks and decided to run this new logic in a separate Task. Existing tests are working fine.
This implementation is not part of any performance enhancement.
for `this.numberOfFailures`, why do you think concurrency needs to be handled. It is a job which runs in every 10 min.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: removed.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
ealsur: How is that custom logic related to this change? This is changing an inline async execution to a background fire-and-forget pattern. This is unrelated to the serialization or the 2Mb problem.

Microsoft.Azure.Cosmos/src/Json/IJsonWriter.cs
```diff
@@ -103,7 +103,8 @@ interface IJsonWriter
         /// Writes an signed 4-byte integer to the internal buffer.
         /// </summary>
         /// <param name="value">The value of the integer to write.</param>
-        void WriteInt32Value(int value);
+        /// <param name="withPrefix">If prefix should be there.</param>
+        void WriteInt32Value(int value, bool withPrefix = true);
```
ealsur, line 107: Undo. No need to use IJsonWriter to do JSon serialization. You can use Newtonsoft.JsonWriter

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: made changes as we discussed.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -170,39 +162,45 @@ private async Task EnrichAndSendAsync()
...
+                                requestInfoSnapshot: requestInfoSnapshot,
+                                cancellationToken: this.cancellationTokenSource.Token);
+
+                        this.numberOfFailures = 0;
+                    }
+                    catch (Exception ex)
+                    {
+                        this.numberOfFailures++;
+
+                        DefaultTrace.TraceError("Telemetry Job Processor failed with error : ", ex);
```
ealsur, line 202:
```suggestion
                        DefaultTrace.TraceError("Telemetry Job Processor failed with error : {0}", ex);
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+            JsonWriter writer = ClientTelemetryPayloadWriter.GetWriter(properties, "operationInfo");
+            
+            if (operationInfoSnapshot != null && operationInfoSnapshot.Count > 0)
+            {
+                var opsEnumerator = operationInfoSnapshot.GetEnumerator();
+                while (opsEnumerator.MoveNext())
+                {
+                    long lengthNow = stringBuilder.Length;
```
ealsur, line 41: Who is appending to this StringBuilder? It nevers grows. Line 41 is checking the length (let's assume 0 initially), then line 60 calls `ToString()` but where is anything that Appends? When is the `writer` content go into the StringBuilder?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -0,0 +1,161 @@
...
+                        writer.WriteEndArray();
+                        writer.WriteEndObject();
+
+                        await callback.Invoke(stringBuilder.ToString());
+
+                        writer = ClientTelemetryPayloadWriter.GetWriter(properties, "operationInfo");
+                    }
+
+                    writer.WriteRawValue(latencyMetrics);
+                    writer.WriteRawValue(requestChargeMetrics);
```
ealsur, line 66: Are these 2 (latencyMetrics, requestChargeMetrics) never going to go beyond 2Mb?

ealsur (approved):

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
sourabh1007: it has to go together (because belong to the same operation)...i cannot send them separately.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -170,39 +162,45 @@ private async Task EnrichAndSendAsync()
...
+                    {
+                        await this.processor
+                        .ProcessAndSendAsync(
+                                clientTelemetryInfo: this.clientTelemetryInfo,
+                                operationInfoSnapshot: operationInfoSnapshot,
+                                cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
+                                requestInfoSnapshot: requestInfoSnapshot,
+                                cancellationToken: this.cancellationTokenSource.Token);
+
+                        this.numberOfFailures = 0;
```
kirankumarkolli, line 196: Is the counter meant for #NWFailures or #EnrichAndSendAsync?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryHelper.cs
```diff
@@ -46,115 +46,61 @@ internal static async Task<AccountProperties> SetAccountNameAsync(GlobalEndpoint
...
             }
 
             DefaultTrace.TraceVerbose("System Usage recorded by telemetry is : {0}", systemUsageHistory);
-
-            systemInfoCollection.Add(TelemetrySystemUsage.GetCpuInfo(systemUsageHistory.Values));
-            systemInfoCollection.Add(TelemetrySystemUsage.GetMemoryRemainingInfo(systemUsageHistory.Values));
-            systemInfoCollection.Add(TelemetrySystemUsage.GetAvailableThreadsInfo(systemUsageHistory.Values));
-            systemInfoCollection.Add(TelemetrySystemUsage.GetThreadWaitIntervalInMs(systemUsageHistory.Values));
-            systemInfoCollection.Add(TelemetrySystemUsage.GetThreadStarvationSignalCount(systemUsageHistory.Values));
+            List<SystemInfo> systemInfoCollection = new List<SystemInfo>
```
kirankumarkolli, line 60: nit: Specify the capacity for know cases which avoid any list re-initializations

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -0,0 +1,161 @@
...
+    using System.IO;
+    using System.Text;
+    using System.Threading.Tasks;
+    using HdrHistogram;
+    using Microsoft.Azure.Cosmos.Telemetry.Models;
+    using Newtonsoft.Json;
+
+    internal static class ClientTelemetryPayloadWriter
+    {
+        private static readonly StringBuilder stringBuilder = new StringBuilder(ClientTelemetryOptions.PayloadSizeThreshold);
```
kirankumarkolli, line 19: Why global state? 
If the goal is to reduce allocations then please prefer pooling instead.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -0,0 +1,161 @@
...
+            Func<string, Task> callback)
+        {
+            if (properties == null)
+            {
+                throw new ArgumentNullException(nameof(properties));
+            }
+
+            JsonWriter writer = ClientTelemetryPayloadWriter.GetWriter(properties, "operationInfo");
+            
+            if (operationInfoSnapshot != null && operationInfoSnapshot.Count > 0)
```
kirankumarkolli, line 36: nit: Prefer Any() instead of Count

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                var opsEnumerator = operationInfoSnapshot.GetEnumerator();
```
kirankumarkolli, line 38: nit: 'foreach (var entry in operationInfoSnapshot)' will be more intutive

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// <summary>
+        /// If  payload size is greater than 2 MB, it breaks the list of objects into multiple chunks and send it one by one.
+        /// else send as it is.
+        /// </summary>
+        /// <param name="clientTelemetryInfo"></param>
+        /// <param name="operationInfoSnapshot"></param>
+        /// <param name="cacheRefreshInfoSnapshot"></param>
+        /// <param name="requestInfoSnapshot"></param>
+        /// <param name="cancellationToken"></param>
+        internal async Task GenerateOptimalSizeOfPayloadAndSendAsync(ClientTelemetryProperties clientTelemetryInfo,
```
kirankumarkolli, line 65: How about inline this method with above one?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: here any failure related to Client Telemetry Service API will increase this counter.
**ref.** https://github.com/Azure/azure-cosmos-dotnet-v3/blob/f262f8941f541332509eba64a2297e9d5e221675/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs#L462
**ref.** https://github.com/Azure/azure-cosmos-dotnet-v3/blob/f262f8941f541332509eba64a2297e9d5e221675/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs#L475

and resetting the counter on any successful call https://github.com/Azure/azure-cosmos-dotnet-v3/blob/f262f8941f541332509eba64a2297e9d5e221675/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs#L468

FYI: it is existing code, not part of this PR.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryHelper.cs
sourabh1007: There will be only 5 elements in the list, So reinitialization won't trigger at all. Setting capacity to "5", it will save some bits.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -0,0 +1,161 @@
...
+            if (properties == null)
+            {
+                throw new ArgumentNullException(nameof(properties));
+            }
+            
+            StringBuilder stringBuilder = new StringBuilder(ClientTelemetryOptions.PayloadSizeThreshold);
+            
+            JsonWriter writer = ClientTelemetryPayloadWriter.GetWriterWithSectionStartTag(stringBuilder, properties, "operationInfo");
+            
+            if (operationInfoSnapshot.Any())
```
kirankumarkolli, line 36: nit: operationInfoSnapshot?.Any()

kirankumarkolli (approved):

kirankumarkolli (approved):

PR #3730, opened 2023-02-28
Query: Fixes System.ArgumentException when using PartitionKey.None on x86, Linux or in Optimistic Direct Execution

### LINKED ISSUE #2030: PartitionKey has fewer components than defined the collection resource. - On Azure not Local
**Describe the bug**
Performing a query similar to the below on a set of records.  Container's partition key is /pk .  Documents have been stored as having PartitionKey.None (its a very small container).  Query is performed using PartitionKey.None:
```
            QueryDefinition query = new QueryDefinition("SELECT TOP 1 h.slug FROM h WHERE ARRAY_CONTAINS(h.knownDomains, @domain)")
                                        .WithParameter("@domain", domain);
            QueryRequestOptions queryOptions = new QueryRequestOptions()
            {
                MaxItemCount = 1,
              PartitionKey = PartitionKey.None
            };
```
This code works in local debugging in Debug or Release modes.
When deployed to Azure Functions, the behavior changes and the following exception is encountered: PartitionKey has fewer components than defined the collection resource.

I have changed deployment methods, suspecting nuget package differences but did not solve.

**Expected behavior**
Identical functionality in local and on Azure, for PartitionKey.None to be accepted.

**Actual behavior**
PartitionKey has fewer components than defined the collection resource encountered in Azure Functions, not in local.

**Environment summary**
SDK Version: Microsoft.Azure.Cosmos versions 3.14.0 and 3.15.0
OS Version: Windows 10 local development, Azure Functions v3

**Additional context**
Work Around: Remove Partition Key in query.

neildsh:
## Description

Microsoft.Azure.Documents.PartitionKeyInternal has the following method:
`public string GetEffectivePartitionKeyString(PartitionKeyDefinition partitionKeyDefinition, bool strict = true)`
Unfortunately, this is a broken contract, since it also has members:
```
        private static readonly PartitionKeyInternal NonePartitionKey = new PartitionKeyInternal();
        public static PartitionKeyInternal None
        {
            get
            {
                return PartitionKeyInternal.NonePartitionKey;
            }
        }
```

This implies that  `GetEffectivePartitionKeyString` throws an `System.ArgumentException` when `this` evaluates to `PartitionKeyInternal.None`.

The higher layer at `Microsoft.Azure.Cosmos.PartitionKey` is also broken since it exposes a `PartitionKey.None` member that maps to `PartitionKeyInternal.None`, instead of mapping to one of `PartitionKeyInternal.Empty` or `PartitionKeyInternal.Undefined` depending on the whether the collection is a legacy fixed collection that was migrated to be system partitioned or if it is regular multi partitioned collection.

This change uses a workaround called `CachedContainerQueryProperties` which currently has the behavior mentioned above.

Also, did some housekeeping in `OptimisticDirectExecutionQueryTests`

## Type of change

## Closing issues

closes #2030

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
@@ -214,7 +214,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
                                 List<Documents.PartitionKeyRange> targetRanges = await cosmosQueryContext.QueryClient.GetTargetPartitionKeyRangesByEpkStringAsync(
                                     cosmosQueryContext.ResourceLink,
                                     containerQueryProperties.ResourceId,
-                                    inputParameters.PartitionKey.Value.InternalKey.GetEffectivePartitionKeyString(partitionKeyDefinition),
+                                    containerQueryProperties.EffectivePartitionKeyString,
                                     forceRefresh: false,
                                     createQueryPipelineTrace);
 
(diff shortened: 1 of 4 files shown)


neildsh:
This change needs further vetting. There is some lack of clarity on what PartitionKey.None means in a physical sense

ealsur:
@neildsh My understanding is that PartitionKey.None means the document has no partition key property, PartitionKey.Null means the item has a partition key with a null value.

sboshra (approved):
:shipit:

PR #3739, opened 2023-03-02
[Internal] ConnectionStateListner: Adds logic to record transport uris and mark replica health statuses

kundadebdatta:
# Pull Request Template

## Description

Connection State Listener Changes.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
@@ -387,56 +387,6 @@ namespace Microsoft.Azure.Cosmos.Routing
             }
         }
 
-        public void TryRemoveAddresses(
-            ServerKey serverKey)
-        {
-            if (serverKey == null)
-            {
-                throw new ArgumentNullException(nameof(serverKey));
-            }
-
-            if (this.serverPartitionAddressToPkRangeIdMap.TryRemove(serverKey, out HashSet<PartitionKeyRangeIdentity> pkRangeIds))
-            {
-                PartitionKeyRangeIdentity[] pkRangeIdsCopy;
-                lock (pkRangeIds)
-                {
-                    pkRangeIdsCopy = pkRangeIds.ToArray();
-                }
-
-                foreach (PartitionKeyRangeIdentity pkRangeId in pkRangeIdsCopy)
-                {
-                    DefaultTrace.TraceInformation("Remove addresses for collectionRid :{0}, pkRangeId: {1}, serviceEndpoint: {2}",
-                       pkRangeId.CollectionRid,
-                       pkRangeId.PartitionKeyRangeId,
-                       this.serviceEndpoint);
-
-                    this.serverPartitionAddressCache.TryRemove(pkRangeId);
-                }
-            }
-        }
-
-        public async Task<PartitionAddressInformation> UpdateAsync(
-            PartitionKeyRangeIdentity partitionKeyRangeIdentity,
-            CancellationToken cancellationToken)
-        {
-            if (partitionKeyRangeIdentity == null)
-            {
-                throw new ArgumentNullException(nameof(partitionKeyRangeIdentity));
-            }
-
-            cancellationToken.ThrowIfCancellationRequested();
-
-            return await this.serverPartitionAddressCache.GetAsync(
-                       key: partitionKeyRangeIdentity,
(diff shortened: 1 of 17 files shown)


Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -387,56 +387,6 @@ public void SetOpenConnectionsHandler(IOpenConnectionsHandler openConnectionsHan
...
 
-        public void TryRemoveAddresses(
-            ServerKey serverKey)
-        {
-            if (serverKey == null)
-            {
-                throw new ArgumentNullException(nameof(serverKey));
-            }
-
-            if (this.serverPartitionAddressToPkRangeIdMap.TryRemove(serverKey, out HashSet<PartitionKeyRangeIdentity> pkRangeIds))
```
kirankumarkolli, line 398: Is serverPartitionAddressToPkRangeIdMap needed as instance state?

Microsoft.Azure.Cosmos/src/direct/Channel.cs
```diff
@@ -235,8 +235,12 @@ private void Initialize()
         /// <summary>
         /// Returns the background channel initialization task.
         /// </summary>
+        /// <param name="activityId">activityId.</param>
+        /// <param name="transportAddressUri">transportAddressUri.</param>
         /// <returns>The initialization task.</returns>
-        public Task OpenChannelAsync(Guid activityId)
+        public Task OpenChannelAsync(
+            Guid activityId,
+            TransportAddressUri transportAddressUri)
```
kirankumarkolli, line 243: Channel already has URI and new TransportAddressURi in Open is ambuguous right?

kirankumarkolli: If there are two replica addressed for the same endpoint will be two different TransprtAddressURI's right?
So won't it result in partial state updates?

kundadebdatta: This is a good point. However, the `Uri` object doesn't have the health statuses inside it. We were interested to keep track of the same transport address uri instance with it's current health status, so that we could mark it to `Unhealthy` when needed.

kundadebdatta: So literally, the `TransportAddressUri` has no impact on opening the channels. I agree that it looks very confusing the way it's implemented.

kundadebdatta: [From this documentation](https://cosmosdbdocs.azurewebsites.net/design/network/dotnet-client-transport.html), I feel that it could be a good idea to clean up the `IChannel` dependency from the `LoadBalancingChannel` so that any changes needed for `LoadBalancingChannel` doesn't impact the `Channel` class, and vice-versa. This will also help reducing any confusions.

image

Microsoft.Azure.Cosmos/src/direct/LoadBalancingPartition.cs
```diff
@@ -259,9 +266,15 @@ public void Dispose()
 
             if (waitForBackgroundInitializationComplete)
             {
-                await newChannel.OpenChannelAsync(activityId);
+                await newChannel.OpenChannelAsync(activityId, transportAddressUri);
             }
 
+            // Invoke the connection state listner to record the transport address uris.
+            this.channelProperties
```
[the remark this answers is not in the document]
kirankumarkolli: Another alternative is to make the connection abstraction support listeners. 
Listeners will be invoked for <EndPoint, Status>. 

Now we can have one global listener which get plugged through maintaining the global state. Thoughts?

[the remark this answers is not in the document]
kundadebdatta: Based on our offline conversation, when a replica status changes/ updates, it may not be applicable for all the other replicas on that specific endpoint. Hence, we will stick with the current approach to record the replica health status, by notifying the connection state listener using the `OnChannelInitializationEvent()` from `LoadBalancingPartition`.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: Good Point! Thinking more on this, I don't think the `serverPartitionAddressToPkRangeIdMap` instance is needed anymore. Hence, I have made the code changes to remove the above dependency from `GatewayAddressCache`.

PR #3742, opened 2023-03-06
[Internal] Open Telemetry: Adds operation type list in batch operation activity

sourabh1007:
## Description
Right now, for batch operations, we don't have any attribute in open telemetry which can tell "what all operations were made in that batch operation". In this PR, adding operation type with count as an attribute

## Type of change


sourabh1007:
not having any business value... hence closing it.

cc @kirankumarkolli @jcocchi

PR #3746, opened 2023-03-08
Adding GetNormalizedRegionName to fix #2330

### LINKED ISSUE #2330: ApplicationRegion parameter does not work with output from ARM
**Describe the bug**
I have a compute resource in Azure, lets say an AKS cluster. Doesn't really matter which one since the problem is the same across multiple services:

In my deployment pipeline I run a CLI command to retrieve the region, for instance:
`az aks show --resource-group my-rg -n mycluster --query "location" --output tsv`

This outputs something like `northeurope` or `eastus2`. This seems consistent across many Azure resources.

Now, I want to use that output to feed into an app setting which I then want to use to configure my Cosmos DB client with ApplicationRegion:

```
CosmosClientBuilder clientBuilder = new CosmosClientBuilder(SysConfig.CosmosEndpointUri, SysConfig.CosmosPrimaryKey);
var builderOptions = clientBuilder.WithApplicationRegion(Environment.GetEnvironmentVariable("AZURE_REGION"));
```

Turns out, you can't. Because the SDK wants `North Europe` or `East US 2`. I don't see any way how to programmatically split this up

**To Reproduce**
See above.

**Expected behavior**
The SDK should accept standard naming input for Azure regions like `northeurope`.

**Actual behavior**
See above.
image

**Environment summary**
SDK Version: 3.14.0

sebader:
## Description
This is a bit of a crude fix for https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2330
But since nothing has happened there and this is a really bugging a lot of people, I went ahead with this quick proposal.

This adds a helper method to get the region name in the shape and form that the SDK wants it

## Type of change
 
* [ ]  Bug fix (non-breaking change which fixes an issue)
* [x]  New feature (non-breaking change which adds functionality)
* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ]  This change requires a documentation update

## Closing issues
#2330


ealsur:
This same requirement is being tackled by https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3857.
Thanks for your valuable contribution and for providing your insight and work to address this issue.

PR #3750, opened 2023-03-10
ClientTelemetry Throttling 

sourabh1007:
# Pull Request Template

## Description

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Telemetry/Sampler/NetworkRequestSampler.cs
```diff
@@ -0,0 +1,75 @@
...
+        private readonly int NetworkFailuresCount;
+        private readonly ISet<RequestInfo> TempStorage;
+        
+        public NetworkRequestSampler(int maxNumberOfFailures)
+        {
+            this.NetworkFailuresCount = maxNumberOfFailures;
+            this.TempStorage = new HashSet<RequestInfo>();
+        }
+        
+        public bool ShouldSample(RequestInfo requestInfo)
```
FabianMeiswinkel, line 22: Sampling should not just use the first random 1 samples - but choose the "100 most relevant" samples based on status code and latency, correct? Doesn't have to be exact science even without much effort we can do better than random selection?

sourabh1007: Here, Sampling will be applied over service endpoints with non-success status code and high latency. and I am capturing top 100 such service endpoints.
What do you mean by "most relevant"?

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/NetworkRequestSampler.cs
```diff
@@ -0,0 +1,75 @@
...
+                return false;
+            }
+            
+            if (this.TempStorage.Count < this.NetworkFailuresCount)
+            {
+                return this.TempStorage.Add(requestInfo);
+            }
+            else
+            {
+                bool isAdded = this.TempStorage.Add(requestInfo);
```
ealsur, line 35: Is this Add/Remove logic to check if the hash of the requestInfo is already present? What properties are in the hash?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -344,11 +349,19 @@ await this.processor
         /// <param name="containerId"></param>
         /// <param name="databaseId"></param>
         /// <param name="storeResponseStatistics"></param>
-        private void RecordRntbdResponses(string containerId, string databaseId, List<StoreResponseStatistics> storeResponseStatistics)
+        /// <param name="droppedRntbdRequestCount"></param>
+        private void RecordRntbdResponses(string containerId, string databaseId, List<StoreResponseStatistics> storeResponseStatistics, out int droppedRntbdRequestCount)
         {
+            using IClientTelemetrySampler<RequestInfo> networkRequestSampler
```
ealsur, line 355: Creating the sample here will only sample over the information of 1 operation (if 1 operation had 2 requests or 200) not across all the operations in the batch that will be sent to client telemetry, correct?

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/NetworkRequestSampler.cs
ealsur: Why not store the hash in the hashset? Would that use less memory than a copy of the full object?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -47,10 +47,19 @@ internal static class ClientTelemetryPayloadWriter
...
+                        = new MetricInfo(
+                                metricsName: ClientTelemetryOptions.DroppedRntbdRequestsName, 
+                                unitName: ClientTelemetryOptions.DroppedRntbdRequestsUnit, 
+                                count: entry.Value.droppedRequestCount);
+                    
                     string latencyMetrics = JsonConvert.SerializeObject(payloadForLatency);
                     string requestChargeMetrics = JsonConvert.SerializeObject(payloadForRequestCharge);
-                    
-                    if (lengthNow + latencyMetrics.Length + requestChargeMetrics.Length > ClientTelemetryOptions.PayloadSizeThreshold)
+                    string droppedRntbdRequestsMetrics = JsonConvert.SerializeObject(payloadForDroppedRntbdRequests);
```
ealsur, line 59: Is this JSON the contract the client telemetry endpoint expects? Why are the other properties just key/value pairs but this is a complex object? Why not "dropped requests": 10 for example?

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/NetworkRequestSampler.cs
sourabh1007: I think top 100 replicas should be enough to give us the overall behavior.

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/NetworkRequestSampler.cs
sourabh1007: Please ref. this for the properties used in hash https://github.com/Azure/azure-cosmos-dotnet-v3/blob/d41f0995618cb344fcaf3c16e002960493aaa016/Microsoft.Azure.Cosmos/src/Telemetry/Models/RequestInfo.cs#L38.
I don't think it really matters; sampler is short lived object with HashSet having maximum 100 objects only.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: yes, Idea is to limit number of network calls made by an operation, for example cross partition query on a container with 1000 of partitions and replicas. So in that case we don't to store everything, we just want to store top 100 replica (or service endpoint) failures.

cc @FabianMeiswinkel

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
sourabh1007: `OperationInfo `and `CacheRefreshInfo`, these 2 objects are part of initial design. hence, for each kind of metrics , there will be a separate JSON object.
e.g.

```
"operationInfo": [
    {
      "regionsContacted": "South Central US",
      "greaterThan1Kb": false,
      "databaseName": "10f50f2a-de1f-4aad-8754-d35cca3961a4",
      "containerName": "03183a28-8ab1-4266-8095-6afc8b826188",
      "operation": "Read",
      "resource": "Document",
      "statusCode": 200,
      "metricInfo": {
        "metricsName": "RequestLatency", //metrics 1
        "unitName": "MilliSecond",
        "mean": 2.369,
        "count": 1,
        "min": 2.369,
        "max": 2.369,
        "percentiles": {
          "50": 2.369,
          "90": 2.369,
          "95": 2.369,
          "99": 2.369,
          "99.9": 2.369
        }
      }
    },
    {
      "regionsContacted": "South Central US",
      "greaterThan1Kb": false,
      "databaseName": "10f50f2a-de1f-4aad-8754-d35cca3961a4",
      "containerName": "03183a28-8ab1-4266-8095-6afc8b826188",
      "operation": "Read",
      "resource": "Document",
      "statusCode": 200,
      "metricInfo": {
        "metricsName": "RequestCharge", //metrics 2
        "unitName": "RU",
        "mean": 2.0,
        "count": 1,
        "min": 2.0,
        "max": 2.0,
        "percentiles": {
          "50": 2.0,
          "90": 2.0,
          "95": 2.0,
          "99": 2.0,
          "99.9": 2.0
        }
      }
    }]
```
So, following the same pattern, I added 3rd new metrics.

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/NetworkRequestSampler.cs
sourabh1007: Discussed it offline with @FabianMeiswinkel ...refactoring code to take list of data as input and return filtered results.

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/NetworkRequestSampler.cs
ealsur: We are using the HashSet just as a distinct+count, we are not using any of the objects inside it, why not just minimize the memory footprint? Sure, it has a max of 100 objects but this is per-operation, the volume of operations can be huge?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
ealsur: The case of the cross-partition query is interesting. Because each partition might be a different ReadNext call. Did we test that on a cross-partition query, we only end up creating a single sampler? (just to confirm the hypothesis)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTests.cs
```diff
@@ -68,7 +68,7 @@ public void TestInitialize()
                         HttpResponseMessage result = new HttpResponseMessage(HttpStatusCode.OK);
 
                         string jsonObject = request.Content.ReadAsStringAsync().GetAwaiter().GetResult();
-
+                        Console.WriteLine(jsonObject);
```
ealsur, line 71: nit: Remove Console output

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/NetworkRequestSampler.cs
sourabh1007: redesigned the logic. Now you will see different sampler

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: Redesign the logic to apply sampling on overall collected data irrespective of operation type

FabianMeiswinkel (commented):
Overall I like the decision making for the sampler. Have a few implementation comments - and definitely want Matias and or Kiran at least to also take a look

kirankumarkolli:
> Minor change: added 412 status code into excluded status code list. So basically telemetry won't be collected for network calls with statuscode 412 and substatuscode 0.

What's the reason for exclude?

sourabh1007:
> > Minor change: added 412 status code into excluded status code list. So basically telemetry won't be collected for network calls with statuscode 412 and substatuscode 0.
> 
> What's the reason for exclude?

412 status code is "Precondition Failure" which we don't want to record. It is one of the valid scenario which we don't want to track in client telemetry @FabianMeiswinkel

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/DataSamplerTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            /*stopWatch.Stop();
+
+            Console.WriteLine("non linq execution time " + stopWatch.ElapsedTicks + " " + Process.GetCurrentProcess().PrivateMemorySize64);
+                */
+            Assert.AreEqual(numberOfGroups * numberOfElementsInEachGroup, sampleDataByLatency.Count);
+
+          /*  stopWatch = Stopwatch.StartNew();
+            stopWatch.Start();
+            List<RequestInfo> sortedList = requestInfoList.GroupBy(r => new
+            {
```
sourabh1007, line 70: I will remove this commented code before merge

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/DataSampler.cs
```diff
@@ -0,0 +1,100 @@
...
+                // Check if similar object is already present
+                if (sampledRawData.TryGetValue(key, out List<KeyValuePair<double, RequestInfo>> sortedData))
+                {
+                    DataSampler.AddToList(orderBy, requestInfo, sortedData);
+
+                    sortedData.Sort(DataComparer.Instance);
+
+                    if (sortedData.Count > ClientTelemetryOptions.NetworkRequestsSampleSizeThrehold)
+                    {
+                        sortedData.RemoveAt(sortedData.Count - 1);
```
ealsur, line 40: not sure I understand this one. Let's say sortedDate has 10 and limit is 5, we are removing just 1? Shouldn't we remove 5?

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/DataSampler.cs
```diff
@@ -0,0 +1,100 @@
...
+                    }
+                    sampledRawData.Remove(key);
```
ealsur, line 42: Instead of Removing and adding, wouldn't replacing achieve the same?

Microsoft.Azure.Cosmos/src/Telemetry/Models/RequestInfo.cs
```diff
@@ -62,5 +68,26 @@ public override bool Equals(object obj)
...
+                {
+                    return metric.Percentiles[ClientTelemetryOptions.Percentile99];
+                }
+            }
+            return Double.MinValue; // least prioity for request info w/o latency info
+        }
+
+        public double GetSampleCount()
+        {
+            return (double)this.Metrics[0].Count;
```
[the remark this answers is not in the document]
sourabh1007: Sample Count will be same for all the metrics. So picked up first metrics from the list (doesn't matter what kind of metrics it is). 
Since it is a list of all metrics, wouldn't know at which position I can find it. Hence for p99 latency I have loop.

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/DataSampler.cs
sourabh1007: here I am removing the last element of the list because list is already crossed the threshold size and I added extra object just before sorting, to make sure I am removing the object with lowest latency/samplecount.

Microsoft.Azure.Cosmos/src/Telemetry/NetworkDataRecorder.cs
```diff
@@ -0,0 +1,145 @@
...
+                    Resource = storeResponseStatistic.RequestResourceType.ToString(),
+                    Operation = storeResponseStatistic.RequestOperationType.ToString(),
+                };
+
+        }
+
+        public List<RequestInfo> GetRequests()
+        {
+            List<RequestInfo> requestInfoList = new List<RequestInfo>();
+            requestInfoList.AddRange(this.GetErroredRequests());
```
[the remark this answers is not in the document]
sourabh1007: did use enumerator but tried to avoid list copy.

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/DataSampler.cs
sourabh1007: Couldn't find replace in dictionary. am I missing something?

Microsoft.Azure.Cosmos/src/Telemetry/Models/RequestInfo.cs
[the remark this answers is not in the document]
ealsur: Is the list always ensured to have at least 1 item? Any change of this line leading to a Null Ref or KeyNotFound?

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/DataSampler.cs
ealsur: sampleRawData[key] = value;

This creates or updates

Microsoft.Azure.Cosmos/src/Telemetry/Models/RequestInfo.cs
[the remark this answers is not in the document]
sourabh1007: Yes we are going to capture at least metrics always and double casting it required to use same data collection in Sampler logic othwerwise somewhere I have to mention that, if "order by latency" then deal with double otherwise deal with "long" which was making sampler logic complicated. Hence typecast to same type i.e double

ealsur (dismissed):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/DataSamplerTests.cs
sourabh1007: done

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/DataSampler.cs
```diff
@@ -0,0 +1,99 @@
...
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Microsoft.Azure.Cosmos.Query.Core.Pipeline.CrossPartition.OrderBy;
+    using Microsoft.Azure.Cosmos.Telemetry.Models;
+
+    /// <summary>
+    /// Sampler to select top N unique records and return true/false on the basis of elements already selected.
+    /// </summary>
+    internal sealed class DataSampler
```
FabianMeiswinkel, line 17: This class needs some more love - I claim to usually be relatively good at reading other people's code - but I can hardly grasp what you are doing here - or why you are doing it this way? Why obvious inefficiencies like sorting an already presorted list? Any insertion seems to be O(log(n)) + O(n) for moving items within the ArrayList - simple checks whether the current item is larger than the comparator at the SampleSize index are missing - etc. - in this class it would help to be more verbose with comments - what are you doing - why certain algorithms were used etc. would help making it more readable.

FabianMeiswinkel: An alternative would be to first only add to a list - then sort it each once - at least explain why you don't use that approach - which seems easier and even better in perf (maybe not memory consumption - and maybe that is the explanation)?

FabianMeiswinkel (changes requested):
Please look at the comments in DataSampler class

Microsoft.Azure.Cosmos/src/Telemetry/Sampler/DataSampler.cs
sourabh1007: Yes I agree I little overcomplicated it. Revisited the logic according to your suggestion.

FabianMeiswinkel (approved):
LGTM now.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryOptions.cs
```diff
@@ -88,15 +88,18 @@ internal static class ClientTelemetryOptions
 
         internal static readonly ResourceType AllowedResourceTypes = ResourceType.Document;
         // Why 5 sec? As of now, if any network request is taking more than 5 millisecond sec, we will consider it slow request this value can be revisited in future
-        private static readonly TimeSpan NetworkLatencyThreshold = TimeSpan.FromMilliseconds(5);
+        internal static readonly TimeSpan NetworkLatencyThreshold = TimeSpan.FromMilliseconds(5);
```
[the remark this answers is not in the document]
sourabh1007: yes it is intentional. in later PRs it will be configurable then we can definitely debate on this value.

PR #3754, opened 2023-03-13
[Internal] Query: Adds OptimisticDirectExecute and RequiresDistribution headers to SDK

akotalwar:
# Pull Request Template

## Description

This PR introduces two new boolean headers for the client. One is the OptimisticDirectExecute header, which is on the request message. It indicates to the backend whether the query can use the OptimisticDirectExecute pipeline or not.  The second header is the RequiresDistribution header and is on the response message. It lets the client know whether the query requires distribution. 

We use the OptimisticDirectExecute pipeline to return results when either of these conditions is true: the query does not require distribution, or the query requires distribution but there is no continuation token. Otherwise, we switch to the specialized pipeline to handle the remaining results if the query requires distribution and there is a continuation token.

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/RequestOptions/InternalRequestOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    using System;
+    using static Microsoft.Azure.Documents.RuntimeConstants;
+
+    /// <summary>
+    /// Internal Cosmos query request options to pass between classes/functions
+    /// </summary>
+    internal class InternalRequestOptions
+    {
+        internal bool OptimisticDirectExecute { get; set; }
```
adityasa, line 15: Please add summary for each.

adityasa, line 15: >set [](http://example.com/codeflow?start=53&length=3)

Please remove setters and have a constructor that forces each parameter.

Microsoft.Azure.Cosmos/src/Pagination/NetworkAttachedDocumentContainer.cs
```diff
@@ -239,6 +239,7 @@ internal sealed class NetworkAttachedDocumentContainer : IMonadicDocumentContain
             SqlQuerySpec sqlQuerySpec,
             FeedRangeState<QueryState> feedRangeState,
             QueryPaginationOptions queryPaginationOptions,
+            InternalRequestOptions internalRequestOptions,
```
adityasa, line 242: It's a bit confusing to have
1. QueryPaginationOptions
2. InternalRequestOptions
3. QueryRequestOptions (below)
that kind of do the same thing - turn options into headers.

How can we merge these three effectively?

akotalwar: We have updated the name to AdditionalRequestHeaders for now.

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        internal string ContentType { get; }
+
+        // Boolean to let the backend know if a given operation is a query or not.
+        internal bool IsQuery { get; }
+
+        // GUID to enable backend to link multiple activityIds that belong to the same operation.
+        internal Guid CorrelatedActivityId { get; }
+
+        public AdditionalRequestHeaders(bool optimisticDirectExecute = false, bool isContinuationExpected = false, Guid correlatedActivityId = new Guid(), bool isQuery = true, string contentType = MediaTypes.QueryJson)
```
adityasa, line 29: >new Guid() [](http://example.com/codeflow?start=143&length=10)

This doesn't seem right.

adityasa, line 29: > = true [](http://example.com/codeflow?start=167&length=7)

Same. Is this class limited to query codepath?

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            this.OptimisticDirectExecute = optimisticDirectExecute;
+            this.IsContinuationExpected = isContinuationExpected;
```
adityasa, line 32: Please assert that ode is true only if isQuery is true as well.

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    using System;
+    using static Microsoft.Azure.Documents.RuntimeConstants;
+
+    /// <summary>
+    /// Internal Cosmos headers to pass between classes/functions
+    /// </summary>
+    internal sealed class AdditionalRequestHeaders
+    {
+        // Boolean to let the backend know if a query is utilizing the OptimisticDirectExecute pipeline.
```
adityasa, line 15: >// [](http://example.com/codeflow?start=8&length=2)

Since you already have a comment, please use C#'s documentation style comments for fields
///

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,38 @@
...
+        // Boolean to let the backend know if continuations are expected for a certain query.
+        internal bool IsContinuationExpected { get; }
+
+        internal string ContentType { get; }
+
+        // Boolean to let the backend know if a given operation is a query or not.
+        internal bool IsQuery { get; }
+
+        // GUID to enable backend to link multiple activityIds that belong to the same operation.
+        internal Guid CorrelatedActivityId { get; }
```
adityasa, line 27: Please sort alphabetically.

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,38 @@
...
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Query.Core
+{
+    using System;
+    using static Microsoft.Azure.Documents.RuntimeConstants;
+
+    /// <summary>
+    /// Internal Cosmos headers to pass between classes/functions
+    /// </summary>
```
adityasa, line 12: Please consider rephrasing.

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
adityasa, line 15: >Boolean [](http://example.com/codeflow?start=11&length=7)

No need to specify the type.

adityasa: You can refer to existing .NET framework/library class documentation for style, format and language.

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,38 @@
...
+    internal sealed class AdditionalRequestHeaders
+    {
+        // Boolean to let the backend know if a query is utilizing the OptimisticDirectExecute pipeline.
+        internal bool OptimisticDirectExecute { get; }
+
+        // Boolean to let the backend know if continuations are expected for a certain query.
+        internal bool IsContinuationExpected { get; }
```
adityasa, line 19: Same as above.

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
adityasa, line 27: Same as above.

Microsoft.Azure.Cosmos/src/Pagination/NetworkAttachedDocumentContainer.cs
```diff
@@ -258,16 +259,17 @@ internal sealed class NetworkAttachedDocumentContainer : IMonadicDocumentContain
             }
 
             QueryRequestOptions queryRequestOptions = this.queryRequestOptions == null ? new QueryRequestOptions() : this.queryRequestOptions;
+            AdditionalRequestHeaders newAdditionalRequestHeaders = new AdditionalRequestHeaders(additionalRequestHeaders.OptimisticDirectExecute, isContinuationExpected: false, this.correlatedActivityId, additionalRequestHeaders.IsQuery, additionalRequestHeaders.ContentType);
```
adityasa, line 262: Please assert this that incoming additionalRequestHeaders.IsContinuationExpected == false and additionalRequestHeaders.CorrelatedActivityId == Guid.Empty

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        internal string ContentType { get; }
+
+        // Boolean to let the backend know if a given operation is a query or not.
+        internal bool IsQuery { get; }
+
```
adityasa, line 25: Is this ever set?

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
```diff
@@ -143,13 +142,14 @@ internal class CosmosQueryClientCore : CosmosQueryClient
...
-                        isContinuationExpected.ToString());
+                        additionalRequestHeaders.IsContinuationExpected.ToString());
                     QueryRequestOptions.FillContinuationToken(
                         cosmosRequestMessage,
                         continuationToken);
-                    cosmosRequestMessage.Headers.Add(HttpConstants.HttpHeaders.ContentType, MediaTypes.QueryJson);
-                    cosmosRequestMessage.Headers.Add(HttpConstants.HttpHeaders.IsQuery, bool.TrueString);
-                    cosmosRequestMessage.Headers.Add(WFConstants.BackendHeaders.CorrelatedActivityId, clientQueryCorrelationId.ToString());
+                    cosmosRequestMessage.Headers.Add(HttpConstants.HttpHeaders.ContentType, additionalRequestHeaders.ContentType);
+                    cosmosRequestMessage.Headers.Add(HttpConstants.HttpHeaders.IsQuery, additionalRequestHeaders.IsQuery.ToString());
```
adityasa, line 150: >IsQuery [](http://example.com/codeflow?start=113&length=7)

This is never set, so will always be false.

adityasa, line 150: >IsQuery [](http://example.com/codeflow?start=79&length=7)

Since this is always true, I would remove it from AdditionalRequestHeaders.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -60,6 +61,34 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
             if (success && !isPartitionSplitException)
             {
                 this.continuationToken = this.Current.Succeeded ? this.Current.Result.State?.Value : null;
+                if (this.Current.Succeeded)
+                {
+                    this.Current.Result.AdditionalHeaders.TryGetValue("x-ms-cosmos-query-requiresdistribution", out string requiresDistribution);
```
adityasa, line 66: >"x-ms-cosmos-query-requiresdistribution" [](http://example.com/codeflow?start=70&length=40)

Friendly reminder to use a const after direct package is updated.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -33,6 +33,7 @@ private enum ExecutionState
         private TryCatch<IQueryPipelineStage> inner;
         private CosmosElement continuationToken;
         private ExecutionState executionState;
+        private bool? previousRequiresDistribution;
```
adityasa, line 36: Please remove, since it doesn't seem like you need this field for anything other than the assert.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -60,6 +61,34 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
             if (success && !isPartitionSplitException)
             {
                 this.continuationToken = this.Current.Succeeded ? this.Current.Result.State?.Value : null;
+                if (this.Current.Succeeded)
```
adityasa, line 64: >this.Current.Succeeded [](http://example.com/codeflow?start=20&length=22)

Doesn't success imply this.Current.Succeeded?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -60,6 +61,34 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                    {
+                        bool queryRequiresDistribution = bool.Parse(requiresDistribution);
+                        if (this.previousRequiresDistribution != null)
+                        {
+                            Debug.Assert(this.previousRequiresDistribution == queryRequiresDistribution, "OptimisticDirectExecuteQueryPipelineStage Assert!", "RequiresDistribution flag cannot switch midway through execution");
+                        }
+                        else
+                        { 
+                            this.previousRequiresDistribution = queryRequiresDistribution;
+                        }
```
adityasa, line 77: Please rewrite without needing previousRequiresDistribution. You can use this.executionState instead.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -60,6 +61,34 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+
+                        if (queryRequiresDistribution && this.continuationToken != null)
```
adityasa, line 79: >this.continuationToken != null [](http://example.com/codeflow?start=57&length=30)

Please do this check outside. We don't need to parse requiresdistribution header if there is no continuation.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -60,6 +61,34 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
             if (success && !isPartitionSplitException)
             {
                 this.continuationToken = this.Current.Succeeded ? this.Current.Result.State?.Value : null;
+                if (this.Current.Succeeded)
+                {
+                    this.Current.Result.AdditionalHeaders.TryGetValue("x-ms-cosmos-query-requiresdistribution", out string requiresDistribution);
+                    if (requiresDistribution != null)
+                    {
```
adityasa, line 68: I think we should assert that this is always present?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
adityasa, line 64: We don't need to check this, only the continuation token.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
adityasa, line 66: >requiresDistribution [](http://example.com/codeflow?start=123&length=20)

requiresDistributionHeaderValue

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -60,6 +61,34 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                        bool queryRequiresDistribution = bool.Parse(requiresDistribution);
```
adityasa, line 69: >queryRequiresDistribution [](http://example.com/codeflow?start=29&length=25)

requiresDistribution

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -60,6 +61,34 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                        if (queryRequiresDistribution && this.continuationToken != null)
+                        {
+                            this.inner = await this.queryPipelineStageFactory(null);
+                            this.executionState = ExecutionState.SpecializedDocumentQueryExecution;
+                            if (this.inner.Failed)
+                            {
+                                return false;
+                            }
+
+                            success = await this.inner.Result.MoveNextAsync(trace);
```
adityasa, line 88: Please create a function for this and call it from here and line 95-102.

adityasa:
>             }

I think we have an existing bug/opportunity for improvement. All of this highlighted code needs to happen only if ExecutionState is ODE.

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs:103 in 992be1b. [](commit_id = 992be1bf500d2f298fd6a3b6451d6c79891ab010, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Pagination/IMonadicQueryDataSource.cs
```diff
@@ -16,6 +16,7 @@ internal interface IMonadicQueryDataSource
             SqlQuerySpec sqlQuerySpec,
             FeedRangeState<QueryState> feedRangeState,
             QueryPaginationOptions queryPaginationOptions,
+            AdditionalRequestHeaders additionalRequestHeaders,
             ITrace trace,
```
adityasa, line 20: 1. Most of the callers to MonadicQueryAsync provide an empty AdditionalRequestHeaders.
2. You need this only for ODE and none of the additional headers.
3. The AdditionalRequestHeaders get "overwritten" anyways before calling ExecuteItemQueryAsync.

For these reasons let's to turn this into a simple boolean with a default value of false. By convention CancellationToken is last parameters, so you will need to add default value of null to it as well.

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
akotalwar: It is never set, but its default value is true so it would always be true

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
akotalwar: We cannot assert that requiresDistribution will always be there because in the case of partition split, fallback pipeline goes through this code too and in that case there would be no requiredDistribution flag.

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
akotalwar: Yes, I believe it is

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -60,6 +61,34 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
                 this.continuationToken = this.Current.Succeeded ? this.Current.Result.State?.Value : null;
+                if (this.Current.Succeeded)
+                {
+                    this.Current.Result.AdditionalHeaders.TryGetValue("x-ms-cosmos-query-requiresdistribution", out string requiresDistribution);
+                    if (requiresDistribution != null)
+                    {
+                        bool queryRequiresDistribution = bool.Parse(requiresDistribution);
+                        if (this.previousRequiresDistribution != null)
+                        {
+                            Debug.Assert(this.previousRequiresDistribution == queryRequiresDistribution, "OptimisticDirectExecuteQueryPipelineStage Assert!", "RequiresDistribution flag cannot switch midway through execution");
```
akotalwar, line 72: >Debug.Assert(this.previousRequiresDistribution == queryRequiresDistribution, "Op [](http://example.com/codeflow?start=28&length=80)

Throw InvalidOperationException

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <summary>
+    /// Capturing additional headers to be sent in the web request.
+    /// </summary>
+    internal sealed class AdditionalRequestHeaders
+    {
+        /// <summary>
+        /// GUID to enable backend to link multiple activityIds that belong to the same operation.
+        /// </summary>
+        internal Guid CorrelatedActivityId { get; }
```
sboshra, line 18: >internal [](http://example.com/codeflow?start=8&length=8)

public

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -57,20 +58,39 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                        this.Current.Result.AdditionalHeaders.TryGetValue("x-ms-cosmos-query-requiresdistribution", out string requiresDistributionHeaderValue);
+                        Debug.Assert(!string.IsNullOrEmpty(requiresDistributionHeaderValue), "OptimisticDirectExecuteQueryPipelineStage Assert!", "Missing requiresDistribution flag in backend response for OptimisticDirectExecute request");
+                        
+                        bool requiresDistribution = bool.Parse(requiresDistributionHeaderValue);
+                        if (this.previousRequiresDistribution != null)
+                        {
+                            if (this.previousRequiresDistribution != requiresDistribution)
+                            {
+                                throw new InvalidOperationException("RequiresDistribution flag cannot switch midway through execution");
+                            }
```
sboshra, line 77: please add a comment that this should never happen - basically, this is more of an asset than an exception

sboshra (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -57,20 +58,39 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
             bool success = hasNext.Succeeded && hasNext.Result;
             bool isPartitionSplitException = hasNext.Succeeded && this.Current.Failed && this.Current.InnerMostException.IsPartitionSplitException();
```
adityasa, line 59: >bool isPartitionSplitException = hasNext.Succeeded && this.Current.Failed && this.Current.InnerMostException.IsPartitionSplitException(); [](http://example.com/codeflow?start=12&length=137)

This can be inside the if below.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -57,20 +58,39 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                    {
+                        this.Current.Result.AdditionalHeaders.TryGetValue("x-ms-cosmos-query-requiresdistribution", out string requiresDistributionHeaderValue);
+                        Debug.Assert(!string.IsNullOrEmpty(requiresDistributionHeaderValue), "OptimisticDirectExecuteQueryPipelineStage Assert!", "Missing requiresDistribution flag in backend response for OptimisticDirectExecute request");
+                        
+                        bool requiresDistribution = bool.Parse(requiresDistributionHeaderValue);
+                        if (this.previousRequiresDistribution != null)
+                        {
+                            if (this.previousRequiresDistribution != requiresDistribution)
+                            {
+                                throw new InvalidOperationException("RequiresDistribution flag cannot switch midway through execution");
```
adityasa, line 78: >RequiresDistribution flag cannot switch midway through execution [](http://example.com/codeflow?start=69&length=64)

Consider:
Unexpected switch in 'x-ms-cosmos-query-requiresdistribution' value. Previous value : '', Current value : ''.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -57,20 +58,39 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                        }
+                        else
+                        {
+                            this.previousRequiresDistribution = requiresDistribution;
+                        }
+
+                        if (requiresDistribution)
+                        {
+                            success = await this.CreateAndRunFallbackPipelineAsync(isPartitionSplitException: false, trace);
+                        }
```
adityasa, line 89: Simplify (line 72 - 87):

if ( this.previousRequiresDistribution.HasValue &&
this.previousRequiresDistribution != requiresDistribution)
{
     throw new ...
}

if(requiresDistribution)
{
   success = ...
}

this.previousRequiresDistribution = requiresDistribution;

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -94,6 +114,21 @@ private CosmosElement TryUnwrapContinuationToken()
             return null;
         }
 
+        private async Task<bool> CreateAndRunFallbackPipelineAsync(bool isPartitionSplitException, ITrace trace)
```
adityasa: >CreateAndRunFallbackPipelineAsync [](http://example.com/codeflow?start=33&length=33)

This name doesn't indicate that there's a side effect (on the state of the instance of this class i.e. it changes the executionState.

Consider
SwitchToFallbackPipelineAsync

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,37 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Query.Core
+{
+    using System;
+    using static Microsoft.Azure.Documents.RuntimeConstants;
+
+    /// <summary>
+    /// Capturing additional headers to be sent in the web request.
```
adityasa, line 11: >Capturing [](http://example.com/codeflow?start=8&length=9)

Captures

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Pagination/IMonadicQueryDataSource.cs
```diff
@@ -16,6 +16,7 @@ internal interface IMonadicQueryDataSource
             SqlQuerySpec sqlQuerySpec,
             FeedRangeState<QueryState> feedRangeState,
             QueryPaginationOptions queryPaginationOptions,
+            bool optimisticDirectExecute,
```
neildsh, line 19: There is no need to add a separate parameter here. You can just add this to `QueryPaginationOptions`

Microsoft.Azure.Cosmos/src/Query/Core/QueryClient/CosmosQueryClient.cs
```diff
@@ -54,12 +54,11 @@ internal abstract class CosmosQueryClient
             string resourceUri,
             Documents.ResourceType resourceType,
             Documents.OperationType operationType,
-            Guid clientQueryCorrelationId,
             FeedRange feedRange,
             QueryRequestOptions requestOptions,
+            AdditionalRequestHeaders additionalRequestHeaders,
```
neildsh, line 59: >AdditionalRequestHeaders additionalRequestHeaders, [](http://example.com/codeflow?start=12&length=50)

`QueryRequestOptions` already has a member `EnableOptimisticDirectExecution`. We should not need this redundant new `AdditionalRequestHeaders.OptimisticDirectExecute`

akotalwar: EnableOptimisticDirectExecute and OptimisticDirectExecute headers serve different purposes. 
EnableOptimisticDirectExecute lets the client know if it can use the Ode pipeline. This does not necessarily mean the Ode pipeline will be used (ex: cross partition query).
OptimisticDirectExecute is used to let the backend know if the Ode pipeline was used or not. 
Hence, just because EnableOptimisticDirectExecute is true does not mean that OptimisticDirectExecute will also be true.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Pagination/IMonadicQueryDataSource.cs
akotalwar: Based on previous discussions, it was decided to not add optimisticDirectExecute header to queryPaginationOptions as this header has nothing to do with Pagination

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -94,6 +110,21 @@ private CosmosElement TryUnwrapContinuationToken()
             return null;
         }
 
+        private async Task<bool> SwitchToFallbackPipelineAsync(bool isPartitionSplitException, ITrace trace)
+        {
+            this.executionState = ExecutionState.SpecializedDocumentQueryExecution;
```
adityasa, line 115: I would (Debug) assert on a line prior line that
executionState == ODE.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/OptimisticDirectExecutionQueryTests.cs
```diff
@@ -12,162 +12,247 @@
...
+                    enableOptimisticDirectExecution: true,
+                    pageSizeOptions: PageSizeOptions.NonGroupByAndNoContinuationTokenPageSizeOptions,
+                    expectedPipelineType: TestInjections.PipelineType.OptimisticDirectExecution),
+                CreateInput(
+                    query: $"SELECT DISTINCT VALUE r.{NumberField} FROM r ORDER BY r.{NumberField} DESC",
+                    expectedResult: first7IntegersReversed,
+                    partitionKey: null,
+                    enableOptimisticDirectExecution: true,
+                    pageSizeOptions: PageSizeOptions.NonGroupWithContinuationTokenPageSizeOptions,
+                    expectedPipelineType: TestInjections.PipelineType.Specialized),
```
adityasa, line 114: This is good. Can we also add variations of this without the order by ?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Pagination/InMemoryContainer.cs
```diff
@@ -1080,6 +1084,12 @@ static Task<TryCatch<Record>> CreateNotFoundException(CosmosElement partitionKey
 
         public IEnumerable<int> PartitionKeyRangeIds => this.partitionKeyRangeIdToHashRange.Keys;
 
+        public bool RequiresDistribution
+        {
+            get => this.requiresDistribution;
```
adityasa, line 1089: >requiresDistribution [](http://example.com/codeflow?start=24&length=20)

You can remove this field.
And replace the property definition as:

public bool RequiresDistribution { get; set; }

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -192,6 +192,145 @@ public async Task TestPipelineForBackendDocumentsOnSinglePartitionAsync()
...
+                new SwitchableRequiresDistributionTestCases("SELECT TOP 5 r.id FROM r WHERE r.id > 5 ORDER BY r.id", 0, 0),
+                new SwitchableRequiresDistributionTestCases("SELECT * FROM r OFFSET 5 LIMIT 3", 1, 3),
+                new SwitchableRequiresDistributionTestCases("SELECT * FROM r WHERE r.id > 5 OFFSET 5 LIMIT 3", 0, 0)
+            };
+
+            foreach (SwitchableRequiresDistributionTestCases testCase in singlePartitionContainerTestCases)
+            {
+                    OptimisticDirectExecutionTestInput input = CreateInput(
+                        description: @"Queries which can switch their requiresDist flag",
+                        query: testCase.Query,
```
adityasa, line 216: Please fix the description. These queries are not expected to have requiresDistribution = true.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -192,6 +192,145 @@ public async Task TestPipelineForBackendDocumentsOnSinglePartitionAsync()
...
+                new SwitchableRequiresDistributionTestCases("SELECT Count(1) as count FROM root r JOIN b IN r.id", 0, 1),
+                new SwitchableRequiresDistributionTestCases("SELECT Avg(1) AS avg FROM root r", 0, 1),
+                new SwitchableRequiresDistributionTestCases("SELECT r.id, Count(1) as count FROM r WHERE r.id > 0 GROUP BY r.id", 0, 0),
+                new SwitchableRequiresDistributionTestCases("SELECT r.id FROM r WHERE r.id > 0 GROUP BY r.id ORDER BY r.id", 0, 0)
+            };
+
+            foreach (SwitchableRequiresDistributionTestCases testCase in singlePartitionContainerTestCases)
+            {
+                OptimisticDirectExecutionTestInput input = CreateInput(
+                    description: @"Queries which can switch their requiresDist flag",
```
adityasa, line 255: Please update the description.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -192,6 +192,145 @@ public async Task TestPipelineForBackendDocumentsOnSinglePartitionAsync()
...
+        }
+
+        [TestMethod]
+        public async Task TestQueriesWhichWillAlwaysRequireDistribution()
+        {
+            // requiresDist = true
+            int numItems = 100;
+            List<SwitchableRequiresDistributionTestCases> singlePartitionContainerTestCases = new List<SwitchableRequiresDistributionTestCases>()
+            {
+                new SwitchableRequiresDistributionTestCases("SELECT Sum(id) as sum_id FROM r JOIN id IN r.id", 0, 1),
```
adityasa, line 239: >Switchable [](http://example.com/codeflow?start=20&length=10)

Consider dropping Switchable from the name to avoid confusion.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -192,6 +192,145 @@ public async Task TestPipelineForBackendDocumentsOnSinglePartitionAsync()
...
+                new SwitchableRequiresDistributionTestCases("SELECT DISTINCT TOP 5 r.id FROM r ORDER BY r.id", 0, 5),
+                new SwitchableRequiresDistributionTestCases("SELECT DISTINCT r.id FROM r GROUP BY r.id", 0,  10),
+                new SwitchableRequiresDistributionTestCases("SELECT DISTINCT r.id, Sum(r.id) as sum_a FROM r GROUP BY r.id",0, 10),
+                new SwitchableRequiresDistributionTestCases("SELECT Count(1) FROM (SELECT DISTINCT r.id FROM root r)", 0, 1),
+                new SwitchableRequiresDistributionTestCases("SELECT DISTINCT id FROM r JOIN id in r.id", 0, 0),
+                new SwitchableRequiresDistributionTestCases("SELECT r.id, Count(1) AS count_a FROM r GROUP BY r.id ORDER BY r.id", 0, 10),
+                new SwitchableRequiresDistributionTestCases("SELECT Count(1) as count FROM root r JOIN b IN r.id", 0, 1),
+                new SwitchableRequiresDistributionTestCases("SELECT Avg(1) AS avg FROM root r", 0, 1),
```
adityasa, line 247: >Avg [](http://example.com/codeflow?start=68&length=3)

Please double check why this always requiresDistribution.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -432,7 +577,13 @@ private static async Task<IQueryPipelineStage> GetOdePipelineAsync(OptimisticDir
                 Version = PartitionKeyDefinitionVersion.V2,
             };
 
-            IMonadicDocumentContainer monadicDocumentContainer = new InMemoryContainer(partitionKeyDefinition);
+            InMemoryContainer inMemoryContainer = new InMemoryContainer(partitionKeyDefinition);
+            if (requiresDist)
+            {
+                inMemoryContainer.RequiresDistribution = true; 
+            }
+
```
adityasa, line 585: Simplify.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -597,6 +743,22 @@ private static QueryRequestOptions GetQueryRequestOptions(bool enableOptimisticD
             }
             };
         }
+        internal readonly struct SwitchableRequiresDistributionTestCases
```
adityasa, line 746: Newline before would be nice.

adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Pagination/IMonadicQueryDataSource.cs
neildsh: `QueryPaginationOptions` already contains unrelated information such as `JsonSerializationFormat`, and it also takes a dictionary of additional headers

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -55,22 +56,37 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
-                this.executionState = ExecutionState.SpecializedDocumentQueryExecution;
-                if (this.inner.Failed)
+                bool isPartitionSplitException = hasNext.Succeeded && this.Current.Failed && this.Current.InnerMostException.IsPartitionSplitException();
+                if (success && !isPartitionSplitException)
                 {
-                    return false;
+                    this.continuationToken = this.Current.Succeeded ? this.Current.Result.State?.Value : null;
+                    if (this.continuationToken != null)
+                    {
+                        this.Current.Result.AdditionalHeaders.TryGetValue("x-ms-cosmos-query-requiresdistribution", out string requiresDistributionHeaderValue);
```
neildsh, line 67: >"x-ms-cosmos-query-requiresdistribution" [](http://example.com/codeflow?start=74&length=40)

Please use a constant (defined with all the other headers) rather than an inline literal

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -55,22 +56,37 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                        Debug.Assert(!string.IsNullOrEmpty(requiresDistributionHeaderValue), "OptimisticDirectExecuteQueryPipelineStage Assert!", "Missing requiresDistribution flag in backend response for OptimisticDirectExecute request");
+                        
+                        bool requiresDistribution = bool.Parse(requiresDistributionHeaderValue);
+                        if (this.previousRequiresDistribution.HasValue && this.previousRequiresDistribution != requiresDistribution)
+                        {
+                            // We should never enter this if statement as requiresDistribution flag can never switch mid execution.
+                            // Hence, this exception should never be thrown.
+                            throw new InvalidOperationException("Unexpected switch in 'x-ms-cosmos-query-requiresdistribution' value. Previous value : " + this.previousRequiresDistribution + " Current value : " + requiresDistribution + ".");
```
neildsh, line 75: >"Unexpected switch in 'x-ms-cosmos-query-requiresdistribution' value. Previous value : " + this.previousRequiresDistribution [](http://example.com/codeflow?start=64&length=124)

[Nit] The convention is to use interpolated strings:

`$"Previous value: {this.previousRequiresDistribution}"`

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -55,22 +56,37 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                    if (this.continuationToken != null)
+                    {
+                        this.Current.Result.AdditionalHeaders.TryGetValue("x-ms-cosmos-query-requiresdistribution", out string requiresDistributionHeaderValue);
+                        Debug.Assert(!string.IsNullOrEmpty(requiresDistributionHeaderValue), "OptimisticDirectExecuteQueryPipelineStage Assert!", "Missing requiresDistribution flag in backend response for OptimisticDirectExecute request");
+                        
+                        bool requiresDistribution = bool.Parse(requiresDistributionHeaderValue);
+                        if (this.previousRequiresDistribution.HasValue && this.previousRequiresDistribution != requiresDistribution)
+                        {
+                            // We should never enter this if statement as requiresDistribution flag can never switch mid execution.
+                            // Hence, this exception should never be thrown.
```
neildsh, line 74: [Nit] Consider removing comment. The exception message is clear enough.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
neildsh, line 36: >previousRequiresDistribution [](http://example.com/codeflow?start=22&length=28)

dont we need to set this value in the case where we hydrate the pipeline from a continuation token ?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Pagination/InMemoryContainer.cs
```diff
@@ -37,6 +37,7 @@ internal class InMemoryContainer : IMonadicDocumentContainer
     {
         private readonly PartitionKeyDefinition partitionKeyDefinition;
         private readonly Dictionary<int, (int, int)> parentToChildMapping;
+        private bool requiresDistribution;
```
neildsh, line 40: >private bool requiresDistribution; [](http://example.com/codeflow?start=8&length=34)

As discussed in past PRs, please do not add one-off test flags to this class. Please create a subclass internal to your test instead and override the `MonadicQueryAsync` method

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
akotalwar: Samer had mentioned that I add this comment during my PR review with him

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -117,6 +154,11 @@ private CosmosElement TryUnwrapContinuationToken()
             }
 
             OptimisticDirectExecutionQueryPipelineStage odePipelineStageMonadicCreate = new OptimisticDirectExecutionQueryPipelineStage(pipelineStage, fallbackQueryPipelineStageFactory, inputParameters.InitialUserContinuationToken);
+            if (inputParameters.InitialUserContinuationToken != null)
+            {
+                odePipelineStageMonadicCreate.previousRequiresDistribution = false;
```
neildsh, line 159: >previousRequiresDistribution = false; [](http://example.com/codeflow?start=46&length=37)

Please consider passing this to the constructor

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -55,22 +58,37 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                    {
+                        this.Current.Result.AdditionalHeaders.TryGetValue(HttpConstants.HttpHeaders.RequiresDistribution, out string requiresDistributionHeaderValue);
+                        Debug.Assert(!string.IsNullOrEmpty(requiresDistributionHeaderValue), "OptimisticDirectExecuteQueryPipelineStage Assert!", "Missing requiresDistribution flag in backend response for OptimisticDirectExecute request");
+                        
+                        bool requiresDistribution = bool.Parse(requiresDistributionHeaderValue);
+                        if (this.previousRequiresDistribution.HasValue && this.previousRequiresDistribution != requiresDistribution)
+                        {
+                            // We should never enter this if statement as requiresDistribution flag can never switch mid execution.
+                            // Hence, this exception should never be thrown.
+                            throw new InvalidOperationException($"Unexpected switch in 'x-ms-cosmos-query-requiresdistribution' value. Previous value : {this.previousRequiresDistribution} Current value : {requiresDistribution}.");
```
neildsh, line 77: >'x-ms-cosmos-query-requiresdistribution' [](http://example.com/codeflow?start=87&length=40)

Please use the same header constant that you use on line 69

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
neildsh: I see. Then consider fixing the comment by mentioning the circumstances that can cause the value to flip

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -28,11 +29,13 @@ private enum ExecutionState
             SpecializedDocumentQueryExecution,
         }
 
+        private const string optimisticDirectExecute = "OptimisticDirectExecute";
```
neildsh, line 32: >private const string optimisticDirectExecute = "OptimisticDirectExecute"; [](http://example.com/codeflow?start=8&length=73)

Please define the header name in HttpConstants just like the other header. Also, the convention is for header names to start with `x-ms-` or `x-ms-cosmos` or `x-ms-documentdb`

neildsh:
> //------------------------------------------------------------

Please move this file to `Microsoft.Azure.Cosmos\src\Query\Core\` so that the path matches the namespace

---
Refers to: Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs:1 in e07bf58. [](commit_id = e07bf58bb3395c0c031ae6e7bc8baa5a2f2b2f6d, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Pagination/NetworkAttachedDocumentContainer.cs
```diff
@@ -29,6 +29,7 @@ namespace Microsoft.Azure.Cosmos.Pagination
 
     internal sealed class NetworkAttachedDocumentContainer : IMonadicDocumentContainer
     {
+        private const string optimisticDirectExecute = "OptimisticDirectExecute";
```
neildsh, line 32: >private const string optimisticDirectExecute = "OptimisticDirectExecute"; [](http://example.com/codeflow?start=8&length=73)

This member is redefined in `OptimisticDirectExecutionQueryPipelineStage` . Please move to a common HttpHeaders location and reference in both places

Microsoft.Azure.Cosmos/src/RequestOptions/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// Let the backend know if continuations are expected for a certain query.
+        /// </summary>
+        public bool IsContinuationExpected { get; }
+
+        /// <summary>
+        /// Let the backend know if a query is utilizing the OptimisticDirectExecute pipeline.
+        /// </summary>
+        public bool OptimisticDirectExecute { get; }
+
+        public AdditionalRequestHeaders(Guid correlatedActivityId = default, bool isContinuationExpected = false, bool optimisticDirectExecute = false)
```
neildsh, line 30: >default, bool isContinuationExpected = false, boo [](http://example.com/codeflow?start=68&length=49)

Please remove default arguments since they are not being used in product code. default arguments are somewhat error prone so it is better to be explicit at call sites.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -126,6 +120,14 @@ public void NegativeOptimisticDirectExecutionOutput()
 
             List<OptimisticDirectExecutionTestInput> testVariations = new List<OptimisticDirectExecutionTestInput>
             {
+                // Test case has been commented out due to an existing bug in the code (https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2030)
```
neildsh, line 123: >(https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2030) [](http://example.com/codeflow?start=87&length=61)

I fixed this last month, so you should be able to uncomment the test case

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -598,6 +738,68 @@ private static QueryRequestOptions GetQueryRequestOptions(bool enableOptimisticD
             };
         }
 
+        internal readonly struct RequiresDistributionTestCases
```
neildsh, line 741: >RequiresDistributionTestCases [](http://example.com/codeflow?start=33&length=29)

[Nit]  `RequiresDistributionTestCase` singular, not plural

Microsoft.Azure.Cosmos/src/Query/Core/AdditionalRequestHeaders.cs
```diff
@@ -0,0 +1,37 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Query.Core
+{
+    using System;
+    using static Microsoft.Azure.Documents.RuntimeConstants;
```
neildsh, line 8: Please remove unused import

neildsh (dismissed):

PR #3636, opened 2023-01-09
CosmosClientOptions: Adds ServerCertificateCustomValidationCallback for Http and TCP

aavasthy:
# Pull Request Template

## Description
[Issue 3336](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3336): **Parameterize server ssl cert name override in direct connectivity mode**
In scenarios where the client service cannot take a dependency on Dns (i.e connecting to cosmosDb server using IP).
The ssl handshake with direct connectivity fails since url doesn't match the SN/SAN of the server certificate.

## Solution
Adding ServerCertificateCustomValidationCallback contract field for HTTP and TCP SSL callbacks. 
Include the callback field in StoreClientFactory in direct package 3.30.0 version via this PR https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB/pullrequest/951432

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Closing issues

To automatically close an issue: closes #3336

PR #3762, opened 2023-03-14
HttpClient: Adds detection of DNS changes through use of SocketsHttpHandler for .NET 6 and above

### LINKED ISSUE #3658: Ensure DNS changes are detected by client
Background: https://azure.status.microsoft/en-us/status/history/ Tracking ID: VN11-JD8

Reference: https://www.meziantou.net/avoid-dns-issues-with-httpclient-in-dotnet.htm

Azure SDK reference: https://github.com/Azure/azure-sdk-for-net/issues/19457

Azure SDK has followed this article and added the default configuration to HttpClient on https://github.com/Azure/azure-sdk-for-net/pull/19524

More code pointers:

* https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs#L146-L157
* https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/ServicePointHelpers.cs#L61

We should evaluate if we want to follow this approach, the Azure SDK went through these changes so we can assume they are ok, but it's worth investigating.

There is the alternative that none of these APIs are available for NETStandard 2.0

NaluTripician:
# Pull Request Template

## Description

When creating the `HttpMessageHandler` we now use Reflection to see if it is possible to create a `SocketsHttpHandler` (only available in .NET 6.0 and above). If not, the SDK will default to the behavior it was using previously. We also set the `PooledConnectionLifetime` to a random time between 5 minutes and 5 minutes and 30 seconds which will ensure that DNS changes will be detected after a maximum of that time span while also ensuring that a large number of connections are closed at the same time on a cluster of multiple machines.  

## Type of change

## Closing issues

To automatically close an issue: closes #3658


Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -101,29 +100,74 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+                {
+                    PropertyInfo maxConnectionsPerSercerInfo = socketHandlerType.GetProperty("MaxConnectionsPerServer");
+                    maxConnectionsPerSercerInfo.SetValue(socketHttpHandler, gatewayModeMaxConnectionLimit, null);
+
+                    if (serverCertificateCustomValidationCallback != null)
+                    {
+                        Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> customCertificateValidationCallback =
+                            (_, certificate2, x509Chain, sslPolicyErrors) => serverCertificateCustomValidationCallback(certificate2, x509Chain, sslPolicyErrors);
+
+                        PropertyInfo serverCertificateCustomValidationCallbackInfo = socketHandlerType.GetProperty("ServerCertificateCustomValidationCallback");
```
ealsur, line 131: Use https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs#L686-L691

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -101,29 +100,74 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
-            HttpClientHandler httpClientHandler = new HttpClientHandler();
-
-            // Proxy is only set by users and can cause not supported exception on some platforms
-            if (webProxy != null)
-            {
-                httpClientHandler.Proxy = webProxy;
-            }
-
-            // https://docs.microsoft.com/en-us/archive/blogs/timomta/controlling-the-number-of-outgoing-connections-from-httpclient-net-core-or-full-framework
+            HttpClientHandler httpClientHandler;
```
ealsur, line 103: not needed

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -101,29 +100,74 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+                catch (PlatformNotSupportedException)
+                {
+                }
+
+                return (HttpMessageHandler)socketHttpHandler;
+
             }
-            // MaxConnectionsPerServer is not supported on some platforms.
-            catch (PlatformNotSupportedException)
+            catch
```
ealsur, line 145: This is a big catch, and can be catching not only the `SocketsHttpHandler` not existing but then also other failures. Which are the pros and cons? Should we first test for `SocketsHttpHandler` and then have a dedicated method that takes care of the other properties if it exists and if something there fails, fail entirely?

ealsur: For example, have 2 methods, one that creates an HttpClientHandler and the other for SocketsHttpHandler.

Decide which to call based on the fact that reflection can instantiate SocketsHttpHandler or not.

Later on, when we multi target, we simply change the reflection check to a `#if NET6` build constants

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -87,7 +87,7 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
                     receivedResponseEventArgs);
             }
 
-            HttpClient httpClient = new HttpClient(httpMessageHandler);
+            HttpClient httpClient = new HttpClient(httpMessageHandler, disposeHandler: false);
```
ealsur, line 90: Don't do this. This should only be done if the user is injecting their own HttpClient.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -98,76 +98,101 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
-        public static HttpMessageHandler CreateHttpClientHandler(int gatewayModeMaxConnectionLimit, IWebProxy webProxy, Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
+        public static HttpMessageHandler CreateHttpClientHandler(
+            int gatewayModeMaxConnectionLimit, 
+            IWebProxy webProxy, 
+            Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
         {
-            HttpClientHandler httpClientHandler;
             try
             {
                 Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
```
ealsur, line 108: Check if null? Is this enough to know if the instance can be created?

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -98,76 +98,101 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+                return createSocketsHttpHandler(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+            }
+            catch
+            {
+                return createHttpClientHandler(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+            }
+        }
 
-                object socketHttpHandler = Activator.CreateInstance(socketHandlerType);
+        private static HttpMessageHandler createSocketsHttpHandler(
```
ealsur, line 117: nit: method naming should be upper camel case

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -98,76 +98,101 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+        public static HttpMessageHandler CreateHttpClientHandler(
+            int gatewayModeMaxConnectionLimit, 
+            IWebProxy webProxy, 
+            Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
         {
-            HttpClientHandler httpClientHandler;
             try
             {
                 Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+                return createSocketsHttpHandler(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
```
ealsur, line 109: When calling a static method, prefix with the class name

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -74,6 +75,9 @@ public async Task ValidateExceptionOnInitTask()
                 Assert.IsFalse(object.ReferenceEquals(ex, cosmosException1));
                 Assert.IsTrue(httpCallCount > 0);
             }
+
+            //Test cleanup
+            cosmosClient.Dispose();
```
ealsur, line 80: This method is creating the SDK with `using CosmosClient...` dispose is not needed

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -338,6 +348,9 @@ public async Task ResourceResponseStreamingTest()
                 {
                     Console.WriteLine("Expected exception while deserializing Resource: " + ex.Message);
                 }
+                
+                //Test cleanup
+                client.Dispose();
```
ealsur, line 353: Ditto. `using` will dispose the client

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -907,6 +932,50 @@ public async Task HttpClientConnectionLimitTest()
...
+                accountEndpoint: "https://localhost:8081",
+                authKeyOrResourceToken: Convert.ToBase64String(Encoding.UTF8.GetBytes(Guid.NewGuid().ToString())));
+
+            //Assert type of message handler
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+            Type clientMessageHandlerType = cosmosClient.ClientContext.DocumentClient.httpClient.HttpMessageHandler.GetType();
+            Assert.AreEqual(socketHandlerType, clientMessageHandlerType);
+
+            //Test cleanup
+            cosmosClient.Dispose();
```
ealsur, line 943: DIspose + using: One or the other, not both

NaluTripician: Will check rest of disposes and update accordingly.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+
+            if (socketHandlerType != null)
+            {
+                return CosmosHttpClientCore.CreateSocketsHttpHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+            }
+            
+            return CosmosHttpClientCore.CreateHttpClientHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+        }
+
+        private static HttpMessageHandler CreateSocketsHttpHandlerHelper(
```
ealsur, line 116: Let's please add a `// TODO: Remove Reflection when multitargetting is possible`

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
 
-        public static HttpMessageHandler CreateHttpClientHandler(int gatewayModeMaxConnectionLimit, IWebProxy webProxy, Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
+        public static HttpMessageHandler CreateHttpClientHandler(
+            int gatewayModeMaxConnectionLimit, 
+            IWebProxy webProxy, 
+            Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
+        {
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+
+            if (socketHandlerType != null)
```
ealsur, line 108: Let's please add a `// TODO: Remove type check and use #if NET6_0_OR_GREATERwhen multitargetting is possible`

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
ealsur, line 116: Make these `public` and add unit tests to make sure both methods create the correct types and sets the correct properties on https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -337,7 +338,7 @@ public async Task ResourceResponseStreamingTest()
                 catch (JsonReaderException ex)
                 {
                     Console.WriteLine("Expected exception while deserializing Resource: " + ex.Message);
-                }
+                }
```
Pilchie, line 341: nit: whitespace

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+                    remoteCertificateValidationCallbackInfo.SetValue(
+                        sslOptions,
+                        new RemoteCertificateValidationCallback((object _, X509Certificate certificate, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors) => serverCertificateCustomValidationCallback(
+                                certificate is { } ? new X509Certificate2(certificate) : null,
+                                x509Chain,
+                                sslPolicyErrors)),
+                        null);
+
+                    ////set SslOptions value
+                    //sslOptionsInfo.SetValue(socketHttpHandler,
```
Pilchie, line 158: Remove or uncomment.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+            if (webProxy != null)
+            {
+                PropertyInfo webProxyInfo = socketHandlerType.GetProperty("Proxy");
+                webProxyInfo.SetValue(socketHttpHandler, webProxy, null);
+            }
+
+            // https://docs.microsoft.com/en-us/archive/blogs/timomta/controlling-the-number-of-outgoing-connections-from-httpclient-net-core-or-full-framework
+            try
+            {
+                PropertyInfo maxConnectionsPerSercerInfo = socketHandlerType.GetProperty("MaxConnectionsPerServer");
```
Pilchie, line 138: spelling

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+            object socketHttpHandler = Activator.CreateInstance(socketHandlerType);
+
+            PropertyInfo pooledConnectionLifetimeInfo = socketHandlerType.GetProperty("PooledConnectionLifetime");
+            pooledConnectionLifetimeInfo.SetValue(socketHttpHandler, TimeSpan.FromMinutes(5), null);
+
+            // Proxy is only set by users and can cause not supported exception on some platforms
+            if (webProxy != null)
+            {
+                PropertyInfo webProxyInfo = socketHandlerType.GetProperty("Proxy");
+                webProxyInfo.SetValue(socketHttpHandler, webProxy, null);
```
Pilchie, line 132: You can use the two argument value, since you're not setting an indexer here.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+            try
+            {
+                PropertyInfo maxConnectionsPerSercerInfo = socketHandlerType.GetProperty("MaxConnectionsPerServer");
+                maxConnectionsPerSercerInfo.SetValue(socketHttpHandler, gatewayModeMaxConnectionLimit, null);
+
+                if (serverCertificateCustomValidationCallback != null)
+                {                     
+                    //Get SslOptions Property
+                    PropertyInfo sslOptionsInfo = socketHandlerType.GetProperty("SslOptions");
+                    object sslOptions = sslOptionsInfo.GetValue(socketHttpHandler, null);
```
Pilchie, line 145:
```suggestion
                    object sslOptions = sslOptionsInfo.GetValue(socketHttpHandler);
```

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+
+                    //Set SslOptions Property with custom certificate validation
+                    PropertyInfo remoteCertificateValidationCallbackInfo = sslOptions.GetType().GetProperty("RemoteCertificateValidationCallback");
+                    remoteCertificateValidationCallbackInfo.SetValue(
+                        sslOptions,
+                        new RemoteCertificateValidationCallback((object _, X509Certificate certificate, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors) => serverCertificateCustomValidationCallback(
+                                certificate is { } ? new X509Certificate2(certificate) : null,
+                                x509Chain,
+                                sslPolicyErrors)),
+                        null);
```
Pilchie, line 155:
```suggestion
                                sslPolicyErrors)));
```

Pilchie (approved):
One nit.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -227,6 +230,17 @@ public int GatewayModeMaxConnectionLimit
         /// <seealso cref="CosmosClientBuilder.WithRequestTimeout(TimeSpan)"/>
         public TimeSpan RequestTimeout { get; set; }
 
+        /// <summary>
+        /// Gets the pooled connection timeout in minutes when connected to the Azure Cosmos DB service.
```
ealsur, line 234:
```suggestion
        /// Gets or sets the HTTP pooled connection timeout in minutes when connected to the Azure Cosmos DB service.
```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -227,6 +230,17 @@ public int GatewayModeMaxConnectionLimit
...
+        /// This property defines maximal connection lifetime in the pool, tracking its age from when the connection was established, 
+        /// regardless of how much time it spent idle or active. Connections are not torn down while actively being used to service requests. 
+        /// This lifetime is useful in order to allow connections to be reestablished periodically so as to better reflect DNS or other network changes.
+        /// This value is only used in .NET version 6.0 and above.
+        /// </summary>
+        /// <value>Default value is 5 minutes.</value>
+        /// <seealso href="https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.pooledconnectionlifetime?view=net-8.0"/>       
+        public TimeSpan PooledConnectionLifetime { get; set; }
```
ealsur, line 242: We need to be explicit about which protocol it affect

```suggestion
        public TimeSpan HttpPooledConnectionLifetime { get; set; }
```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -227,6 +230,17 @@ public int GatewayModeMaxConnectionLimit
         /// <seealso cref="CosmosClientBuilder.WithRequestTimeout(TimeSpan)"/>
         public TimeSpan RequestTimeout { get; set; }
 
+        /// <summary>
+        /// Gets the pooled connection timeout in minutes when connected to the Azure Cosmos DB service.
+        /// This property defines maximal connection lifetime in the pool, tracking its age from when the connection was established, 
+        /// regardless of how much time it spent idle or active. Connections are not torn down while actively being used to service requests. 
+        /// This lifetime is useful in order to allow connections to be reestablished periodically so as to better reflect DNS or other network changes.
+        /// This value is only used in .NET version 6.0 and above.
```
ealsur, line 238: Add this part as a `<remarks>`

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -227,6 +230,17 @@ public int GatewayModeMaxConnectionLimit
...
+        /// </summary>
+        /// <value>Default value is 5 minutes.</value>
+        /// <seealso href="https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.pooledconnectionlifetime?view=net-8.0"/>
```
ealsur, line 241:
```suggestion
        /// <seealso href="https://learn.microsoft.com/dotnet/fundamentals/networking/http/httpclient-guidelines#dns-behavior"/>       
```

Avoid language specific urls

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -924,7 +924,8 @@ private async Task OpenPrivateAsync(CancellationToken cancellationToken)
                 this.ConnectionPolicy,
                 handler,
                 this.sendingRequest,
-                this.receivedResponse);
+                this.receivedResponse,
+                this.ConnectionPolicy.PooledConnectionLifetime);
```
ealsur, line 928: This method is already receiving the ConnectionPolicy 4 lines above

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -76,7 +76,8 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
                 httpMessageHandler = CosmosHttpClientCore.CreateHttpClientHandler(
                         gatewayModeMaxConnectionLimit: connectionPolicy.MaxConnectionLimit,
                         webProxy: null,
-                        serverCertificateCustomValidationCallback: connectionPolicy.ServerCertificateCustomValidationCallback);
+                        serverCertificateCustomValidationCallback: connectionPolicy.ServerCertificateCustomValidationCallback,
+                        pooledConnectionLifetime);
```
ealsur, line 80:
```suggestion
                        connectionPolicy.PooledConnectionLifetime);
```

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -45,7 +44,8 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
             ConnectionPolicy connectionPolicy,
             HttpMessageHandler httpMessageHandler,
             EventHandler<SendingRequestEventArgs> sendingRequestEventArgs,
-            EventHandler<ReceivedResponseEventArgs> receivedResponseEventArgs)
+            EventHandler<ReceivedResponseEventArgs> receivedResponseEventArgs,
+            TimeSpan pooledConnectionLifetime)
```
ealsur, line 48: remove

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -907,6 +907,44 @@ public async Task HttpClientConnectionLimitTest()
...
+                authKeyOrResourceToken: Convert.ToBase64String(Encoding.UTF8.GetBytes(Guid.NewGuid().ToString())));
+
+            //Assert type of message handler
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+            Type clientMessageHandlerType = cosmosClient.ClientContext.DocumentClient.httpClient.HttpMessageHandler.GetType();
+            Assert.AreEqual(socketHandlerType, clientMessageHandlerType);
+        }
+
+        [TestMethod]
+        public void CustomCertificateValidatorTest()
```
ealsur, line 925: Isn't this already tested with ` Verify_CertificateCallBackGetsCalled_ForTCP_HTTP`?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -16,6 +16,10 @@ namespace Microsoft.Azure.Cosmos.Tests
     using System.Collections.Generic;
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Cosmos.Tracing.TraceData;
+    using System.Net.Security;
+    using System.Security.Cryptography.X509Certificates;
+    using FluentAssertions;
```
ealsur, line 21: nit: is this using needed?
Also, reorder usings (you can do it with right-click Remove and reorder)

ealsur (approved):
LGTM, left some extra comments. Let's bring this PR to the next scrum to make sure we are all on the same page. Let's look at the public doc text and see if we are all in agreement. 

Another idea to discuss -> Should we have 2 defaults, one for Gateway mode and another for Direct mode?

You will also need to update public contract files

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
NaluTripician: You are correct, I missed that test, will remove this one.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+        {
+            // TODO: Remove Reflection when multitargetting is possible
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+
+            object socketHttpHandler = Activator.CreateInstance(socketHandlerType);
+
+            PropertyInfo pooledConnectionLifetimeInfo = socketHandlerType.GetProperty("PooledConnectionLifetime");
+            
+            //Sets the timeout for unused connections to a random time between 5 minutes and 5 minutes and 30 seconds.
+            //This is to avoid the issue where a large number of connections are closed at the same time.
```
ealsur, line 130:
```suggestion
            //This is to avoid the issue where a large number of connections are closed at the same time on a cluster of multiple machines
```

ealsur (dismissed):

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,80 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+            // TODO: Remove Reflection when multitargetting is possible
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+
+            object socketHttpHandler = Activator.CreateInstance(socketHandlerType);
+
+            PropertyInfo pooledConnectionLifetimeInfo = socketHandlerType.GetProperty("PooledConnectionLifetime");
+            
+            //Sets the timeout for unused connections to a random time between 5 minutes and 5 minutes and 30 seconds.
+            //This is to avoid the issue where a large number of connections are closed at the same time on a cluster of multiple machines
+            TimeSpan connectionTimeSpan = TimeSpan.FromMinutes(5) + TimeSpan.FromSeconds(30 * new Random().NextDouble());
```
Pilchie, line 131: You probably want `Random.Shared.NextDouble()` here.  See https://learn.microsoft.com/dotnet/api/system.random#avoiding-multiple-instantiations.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +98,86 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+            return CosmosHttpClientCore.CreateHttpClientHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+        }
+
+        public static HttpMessageHandler CreateSocketsHttpHandlerHelper(
+            int gatewayModeMaxConnectionLimit, 
+            IWebProxy webProxy, 
+            Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
+        {
+            // TODO: Remove Reflection when multitargetting is possible
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
```
sourabh1007, line 123: Should we put this in try-catch and failover to default implementation if anything fails in its execution? we never know what will change in future .net frameworks.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosMultiHashTest.cs
```diff
@@ -18,9 +18,13 @@ public class CosmosMultiHashTest
         private Container container = null;
         private ContainerProperties containerProperties = null;
 
+        private readonly string currentVersion = HttpConstants.Versions.CurrentVersion;
+
+
         [TestInitialize]
         public async Task TestInitialize()
         {
+            HttpConstants.Versions.CurrentVersion = "2020-07-15";
```
sourabh1007, line 27: just curious why are making this change?

NaluTripician: This was changed in #3763, unsure why it is undoing the changes here. Will remove.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -106,12 +107,18 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+                if (socketHandlerType != null)
+                {
+                    return CosmosHttpClientCore.CreateSocketsHttpHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+                }
             }
-            
-            return CosmosHttpClientCore.CreateHttpClientHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+            catch (Exception e) 
+            {
+                DefaultTrace.TraceError("Failed to create SocketsHttpHandler: {0}", e.ToString());
```
ealsur, line 119:
```suggestion
                DefaultTrace.TraceError("Failed to create SocketsHttpHandler: {0}", e);
```

Don't materialize strings before hand

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +99,92 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+        {
+            // TODO: Remove type check and use #if NET6_0_OR_GREATER when multitargetting is possible
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+
+            try
+            {
+                if (socketHandlerType != null)
+                {
+                    return CosmosHttpClientCore.CreateSocketsHttpHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+                }
```
ealsur, line 115: You are now missing the `else` case. Something like:

```
if (socketHandlerType != null)
{
	try
	{	
		return CosmosHttpClientCore.CreateSocketsHttpHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
	}
	catch (Exception e) 
	{
		DefaultTrace.TraceError("Failed to create SocketsHttpHandler: {0}", e);
	}  
}

return CosmosHttpClientCore.CreateHttpClientHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
```

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +99,90 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
         }
 
-        public static HttpMessageHandler CreateHttpClientHandler(int gatewayModeMaxConnectionLimit, IWebProxy webProxy, Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
+        public static HttpMessageHandler CreateHttpClientHandler(
+            int gatewayModeMaxConnectionLimit, 
+            IWebProxy webProxy, 
+            Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
+        {
+            // TODO: Remove type check and use #if NET6_0_OR_GREATER when multitargetting is possible
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
```
ealsur, line 108: You removed the `if` for this `socketHandlerType` check on your last commit

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
ealsur: You missed the `if`

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +99,93 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+
+            if (socketHandlerType != null)
+            {
+                try
+                {               
+                    return CosmosHttpClientCore.CreateSocketsHttpHandlerHelper(gatewayModeMaxConnectionLimit, webProxy, serverCertificateCustomValidationCallback);
+                }
+                catch (Exception e)
+                {
+                    DefaultTrace.TraceError("Failed to create SocketsHttpHandler: {0}", e);
```
ealsur, line 118: @kirankumarkolli / @Pilchie are we ok with falling back to using HttpClientHandler if using SocketsHttpHandler fails, or do you think failing completely would be better? Because we are using Reflection, my only concern is some platform not supporting some of the reflection types and crashing without recovery.

Pilchie: I support falling back to `HttpClientHandler`.

ealsur (approved):

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +99,93 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
                 eventSource: eventSource);
         }
 
-        public static HttpMessageHandler CreateHttpClientHandler(int gatewayModeMaxConnectionLimit, IWebProxy webProxy, Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
+        public static HttpMessageHandler CreateHttpClientHandler(
+            int gatewayModeMaxConnectionLimit, 
+            IWebProxy webProxy, 
+            Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
+        {
+            // TODO: Remove type check and use #if NET6_0_OR_GREATER when multitargetting is possible
```
kirankumarkolli, line 107: TODO's: Are they for future planned work?

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +99,93 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+            // TODO: Remove Reflection when multitargetting is possible
+            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+
+            object socketHttpHandler = Activator.CreateInstance(socketHandlerType);
+
+            PropertyInfo pooledConnectionLifetimeInfo = socketHandlerType.GetProperty("PooledConnectionLifetime");
+
+            //Sets the timeout for unused connections to a random time between 5 minutes and 5 minutes and 30 seconds.
+            //This is to avoid the issue where a large number of connections are closed at the same time.
+            Type threadSafeRandomType = Type.GetType("System.Random+ThreadSafeRandom, System.Private.CoreLib");
```
kirankumarkolli, line 139: ThreadSafeRandom reflection the only way to get the random?

There are other alternative naive types in .NET standard for random right?

NaluTripician: The `ThreadSafeRandom` reflection is the only way to get the thread-safe random that @Pilchie suggested.  This was added in .NET 6.0 and does not work in earlier versions so we have to use reflection if we want the thread-safe random fuctions. See: https://learn.microsoft.com/dotnet/api/system.random#avoiding-multiple-instantiations.

kirankumarkolli: One handler per client lifetime. ThreadSafeRandom using reflection vs type. My preference is to alter. 
@Pilchie thoughts?

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -99,7 +99,93 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
...
+                    remoteCertificateValidationCallbackInfo.SetValue(
+                        sslOptions,
+                        new RemoteCertificateValidationCallback((object _, X509Certificate certificate, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors) => serverCertificateCustomValidationCallback(
+                                certificate is { } ? new X509Certificate2(certificate) : null,
+                                x509Chain,
+                                sslPolicyErrors)));                   
+                }
+            }
+            // MaxConnectionsPerServer is not supported on some platforms.
+            catch (PlatformNotSupportedException)
```
Pilchie, line 178: Should this catch be before the `if` to set the `serverCertificateCustomValidationCallback`?

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
Pilchie: Alter to what?  I think options are:

1. `new Random().NextDouble` - issue is that if that happens on multiple threads close enough together, because the seed is chosen based on the clock, they will all end up with the same value.
2. Make a static field to hold the random.  Works, but calls to `NextDouble` aren't thread safe, so we'd need to introduce locks.
3. Use relection as the current change does.
4. Give up on randomly spreading the time, and just use a fixed time.

@kirankumarkolli is the concern we're trying to mitigate with the random significant enough to warrant 2 or 3?

kirankumarkolli: A variation of #2 with serializing in our code (anyways once generation) right?

kirankumarkolli: SocketsHttpHandler is suppsoed from core 3.0 onwards, where as ThreadSafeRandom is from >net 6.0. 

If we are basing .NET 6 as baseline then its okey.

ealsur: All net core versions are deprecated, net6 is the only LTS atm

Pilchie: True, but lots of customers still on .NET Core 3.1.  This change in its current form would break for them, right?  Because we would succeed in finding SocketsHttpHandler, but not ThreadSafeRandom?

.NET Team typically treats removing support for a TFM as a breaking change.

@kirankumarkolli can you expand on what you mean (maybe some sample code)?

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
NaluTripician: @Pilchie The catch coming after the if is based on the original behavior (see the `CreateHttpClientHandlerHelper` method on Lines 199-210). If you think this is a mistake then I can change it on both methods.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
ealsur: @Pilchie @kirankumarkolli how about we use the same Random we use for address randomization? It has a salt

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/94e811f848bbed1ededd422bd1efce9305721632/Microsoft.Azure.Cosmos/src/direct/CustomTypeExtensions.cs#L176

kirankumarkolli: > lock() { random.NextDouble() }
This can cover the threadsafe part. And its called once per client so full serial execution is acceptable.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
ealsur: This might be a bug introduced in https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3636. Setting the `serverCertificateCustomValidationCallback` should be outside (after the catch) in both methods. Good catch @Pilchie

ealsur (approved):

Pilchie (approved):

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
Pilchie: > Wouldn't the salt generated by the each unique execution help in that area? While not 100%, wouldn't it be good enough? The worst scenario is multiple machines in a cluster coming up at the same time (which is already rare) but wouldn't each machine's crypto layer generate a different salt?

It's been a while since I looked at the implementation, but I believe the seed to most non-crypto RNGs is the time.  So, across machines, it's probably fine because of clock sku, but on a single machine multiple threads, or multiple instances created within the same tick can all produce the same sequence of numbers.

PR #3783, opened 2023-03-29
[Internal] Client Telemetry: Refactors code to run client telemetry data processing task in background.

### LINKED ISSUE #3539: Client Telemetry: Rerun job if it failed and stopped after 3 retry
If Client telemetry job failed and stopped after 3 retries. then restart this job after half an hour again. because we don't want to be in a situation where job stopped due to network failure but it never started.

sourabh1007:
## Description
1. Right now, if processing takes 1 minute, then client telemetry window size will increase from 10 min to 10 + 1 = 11 min. Hence running it as a separate background job triggered by client telemetry job.
2. Cancelling the processing job if it is taking more than 5 min to give a breathing space to next round of telemetry.
3. Remove logic to stop client telemetry job.  Client Telemetry Job should start/stop by the API.

## Type of change

Closes #3539

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
@@ -173,16 +173,18 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                     try
                     {
                         CancellationTokenSource cancellationToken = new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryProcessorTimeOut);
+                        Task processorTask = Task.Run(() => this.processor
+                                                                    .ProcessAndSendAsync(
+                                                                            clientTelemetryInfo: this.clientTelemetryInfo,
+                                                                            operationInfoSnapshot: operationInfoSnapshot,
+                                                                            cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
+                                                                            requestInfoSnapshot: requestInfoSnapshot,
+                                                                            processorCancelToken: cancellationToken), cancellationToken.Token);
 
                         // Initiating Telemetry Data Processor task which will serialize and send telemetry information to Client Telemetry Service
                         // Not disposing this task. If we dispose a client then, telemetry job(telemetryTask) should stop but processor task(processorTask) should make best effort to finish the job in background.
-                        _ = Task.Run(() => ClientTelemetry
-                                           .RunProcessorTaskAsync(this.processor.ProcessAndSendAsync(
-                                                                      clientTelemetryInfo: this.clientTelemetryInfo,
-                                                                      operationInfoSnapshot: operationInfoSnapshot,
-                                                                      cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
-                                                                      requestInfoSnapshot: requestInfoSnapshot,
-                                                                      cancellationToken: cancellationToken)), cancellationToken.Token);
+                        _ = ClientTelemetry.RunProcessorTaskAsync(this.clientTelemetryInfo.DateTimeUtc, processorTask, ClientTelemetryOptions.ClientTelemetryProcessorTimeOut);
+
                     }
                     catch (Exception ex)
                     {
@@ -198,15 +200,30 @@ namespace Microsoft.Azure.Cosmos.Telemetry
             DefaultTrace.TraceInformation("Telemetry Job Stopped.");
         }
 
-        internal static async Task RunProcessorTaskAsync(Task processingTask)
+        /// <summary>
+        /// This Task makes sure, processing task is timing out after 5 minute of timeout
+        /// </summary>
+        /// <param name="telemetryDate"></param>
+        /// <param name="processingTask"></param>
+        /// <param name="timeout"></param>
+        internal static async Task RunProcessorTaskAsync(string telemetryDate, Task processingTask, TimeSpan timeout)
         {
-            try
+            using (CancellationTokenSource tokenForDelayTask = new CancellationTokenSource())
             {
-               await processingTask;
-            }
-            finally
-            {
-                processingTask.Dispose(); // release all resources
+                Task delayTask = Task.Delay(timeout, tokenForDelayTask.Token);
+                
+                Task resultTask = await Task.WhenAny(processingTask, delayTask);
+                if (resultTask == delayTask)
+                {
+                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is cancelled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
+                    // Operation cancelled
+                    throw new TimeoutException($"Processor task with date as {telemetryDate} is cancelled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
+                }
+                else
+                {
+                    // Cancel the timer task so that it does not fire
+                    tokenForDelayTask.Cancel();
+                }
             }
         }
 
--- a/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
@@ -36,7 +36,6 @@ namespace Microsoft.Azure.Cosmos.Telemetry
         private readonly NetworkDataRecorder networkDataRecorder;
         
         private readonly CancellationTokenSource cancellationTokenSource;
-        
         private readonly GlobalEndpointManager globalEndpointManager;
 
         private Task telemetryTask;
--- a/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
@@ -148,7 +148,7 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                         AccountProperties accountProperties = await ClientTelemetryHelper.SetAccountNameAsync(this.globalEndpointManager);
                         this.clientTelemetryInfo.GlobalDatabaseAccountName = accountProperties.Id;
                     }
-                    
+
                     await Task.Delay(observingWindow, this.cancellationTokenSource.Token);
 
                     this.clientTelemetryInfo.DateTimeUtc = DateTime.UtcNow.ToString(ClientTelemetryOptions.DateFormat);
@@ -170,24 +170,24 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                         = Interlocked.Exchange(ref this.cacheRefreshInfoMap, new ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram>());
 
                     List<RequestInfo> requestInfoSnapshot = this.networkDataRecorder.GetRequests();
-                    
+
                     try
                     {
                         CancellationTokenSource cancellationToken = new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryProcessorTimeOut);
-                        Task processorTask = Task.Run(() => this.processor
-                                                                    .ProcessAndSendAsync(
-                                                                            clientTelemetryInfo: this.clientTelemetryInfo,
-                                                                            operationInfoSnapshot: operationInfoSnapshot,
-                                                                            cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
-                                                                            requestInfoSnapshot: requestInfoSnapshot), cancellationToken.Token);
 
                         // Initiating Telemetry Data Processor task which will serialize and send telemetry information to Client Telemetry Service
                         // Not disposing this task. If we dispose a client then, telemetry job(telemetryTask) should stop but processor task(processorTask) should make best effort to finish the job in background.
-                        _ = ClientTelemetry.RunProcessorTaskAsync(this.clientTelemetryInfo.DateTimeUtc, processorTask);
+                        _ = Task.Run(() => ClientTelemetry
+                                           .RunProcessorTaskAsync(this.processor.ProcessAndSendAsync(
+                                                                      clientTelemetryInfo: this.clientTelemetryInfo,
+                                                                      operationInfoSnapshot: operationInfoSnapshot,
+                                                                      cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
+                                                                      requestInfoSnapshot: requestInfoSnapshot,
+                                                                      cancellationToken: cancellationToken)), cancellationToken.Token);
                     }
                     catch (Exception ex)
                     {
-                        DefaultTrace.TraceError("Telemetry Job Processor failed to run with error : {0}", ex);
+                        DefaultTrace.TraceError("Exception while processing the data : {0} with telemetry date as {1}", ex.Message, this.clientTelemetryInfo.DateTimeUtc);
                     }
                 }
             }
@@ -199,18 +199,15 @@ namespace Microsoft.Azure.Cosmos.Telemetry
             DefaultTrace.TraceInformation("Telemetry Job Stopped.");
         }
 
-        internal static async Task RunProcessorTaskAsync(string telemetryDate, Task processingTask)
+        internal static async Task RunProcessorTaskAsync(Task processingTask)
         {
-            Task[] tasks = new Task[] { Task.Delay(ClientTelemetryOptions.ClientTelemetryProcessorTimeOut), processingTask };
-            Task completedTask = await Task.WhenAny(tasks);
-
-            if (!object.ReferenceEquals(completedTask, processingTask))
+            try
             {
-                DefaultTrace.TraceError($"Telemetry Processor Task timed out after {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut.TotalMilliseconds}ms with telemetry date as {telemetryDate}");
-
-                processingTask.Dispose();
-                
-                throw new TimeoutException($"Telemetry Processor Task timed out after {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut.TotalMilliseconds}ms with telemetry date as {telemetryDate}");
+               await processingTask;
+            }
+            finally
+            {
+                processingTask.Dispose(); // release all resources
             }
         }
 
--- a/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
@@ -179,7 +179,7 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                                                                             operationInfoSnapshot: operationInfoSnapshot,
                                                                             cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
                                                                             requestInfoSnapshot: requestInfoSnapshot,
-                                                                            processorCancelToken: cancellationToken), cancellationToken.Token);
+                                                                            cancellationToken: cancellationToken.Token), cancellationToken.Token);
 
                         // Initiating Telemetry Data Processor task which will serialize and send telemetry information to Client Telemetry Service
                         // Not disposing this task. If we dispose a client then, telemetry job(telemetryTask) should stop but processor task(processorTask) should make best effort to finish the job in background.
@@ -188,7 +188,7 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                     }
                     catch (Exception ex)
                     {
-                        DefaultTrace.TraceError("Exception while processing the data : {0} with telemetry date as {1}", ex.Message, this.clientTelemetryInfo.DateTimeUtc);
+                        DefaultTrace.TraceError("Exception while initiating processing task : {0} with telemetry date as {1}", ex.Message, this.clientTelemetryInfo.DateTimeUtc);
                     }
                 }
             }
@@ -215,9 +215,9 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                 Task resultTask = await Task.WhenAny(processingTask, delayTask);
                 if (resultTask == delayTask)
                 {
-                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is cancelled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
+                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is canceled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut}");
                     // Operation cancelled
-                    throw new TimeoutException($"Processor task with date as {telemetryDate} is cancelled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
+                    throw new OperationCanceledException($"Processor task with date as {telemetryDate} is canceled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut}");
                 }
                 else
                 {
--- a/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
@@ -174,13 +174,15 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                     
                     try
                     {
+                        CancellationTokenSource cancellationToken = new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryProcessorTimeOut);
+                        
                         // Initiating Telemetry Data Processor task which will serialize and send telemetry information to Client Telemetry Service service
                         this.processorTask = Task.Run(() => ClientTelemetry.RunProcessorTaskAsync(this.processor
                                                                 .ProcessAndSendAsync(
                                                                     clientTelemetryInfo: this.clientTelemetryInfo,
                                                                     operationInfoSnapshot: operationInfoSnapshot,
                                                                     cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
-                                                                    requestInfoSnapshot: requestInfoSnapshot)), new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryProcessorTimeOut).Token);
+                                                                    requestInfoSnapshot: requestInfoSnapshot)), cancellationToken.Token);
                     }
                     catch (Exception ex)
                     {
@@ -200,7 +202,12 @@ namespace Microsoft.Azure.Cosmos.Telemetry
         {
             try
             {
-                await processingTask;
+                Task completedTask = await Task.WhenAny(processingTask, Task.Delay(ClientTelemetryOptions.ClientTelemetryProcessorTimeOut));
+                
+                if (completedTask != processingTask)
+                {
+                    throw new TimeoutException($"Telemetry Processor Task timed out after {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut.TotalMilliseconds}ms");
+                }
             }
             catch (Exception ex)
             {
--- a/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
@@ -215,9 +215,9 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                 Task resultTask = await Task.WhenAny(processingTask, delayTask);
                 if (resultTask == delayTask)
                 {
-                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is canceled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut}");
+                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is canceled as it did not finish in {timeout}");
                     // Operation cancelled
-                    throw new OperationCanceledException($"Processor task with date as {telemetryDate} is canceled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut}");
+                    throw new OperationCanceledException($"Processor task with date as {telemetryDate} is canceled as it did not finish in {timeout}");
                 }
                 else
                 {
--- a/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
@@ -28,8 +28,6 @@ namespace Microsoft.Azure.Cosmos.Telemetry
     /// </summary>
     internal class ClientTelemetry : IDisposable
     {
-        private const int allowedNumberOfFailures = 3;
-
         private static readonly TimeSpan observingWindow = ClientTelemetryOptions.GetScheduledTimeSpan();
 
         private readonly ClientTelemetryProperties clientTelemetryInfo;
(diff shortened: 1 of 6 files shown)


Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -180,23 +174,32 @@ private async Task EnrichAndSendAsync()
...
+                        Task tempProcessorTask = Task.Run(() => this.processor
                             .ProcessAndSendAsync(
                                 clientTelemetryInfo: this.clientTelemetryInfo,
                                 operationInfoSnapshot: operationInfoSnapshot,
                                 cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
-                                requestInfoSnapshot: this.networkDataRecorder.GetRequests(),
-                                cancellationToken: this.cancellationTokenSource.Token);
+                                requestInfoSnapshot: requestInfoSnapshot,
+                                cancellationToken: this.cancellationTokenSource.Token));
+                        tempProcessorTask.Start(); // run it in background
```
ealsur, line 193: Doesn't `Run` already start the task?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -180,23 +174,32 @@ private async Task EnrichAndSendAsync()
...
                         = Interlocked.Exchange(ref this.cacheRefreshInfoMap, new ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram>());
+
+                    List<RequestInfo> requestInfoSnapshot = this.networkDataRecorder.GetRequests();
                     
                     try
                     {
-                        await this.processor
+                        // Use the Wait method with a CancellationToken to wait for the task to complete.
+                        // If anyhow prev task was not finished sucessfully in 5 min then ,Do not trigger a new task. 
+                        this.processorTask?.Wait(this.processorCancellationTokenSource.Token);
```
ealsur, line 184: This is a blocking call, why not `await` instead?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -180,23 +174,32 @@ private async Task EnrichAndSendAsync()
...
-                                requestInfoSnapshot: this.networkDataRecorder.GetRequests(),
-                                cancellationToken: this.cancellationTokenSource.Token);
+                                requestInfoSnapshot: requestInfoSnapshot,
+                                cancellationToken: this.cancellationTokenSource.Token));
+                        tempProcessorTask.Start(); // run it in background
 
-                        this.numberOfFailures = 0;
+                        this.processorTask = tempProcessorTask;
+                    }
+                    catch (OperationCanceledException)
```
ealsur, line 197: Why not use the pattern with Task.WhenAny and a Task.Delay along with the processorTask? If the Task.Delay finishes sooner, then it means the processorTask timed out and should be canceled

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryOptions.cs
```diff
@@ -79,7 +79,7 @@ internal static class ClientTelemetryOptions
         internal const double Percentile99 = 99.0;
         internal const double Percentile999 = 99.9;
         internal const string DateFormat = "yyyy-MM-ddTHH:mm:ssZ";
-        
+        internal const int ProcessorTimeOutInMs = 5 * 60 * 1000; // 5 minutes
```
ealsur, line 82: Why not use TimeSpan.FromMinutes(5)?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -149,12 +149,6 @@ private async Task EnrichAndSendAsync()
             {
                 while (!this.cancellationTokenSource.IsCancellationRequested)
                 {
-                    if (this.numberOfFailures == allowedNumberOfFailures)
```
ealsur, line 152: Who is using `allowedNumberOfFailures`? Should it be removed?

ealsur, line 152: What is `this.numberOfFailures` used for now?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
ealsur: Also, the `this.processorCancellationTokenSource` has 5 minute lifespan, you are not resetting it, which means, after 5 min it is completely cancelled and it will forever stop the processing

sourabh1007: yes if particular batch is taking more than 5 min in processing.. just kill that task and give space to another upcoming task, I don't want to be in a situation when all the tasks are either piled up or eating resources in order to complete the processing task. I want to keep moving forward with whatever we have in our hand.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: I have removed this code.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -180,23 +174,32 @@ private async Task EnrichAndSendAsync()
...
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        DefaultTrace.TraceError("Telemetry Job Processor failed due to timeout in 5 min");
                     }
                     catch (Exception ex)
                     {
-                        this.numberOfFailures++;
-
                         DefaultTrace.TraceError("Telemetry Job Processor failed with error : {0}", ex);
```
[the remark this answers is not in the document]
sourabh1007: added continue with block for this task and inside processor, there are other try-catchs available.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: already removed

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -180,24 +169,32 @@ private async Task EnrichAndSendAsync()
...
-                        this.numberOfFailures = 0;
+                        // Initiating Telemetry Data Processor task which will serialize and send telemetry information to Client Telemetry Service service
+                        this.processorTask = Task.Run(
+                            function: () => this.processor
+                                                    .ProcessAndSendAsync(
+                                                        clientTelemetryInfo: this.clientTelemetryInfo,
+                                                        operationInfoSnapshot: operationInfoSnapshot,
+                                                        cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
+                                                        requestInfoSnapshot: requestInfoSnapshot),
+                            cancellationToken: this.processorCancellationTokenSource.Token)
```
ealsur, line 185: `processorCancellationTokenSource` is created, but it is never reset. It is being created as a readonly on the class scope, you cannot reuse the CancellationTokenSource, it is not how it works.

When you create a CancellationTokenSource, it can be used once, it will cancel 5 minutes after it is created. You need to re-create it in order to use it in another subsequent request/operation.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -180,24 +169,32 @@ private async Task EnrichAndSendAsync()
...
+                                                .ContinueWith((i) =>
```
ealsur, line 186: Avoid `ContinueWith`. You can do the same thing using an async method:

```
private static async Task ProcessAndObserve(Task processingTask)
{
    try
    {
        await processingTask;
    }
    catch (Exception ex)
    {
        DefaultTrace.TraceError($"Client Telemetry data processing task faulted. {0}", ex);
    }}
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -23,7 +23,8 @@ internal class ClientTelemetryProcessor
         
         private readonly AuthorizationTokenProvider tokenProvider;
         private readonly CosmosHttpClient httpClient;
-            
+        private readonly CancellationTokenSource serviceCancellationToken = new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryServiceTimeOut);
```
ealsur, line 26: Same problem here. You are creating a CancellationTokenSource and reusing it, this will only work once

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
ealsur, line 185: You are using `Task.Run` with CancellationToken parameter -> https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=net-8.0#system-threading-tasks-task-run(system-func((system-threading-tasks-task))-system-threading-cancellationtoken)

This does not work for a timeout enforcement, because `Task.Run` only check the CancellationToken state **before** the async work starts.

> A cancellation token that can be used to cancel the work if it has not yet started. [Run(Func<Task>, CancellationToken)](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=net-8.0#system-threading-tasks-task-run(system-func((system-threading-tasks-task))-system-threading-cancellationtoken)) does not pass cancellationToken to action.

ealsur (changes requested):
CancellationToken wiring on ProcessAndSendAsync is not enforcing timeouts and re-usage of CancellationTokenSource needs to be fixed

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: made changes

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -180,24 +168,26 @@ private async Task EnrichAndSendAsync()
...
-                        await this.processor
-                            .ProcessAndSendAsync(
-                                clientTelemetryInfo: this.clientTelemetryInfo,
-                                operationInfoSnapshot: operationInfoSnapshot,
-                                cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
-                                requestInfoSnapshot: this.networkDataRecorder.GetRequests(),
-                                cancellationToken: this.cancellationTokenSource.Token);
-
-                        this.numberOfFailures = 0;
+                        CancellationTokenSource cancellationToken = new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryProcessorTimeOut);
```
ealsur, line 176: The code in this method is not really enforcing the timeout, just passing the CT, if the enforcement is on the `ProcessAndSendAsync` why not create the CancellationTokenSource there?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -209,6 +199,18 @@ await this.processor
             DefaultTrace.TraceInformation("Telemetry Job Stopped.");
         }
 
+        internal static async Task RunProcessorTaskAsync(Task processingTask)
+        {
+            try
+            {
+               await processingTask;
+            }
+            finally
```
ealsur, line 208: Should the catch with the TraceError be moved here?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryOptions.cs
```diff
@@ -79,13 +79,14 @@ internal static class ClientTelemetryOptions
         internal const double Percentile99 = 99.0;
         internal const double Percentile999 = 99.9;
         internal const string DateFormat = "yyyy-MM-ddTHH:mm:ssZ";
-        
+
```
ealsur, line 82: nit: remove extra spaces

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -106,6 +118,11 @@ internal static class ClientTelemetryPayloadWriter
                 
                 foreach (RequestInfo entry in sampledRequestInfo)
                 {
+                    if (cancellationToken.IsCancellationRequested)
```
ealsur, line 121: Shouldn't the `callback.Invoke` also take the CancellationToken to cancel network calls?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -39,27 +40,38 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
             ConcurrentDictionary<OperationInfo, (LongConcurrentHistogram latency, LongConcurrentHistogram requestcharget)> operationInfoSnapshot,
             ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram> cacheRefreshInfoSnapshot,
             IReadOnlyList<RequestInfo> requestInfoSnapshot,
-            CancellationToken cancellationToken)
+            CancellationTokenSource cancellationToken)
         {
+            CancellationTokenSource serviceCancellationToken = new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryServiceTimeOut);
             try
             {
+                if (cancellationToken.IsCancellationRequested)
```
ealsur, line 48: This is only checking the 5 minute at the beginning, when do we expect 5 min to be elapsed if we check it only on the beginning? Should the 5 min better be enforced with a Task.Delay and WhenAny along with `SerializedPayloadChunksAsync`

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: Here requirement is to check if processor task is taking more time than the configured timeout. There are 2 ways to solve this problem:
1. Using WhenAny() with Delay Thread or similar function to identify if my task is taking more than configured delay task : With this approach I see 2 problems, 
a) I will be spinning 2 extra tasks to identify the time taken by processor. Here, I want to avoid spinning up more threads. 
b) once it identified that processor task is taking more time, we get the instance of delay task but won't be able to cancel the processor task actually.

2. Cancellation Token approach (current approach): It is not actual timeout approach but it will serve the purpose, here I have a cancellation token with configured time, and I am checking for cancellation token status at the points where I think more time can be consumed. Now if in one of those places if it finds that token has been cancelled. it will just come out of the task and will stop consuming more resources.

**why not create the CancellationTokenSource there?** To include the time taken in spinning up a new thread (imagine a scenario with thread contention and anyhow processor took time to start)

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: I intentionally moved it out of this. It is easier to write test along with this function otherwise, It will throw exception and just print the error. And I need to use mocking or something to do assertions.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: it is not just in the beginning, it is there in all the data loops, these data loops might take time, I don't see any other code where time can be consumed.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
ealsur: First, a Task is not a thread. Second, creating the CT here is not accounting the time to spin up the Task because you are not awaiting it. Third, because a Task is not a Thread, using solution one does not use more threads. And last, you can certainly cancel the processor task when the timeout happens, you can signal a CT inside the processor or dispose the Task directly.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
ealsur: But your catch on the other function will not catch anything, because you are creating the Task using Task.Run (fire and forget), so you are dropping any exception logging.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
ealsur: Nevermind, I just saw the different parameters

ealsur, line 48:
```suggestion
cancellationToken.ThrowIfCancellationRequested();
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryPayloadWriter.cs
```diff
@@ -37,6 +39,11 @@ internal static class ClientTelemetryPayloadWriter
             {
                 foreach (KeyValuePair<OperationInfo, (LongConcurrentHistogram latency, LongConcurrentHistogram requestcharge)> entry in operationInfoSnapshot)
                 {
+                    if (cancellationToken.IsCancellationRequested)
```
ealsur, line 42: Replace all the `throw new TimeoutException` with

```suggestion
                    cancellationToken.ThrowIfCancellationRequested();
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -39,27 +40,38 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
             ConcurrentDictionary<OperationInfo, (LongConcurrentHistogram latency, LongConcurrentHistogram requestcharget)> operationInfoSnapshot,
             ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram> cacheRefreshInfoSnapshot,
             IReadOnlyList<RequestInfo> requestInfoSnapshot,
-            CancellationToken cancellationToken)
+            CancellationTokenSource cancellationToken)
         {
+            CancellationTokenSource serviceCancellationToken = new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryServiceTimeOut);
```
ealsur, line 45: Should this be a Linked CancellationToken?

```suggestion
            CancellationTokenSource serviceCancellationToken = new CancellationTokenSource(ClientTelemetryOptions.ClientTelemetryServiceTimeOut);
            CancellationTokenSource serviceCancellationToken = new CancellationTokenSource.CreateLinkedTokenSource(serviceCancellationToken.Token, cancellationToken);
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
ealsur: Example of how we enforce timeout on a Task -> https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedProcessing/FeedProcessorCore.cs#L52-L65

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: Linking service token with processor token.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: removed cancellation

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -39,16 +40,18 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
             ConcurrentDictionary<OperationInfo, (LongConcurrentHistogram latency, LongConcurrentHistogram requestcharget)> operationInfoSnapshot,
             ConcurrentDictionary<CacheRefreshInfo, LongConcurrentHistogram> cacheRefreshInfoSnapshot,
             IReadOnlyList<RequestInfo> requestInfoSnapshot,
-            CancellationToken cancellationToken)
+            CancellationTokenSource processorCancelToken)
```
ealsur, line 43:
```suggestion
            CancellationToken processorCancellationToken)
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -39,16 +40,18 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
...
         {
             try
             {
+                using CancellationTokenSource cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(processorCancelToken.Token);
```
ealsur, line 47: This is not needed because you are not linking 2 tokens with different time

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -39,16 +40,18 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
...
+
                 await ClientTelemetryPayloadWriter.SerializedPayloadChunksAsync(
                     properties: clientTelemetryInfo,
                     operationInfoSnapshot: operationInfoSnapshot,
                     cacheRefreshInfoSnapshot: cacheRefreshInfoSnapshot,
                     sampledRequestInfo: requestInfoSnapshot,
-                    callback: async (payload) => await this.SendAsync(clientTelemetryInfo.GlobalDatabaseAccountName, payload, cancellationToken));
+                    callback: async (payload) => await this.SendAsync(clientTelemetryInfo.GlobalDatabaseAccountName, payload, cancellationToken.Token));
```
ealsur, line 54:
```suggestion
                    callback: async (payload) => await this.SendAsync(clientTelemetryInfo.GlobalDatabaseAccountName, payload, processorCancellationToken));
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -180,24 +167,28 @@ private async Task EnrichAndSendAsync()
...
+                        // Not disposing this task. If we dispose a client then, telemetry job(telemetryTask) should stop but processor task(processorTask) should make best effort to finish the job in background.
+                        _ = ClientTelemetry.RunProcessorTaskAsync(this.clientTelemetryInfo.DateTimeUtc, processorTask, ClientTelemetryOptions.ClientTelemetryProcessorTimeOut);
+
                     }
                     catch (Exception ex)
                     {
-                        this.numberOfFailures++;
-
-                        DefaultTrace.TraceError("Telemetry Job Processor failed with error : {0}", ex);
+                        DefaultTrace.TraceError("Exception while processing the data : {0} with telemetry date as {1}", ex.Message, this.clientTelemetryInfo.DateTimeUtc);
```
ealsur, line 191: This catch block is not catching any processing error.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -209,6 +200,33 @@ await this.processor
...
+            using (CancellationTokenSource tokenForDelayTask = new CancellationTokenSource())
+            {
+                Task delayTask = Task.Delay(timeout, tokenForDelayTask.Token);
+                
+                Task resultTask = await Task.WhenAny(processingTask, delayTask);
+                if (resultTask == delayTask)
+                {
+                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is cancelled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
+                    // Operation cancelled
+                    throw new TimeoutException($"Processor task with date as {telemetryDate} is cancelled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
```
ealsur, line 220:
```suggestion
                    throw new OperationCanceledException($"Processor task with date as {telemetryDate} is cancelled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -209,6 +200,33 @@ await this.processor
...
+        internal static async Task RunProcessorTaskAsync(string telemetryDate, Task processingTask, TimeSpan timeout)
+        {
+            using (CancellationTokenSource tokenForDelayTask = new CancellationTokenSource())
+            {
+                Task delayTask = Task.Delay(timeout, tokenForDelayTask.Token);
+                
+                Task resultTask = await Task.WhenAny(processingTask, delayTask);
+                if (resultTask == delayTask)
+                {
+                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is cancelled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
```
ealsur, line 218:
```suggestion
                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is canceled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut} milliseconds.");
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryOptions.cs
```diff
@@ -101,7 +101,8 @@ internal static class ClientTelemetryOptions
 
         internal static readonly int NetworkTelemetrySampleSize = 200;
         internal static int PayloadSizeThreshold = 1024 * 1024 * 2; // 2MB
-
+        internal static TimeSpan ClientTelemetryProcessorTimeOut = TimeSpan.FromMinutes(5); // 5 minutes
```
ealsur, line 104:
```suggestion
        internal static TimeSpan ClientTelemetryProcessorTimeOut = TimeSpan.FromMinutes(5);
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -209,6 +200,33 @@ await this.processor
...
+        internal static async Task RunProcessorTaskAsync(string telemetryDate, Task processingTask, TimeSpan timeout)
+        {
+            using (CancellationTokenSource tokenForDelayTask = new CancellationTokenSource())
+            {
+                Task delayTask = Task.Delay(timeout, tokenForDelayTask.Token);
+                
+                Task resultTask = await Task.WhenAny(processingTask, delayTask);
+                if (resultTask == delayTask)
+                {
+                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is canceled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut}");
```
ealsur, line 218: just realized the method takes the timespan as parameter. If we want to use `ClientTelemetryOptions.ClientTelemetryProcessorTimeOut` on the error trace then we need to use it on the Task.Delay and remove it from the parameters, or, change the message to:

```suggestion
                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is canceled as it did not finish in {timeout}");
```

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -209,6 +200,33 @@ await this.processor
...
+                    // Operation cancelled
+                    throw new OperationCanceledException($"Processor task with date as {telemetryDate} is canceled as it did not finish in {ClientTelemetryOptions.ClientTelemetryProcessorTimeOut}");
```
ealsur, line 220: Same as the other comment:

```suggestion
                    throw new OperationCanceledException($"Processor task with date as {telemetryDate} is canceled as it did not finish in {timeout}");
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/ClientTelemetryTests.cs
```diff
@@ -245,14 +244,116 @@ public async Task CheckIfPayloadIsDividedCorrectlyAsync(int expectedOperationInf
...
             Assert.AreEqual(expectedOperationInfoSize, actualOperationInfoSize, "Operation Info is not correct");
             Assert.AreEqual(expectedCacheRefreshInfoSize, actualCacheRefreshInfoSize, "Cache Refresh Info is not correct");
             Assert.AreEqual(expectedRequestInfoSize, actualRequestInfoSize, "Request Info is not correct");
         }
-        
+
+        [TestMethod]
+        public async Task ClientTelmetryProcessor_should_timeout()
+        {
+            Environment.SetEnvironmentVariable(ClientTelemetryOptions.EnvPropsClientTelemetryEndpoint, "http://dummy.telemetry.endpoint/");
```
ealsur, line 258: Shouldn't we reset this value after the test is done?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/ClientTelemetryTests.cs
```diff
@@ -245,14 +244,116 @@ public async Task CheckIfPayloadIsDividedCorrectlyAsync(int expectedOperationInf
...
+            try
+            {
+                Task processorTask = Task.Run(() => processor.ProcessAndSendAsync(
+                                                     clientTelemetryProperties,
+                                                     operationInfoSnapshot,
+                                                     cacheRefreshInfoSnapshot,
+                                                     default,
+                                                     new CancellationTokenSource(TimeSpan.FromSeconds(1)).Token));
+
+                await ClientTelemetry.RunProcessorTaskAsync(DateTime.Now.ToString(), processorTask, TimeSpan.FromTicks(1));
```
ealsur, line 347:
```suggestion
                await Assert.ThrowsAsync<OperationCanceledException>(() -> ClientTelemetry.RunProcessorTaskAsync(DateTime.Now.ToString(), processorTask, TimeSpan.FromTicks(1)));
```

This removes the try/catch and all the other Asserts

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/ClientTelemetryTests.cs
sourabh1007: doesn't matter...i need to set this value to pass some validation in client teemetry code. and If I need it in other tests, I always setting it some value.

ealsur: I understand, but from the correctness perspective, this test is setting an Environment variable, which affects all tests that run after it, and you cannot tell which are them, or if any of them, now or in the future, might be affected by this global variable. Shouldn't we reset the variable to the previous state when the test is done? You can read the current value, run the test and on a `finally` clause, set the value to what it was originally. That way the test cleans up after itself and has no posibility of impacting a future test.

sourabh1007: cleaning up this in test clean up.

ealsur (dismissed):

ealsur (approved):

Issue #3459, opened 2022-09-14
[Internal] Designing capabilities to support Distributed Tracing


philipthomas-MSFT:
**Is your feature request related to a problem? Please describe.**
To be able to implement and deliver distributed tracing transactions downstream of the Cosmos DB SDK into the compute gateway, routing gateway, and/or the backend services.

**Describe the solution you'd like**
While conforming to the  OpenTelemetry Guidelines, Standards and Best Practices, we will build the following capabilities:

- The ability to generate a new _TraceId_, as well as additional intrinsic and extrinsic context information, that represents a new trace transaction from the SDK client automatically, when one is not provided by the user.

- The ability to create accept a _TraceId_, as well as additional intrinsic and extrinsic contextual information, that represents a new or existing trace transaction from the SDK client manually, when one is provided by the user.

- The ability to transport a generated or manually created _TraceId_, as well as additional intrinsic and extrinsic contextual information, from the SDK into other components that live within that timeline downstream of the SDK.

- The ability to generate _SpanId_, as well as additional intrinsic and extrinsic contextual information, that represents a new component, or operation within that trace's transaction.

**Describe alternatives you've considered**
TBD

**Additional context**
TBD

philipthomas-MSFT:
Issues with Compute and Routing Gateway build failure.
PR is out, need more reviews.

PR #3801, opened 2023-04-05
[Internal] OpenTelemetry : Adds Telemetry Distributed Tracing functionality

aavasthy:
# Pull Request Template

## Description
[Feature 3459:](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3459) Adds open telemetry support to the SDK for network level requests for both HTTP and TCP requests.
Second PR for Distributed Tracing changes:

- Update OpentelmetryRecorder and factory for including logic to have a parent Id at operation level
- Add distributed trace id in Diagnostics 

Performance Test Results
| | **Master** | **Distributed Code Change** | **Distributed Code Change with exporter in Benchmark** |
| -------| -------- | ---------- | --------|
| EnableDistributedTracingFlag | False | True | True |
| 1. | 8891 rps| 9045 rps| 9356 rps|
| 2. | 9919 rps| 9377 rps| 9981 rps|
| 3. | 9566 rps| 9398 rps| 9508 rps|
| 4. | 9180 rps| 11308 rps| 9625 rps|
| 5. | 8798 rps| 9428 rps| 9522 rps|
| 6. | 8879 rps| 9611 rps| 9315 rps|
| 7. | 10634 rps | 9416 rps| 9118 rps|
| 8. | 8952 rps| 9503 rps| 9372 rps|
| 9. | 8958 rps| 9842 rps| 9516 rps|
| 10. | 8887 rps| 9314 rps| 9334 rps|

## Type of change


Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -30,6 +32,16 @@ internal struct OpenTelemetryCoreRecorder : IDisposable
...
         };
 
+        public OpenTelemetryCoreRecorder(DiagnosticScope scope)
+        {
+            this.scope = scope;
+            this.scope.Start();
+        }
+        public OpenTelemetryCoreRecorder(string operationName)
+        {
+            this.activity = new Activity(operationName);
```
ealsur, line 42: What is the default value of scope? And is it Disposable?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -30,6 +32,16 @@ internal struct OpenTelemetryCoreRecorder : IDisposable
             { typeof(ChangeFeedProcessorUserException), (exception, scope) => ChangeFeedProcessorUserException.RecordOtelAttributes((ChangeFeedProcessorUserException)exception, scope)}
         };
 
+        public OpenTelemetryCoreRecorder(DiagnosticScope scope)
+        {
+            this.scope = scope;
+            this.scope.Start();
+        }
+        public OpenTelemetryCoreRecorder(string operationName)
```
ealsur, line 40: nit: separate functions with empty lines

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -30,6 +32,16 @@ internal struct OpenTelemetryCoreRecorder : IDisposable
             { typeof(ChangeFeedProcessorUserException), (exception, scope) => ChangeFeedProcessorUserException.RecordOtelAttributes((ChangeFeedProcessorUserException)exception, scope)}
         };
 
+        public OpenTelemetryCoreRecorder(DiagnosticScope scope)
```
ealsur, line 35: This class has now 3 constructors, which is used when? And would it be better to have a private constructor and static methods with different names? For example: CreateWithDiagnosticScope() or CreateWithActivity(). Particularly the second one (activity) takes a string, it is rather obscure that it ends up creating an Activity instead of a scope

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
```diff
@@ -49,7 +49,6 @@ public ClientSideRequestStatisticsTraceDatum(DateTime startTime, TraceSummary su
         }
 
         public DateTime RequestStartTimeUtc { get; }
-
```
ealsur, line 52: undo this file?

sourabh1007:
Can you please give proper title to this PR, operation level Open telemetry is already there. This PR is related to distributed tracing and generating and logging operation level traceid.

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
```diff
@@ -496,6 +497,7 @@ protected virtual void Dispose(bool disposing)
                                     databaseName: databaseName,
                                     operationType: operationType,
                                     requestOptions: requestOptions,
+                                    trace: trace,
```
sourabh1007, line 500: Instead of passing Trace in factory. you can add datum with key vaue pair here.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryAttributeKeys.cs
```diff
@@ -10,6 +10,7 @@ internal sealed class OpenTelemetryAttributeKeys
         public const string DiagnosticNamespace = "Azure.Cosmos";
         public const string ResourceProviderNamespace = "Microsoft.DocumentDB";
         public const string OperationPrefix = "Operation";
+        public const string NetworkLevelPrefix = "Request";
```
sourabh1007, line 13: Why are we adding it here? You must be added it already in direct package. please use the same reference here

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
sourabh1007: Open telemetry Factory is already there to take this decision, which constructor needs to be called. May be adding a proper documentation here would make more sense.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -179,6 +191,10 @@ Documents.OperationType operationType
 
                 this.scope.Dispose();
             }
+            else
```
kirankumarkolli, line 194: Isn't this.scope null check required for above if?

kirankumarkolli: Also all this.scope usages in this file right?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
kirankumarkolli: The three CTOR's contracts are confusing. 
Last one is a union one.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryAttributeKeys.cs
aavasthy: In Direct Package, NetworkLevelPrefix is not part of a constants file. It is being used in OpenTelemetryRecorderFactory
https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB?path=/Product/Microsoft.Azure.Documents/SharedFiles/Telemetry/OpenTelemetryRecorderFactory.cs&_a=contents&version=GBmaster

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: Default value is null and it is disposable. The call to openTelemetryRecorder is within using block so it will get disposed.
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/789f70183a2323259d6518eb7d463aca38f4ae08/Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs#L492

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: Added comments for better understanding of the use case of each constructor. The three constructors are being used for three different scenarios depending on which listener is available.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
```diff
@@ -49,9 +52,17 @@ internal static class OpenTelemetryRecorderFactory
...
+#if !INTERNAL
+                else if (Activity.Current is null)
+                {
+                    DiagnosticScope requestScope = LazyScopeFactory.Value.CreateScope(name: $"{OpenTelemetryAttributeKeys.NetworkLevelPrefix}.{operationName}");
 
-            return default;
+                    openTelemetryRecorder = requestScope.IsEnabled ? new OpenTelemetryCoreRecorder(scope: requestScope) : new OpenTelemetryCoreRecorder(operationName);
+                }
+#endif
+                trace.AddDatum("DistributedTraceId", Activity.Current?.Id);
```
sourabh1007, line 63: Shouldn't it be just "TraceId"

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            CustomOtelExporter.ResetData();
+        }
+
+        [DataRow(true, true, $"{OpenTelemetryAttributeKeys.DiagnosticNamespace}.Operation", 1)]
+        [DataRow(true, true, $"{OpenTelemetryAttributeKeys.DiagnosticNamespace}.Request", 1)]
+        [TestMethod]
+        public async Task OperationScopeEnabled(bool enableDistributingTracing, bool enableActivitySource, string source, int activityCount)
```
sourabh1007, line 40: if enabledistributingtracing, enableActivitySource, activitycount are not changing in Datarow, please hardcode it in the test itself.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,101 @@
...
+    using AzureCore = global::Azure.Core;
+    using Microsoft.Azure.Cosmos.Telemetry;
+    using System.Diagnostics;
+    using Microsoft.Azure.Cosmos.Tracing;
+    using System.Runtime.InteropServices;
+    using System.Globalization;
+    using System.Threading;
+
+    [VisualStudio.TestTools.UnitTesting.TestClass]
+    public sealed class DistributedTracingOTelTests
```
sourabh1007, line 25: This should cover all the scenarios like what if operation and request listeners are subscribed together or nothing is subscribed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+                client = TestCommon.CreateCosmosClient(
+                    useGateway: false,
+                    enableDistributingTracing: enableDistributingTracing);
+
+                database = await client.CreateDatabaseAsync(
+                        Guid.NewGuid().ToString(),
+                        cancellationToken: default);
+                Assert.AreEqual(activityCount, CustomOtelExporter.CollectedActivities.Count());
+                Assert.AreEqual(source, CustomOtelExporter.CollectedActivities.FirstOrDefault().Source.Name);
```
sourabh1007, line 57: please assert on header also which we are setting for context propagation.

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
```diff
@@ -102,7 +102,7 @@ public class BenchmarkConfig
         public bool EnableTelemetry { get; set; }
 
         [Option(Required = false, HelpText = "Enable Distributed Tracing")]
-        public bool EnableDistributedTracing { get; set; }
+        public bool EnableDistributedTracing { get; set; } = true;
```
sourabh1007, line 105: by default set it false.

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
sourabh1007: for any scenario, activity instance should be present here. because we are disposing it with recorder.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
ealsur: If the default value is null, then shouldn't we check before using it? On the Dispose method in this class, there is no guard against it being null

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
ealsur: One pattern to make it easier on this cases is making the constructors private and have X static methods with descriptive names.

For example:

```
public static OpenTelemetryCoreRecorder CreateForOneScenario(some parameters)

public static OpenTelemetryCoreRecorder CreateForAnotherScenario(some other parameters)

```

That way it is clear for the consumer which to use.

Last constructor seems to allow all cases though? Is that correct?

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
aavasthy: if it is kept outside in clientContextCore then IsDistributedTracingEnabled check would need to be added explicitly. But currently it is added in OpenTelemetryRecorderFactory for scenario where EnableDistributedTracing flag is on and there is already a parent activity but no listeners on network or operation level.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
aavasthy: updated and added tests for asserting header for gateway as well as direct mode.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
aavasthy: Updated tests and covering scenarios with different sources.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: Added 3 new public static methods for each constructor with description. The last constructor doesn't allow all cases, the parameters in last constructor are not optional.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: Diagnostic Scope is a struct and wouldn't be null.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: Updated the above comment. DiagnosticScope is not a class but a struct so it would not be null and would have a valid instance. So, default(DiagnosticScope) would return an instance of DiagnosticScope with '_activityAdapter':null, '_suppressedNestedClientActivities': false and 'IsEnabled': false.
https://github.com/Azure/azure-sdk-for-net/blob/dbd2b6d7b1db29e109255c1ddd2d9b284eff5196/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs#L34

DiagnosticScope dispose will dispose 'ActivityAdapter' object and will call 'Activity.Current' to end the activity and stop tracing https://github.com/Azure/azure-sdk-for-net/blob/dbd2b6d7b1db29e109255c1ddd2d9b284eff5196/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs#L398

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -54,6 +68,44 @@ internal struct OpenTelemetryCoreRecorder : IDisposable
...
+        /// <summary>
+        /// Used for creating parent activity in scenario where there are no listeners at operation level and network level
+        /// </summary>
+        public static OpenTelemetryCoreRecorder CreateParentActivity(string operation)
+        {
+            return new OpenTelemetryCoreRecorder(operation);
+        }
+
+        /// <summary>
+        /// Used for creating parent activity in scenario where there are listeners at operation level
```
ealsur, line 89: If there are present at network level, would it make a difference? My understanding is no, from reading this, but wanted to confirm

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -54,6 +68,44 @@ internal struct OpenTelemetryCoreRecorder : IDisposable
             }
         }
 
+        /// <summary>
+        /// Used for creating parent activity in scenario where there are no listeners at operation level 
+        /// but they are present at network level
+        /// </summary>
+        public static OpenTelemetryCoreRecorder CreateNetworkLevelParentActivity(DiagnosticScope scope)
```
ealsur, line 75: This is the parent scope for network? If yes, can we rename the variable accordingly to differentiate from other `scope` in the other methods?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -54,6 +68,44 @@ internal struct OpenTelemetryCoreRecorder : IDisposable
...
+        {
+            return new OpenTelemetryCoreRecorder(scope);
+        }
+
+        /// <summary>
+        /// Used for creating parent activity in scenario where there are no listeners at operation level and network level
+        /// </summary>
+        public static OpenTelemetryCoreRecorder CreateParentActivity(string operation)
```
ealsur, line 83: Nit: for consistency, this is the operation name? In the other method we used `operationName`

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: No it would not make a difference in this case if in there are listeners at network level or not. We are creating an activity here as there is listener at operation level.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -54,6 +68,44 @@ internal struct OpenTelemetryCoreRecorder : IDisposable
...
+        {
+            return new OpenTelemetryCoreRecorder(operation);
+        }
+
+        /// <summary>
+        /// Used for creating parent activity in scenario where there are listeners at operation level 
+        /// </summary>
+        public static OpenTelemetryCoreRecorder CreateOperationLevelParentActivity(
+        DiagnosticScope scope,
```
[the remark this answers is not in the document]
aavasthy: updated

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
```diff
@@ -15,43 +17,52 @@ internal static class OpenTelemetryRecorderFactory
...
+#if !INTERNAL
+                else if (Activity.Current is null)
+                {
+                    DiagnosticScope requestScope = LazyScopeFactory.Value.CreateScope(name: $"{OpenTelemetryAttributeKeys.NetworkLevelPrefix}.{operationName}");
 
-            return default;
+                    openTelemetryRecorder = requestScope.IsEnabled ? OpenTelemetryCoreRecorder.CreateNetworkLevelParentActivity(networkScope: requestScope) : OpenTelemetryCoreRecorder.CreateParentActivity(operationName);
+                }
+#endif
+                trace.AddDatum("DistributedTraceId", Activity.Current?.Id);
```
sourabh1007, line 63: Add only trace id in the request diagnostics. Instead of whole activity id.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/TestCommon.cs
```diff
@@ -149,6 +150,10 @@ internal static (string endpoint, string authKey) GetAccountInfo()
                     LatencyThresholdForDiagnosticEvent = TimeSpan.FromMilliseconds(0)
                 });
             }
+            if(httpClientFactory != null)
+            {
+                cosmosClientBuilder = cosmosClientBuilder.WithHttpClientFactory(httpClientFactory);
```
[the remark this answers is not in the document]
ealsur: You can use the `customizeClientBuilder` parameter for this.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -54,6 +68,44 @@ internal struct OpenTelemetryCoreRecorder : IDisposable
...
+        public static OpenTelemetryCoreRecorder CreateParentActivity(string operationName)
+        {
+            return new OpenTelemetryCoreRecorder(operationName);
+        }
+
+        /// <summary>
+        /// Used for creating parent activity in scenario where there are listeners at operation level 
+        /// </summary>
+        public static OpenTelemetryCoreRecorder CreateOperationLevelParentActivity(
+        DiagnosticScope operationScope,
```
ealsur, line 92: nit: padding, the parameters are normally padded in

ealsur (dismissed):

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
```diff
@@ -15,43 +17,52 @@ internal static class OpenTelemetryRecorderFactory
...
                         databaseName: databaseName,
                         operationType: operationType,
                         clientContext: clientContext,
                         config: requestOptions?.DistributedTracingOptions ?? clientContext.ClientOptions?.DistributedTracingOptions);
                 }
-            }
+#if !INTERNAL
+                else if (Activity.Current is null)
+                {
+                    DiagnosticScope requestScope = LazyScopeFactory.Value.CreateScope(name: $"{OpenTelemetryAttributeKeys.NetworkLevelPrefix}.{operationName}");
```
kirankumarkolli, line 58: Can you please expand more context? 
Scope is changing from operation -> NWLevel and its unclear to me.

aavasthy: The main aim is to have a parent activity created in all scenarios.
The first scope here checks for listeners at Operation Level. If there are listeners at operation level then scope is enabled and activity is created.
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/2837201bf9fc7c55c8a3177f6dd57306eb48ccc6/Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs#L40

The second check is to see if there isn't any existing parent id(no activity created by the client) then check if there is a listener at Network level. If there is a source at network level then create a dummy scope that creates a parent id.
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/2837201bf9fc7c55c8a3177f6dd57306eb48ccc6/Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs#L58

And if there is no listener at operation and network level then create a dummy activity with poperationName that acts as a parent Activity.
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/2837201bf9fc7c55c8a3177f6dd57306eb48ccc6/Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs#L60

aavasthy:
@msftbot make sure to wait for @sourabh1007 and @kirankumarkolli to approve

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
kirankumarkolli: > second check is to see if there isn't any existing parent id(no activity created by the client) then check if there is a listener at Network level.

Do this check needs to be conditional on no-activity?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/TestCommon.cs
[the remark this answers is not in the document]
aavasthy: Updated and now using customizeClientBuilder.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
aavasthy: Yes we only generate a network level parent activity if there isn't an existing parent activity. The whole point is to have a parentActivity at root level.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
aavasthy: Updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializer.xml
```diff
@@ -0,0 +1,48 @@
+﻿<Results>
+  <Result>
```
ealsur, line 2: This file might not be needed, doesn't sound like you are changing anything that should affect it, why is it being generated I don't know though, but doesn't sound like it should be being pushed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,212 @@
...
+                Assert.AreEqual(0, CustomOtelExporter.CollectedActivities.Count());
+            }
+        }
+
+        [TestCleanup]
+        public async Task CleanUp()
+        {
+            if (this.database != null)
+            {
+                await this.database.DeleteStreamAsync();
```
ealsur, line 205: Please use this.TestCleanUp, otherwise you are not disposing the underlying client.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/BaseCosmosClientHelper.cs
```diff
@@ -47,6 +47,18 @@ public async Task TestInit()
             await this.BaseInit(this.cosmosClient);
         }
 
+        public async Task TestInit(
```
ealsur, line 50: Isn't distributed tracing available in `CosmosClientBuilder.WithDistributedTracing`? You don't need to add this method, you can simply do:

```
customizeClientBuilder: builder => builder.WithDistributedTracing(true)
```

WIth the existing TestInit

ealsur (changes requested):
Waiting on the tests cleanup

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -179,6 +232,10 @@ Documents.OperationType operationType
 
                 this.scope.Dispose();
             }
+            else
+            {
+                this.activity?.Stop();
```
kirankumarkolli, line 237: this.activity is set only in else case right? 

One generic alternative might be to Stop() conditional on JUST this.activity set.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
```diff
@@ -15,43 +17,52 @@ internal static class OpenTelemetryRecorderFactory
...
 
-            return default;
+                    openTelemetryRecorder = requestScope.IsEnabled ? OpenTelemetryCoreRecorder.CreateNetworkLevelParentActivity(networkScope: requestScope) : OpenTelemetryCoreRecorder.CreateParentActivity(operationName);
```
kirankumarkolli, line 60: Is namespace not needed for activity creation?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
aavasthy: updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/BaseCosmosClientHelper.cs
aavasthy: Updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,218 @@
...
+
+    [VisualStudio.TestTools.UnitTesting.TestClass]
+    [TestCategory("UpdateContract")]
+    public sealed class DistributedTracingOTelTests : BaseCosmosClientHelper
+    {
+        private CosmosClientBuilder cosmosClientBuilder;
+        [TestInitialize]
+        public async Task TestInitialize()
+        {
+            await base.TestInit();
```
ealsur, line 41: Inside each test, you are creating a custom client, and also a database. What is the point then to call `base.TestInit` if nothing the TestInit is doing is something you are taking advantage of?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,218 @@
...
+            
+            string[] sources = new string[] {firstSource, secondSource};
+            sources = sources.Where(x => x != null).ToArray();
+
+            using (TracerProvider provider = Sdk.CreateTracerProviderBuilder()
+                .AddCustomOtelExporter()
+                .AddSource(sources)
+                .Build())
+            {
+                this.cosmosClientBuilder = this.cosmosClientBuilder
```
ealsur, line 63: Why not use instead: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/BaseCosmosClientHelper.cs#L38-L48

```
await base.TestInit(
  validateSinglePartitionKeyRangeCacheCall: false,
  customizeClientBuilder : (builder) => builder.WithDistributedTracing(true));
```

It sets the client, and also creates the database.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,218 @@
...
+                    return null;
+                }
+            };
+
+            using (TracerProvider provider = Sdk.CreateTracerProviderBuilder()
+                .AddCustomOtelExporter()
+                .AddSource(sources)
+                .Build())
+            {
+                this.cosmosClientBuilder = this.cosmosClientBuilder
```
ealsur, line 135: Same here and everywhere you are creating a custom client

kirankumarkolli (dismissed):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
ealsur:
```
await base.TestInit(
  validateSinglePartitionKeyRangeCacheCall: false,
  customizeClientBuilder : (builder) => builder
      .WithHttpClientFactory(() => new HttpClient(httpClientHandlerHelper))
      .WithDistributedTracing(true));
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,218 @@
...
+        {
+            AppContext.SetSwitch("Azure.Experimental.EnableActivitySource", false);
+            AzureCore.ActivityExtensions.ResetFeatureSwitch();
+            using (TracerProvider provider = Sdk.CreateTracerProviderBuilder()
+                .AddCustomOtelExporter()
+                .Build())
+            {
+                this.cosmosClientBuilder = this.cosmosClientBuilder
+                                           .WithDistributedTracing(enableDistributingTracing);
+                this.SetClient(useGateway
```
ealsur, line 178: Replace with:

```
if (useGateway)
{
await base.TestInit(
  validateSinglePartitionKeyRangeCacheCall: false,
  customizeClientBuilder : (builder) => builder
     .WithConnectionModeGateway()
     .WithDistributedTracing(true));
}
else
{
await base.TestInit(
  validateSinglePartitionKeyRangeCacheCall: false,
  customizeClientBuilder : (builder) => builder
     .WithDistributedTracing(true));
}
```

aavasthy: Updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
aavasthy: updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
aavasthy: Updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
aavasthy: updated and using the existing TestInit() now

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
aavasthy: discussed offline

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
aavasthy: discussed offline

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
```diff
@@ -15,43 +17,52 @@ internal static class OpenTelemetryRecorderFactory
...
+#if !INTERNAL
+                else if (Activity.Current is null)
+                {
+                    DiagnosticScope requestScope = LazyScopeFactory.Value.CreateScope(name: operationName);
 
-            return default;
+                    openTelemetryRecorder = requestScope.IsEnabled ? OpenTelemetryCoreRecorder.CreateNetworkLevelParentActivity(networkScope: requestScope) : OpenTelemetryCoreRecorder.CreateParentActivity(operationName);
+                }
+#endif
+                trace.AddDatum("DistributedTraceId", Activity.Current?.TraceId);
```
kirankumarkolli, line 63: Is it explicitly needed?
Our activityid's capture down steam mgiht alreay have this context right?

aavasthy: This gets added to the Diagnostic Log. We currently don't have DistributedTraceId in that.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/BaseCosmosClientHelper.cs
```diff
@@ -60,6 +60,8 @@ public async Task TestCleanup()
 
             // Only dispose if the caller set a custom client
             this.cosmosClient?.Dispose();
+
+            this.cosmosClient = null;
```
ealsur, line 64: nit: Why is this needed?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
sourabh1007: @kirankumarkolli  I don't thing we capture `Activity.Current?.TraceId` anywhere except this place. if you think that something like this already exists and should be addressed, please provide some context, maybe we can address it as part of next PR.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,210 @@
...
+    using OpenTelemetry.Trace;
+    using OpenTelemetry;
+    using AzureCore = global::Azure.Core;
+    using Microsoft.Azure.Cosmos.Telemetry;
+    using System.Diagnostics;
+    using Microsoft.Azure.Cosmos.Tracing;
+    using System.Net.Http;
+
+    [VisualStudio.TestTools.UnitTesting.TestClass]
+    public sealed class DistributedTracingOTelTests : BaseCosmosClientHelper
```
sourabh1007, line 24: Please add few more assertions:
1. Assert if we have traceid in header (Direct and TCP) both if this feature is enable. (like we have in Direct package)
2. Assert Whatever `traceid `we are putting in request diagnostics is same as what we are sending in headers.
3. Assert when we are creating dummy activity at operation level. It doesn't have any attribute. Ideally, all the attributes should be null.

We should assert everything what we expect from a feature and what we don't expect. It will be a good reference point for somebody who wants to understand the feature.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -14,11 +15,12 @@ namespace Microsoft.Azure.Cosmos.Telemetry
...
-        
-        private readonly DiagnosticScope scope;
-        private readonly DistributedTracingOptions config;
 
-        private readonly Documents.OperationType operationType;
+        private readonly DiagnosticScope scope = default;
+        private readonly DistributedTracingOptions config = null;
+        private readonly Activity activity = null;
+
+        private readonly Documents.OperationType operationType = Documents.OperationType.Invalid;
```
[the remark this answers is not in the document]
aavasthy: OperationType doesn't get set for scenarios when there is no subscription at operation level.
Documents.OperationType operationType = Documents.OperationType.Invalid has been kept as invalid as the default. OperationType is non-nullable field.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
aavasthy: - Assert if we have traceid in header (Direct and TCP) both if this feature is enable. (like we have in Direct package)
The existing test cases are already asserting that. First and second test case already assert TraceId is present in header in direct and gateway mode.
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/c217a21b5e5e3d720f8f3a631f929895ccd43d4b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs#L65
- Assert Whatever traceid we are putting in request diagnostics is same as what we are sending in headers.
Currently it is asserting that generated activity traceId is same as the diagnostics logs and same as reesponse  traceid.
- Assert when we are creating dummy activity at operation level. It doesn't have any attribute. Ideally, all the attributes should be null.
For dummy activity created via network scope it has default tags which gets added when scope is created. For summy activty created via Activity.Start(), CustomOtelExporter doesn't capture those type of activities.

sourabh1007 (approved):
LGTM.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/BaseCosmosClientHelper.cs
aavasthy: removed as part of code cleanup

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,216 @@
...
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            AppContext.SetSwitch("Azure.Experimental.EnableActivitySource", true);
+            AzureCore.ActivityExtensions.ResetFeatureSwitch();
+        }
+
+        [DataRow($"{OpenTelemetryAttributeKeys.DiagnosticNamespace}.Operation", $"{OpenTelemetryAttributeKeys.DiagnosticNamespace}.Request")]
+        [DataRow($"{OpenTelemetryAttributeKeys.DiagnosticNamespace}.Operation", null)]
+        [TestMethod]
```
ealsur, line 35: nit: If you are using DataRow, doesn't this need to be a DataTestMethod?

Prefer to add names to DataRows so when the test fails, it is clear what was being tested, for example:

```
[DataRow(param1, param2,...., DisplayName = "Checks for scenario1")]
[DataRow(param1, param2,...., DisplayName = "Checks for scenario2")]
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -0,0 +1,216 @@
...
+                                                                    .FirstOrDefault()
+                                                                    .TraceId
+                                                                    .ToString();
+            Assert.AreEqual(distributedTraceId, traceIdCreateContainer);
+
+            //Assert activity creation
+            Assert.IsNotNull(CustomOtelExporter.CollectedActivities);
+        }
+
+        [DataRow(false, true)]
```
ealsur, line 156: nit: Consider adding `DisplayName` to the DataRow so it's a bit more clear what the test case is when it fails

ealsur (dismissed):

sourabh1007 (approved):

kirankumarkolli (dismissed):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
aavasthy: updated

sourabh1007 (approved):

PR #3803, opened 2023-04-06
HttpClient: Adds Properties to the Http messages if available

ealsur:
An operation is mapped to a DocumentServiceRequest. When the DocumentServiceRequest is sent through Gateway mode, it is converted to an HttpRequestMessage.

The DocumentServiceRequest contains a list of Headers and Properties. 

Today we are just copying the Headers but not the Properties on the HttpRequestMessage.


kirankumarkolli:
Any scenario impacted by this gap?

kirankumarkolli (approved):

ealsur:
> Any scenario impacted by this gap?

Nope, no existing scenarios.

PR #3804, opened 2023-04-06
[Internal] Upgrade Resiliency: Adds Aggressive Replica Validation Logic.

kundadebdatta:
# Pull Request Template

## Description

The purpose of this PR is to validate the Unknown replicas aggressively, only for the first preferred read region and modify the AddressEnumerator sorting logic as Connected > Unknown > UnhealthyPending > Unhealthy. Because validating the Unknown replicas are aggressive in nature, our concurrency logic in RntbdOpenConnectionHandler should handle it gracefully.

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
@@ -857,17 +857,8 @@ namespace Microsoft.Azure.Cosmos.Routing
                 throw new ArgumentNullException(nameof(addresses));
             }
 
-            IEnumerable<TransportAddressUri> addressesNeedToValidateStatus = addresses
-                .Where(address => this.isFirstPreferredReadRegion ?
-                    address
-                    .GetCurrentHealthState()
-                    .GetHealthStatus() is
-                    TransportAddressHealthState.HealthStatus.Unknown or
-                    TransportAddressHealthState.HealthStatus.UnhealthyPending :
-                    address
-                    .GetCurrentHealthState()
-                    .GetHealthStatus() is
-                    TransportAddressHealthState.HealthStatus.UnhealthyPending);
+            IEnumerable<TransportAddressUri> addressesNeedToValidateStatus = this.GetAddressesNeededToValidateStatus(
+                    transportAddresses: addresses);
 
             if (addressesNeedToValidateStatus.Any())
             {
@@ -928,6 +919,31 @@ namespace Microsoft.Azure.Cosmos.Routing
             }
 
             return newAddresses;
+        }
+
+        /// <summary>
+        /// Returns a list of <see cref="TransportAddressUri"/> needed to validate their health status. Validating
+        /// a uri is done by opening Rntbd connection to that replica, which is a costly operation by nature. Therefore
+        /// vaidating both Unhealthy and Unknown replicas at the same time could impose a high CPU utilization. To avoid
+        /// this situation, validating both Unknown and Unhealthy replicas at the same time is restricted only for the
+        /// first preferred read region at the moment.
+        /// </summary>
+        /// <param name="transportAddresses">A read only list of <see cref="TransportAddressUri"/>s.</param>
+        /// <returns>A list of <see cref="TransportAddressUri"/> that needs to validate their status.</returns>
+        private IEnumerable<TransportAddressUri> GetAddressesNeededToValidateStatus(
+            IReadOnlyList<TransportAddressUri> transportAddresses)
+        {
+            return transportAddresses
+                .Where(address => this.isFirstPreferredReadRegion ?
+                    address
+                        .GetCurrentHealthState()
+                        .GetHealthStatus() is
+                            TransportAddressHealthState.HealthStatus.Unknown or
+                            TransportAddressHealthState.HealthStatus.UnhealthyPending :
+                    address
+                        .GetCurrentHealthState()
+                        .GetHealthStatus() is
+                            TransportAddressHealthState.HealthStatus.UnhealthyPending);
         }
 
         protected virtual void Dispose(bool disposing)
--- a/Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
@@ -44,6 +44,7 @@ namespace Microsoft.Azure.Cosmos.Routing
         private readonly string protocolFilter;
         private readonly ICosmosAuthorizationTokenProvider tokenProvider;
         private readonly bool enableTcpConnectionEndpointRediscovery;
+        private readonly bool isFirstPreferredReadRegion;
 
         private readonly CosmosHttpClient httpClient;
         private readonly bool isReplicaAddressValidationEnabled;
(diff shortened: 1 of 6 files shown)


Microsoft.Azure.Cosmos/src/Routing/GlobalAddressResolver.cs
```diff
@@ -215,6 +215,11 @@ public void SetOpenConnectionsHandler(IOpenConnectionsHandler openConnectionsHan
             bool forceRefresh,
             CancellationToken cancellationToken)
         {
+            request.RequestContext.FirstPreferredReadRegion = this.endpointManager
```
ealsur, line 218: This is the first Read region, not always the preferred. If the customer is not setting any preference, this is the main region or if the Preferred is marked as unavailable.

Microsoft.Azure.Cosmos/src/Routing/GlobalAddressResolver.cs
```diff
@@ -281,7 +286,8 @@ private EndpointCache GetOrAddEndpoint(Uri endpoint)
                         this.serviceConfigReader,
                         this.httpClient,
                         this.openConnectionsHandler,
-                        enableTcpConnectionEndpointRediscovery: this.enableTcpConnectionEndpointRediscovery);
+                        enableTcpConnectionEndpointRediscovery: this.enableTcpConnectionEndpointRediscovery,
+                        isFirstPreferredReadRegion: endpoint.Equals(this.endpointManager.ReadEndpoints.First()));
```
ealsur, line 290: String comparison, use CultureInvariant.IgnoreCase?

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -840,28 +844,35 @@ private static string ProtocolString(Protocol protocol)
...
         {
             if (addresses == null)
             {
                 throw new ArgumentNullException(nameof(addresses));
             }
 
-            IEnumerable<TransportAddressUri> addressesNeedToValidation = addresses
-                .Where(address => address
+            IEnumerable<TransportAddressUri> addressesNeedToValidateStatus = addresses
+                .Where(address => this.isFirstPreferredReadRegion ?
```
kirankumarkolli, line 861: Nit: May be modeling it as a method and calling might improve readability.

Microsoft.Azure.Cosmos/src/direct/StoreReader.cs
```diff
@@ -214,13 +214,15 @@ private static ReferenceCountedDisposable<StoreResult> GetStoreResultOrThrowGone
 
             bool hasGoneException = false;
             bool hasCancellationException = false;
+            bool aggressiveValidationEnabled = entity.RequestContext.RegionName?.Equals(entity.RequestContext.FirstPreferredReadRegion) ?? false;
```
ealsur, line 217: Who is populating `RegionName`?

ealsur, line 217: string comparison, should we use the StringComparison.OrdinalIgnoreCase ?

kundadebdatta: This is done by the `AddressResolver`. See this [lines](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/users/kundadebdatta/implememt_aggressive_validation_draft/Microsoft.Azure.Cosmos/src/Routing/AddressResolver.cs#L72) of code.

Microsoft.Azure.Cosmos/src/Routing/GlobalAddressResolver.cs
kundadebdatta: I think this logic should work for our implementation. Let me know your thoughts as well @xinlian12 !

Microsoft.Azure.Cosmos/src/direct/AddressEnumerator.cs
```diff
@@ -238,35 +243,61 @@ public static IEnumerable<TransportAddressUri> GetTransportAddressUrisWithPredef
...
+                if (validateUnknownReplicasAggressively)
                 {
-                    yield return transportAddressUri;
+                    if (status == TransportAddressHealthState.HealthStatus.Connected)
+                    {
+                        yield return transportAddressUri;
+                    }
+                    else if (status == TransportAddressHealthState.HealthStatus.Unknown)
+                    {
+                        unknownReplicas ??= new ();
```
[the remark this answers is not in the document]
kundadebdatta: Sure, per our earlier experiments ([see this discussion section](https://msdata.visualstudio.com/DefaultCollection/CosmosDB/_git/CosmosDB/pullRequest/945286#1670885937)), the results indicated towards using `yield return`, instead of using any other approach.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -61,7 +62,8 @@ internal class GatewayAddressCache : IAddressCache, IDisposable
             CosmosHttpClient httpClient,
             IOpenConnectionsHandler openConnectionsHandler,
             long suboptimalPartitionForceRefreshIntervalInSeconds = 600,
-            bool enableTcpConnectionEndpointRediscovery = false)
+            bool enableTcpConnectionEndpointRediscovery = false,
+            bool isFirstPreferredReadRegion = false)
```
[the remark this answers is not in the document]
kundadebdatta: So, basically our goal is to consider the "current read region" for the aggressive validation. This means, going by the code path, it will not consider a region that is unavailable.

One possible edge case for this could be, if the "current read region" is unavailable and the SDK initializes and warms up the caches on a different region other than the "current read region", and later the same unavailable "current read region" become available. What will happen is, the SDK will then start validating the replicas aggressively to that region by opening many Rntbd connections to the Unknown replicas. However, we both feel that we already have the concurrency control in place (in the `RntbdOpenConnectionHandler`) which should cover this edge case by gracefully establishing the connectivity. Cc: Annie.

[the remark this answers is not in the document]
xinlian12: this flag is setup per region, so even when the region becomes available again, we are not going to validate replicas aggressively. 

we might become too aggressive when: 
RegionA 
RegionB

- OpenAsync -> Region A is unavailable, so all the warmed up connection will be to regionB
- Region A is available, so traffic shifted back to Region A
- Connections in RegionB got closed due to idleConnectionTimeout or idleEndpointTimeout
- Region A is unavailable again, so traffic shifted to RegionB again. Because the previous established connections have closed, so we are going for aggressive validation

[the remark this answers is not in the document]
kundadebdatta: Hi @xinlian12 : Me and @kirankumarkolli had a sync up on this and we both feel that at the moment, we can use the aggressive validation for all the regions, and not just restrict it for the current read region. We can fine tune our concurrency control logic in `RntbdOpenConnectionHandler` if needed to tackle any future regressions, if at all observed.

So far, per our replica validation in perf environment, we haven't yet observed any regressions. Let me know your thoughts on this.

PR #3806, opened 2023-04-10
Client Encryption: Adds raw key value to MDE Encryption

vipulvishal-ms:
## Description

Update is to add raw key to MDE algorithm. Clients using legacy algorithm use the same key storage and API not only for Cosmos DB encryption, but also for others (Azure Storage, generic encryptor...). For this, we leverage the DataEncryptionKeyProvider.FetchDataEncryptionKeyAsync method to load the unwrapped key from the container.
But in current MdeEncryptionAlgorithm, the RawKey property is null by design. 

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos.Encryption.Custom/src/MdeServices/MdeEncryptionAlgorithm.cs
+++ b/Microsoft.Azure.Cosmos.Encryption.Custom/src/MdeServices/MdeEncryptionAlgorithm.cs
@@ -14,8 +14,10 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom
     {
         private readonly AeadAes256CbcHmac256EncryptionAlgorithm mdeAeadAes256CbcHmac256EncryptionAlgorithm;
 
+        private readonly byte[] unwrapKey;
+
         // unused for MDE Algorithm.
-        public override byte[] RawKey => null;
+        public override byte[] RawKey => this.unwrapKey;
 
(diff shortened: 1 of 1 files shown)


Microsoft.Azure.Cosmos.Encryption.Custom/src/MdeServices/MdeEncryptionAlgorithm.cs
```diff
@@ -90,9 +92,11 @@ internal sealed class MdeEncryptionAlgorithm : DataEncryptionKey
         /// <param name="dataEncryptionKey"> Data Encryption Key </param>
         /// <param name="encryptionType"> Encryption type </param>
         public MdeEncryptionAlgorithm(
+            byte[] unwrapKey,
```
kr-santosh, line 95: this is an internal class. Who is setting this ?

Microsoft.Azure.Cosmos.Encryption.Custom/src/MdeServices/MdeEncryptionAlgorithm.cs
```diff
@@ -29,6 +31,7 @@ internal sealed class MdeEncryptionAlgorithm : DataEncryptionKey
         /// <param name="encryptionType"> Encryption type </param>
         /// <param name="encryptionKeyStoreProvider"> EncryptionKeyStoreProvider for wrapping and unwrapping </param>
         public MdeEncryptionAlgorithm(
+            byte[] unwrapKey,
```
kr-santosh, line 34: not required. DataEncryptionKeyProperties has the unwrapped key.

PR #3816, opened 2023-04-18
Client Telemetry: Adds Client Config API support to control the feature

### LINKED ISSUE #3427: Cosmos account: Telemetry context (endpoint, enabled/disabled)

sourabh1007:
### Background 
Till now, Client Telemetry feature was controlled by environment variable. It means customer has to set a particular environment variable in order to enable client telemetry on SDK. but ideally, it should be controlled by portal. In portal, there will be a switch to enable/disable client telemetry. And SDK will get this information using a gateway API (similar to database account properties) and act accordingly.

As part of this PR, Gateway API is integrated, and environment variable config is removed. SDK will call gateway API in every 10 minutes to refresh this information.

### What is included in this PR?
1. Removed existing environment variable related to Client Telemetry.
2. Made sure Client Config API is called during initialization.
3. Made sure there is a background job running which keep checking the status of client telemetry config every 10min.
4. Made sure client telemetry job is starting and stopping with the flag without any re-initialization.
5. There is only one way to enable client telemetry i.e. through portal but it can be disabled by 2 ways:
   a) Instance level: by calling client builder function or setting `CosmosClientOptions.EnableClientTelemetry` to `false`.
   ~~b) Machine level : by setting environment variable `COSMOS.CLIENT_TELEMETRY_ENABLED` to `false`.~~ As per discussion with team, it is not required we can recommend customer to switch off client telemetry completely.
6. Fixing references in related projects like Benchmark and CTL.

### Public API
Not part of this PR

### Design Summary
During initialization of client, starting a Task in `DoumentClient ` which check for Account Client Config and initialize client telemetry job if it is enabled(ref. `TelemetryToServiceHelper`). This job will dynamically update the `clientTelemetryInstance` if client telemetry job is running.

### Test Coverage
1. Updated existing `ClientTelemetryTest` Emulator test, to mock ClientConfig API and always return enabled state of client telemetry
~2.  Added a Test in `GatewayTests ` to validate ClientConfig API with gateway end point~
~3.  Added logic to access ClientConfig API using AAD in `CosmosAadTests.cs`~
4.  Added tests to validate client telemetry is switching off and switching on with API response on refresh. `ClientTelemetryConfigurationTest.cs`

## Type of change

## Closing issues
To automatically close an issue: closes #3427


Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
```diff
@@ -98,8 +98,8 @@ public class BenchmarkConfig
         [Option(Required = false, HelpText = "Disable core SDK logging")]
         public bool DisableCoreSdkLogging { get; set; }
 
-        [Option(Required = false, HelpText = "Enable Client Telemetry")]
-        public bool EnableTelemetry { get; set; }
+        [Option(Required = false, HelpText = "Disable Client Telemetry")]
+        public bool DisableTelemetry { get; set; }
```
kirankumarkolli, line 102: Can you please expand on why its renamed?

Microsoft.Azure.Cosmos.Samples/Tools/CTL/Program.cs
```diff
@@ -28,9 +28,6 @@ public static async Task Main(string[] args)
             try
             {
                 CTLConfig config = CTLConfig.From(args);
-
```
kirankumarkolli, line 31: What's CTL environment accounts default? 
And do CTL workflows update to the target desired state for us?

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
sourabh1007: Client Telemetry should be enabled at account level while creating an account. But providing an option to disable it if required.

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
```diff
@@ -198,25 +198,9 @@ internal Microsoft.Azure.Cosmos.CosmosClient CreateCosmosClient(string accountKe
...
-                        Convert.ToString(this.TelemetryScheduleInSec));
-                }
-
-                if (!string.IsNullOrEmpty(this.TelemetryEndpoint))
-                {
-                    Environment.SetEnvironmentVariable(
-                        Microsoft.Azure.Cosmos.Telemetry.ClientTelemetryOptions.EnvPropsClientTelemetryEndpoint, 
-                        this.TelemetryEndpoint);
-                }
+                clientOptions.EnableClientTelemetry = false;
```
ealsur, line 203: Isn't this property obsolete now in CosmosClientOptions? On which case would the customer have a config set as `false`?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1044,33 +1040,54 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
...
+        /// <summary>
+        /// Trigger Client Telemetry job when it is enabled and not already running.
+        /// </summary>
+        private void InitializeClientTelemetry(AccountClientConfigProperties databaseAccountClientConfigs)
         {
-            if (this.ConnectionPolicy.EnableClientTelemetry)
+            if (databaseAccountClientConfigs.IsClientTelemetryEnabled() && this.ConnectionPolicy.EnableClientTelemetry)
             {
-                try
+                if (this.ClientTelemetryTask == null)
```
ealsur, line 1050: When will this initialization method be called twice?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1044,33 +1040,54 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
...
+                        DefaultTrace.TraceWarning($"Error While starting Telemetry Job : {ex.Message}. Hence disabling Client Telemetry");
+                        this.ConnectionPolicy.EnableClientTelemetry = false;
+                    }
                 }
-                catch (Exception ex)
+                else
                 {
-                    DefaultTrace.TraceInformation($"Error While starting Telemetry Job : {ex.Message}. Hence disabling Client Telemetry");
-                    this.ConnectionPolicy.EnableClientTelemetry = false;
+                    DefaultTrace.TraceInformation("Client Telemetry Job already running.");
```
ealsur, line 1075: This might get noisy if it's always tracing Information

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1044,33 +1040,54 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
...
+                if (this.ClientTelemetryTask != null)
+                {
+                    DefaultTrace.TraceInformation("Stopping Client Telemetry Job.");
+                    
+                    this.ClientTelemetryTask.Dispose();
+                    this.ClientTelemetryTask = null;
+                }
+                else
+                {
+                    DefaultTrace.TraceInformation("Client Telemetry Disabled.");
```
ealsur, line 1089: Same here, writing this for every refresh might be noisy?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6728,8 +6788,20 @@ private async Task InitializeGatewayConfigurationReaderAsync()
...
-
+            
             this.GlobalEndpointManager.InitializeAccountPropertiesAndStartBackgroundRefresh(accountProperties);
+
+            // Disable system usage for internal builds. Cosmos DB owns the VMs and already logs
+            // the system information so no need to track it.
+#if !INTERNAL
+            AccountClientConfigProperties accountClientConfigProperties = await accountReader.GetDatabaseAccountClientConfigAsync(new Uri(this.ServiceEndpoint + Paths.ClientConfigPathSegment));
+            // Safety check to make sure, client initialization won't fail if this API is not available
+            if (accountClientConfigProperties != null)
```
ealsur, line 6799: Is there a better contract than null to detect potential failures within the `GetDatabaseAccountClientConfigAsync` call?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6728,8 +6788,20 @@ private async Task InitializeGatewayConfigurationReaderAsync()
...
+            {
+                this.InitializeClientTelemetry(accountClientConfigProperties);
+                this.GlobalEndpointManager.InitializeClientTelemetryTaskAndStartBackgroundRefresh();
```
ealsur, line 6802: If the account had client telemetry disabled, this is not called. How is this called if the client enables this later on?

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -10,26 +10,31 @@ namespace Microsoft.Azure.Cosmos.Handlers
...
     using Microsoft.Azure.Cosmos.Telemetry;
+    using Microsoft.Azure.Cosmos.Tracing;
 
     internal class TelemetryHandler : RequestHandler
     {
-        private readonly ClientTelemetry telemetry;
+        private readonly CosmosClient Client;
 
-        public TelemetryHandler(ClientTelemetry telemetry)
+        public TelemetryHandler(CosmosClient client)
```
ealsur, line 19: Why pass the CosmosClient to always access nested classes? Why not pass the nested classes or the IDocumentClient interface?

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -10,26 +10,31 @@ namespace Microsoft.Azure.Cosmos.Handlers
...
             ResponseMessage response = await base.SendAsync(request, cancellationToken);
-            if (this.IsAllowed(request))
+
+            // Check if this particular operation is eligible for client telemetry collection
+            if (this.Client.DocumentClient.ClientTelemetryTask != null && this.IsAllowed(request))
             {
                 try
                 {
-                    this.telemetry
+                    this.Client.DocumentClient.ClientTelemetryTask
```
ealsur, line 37: Isn't there a better way than just jumping through this many hoops? It is going through 3 different classes.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -10,26 +10,31 @@ namespace Microsoft.Azure.Cosmos.Handlers
...
             RequestMessage request,
             CancellationToken cancellationToken)
         {
+            await this.Client.DocumentClient.EnsureValidClientAsync(NoOpTrace.Singleton);
+
             ResponseMessage response = await base.SendAsync(request, cancellationToken);
-            if (this.IsAllowed(request))
+
+            // Check if this particular operation is eligible for client telemetry collection
+            if (this.Client.DocumentClient.ClientTelemetryTask != null && this.IsAllowed(request))
```
ealsur, line 33: Any cleaner way than checking for a variable being not null to infer this? Can this be abstracted as a method that says if its enabled or not?

Microsoft.Azure.Cosmos/src/IDocumentClient.cs
```diff
@@ -78,6 +79,11 @@ internal interface IDocumentClient
         /// <seealso cref="Microsoft.Azure.Cosmos.ConsistencyLevel"/>
         Documents.ConsistencyLevel ConsistencyLevel { get; }
 
+        /// <summary>
+        /// Gets Client Telemetry Task Instance
+        /// </summary>
+        ClientTelemetry ClientTelemetryTask { get; set; }
```
ealsur, line 85: Exposing getter and setter sounds a bit weird, what is the setter for?

ealsur (changes requested):

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
sourabh1007: No, this property is not obsolete. it will be always there to give an option to customer, to disable client telemetry forcefully for a particular client instance.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: Traces are anyway noisy and it will come here, once in 10 minutes. So I don't think it is that bad.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1396,6 +1420,8 @@ public void Dispose()
 
         internal virtual Task<QueryPartitionProvider> QueryPartitionProvider => this.queryPartitionProvider.Value;
 
+        public ClientTelemetry ClientTelemetryTask { get; set; }
```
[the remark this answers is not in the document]
sourabh1007: It is clienttelemetry instance which has task in it.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6728,8 +6788,20 @@ private async Task InitializeGatewayConfigurationReaderAsync()
...
             AccountProperties accountProperties = this.accountServiceConfiguration.AccountProperties;
             this.UseMultipleWriteLocations = this.ConnectionPolicy.UseMultipleWriteLocations && accountProperties.EnableMultipleWriteLocations;
-
+            
             this.GlobalEndpointManager.InitializeAccountPropertiesAndStartBackgroundRefresh(accountProperties);
+
+            // Disable system usage for internal builds. Cosmos DB owns the VMs and already logs
+            // the system information so no need to track it.
+#if !INTERNAL
+            AccountClientConfigProperties accountClientConfigProperties = await accountReader.GetDatabaseAccountClientConfigAsync(new Uri(this.ServiceEndpoint + Paths.ClientConfigPathSegment));
```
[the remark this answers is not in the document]
sourabh1007: I don't think, we need any resiliency for this API as a first draft.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: here we have only 2 cases, either I got the data or I don't get the data. If get the data, then initialize the telemetry job else lets try again, after 10 min.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: fixing it. Thanks for the catch.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
ealsur: I understand the intent, the part that feels a bit weird is using `null`. For example, methods like `bool.TryParse` they don't rely on returning `null`, there is a clear contract that either I could parse it or I could not, here, the `null` check is a responsibility of the caller, and if it fails to do it, it generates a NullRef exception. 

You are checking for `null` by giving `null` the meaning "The API call failed", but `null` by itself does not have that meaning. Someone reading this code without context then needs to wonder, are there different cases where this could be `null`? Which are they? It is not clear just by looking at the code.

Maybe using the `TryCatch<AccountClientConfigProperties >`?

@Pilchie / @kirankumarkolli any thoughts?

sourabh1007: Changed return type to this `TryCatch<AccountClientConfigProperties>`. I think, it is better now.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
sourabh1007: did some refactoring.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
sourabh1007: Document Client is null at this place during initialization . Initializing Document Client takes time.

Microsoft.Azure.Cosmos/src/IDocumentClient.cs
sourabh1007: I don't need setter.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1044,33 +1041,54 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
...
+                            preferredRegions: this.ConnectionPolicy.PreferredLocations,
+                            globalEndpointManager: this.GlobalEndpointManager,
+                            databaseAccountClientConfigs: databaseAccountClientConfigs);
+
+                        DefaultTrace.TraceInformation("Client Telemetry Enabled.");
+                    }
+                    catch (Exception ex)
+                    {
+                        DefaultTrace.TraceWarning($"Error While starting Telemetry Job : {ex.Message}. Hence disabling Client Telemetry");
+                        this.ConnectionPolicy.EnableClientTelemetry = false;
```
ealsur, line 1071: Isn't this a user provided config? We shouldn't be modifying it?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1396,6 +1421,8 @@ public void Dispose()
 
         internal virtual Task<QueryPartitionProvider> QueryPartitionProvider => this.queryPartitionProvider.Value;
 
+        public ClientTelemetry ClientTelemetryInstance { get; set; }
```
ealsur, line 1424:
```suggestion
        public ClientTelemetry ClientTelemetryInstance { get; private set; }
```

Shouldn't the setter be private?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6728,8 +6793,22 @@ private async Task InitializeGatewayConfigurationReaderAsync()
             await this.accountServiceConfiguration.InitializeAsync();
             AccountProperties accountProperties = this.accountServiceConfiguration.AccountProperties;
             this.UseMultipleWriteLocations = this.ConnectionPolicy.UseMultipleWriteLocations && accountProperties.EnableMultipleWriteLocations;
-
+
```
ealsur, line 6796: Nit: remove blanks

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -494,6 +497,57 @@ private async void StartLocationBackgroundRefreshLoop()
...
+            }
+            catch
+            {
+                this.isBackgroundAccountRefreshActive = false;
+                throw;
+            }
+        }
+        
+#pragma warning disable VSTHRD100 // Avoid async void methods
+        public async void RefreshAccountClientConfigsAndStartClientTelemetryJob()
```
ealsur, line 519: Nit: Do you know why we use `async void` and not `async Task`?

Pilchie: Yes, please avoid `async void`

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -48,10 +51,42 @@ public TelemetryHandler(ClientTelemetry telemetry)
...
+
+        /// <summary>
+        /// Check if Client Telemetry Job is running in background.
+        /// </summary>
+        /// <param name="clientTelemetryJob"></param>
+        /// <returns>true/false</returns>
+        private bool IsClientTelemetryJobRunning(out ClientTelemetry clientTelemetryJob)
+        {
+            clientTelemetryJob = this.Client.DocumentClient.ClientTelemetryInstance;
+            if (this.Client.DocumentClient.ClientTelemetryInstance == null)
```
ealsur, line 77: Why are you checking `this.Client.DocumentClient.ClientTelemetryInstance`? If you just snapshoted on `clientTelemetryJob` shouldn't you be able to use that?

Also, this could be simplified?

```suggestion
           return clientTelemetryJob != null;
```

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
jcocchi: I think there is still an open thread on naming for this feature/ config name. We can consider something like "DisableSendingTelemetryToService" but want to make sure it's not confused with OpenTelemetry support or other types of telemetry

sourabh1007: Right now, it is not publicly exposed. So, we can go with it as it is. public service API can be discussed and covered as separate PR.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
ealsur: The `IDocumentClient` interface should suffice as input to this class.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -814,6 +816,14 @@ internal virtual ConnectionPolicy GetConnectionPolicy(int clientId)
             return connectionPolicy;
         }
 
+        private static bool IsClientTelemetryEnabled()
+        {
+            bool isTelemetryEnabled = ConfigurationManager
```
ealsur, line 821: The Description of the PR says that that environment variable is removed, should the Description be updated? Seems we still use an environment variable?

sourabh1007: I kept it to disable this feature at "Machine Level" but as per our discussion today, we don't want it. So I am going to remove it.

Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Tests/Contracts/DotNetSDKEncryptionCustomAPI.json
```diff
@@ -51,13 +51,6 @@
           ],
           "MethodInfo": "System.Threading.Tasks.Task`1[Microsoft.Azure.Cosmos.Encryption.Custom.DataEncryptionKey] FetchDataEncryptionKeyAsync(System.String, System.String, System.Threading.CancellationToken);IsAbstract:False;IsStatic:False;IsVirtual:True;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
         },
-        "System.Threading.Tasks.Task`1[Microsoft.Azure.Cosmos.Encryption.Custom.DataEncryptionKey] FetchDataEncryptionKeyWithoutRawKeyAsync(System.String, System.String, System.Threading.CancellationToken)[System.Runtime.CompilerServices.AsyncStateMachineAttribute(typeof(Microsoft.Azure.Cosmos.Encryption.Custom.CosmosDataEncryptionKeyProvider+<FetchDataEncryptionKeyWithoutRawKeyAsync>))]": {
```
ealsur, line 54: Is this related to this PR?

sourabh1007: no

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
ealsur: Had offline discussion. This is not ideal, but we are doing it on the TransportHandler too. Eventually we want to remove the DocumentClient and receiving the CosmosClient is better for that plan

ealsur (dismissed):

Microsoft.Azure.Cosmos.Samples/Tools/CTL/CTLConfig.cs
```diff
@@ -132,8 +132,7 @@ internal CosmosClient CreateCosmosClient()
         {
             CosmosClientOptions clientOptions = new CosmosClientOptions()
             {
-                ApplicationName = CTLConfig.UserAgentSuffix,
-                EnableClientTelemetry = true
+                ApplicationName = CTLConfig.UserAgentSuffix
```
kirankumarkolli, line 135: Do CTL accounts always have it enabled? Or does it need a follow-up?
At-least for perf federations we should not enable it by-default (otherwise will show-up as regression)

Microsoft.Azure.Cosmos.Samples/Tools/CTL/Program.cs
```diff
@@ -54,9 +51,7 @@ public static async Task Main(string[] args)
 
                     logger.LogInformation("Initialization completed.");
 
-                    if(client.ClientOptions.EnableClientTelemetry.GetValueOrDefault()) {
-                        logger.LogInformation("Telemetry is enabled for CTL.");
```
kirankumarkolli, line 58: Enable logging is excluded is it intentional?

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
```diff
@@ -48,7 +48,7 @@ public ConnectionPolicy()
             this.MaxConnectionLimit = defaultMaxConcurrentConnectionLimit;
             this.RetryOptions = new RetryOptions();
             this.EnableReadRequestsFallback = null;
-            this.EnableClientTelemetry = ClientTelemetryOptions.IsClientTelemetryEnabled();
+            this.DisableClientTelemetryToService = false; // it means client telemetry feature is NOT disabled forcefully for this client instance.
```
kirankumarkolli, line 51: Defaults we can inline with the property it-self. 

ex: internal book DisableXX { get; set; }= false;

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -730,9 +731,10 @@ internal Protocol ConnectionProtocol
         internal bool? EnableCpuMonitor { get; set; }
 
         /// <summary>
-        /// Flag to enable telemetry
+        /// Telemetry feature is under preview right now.
```
kirankumarkolli, line 734: If its preview, isn't below unconditonal part enabling for GA as well?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -730,9 +731,10 @@ internal Protocol ConnectionProtocol
...
+        /// By default, this can be controlled by enabling it on portal but alternatively, it can be disabled by this flag for particular CosmosClient instance.
         /// </summary>
-        internal bool? EnableClientTelemetry { get; set; }
+        public bool? DisableClientTelemetryToService { get; set; }
```
kirankumarkolli, line 737: Is the API discussion on (offline mailthread) will be reflected here?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -774,9 +776,9 @@ internal virtual ConnectionPolicy GetConnectionPolicy(int clientId)
                 ServerCertificateCustomValidationCallback = this.ServerCertificateCustomValidationCallback
             };
 
-            if (this.EnableClientTelemetry.HasValue)
+            if (this.DisableClientTelemetryToService.HasValue)
             {
-                connectionPolicy.EnableClientTelemetry = this.EnableClientTelemetry.Value;
+                connectionPolicy.DisableClientTelemetryToService = this.DisableClientTelemetryToService.Value;
```
kirankumarkolli, line 781:
```suggestion
                connectionPolicy.DisableClientTelemetryToService = this.DisableClientTelemetryToService.HasValue && this.DisableClientTelemetryToService.Value;
```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli, line 737: Do name needs to convey that its the override?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -681,23 +681,13 @@ internal CosmosClientBuilder WithCpuMonitorDisabled()
         }
 
         /// <summary>
-        /// Disable Telemetry if enabled using environment properties
+        /// Telemetry feature is under preview right now.
+        /// By default, this can be controlled by enabling it on portal but alternatively, it can be disabled by this flag for particular CosmosClient instance.
         /// </summary>
         /// <returns>The <see cref="CosmosClientBuilder"/> object</returns>
-        internal CosmosClientBuilder WithTelemetryDisabled()
+        public CosmosClientBuilder WithTelemetryToServiceDisabled()
```
kirankumarkolli, line 688: Naming: With prefix represents kind of opt-in or enable. 
For disabling how about DisableTelemetry....()

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli, line 737: For compute integration this always needs to be disable for now till we work-out the integration path forward.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6732,12 +6792,25 @@ private async Task InitializeGatewayConfigurationReaderAsync()
...
             AccountProperties accountProperties = this.accountServiceConfiguration.AccountProperties;
             this.UseMultipleWriteLocations = this.ConnectionPolicy.UseMultipleWriteLocations && accountProperties.EnableMultipleWriteLocations;
-
+            
             this.GlobalEndpointManager.InitializeAccountPropertiesAndStartBackgroundRefresh(accountProperties);
+  
+            // Disable system usage for internal builds. Cosmos DB owns the VMs and already logs
+            // the system information so no need to track it.
+#if !INTERNAL
+            TryCatch<AccountClientConfigProperties> accountClientConfigProperties = await accountReader.GetDatabaseAccountClientConfigAsync();
```
kirankumarkolli, line 6804: Blocking client initialization for telemetry set-up right? 
Is it needed? Or can the telemetry set-up be done asynchronous in background?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6732,12 +6792,25 @@ private async Task InitializeGatewayConfigurationReaderAsync()
...
+            // Make sure to intialize the client telemetry job if the client config is available
+            if (accountClientConfigProperties.Succeeded)
+            {
+                this.InitializeClientTelemetry(accountClientConfigProperties.Result);
+            }
+            
+            // No matter what, start the background job to retry calling client config API
+            this.accountClientConfigTask = this.GlobalEndpointManager.InitializeClientTelemetryTaskAndStartBackgroundRefreshAsync();
```
kirankumarkolli, line 6812: Separate refresh task vs merge with the account-refresh task. Thoughts?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1046,33 +1045,53 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
...
                 }
             }
             else
             {
-                DefaultTrace.TraceInformation("Client Telemetry Disabled.");
+                if (this.ClientTelemetryInstance != null)
+                {
+                    DefaultTrace.TraceInformation("Stopping Client Telemetry Job.");
+                    
+                    this.ClientTelemetryInstance.Dispose();
```
kirankumarkolli, line 1088: Possible race-condition: Telemetry Handler might be consuming the instance while new state(disable) concurrently disposing the instance.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1046,33 +1045,53 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
...
         }
 
-        private void InitializeClientTelemetry()
+        /// <summary>
+        /// Trigger Client Telemetry job when it is enabled and not already running.
+        /// </summary>
+        private void InitializeClientTelemetry(AccountClientConfigProperties databaseAccountClientConfigs)
         {
-            if (this.ConnectionPolicy.EnableClientTelemetry)
+            if (databaseAccountClientConfigs.IsClientTelemetryEnabled() && !this.ConnectionPolicy.DisableClientTelemetryToService)
```
kirankumarkolli, line 1053: Assuming that no-collection overhead is low/none, then other alternative model might be a flag-based collection but same instance. Thoughts?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1398,6 +1418,8 @@ public void Dispose()
 
         internal virtual Task<QueryPartitionProvider> QueryPartitionProvider => this.queryPartitionProvider.Value;
 
+        public ClientTelemetry ClientTelemetryInstance { get; private set; }
```
kirankumarkolli, line 1421: Feels inverse relationships, TelemetryHandler accessing the instance through DocumentClient.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -10,26 +10,29 @@ namespace Microsoft.Azure.Cosmos.Handlers
...
     using Microsoft.Azure.Cosmos.Core.Trace;
     using Microsoft.Azure.Cosmos.Telemetry;
+    using Microsoft.Azure.Cosmos.Tracing;
 
     internal class TelemetryHandler : RequestHandler
     {
-        private readonly ClientTelemetry telemetry;
-
-        public TelemetryHandler(ClientTelemetry telemetry)
+        private readonly CosmosClient Client;
```
kirankumarkolli, line 17: CosmosClient direct access inside handler is too-deep. (Earlier we did it as hack) any other alternatives?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -494,6 +497,38 @@ private async void StartLocationBackgroundRefreshLoop()
             this.StartLocationBackgroundRefreshLoop();
         }
 
+        public virtual Task InitializeClientTelemetryTaskAndStartBackgroundRefreshAsync()
+        {
+            return Task.Run(() => this.RefreshAccountClientConfigsAndStartClientTelemetryJobAsync(), this.cancellationTokenSource.Token);
```
kirankumarkolli, line 502: Is Task.Run necessary?

kirankumarkolli (changes requested):
Please check comments

Microsoft.Azure.Cosmos.Samples/Tools/CTL/CTLConfig.cs
sourabh1007: yes thats why I removed it...now it won't be enabled by default...you can enable/disable client telemetry for the database account, used for the CTL run, from the portal only.

Microsoft.Azure.Cosmos.Samples/Tools/CTL/Program.cs
sourabh1007: yes because you can not enable it using clientOptions. You can only disable it from clientoptions.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: For GA and Preview Package, if we can control the client telemetry from the portal itself, then we would need API to disable it in both kind of packages.
Hence, I didn't put them under Preview Condition but mentioned that in the comment.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: yes...if we come up with any specific name to disable telemetry...I think we didn't finalize anything yet.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
ealsur: Good point. There is telemetry config refresh that happens in the back, the only issue is that if we depend on the background to initialize, we might be dropping or missing the initial operations.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
ealsur: I had the same comment. We did it on TransportHandler too. I think the reasoning was that eventually we wanted to remove DocumentClient, so adding the DocumentClient reference instead would make future removal harder.

Polishing this (removal of CosmosClient from handlers) might be tackled afterwards to avoid bloating the current PR, but I had the same concern: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3816#discussion_r1198352582

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
ealsur: Missed this

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: In a sense it does override, because if the portal enables it and this disables it, it overrides the "enablement" on the portal?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
sourabh1007: fine with any name actually.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: Client Telemetry job doesn't exist for INTERNAL packages ref. https://github.com/Azure/azure-cosmos-dotnet-v3/blob/70632ba77ac294dce8bf923a11292b082a080624/Microsoft.Azure.Cosmos/src/DocumentClient.cs#L6803

But yes I will set also false for Internal packages

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: what is the concern here? Initialization getting delayed? 

IMO initialization happens only once in lifetime, and it is just an attempt to load client config also where we are not even failing the initialization if this call fails. 

After that, already there is background job which refreshed this configuration.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: I just wanted to keep it separate. If required tomorrow, we can consider that option.

Anyway I already tried that and felt that it making code complicated to understand. If you want to see how it looks like if we merge them, here is the PR https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3675

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
sourabh1007: creating an issue to track this https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3982. Please feel free to edit the description if required.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
jcocchi: Override could add confusion about overriding the enable path as well, I think it's fine to leave that off

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
jcocchi: The name should match whatever we decide for the client option too, I don't think we closed on it yet

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: Can you please explain more about it? 

Here idea is `DocumentClient` is responsible for loading any information during initialization (or trigger any job). TelemetryHandler is just checking if client telemetry job is running or not. If its running means data needs to be collected otherwise just move forward.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: addressed.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
sourabh1007: addressed.

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
sourabh1007: setting it in clientoptions

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: Set it default to true in CosmosCLientOptions for Internal packages

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6432,7 +6454,45 @@ public Task<AccountProperties> GetDatabaseAccountAsync()
         {
             return TaskHelper.InlineIfPossible(() => this.GetDatabaseAccountPrivateAsync(this.ReadEndpoint), this.ResetSessionTokenRetryPolicy.GetRequestPolicy());
         }
+        
+        async Task IDocumentClientInternal.RefreshDatabaseAccountClientConfigInternalAsync()
```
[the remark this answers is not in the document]
sourabh1007: yes..followed the similar approach on how we are loading `AccountProperties`. In `GlobalEndpointManager `we don't have instance of the reader.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -494,6 +497,38 @@ private async void StartLocationBackgroundRefreshLoop()
             this.StartLocationBackgroundRefreshLoop();
         }
 
+        public virtual Task InitializeClientTelemetryTaskAndStartBackgroundRefreshAsync()
```
kirankumarkolli, line 500: Is this method needed?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
kirankumarkolli: Assuming that the ClientTelemetry will be a singleton per client, then the Handler can be initialized with the client telemetry instance directly. 

Its the dynamic reset of the ClientTelemetry instance that's resulting in accessing from DocumentClient.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
kirankumarkolli: Please check: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3816#discussion_r1263167135

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
kirankumarkolli: Please check: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3816#discussion_r1263167135

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli: Is this change ready to-go right now? Or has dependency on service?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6432,7 +6456,45 @@ public Task<AccountProperties> GetDatabaseAccountAsync()
...
+
+                    foreach (string key in headersCollection.AllKeys())
+                    {
+                        request.Headers.Add(key, headersCollection[key]);
+                    }
+
+                    return request;
+                }
 
+                TryCatch<AccountClientConfigProperties> databaseAccountClientConfigs = await gatewayModel.GetDatabaseAccountClientConfigAsync(CreateRequestMessage,
```
kirankumarkolli, line 6489: Does it needs to be conditional on `this.ConnectionPolicy.DisableClientTelemetryToService`

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -660,7 +665,7 @@ private async Task OpenPrivateAsync(CancellationToken cancellationToken)
                     storeModel: this.GatewayStoreModel, 
                     tokenProvider: this, 
                     retryPolicy: this.retryPolicy,
-                    clientTelemetry: this.clientTelemetry);
+                    clientTelemetry: this.ClientTelemetryInstance);
```
kirankumarkolli, line 668: When at client starts its not enabled but later its enabled, how is caches telemetry instance refreshed?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1046,33 +1045,55 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
...
+                            httpClient: this.httpClient,
+                            userAgent: this.ConnectionPolicy.UserAgentContainer.BaseUserAgent,
+                            connectionMode: this.ConnectionPolicy.ConnectionMode,
+                            authorizationTokenProvider: this.cosmosAuthorization,
+                            diagnosticsHelper: DiagnosticsHandlerHelper.Instance,
+                            preferredRegions: this.ConnectionPolicy.PreferredLocations,
+                            globalEndpointManager: this.GlobalEndpointManager,
+                            databaseAccountClientConfigs: databaseAccountClientConfigs);
+
+                        DefaultTrace.TraceInformation("Client Telemetry Enabled.");
```
kirankumarkolli, line 1070: TraceVerbose

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1046,33 +1045,55 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
...
             {
-                DefaultTrace.TraceInformation("Client Telemetry Disabled.");
+                if (this.ClientTelemetryInstance != null)
+                {
+                    DefaultTrace.TraceInformation("Stopping Client Telemetry Job.");
+
+                    ClientTelemetry tempTask = this.ClientTelemetryInstance;
+                    this.ClientTelemetryInstance = null;
+
+                    tempTask.Dispose();
```
kirankumarkolli, line 1091: Race condition: handler might still be using the instance while its dispose here right?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: Deployment of client configuration service is already started. i will merge this PR...as soon as it is deployed in production completely.

sourabh1007: Order of release would be ClientConfig API -> SDK release -> portal release

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
sourabh1007: Making this internal and I will create a new PR covering only all public API related to telemetry.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: We need client telemetry instance to pass it in collection cache call (or more calls). Having it in `DocumentClient `, gives me flexibility to pass it anywhere and call collect method.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: fixed and added test for this

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -48,10 +51,42 @@ public TelemetryHandler(ClientTelemetry telemetry)
...
+        }
+
+        /// <summary>
+        /// Check if Client Telemetry Job is running in background.
+        /// </summary>
+        /// <param name="clientTelemetryJob"></param>
+        /// <returns>true/false</returns>
+        private bool IsClientTelemetryJobRunning(out ClientTelemetry clientTelemetryJob)
+        {
+            clientTelemetryJob = this.TelemetryToServiceHelper.clientTelemetryInstance;
```
kirankumarkolli, line 76: Concurrency: Dispose and usage still possible right?

Microsoft.Azure.Cosmos/src/Routing/ClientCollectionCache.cs
```diff
@@ -215,10 +215,11 @@ await this.storeModel.ProcessMessageAsync(request))
                             {
                                 ContainerProperties containerProperties = CosmosResource.FromStream<ContainerProperties>(response);
 
-                                if (this.clientTelemetry != null)
+                                ClientTelemetry telemetryInstance = this.telemetryToServiceHelper?.clientTelemetryInstance;
+                                if (telemetryInstance != null)
```
kirankumarkolli, line 219: If-else getting spilled all through the collection places right? 
Like if we are adding more collection places (ex: NW etc..) then the possibility of coding gaps shows-up right? Thoughts?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,187 @@
...
+            this.clientId = clientId;
+            this.cosmosAuthorization = cosmosAuthorization;
+            this.httpClient = httpClient;
+            this.connectionPolicy = connectionPolicy;
+            this.serviceEnpoint = serviceEndpoint;
+            this.globalEndpointManager = globalEndpointManager;
+            this.cancellationTokenSource = cancellationTokenSource;
+        }
+
+        public static TelemetryToServiceHelper StartBackgroundJobForClientConfigAndTelemetry(string clientId,
```
kirankumarkolli, line 54: nit: Naming of factory method will be like CreateAndInitialize etc...

kirankumarkolli, line 54: Do every client instance creates a new ToServiceHelper or shares?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,187 @@
...
+            ConnectionPolicy connectionPolicy,
+            AuthorizationTokenProvider cosmosAuthorization,
+            CosmosHttpClient httpClient,
+            Uri serviceEndpoint,
+            GlobalEndpointManager globalEndpointManager,
+            CancellationTokenSource cancellationTokenSource)
+        {
+            if (connectionPolicy.DisableClientTelemetryToService)
+            {
+                return null;
```
kirankumarkolli, line 64: it will result in NRE right?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        private void Initialize()
+        {
+            this.accountClientConfigTask = Task.Run(() => this.RefreshDatabaseAccountClientConfigInternalAsync());
+        }
+
+        private async Task RefreshDatabaseAccountClientConfigInternalAsync()
+        {
+            while (!this.cancellationTokenSource.IsCancellationRequested)
```
kirankumarkolli, line 81: Error handling?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,187 @@
...
+                else
+                {
+                    DefaultTrace.TraceInformation("Client Telemetry Disabled.");
+                }
+            }
+        }
+
+        public void Dispose()
+        {
+            this.clientTelemetryInstance?.Dispose();
```
kirankumarkolli, line 184: cancellationToken cancel?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: yes every client instance has its own ToServiceHelper...it is not shared between clients

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
sourabh1007: No it is part of handler, if request is here means there is an active request going on right now, during that I don't think anybody would dispose the client.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: Added Trace

Microsoft.Azure.Cosmos/src/Routing/ClientCollectionCache.cs
sourabh1007: Implemented NoOps Client Telemetry Collector

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+
+            telemetryToServiceHelper.Initialize();
+
+            return telemetryToServiceHelper;
+#endif
+        }
+
+        private void Initialize()
+        {
+            this.accountClientConfigTask = Task.Run(() => this.RefreshDatabaseAccountClientConfigInternalAsync());
```
kirankumarkolli, line 87: Task.Run() is it needed?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+            while (!this.cancellationTokenSource.IsCancellationRequested)
+            {
+                Uri serviceEndpointWithPath = new Uri(this.serviceEnpoint + Paths.ClientConfigPathSegment);
+
+                TryCatch<AccountClientConfigProperties> databaseAccountClientConfigs = await this.GetDatabaseAccountClientConfigAsync(this.cosmosAuthorization, this.httpClient, serviceEndpointWithPath);
+                if (databaseAccountClientConfigs.Succeeded)
+                {
+                    this.InitializeClientTelemetry(databaseAccountClientConfigs.Result);
+                }
+                else if (!this.cancellationTokenSource.IsCancellationRequested)
```
kirankumarkolli, line 101: cancellationTokenSource is not wired into the AccountConfigAsync() right? 
Is this check really needed?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+                    DefaultTrace.TraceVerbose("Client Telemetry Job already running.");
+                }
+            }
+            else
+            {
+                if (this.clientTelemetryCollector is not ClientTelemetryCollectorsNoOp)
+                {
+                    DefaultTrace.TraceInformation("Stopping Client Telemetry Job.");
+
+                    this.clientTelemetryCollector.Dispose();
```
kirankumarkolli, line 187: Concurrent Dispose() while it might be in-use for telemetry collection right? 

Is explicit Dispose() really needed? Can we leave it to Finalizer()?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+    using Microsoft.Azure.Cosmos.Routing;
+    using Microsoft.Azure.Cosmos.Tracing;
+    using Microsoft.Azure.Documents;
+    using Microsoft.Azure.Documents.Collections;
+
+    internal class TelemetryToServiceHelper : IDisposable
+    {
+        internal static int DefaultBackgroundRefreshClientConfigTimeIntervalInMS = 10 * 60 * 1000;
+
+        internal IClientTelemetryCollectors clientTelemetryCollector = new ClientTelemetryCollectorsNoOp();
```
kirankumarkolli, line 24: It's an implicit contract on using collector right? 
At-least please make it an explicit public contract.

Microsoft.Azure.Cosmos/src/Telemetry/IClientTelemetryCollectors.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Net;
+    using Microsoft.Azure.Cosmos.Tracing;
+    using Microsoft.Azure.Documents;
+
+    internal interface IClientTelemetryCollectors : IDisposable
+    {
+        public void CollectCacheInfo(string cacheRefreshSource,
```
kirankumarkolli, line 15: Modeling arguments as type and then having an API like Collection(Action<CacheInfo> ) will enable TelemetryToServiceHelper be the collector and make decisions on enable/disable with-out exposing it upstream. Thoughts?

sourabh1007: It would be a major design change. I would take it as separate PR (just after, this one is approved) ...in this PR I just did refactor of the existing code.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: somebody has to tell the collector that..now stop the background task.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+        }
+
+        public void Dispose()
+        {
+            if (this.cancellationTokenSource != null && !this.cancellationTokenSource.IsCancellationRequested)
+            {
+                this.cancellationTokenSource.Cancel();
+                this.cancellationTokenSource.Dispose();
```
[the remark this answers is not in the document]
sourabh1007: Now passing cancellation token to account config also. So cancelling this call if dispose is called.

Microsoft.Azure.Cosmos/src/IDocumentClient.cs
```diff
@@ -78,6 +79,11 @@ internal interface IDocumentClient
         /// <seealso cref="Microsoft.Azure.Cosmos.ConsistencyLevel"/>
         Documents.ConsistencyLevel ConsistencyLevel { get; }
 
+        /// <summary>
+        /// Gets Client Telemetry Instance with Task
+        /// </summary>
+        TelemetryToServiceHelper TelemetryToServiceHelper { get; set; }
```
kirankumarkolli, line 85: Who is the consumer?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1398,6 +1370,9 @@ public void Dispose()
 
         internal virtual Task<QueryPartitionProvider> QueryPartitionProvider => this.queryPartitionProvider.Value;
 
+        internal TelemetryToServiceHelper TelemetryToServiceHelper { get; set; }
+        TelemetryToServiceHelper IDocumentClient.TelemetryToServiceHelper { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
kirankumarkolli, line 1374: Who is consumer? Is it needed?

Microsoft.Azure.Cosmos/src/Handler/ClientPipelineBuilder.cs
```diff
@@ -45,15 +45,16 @@ internal class ClientPipelineBuilder
 #if !INTERNAL
```
kirankumarkolli, line 45: Unrelated to this PR: 

For if-else combination as convention its recommended to use #if <DEFINE> 
!<DEFINE> impacts readability.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -10,48 +10,71 @@ namespace Microsoft.Azure.Cosmos.Handlers
...
 
-        private bool IsAllowed(RequestMessage request)
+        /// <summary>
+        /// Check if Collection should happen or not, if yes then return the client job instance where information needs to send
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns>true/false</returns>
+        private bool IsEligibleForTelemetryCollection(RequestMessage request)
+        {
+            return this.IsRequestAllowed(request);
```
kirankumarkolli, line 69: Is a wrapper method here needed?

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -10,48 +10,71 @@ namespace Microsoft.Azure.Cosmos.Handlers
...
             return response;
         }
 
-        private bool IsAllowed(RequestMessage request)
+        /// <summary>
+        /// Check if Collection should happen or not, if yes then return the client job instance where information needs to send
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns>true/false</returns>
+        private bool IsEligibleForTelemetryCollection(RequestMessage request)
```
kirankumarkolli, line 67: Can be static right?

Microsoft.Azure.Cosmos/src/Telemetry/Collector/IClientTelemetryCollectors.cs
```diff
@@ -0,0 +1,16 @@
...
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Telemetry
+{
+    using System;
+    using Microsoft.Azure.Cosmos.Telemetry.Collector;
+
+    internal interface IClientTelemetryCollectors : IDisposable
+    {
+        public void CollectCacheInfo(Func<CacheTelemetryData> functionFordata);
```
kirankumarkolli, line 12: Please add documentation on why Fun<T> arguments are used?

Microsoft.Azure.Cosmos/src/Telemetry/Collector/OperationTelemetryData.cs
```diff
@@ -0,0 +1,25 @@
...
+        internal HttpStatusCode statusCode { get; set; }
+        internal long responseSizeInBytes { get; set; }
+        internal string containerId { get; set; }
+        internal string databaseId { get; set; }
+        internal OperationType operationType { get; set; }
+        internal ResourceType resourceType { get; set; }
+        internal string consistencyLevel { get; set; }
+        internal double requestCharge { get; set; }
+        internal SubStatusCodes subStatusCode { get; set; }
+        internal ITrace trace { get; set; }
```
kirankumarkolli, line 23: Its not data? Why it it part of the data model?

Microsoft.Azure.Cosmos/src/Routing/ClientCollectionCache.cs
```diff
@@ -215,20 +216,19 @@ await this.storeModel.ProcessMessageAsync(request))
...
-                                                    requestLatency: response.RequestStats.RequestLatency,
-                                                    statusCode: response.StatusCode,
-                                                    containerId: collectionName,
-                                                    operationType: request.OperationType,
-                                                    resourceType: request.ResourceType,
-                                                    subStatusCode: response.SubStatusCode,
-                                                    databaseId: databaseName);
-                                }
+                                this.telemetryToServiceHelper.CollectCacheInfo(
+                                    () =>
```
kirankumarkolli, line 220: containerId and databaseId are excluded is that expected? 

If so do the data-model really need them?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: Removed

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -38,6 +37,8 @@ internal class ClientTelemetry : IDisposable
         private readonly CancellationTokenSource cancellationTokenSource;
         private readonly GlobalEndpointManager globalEndpointManager;
 
+        private readonly Uri endpointUrl;
+
```
kirankumarkolli, line 41: With factory pattern CTOR is expected to be private.

Microsoft.Azure.Cosmos/src/IDocumentClient.cs
sourabh1007: removed. remaining of last changes :(

Microsoft.Azure.Cosmos/src/Handler/ClientPipelineBuilder.cs
sourabh1007: switched

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,253 @@
...
+    using Microsoft.Azure.Cosmos.Handler;
+    using Microsoft.Azure.Cosmos.Query.Core.Monads;
+    using Microsoft.Azure.Cosmos.Resource.Settings;
+    using Microsoft.Azure.Cosmos.Routing;
+    using Microsoft.Azure.Cosmos.Telemetry.Collector;
+    using Microsoft.Azure.Cosmos.Tracing;
+    using Microsoft.Azure.Documents;
+    using Microsoft.Azure.Documents.Collections;
+
+    internal class TelemetryToServiceHelper : IClientTelemetryCollectors, IDisposable
```
kirankumarkolli, line 21: How about name it as "Collector" suffix in-place of Helper.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
sourabh1007: removed

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
sourabh1007: n/a

Microsoft.Azure.Cosmos/src/Telemetry/Collector/OperationTelemetryData.cs
sourabh1007: It is not part of data model which we save to kusto. It is needed to get the network level information for the operation.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: it is private. other constructor is removed

Microsoft.Azure.Cosmos/src/Routing/ClientCollectionCache.cs
sourabh1007: it is expected. in this case containerd and databaseid is getting fetched from the `collectionLink`. This is not related to data modal. It is required to generate the data model by the collector.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1346,9 +1318,10 @@ public void Dispose()
                 this.initTaskCache = null;
             }
 
-            if (this.clientTelemetry != null)
+            if (this.TelemetryToServiceHelper != null)
             {
-                this.clientTelemetry.Dispose();
+                this.TelemetryToServiceHelper.Dispose();
```
kirankumarkolli, line 1323: GEM is used by TelemetryServiceHelper and order of dispose is conflicting. All dependencies of GEM should be disposed before its disposal.

Microsoft.Azure.Cosmos/src/IDocumentClient.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Azure.Cosmos
     using System.Threading.Tasks;
     using Microsoft.Azure.Cosmos.Linq;
     using Microsoft.Azure.Cosmos.Query.Core;
+    using Microsoft.Azure.Cosmos.Telemetry;
```
kirankumarkolli, line 13: Undo changes to this file

Microsoft.Azure.Cosmos/src/Telemetry/Collector/OperationTelemetryData.cs
kirankumarkolli: TelemetryData types are ambiguous used for Handler -> Collector and Collector -> Service contracts interchangeably. 

It's unclear which part are expected at what layer.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
kirankumarkolli: Please check for other such dependencies, ex: Httpclient

Microsoft.Azure.Cosmos/src/Telemetry/Collector/TelemetryToServiceCollector.cs
```diff
@@ -0,0 +1,252 @@
...
+        public void CollectCacheInfo(Func<CacheTelemetryInformation> functionFordata)
+        {
+            if (this.IsClientTelemetryJobNotRunning())
+            {
+                return;
+            }
+
+            CacheTelemetryInformation data = functionFordata();
+
+            if (data.collectionLink != null)
```
kirankumarkolli, line 100: Is null expected?

kirankumarkolli: How to interpret data with-out database/collection information

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: _httpclient_ and _authorizaion_ are used in `ClientTelemetry` class which gets disposed as part of the `TelemetryServiceHelper`. Hence disposal is required before disposal of `TelemetryToServiceHelper` class.

Microsoft.Azure.Cosmos/src/Telemetry/Collector/TelemetryToServiceCollector.cs
```diff
@@ -0,0 +1,252 @@
...
+                if (databaseAccountClientConfigs.Succeeded)
+                {
+                    this.InitializeClientTelemetry(databaseAccountClientConfigs.Result);
+                }
+                else if (!this.cancellationTokenSource.IsCancellationRequested)
+                {
+                    DefaultTrace.TraceWarning($"Exception while calling client config " + databaseAccountClientConfigs.Exception.ToString());
+                }
+
+                await Task.Delay(DefaultBackgroundRefreshClientConfigTimeIntervalInMS);
```
kirankumarkolli, line 142: Wire cancellationToken otherwise the client might needs to wait for DefaultBackgroundRefreshClientConfigTimeIntervalInMS.

Also Task.Delay might throw with CancellationToken

Microsoft.Azure.Cosmos/src/Telemetry/Collector/OperationTelemetryData.cs
sourabh1007: All the Collector related files are here: https://github.com/Azure/azure-cosmos-dotnet-v3/tree/9877de40a7dbe5979bd857e8f4bd51869d1d9dd4/Microsoft.Azure.Cosmos/src/Telemetry/Collector

Service Contract related files are here: https://github.com/Azure/azure-cosmos-dotnet-v3/tree/ebd1b91a26547721ef232ef597d878144889b80b/Microsoft.Azure.Cosmos/src/Telemetry/Models

If you have any specific suggestion, please let me know I would definitely cover it as part of another PR. Already I have made lots of design changes as part of this PR. We can have focused discussion also without blocking the actual functionality.

kirankumarkolli:
Summary of offline syncup:

Functional: 
- Post disable detection: near time (possible some delay) stop collection
- The workload characteristics to be uniform: Spiky exceptions (possibel high CPU)
	- Few exception: Is okey
	- Exceptions of order of requests: Is not okey (10K rps)


C# model: 
- Ideal: Not to have pattern accessing Disposed Object


GoodToHave: 
- [Dispose] ClientTelemetry object ASAP (assuming limited instances or single instance)


Option1:
	- RealCollector & NoOpCollector
	- this.clientTelemetry: Keep switching between them? 
		- Inflight ones: using any will get drained out

sourabh1007:
As @kirankumarkolli suggested splitting this PR into small PRs:
1. Collector abstraction (it seems this is independent separate change) right?
2. Making Telemetry dynamically turn ON/OFF (handling the concurrency/dispose aspects) - Default OFF.
3. Integrating the Account config (With default still OFF, i.e. customer opt-in)

sourabh1007:
CLosing this PR as it is covered as part of attached PRs

PR #3821, opened 2023-04-19
Performance: Adds AuthorizationHelper Improvements to ConvertToBase64string method

aavasthy:
# Pull Request Template

## Description
Instead of using ‘ArrayPool<byte>.Shared.Rent(capacity)’ to allocate byte array temporarily, we can use stackalloc byte[]. It avoids the overhead of allocating and deallocating the buffer .

| **Method**| **Mean** | **Error** | **StdDev** | **StdErr**| **Min** | **Q1** | **Median** |**Q3**| **Max** | **P80** |  |**memory**|
| -------| -------- | ---------- | --------| -------| -------- | ---------- | --------|-------| -------- | ---------- | --------|-------| 
| CreateSignatureGeneration-Master | 685.5 ns| 12.20 ns| 17.50 ns| 3.31 ns | 666.7 ns| 673.9 ns| 679.6 ns| 694.6 ns | 738.8 ns| 694.6 ns| 1458899| 536 B | 
| ReadSignatureGeneration-Master | 679.5 ns| 8.46 ns| 12.84 ns| 2.53 ns | 657.5 ns| 671.5 | 676.4 ns | 685.9 ns| 713.9 ns| 669.3 ns| 1471895 | 536 B|
|CreateSignatureGeneration-Updated | 704.3 ns| 11.81 ns| 17.67 ns| 3.23 ns| 666 ns | 692.2 ns | 707.4 ns| 716.9 ns| 739.3 ns| 719.2 ns | 1,419,871.9| 544 B| 
| ReadSignatureGeneration-Updated| 734.4 ns| 20.05 ns| 30.00 ns| 5.48 ns | 670.6 ns| 718.0 ns| 731.9 ns| 746.7 ns | 789.4 ns| 761.8 ns| 1361727.4 | 544 B |

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,34 +782,18 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
             // Create a large enough buffer that URL encode can use it.
             // Increase the buffer by 3x so it can be used for the URL encoding
             int capacity = Base64.GetMaxEncodedToUtf8Length(hashPayLoad.Length) * 3;
-            byte[] rentedBuffer = ArrayPool<byte>.Shared.Rent(capacity);
+            Span<byte> encodingBuffer = stackalloc byte[capacity];
```
visridha, line 785: How do we prevent stackoverflows here? This is an arbitrary length entity. We cannot allocate random sizes on the stack.

visridha (changes requested):

kirankumarkolli:
Would be great to include memory also part of diagnostics.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
kirankumarkolli, line 785: stackalloc might return non-initialized allocation. If boundary aligned etc... might have unexpected data, Please clear it.

aavasthy: Can include a MaxCapacity value. This will prevent stack overflow.
int capacity = Math.Min(Base64.GetMaxEncodedToUtf8Length(hashPayLoad.Length) * 3, MaxCapacity)

aavasthy: Updated

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -781,35 +781,21 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
         {
             // Create a large enough buffer that URL encode can use it.
             // Increase the buffer by 3x so it can be used for the URL encoding
-            int capacity = Base64.GetMaxEncodedToUtf8Length(hashPayLoad.Length) * 3;
-            byte[] rentedBuffer = ArrayPool<byte>.Shared.Rent(capacity);
+            const int MaxCapacity = 524288; //0.5 MB
```
ealsur, line 784: nit: Should the const be outside the method scope?

aavasthy: updated

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
visridha: Allocating 0.5 MB on the stack is not great - especially when you have 32-bit applications. Given this thread:
https://social.msdn.microsoft.com/Forums/vstudio/en-US/d3ee9fd7-eb73-4902-9c56-7dbe44aaea41/max-size-possible-allocated-by-stackalloc?forum=csharpgeneral

You're using up to 1/2 of the thread stack here. Given any reasonably deep stack this will StackOverflow... 

Please also see this thread: https://stackoverflow.com/questions/785226/practical-use-of-stackalloc-keyword

Realistically i would not allocate more than 4-5K on the stack. @kirankumarkolli thoughts?

kirankumarkolli: That's a great point @visridha. 4K seems good. 
Am still unsure about the ROI.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -764,7 +766,10 @@ public static Uri GenerateUriFromAddressRequestUri(Uri uri)
 
                 payload = new ArrayOwner(ArrayPool<byte>.Shared, new ArraySegment<byte>(buffer, 0, length));
                 byte[] hashPayLoad = stringHMACSHA256Helper.ComputeHash(payload.Buffer);
-                return AuthorizationHelper.OptimizedConvertToBase64string(hashPayLoad, urlEncode);
+                using (Trace trace = Trace.GetRootTrace(nameof(GenerateAuthorizationTokenWithHashCore), TraceComponent.Authorization, TraceLevel.Info))
```
kirankumarkolli, line 769: Trace will be more overhead than this optimization brings. 
Each trace nesting creates lists over heap.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -25,6 +26,7 @@ internal static class AuthorizationHelper
         public const int DefaultMasterTokenExpiryInSeconds = 900;
         private const int MaxAadAuthorizationHeaderSize = 16 * 1024;
         private const int MaxResourceTokenAuthorizationHeaderSize = 8 * 1024;
+        private const int MaxCapacity = 4096;
```
kirankumarkolli, line 29: nit: Suffix the units (ex: Bytes)

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
aavasthy: Yes I checked the performance test's mean metric increased because of this trace change

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -777,39 +782,25 @@ public static Uri GenerateUriFromAddressRequestUri(Uri uri)
...
         /// This avoids the over head of converting it to a string and back to a byte[].
         /// </summary>
-        private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad, bool urlEncode)
+        private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad, bool urlEncode, ITrace trace = null)
         {
             // Create a large enough buffer that URL encode can use it.
             // Increase the buffer by 3x so it can be used for the URL encoding
-            int capacity = Base64.GetMaxEncodedToUtf8Length(hashPayLoad.Length) * 3;
-            byte[] rentedBuffer = ArrayPool<byte>.Shared.Rent(capacity);
+            int capacity = Math.Min(Base64.GetMaxEncodedToUtf8Length(hashPayLoad.Length) * 3, MaxCapacity);
```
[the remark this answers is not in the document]
aavasthy: I can include a logic to increase the capacity of the buffer until it is large enough to fit the data in case when computed length is greater than the max capacity

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
kirankumarkolli: I suppose this is used just for measurement and will be undone before commit right?

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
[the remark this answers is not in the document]
kirankumarkolli: One other alternative is to use Heap pool allocation like what was before as fall-back. 
Kind of try with stackalloc model.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -781,35 +785,26 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
...
-                }
+            OperationStatus status = Base64.EncodeToUtf8(hashPayLoad, encodingBuffer, out int _, out int bytesWritten);
 
-                return urlEncode 
-                    ? AuthorizationHelper.UrlEncodeBase64SpanInPlace(encodingBuffer, bytesWritten)
-                    : Encoding.UTF8.GetString(encodingBuffer.Slice(0, bytesWritten));
+            if (bytesWritten > MaxCapacity)
+            {
+               capacity = bytesWritten;
+               encodingBuffer = new byte[capacity];
```
kirankumarkolli, line 797: Isn't previous ArrayPool.Rent better than the new allocation?

aavasthy: Original code:
The original code uses ArrayPoolvwhich stores the encoded bytes and performs url encoding. It does release the rented buffer back to the shared pool once it is no longer needed but it has memory allocation issue. It always rents a buffer with a capacity which is 3 times the maximum encoded length of the input byte array. So if the input byte array is small even then the rented buffer is larger than necessary resulting in wasted memory.
Additionally, if the input byte array is very large then the rented buffer may not be large enough to accommodate the encoded bytes.

Updated code:
The updated code seems better in terms of memory allocation. The code first sets a maximum capacity and then computes the required capacity for the encoded byte array. The computed capacity is capped at the maximum capacity to avoid allocating a buffer that is too large. This avoids wasting memory.

The code uses stackalloc which allocates memory on the stack and accessing memory on stack is faster. It also avoids the overhead of calling ArrayPool.Shared.Rent and ArrayPool.Shared.Return.

The updated code also handles the scenario when computed length is greater than maxCapacity.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
aavasthy: updated to 4096

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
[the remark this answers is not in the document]
aavasthy: The current implementation is a hybrid where it uses stackalloc in case the calculated capacity is less than MaxCapacity and a new byte array is allocated on the heap when capacity is higher.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
kirankumarkolli: Is it expected?

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,34 +786,19 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
...
-                    hashPayLoad,
-                    encodingBuffer,
-                    out int _,
-                    out int bytesWritten);
-
-                if (status != OperationStatus.Done)
-                {
-                    throw new ArgumentException($"Authorization key payload is invalid. {status}");
-                }
+            encodingBuffer.Clear();
```
kirankumarkolli, line 791: For heap allocation its unnecessary overhead.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,34 +786,19 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
             // Create a large enough buffer that URL encode can use it.
             // Increase the buffer by 3x so it can be used for the URL encoding
             int capacity = Base64.GetMaxEncodedToUtf8Length(hashPayLoad.Length) * 3;
-            byte[] rentedBuffer = ArrayPool<byte>.Shared.Rent(capacity);
+            Span<byte> encodingBuffer = capacity <= MaxCapacityInBytes ? stackalloc byte[capacity] : new byte[capacity];
```
kirankumarkolli, line 789: New allocation vs ArrayPool: Last time moved to ArrayPool part of optimization. 
Is this change intentional?

kirankumarkolli (changes requested):
Please check comments

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
aavasthy: So the implementation involves using stackalloc only when the computed capacity is less than or equal to MaxCapacity otherwise use heap allocation.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
aavasthy: I need it in case of stack allocation. I can put it under if block to remove this overhead.

aavasthy: Added under the if check only for stack allocation

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
aavasthy: using Array pool for scenarios where calculated capacity > maximum capacity

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,32 +785,30 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
...
-                    ? AuthorizationHelper.UrlEncodeBase64SpanInPlace(encodingBuffer, bytesWritten)
-                    : Encoding.UTF8.GetString(encodingBuffer.Slice(0, bytesWritten));
+                return urlEncode
+                ? AuthorizationHelper.UrlEncodeBase64SpanInPlace(encodingBuffer, bytesWritten)
+                : Encoding.UTF8.GetString(encodingBuffer.Slice(0, bytesWritten));
             }
             finally
             {
-                if (rentedBuffer != null)
+                if (capacity > MaxCapacityInBytes)
```
kirankumarkolli, line 809: Decision of maxCapacity was repeated multiple times. It can be captured as a flag above and re-used across.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,32 +785,31 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
...
+                ? AuthorizationHelper.UrlEncodeBase64SpanInPlace(encodingBuffer, bytesWritten)
+                : Encoding.UTF8.GetString(encodingBuffer.Slice(0, bytesWritten));
             }
             finally
             {
-                if (rentedBuffer != null)
+                if (!useStackAlloc)
                 {
-                    ArrayPool<byte>.Shared.Return(rentedBuffer);
+                    ArrayPool<byte>.Shared.Return(encodingBuffer.ToArray());
```
kirankumarkolli, line 812: https://learn.microsoft.com/en-us/dotnet/api/system.span-1.toarray?view=net-7.0#system-span-1-toarray

its a copy right? not the original array rented?
Alternative: Hold byte[] reference and use it.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,32 +785,33 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
...
 
-                return urlEncode 
+                return urlEncode
                     ? AuthorizationHelper.UrlEncodeBase64SpanInPlace(encodingBuffer, bytesWritten)
                     : Encoding.UTF8.GetString(encodingBuffer.Slice(0, bytesWritten));
             }
             finally
             {
-                if (rentedBuffer != null)
+                if (rentedArray != null)
```
kirankumarkolli, line 812: Decision model is based on multiple conditions, readability is non-trivial. 
One option is to base it on one/few conditions ex: useStackAlloc

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,32 +785,33 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
...
-                // This replaces the Convert.ToBase64String
-                OperationStatus status = Base64.EncodeToUtf8(
-                    hashPayLoad,
-                    encodingBuffer,
-                    out int _,
-                    out int bytesWritten);
+                if (useStackAlloc)
+                {
+                    encodingBuffer.Clear();
+                }
```
DmitriMelnikov, line 798: This is redundant operation.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,32 +785,33 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
...
                     ? AuthorizationHelper.UrlEncodeBase64SpanInPlace(encodingBuffer, bytesWritten)
                     : Encoding.UTF8.GetString(encodingBuffer.Slice(0, bytesWritten));
             }
             finally
             {
-                if (rentedBuffer != null)
+                if (rentedArray != null)
                 {
-                    ArrayPool<byte>.Shared.Return(rentedBuffer);
+                    ArrayPool<byte>.Shared.Return(rentedArray);
```
DmitriMelnikov, line 814: You could not return if useStackAlloc == true.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
aavasthy: I just want to do encodingBuffer.Clear() with stackAlloc and not with ArrayPool

kirankumarkolli: > The content of the newly allocated memory is undefined. You should initialize it before the use. For example, you can use the [Span<T>.Clear](https://learn.microsoft.com/en-us/dotnet/api/system.span-1.clear) method that sets all the items to the default value of type T

If usage pattern guarantee that all bytes will be overwrittern then clear() is a overhead, that's not the case here right?

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
```diff
@@ -782,32 +785,33 @@ private static unsafe string OptimizedConvertToBase64string(byte[] hashPayLoad,
             // Create a large enough buffer that URL encode can use it.
             // Increase the buffer by 3x so it can be used for the URL encoding
             int capacity = Base64.GetMaxEncodedToUtf8Length(hashPayLoad.Length) * 3;
-            byte[] rentedBuffer = ArrayPool<byte>.Shared.Rent(capacity);
+            bool useStackAlloc = capacity <= MaxCapacityInBytes;
+
+            byte[] rentedArray = null;
+            Span<byte> encodingBuffer = useStackAlloc ? stackalloc byte[capacity] : (rentedArray = ArrayPool<byte>.Shared.Rent(capacity));
```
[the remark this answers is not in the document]
aavasthy: But in this change we use rentedArray only when we are implementing ArrayPool and we don't rent array when using stackAlloc. ArrayPool is only being used for cases when calculated capacity > MaxCapacity otherwise use stackAlloc.

Microsoft.Azure.Cosmos/src/Authorization/AuthorizationHelper.cs
aavasthy: yes, encodingBuffer.Clear() is used here when the buffer is created using stackalloc, not when it's obtained from ArrayPool. In this case, calling Clear() ensures that any arbitrary values from the stack are cleared and sets all elements of the Span to their default value.

aavasthy:
Closing the issue for now as there is lack of clarity around the ask and still waiting on requesting team.

PR #3830, opened 2023-04-28
[Internal] Query : Adds Tests for Linq Custom Serializer

adityasa:
# [Internal] Query : Adds Tests for Linq Custom Serializer

## Description

Bootstrapping some tests for LINQ custom serializer. Building on something Minh added before but was undone since we had to revert that change as a regression. Also baselining current behavior around newtonsoft v/s json.net attributes as far as custom serializer is concerned.

## Type of change


sboshra (approved):
:shipit:

kirankumarkolli:

Issue #3635, opened 2023-01-09
[Cosmos Service Upgrade - .NET] Replica Validation - Update Diagnostics to Add Replica Health States


kundadebdatta:
- Update Diagnostics to Add Replica Health States. This requires changes in both `v3` and `Cosmos.Direct` packages.

philipthomas-MSFT:
Waiting for Direct package merge. PR is out.

philipthomas-MSFT:
Complete.

PR #3835, opened 2023-05-01
[Internal] Upgrade Resiliency: Adds Replica Health State Diagnostics.

### LINKED ISSUE #3773: Rntbd Connection - Health Check Improvements
This task proposes adding more aggressive timeout check in the Rntbd connection health check.

The current flow:

- Per connection, if there are 3 consecutive request timeout within 65s, then connection will be closed.

Improvements proposed:

- If the last request time is 65 seconds ago or any threshold configured, close on first timeout
- On primary replicas, close connections more aggressively (first timeout + 5.5s/6s/configured threshold) or Math.max (3 times request timeout, 6 seconds)
- These aggressive closure should be a opt-in behavior for now

kundadebdatta:
# Pull Request Template

## Description

The scope of this PR is given below:

- **Code changes to read replica health states from gateway address cache:** Today, the existing approach reads the replica health states from the returned address from address resolver and snapshots them into a new list for the purpose of diagnostics. This approach has severe drawbacks as it's been done on a hot-path. Therefore, to fix this behavior, this PR refactors the GatewayAddressCache to return the replica health states along with the list of replicas.

- **Adds the replica health statuses into the cosmos sdk diagnostics JSON:** The updated diagnostics will look similar to the below.

  ```
    "RequestCharge": 1,
    "RetryAfterInMs": null,
    "BELatencyInMs": "0.229",
    "ReplicaHealthStatuses": [
        "(port: 14387 | status: Connected | lkt: 06/06/2023 00:52:01)",
        "(port: 14007 | status: Connected | lkt: 06/06/2023 00:52:01)",
        "(port: 14017 | status: Unknown | lkt: 06/06/2023 00:52:01)",
        "(port: 14026 | status: Unhealthy | lkt: 06/06/2023 00:52:01)"
    ],
    "transportRequestTimeline": {
        "requestTimeline": [
            {
                "event": "Created",
                "startTimeUtc": "2023-06-06T00:52:05.7164361Z",
                "durationInMs": 0.0039
            }
      ]
  }
  ```

- **`Cosmos.Direct` Connection health check improvements:** Refactors the `Connection` health check validation logic to add aggressive connection timeout detection. This is done by increasing the `Cosmos.Direct` version to `3.31.1`.

## Type of change

## Closing issues

To automatically close an issue: closes #3773 #3635


philipthomas-MSFT (approved):
LGTM

philipthomas-MSFT (dismissed):
approving again but talked with Deb offline on the change to 5 seconds.

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.Cosmos
     /// </summary>
     internal sealed class ConnectionPolicy
     {
-        private const int defaultRequestTimeout = 10;
+        private const int defaultRequestTimeout = 5;
```
FabianMeiswinkel, line 22: Can you explain the motivation for this in the PR description? Backend request timeout is 5 seconds - using client-side request timeout of 5 seconds as well by default means we can fail even when backend process request successfully. For queries, stored proc executions and even point operations with really large documents that seems risky?

xinlian12: i think this maps to the transient timeout we used in java SDK which is 5s by default

xinlian12 (dismissed):
LGTM, thanks

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
FabianMeiswinkel: But we even know of cases where cx had to increase it when backend took regularly close to 5s - why move into the same problem in .Net - especially because this can be considered a breaking change?

xinlian12: i think it is based on during upgrades, we observed that in .Net the maxLatency can go up to 10s because of the request timeout config @kundadebdatta please correct me here for the reasoning. But if the 10s is particularly chosen due to the above scenarios, then yes we should still keep as 10s.

FabianMeiswinkel: What about using 6 seconds (or maybe 5.5 seocnds) - still some buffer when backend takes 5 seconds - but reduces the tail latency?

FabianMeiswinkel: And please call this change out for Gateway teams - especially on Mongo (where they changes response size to up to 16 MB they might need to tune this)

kundadebdatta: Sounds like a good proposal. I agree to push it to '6` seconds, by keeping a second aside as a buffer.

FabianMeiswinkel (approved):
LGTM when changing the request timeout to 6 seconds

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
kirankumarkolli: https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB?path=/Product/Cosmos/Compute/Host/TenantClientResolver.cs&version=GBmaster&line=509&lineEnd=510&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

Its explicitly configured and default is not used.

kirankumarkolli (changes requested):
Changes LGTM
Let's please split it into multiple PRs

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
kundadebdatta: Broken this PR into 2 parts. The second part can be [found here](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3902).

FabianMeiswinkel (approved):
LGTM

xinlian12 (approved):
LGTM, thanks

kirankumarkolli (approved):

PR #3836, opened 2023-05-03
[Preview] Integrated cache: Adds BypassIntegratedCache to DedicatedGatewayRequestOptions

jiajunpeng-msft:
# Pull Request Template

## Description

This is to add a new DedicatedGatewayRequestOption "BypassIntegratedCache".

Currently, integrated cache is used by default for Dedicated Gateway. Customers cannot skip cache for particular requests or data unless they shift to multi-tenant Gateway,which will lose the benefits of Dedicated Gateway.

For customers to have more control over integrated cache, we're introducing a new "RequestOption" called "BypassIntegratedCache". This option will allow the customer to decide whether to use integrated cache for each request or not. If this value is set to true, the item/query will be served from backend and won't be cached in Dedicated Gateway.

## Type of change

## Closing issues


ealsur:
@msftbot wait for @kirankumarkolli to approve, this PR modifies public API surface

Microsoft.Azure.Cosmos/src/Resource/Settings/DedicatedGatewayRequestOptions.cs
```diff
@@ -23,6 +23,12 @@ public class DedicatedGatewayRequestOptions
         /// </remarks> 
         public TimeSpan? MaxIntegratedCacheStaleness { get; set; }
 
+        /// <summary>
+        /// Gets or sets if bypass the integrated cache or not associated with the request in the Azure CosmosDB service.
+        /// </summary>
+        /// <value>Default value is false.</value>
+        public bool? BypassIntegratedCache { get; set; }
```
ealsur, line 30: If this is for Public Preview, it should not be public, it should be conditionally public. For example: 
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/17203a5966e906e0d66d52a95cf4d482add3c51e/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs#L1018-L1023

Or is this already GAed?

Is this already planned to be added to other SDKs?

ealsur (changes requested):
Waiting on confirmation wheter this is preview or GA

Microsoft.Azure.Cosmos/src/Resource/Settings/DedicatedGatewayRequestOptions.cs
ealsur: Which SDK is it on is orthogonal. Is this feature GA? Or are you introducing it for users to test it (public preview)?

jiajunpeng-msft: After discussed with my manager, we decided to push this under public preview. Will update the PR soon. Previously we were thinking to directly GA this feature as this is a small change but feels like under public preview is safer.

Microsoft.Azure.Cosmos/src/Resource/Settings/DedicatedGatewayRequestOptions.cs
```diff
@@ -36,6 +47,11 @@ internal static void PopulateMaxIntegratedCacheStalenessOption(DedicatedGatewayR
 
                 request.Headers.Set(HttpConstants.HttpHeaders.DedicatedGatewayPerRequestCacheStaleness, cacheStalenessInMilliseconds.ToString(CultureInfo.InvariantCulture));
             }
+
+            if (dedicatedGatewayRequestOptions?.BypassIntegratedCache != null)
```
ealsur, line 51: This is a `bool?` the correct way would be to use HasValue to check if it has been set, and Value to extract the value and compare if it's true.

ealsur, line 51: Please add tests to confirm the header is set only when the property is set to True, and not when it's not set or set to False

Microsoft.Azure.Cosmos/src/Resource/Settings/DedicatedGatewayRequestOptions.cs
```diff
@@ -23,6 +23,17 @@ public class DedicatedGatewayRequestOptions
         /// </remarks> 
         public TimeSpan? MaxIntegratedCacheStaleness { get; set; }
 
+        /// <summary>
+        /// Gets or sets if bypass the integrated cache or not associated with the request in the Azure CosmosDB service.
```
kirankumarkolli, line 27: Please expand with more context on what it does if possible, with an example. 
At-least we need to cover that it doesn't server from cache and also the response is also not cached.

Microsoft.Azure.Cosmos/src/Resource/Settings/DedicatedGatewayRequestOptions.cs
jiajunpeng-msft: Updated.

Microsoft.Azure.Cosmos/src/Resource/Settings/DedicatedGatewayRequestOptions.cs
jiajunpeng-msft: Updated.

Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Tests/Contracts/DotNetSDKEncryptionCustomAPI.json
```diff
@@ -51,6 +51,13 @@
           ],
           "MethodInfo": "System.Threading.Tasks.Task`1[Microsoft.Azure.Cosmos.Encryption.Custom.DataEncryptionKey] FetchDataEncryptionKeyAsync(System.String, System.String, System.Threading.CancellationToken);IsAbstract:False;IsStatic:False;IsVirtual:True;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
         },
+        "System.Threading.Tasks.Task`1[Microsoft.Azure.Cosmos.Encryption.Custom.DataEncryptionKey] FetchDataEncryptionKeyWithoutRawKeyAsync(System.String, System.String, System.Threading.CancellationToken)[System.Runtime.CompilerServices.AsyncStateMachineAttribute(typeof(Microsoft.Azure.Cosmos.Encryption.Custom.CosmosDataEncryptionKeyProvider+<FetchDataEncryptionKeyWithoutRawKeyAsync>))]": {
+          "Type": "Method",
+          "Attributes": [
```
sourabh1007, line 56: Does this PR includes change n encryption project also? Please revert this file if not.

jiajunpeng-msft: Reverted.

ealsur (approved):

kirankumarkolli (approved):

Issue #3582, opened 2022-11-24
NuGet LicenseFile Info missing


tinohager:
The `Microsoft.Azure.Cosmos` NuGet Package miss the new `<PackageLicenseFile>LICENSE.txt</PackageLicenseFile>` in the project file.

```
<PropertyGroup>
  <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
</PropertyGroup>

<ItemGroup>
  <None Include="..\LICENSE.txt" Pack="true" PackagePath="LICENSE.txt" />
</ItemGroup>
```

Therefore, the license used is not displayed in NuGet, [Microsoft.Azure.Cosmos](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj)
![image](https://user-images.githubusercontent.com/8215186/203745061-f6e3181e-38a5-4394-a268-10cc18f4ad69.png)

Instead of [Microsoft.Azure.Amqp](https://github.com/Azure/azure-amqp/blob/master/src/Microsoft.Azure.Amqp.csproj)
![image](https://user-images.githubusercontent.com/8215186/203745555-a8ab338f-2223-4133-b45e-bd5ba7904e66.png)

ealsur:
@philipthomas-MSFT Could you take a look at this report and see if this property is required?

Issue #3667, opened 2023-01-25
3rd party document.


AASPLLC:
This is actually just a simple question. Do we have to keep the 3rd party document (ThirdPartyNotice.txt) in every cosmos project?
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/ThirdPartyNotice.txt

bartelink:
i.e. is the copying of that file as Content in the nuget package intentional, or should it be changed to `None` in order to stop it leaking into all projects referencing it?

<img width="172" alt="image" src="https://user-images.githubusercontent.com/206668/214793217-66904563-8764-4eda-b244-fa5f6a67dbeb.png">

ealsur:
@philipthomas-MSFT Could you take a look to see what is the purpose of the file and if we are bundling it correctly?

walliski:
Its a major annoyance that it shows up in all Projects that references the Cosmos NuGet. I haven't found a good way to remove it for all developers either, as the Path for it is the users' NuGet cache in their home folder.

Pilchie:
I think the issue here is that this file shouldn't be in the "content" or "contentFiles" folders of the package.  Instead it should just be a loose file in the root of the nupkg, similar to Icon.png.

bartelink:
This is where it's pulled in: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/1a8b7f0d0bad48d7517f8871cb5174a4f929118f/Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj#LL61C1-L63C14

PR #3847, opened 2023-05-12
PackageLicense: Removes PackageLicenseUrl and Adds PackageLicenseFile since PackageLicenseUrl is deprecated

philipthomas-MSFT:
## Description

This originated from an [3582](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3582) issue where the license was using an Url and not showing up in _Nuget.org_ in the details of the package. **PackageLicenseUrl** was deprecated and it was recommended to use either **PackageLicenseFile** or **PackageLicenseExpression** for immutability reasons. I am providing link to the original issue and [discussion](https://github.com/NuGet/Home/issues/4628). 

There was an option to leave LICENSE without extension or rename with .TXT extension. I opted to use the original without the .TXT as I do not see either a PRO or a CON to one over the other, but I have no qualms with either option.

In regard to not using **PackageLicenseExpression**. I am not really seeing a PRO or a CON on the surface as to using this or not for this particular repo.

## Type of change

.NET project file to include **PackageLicenseFile** and exclude **PackageLicenseUrl**.

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
+++ b/Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
@@ -24,7 +24,6 @@
 		<PackageId>Microsoft.Azure.Cosmos</PackageId>
 		<PackageTags>microsoft;azure;cosmos;cosmosdb;documentdb;docdb;nosql;azureofficial;dotnetcore;netcore;netstandard</PackageTags>
 		<PackageReleaseNotes>The change log for this SDK is made available at https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/changelog.md at the time of release.</PackageReleaseNotes>
-		<PackageLicenseUrl>https://aka.ms/netcoregaeula</PackageLicenseUrl>
 		<PackageProjectUrl>https://github.com/Azure/azure-cosmos-dotnet-v3</PackageProjectUrl>
 		<PublishRepositoryUrl Condition=" '$(ProjectRef)' != 'True' ">true</PublishRepositoryUrl>
 		<PackageIconUrl>http://go.microsoft.com/fwlink/?LinkID=288890</PackageIconUrl>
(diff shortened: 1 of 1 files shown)


NaluTripician:
LGTM

ealsur (approved):
LGTM. Can you run dotnet pack on the project and inspect the package with Nuget Package Explorer to verify the license is inside?

philipthomas-MSFT:
> LGTM. Can you run dotnet pack on the project and inspect the package with Nuget Package Explorer to verify the license is inside?

Yes, I already did this and verified. Using all options in the description.

ealsur:
@msftbot wait for @kirankumarkolli to approve

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
```diff
@@ -43,8 +42,13 @@
 		<NoWarn>NU5125</NoWarn>
 		<Optimize Condition="'$(Configuration)'=='Release'">true</Optimize>
 		<LangVersion>$(LangVersion)</LangVersion>
+		<PackageLicenseFile>LICENSE</PackageLicenseFile>
 	</PropertyGroup>
 
+	<ItemGroup>
+		<None Include="..\..\LICENSE" Pack="true" PackagePath=""/>
```
kirankumarkolli, line 49: Is it expected to be copied into output folder?
Can you please check what .net core packages do and replicate the same?

/cc: @Pilchie

Pilchie: Being precise, this *doesn't* copy it to the output folder of the build, but does include it in the NuGet package.  That seems to match .NET does for licenses that aren't supported by `PackageLicenseExpression` (https://github.com/dotnet/arcade/blob/1aff4eb33aa7cbf26ccd9fc43c17cb609a14dad4/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets#L33C2-L43).

Note: You could consider adding the `Visible='False'` metadata if you don't want the license file to show up in Solution Explorer.

philipthomas-MSFT: Updated to include `Visible="false"`. @Pilchie is there a reason why we want this invisible, but other files like ThirdPartyNotice.txt are visible? 

`<Content Include="..\..\ThirdPartyNotice.txt" Link="ThirdPartyNotice.txt" />`

This may play into the other issue I have that I will be working on next. #3667 

Just FYI, here is the package that I `dotnet pack` it locally.

image

sourabh1007 (approved):

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
Pilchie: > is there a reason why we want this invisible

I think it's just a preference thing. I have no problem keeping it visible, though relative paths can end up being displayed strangely is all.

kirankumarkolli (approved):

PR #3857, opened 2023-05-18
Support multiple formats of Azure region names for CosmosClientOptions

### LINKED ISSUE #2330: ApplicationRegion parameter does not work with output from ARM

pradeep-chellappan:
## Description

This is a proposed fix for - https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2330
We want to set CosmosClientOptions.ApplicationRegion and CosmosClientOptions.ApplicationPreferredRegions in the format that ARM exposes it. For e.g. if the region is West US 2, the CosmosClientOptions should accept both "West US 2" and "WestUs2" and "westus2". This gives maximum flexibility to the users of the SDK.
To enable this, I keep a mapping between the normalized region name and the one that CosmosDB SDK uses internally. When CosmosClientOptions.ApplicationRegion or CosmosClientOptions.ApplicationPreferredRegions is set, the region names are converted, if required.

## Type of change

## Closing issues

To automatically close an issue: closes #2330

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -72,6 +72,8 @@ namespace Microsoft.Azure.Cosmos
         private IWebProxy webProxy;
         private Func<HttpClient> httpClientFactory;
         private string applicationName;
+        private string applicationRegion;
+        private IReadOnlyList<string> applicationPreferredRegions;
 
         /// <summary>
         /// Creates a new CosmosClientOptions
@@ -154,7 +156,11 @@ namespace Microsoft.Azure.Cosmos
         /// </example>
         /// <seealso cref="CosmosClientBuilder.WithApplicationRegion(string)"/>
         /// <seealso href="https://docs.microsoft.com/azure/cosmos-db/high-availability#high-availability-with-cosmos-db-in-the-event-of-regional-outages">High availability on regional outages</seealso>
-        public string ApplicationRegion { get; set; }
+        public string ApplicationRegion
+        {
+            get => this.applicationRegion;
+            set => this.applicationRegion = RegionNameMapping.GetCosmosDBRegionName(value);
+        }
 
         /// <summary>
         /// Gets and sets the preferred regions for geo-replicated database accounts in the Azure Cosmos DB service. 
@@ -189,7 +195,11 @@ namespace Microsoft.Azure.Cosmos
         /// </code>
         /// </example>
         /// <seealso href="https://docs.microsoft.com/azure/cosmos-db/high-availability#high-availability-with-cosmos-db-in-the-event-of-regional-outages">High availability on regional outages</seealso>
-        public IReadOnlyList<string> ApplicationPreferredRegions { get; set; }
+        public IReadOnlyList<string> ApplicationPreferredRegions
+        {
+            get => this.applicationPreferredRegions;
+            set => this.applicationPreferredRegions = value?.Select(RegionNameMapping.GetCosmosDBRegionName).ToList();
+        }
 
         /// <summary>
         /// Get or set the maximum number of concurrent connections allowed for the target
--- a/Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
+++ b/Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
@@ -1,0 +1,104 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Maps a normalized region name to the format that CosmosDB is expecting (for e.g. from 'westus2' to 'West US 2')
+    /// </summary>
+    internal class RegionNameMapping
+    {
+        private static readonly IReadOnlyDictionary<string, string> normalizedToCosmosDBRegionNameMapping =
+            new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            {
+                {"westus", Regions.WestUS},
+                {"westus2", Regions.WestUS2},
+                {"westcentralus", Regions.WestCentralUS},
+                {"eastus", Regions.EastUS},
+                {"eastus2", Regions.EastUS2},
+                {"centralus", Regions.CentralUS},
+                {"southcentralus", Regions.SouthCentralUS},
+                {"northcentralus", Regions.NorthCentralUS},
+                {"westeurope", Regions.WestEurope},
+                {"northeurope", Regions.NorthEurope},
+                {"eastasia", Regions.EastAsia},
+                {"southeastasia", Regions.SoutheastAsia},
+                {"japaneast", Regions.JapanEast},
+                {"japanwest", Regions.JapanWest},
+                {"australiaeast", Regions.AustraliaEast},
+                {"australiasoutheast", Regions.AustraliaSoutheast},
+                {"centralindia", Regions.CentralIndia},
+                {"southindia", Regions.SouthIndia},
+                {"westindia", Regions.WestIndia},
+                {"canadaeast", Regions.CanadaEast},
+                {"canadacentral", Regions.CanadaCentral},
+                {"germanycentral", Regions.GermanyCentral},
+                {"germanynortheast", Regions.GermanyNortheast},
+                {"chinanorth", Regions.ChinaNorth},
+                {"chinaeast", Regions.ChinaEast},
+                {"chinanorth2", Regions.ChinaNorth2},
+                {"chinaeast2", Regions.ChinaEast2},
+                {"koreasouth", Regions.KoreaSouth},
+                {"koreacentral", Regions.KoreaCentral},
+                {"ukwest", Regions.UKWest},
+                {"uksouth", Regions.UKSouth},
+                {"brazilsouth", Regions.BrazilSouth},
+                {"usgovarizona", Regions.USGovArizona},
+                {"usgovtexas", Regions.USGovTexas},
+                {"usgovvirginia", Regions.USGovVirginia},
+                {"eastus2euap", Regions.EastUS2EUAP},
+                {"centraluseuap", Regions.CentralUSEUAP},
+                {"francecentral", Regions.FranceCentral},
+                {"francesouth", Regions.FranceSouth},
+                {"usdodcentral", Regions.USDoDCentral},
+                {"usdodeast", Regions.USDoDEast},
+                {"australiacentral", Regions.AustraliaCentral},
+                {"australiacentral2", Regions.AustraliaCentral2},
+                {"southafricanorth", Regions.SouthAfricaNorth},
+                {"southafricawest", Regions.SouthAfricaWest},
+                {"uaecentral", Regions.UAECentral},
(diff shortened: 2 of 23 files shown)


pradeep-chellappan:
@microsoft-github-policy-service agree company="DocuSign"

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// <summary>
+    /// Maps a normalized region name to the format that CosmosDB is expecting (for e.g. from 'westus2' to 'West US 2')
+    /// </summary>
+    internal class RegionNameMapping
+    {
+        private static readonly IReadOnlyDictionary<string, string> normalizedToCosmosDBRegionNameMapping =
+            new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            {
+                {"westus", Regions.WestUS},
```
kirankumarkolli, line 18: Recommended to get this mapping dynamically from the RegionUtility map. 
With-out it for new regions multiple places needs to be touched (high possibility is mis)

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
```diff
@@ -0,0 +1,104 @@
...
+                {"polandcentral", Regions.PolandCentral},
+            };
+
+        /// <summary>
+        /// Given a normalized region name, this function retrieves the region name in the format that CosmosDB expects.
+        /// If the region is not known, the same value as input is returned.
+        /// </summary>
+        /// <param name="normalizedRegionName">An Azure region name in a normalized format. The input is not case sensitive.</param>
+        /// <returns>A string that contains the region name in the format that CosmosDB expects.</returns>
+        public static string GetCosmosDBRegionName(string normalizedRegionName)
```
kirankumarkolli, line 94: Our proposal is to not have it as public contract but cover it as implementation when a normalized region name it set.

kirankumarkolli (changes requested):
Please check comments

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
pradeep-chellappan: I am not sure if I understood this comment. Did you mean you don't want this as a public method in the SDK? The class itself is internal. @kirankumarkolli

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
pradeep-chellappan: I was trying to avoid any reflection, but I'll send a change which avoid the need to update the list of regions in multiple places.

pradeep-chellappan: You meant RegionProximityUtil, right?

pradeep-chellappan: Made changes to avoid having a duplicate map.

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
```diff
@@ -0,0 +1,45 @@
...
+        private static readonly IDictionary<string, string> normalizedToCosmosDBRegionNameMapping;
+
+        static RegionNameMapping()
+        {
+            FieldInfo[] fields = typeof(Regions).GetFields(BindingFlags.Public | BindingFlags.Static);
+            normalizedToCosmosDBRegionNameMapping = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+            foreach (FieldInfo field in fields)
+            {
+                normalizedToCosmosDBRegionNameMapping[field.Name.ToLowerInvariant()] = field.GetValue(null).ToString();
```
sourabh1007, line 25: how did we make sure that it is applicable for all the regions? Is other SDKs are following this logic? or it is just observation and relying on adding condition if we found/add any new region which is not following this pattern?

pradeep-chellappan: It is applicable to all regions where CosmosDB currently runs (based on the Regions class).

kirankumarkolli: @sourabh1007 its possible that some other azure RP's returning/expecting in different form. 
One example that was mentioned with "-" by @FabianMeiswinkel 

We will learn more as we uncover and fill the gaps forward

kirankumarkolli:
@pradeep-chellappan can you please take a look at comments?

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
```diff
@@ -0,0 +1,45 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+
+    /// <summary>
+    /// Maps a normalized region name to the format that CosmosDB is expecting (for e.g. from 'westus2' to 'West US 2')
+    /// </summary>
+    internal class RegionNameMapping
+    {
+        private static readonly IDictionary<string, string> normalizedToCosmosDBRegionNameMapping;
```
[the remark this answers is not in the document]
pradeep-chellappan: Sorry, but what is the caching that you are referring to?

kirankumarkolli:
@pradeep-chellappan Can you please look into the comments. 

We want to ship this feature ASAP. We will wait for your update till 07/14 and after that we will move forward with a new PR replicating the changes.

pradeep-chellappan:
> @pradeep-chellappan Can you please look into the comments.
> 
> We want to ship this feature ASAP. We will wait for your update till 07/14 and after that we will move forward with a new PR replicating the changes.

Hey Kiran, could you please reply to my prior comment regarding the caching?

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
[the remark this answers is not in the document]
pradeep-chellappan: @kirankumarkolli - I did not understand your comment. Could you please clarify?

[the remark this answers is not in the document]
kirankumarkolli: Static dict<> lifetime is process lifetime. Where as the usage of the dict<> is just one method call at the ClientInitialization time only. 

So static instance/cache is unnecessary, it can be created once as local state.

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        static RegionNameMapping()
+        {
+            FieldInfo[] fields = typeof(Regions).GetFields(BindingFlags.Public | BindingFlags.Static);
```
kirankumarkolli, line 20: @ealsur how about include all (non-public) also? 

Its more about future proofing where the region might be added as internal before going public. And string/text based API's will still be functional right?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -154,7 +156,11 @@ public string ApplicationName
         /// </example>
         /// <seealso cref="CosmosClientBuilder.WithApplicationRegion(string)"/>
         /// <seealso href="https://docs.microsoft.com/azure/cosmos-db/high-availability#high-availability-with-cosmos-db-in-the-event-of-regional-outages">High availability on regional outages</seealso>
-        public string ApplicationRegion { get; set; }
+        public string ApplicationRegion
+        {
+            get => this.applicationRegion;
+            set => this.applicationRegion = RegionNameMapping.GetCosmosDBRegionName(value);
```
kirankumarkolli, line 162: Application input is updated right on setter. Functionally works. 
One other alternative is to keep original values as Application configured and then later during CosmosClient initialization do conversation. 

@ealsur any preferences?

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
kirankumarkolli, line 25: Coverage for spaces" Ref: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2330#issuecomment-1570900320

I.e. if the incoming name doesn't have spaces also it should match.

kirankumarkolli:
> > @pradeep-chellappan Can you please look into the comments.
> > We want to ship this feature ASAP. We will wait for your update till 07/14 and after that we will move forward with a new PR replicating the changes.
> 
> Hey Kiran, could you please reply to my prior comment regarding the caching?

Thanks you @pradeep-chellappan, please check updated comments.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
FabianMeiswinkel: I would prefer also having the original UserAgent input in the diagnostics - so, both the normalized and original values. Keeping the original value in ApplicationRegion and then client on initialization normalizing it  is porbably the easiest way to get above.

Microsoft.Azure.Cosmos/src/RegionNameMapping.cs
ealsur: We don't have cases where "Regions" contains internal names. We have cases where "LocationNames" (which is internal itself) contains regions that are not ready though. When they are not ready, there is no mapping in RegionProximityUtil either. 
So adding also Internals here would not enable any scenario.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: No preference. On the Diagnostics we log the value of the public getter:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/d7fc28225057ac402a7c862daf7707d08a2f70d3/Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientConfigurationTraceDatum.cs#L28-L29

The only thing with this approach is that the getter does not return the same value that was set, so if anyone is validating that if I set West US, I get West US, the test will fail.

Making the conversion internally on:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/d7fc28225057ac402a7c862daf7707d08a2f70d3/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs#L782-L790

Would achieve the same thing while keeping the getter value matching the user input.

ealsur:
Closed through https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4016

PR #3862, opened 2023-05-18
LINQ: improve support for System.Text.Json JsonConverter and JsonPropertyName

### LINKED ISSUE #3250: LINQ: Add support for System.Text.Json attributes
If you follow this sample and use the System.Text.Json as the default serializer LINQ will not translate the attribute so the incorrect query will be generated. This will cause the query to return no or incorrect results.

Follow the sample listed here: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/SystemTextJson/CosmosSystemTextJsonSerializer.cs

Update the class like the sample:
```c#
[JsonPropertyName("poolId")] // after change
//[JsonProperty("poolId")] // before change
public string PoolId { get; set; }
```

Try to do a LINQ query:
```c#
 container.GetItemLinqQueryable<T>().Where(x => x.PoolId != null).ToFeedIterator();
```

The query will execute but it will return no results because it will be translated as 'PoolId' instead of the expected 'poolId'.

It seems the following code needs to be updated to handle the new System.Text.Json types:
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/cee34ce3cf41dabb2a073f64a289edb2e922177a/Microsoft.Azure.Cosmos/src/Linq/TypeSystem.cs#L30

ccurrens:
# Pull Request Template

## Description

One of the CosmosDB samples shows how to add [System.Text.Json as the cosmos serializer](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/2b7d709b9fab04a09f373d4d64e2a1ff7c16f849/Microsoft.Azure.Cosmos.Samples/Usage/SystemTextJson/CosmosSystemTextJsonSerializer.cs). This works well to read and write documents to storage, but fails short when it comes to querying. There are two scenarios that work with `Newtonsoft.Json` that don't work today with `System.Text.Json`:

### Using customized property names in LINQ queries.

These properties work for serialization because `System.Text.Json` can be specified as the custom serializer. It does not work in LINQ queries, because the `TypeSystem` has no knowledge of `JsonPropertyNameAttribute`. This MR adds support for that attribute in `TypeSystem`.

For backwards compatibility, `Newtonsoft.Json.JsonProperty` takes precedence, then `DataMember`, and if still `null`, then `JsonPropertyNameAttribute` is used if present.

### Custom converters on properties (e.g., `StringEnumConverter`)

This is currently supported for `Newtonsoft.Json` converters. This MR adds support for `System.Text.Json.JsonConverter` types. Similarly, the `JsonConverters` for `Newtonsoft.Json` will take precedence over STJ types.

## Type of change

## Closing issues

closes #3250


Arcalise08:
Why isnt this being pulled in. These changes seem necessary for proper System.Text.Json support.

ilmax:
Hey @kirankumarkolli, what's holding this one? I'm not sure why it takes forever to fix a BUG like this one and no, no one will convince me that this is not a bug.

Arcalise08:
Anyone looking at this PR. They arent going to approve this one. Or even look at it. They have their own devs working on the functionality though. I found the PR below.

https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4138

PR #3866, opened 2023-05-19
CosmosClient: Fixes missing Trace when converting HTTP Timeout to 503

### LINKED ISSUE #3831: Missing trace when converting HTTP Timeout to 503
When we convert HTTP timeouts to a 503 response, we are failing to pass the `trace` parameter.

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/e5972ac0ffd53595b2f1f0f10a3faf913ff882e1/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs#L381-L398

`CosmosExceptionFactory.CreateServiceUnavailableException` has a `trace` parameter that we are not passing.

This generates empty diagnostics:

```
Response status code does not indicate success: ServiceUnavailable (503); Substatus: 20003; ActivityId: xxxxx; Reason: (GatewayStoreClient Request Timeout. Start Time UTC:xx/xx/xxxx xx:xx:xx xx; Total Duration:65013.1998 Ms; Request Timeout 65000 Ms; Http Client Timeout:65000 Ms; Activity id: xxxx;);; 
Diagnostics:{"Summary":{},"name":"NoOp","start time":"00:00:00:000","duration in milliseconds":0}
```

rinatmini:
# Pull Request Template

## Description

Added trace when converting HTTP Timeout to 503 to avoid generating empty diagnostics.
Example of response after passing `trace` in `CosmosExceptionFactory.CreateServiceUnavailableException` (grabbed using unit-tests).
```json
{
    Response status code does not indicate success: ServiceUnavailable (503); Substatus: 20003; ActivityId: 00000000-0000-0000-0000-000000000000; Reason: (GatewayStoreClient Request Timeout. Start Time UTC:5/24/2023 11:55:30 PM; Total Duration:1421.2407 Ms; Request Timeout 65000 Ms; Http Client Timeout:65000 Ms; Activity id: 00000000-0000-0000-0000-000000000000;);; 
    Diagnostics:
        {
            "Summary": {},
            "name": "DocumentClient_BuildHttpClientFactory_WithHandler",
            "start datetime": "2023-05-24T23: 55: 30.256Z",
            "duration in milliseconds": 1742.3758
        }
}
```

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Closing issues

To automatically close an issue: closes #3831


Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -307,20 +308,25 @@ ValueTask<HttpRequestMessage> CreateRequestMessage()
...
         }
 
         private async Task<HttpResponseMessage> SendHttpHelperAsync(
             Func<ValueTask<HttpRequestMessage>> createRequestMessageAsync,
             ResourceType resourceType,
             HttpTimeoutPolicy timeoutPolicy,
             IClientSideRequestStatistics clientSideRequestStatistics,
-            CancellationToken cancellationToken)
+            CancellationToken cancellationToken,
+            ITrace trace)
```
ealsur, line 329: nit: CancellationToken should always be the last parameter.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -307,20 +308,25 @@ ValueTask<HttpRequestMessage> CreateRequestMessage()
                 throw new ArgumentNullException(nameof(createRequestMessageAsync));
             }
 
-            return this.SendHttpHelperAsync(
-                createRequestMessageAsync,
-                resourceType,
-                timeoutPolicy,
-                clientSideRequestStatistics,
-                cancellationToken);
+            using (ITrace trace = Trace.GetRootTrace(nameof(SendHttpAsync), TraceComponent.Transport, TraceLevel.Info))
```
ealsur, line 311: The idea is that whoever calls CosmosHttpClientCore already has a Trace, that is the trace that needs to be passed down here. 

So we need to find who calls this SendHttpAsync method and pass it down.

Interestingly, this method receives already

`IClientSideRequestStatistics clientSideRequestStatistics`, which is part of the Trace, in fact, below this code does:

```
if (clientSideRequestStatistics is ClientSideRequestStatisticsTraceDatum datum)
```

ClientSideRequestStatisticsTraceDatum contains the reference to the trace through its `TraceSummary` property.

Either use this to access the original Trace or receive the Trace along with the `IClientSideRequestStatistics` and make sure all the callers pass it (this last option might be more involved in terms of code changes)

ealsur (changes requested):

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
rinatmini: @ealsur @sourabh1007
Added receiving the Trace along with the `IClientSideRequestStatisticsIn` in the hierarchy of calls. 
Most of the callers don't have `Trace` or don't use it - so in most cases, `Trace/Diagnostic` is still empty or default (`NoOps`). Looks like will be required additional work to add using and populating the `Trace` object during calls to get correct and desired diagnostics.
Also `CosmosExceptionFactory.CreateServiceUnavailableException` expects `ITrace` and `TraceSummary` and cannot be used to get access the original Trace.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6473,8 +6473,10 @@ async ValueTask<HttpRequestMessage> CreateRequestMessage()
                     return request;
                 }
 
-                AccountProperties databaseAccount = await gatewayModel.GetDatabaseAccountAsync(CreateRequestMessage,
-                                                                                               clientSideRequestStatistics: null);
+                AccountProperties databaseAccount = await gatewayModel.GetDatabaseAccountAsync(
+                    CreateRequestMessage,
+                    clientSideRequestStatistics: null,
+                    trace: null);
```
sourabh1007, line 6479: No need to pass null, you can make it optional in the function itself

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -105,6 +107,7 @@ public virtual async Task<DocumentServiceResponse> ProcessMessageAsync(DocumentS
                 ResourceType.DatabaseAccount,
                 HttpTimeoutPolicyControlPlaneRead.Instance,
                 clientSideRequestStatistics,
+                trace,
```
sourabh1007, line 110: Why are you passing trace here? are we going to capture traces from GetDatabaseAccount Call also?

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
ealsur: Have we tested this (that the ClientSideRequestStatistics have no Trace) with the SDK in GatewayMode? We need to remember that there are 2 groups of calls, there are initialization calls that are for Metadata and then there are Data Plane calls.

You mentioned that most don't have `Trace`, which ones do?

When I see in the `TransportHandler` how the ClientSideRequestStatistics is created, it should have a Summary: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Handler/TransportHandler.cs#L96-L97

I wonder, can we check for this and IF there is a Summary, then pass the Trace to the `CreateServiceUnavailableException` call and if not (there is no Summary), then pass a `NoOpTrace.Singleton`.

But first understand which are the cases where there is a Summary and which where there isn't?

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
ealsur: GetDatabaseAccount is part of the initialization, there is no harm in adding the trace there?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -111,15 +114,18 @@ async Task<HttpResponseMessage> sendFunc(HttpRequestMessage request, Cancellatio
                 HttpMessageHandler messageHandler = new MockMessageHandler(sendFunc);
                 using CosmosHttpClient cosmoshttpClient = MockCosmosUtil.CreateCosmosHttpClient(() => new HttpClient(messageHandler));
 
-                HttpResponseMessage responseMessage = await cosmoshttpClient.SendHttpAsync(() =>
+                using (ITrace trace = Trace.GetRootTrace(nameof(RetryTransientIssuesTestAsync)))
```
ealsur, line 117: Where is this trace used?

ealsur (dismissed):

sourabh1007 (approved):

sourabh1007 (approved):

ealsur (approved):

PR #3879, opened 2023-05-30
Subpartitioning: Fixes handling of split physical partitions

### LINKED ISSUE #3865: ChangeFeedIterator gets stuck on multihashed containers for prefix partitions keys that span multiple physical partitions.
**Describe the bug**
When reading the `ChangeFeed` for a prefix partition key that spans multiple physical partitions the `ChangeFeed` will not be able to handle the partition split and will get stuck in an infinite loop.

**Expected behavior**
Results should be returned.

**Actual behavior**
`ReadNextAsync` becomes stuck in an infinite loop.

```csharp
// Create Container Hierarchical PartitionKeys
this.containerProperties = new ContainerProperties("mycoll", new List<string> { "/State", "/City", "/ZipCode" });
this.container = await database.CreateContainerAsync(this.containerProperties);

// ...
//Container is then filled with enough items span multiple physical partitions 
// ...

//Create a Prefix PartitionKey
Cosmos.PartitionKey prefixPK = new PartitionKeyBuilder().Add("Redmound").Build();

//Create ChangeFeedIterator
FeedRange feedRange = FeedRange.FromPartitionKey(shardPartitionKey);
FeedIterator<dynamic> feedIterator = container.GetChangeFeedIterator<dynamic>(ChangeFeedStartFrom.Beginning(shardFeedRange), ChangeFeedMode.Incremental, new ChangeFeedRequestOptions() { PageSizeHint = 10 });

while (feedIterator .HasMoreResults)
{
    FeedResponse<dynamic> response = await feedIterator .ReadNextAsync(); // BUG: Code gets stuck here!
}

```

NaluTripician:
# Pull Request Template

## Description

Bug where SDK would get hung up on ChangeFeedIterator and Query if one prefix partition key spanned multiple physical partitions is now fixed.

## Type of change

## Closing issues

To automatically close an issue: closes #3865


NaluTripician:
@msftbot wait to merge until @kirankumarkolli and @ealsur approve

Microsoft.Azure.Cosmos/src/FeedRange/FeedRanges/FeedRangePartitionKey.cs
```diff
@@ -31,8 +33,13 @@ public FeedRangePartitionKey(PartitionKey partitionKey)
...
                 {
-                    Documents.Routing.Range<string>.GetPointRange(
-                        this.PartitionKey.InternalKey.GetEffectivePartitionKeyString(partitionKeyDefinition))
+                    Documents.Routing.PartitionKeyInternal.GetEffectivePartitionKeyRange(
+                        partitionKeyDefinition,
+                        new Documents.Routing.Range<Documents.Routing.PartitionKeyInternal>(
+                            this.PartitionKey.InternalKey,
+                            this.PartitionKey.InternalKey,
+                            true,
+                            true))
```
ealsur, line 42: Please use named parameters.

ealsur, line 42: Is it expected that the MaxINclusive is true?

ealsur: Also, does this method depend on Direct 3.31? If yes, how would this PR be cherrypicked for a hotfix release? Unless we plan to do a full release?

sourabh1007: and we cannot do a full release until unless we have 0415 deployed everywhere.

sourabh1007: releasing with 0109 was almost impossible because there were few commits required from latest release in order to pass the build. Thats why, we decided to go forward with 0415 release as it in progress and expected to be deployed everywhere very soon. cc @kirankumarkolli

ealsur: Makes sense. In that case though, we effectively blocked ourselves from releasing, including this hotfix, until deployment completes?

ealsur: An alternative would be to not use the API that comes with the Direct package (it's code we can replicate on our side)

sourabh1007: or Direct package commit which is required for this PR, can be cherry-picked to the new hotfix release of 3.30.x. and use this new version of direct package + this PR, to do a minor release of V3.

sourabh1007: no i was talking about making hotfix on older direct branch which is on 202208xx. but then you will run into same problem, in which I ran. 
So, we should wait for 0415 to be deployed everywhere.

NaluTripician: > Is it expected that the MaxINclusive is true?

Yes, if it is false then it will not include the whole range.

Microsoft.Azure.Cosmos/src/FeedRange/FeedRanges/FeedRangePartitionKey.cs
```diff
@@ -31,8 +33,13 @@ public FeedRangePartitionKey(PartitionKey partitionKey)
...
                 {
-                    Documents.Routing.Range<string>.GetPointRange(
-                        this.PartitionKey.InternalKey.GetEffectivePartitionKeyString(partitionKeyDefinition))
+                    Documents.Routing.PartitionKeyInternal.GetEffectivePartitionKeyRange(
+                        partitionKeyDefinition,
+                        new Documents.Routing.Range<Documents.Routing.PartitionKeyInternal>(
+                            min: this.PartitionKey.InternalKey,
+                            max: this.PartitionKey.InternalKey,
+                            isMinInclusive: true,
+                            isMaxInclusive: true))
```
kirankumarkolli, line 42: Is this expected?

kirankumarkolli: Its already answered for Matias, need to spend little more time.

FabianMeiswinkel: @kirankumarkolli - the question above is whether it is expected to use a range with min and max inclusive? I think for a PartitionKeyFeedRange it has always been the case - but it is a good question - in Java we normalize to EPKs (always Min inclusive / max Exclusive) in query, changefeed etc. and wherever we need EPKs. And we also created a helper function to create a NormalizedEPK (with min Inclusive/max Exclusive) even publicly in the Spark Connector. So, agreed - would be good to discuss whether we should just always use normalized EPK.

kirankumarkolli: @ealsur that's my take-away from last Nalu update. 
Where-as EPK -> Physical partition ownership is [Min, Max).

kirankumarkolli (approved):

SrinikhilReddy (approved):

ealsur (approved):

Microsoft.Azure.Cosmos/src/FeedRange/FeedRanges/FeedRangePartitionKey.cs
hejtmii: @kirankumarkolli @ealsur Hi, was the conclusion about this that it is OK, or did it somehow slip through the review? 

We are experiencing issue with the same symptoms that are discussed in this PR in the latest version of the SDK (3.38.1) and in https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3934 (SDK stuck in an infinite loop of updating `pkranges` after partition split) and I can see in the current code of the SDK that both are still inclusive.

I can also see the same thing in `CosmosQueryClientCore.GetCachedContainerQueryPropertiesAsync`

image

image

We are getting 410 Gone from the `pkranges` requests after partition split happened on our DB which puts the SDK into an infinite loop.

So I am wondering if it could be caused by this

We will create a new issue for it, just collecting any possibly related facts... please let me know if this is a bad clue or a good one

PR #3882, opened 2023-06-01
[Internal] Query: Adds OptimisticDirectExecute and RequiresDistribution headers

akotalwar:
# Pull Request Template

## Description

This PR introduces two new boolean headers for the client. One is the OptimisticDirectExecute header, which is on the request message. It indicates to the backend whether the query can use the OptimisticDirectExecute pipeline or not. The second header is the RequiresDistribution header and is on the response message. It lets the client know whether the query requires distribution.

We use the OptimisticDirectExecute pipeline to return results when either of these conditions is true: the query does not require distribution, or the query requires distribution but there is no continuation token. Otherwise, we switch to the specialized pipeline to handle the remaining results if the query requires distribution and there is a continuation token.

## Type of change

## Closing issues


adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -55,22 +63,41 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+                    this.continuationToken = this.Current.Succeeded ? this.Current.Result.State?.Value : null;
+                    if (this.continuationToken != null)
+                    {
+                        this.Current.Result.AdditionalHeaders.TryGetValue(HttpConstants.HttpHeaders.RequiresDistribution, out string requiresDistributionHeaderValue);
+
+                        bool requiresDistribution = true;
+                        if (requiresDistributionHeaderValue != null)
+                        {
+                            requiresDistribution = bool.Parse(requiresDistributionHeaderValue);
+                        }
```
neildsh, line 80: `
bool requiresDistribution;
if(this.Current.Result.AdditionalHeaders.TryGetValue(HttpConstants.HttpHeaders.RequiresDistribution, out string requiresDistributionHeaderValue)
{
    requiresDistribution = bool.Parse(requiresDistributionHeaderValue);
}
else
{
    requiresDistribution = true;
}
`

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -55,22 +63,41 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
+
+                        if (this.previousRequiresDistribution.HasValue && this.previousRequiresDistribution != requiresDistribution)
+                        {
+                            // We should never enter this if statement as requiresDistribution flag can never switch mid execution.
+                            // Hence, this exception should never be thrown.
```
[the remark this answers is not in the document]
akotalwar: Samer and Aditya S had asked me to add a comment here explaining why we would never enter this if statement. What should I change in this comment to make it have more value?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -55,22 +63,41 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace)
...
             {
-                this.inner = await this.queryPipelineStageFactory(this.TryUnwrapContinuationToken());
-                this.executionState = ExecutionState.SpecializedDocumentQueryExecution;
-                if (this.inner.Failed)
+                bool isPartitionSplitException = hasNext.Succeeded && this.Current.Failed && this.Current.InnerMostException.IsPartitionSplitException();
+                if (success && !isPartitionSplitException)
                 {
-                    return false;
+                    this.continuationToken = this.Current.Succeeded ? this.Current.Result.State?.Value : null;
+                    if (this.continuationToken != null)
```
neildsh, line 72: >(this.continuationToken != null) [](http://example.com/codeflow?start=23&length=32)

Why are we checking requiresDistribution only if there is a continuation? This implies that some queries can stop working if data volume increases.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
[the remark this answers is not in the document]
neildsh: fwiw, I think the exception message is clear enough...

neildsh:
>         private const string optimisticDirectExecutionToken = "OptimisticDirectExecutionToken";

Coding convention: constants should start with an uppercase letter: `optimisticDirectExecutionToken`

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs:32 in b88eecb. [](commit_id = b88eecbdfc28a2b52cdcadc450824f6c6cb40520, deletion_comment = False)

neildsh:
>         private class OptimisticDirectExecutionQueryPipelineImpl : IQueryPipelineStage

`sealed class` please

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs:164 in b88eecb. [](commit_id = b88eecbdfc28a2b52cdcadc450824f6c6cb40520, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
```diff
@@ -94,19 +121,33 @@ private CosmosElement TryUnwrapContinuationToken()
             return null;
         }
 
+        private async Task<bool> SwitchToFallbackPipelineAsync(bool useContinuation, ITrace trace)
```
neildsh, line 124: >bool useContinuation [](http://example.com/codeflow?start=63&length=20)

It would be better to take the continuationToken as an argument here.

neildsh:
>         private CosmosElement TryUnwrapContinuationToken()

Consider making this method `static` and using an early return:
`
if (continuationToken == null) return null;

<body of method>
`

It makes the code easier to read.

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs:111 in b88eecb. [](commit_id = b88eecbdfc28a2b52cdcadc450824f6c6cb40520, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
akotalwar: We can return the entire result set to the user without any further action from the pipeline or backend if there is no continuation token, regardless of the value of requiresDistribution. Only if there is a continuation token and requiresDistribution is true will we have to fallback to a different pipeline. 

For queries with large data volumes that are more likely to have a continuation token, we can fall back on the other pipelines that can handle them and deliver the correct results to the user.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
akotalwar: I think we should keep the current format. It clarifies when we need a continuation token and when we dont, unlike the alternative solution. In the alternative solution, we would either need to make the continuation token parameter an optional parameter or pass in a null value for it. Both these solutions are harder to read than the current solution.

akotalwar:
>         private CosmosElement TryUnwrapContinuationToken()

Added an early return

---
In reply to: [1574158577](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3882#issuecomment-1574158577) [](http://example.com/codeflow?ancestors=1574158577,1574158577)

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs:111 in b88eecb. [](commit_id = b88eecbdfc28a2b52cdcadc450824f6c6cb40520, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
neildsh: This implies some queries will stop working or return incorrect results when data volume increases. I am okay with this since the behavior is opt-in, but it's good to make a conscious note that we are making that choice here.

neildsh:
>         private CosmosElement TryUnwrapContinuationToken()

what is the objection to adding a method parameter?

---
In reply to: [1574246115](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3882#issuecomment-1574246115) [](http://example.com/codeflow?ancestors=1574246115,1574158577)

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs:111 in b88eecb. [](commit_id = b88eecbdfc28a2b52cdcadc450824f6c6cb40520, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionQueryPipelineStage.cs
neildsh: Just explicitly pass a null value, since that is what is happening under the covers anyway. This also decouples this method from the `TryUnwrapContinuationToken` call

neildsh (approved):

sboshra (approved):
:shipit:

PR #3900, opened 2023-06-08
[Internal] Client Telemetry: Adds Client Telemetry pipeline sending data to service

sourabh1007:
## Description

Right now, if we make any change unknowingly which breaks the contract with service, we wouldn't know until unless we tested it explicitly or we get it from customers once we shipped a version of SDK with the regression.

**Covering following things:**
1. Adding 2 new pipelines for Client Telemetry feature:
   a) **Emulator Tests ClientTelemetry:** It is normal integration test against Emulator account. We don't call Client Telemetry Service here. It runs as part of our regular pipelines.
   b) **Integration Tests ClientTelemetry:** It runs as part of SDK release and here we actually call the service to validate if current SDK supports current version of Client Telemetry Service.
Proof of testing : image

3. Using windows-2019 VM as services doesn't support latest VM as of now. tools team is working on it.

## Type of change


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTests.cs
```diff
@@ -125,8 +127,10 @@ public void TestInitialize()
                 { ClientTelemetryOptions.ThreadWaitIntervalInMsName, ClientTelemetryOptions.ThreadWaitIntervalInMsUnit }
             };
 
-            this.cosmosClientBuilder = TestCommon.GetDefaultConfiguration()
-                                        .WithApplicationPreferredRegions(this.preferredRegionList);
+            string connectionString = Environment.GetEnvironmentVariable("COSMOS.DB_CONNECTION_STRING");
```
ealsur, line 130: Was this config value added to the pipeline configuration? How is it being populated?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTests.cs
```diff
@@ -125,8 +127,10 @@ public void TestInitialize()
...
+            this.cosmosClientBuilder = new CosmosClientBuilder(connectionString: connectionString)
+                                                .WithCustomSerializer(new CosmosJsonDotNetSerializer())
```
ealsur, line 132: Why do we need to specify this Serializer?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTests.cs
sourabh1007: yes, it is added in the pipeline variables.

kirankumarkolli:
> Right now, if we make any change unknowingly which breaks the contract with service, we wouldn't know until unless we tested it explicitly or we get it from customers once we shipped a version of SDK with the regression.

Is this to cover for scenarios when payload didn't reach service? 
If it's after, then service side alerts should cover it right? If tis a gap right now, isn't that the right place to, have it?

kirankumarkolli (changes requested):
https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3900#issuecomment-1659302730

Waiting for clarification on above.

sourabh1007:
> > Right now, if we make any change unknowingly which breaks the contract with service, we wouldn't know until unless we tested it explicitly or we get it from customers once we shipped a version of SDK with the regression.
> 
> Is this to cover for scenarios when payload didn't reach service? If it's after, then service side alerts should cover it right? If tis a gap right now, isn't that the right place to, have it?

it is to make sure, as part of our changes in SDK, we are not breaking any contract with client telemetry service unknowingly. FYI. Java SDK already has similar pipeline.

Microsoft.Azure.Cosmos/src/Telemetry/Models/MetricInfo.cs
```diff
@@ -81,6 +90,10 @@ internal MetricInfo SetAggregators(LongConcurrentHistogram histogram, double adj
                 };
                 this.Percentiles = percentile;
             }
+            else
+            {
+                this.Percentiles = MetricInfo.defaultPercentiles;
```
kirankumarkolli, line 95: Is this change a functional MUST? (Ideal to have them as separate PRs for its visibility)

What do service do if the client doesn't pass percentile section?

sourabh1007: Basically, service doesn't allow Metrics with partial or no information. Other fields are of primitive type hence always have 0 value except percentile which is an object type, hence has `null `value.

**Why I included it as part of this PR?** I identified it during my testing of this PR hence made this change as believed it is just a null check. If you think that it is something should be taken as separate PR. I can make this change separately also.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    [TestClass]
+    [TestCategory("ClientTelemetryRelease")]
+    public class ClientTelemetryReleaseTests : ClientTelemetryTests
+    {
+        [ClassInitialize]
+        public static new void ClassInitialize(TestContext context)
+        {
+            ClientTelemetryTests.ClassInitialize(context);
+
+            // It will go away in next PR
```
kirankumarkolli, line 26: Below will flow through the account config right?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
```diff
@@ -0,0 +1,49 @@
...
+            Environment.SetEnvironmentVariable(ClientTelemetryOptions.EnvPropsClientTelemetryEnabled, "true");
+            Environment.SetEnvironmentVariable(ClientTelemetryOptions.EnvPropsClientTelemetrySchedulingInSeconds, "1");
+            Environment.SetEnvironmentVariable(ClientTelemetryOptions.EnvPropsClientTelemetryEndpoint, "https://tools.cosmos.azure.com/api/clienttelemetry/trace");
```
kirankumarkolli, line 29: Any possible impact of it on the tools livesite? (till these are excluded/removed)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
sourabh1007: yes...already implemented it in another PR..once that is in..I will remove these.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/HttpHandlerHelper.cs
```diff
@@ -39,7 +41,14 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                 }
             }
 
-            httpResponse =  await base.SendAsync(request, cancellationToken);
+            try
+            {
+                httpResponse = await base.SendAsync(request, cancellationToken);
+            }
+            catch (Exception ex) {
+                this.ExceptionIntercepter?.Invoke(request, ex);
```
kirankumarkolli, line 49: In-case of intetrceptor not configured, it results in exception dropped right? 

Is invoker expected to define the exception throw?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
sourabh1007: These configurations are used only by SDK. So, I don't think it should have any impact on tools live site.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/HttpHandlerHelper.cs
sourabh1007: ohh yes..I should throw it as it is if intercepter is not configured.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
kirankumarkolli: I see that the handler is intercepting it => good

templates/build-test.yml
```diff
@@ -205,3 +207,37 @@ jobs:
       nugetConfigPath: NuGet.config
       publishTestResults: true
       testRunTitle: Microsoft.Azure.Cosmos.Encryption.Custom.EmulatorTests
+
+- job:
+  displayName: ${{ parameters.ClientTelemetryTestName }} ${{ parameters.BuildConfiguration }}
+  timeoutInMinutes: 120
+  condition: and(succeeded(), eq('${{ parameters.OS }}', 'Windows'))
+  pool:
+    vmImage: windows-2019
```
ealsur, line 216: Why not `vmImage: ${{ parameters.VmImage }}` like the others?

ealsur: We are using `windows-2022`.

image

templates/build-test.yml
```diff
@@ -205,3 +207,37 @@ jobs:
       nugetConfigPath: NuGet.config
       publishTestResults: true
       testRunTitle: Microsoft.Azure.Cosmos.Encryption.Custom.EmulatorTests
+
+- job:
+  displayName: ${{ parameters.ClientTelemetryTestName }} ${{ parameters.BuildConfiguration }}
```
ealsur, line 212: Where is the filtering that makes this apply only on Release pipeline?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/HttpHandlerHelper.cs
```diff
@@ -39,7 +41,19 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
...
+            {
+                httpResponse = await base.SendAsync(request, cancellationToken);
+            }
+            catch (Exception ex) {
+
+                if (this.ExceptionIntercepter == null)
+                {
+                    throw;
+                }
+                this.ExceptionIntercepter?.Invoke(request, ex);
```
ealsur, line 54: There is already a check for null above, at this point, we know the variable is not null, right?

```suggestion
                this.ExceptionIntercepter.Invoke(request, ex);
```

templates/build-test.yml
```diff
@@ -5,11 +5,13 @@ parameters:
   Arguments: ''
   VmImage: '' # https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops
   OS: 'Windows'
-  EmulatorPipeline1Arguments: ' --filter "TestCategory!=Quarantine & TestCategory!=Functional & (TestCategory=Query|TestCategory=ReadFeed|TestCategory=Batch|TestCategory=ChangeFeed)" --verbosity normal '
-  EmulatorPipeline2Arguments: ' --filter "TestCategory!=Quarantine & TestCategory!=Functional & TestCategory!=Query & TestCategory!=ReadFeed & TestCategory!=Batch & TestCategory!=ChangeFeed" --verbosity normal '
+  EmulatorPipeline1Arguments: ' --filter "TestCategory!=Quarantine & TestCategory!=Functional & ClassName!~Microsoft.Azure.Cosmos.SDK.EmulatorTests.ClientTelemetry & (TestCategory=Query|TestCategory=ReadFeed|TestCategory=Batch|TestCategory=ChangeFeed)" --verbosity normal '
```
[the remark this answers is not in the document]
sourabh1007: It was not filtering with test categories hence switched to classname

templates/build-test.yml
sourabh1007: As mentioned in PR description, client telemetry service is throwing SSL error in windows 2022 hence using windows 2019, will update later when support for windows 2022 is ready.

templates/build-test.yml
sourabh1007: As mentioned in PR description, client telemetry service is throwing SSL error with windows 2022, hence using windows 2019, will update later when support for windows 2022 is ready.

templates/build-test.yml
sourabh1007: it is driven from parameters

default value is (in _build-test.yml_)
```
ClientTelemetryTestName: 'EmulatorTests ClientTelemetry'
ClientTelemetryTestClassName: 'Microsoft.Azure.Cosmos.SDK.EmulatorTests.ClientTelemetryTests'
```
for release pipeline, these value gets override with 
```
ClientTelemetryTestName: IntegrationTest ClientTelemetry
ClientTelemetryTestClassName: Microsoft.Azure.Cosmos.SDK.EmulatorTests.ClientTelemetryReleaseTests
```
in _azure-pipelines-official.yml_

templates/build-test.yml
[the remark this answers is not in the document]
ealsur: That's odd, where is the TestCategory defined? In the base class?

templates/build-test.yml
ealsur: We can already identify when the pipeline is for Release or CI, why do we need the extra variables? build-test can define an extra parameter that is Release true/false depending on the scenario?

templates/build-test.yml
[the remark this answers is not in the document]
sourabh1007: both base and parent..and both are different test classes...in the release class, overriding few functions

templates/build-test.yml
```diff
@@ -205,3 +207,37 @@ jobs:
...
+    clean: true  # if true, execute `execute git clean -ffdx && git reset --hard HEAD` before fetching
+
+  # Add this Command to Include the .NET 6 SDK
+  - task: UseDotNet@2
+    displayName: Use .NET 6.0
+    inputs:
+      packageType: 'sdk'
+      version: '6.x'
+  
+  - ${{ if ne(parameters.ReleasePackage, true) }}:
```
ealsur, line 229: How does this parameter work? It is not in the definition of the yml file

templates/build-test.yml
```diff
@@ -205,3 +207,37 @@ jobs:
...
+    condition: succeeded()
+    inputs:
+      command: test
+      projects: 'Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/*.csproj'
+      arguments: --filter "ClassName=${{ parameters.ClientTelemetryTestClassName }}" --verbosity normal --configuration Debug /p:OS=${{ parameters.OS }}
+      nugetConfigPath: NuGet.config
+      publishTestResults: true
+      testRunTitle: Microsoft.Azure.Cosmos.EmulatorTests
+    env:
+      COSMOSDB_ACCOUNT_CONNECTION_STRING: $(COSMOSDB_ACCOUNT_CONNECTION_STRING)
```
ealsur, line 243: Why is this needed? THe other Emulator tests are not passing this

templates/build-test.yml
```diff
@@ -205,3 +207,37 @@ jobs:
...
+  - ${{ if ne(parameters.ReleasePackage, true) }}:
+    - template:  emulator-setup.yml
+
+  - task: DotNetCoreCLI@2
+    displayName: Microsoft.Azure.Cosmos.EmulatorTests
+    condition: succeeded()
+    inputs:
+      command: test
+      projects: 'Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/*.csproj'
+      arguments: --filter "ClassName=${{ parameters.ClientTelemetryTestClassName }}" --verbosity normal --configuration Debug /p:OS=${{ parameters.OS }}
```
ealsur, line 238: Why is it debug configuration and verbosity normal? Compared with others:

```

      arguments: .......... --configuration ${{ parameters.BuildConfiguration }} /p:OS=${{ parameters.OS }}

```

templates/build-test.yml
ealsur: Think about this like writing code. The caller does not care about class names, right? It somehow leaks what is happening to the caller by making the caller decide which ClassName to use, if in the future we don't want to use ClassName, we need to change multiple places.

Why not have a parameter that is something like "IsRelease" defaults to false, can be sent to true. And inside this file depending on this, we set a variable with the ClassNames and/or test name?

templates/build-test.yml
sourabh1007: It is needed for the Integration Test which will run in release pipeline where it will talk to real database account and send telemetry to the service.

templates/build-test.yml
sourabh1007: changed debug configuration as parameter, verbosity is already set to normal in parameters for other pipelines.

templates/build-test.yml
sourabh1007: May be it was always false :( ..i didn't realize because i just switched the condition to test if this condition is working but it makes sense add this parameter in the definition.

ealsur: But this doesn't make sense. `emulator-setup.yml` starts the Emulator. WIthout the Emulator, the Emulator tests below won't ever work.

ealsur (approved):

kirankumarkolli (changes requested):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    [TestCategory("ClientTelemetryRelease")]
+    public class ClientTelemetryReleaseTests : ClientTelemetryTests
+    {
+        [ClassInitialize]
+        public static new void ClassInitialize(TestContext context)
+        {
+            ClientTelemetryTests.ClassInitialize(context);
+
+            // It will go away in next PR
+            Environment.SetEnvironmentVariable(ClientTelemetryOptions.EnvPropsClientTelemetryEnabled, "true");
```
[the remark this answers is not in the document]
sourabh1007: It will remove in next pr

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    using Microsoft.Azure.Cosmos.Fluent;
+    using Microsoft.Azure.Cosmos.Telemetry;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    /// <summary>
+    /// In Release pipeline, no need to mock Client Telemetry Service Call and Test will talk to the real database account.
+    /// </summary>
+    [TestClass]
+    [TestCategory("ClientTelemetryRelease")]
+    public class ClientTelemetryReleaseTests : ClientTelemetryTests
```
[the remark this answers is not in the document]
sourabh1007: Idea is to keep release and emulator test separate....while they will run same set of test with slight modifications.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTests.cs
```diff
@@ -65,68 +79,89 @@ public void TestInitialize()
...
 
-                        lock (this.actualInfo)
+                        lock(this.actualInfo)
                         {
                             this.actualInfo.Add(JsonConvert.DeserializeObject<ClientTelemetryProperties>(jsonObject));
                         }
+                    }
 
-                        return Task.FromResult(result);
+                    return this.HttpHandlerRequestCallbackChecks(request);
```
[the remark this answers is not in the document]
sourabh1007: Returning null means return as it is.

templates/build-test.yml
```diff
@@ -205,3 +206,44 @@ jobs:
...
+- job:
+  variables:
+    ${{ if ne(parameters.ReleasePackage, true) }}:
+      TestName: Microsoft.Azure.Cosmos.EmulatorTests
+      ClientTelemetryTestClassName: Microsoft.Azure.Cosmos.SDK.EmulatorTests.ClientTelemetryTests
+    ${{ if eq(parameters.ReleasePackage, true) }}:
+      TestName: Integration Test With Client Telemetry Service
+      ClientTelemetryTestClassName: Microsoft.Azure.Cosmos.SDK.EmulatorTests.ClientTelemetryReleaseTests
+  displayName: TelemetryToService ${{ parameters.BuildConfiguration }}
+  timeoutInMinutes: 120
```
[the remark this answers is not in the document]
sourabh1007: Taken from existing tests only.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
```diff
@@ -0,0 +1,158 @@
...
+    /// <summary>
+    /// In Release pipeline, no need to mock Client Telemetry Service Call and Test will talk to the real database account.
+    /// </summary>
+    [TestClass]
+    [TestCategory("ClientTelemetryRelease")]
+    public class ClientTelemetryReleaseTests : ClientTelemetryTestsBase
+    {
+        public override CosmosClientBuilder GetBuilder()
+        {
+            string connectionString = ConfigurationManager.GetEnvironmentVariable<string>("COSMOSDB_ACCOUNT_CONNECTION_STRING", null);
```
kirankumarkolli, line 23: If ENV variable can be set conditionally for the release only then Testbase GetBuilder() it-self can abstract it out. 

GetBuilder() might then look like
{
   if (ENV(COSMOSDB_ACCOUNT_CONNECTION_STRING) != null}
{
   return connectionstring based init;
}
return generic test cllient;

sourabh1007: We can put env under condition..but i don't see much value there....right now it is very clear that release is using env variable and emulator is using default conf. If we have such logic, somebody need to understand the code why would we have such condition.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
```diff
@@ -0,0 +1,158 @@
...
+            return new CosmosClientBuilder(connectionString: connectionString);
+        }
+
+        public override Task<HttpResponseMessage> HttpHandlerRequestCallbackChecks(HttpRequestMessage request)
+        {
+            return null;
```
kirankumarkolli, line 29: Please add code comment that `null` will force sending real call down stream in pipeline.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        [ClassInitialize]
+        public static new void ClassInitialize(TestContext context)
+        {
+            ClientTelemetryTestsBase.ClassInitialize(context);
+
+            // It will go away in next PR
+            Environment.SetEnvironmentVariable(ClientTelemetryOptions.EnvPropsClientTelemetryEnabled, "true");
```
kirankumarkolli, line 38: Are the ENV variables setting needed pre-client creation? 
Post set-up is non-deterministic right?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
sourabh1007: I can add that comment but that is how httphandler works in test everywhere....it is not something I am adding newly.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
kirankumarkolli: I see that its only setting the ENV not creating client, so should be good. 
Is my understanding right?

sourabh1007: It will go away for release test in next PR

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTests.cs
```diff
@@ -4,1121 +4,154 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
 
+    /// <summary>
+    /// In Emulator Mode, Run test against emulator and mock client telemetry service calls
+    /// </summary>
     [TestClass]
-    public class ClientTelemetryTests : BaseCosmosClientHelper
+    [TestCategory("ClientTelemetryEmulator")]
+    public class ClientTelemetryTests : ClientTelemetryTestsBase
```
kirankumarkolli, line 19: Rename the class to include "Emulator' in type for better readability

sourabh1007: By default all tests are emulator test in this project....I have comments also to explain the same.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        [TestCleanup]
+        public override async Task Cleanup()
+        {
+            await base.Cleanup();
+        }
+
+        [TestMethod]
+        [DataRow(ConnectionMode.Direct, true)]
```
kirankumarkolli, line 62: Replicating the DataRows definitions is not ideal, thoughts?

sourabh1007: What do you mean by this?

sourabh1007: Do you mean using same datarow def for release and emulator tests?

templates/build-test.yml
```diff
@@ -205,3 +206,44 @@ jobs:
       nugetConfigPath: NuGet.config
       publishTestResults: true
       testRunTitle: Microsoft.Azure.Cosmos.Encryption.Custom.EmulatorTests
+
+- job:
```
kirankumarkolli, line 210: Every gate run now needs this extra check right? (even non-release)? 

For non-release the emulator client telemetry tests can be run part of above emulator tests right? 
And only do release test for official (this job can be moved into official yml if needed)

/cc: @ealsur, @sourabh1007 thoughts?

sourabh1007: What is the benefit of this approach? Right now, we have a single job serving release and emulator tests both. So IMO readability wise it is quite clear to understand what we are doing really with client telemetry test in release and non release mode.
Otherwise, readability wise somebody needs to understand where we are running client telemetry release and emulator test.

sourabh1007: Do we really worry about this concurrency load. I am not sure about it.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTests.cs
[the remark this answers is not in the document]
sourabh1007: added documentation in release class.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
sourabh1007: added comment

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryReleaseTests.cs
sourabh1007: already discussed this with @ealsur and we decided to keep release and emulator test separately. 
1. `dotnet test --filter` works with category name only with this model.
2. It keeps emulator and release tests separate and flexibility to change them in the future independently.

templates/build-test.yml
ealsur: > I know there is a limit but I think we are far from this limit.

image

There is a limit of 26 parallel jobs. This sounds like it is making 1 PR run 19 jobs (adding 1 from the 18 that we normally run). It is not a huge impact overall compared with the existing list, but I do agree that it could possibly be folded into the existing one. We are already in a number that is pretty high (1 PR generating 18 out of 26) so if we can at least try to not add more it might be worth it.

Based on the PR it looks like:

1. If ReleasePackage flag is true, then running the extra job makes sense, as it's a special scenario that requires using a different connection string.
2. If ReleasePackage flag is false, then possibly this could be folded into the existing run (EmulatorPipeline1Arguments)

templates/build-test.yml
```diff
@@ -5,9 +5,9 @@ parameters:
   Arguments: ''
   VmImage: '' # https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops
   OS: 'Windows'
-  EmulatorPipeline1Arguments: ' --filter "TestCategory!=Quarantine & TestCategory!=Functional & (TestCategory=Query|TestCategory=ReadFeed|TestCategory=Batch|TestCategory=ChangeFeed)" --verbosity normal '
-  EmulatorPipeline2Arguments: ' --filter "TestCategory!=Quarantine & TestCategory!=Functional & TestCategory!=Query & TestCategory!=ReadFeed & TestCategory!=Batch & TestCategory!=ChangeFeed" --verbosity normal '
-  EmulatorPipeline1CategoryListName: ' Query, ChangeFeed, ReadFeed, Batch ' # Divided in 2 categories to run them in parallel and reduce the PR feedback time
+  EmulatorPipeline1Arguments: ' --filter "TestCategory!=Quarantine & TestCategory!=Functional & TestCategory!=ClientTelemetryRelease & (TestCategory=ClientTelemetryEmulator|TestCategory=Query|TestCategory=ReadFeed|TestCategory=Batch|TestCategory=ChangeFeed)" --verbosity normal '
```
kirankumarkolli, line 8: `TestCategory!=ClientTelemetryRelease` is this condition needed? 
OR condition below is explicitly excludes already right?

kirankumarkolli (approved):

templates/build-test.yml
sourabh1007: in both the pipelines we need to exclude it explicitly , similar to `functional`, `quarantine`

Issue #3416, opened 2022-08-25
ToFeedIterator throws a non-specific exception when an invalid continuation token is passed in the query


vit-svoboda:
>We are continuously addressing and improving the SDK, if possible, make sure the problem persist in the [latest SDK version](https://www.nuget.org/packages/Microsoft.Azure.Cosmos).

**Describe the bug**
`IQueryable<T>` extension method `ToFeedIterator` makes it impossible to write error handling for an invalid continuation token being  passed in the query. A general `JsonInvalidTokenException` wrapped in `ExceptionWithStackTraceException` is thrown instead of a descriptive `CosmosException`.

**To Reproduce**
```
using Microsoft.Azure.Cosmos;
using Microsoft.Azure.Cosmos.Fluent;
using Microsoft.Azure.Cosmos.Linq;

var client = new CosmosClientBuilder("<cosmos connection string>")
    .WithConnectionModeDirect()
    .Build();

var container = client.GetContainer("<database name>", "<collection name>");
var continuationToken = "not json goes boom";

var queryable = container.GetItemLinqQueryable<dynamic>(continuationToken: continuationToken);

// Throws JsonInvalidTokenException wrapped in ExceptionWithStackTraceException 
var iterator = queryable.ToFeedIterator(); 

var response = await iterator.ReadNextAsync();
```

**Expected behavior**
Either `ToFeedIterator` or `ReadNextAsync` throws `CosmosException` saying "The continuation token specified in the request is malformed. Re-run the query without it and iterate." with the `SubStatusCode` mentioned in https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3074.

**Actual behavior**
`ToFeedIterator` throws `ExceptionWithStackTraceException` with inner `JsonInvalidTokenException` saying "Encountered an element that is not a valid JSON value (false / null / true / object / array / number / string), Windows/10.0.19044 cosmos-netstandard-sdk/3.29.1" but no mention of what has been parsed.

**Environment summary**
SDK Version: 3.30.0
OS Version (e.g. Windows, Linux, MacOSX): Windows 10

**Additional context**
The descriptive `CosmosException` is thrown from the `ReadNextAsync` if the `FeedIterator` is created like so:
```
var queryDefinition = queryable.ToQueryDefinition();

var iterator = client.GetDatabaseQueryIterator<object>(
    queryDefinition,
    continuationToken);
```
[...]

PR #3917, opened 2023-06-15
Query: Fixes malformed continuation token exception type and message

akotalwar:
# Pull Request Template

## Description

This change fixes the bug (https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3416) where the malformed continuation token exception was not a CosmosException and lacked enough details to identify the cause. Now, users will get a CosmosException with a clear message and the status code and sub status code to indicate the error.

## Type of change

## Closing issues

To automatically close an issue: closes https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3416


Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
```diff
@@ -187,7 +189,28 @@ public static bool TryCreateFromBuffer<TCosmosElement>(ReadOnlyMemory<byte> buff
         public static CosmosElement Parse(string json)
         {
             TryCatch<CosmosElement> tryParse = CosmosElement.Monadic.Parse(json);
-            tryParse.ThrowIfFailed();
+            bool isMalformedJSON = tryParse.InnerMostException.Message.Contains("Encountered an element that is not a valid JSON")
```
ealsur, line 192: Is there a way to do this by testing the exception type instead of string comparison?

akotalwar: I could not find a suitable exception type to compare with, so I had to use string comparison instead.

akotalwar: I could not find a suitable exception type to compare with, so I had to use string comparison instead.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
```diff
@@ -211,6 +211,28 @@ public async Task ItemLINQQueryTest()
...
+                    continuationToken: "Malformed String");
+                try
+                {
+                    FeedIterator<ToDoActivity> iterator = querable.ToFeedIterator();
+                }
+                catch (CosmosException exception)
+                {
+                    Assert.IsTrue(exception.StatusCode == System.Net.HttpStatusCode.BadRequest);
+                    Assert.IsTrue(exception.SubStatusCode == 20007);
+                    return;
```
adityasa, line 231: Nit, not needed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
```diff
@@ -211,6 +211,28 @@ public async Task ItemLINQQueryTest()
...
+                }
+            }
+        }
```
adityasa, line 234: >} [](http://example.com/codeflow?start=8&length=1)

Kindly assert that exception is (always) received.

Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
adityasa, line 192: >InnerMostException [](http://example.com/codeflow?start=44&length=18)

Wouldn't this be null in cases when tryParse didn't fail?

Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
```diff
@@ -187,7 +189,28 @@ public static bool TryCreateFromBuffer<TCosmosElement>(ReadOnlyMemory<byte> buff
...
+                    message: $"Malformed Continuation Token: {json}.",
+                    headers: CosmosQueryResponseMessageHeaders.ConvertToQueryHeaders(
+                        new Headers(),
+                        default,
+                        default,
+                        20007,
+                        default),
+                    stackTrace: malformedContinuationTokenException.StackTrace,
+                    innerException: malformedContinuationTokenException,
+                    trace: null);
```
adityasa, line 208: It seems like this should be pushed inside ThrowIfFailed.

adityasa: Actually I take this back. Why is this code in CosmosElement? Continuation token is a Query concept. CosmosElement is more general. Shouldn't this be a part of appropriate query execution codepath (only) ?

Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
ealsur: Based on how `CosmosElement.Monadic.Parse` calls into `CreatFromBuffer` wouldn't this exception, if failed due to JSON parsing, be of type `JsonParseException`?

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/8684535d233ea6a7147d857fae9a6c69e467c827/Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs#L106-L115

Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
ealsur: Based on my other comment, I believe you are right, the caller should be getting a `JsonParseException` when `ThrowIfFailed` is called, and can act on the `JsonParseException`?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
akotalwar: We have to return from the catch block because otherwise we will never exit the loop. This is because the feedIterator.HasMoreResults will remain true even after the exception occurs. Therefore, we should either return from the catch block or update the feedIterator in the catch block.

Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
```diff
@@ -187,7 +189,32 @@ public static bool TryCreateFromBuffer<TCosmosElement>(ReadOnlyMemory<byte> buff
         public static CosmosElement Parse(string json)
         {
             TryCatch<CosmosElement> tryParse = CosmosElement.Monadic.Parse(json);
-            tryParse.ThrowIfFailed();
+            try
+            {
+                tryParse.ThrowIfFailed();
+            }
+            catch (Exception exception)
```
ealsur, line 196:
```suggestion
            catch (Exception exception) when (exception.InnerException is JsonParseException)
```

Removes the need for the if/else

Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
```diff
@@ -187,7 +189,32 @@ public static bool TryCreateFromBuffer<TCosmosElement>(ReadOnlyMemory<byte> buff
...
+                if (exception.InnerException is JsonParseException)
+                {
+                    MalformedContinuationTokenException malformedContinuationTokenException = new MalformedContinuationTokenException(exception.Message);
+                    throw CosmosExceptionFactory.CreateBadRequestException(
+                            message: $"Malformed Continuation Token: {json}.",
+                            headers: CosmosQueryResponseMessageHeaders.ConvertToQueryHeaders(
+                                new Headers(),
+                                default,
+                                default,
+                                20007,
```
ealsur, line 207: Is this value a const that is somewhere?

Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
akotalwar: With this change you would replace the if else with two catch blocks. The first block would be catch (Exception exception) when (exception.InnerException is JsonParseException) and the second would be catch(Exception ex) where we would throw the exception. 

Instead of having these two catch blocks I think its better to have the if/else we currently have.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
```diff
@@ -211,6 +211,31 @@ public async Task ItemLINQQueryTest()
...
+                }
+                catch (CosmosException exception)
+                {
+                    Assert.IsTrue(exception.StatusCode == System.Net.HttpStatusCode.BadRequest);
+                    Assert.IsTrue(exception.SubStatusCode == 20007);
+                    Assert.IsTrue(exception.Message.Contains("Malformed String"));
+                    return;
+                }
+
+                Assert.IsFalse(true, "Should never reach till here, hence ensuring that an exception is always recieved");
```
adityasa, line 235: >Assert.IsFalse(true [](http://example.com/codeflow?start=16&length=19)

Instead, please use
Assert.Fail

Microsoft.Azure.Cosmos/src/CosmosElements/CosmosElement.cs
```diff
@@ -187,7 +190,29 @@ public static bool TryCreateFromBuffer<TCosmosElement>(ReadOnlyMemory<byte> buff
...
+                MalformedContinuationTokenException malformedContinuationTokenException = new MalformedContinuationTokenException(exception.Message);
+                throw CosmosExceptionFactory.CreateBadRequestException(
+                        message: $"Malformed Continuation Token: {json}.",
+                        headers: CosmosQueryResponseMessageHeaders.ConvertToQueryHeaders(
+                            new Headers(),
+                            default,
+                            default,
+                            (int)SubStatusCodes.MalformedContinuationToken,
+                            default),
+                        stackTrace: malformedContinuationTokenException.StackTrace,
```
ealsur, line 208: nit: Should the callstack be the stack of `exception`? `malformedContinuationTokenException.StackTrace` just contains this method (because the exception is created here)

Microsoft.Azure.Cosmos/src/ReadFeed/ReadFeedIteratorCore.cs
```diff
@@ -113,7 +116,25 @@ internal sealed class ReadFeedIteratorCore : FeedIteratorInternal
                             else
                             {
                                 CosmosString tokenAsString = (CosmosString)token;
-                                state = ReadFeedState.Continuation(CosmosElement.Parse(tokenAsString.Value));
+                                try
+                                {
+                                    state = ReadFeedState.Continuation(CosmosElement.Parse(tokenAsString.Value));
```
ealsur, line 121: This file is only for ReadFeed (when the query is empty), what about when the query has some value?

akotalwar: This function, CosmosElement.Parse, is not used by all query paths. However, we can extend similar support to the query paths that do use it, such as DistinctQueryPipeline, SkipQueryPipeline and TakeQueryPipeline.

akotalwar: I take back what I said above. Below is my updated response to your comment.

The Parse() internally calls a Monadic.Parse(). This is the function which actually throws the JsonParseException. Monadic.Parse() has 8 callers. Out of the 8, one is the Parse(), where we don't properly handle the failure. The other 7 callers (ChangeFeedCrossFeedRangeState, ChangeFeedIteratorCore, QueryIterator etc.) of this function properly handle the exception, where they wrap it in a different exception which is more descriptive before sending it to the user. 

So now, let's talk about the Parse(). 
In this, we just throw whatever exception Monadic.Parse() provides, and our goal is to handle this thrown exception better in the callers of Parse(). 

So, there are 4 locations where we call Parse(). Out of these 4, one has already been handled (ReadFeedIteratorCore) and the remaining three calls sit in the MoveNextAsync for the 3 pipeline stages (DistinctQueryPipeline, SkipQueryPipeline and TakeQueryPipeline). 
I dont believe we need to add support for these 3 cases for the following reason.
The MoveNextAsync() does not take a continuation token as input. Hence, a user cannot provide a continuation token to this function.  This means that the the only place where a user can provide a continuation token for these pipelines is in their MonadicCreate calls. However, in these MonadicCreate function, there is already a check to ensure that the continuation token is valid and if its not then a MalformedContTokenException is thrown.  What this means is that only a valid continuation token can be provided to a MoveNextAsync(), hence making any continuation check in this function redundant. 
Hence, the only case we need to improve is when a user provides a continuation token in the ReadFeedIterator and an ambiguous exception is thrown and this has already been handled by this PR.

akotalwar: Yes, they all return a MalformedContinuationTokenException similar to what we do in ReadFeedIteratorCore

adityasa (approved):
:shipit:

ealsur (approved):

PR #3934, opened 2023-06-22
Subpartitioning: Fixes bug for queries on subpartitioned containers

### LINKED ISSUE #3933: Subpartitioning: Query with top 50 results are getting into infinite loop.

**Describe the bug**
If customer has a hierarchal partition key
`select * from c ` query with partiontion is working but `select top 50 * from c` query is going into infinite loop.

**Expected behavior**
`select top 50 * from c` query should return top 50 results

**Environment summary**
SDK Version: 3.25.0
OS Version (e.g. Windows, Linux, MacOSX) any

NaluTripician:
# Pull Request Template

## Description

Similar to #3879.  For queries across physical partition with prefix partition keys using `top` in the queries, `ReadNextAsync` will get stuck in an infinite loop. 

## Type of change

## Closing issues

To automatically close an issue: closes #3933


KevinPoorDeveloper:
Hi Nalu,
Thank you for your quick work on this. 

I'm pretty convinced that this isn't the only issue here. Our partial partition key that we discussed crossed 40GB, and I assume an additional physical partition was created. Since then, queries work on localhost through VS, but fail on production, whenever we only include part of the partition key in the query. All queries work fine with the full partition query.
Perhaps we shall speak again tomorrow.
Thank you for your time and help,
-Kevin

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -24,9 +25,11 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
     using Microsoft.Azure.Cosmos.Query.Core.Pipeline.Tokens;
     using Microsoft.Azure.Cosmos.Query.Core.QueryClient;
     using Microsoft.Azure.Cosmos.Query.Core.QueryPlan;
+    using Microsoft.Azure.Cosmos.Serialization.HybridRow.Schemas;
```
ealsur, line 28: Why is this dependency needed?

NaluTripician: Looks like it was added automatically by intellisense and I missed it in the cleanup. Removed in latest push.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -9,6 +9,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
+    using global::Azure;
```
ealsur, line 12: nit: is this using needed?

Microsoft.Azure.Cosmos/src/Query/Core/QueryClient/ContainerQueryProperties.cs
```diff
@@ -4,24 +4,26 @@
...
             this.ResourceId = resourceId;
-            this.EffectivePartitionKeyString = effectivePartitionKeyString;
+            this.EffectivePartitionKeyRanges = effectivePartitionKeyRanges;
             this.PartitionKeyDefinition = partitionKeyDefinition;
             this.GeospatialType = geospatialType;
         }
 
         public string ResourceId { get; }
-        public string EffectivePartitionKeyString { get; }
+        public List<Range<string>> EffectivePartitionKeyRanges { get; }
```
ealsur, line 26:
```suggestion
        public List<Range<string>> EffectiveRanges { get; }
```

These are just Ranges, not PartitionKeyRanges

ealsur, line 26: Why a list of Ranges if `GetEffectivePartitionKeyRange` returns a single `Range<string>`?

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
```diff
@@ -63,17 +63,23 @@ internal class CosmosQueryClientCore : CosmosQueryClient
                 trace,
                 cancellationToken);
 
-            string effectivePartitionKeyString = null;
+            List<Range<string>> effectivePartitionKeyRange = null;
```
ealsur, line 66: nit: Shouldn't this be simply:

```suggestion
            Range<string> effectivePartitionKeyRange = null;
```

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
```diff
@@ -63,17 +63,23 @@ internal class CosmosQueryClientCore : CosmosQueryClient
...
             if (partitionKey != null)
             {
                 // Dis-ambiguate the NonePK if used 
                 PartitionKeyInternal partitionKeyInternal = partitionKey.Value.IsNone ? containerProperties.GetNoneValue() : partitionKey.Value.InternalKey;
-                effectivePartitionKeyString = partitionKeyInternal.GetEffectivePartitionKeyString(containerProperties.PartitionKey);
+                effectivePartitionKeyRange.Add(PartitionKeyInternal.GetEffectivePartitionKeyRange(
```
ealsur, line 71:
```suggestion
                effectivePartitionKeyRange = PartitionKeyInternal.GetEffectivePartitionKeyRange(
```

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -801,6 +802,24 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
             return null;
         }
 
+        private static Documents.Routing.PartitionKeyInternal GetNoneValue(
```
ealsur, line 805: When is this used?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -27,6 +28,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
     using Microsoft.Azure.Cosmos.SqlObjects;
     using Microsoft.Azure.Cosmos.SqlObjects.Visitors;
     using Microsoft.Azure.Cosmos.Tracing;
+    using Microsoft.Azure.Documents;
```
ealsur, line 31: Undo this and use the `Documents.` where it is needed, otherwise you have to also do the `Cosmos.PartitionKey` at the bottom of the file

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -211,12 +213,12 @@ internal static class CosmosQueryExecutionContextFactory
...
                                 Documents.PartitionKeyDefinition partitionKeyDefinition = GetPartitionKeyDefinition(inputParameters, containerQueryProperties);
-                                List<Documents.PartitionKeyRange> targetRanges = await cosmosQueryContext.QueryClient.GetTargetPartitionKeyRangesByEpkStringAsync(
+                                List<PartitionKeyRange> targetRanges = await cosmosQueryContext.QueryClient.GetTargetPartitionKeyRangesAsync(
                                     cosmosQueryContext.ResourceLink,
                                     containerQueryProperties.ResourceId,
-                                    containerQueryProperties.EffectivePartitionKeyString,
+                                    containerQueryProperties.EffectivePartitionKeyRanges,
                                     forceRefresh: false,
-                                    createQueryPipelineTrace);
+                                    trace);
```
ealsur, line 221: Why change `createQueryPipelineTrace` for `trace`?

Microsoft.Azure.Cosmos/src/Query/Core/QueryClient/ContainerQueryProperties.cs
ealsur: Is it because `QueryClient.GetTargetPartitionKeyRangesAsync` takes a list?

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
ealsur, line 71: WOuldn't this throw? `effectivePartitionKeyRange` is `null`, you didn't initialize it.

ealsur (changes requested):
Please see the comments

Microsoft.Azure.Cosmos/src/Query/Core/QueryClient/ContainerQueryProperties.cs
NaluTripician: Yes, in all of the cases where we would use this it takes a list.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
NaluTripician: Left in from first iteration. Removed in latest push.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -799,7 +798,7 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
             }
 
             return null;
-        }
+        }
```
ealsur, line 801: nit: undo line

ealsur (approved):
Please get a review from the query folks

Microsoft.Azure.Cosmos/src/Query/Core/QueryClient/ContainerQueryProperties.cs
adityasa, line 26: >List [](http://example.com/codeflow?start=15&length=4)

IReadOnlyList would be nice (if possible).

adityasa:
Hi Nalu, thanks for taking this on, much appreciated.
Can you kindly add coverage/regression test for the specific query in question?

NaluTripician:
> Hi Nalu, thanks for taking this on, much appreciated. Can you kindly add coverage/regression test for the specific query in question?

Yes, @SrinikhilReddy is currently working on the tests.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
neildsh: I believe this is so that the trace is properly nested

Microsoft.Azure.Cosmos/src/Query/Core/QueryClient/ContainerQueryProperties.cs
neildsh: I do not quite understand this. If we always return a list with a single element, why return a list at all ? If the caller needs a list, they can wrap it themselves.

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
```diff
@@ -63,17 +63,26 @@ internal class CosmosQueryClientCore : CosmosQueryClient
...
-                effectivePartitionKeyString = partitionKeyInternal.GetEffectivePartitionKeyString(containerProperties.PartitionKey);
+                effectivePartitionKeyRange = new List<Range<string>>
+                {
+                    PartitionKeyInternal.GetEffectivePartitionKeyRange(
+                        containerProperties.PartitionKey,
+                        new Range<PartitionKeyInternal>(
+                            min: partitionKeyInternal,
+                            max: partitionKeyInternal,
+                            isMinInclusive: true,
+                            isMaxInclusive: true))
```
neildsh, line 79: `Range<PartitionKeyInternal>.GetPointRange(partitionKeyInternal)`

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -211,10 +211,10 @@ internal static class CosmosQueryExecutionContextFactory
 
                                 // Only thing that matters is that we target the correct range.
                                 Documents.PartitionKeyDefinition partitionKeyDefinition = GetPartitionKeyDefinition(inputParameters, containerQueryProperties);
-                                List<Documents.PartitionKeyRange> targetRanges = await cosmosQueryContext.QueryClient.GetTargetPartitionKeyRangesByEpkStringAsync(
```
neildsh, line 214: >GetTargetPartitionKeyRangesByEpkStringAsync [](http://example.com/codeflow?start=118&length=43)

I think we can delete this method from `CosmosQueryClient`

NaluTripician: It looks like it is used in other places. 
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/d0aac27bfdd5fb90a7ab0523491897454f69ba6b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs#L649

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
NaluTripician: Here I think we do not want to use a point range. This is similar to the earlier fix we had to do. For split physical partitions a point range will not work, but in the case where there is only one physical partition this should return the correct result. Here is the[ previous PR](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3879/files) for more context.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
neildsh: wouldn't we need the same fix in other places too, then?

NaluTripician: Other than the place that I mentioned above are there other places in Query that are assuming PK -> Single Range? And would the changes we made cause the code to enter the first if statement in this block?

neildsh: I think pretty much all code prior to the subpartitioning change assumes that a PK value implies a single range

Microsoft.Azure.Cosmos/src/Query/Core/QueryClient/ContainerQueryProperties.cs
```diff
@@ -4,24 +4,26 @@
...
             this.ResourceId = resourceId;
-            this.EffectivePartitionKeyString = effectivePartitionKeyString;
+            this.EffectivePartitionKeyRanges = effectivePartitionKeyRanges;
             this.PartitionKeyDefinition = partitionKeyDefinition;
             this.GeospatialType = geospatialType;
         }
 
         public string ResourceId { get; }
-        public string EffectivePartitionKeyString { get; }
+        public IReadOnlyList<Range<string>> EffectivePartitionKeyRanges { get; }
```
ealsur, line 26: nit: The name of this property suggest it can be used for any Range, but it is only used for cases when there is a PartitionKey. Would a better name for this be something like `EffectiveRangesForPartitionKey` and add a comment on the property explaining that a PartitionKey equals to 1 range when it's a full Partition Key Value or it can span many if it's a Partial Partition Key value from Subpartitioning?

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
```diff
@@ -35,12 +35,40 @@ namespace Microsoft.Azure.Cosmos.Routing
...
     {
+        internal readonly static PartitionKeyHash None = new PartitionKeyHash(0);
         public PartitionKeyHash(UInt128 value)
+            : this(new UInt128[] { value })
         {
-            this.Value = value;
         }
 
-        public UInt128 Value { get; }
+        public PartitionKeyHash(UInt128[] values)
```
ealsur, line 44: Is this code used just for testing?

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
```diff
@@ -35,12 +35,40 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+                byte[] hashBytes = UInt128.ToByteArray(value);
+                stringBuilder.Append(HexConvert.ToHex(hashBytes, 0, hashBytes.Length));
+            }
+
+            this.Value = stringBuilder.ToString();
+        }
+
+        public string Value { get; }
+
+        public UInt128[] hashValues
```
ealsur, line 58: Is this really public? If so, should the casing be `HashValues`?

Also, the getter is recalculating it always, not not just equal this to the original `UInt128[] values` and instead do the String ToHex only when ToString is called?

ealsur: Like, constructor just sets the value of `HashValues`, `ToString` can do the StringBuilder thing, or even better, do it once and cache it in a private property if needed.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
```diff
@@ -218,6 +246,53 @@ private static PartitionKeyHash Hash(ReadOnlySpan<byte> bytesForHashing)
             }
         }
 
+        internal static class HexConvert
```
ealsur, line 249: Isn't this the same that exists in PartitionKeyInternal type?

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
SrinikhilReddy: Changed as suggested, thanks.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
SrinikhilReddy: Yes, removed the reused methods, had to retain one method which is not a public method in PartitionKeyInternal.cs

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
SrinikhilReddy: yes, changed to the modifier to `internal`

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
```diff
@@ -35,12 +35,29 @@ namespace Microsoft.Azure.Cosmos.Routing
...
     {
+        internal readonly static PartitionKeyHash None = new PartitionKeyHash(0);
         public PartitionKeyHash(UInt128 value)
+            : this(new UInt128[] { value })
         {
-            this.Value = value;
         }
 
-        public UInt128 Value { get; }
+        internal PartitionKeyHash(UInt128[] values)
```
ealsur, line 44: nit: Mix/match of public/internal constructors. This class is already `internal`, marking properties or methods as `internal` does not yield any benefits.

Either use `private` if it is not meant to be used outside this type or `public`.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
```diff
@@ -35,12 +35,29 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+                stringBuilder.Append(PartitionKeyInternal.HexConvert.ToHex(hashBytes, 0, hashBytes.Length));
+            }
+
+            this.Value = stringBuilder.ToString();
+            this.values = values;
+        }
+
+        public string Value { get; }
+
+        private readonly UInt128[] values;
```
ealsur, line 59: nit: reorder. `private` local variables go before the constructor.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHashRange.cs
```diff
@@ -180,5 +180,26 @@ public override string ToString()
 
             return stringBuilder.ToString();
         }
+
+        public PartitionKeyHash GetMidPoint()
```
ealsur, line 184: Where is this method used in the PR code?

ealsur (commented):
Some nits and final comments

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHashRange.cs
NaluTripician: removing

ealsur (dismissed):

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
NaluTripician: removing method.

ealsur (dismissed):

ealsur (approved):

Microsoft.Azure.Cosmos/src/Query/Core/QueryClient/ContainerQueryProperties.cs
```diff
@@ -4,24 +4,29 @@
...
+    using System.Collections.Generic;
     using Microsoft.Azure.Documents;
+    using Microsoft.Azure.Documents.Routing;
 
     internal readonly struct ContainerQueryProperties
     {
         public ContainerQueryProperties(
             string resourceId,
-            string effectivePartitionKeyString,
+            List<Range<string>> effectivePartitionKeyRanges,
```
neildsh, line 15: >List [](http://example.com/codeflow?start=12&length=4)

`IReadOnlyList` please

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
```diff
@@ -35,12 +35,30 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+
+            this.Value = stringBuilder.ToString();
+            this.values = values;
+        }
+
+        public readonly static PartitionKeyHash None = new PartitionKeyHash(0);
+
+        public string Value { get; }
+
+        public readonly UInt128[] HashValues => this.values;
```
neildsh, line 61: >UInt128[] [](http://example.com/codeflow?start=24&length=9)

We should not expose this as a mutable array. Please consider changing this to be immutable

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -696,28 +684,6 @@ private static void SetTestInjectionPipelineType(InputParameters inputParameters
             }
         }
 
-        private static bool TryGetEpkProperty(
```
neildsh, line 699: >TryGetEpkProperty [](http://example.com/codeflow?start=28&length=17)

I am not sure this change is safe, since the earlier version was allowing the callers to pass in the pk in the properties and the new version just ignores this

NaluTripician: Added back check for Pk in properties, then convert the string to a range.

ealsur (dismissed):

NaluTripician:
@msftbot wait to merge until @ealsur and @neildsh approve

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Linq;
+    using System.Runtime.CompilerServices;
```
neildsh, line 10: >using System.Runtime.CompilerServices; [](http://example.com/codeflow?start=4&length=38)

is this required?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -635,67 +636,47 @@ internal static class CosmosQueryExecutionContextFactory
...
-                    containerQueryProperties.ResourceId,
-                    effectivePartitionKeyString,
-                    forceRefresh: false,
-                    trace);
-            }
-            else if (feedRangeInternal != null)
-            {
-                targetRanges = await queryClient.GetTargetPartitionKeyRangeByFeedRangeAsync(
+                List<Documents.Routing.Range<string>> effectiveRanges = new List<Documents.Routing.Range<string>>
+                    { new Documents.Routing.Range<string>(effectivePartitionKeyString, effectivePartitionKeyString, true, true) };
```
neildsh, line 651: >(effectivePartitionKeyString, effectivePartitionKeyString, true, true) [](http://example.com/codeflow?start=57&length=70)

Do we have to worry about the same prefix issue with hierarchical partition keys here?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -635,67 +636,47 @@ internal static class CosmosQueryExecutionContextFactory
...
             }
             else
             {
-                targetRanges = await queryClient.GetTargetPartitionKeyRangesAsync(
-                    resourceLink,
-                    containerQueryProperties.ResourceId,
-                    partitionedQueryExecutionInfo.QueryRanges,
-                    forceRefresh: false,
-                    trace);
+                targetRanges = feedRangeInternal != null
```
neildsh, line 661: >feedRangeInternal != null [](http://example.com/codeflow?start=31&length=25)

It would be easier to read if you pull this condition up as an `else if` since we are already in a conditional ladder

neildsh:
>     internal abstract class PartitionKeyHashRangeSplitterAndMerger

looks like this class is not used anywhere except in test code. Should this be moved to the tests folder?

---
Refers to: Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHashRangeSplitterAndMerger.cs:14 in 48557c9. [](commit_id = 48557c97f6b2bdbebc45ff74a614a29255d0d505, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHashRangeSplitterAndMerger.cs
```diff
@@ -82,6 +82,39 @@ public static SplitOutcome TrySplitRange(PartitionKeyHashRange partitionKeyHashR
...
+        {
+            SplitOutcome splitOutcome = PartitionKeyHashRangeSplitterAndMerger.TrySplitRange(
+                partitionKeyHashRange,
+                explicitSplitPoint,
+                out PartitionKeyHashRanges splitRanges);
+
+            return splitOutcome switch
+            {
+                SplitOutcome.Success => splitRanges,
+                _ => throw new ArgumentOutOfRangeException($"Unknown {nameof(SplitOutcome)}: {splitOutcome}."),
```
neildsh, line 95: >ArgumentOutOfRangeException [](http://example.com/codeflow?start=31&length=27)

This is not quite correct. There are other split outcomes possible. Please consider a custom `RangeSplitException` instead

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
```diff
@@ -35,12 +35,30 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+        public PartitionKeyHash(UInt128[] values)
+        {
+            StringBuilder stringBuilder = new StringBuilder();
+            foreach (UInt128 value in values)
+            {
+                stringBuilder.Append(value.ToString());
+            }
+
+            this.Value = stringBuilder.ToString();
+            this.values = values;
```
neildsh, line 54: This is very dangerous. `this.Value` is an immutable string, so far so good. However, `this.values` is a mutable array, which allows both of these members to go out of sync. Please use an immutable structure.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs
```diff
@@ -35,12 +35,30 @@ namespace Microsoft.Azure.Cosmos.Routing
     /// </example>
     internal readonly struct PartitionKeyHash : IComparable<PartitionKeyHash>, IEquatable<PartitionKeyHash>
     {
+        private readonly UInt128[] values;
```
neildsh, line 38: >UInt128[] values; [](http://example.com/codeflow?start=25&length=17)

Can you please add unit tests that exercise the new logic (the full array of UInt128[] values)? PartitionKeyHashRanges also has some significant new logict hat does not seem to be covered.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
NaluTripician: removed.

neildsh:
>             if (!UInt128.TryParse(value, out UInt128 uInt128))

This only parses a single UInt128. If we expect there to be multiple UINt128's , then this would have to change

---
Refers to: Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHash.cs:91 in 48557c9. [](commit_id = 48557c97f6b2bdbebc45ff74a614a29255d0d505, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyHashRanges.cs
```diff
@@ -132,9 +132,9 @@ public static PartitionKeyHashRanges Create(IEnumerable<PartitionKeyHashRange> p
             {
                 if (partitionKeyHashRange.StartInclusive.HasValue)
                 {
-                    if (partitionKeyHashRange.StartInclusive.Value.Value < minStart)
+                    if (partitionKeyHashRange.StartInclusive.Value.HashValues[0] < minStart)
```
neildsh, line 135: >partitionKeyHashRange.StartInclusive.Value.HashValues[0] < minStart [](http://example.com/codeflow?start=24&length=67)

Would this still work if we have multiple ranges with the same hash prefix, but differ in Value.HashValues[1]  ?

neildsh (approved):

ealsur (approved):

NaluTripician:
A separate PR will be out to move several files only used for testing to the Test library.

PR #3944, opened 2023-06-27
HttpTransport: Fixes HttpTimeoutPolicies to not accidentally suppress retries

FabianMeiswinkel:
# Pull Request Template

## Description

Current HttpTimeoutPolicies had too short MaxRetryTime - effectively suppressing any retries

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
@@ -438,8 +438,7 @@ namespace Microsoft.Azure.Cosmos
             DateTime startDateTimeUtc,
             IEnumerator<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> timeoutEnumerator)
         {
-            return (DateTime.UtcNow - startDateTimeUtc) > timeoutPolicy.MaximumRetryTimeLimit || // Maximum of time for all retries
-                !timeoutEnumerator.MoveNext(); // No more retries are configured
+            return !timeoutEnumerator.MoveNext(); // No more retries are configured
         }
 
         private async Task<HttpResponseMessage> ExecuteHttpHelperAsync(
--- a/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Azure.Cosmos
 
         public override string TimeoutPolicyName => HttpTimeoutPolicyControlPlaneRetriableHotPath.Name;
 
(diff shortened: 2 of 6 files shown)


Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -29,7 +29,7 @@ private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRe
 
         public override string TimeoutPolicyName => HttpTimeoutPolicyControlPlaneRetriableHotPath.Name;
```
sourabh1007, line 30: What about having a logic in `HttpTimeoutPolicy` which will have `MaximumRetryTimeLimit `as sum of all the time configured here i.e `TimeoutsAndDelays` + `GatewayTimeout`. Like this we don't have to worry about this calculation if we add new time out policy in future.

FabianMeiswinkel: I think the concept is intentionally decoupled - for example the NoRetry policy sets the MaxRetry to 0. I don't expect us to create many more timeout policies - and not too hard to get this right.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -29,7 +29,7 @@ private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRe
...
 
-        public override TimeSpan MaximumRetryTimeLimit => CosmosHttpClient.GatewayRequestTimeout;
+        public override TimeSpan MaximumRetryTimeLimit => CosmosHttpClient.GatewayRequestTimeout + TimeSpan.FromSeconds(7.5); // 0.5s + 5 s + 65s + 1s delay + 1s buffer
```
FabianMeiswinkel, line 32: This change is not strictly necessary - can revert it if preferred - the check whether start time and now are more than 65 seconds apart will succeed after the first timeouts - so, the original version would have worked.

FabianMeiswinkel: Just think it is a bit more transparent this way - but again can revert to minimize noise if that is preferred.

kirankumarkolli (approved):

ealsur (approved):

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -438,8 +438,7 @@ ValueTask<HttpRequestMessage> CreateRequestMessage()
             DateTime startDateTimeUtc,
```
ealsur, line 438: nit: We could remove the input parameters in the future

PR #3957, opened 2023-06-28
[Internal] Dependencies: Fixes dependabot alert for System.Linq.Dynamic.Core

ealsur:
`System.Linq.Dynamic.Core` was not really being used anywhere but we had the dependency on our Test projects.

There is now a security alert for this: https://github.com/Azure/azure-cosmos-dotnet-v3/security/dependabot/23 so might as well clean this up.


philipthomas-MSFT (approved):
ok. So just removing it altogether? LGTM,

kirankumarkolli (approved):

PR #3967, opened 2023-07-07
[Internal] Upgrade Resiliency: Adds Code to Enable Replica Validation Feature Through `CosmosClientOptions`

### LINKED ISSUE #3922: [Cosmos Service Upgrade - .NET] Replica Validation - Expose Replica Validation Feature as a Public Contract
Acceptance Criteria:

- Refactor the `CosmosClientOptions` to expose a `public` argument to enable/ disable the replica validation feature.

kundadebdatta:
# Pull Request Template

## Description

In reference to a recent conversation with the compute gateway team, we agreed to add an `internal` Boolean flag in `CosmosClientOptions` to dynamically enable the feature. This PR adds the flag in `CosmosClientOptions` to enable or disable the replica validation using the client options.

## Type of change

## Closing issues

To automatically close an issue: closes #3922

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -347,6 +347,23 @@ namespace Microsoft.Azure.Cosmos
         public bool? EnableContentResponseOnWrite { get; set; }
 
         /// <summary>
+        /// Gets or sets the advanced replica selection flag. The advanced replica selection logic keeps track of the replica connection
+        /// status, and based on status, it prioritizes the replicas which show healthy stable connections, so that the requests can be sent
+        /// confidently to the particular replica. This helps the cosmos client to become more resilient and effective to any connectivity issues.
+        /// The default value for this parameter is 'false'.
+        /// </summary>
+        /// <remarks>
+        /// <para>This is optimal for latency-sensitive workloads. Does not apply if <see cref="ConnectionMode.Gateway"/> is used.</para>
+        /// </remarks>
+        /// <seealso cref="CosmosClientBuilder.WithAdvancedReplicaSelectionEnabledForTcp()"/>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        bool EnableAdvancedReplicaSelectionForTcp { get; set; }
+
+        /// <summary>
         /// (Direct/TCP) Controls the amount of idle time after which unused connections are closed.
         /// </summary>
         /// <value>
@@ -758,6 +775,7 @@ namespace Microsoft.Azure.Cosmos
                 EnablePartitionLevelFailover = this.EnablePartitionLevelFailover,
                 PortReuseMode = this.portReuseMode,
                 EnableTcpConnectionEndpointRediscovery = this.EnableTcpConnectionEndpointRediscovery,
+                EnableAdvancedReplicaSelectionForTcp = this.EnableAdvancedReplicaSelectionForTcp,
                 HttpClientFactory = this.httpClientFactory,
                 ServerCertificateCustomValidationCallback = this.ServerCertificateCustomValidationCallback
             };
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -355,7 +355,6 @@ namespace Microsoft.Azure.Cosmos
         /// <remarks>
         /// <para>This is optimal for latency-sensitive workloads. Does not apply if <see cref="ConnectionMode.Gateway"/> is used.</para>
         /// </remarks>
-        /// <seealso cref="CosmosClientBuilder.WithAdvancedReplicaSelectionEnabledForTcp()"/>
         internal bool EnableAdvancedReplicaSelectionForTcp { get; set; }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -621,6 +621,19 @@ public CosmosClientBuilder WithContentResponseOnWrite(bool contentResponseOnWrit
...
         }
 
+        /// <summary>
+        /// Enables the advanced replica selection flag. The advanced replica selection logic keeps track of the replica connection status,
+        /// and based on status, it prioritizes the replicas which are connected to the backend, so that the requests can be sent
+        /// confidently to the particular replica. This helps the cosmos client to become more resilient and effictive to any connection
+        /// timeouts. The default value for this parameter is false.
+        /// </summary>
+        /// <returns>The <see cref="CosmosClientBuilder"/> object</returns>
+        internal CosmosClientBuilder WithAdvancedReplicaSelectionEnabledForTcp()
```
kirankumarkolli, line 631: Compute doesn't use builder API's is not needed.

kundadebdatta: Thanks! Removed the method in the latest commit.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -346,6 +346,17 @@ public ConnectionMode ConnectionMode
...
+        /// <summary>
+        /// Gets or sets the advanced replica selection flag. The advanced replica selection logic keeps track of the replica connection
+        /// status, and based on status, it prioritizes the replicas which show healthy stable connections, so that the requests can be sent
+        /// confidently to the particular replica. This helps the cosmos client to become more resilient and effective to any connectivity issues.
+        /// The default value for this parameter is 'false'.
+        /// </summary>
+        /// <remarks>
+        /// <para>This is optimal for latency-sensitive workloads. Does not apply if <see cref="ConnectionMode.Gateway"/> is used.</para>
+        /// </remarks>
+        internal bool EnableAdvancedReplicaSelectionForTcp { get; set; }
```
kirankumarkolli, line 358: Is explicit differentiation about opted/not vs default behavior for compute? 
If so one option is to model it as nullable. Thoughts?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -233,7 +232,6 @@ internal partial class DocumentClient : IDisposable, IAuthorizationTokenProvider
 
             this.Initialize(serviceEndpoint, connectionPolicy, desiredConsistencyLevel);
             this.initTaskCache = new AsyncCacheNonBlocking<string, bool>(cancellationToken: this.cancellationTokenSource.Token);
-            this.isReplicaAddressValidationEnabled = ConfigurationManager.IsReplicaAddressValidationEnabled();
```
kirankumarkolli, line 236: Is CosmosCLientOptions needs to default to ConfigurationManager.IsReplicaAddressValidationEnabled()?

ealsur: Did we ship a version that supported setting the environment variable and is anyone using it? If yes, we might want to have backward compatibility?

kundadebdatta: `CosmosClientOptions` does not need to default to `ConfigurationManager.IsReplicaAddressValidationEnabled()`. This was an earlier implementation, which was reading the environment variable to set the replica validation flag. Since we never shipped this feature with any version of the SDK, we can safely remove this and rely on the `CosmosClientOptions`.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Well, not setting this flag `EnableAdvancedReplicaSelectionForTcp ` will automatically make the SDK to opt out from replica validation. In order to opt the feature, Compute doesn't need to specify any value for this field and the default value for this flag is always `false`.

I think this brings more simplicity, since we are anyhow not going to make this flag to public.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
ealsur: > Since we never shipped this feature with any version of the SDK, we can safely remove this 

Then yeah, no issues from my side.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: You can make the SDK default to disabled if the `bool?` has not value by just capturing: 

```
bool isEnabled = EnableAdvancedReplicaSelectionForTcp.HasValue && EnableAdvancedReplicaSelectionForTcp.Value;
```

I think Kiran's point is that semantically it's always better to make settings `bool?` like `EnableContentResponseOnWrite` because it allows us (for whatever reason) to tell apart the scenario where the property's setter was not touched (use default) and when the user did set some preference (either true or false). In this particular case there might not be a special case for not touched, but for consistency's sake, if we are using `bool?` for other settings, might as well use it here.

kirankumarkolli: Its an optional property and hence the conversation. 

If compute if always set it explicitly (true/false) then for current requirements its makes sense.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
kirankumarkolli: How do external CX enable/disable this feature?

kundadebdatta:
This PR is no longer required as an [alternate PR](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3974) has been merged to `master` to address the issue.

PR #3974, opened 2023-07-10
[Internal] Upgrade Resiliency: Refactors Code to Enable Replica Validation Feature Through `CosmosClientOptions` And Environment Variable

### LINKED ISSUE #3922: [Cosmos Service Upgrade - .NET] Replica Validation - Expose Replica Validation Feature as a Public Contract

kundadebdatta:
# Pull Request Template

## Description

In reference to a recent conversation with the compute gateway team, we agreed to add an `internal` Boolean flag in `CosmosClientOptions` to programmatically enable (or disable) the feature. This PR adds the flag in `CosmosClientOptions` to enable or disable the replica validation using the cosmos client options. Additionally, it keeps the ability for the external customers, to use the environment variable `AZURE_COSMOS_REPLICA_VALIDATION_ENABLED` to enable or disable the feature. 

## Type of change

## Closing issues

To automatically close an issue: closes #3922

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -347,6 +347,17 @@ namespace Microsoft.Azure.Cosmos
         public bool? EnableContentResponseOnWrite { get; set; }
 
         /// <summary>
+        /// Gets or sets the advanced replica selection flag. The advanced replica selection logic keeps track of the replica connection
+        /// status, and based on status, it prioritizes the replicas which show healthy stable connections, so that the requests can be sent
+        /// confidently to the particular replica. This helps the cosmos client to become more resilient and effective to any connectivity issues.
+        /// The default value for this parameter is 'false'.
+        /// </summary>
+        /// <remarks>
+        /// <para>This is optimal for latency-sensitive workloads. Does not apply if <see cref="ConnectionMode.Gateway"/> is used.</para>
+        /// </remarks>
+        internal bool? EnableAdvancedReplicaSelectionForTcp { get; set; }
+
+        /// <summary>
         /// (Direct/TCP) Controls the amount of idle time after which unused connections are closed.
         /// </summary>
         /// <value>
@@ -758,6 +769,7 @@ namespace Microsoft.Azure.Cosmos
                 EnablePartitionLevelFailover = this.EnablePartitionLevelFailover,
                 PortReuseMode = this.portReuseMode,
                 EnableTcpConnectionEndpointRediscovery = this.EnableTcpConnectionEndpointRediscovery,
+                EnableAdvancedReplicaSelectionForTcp = this.EnableAdvancedReplicaSelectionForTcp.HasValue && this.EnableAdvancedReplicaSelectionForTcp.Value,
                 HttpClientFactory = this.httpClientFactory,
                 ServerCertificateCustomValidationCallback = this.ServerCertificateCustomValidationCallback
             };
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -769,7 +769,7 @@ namespace Microsoft.Azure.Cosmos
                 EnablePartitionLevelFailover = this.EnablePartitionLevelFailover,
                 PortReuseMode = this.portReuseMode,
                 EnableTcpConnectionEndpointRediscovery = this.EnableTcpConnectionEndpointRediscovery,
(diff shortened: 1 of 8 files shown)


Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
```diff
@@ -32,13 +32,20 @@ public static T GetEnvironmentVariable<T>(string variable, T defaultValue)
...
         /// <returns>A boolean flag indicating if replica validation is enabled.</returns>
-        public static bool IsReplicaAddressValidationEnabled()
+        public static bool IsReplicaAddressValidationEnabled(
+            ConnectionPolicy connectionPolicy)
         {
             bool replicaValidationDefaultValue = false;
 #if PREVIEW
             replicaValidationDefaultValue = true;
 #endif
+            if (connectionPolicy != null
```
kundadebdatta, line 44: Does it make sense to use `#if COSMOSCLIENT` check here ? Something like:

```
        public static bool IsReplicaAddressValidationEnabled(
            ConnectionPolicy connectionPolicy)
        {
            bool replicaValidationDefaultValue = false;
#if PREVIEW
            replicaValidationDefaultValue = true;
#endif

#if COSMOSCLIENT
            return connectionPolicy != null && connectionPolicy.EnableAdvancedReplicaSelectionForTcp;
#else
            return ConfigurationManager
                    .GetEnvironmentVariable(
                        variable: ConfigurationManager.ReplicaConnectivityValidationEnabled,
                        defaultValue: replicaValidationDefaultValue);
#endif
        }
```

AFAIR - Compute uses `COSMOSCLIENT` pre-processor to identify themselves.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -346,6 +346,17 @@ public ConnectionMode ConnectionMode
...
+        /// <summary>
+        /// Gets or sets the advanced replica selection flag. The advanced replica selection logic keeps track of the replica connection
+        /// status, and based on status, it prioritizes the replicas which show healthy stable connections, so that the requests can be sent
+        /// confidently to the particular replica. This helps the cosmos client to become more resilient and effective to any connectivity issues.
+        /// The default value for this parameter is 'false'.
+        /// </summary>
+        /// <remarks>
+        /// <para>This is optimal for latency-sensitive workloads. Does not apply if <see cref="ConnectionMode.Gateway"/> is used.</para>
+        /// </remarks>
+        internal bool? EnableAdvancedReplicaSelectionForTcp { get; set; }
```
kirankumarkolli, line 358: 'Advanced' is not mean much for the context right? 

Some names we went over in the past PRs are much better.

kundadebdatta: Me and @ealsur had a conversation around it a week back and we both agreed that `PrioritizeHealthyReplicas` OR `PrioritizeHealthyEndpoints` are probably not a good choice for naming the attribute either. 

With the name `EnableAdvancedReplicaSelectionForTcp`, I agree that it doesn't provide much context however the description covers it right? I am open to make the changes, but wanted to hear @ealsur 's thought on that.

ealsur: The key thing is that **it should** contain "Tcp" and should contain "Enable", **should not** contain "Timeout" or "Upgrade" or "Healthy". 

Whatever name we pick, the GA version will not exist, so really I don't think we should block/spend too much time cranking our heads for a name that will eventually be removed, but that's my humble opinion :)

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
ealsur: We set COSMOSCLIENT when building ourselves:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/51c9d9c6fda1656cb1d3a28d2bb23df3e1ce43ad/Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj#L173

kundadebdatta: Got you! Then the current logic should suffice. No need to over complicate it. I will resolve the comment.

ealsur (approved):
LGTM, I don't have much concern on the name, the feature will eventually be enabled by default and when GAed it will not have a public API toggle

kirankumarkolli (approved):

PR #3997, opened 2023-07-20
[Preview] Integrated cache: Adds ShardKey to DedicatedGatewayRequestOptions

jiajunpeng-msft:
# Pull Request Template

## Description

Currently, integrated cache is not distributed. Requests will randomly land on any of the Dedicated Gateway instances, which causes inconsistency.

We are introducing a new "RequestOption" called "ShardKey" to specify the Dedicated Gateway request shard key. Shard key will be used to route the request to the correlated instance.

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
+++ b/Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Azure.Cosmos
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos.Core.Trace;
     using Microsoft.Azure.Cosmos.Handlers;
     using Microsoft.Azure.Cosmos.Tracing.TraceData;
     using Microsoft.Azure.Documents;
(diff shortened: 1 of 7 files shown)


Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -317,6 +317,16 @@ private static async Task<Stream> BufferContentIfAvailableAsync(HttpResponseMess
                             requestMessage.Headers.TryAddWithoutValidation(key, request.Headers[key]);
                         }
                     }
+
+                    if (key.Equals(HttpConstants.HttpHeaders.DedicatedGatewayShardKey, StringComparison.OrdinalIgnoreCase) && request.Headers[key] != null)
+                    {
+                        string useHttp2String = Environment.GetEnvironmentVariable("UseHttp2");
```
sourabh1007, line 323: you can use this to read environment variable https://github.com/Azure/azure-cosmos-dotnet-v3/blob/5ec7b4bcd7e2e0698736cd374f2196682227c913/Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs#L18

jiajunpeng-msft: Thanks for sharing this! Changed to use `ConfigurationManager`.

sourabh1007 (dismissed):

Microsoft.Azure.Cosmos/src/Resource/Settings/DedicatedGatewayRequestOptions.cs
```diff
@@ -57,6 +57,42 @@ public class DedicatedGatewayRequestOptions
...
+        /// 
+        /// // For QueryRequestOptions
+        /// QueryRequestOptions requestOptions = new QueryRequestOptions
+        /// {
+        ///     DedicatedGatewayRequestOptions = dedicatedGatewayRequestOptions
+        /// };
+        /// ]]>
+        /// </code>
+        /// </example>
+#if PREVIEW
```
kirankumarkolli, line 89: When is backend ready for public preview? 
Ideally this needs to be public after-wards. 

Private preview: As discussed before public contract is not a requirement and can be worked around through using custom-headers right? 

/cc: @jcocchi

kirankumarkolli (changes requested):
Preview contracts are for public previews. Waiting for clarifiation

Microsoft.Azure.Cosmos/src/Resource/Settings/DedicatedGatewayRequestOptions.cs
jcocchi: Sharding public preview is several months out and our private preview will be limited, at least in the initial phase. Using custom headers makes sense for now so we don't confuse existing customers with new options that aren't widely available yet

jiajunpeng-msft: This makes sense to me. I removed ShardKey request option and modified the http2 to be used when it's Dedicated Gateway request and has the environment variable set.

Let me know if this makes sense to you @kirankumarkolli @jcocchi.

kirankumarkolli:

PR #4001, opened 2023-07-21
[Query] Adds public backend metrics property to Diagnostics

### LINKED ISSUE #3357: Exposing QueryMetrics
In SDK V2, the query response contained the actual QueryMetrics object which allowed us to aggregate the metrics of multiple calls for more relevant logging, especially on things like paged queries. Now, the only way to get QueryMetrics information is through the ToString() method of CosmosDiagnostics. This unfortunately removes the consumer's ability to aggregate metrics as they see fit.

Changing the QueryMetrics object to public would allow us to actually have full control over the metrics, instead of piecing them together manually from the serialized string.

If the QueryMetrics class itself cannot be made public, then perhaps exposing a new Metrics object that can still be similarly aggregated (i.e., + operator is overloaded) and has all the fields that were available in V2 could work.

### LINKED ISSUE #3837: Modify querymetrics to be exposed, not a private sealed class.  Maybe .GetQueryMetrics?
QueryMetrics provides very useful information, but here is how I have to get it using the newest cosmos SDK.

```
traces
| where message has "Retrieved Document Count"
| extend Query_Metrics_ = tostring(parse_json(tostring(parse_json(tostring(parse_json(tostring(parse_json(tostring(parse_json(tostring(parse_json(message).children))[1].children))[1].children))[0].children))[0].data)).["Query Metrics"])
```
Look at how many levels deep that is!!!

I can already get Diagnostics like this:

```csharp
var result = await this.container.UpsertItemAsync(entity, entity.ResourceIdentity.GetPartitionKey());
                    this.logger?.LogDebug("{operation} of record {entity} elapsed time {elapsed} query incurred charge of {resourceUnits} RUs",
                        "Update",
                        entity.Id,
                        result.Diagnostics?.GetClientElapsedTime(),
                        result.RequestCharge);
```

Simply expose QueryMetrics as well, via `Diagnostics.GetQueryMetrics()`, instead of having it be a very, very, very deeply nested property.

I began to just import the QueryMetrics and other types into my own project and write an extension method to do this, but the code style doesn't mesh well with our project and style-cop was causing issues.

I also tried to write an extension method to handle the conversion from Diagnostics.ToString(), but it looked really gnarly.

**Additional context**

Maya-Painter:
## Description

These changes expose new strongly typed metrics objects so that users can easily access query metrics via Diagnostics.

Users will be able to retrieve metrics using a new GetQueryMetrics() function.

example:
```
FeedResponse<T> response = await feedIterator.ReadNextAsync();
ServerSideCumulativeMetrics metrics = response.Diagnostics.GetQueryMetrics();
```

New public classes are the following:

image

This PR also includes a refactoring of internal metrics classes.

## Type of change

## Closing issues
closes #3357
closes #3837


Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendAccumulatedMetrics.cs
```diff
@@ -0,0 +1,140 @@
...
+namespace Microsoft.Azure.Cosmos
+{
+    using System;
+    using Microsoft.Azure.Cosmos.Query.Core.Metrics;
+
+    /// <summary>
+    /// Exposed metrics received for queries from the backend.
+    /// </summary>
+    public sealed class BackendAccumulatedMetrics
+    {
```
adityasa, line 13: >{ [](http://example.com/codeflow?start=4&length=1)

To allow for accumulation of metrics across roundtrips, you will need another public parameterized constructor.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendAccumulatedMetrics.cs
```diff
@@ -0,0 +1,140 @@
...
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos
+{
+    using System;
+    using Microsoft.Azure.Cosmos.Query.Core.Metrics;
+
+    /// <summary>
+    /// Exposed metrics received for queries from the backend.
+    /// </summary>
+    public sealed class BackendAccumulatedMetrics
```
adityasa, line 12: >BackendAccumulatedMetrics [](http://example.com/codeflow?start=24&length=25)

Just BackendMetrics?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendAccumulatedMetrics.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// <summary>
+        /// Gets the total query runtime execution time in the Azure Cosmos DB service.
+        /// </summary>
+        public TimeSpan QueryEngineExecutionTime { get; }
+
+        /// <summary>
+        /// String representation of the QueryBackendMetrics.
+        /// </summary>
+        /// <returns>Metric text</returns>
+        public override string ToString()
```
adityasa, line 134: I would keep this logic out of this class.
An option is to move this into a new static FormatTrace method (optionally extension method) inside a TraceUtil or some relevant class.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -49,11 +50,42 @@ public override IReadOnlyList<(string regionName, Uri uri)> GetContactedRegions(
...
+            this.WalkTraceTreeForQueryMetrics(this.Value, ref queryMetrics);
+            return new BackendAccumulatedMetrics(queryMetrics);
+        }
+
         internal bool IsGoneExceptionHit()
         {
             return this.WalkTraceTreeForGoneException(this.Value);
         }
 
+        private bool WalkTraceTreeForQueryMetrics(ITrace currentTrace, ref QueryMetrics queryMetrics)
```
adityasa, line 65: >ref QueryMetrics queryMetrics [](http://example.com/codeflow?start=71&length=29)

ref works, but a better pattern is to provide an accumulator instance that just accumulates individual QueryMetrics to a list and exposes a method to return the result by summing those all up.

adityasa, line 65: >bool [](http://example.com/codeflow?start=16&length=4)

I don't see the return value being used. Is this needed?

adityasa, line 65: >WalkTraceTreeForQueryMetrics [](http://example.com/codeflow?start=21&length=28)

Ideally this method will be a private method inside accumulator.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendAccumulatedMetrics.cs
adityasa: I see two options:

Option1 - Expose 2 public classes with a potential for 3rd in future:
public class QueryMetrics
{
   public BackendMetrics BackendMetrics { get; }
   // Evantually public, assuming this is a good/ok idea
   // for now internal
   ClientMetrics ClientMetrics { get; }
}

Option2 - Only expose backend metrics, with a different name:
public class QueryMetrics
{
  public int RetrievedDocumentCount {get;}
  . . .
  public TimeSpan DocumentWriteTime {get;}
}

I think we should go with Option1 for now and seek feedback from Samer & broader sdk team.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetrics.cs
```diff
@@ -155,7 +152,7 @@ public static BackendMetrics ParseFromDelimitedString(string delimitedString)
             return backendMetrics;
         }
 
-        public ref struct Accumulator
+        internal ref struct Accumulator
```
adityasa, line 155: >Accumulator [](http://example.com/codeflow?start=28&length=11)

Kindly move to a separate file as top level (internal) class.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        public void Accumulate(BackendMetrics backendMetrics)
+        {
+            this.TotalTime += backendMetrics.TotalTime;
+            this.RetrievedDocumentCount += backendMetrics.RetrievedDocumentCount;
+            this.RetrievedDocumentSize += backendMetrics.RetrievedDocumentSize;
+            this.OutputDocumentCount += backendMetrics.OutputDocumentCount;
+            this.OutputDocumentSize += backendMetrics.OutputDocumentSize;
+            this.IndexHitRatio = ((this.OutputDocumentCount * this.IndexHitRatio) + (backendMetrics.OutputDocumentCount * backendMetrics.IndexHitRatio)) / (this.RetrievedDocumentCount + backendMetrics.RetrievedDocumentCount);
+            this.QueryPreparationTimesAccumulator.Accumulate(backendMetrics.QueryPreparationTimes);
```
Maya-Painter, line 62: @adityasa I'm not sure if I'm missing something obvious here but nested accumulators (with the new pattern of saving state on the accumulator itself rather than returning a new accumulator) do not seem to work.  On this line the QueryPreperationTimesAccumulator.Accumulate function is ran with a non-zero QueryPreperationTime, but the QueryPreperationTimesAccumulator properties are not updated.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/QueryPreparationTimesAccumulator.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        public QueryPreparationTimesAccumulator(TimeSpan queryCompliationTime, TimeSpan logicalPlanBuildTime, TimeSpan physicalPlanBuildTime, TimeSpan queryOptimizationTime)
+        {
+            this.QueryCompilationTime = queryCompliationTime;
+            this.LogicalPlanBuildTime = logicalPlanBuildTime;
+            this.PhysicalPlanBuildTime = physicalPlanBuildTime;
+            this.QueryOptimizationTime = queryOptimizationTime;
+        }
+
+        public TimeSpan QueryCompilationTime { get; set; }
```
adityasa, line 19: >public [](http://example.com/codeflow?start=8&length=6)

private please. Same for other fields/accumulators.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/QueryPreparationTimesAccumulator.cs
```diff
@@ -0,0 +1,46 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Query.Core.Metrics
+{
+    using System;
+
+    internal struct QueryPreparationTimesAccumulator
```
adityasa, line 9: >struct [](http://example.com/codeflow?start=13&length=6)

class

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/QueryPreparationTimesAccumulator.cs
adityasa, line 19: >{ get; set; } [](http://example.com/codeflow?start=45&length=13)

Strictly speaking, no need to ruse properties for private only fields. It's less code (both written and generated) if private mutable fields are used.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/QueryPreparationTimesAccumulator.cs
```diff
@@ -0,0 +1,46 @@
...
+                throw new ArgumentNullException(nameof(queryPreparationTimes));
+            }
+
+            this.QueryCompilationTime += queryPreparationTimes.QueryCompilationTime;
+            this.LogicalPlanBuildTime += queryPreparationTimes.LogicalPlanBuildTime;
+            this.PhysicalPlanBuildTime += queryPreparationTimes.PhysicalPlanBuildTime;
+            this.QueryOptimizationTime += queryPreparationTimes.QueryOptimizationTime;
+        }
+
+        public static QueryPreparationTimes ToQueryPreparationTimes(QueryPreparationTimesAccumulator accumulator)
```
adityasa, line 37: I would use instance method instead (drop static and drop input parameter).

adityasa, line 37: >ToQueryPreparationTimes [](http://example.com/codeflow?start=44&length=23)

GetQueryPreparationTimes?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
adityasa: Nothing obvious jumps out to me. Only things I could think of, is either input QueryPreparationTimes are 0 (which you already covered) or there's a bug in the accumulator (I can't see any). Other than debugging this, stepping into the Accumulate method and stepping out all the while ensuring the state of the object is consistent before/after, nothing obvious I can point to at the moment. 
Not sure if my other comment has anything to do with this issue.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
```diff
@@ -0,0 +1,111 @@
...
+        public double IndexHitRatio { get; set; }
+        public QueryPreparationTimesAccumulator QueryPreparationTimesAccumulator { get; set; }
+        public TimeSpan IndexLookupTime { get; set; }
+        public TimeSpan DocumentLoadTime { get; set; }
+        public RuntimeExecutionTimesAccumulator RuntimeExecutionTimesAccumulator { get; set; }
+        public TimeSpan DocumentWriteTime { get; set; }
+        public TimeSpan VMExecutionTime { get; set; }
+
+        public void Accumulate(BackendMetrics backendMetrics)
+        {
```
adityasa, line 55: This input parameter type strikes a bit odd and has circular dependency.
The way to get BackendMetrics is to use BackendMetricsAccumulator.Accumulate.
Accumulate method requires an instance of type BackendMetrics as input parameter.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
```diff
@@ -0,0 +1,111 @@
...
+            if (currentTrace == null)
+            {
+                return;
+            }
+
+            foreach (object datum in currentTrace.Data.Values)
+            {
+                if (datum is QueryMetricsTraceDatum queryMetricsTraceDatum)
+                {
+                    accumulator.Accumulate(queryMetricsTraceDatum.QueryMetrics.BackendMetrics);
```
adityasa, line 98: >queryMetricsTraceDatum.QueryMetrics.BackendMetrics [](http://example.com/codeflow?start=43&length=50)

I am curious as to where does this get a valid (non-zero) value from.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/QueryPreparationTimesAccumulator.cs
Maya-Painter: updated

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/QueryPreparationTimesAccumulator.cs
Maya-Painter: Sorry, what do you mean by "reuse properties" in this case?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
Maya-Painter: It does seem like somewhat of a circular dependency, but I think the parameter makes sense if we want to be accumulating the state per BackendMetric. What other alternatives do we have here?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
Maya-Painter: It ended up being solved by making the accumulators structs instead of classes. Since they were being passed between different functions it would make sense that the need to be ref types and not value types.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
Maya-Painter: As in the mechanism that builds the trace?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendAccumulatedMetrics.cs
Maya-Painter: Moved out of this class and added as an extension method in a new BackendMetricUtils class.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsUtils.cs
```diff
@@ -0,0 +1,13 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.Query.Core.Metrics
+{
+    internal static class BackendMetricsUtils
+    {
+        public static string FormatTrace(this BackendMetrics backendMetrics)
```
Maya-Painter, line 8: @adityasa I'm fine with this but I think there's an argument to be made that customers would expect ToString() to be a usable serialization of the contents of the class. Existing documentation https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query-metrics-performance#:~:text=//%20At%20this%20point%20you%20have%20QueryMetrics%20which%20you%20can%20serialize%20using%20.ToString() implies this as well, although this will need to be updated in any case.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosTraceDiagnosticsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+                    ToDoActivity response = iter.First();
+                    Assert.AreEqual(find.id, response.id);
+                }
+
+                BackendMetrics backendMetricsFromDiagnostics = iter.Diagnostics.GetQueryMetrics();
+                bool tryParseResult = BackendMetrics.TryParseFromDelimitedString(iter.Headers.QueryMetricsText, out BackendMetrics backendMetricsFromHeaders);
+                Assert.IsTrue(tryParseResult);
+
+                headerMetricsAccumulator.Accumulate(backendMetricsFromHeaders);
+                Assert.IsTrue(headerMetricsAccumulator.GetBackendMetrics().FormatTrace() == backendMetricsFromDiagnostics.FormatTrace());
```
Maya-Painter, line 91: @adityasa Trying to come up with more ways to test this.  The FeedResponse.Headers.QueryMetricsText property only seems to be the first of multiple backend calls (1 for each partition) so it isn't an accurate accumulation of metrics for multi partition queries.  I found this surprising honestly, it doesn't seem like very useful if it only provides the stats for one partition of many.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetSDKAPI.json
```diff
@@ -3124,6 +3124,11 @@
           "Attributes": [],
           "MethodInfo": "Int32 GetFailedRequestCount();IsAbstract:False;IsStatic:False;IsVirtual:True;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
         },
+        "Microsoft.Azure.Cosmos.Query.Core.Metrics.BackendMetrics GetQueryMetrics()": {
```
kirankumarkolli, line 3127: Are these meant for public consumption? 

Public API needs to go-through API review process. Can you please schedule one with SDK crew.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
Maya-Painter: Updated to maintain member list in accumulator and only accumulate if requested.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetrics.cs
```diff
@@ -124,114 +132,30 @@ sealed class BackendMetrics
...
-        }
-
-        public static BackendMetrics CreateFromIEnumerable(IEnumerable<BackendMetrics> backendMetricsEnumerable)
+        internal static BackendMetrics CreateFromIEnumerable(IEnumerable<BackendMetrics> backendMetricsEnumerable)
```
adityasa, line 135: >CreateFromIEnumerable [](http://example.com/codeflow?start=39&length=21)

Nit, simply, Create?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
```diff
@@ -0,0 +1,101 @@
...
+    using Microsoft.Azure.Cosmos.Tracing.TraceData;
+
+    internal class BackendMetricsAccumulator
+    {
+        public BackendMetricsAccumulator()
+        {
+            this.BackendMetricsList = new List<BackendMetrics>();
+        }
+
+        private readonly List<BackendMetrics> BackendMetricsList;
```
adityasa, line 19: >BackendMetricsList [](http://example.com/codeflow?start=46&length=18)

Field names use camelCasing.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
```diff
@@ -0,0 +1,101 @@
...
+                throw new ArgumentNullException(nameof(backendMetrics));
+            }
+
+            this.BackendMetricsList.Add(backendMetrics);
+        }
+
+        public BackendMetrics GetBackendMetrics()
+        {
+            TimeSpan totalTime = default;
+            long retrievedDocumentCount = default;
```
adityasa, line 34: >default [](http://example.com/codeflow?start=42&length=7)

0

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+                throw new ArgumentNullException(nameof(backendMetrics));
+            }
+
+            this.BackendMetricsList.Add(backendMetrics);
+        }
+
+        public BackendMetrics GetBackendMetrics()
+        {
+            TimeSpan totalTime = default;
```
adityasa, line 33: >default [](http://example.com/codeflow?start=33&length=7)

It's better to use an explicit default value rather than rely on C#'s default. Also clearly communicates to the reader.

adityasa: Also in other places in this method.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsUtils.cs
adityasa: It's not advisable for user code to take a dependency on QueryMetrics' (or for that matter most objects') specific serialized form. Also, I cannot think of a scenario where this is desirable and needs to be implemented by the sdk (rather than user code).

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ClientSideMetricsAccumulator.cs
```diff
@@ -0,0 +1,48 @@
...
+    using System.Linq;
+
+    internal class ClientSideMetricsAccumulator
+    {
+        public ClientSideMetricsAccumulator()
+        {
+            this.ClientSideMetricsList = new List<ClientSideMetrics>();
+        }
+
+        private readonly List<ClientSideMetrics> ClientSideMetricsList;
```
adityasa, line 17: >ClientSideMetricsList [](http://example.com/codeflow?start=49&length=21)

Same as before.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ClientSideMetricsAccumulator.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                throw new ArgumentNullException(nameof(clientSideMetrics));
+            }
+
+            this.ClientSideMetricsList.Add(clientSideMetrics);
+        }
+
+        public ClientSideMetrics GetClientSideMetrics()
+        {
+            long retries = default;
```
adityasa, line 31: >default [](http://example.com/codeflow?start=27&length=7)

Same as before.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ClientSideMetricsAccumulator.cs
```diff
@@ -0,0 +1,48 @@
...
+            double requestCharge = default;
+            IEnumerable<FetchExecutionRange> fetchExecutionRanges = default;
```
adityasa, line 33: >IEnumerable [](http://example.com/codeflow?start=12&length=11)

List

adityasa, line 33: >default [](http://example.com/codeflow?start=68&length=7)

new List

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ClientSideMetricsAccumulator.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            foreach (ClientSideMetrics clientSideMetrics in this.ClientSideMetricsList)
+            {
+                retries += clientSideMetrics.Retries;
+                requestCharge += clientSideMetrics.RequestCharge;
+                fetchExecutionRanges = (fetchExecutionRanges ?? Enumerable.Empty<FetchExecutionRange>()).Concat(clientSideMetrics.FetchExecutionRanges);
```
adityasa, line 39: >fetchExecutionRanges ??  [](http://example.com/codeflow?start=40&length=24)

Not needed after correct initialization above.

adityasa, line 39: >Concat [](http://example.com/codeflow?start=105&length=6)

AddRange (when the type is changed to IEnumerable

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ClientSideMetricsAccumulator.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            return new ClientSideMetrics(
+                retries: retries,
+                requestCharge: requestCharge,
+                fetchExecutionRanges: fetchExecutionRanges);
```
adityasa, line 45: >fetchExecutionRanges [](http://example.com/codeflow?start=38&length=20)

As an alternative to comments above, you can do fetchExecutionRanges.ToList here.
IEnumerable is awesome. However, since it's lazy evaluated, it can have unintended consequences for e.g. getting different set of values at each enumeration time (if the source is modified).

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ClientSideMetricsAccumulator.cs
adityasa: Or Enumerable.Empty

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/IndexUtilizationInfoAccumulator.cs
```diff
@@ -0,0 +1,53 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Query.Core.Metrics
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    internal class IndexUtilizationInfoAccumulator
```
adityasa, line 11: >IndexUtilizationInfoAccumulator [](http://example.com/codeflow?start=19&length=31)

Same comments as the other class.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetricsAccumulator.cs
adityasa, line 19: FIelds would be declared before the constructor.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ClientSideMetricsAccumulator.cs
Maya-Painter: Good to know! Updated to use List here

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/BackendMetrics.cs
```diff
@@ -10,19 +10,12 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Metrics
...
     /// </summary>
-#if INTERNAL
-#pragma warning disable SA1600
-#pragma warning disable CS1591
-    public
-#else
-    internal
-#endif
-    sealed class BackendMetrics
+    public sealed class BackendMetrics
```
sboshra, line 13: >BackendMetrics [](http://example.com/codeflow?start=24&length=14)

For public interface, I'm not too sure about Backend in the name here. Since we have ClientSideMetrics, should this be called ServerSideMetrics?

Maya-Painter: Makes sense, updated.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
```diff
@@ -0,0 +1,94 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Query.Core.Metrics
```
ealsur, line 5: Consider not adding a new public namespace and use what it currently exists.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    using System;
+
+    /// <summary>
+    /// internal implementation of metrics received for queries from the backend.
+    /// </summary>
+    public sealed class ServerSideMetrics
```
ealsur, line 12: If this is sealed and the constructor is internal, if a user wants to Mock a FeedResponse and Mock the Diagnostics, is this approach compatible?

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -49,6 +49,13 @@ public override IReadOnlyList<(string regionName, Uri uri)> GetContactedRegions(
             return this.Value?.Summary?.RegionsContacted;
         }
 
+        public override ServerSideMetrics GetQueryMetrics()
+        {
+            ServerSideMetricsAccumulator accumulator = new ServerSideMetricsAccumulator();
```
ealsur, line 54: Should this be cached? To avoid walking the entire trace again and again if the user accesses this property (for whatever reason) many times?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosTraceDiagnosticsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+                FeedResponse<ToDoActivity> iter = await feedIterator.ReadNextAsync();
+                Assert.IsTrue(iter.Count() <= 1);
+                if (iter.Count() == 1)
+                {
+                    found = true;
+                    ToDoActivity response = iter.First();
+                    Assert.AreEqual(find.id, response.id);
+                }
+
+                ServerSideMetrics serverSideMetricsFromDiagnostics = iter.Diagnostics.GetQueryMetrics();
```
ealsur, line 86: Is this something that can be added to an existing Query Emulator test? There are probably already Query tests that perform simple queries like this? It would reduce the overhead of creating yet another client/database/container

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetSDKAPI.json
Maya-Painter: Scheduled to discuss Monday 8/7 during the SDK sync slot

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
Maya-Painter: Updated all public classes from `Microsoft.Azure.Cosmos.Query.Core.Metrics` to `Microsoft.Azure.Cosmos` namespace

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
Maya-Painter: Ah I didn't consider mocking.  In that case we could make the constructors for ServerSideMetrics, QueryPreparationTimes, and RuntimeExecutionTimes public?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
Maya-Painter: Ah I didn't consider mocking. In that case we could make the constructors for ServerSideMetrics, QueryPreparationTimes, and RuntimeExecutionTimes public?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosTraceDiagnosticsTests.cs
Maya-Painter: Makes sense. Added to existing tests in CosmosItemTests.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
ealsur: Right, imagine you'd have to mock this. You probably need a public constructor that it's empty and make the properties virtual. For RuntimeExecutionTimes and QueryPreparationTimes is the same.  Similar to what `ResponseMessage` has https://github.com/Azure/azure-cosmos-dotnet-v3/blob/ebd1b91a26547721ef232ef597d878144889b80b/Microsoft.Azure.Cosmos/src/Handler/ResponseMessage.cs#L22

Maya-Painter: Do we want a public default constructor and one that allows the user to set property values (like in ResponseMessage), or just a default constructor?

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
Maya-Painter: Updated to cache the ServerSideMetrics when it's retrieved.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -9,12 +9,14 @@ namespace Microsoft.Azure.Cosmos.Diagnostics
...
     using System.Text;
     using Microsoft.Azure.Cosmos.Json;
+    using Microsoft.Azure.Cosmos.Query.Core.Metrics;
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Cosmos.Tracing.TraceData;
     using static Microsoft.Azure.Cosmos.Tracing.TraceData.ClientSideRequestStatisticsTraceDatum;
 
     internal sealed class CosmosTraceDiagnostics : CosmosDiagnostics
     {
+        private ServerSideAccumulatedMetrics cachedServerSideAccumulatedMetrics { get; set; }
```
adityasa, line 19: >{ get; set; } [](http://example.com/codeflow?start=80&length=13)

I would keep this to field and remove {get;set;} declaration that makes it a property (since it's not accessed from outside the class.

adityasa, line 19: Newline after please.

adityasa, line 19: private readonly Laxy<ServerSideAccumulatedMetrics> accumulatedMetrics;

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -30,6 +32,7 @@ public CosmosTraceDiagnostics(ITrace trace)
             }
 
             this.Value = rootTrace;
+            this.cachedServerSideAccumulatedMetrics = null;
```
adityasa, line 35: this.accumulatedMetrics = new Lazy<ServerSideAccumulatedMetrics(() => this.GetQueryMetrics())
... or something like that (that allows for initialization of actual metrics).

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/PartitionedServerSideMetrics.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Per-partition metrics for queries from the backend.
+    /// </summary>
+    public sealed class PartitionedServerSideMetrics
+    {
+        internal PartitionedServerSideMetrics(ServerSideMetricsInternal serverSideMetricsInternal)
+        {
+            this.ServerSideMetrics = new ServerSideMetrics(serverSideMetricsInternal);
+            this.FeedRange = serverSideMetricsInternal.FeedRange;
+            this.PartitionKeyRangeId = serverSideMetricsInternal.PartitionKeyRangeId;
```
adityasa, line 18: Instead of writing this by hand, please chain the constructors:

internal PartitionedServerSideMetrics(ServerSideMetricsInternal serverSideMetricsInternal)
: this(new ServerSideMetrics(serverSideMetricsInternal), serverSideMetricsInternal.FeedRange, serverSideMetricsInternal.PartitionKeyRangeId)

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/PartitionedServerSideMetrics.cs
```diff
@@ -0,0 +1,49 @@
...
+            this.PartitionKeyRangeId = partitionKeyRangeId;
+        }
+
+        /// <summary>
+        /// Gets the backend metrics for the request.
+        /// </summary>
+        public ServerSideMetrics ServerSideMetrics { get; }
+
+        /// <summary>
+        /// Gets the FeedRange for a single backend call.
```
adityasa, line 40: >a single backend call [](http://example.com/codeflow?start=35&length=21)

Instead:
Gets the FeedRange for the partition
or
Gets the FeedRange for the partition to which the metrics belong

Same for the property below.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/PartitionedServerSideMetrics.cs
```diff
@@ -0,0 +1,49 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    using Microsoft.Azure.Cosmos.Query.Core.Metrics;
+
+    /// <summary>
+    /// Per-partition metrics for queries from the backend.
+    /// </summary>
```
adityasa, line 11: Represents server side metrics specific for a single partition.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/RuntimeExecutionTimesInternal.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    using System;
+
+    /// <summary>
+    /// Query runtime execution times in the Azure Cosmos DB service.
+    /// </summary>
+    internal sealed class RuntimeExecutionTimesInternal
+    {
+        public static readonly RuntimeExecutionTimesInternal Empty = new RuntimeExecutionTimesInternal(
+            queryEngineExecutionTime: default,
```
adityasa, line 14: >default [](http://example.com/codeflow?start=38&length=7)

Explicit values please.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternal.cs
```diff
@@ -0,0 +1,177 @@
...
+        public string FeedRange { get; set; }
+
+        /// <summary>
+        /// Gets the partition key range id for a single backend call.
+        /// </summary>
+        public string PartitionKeyRangeId { get; set; }
+
+        public static ServerSideMetricsInternal Create(IEnumerable<ServerSideMetricsInternal> serverSideMetricsEnumerable)
+        {
+            ServerSideMetricsAccumulator accumulator = default;
```
adityasa, line 153: >default [](http://example.com/codeflow?start=55&length=7)

Doesn't default value here mean a null?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsUtils.cs
```diff
@@ -0,0 +1,18 @@
...
+        }
+
+        public static string FormatTrace(this ServerSideMetrics serverSideMetrics)
```
adityasa, line 13: Is this required (for the public class) ?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternal.cs
adityasa: Likely missing coverage for this since L156 should throw NullReferenceException?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsUtils.cs
Maya-Painter: Updated to be internal, not public

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternal.cs
Maya-Painter: Interesting, this existed before my changes.  Guess this bug was never caught?  In any case, it's not being used.  Do we just want to remove it? @adityasa

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsUtils.cs
Maya-Painter: ServerSideMetricsUtils is internal so the function isn't public.  I was using it to compare object equality in testing in lieu of overriding the equality operator.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternal.cs
adityasa: Structs have a "valid" default value, but with this being a class, that changes. What you have now is expected, although the missing coverage aspect still applies. Were you able to craft a test that would exercise this codepath (and therefore would hit NullRef with earlier version of the code) ?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsUtils.cs
adityasa: I would write a Separate test specific utility function for comparison instead of additional product code.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternal.cs
Maya-Painter: Added a test in ServerSizeMetricsTests that exercises this path.  It did hit the nullref with the previous version of the code but now runs as expected.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsUtils.cs
Maya-Painter: Sounds good.  The function has been removed from this class and added as a utility func to the test class where it is used.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
ealsur: If the properties are virtual then the user can mock the values, it does not need to use any setters

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -30,6 +35,7 @@ public CosmosTraceDiagnostics(ITrace trace)
             }
 
             this.Value = rootTrace;
+            this.accumulatedMetrics = this.accumulatedMetrics = new Lazy<ServerSideCumulativeMetrics>(() => this.GetQueryMetrics());
```
adityasa, line 38: >this.accumulatedMetrics = this.accumulatedMetrics [](http://example.com/codeflow?start=12&length=49)

Multiple assignments to this.accumulatedMetrics.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -49,6 +55,24 @@ public override IReadOnlyList<(string regionName, Uri uri)> GetContactedRegions(
             return this.Value?.Summary?.RegionsContacted;
         }
 
+        public override ServerSideCumulativeMetrics GetQueryMetrics()
```
adityasa, line 58: >GetQueryMetrics [](http://example.com/codeflow?start=52&length=15)

Since this is the public overriden method on CosmosDiagnostics, it should just return this.accumulatedMetrics.Value;

adityasa: You will need a separate, private method that populates ServerSideCumulativeMetrics for the initialization of this.accumulatedMetrics, something like PopulateServerSideCumulativeMetrics.
Typically such a method is a private static, making it harder to be called even accidentally.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -9,12 +9,17 @@ namespace Microsoft.Azure.Cosmos.Diagnostics
...
+    using Microsoft.Azure.Cosmos.Query.Core.Metrics;
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Cosmos.Tracing.TraceData;
     using static Microsoft.Azure.Cosmos.Tracing.TraceData.ClientSideRequestStatisticsTraceDatum;
 
     internal sealed class CosmosTraceDiagnostics : CosmosDiagnostics
     {
+        private readonly Lazy<ServerSideCumulativeMetrics> accumulatedMetrics;
+
+        public ServerSideCumulativeMetrics AccumulatedMetrics => this.accumulatedMetrics.Value;
```
adityasa, line 21: Not required. Also callers won't be able to call this since it's not a part of common base class, which is how callers interact with diagnostics.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideCumulativeMetrics.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServerSideCumulativeMetrics"/> class.
+        /// </summary>
+        /// <param name="cumulativeMetrics"></param>
+        /// <param name="serverSideMetricsList"></param>
+        public ServerSideCumulativeMetrics(ServerSideMetrics cumulativeMetrics, List<ServerSidePartitionedMetrics> serverSideMetricsList)
+        {
+            this.CumulativeMetrics = cumulativeMetrics;
+            this.PartitionedMetrics = serverSideMetricsList;
```
adityasa, line 21: >serverSideMetricsList [](http://example.com/codeflow?start=38&length=21)

serverSideMetricsList.ToList()

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideCumulativeMetrics.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// Metrics received for queries from the backend.
+    /// </summary>
+    public sealed class ServerSideCumulativeMetrics
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServerSideCumulativeMetrics"/> class.
+        /// </summary>
+        /// <param name="cumulativeMetrics"></param>
+        /// <param name="serverSideMetricsList"></param>
+        public ServerSideCumulativeMetrics(ServerSideMetrics cumulativeMetrics, List<ServerSidePartitionedMetrics> serverSideMetricsList)
```
adityasa, line 18: >ServerSideMetrics cumulativeMetrics,  [](http://example.com/codeflow?start=43&length=37)

For consistency, it's better to avoid taking this as a parameter and rather derive it from the second parameter (list).

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
Maya-Painter: Sorry, I thought we wanted to cache this to prevent needing to rewalk the diagnostics every time GetQueryMetrics is called?

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
Maya-Painter: When should this method be called?  upon Diagnostics creation?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideCumulativeMetrics.cs
Maya-Painter: The way this is implemented now it will require adding a support for accumulating ServerSideMetrics.  Right now we just support accumulating ServerSideMetricsInternal, and then convert to ServerSideMetrics.  To avoid this I updated the constructor to take the accumulator instead.  Let me know if this works.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
Maya-Painter: Updated to unseal classes, make properties virtual, and add empty constructor for all 3 public classes.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
Maya-Painter: Static PopulateServerSideCumulativeMetrics function added.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
```diff
@@ -1261,6 +1268,7 @@ public async Task QuerySinglePartitionItemStreamTest(int perPKItemCount, int max
             do
             {
                 iterationCount++;
+                ServerSideMetricsAccumulator headerMetricsAccumulator = new ServerSideMetricsAccumulator();
```
adityasa, line 1271: > ServerSideMetricsAccumulator headerMetricsAccumulator = new ServerSideMetricsAccumulator(); [](http://example.com/codeflow?start=15&length=92)

Doesn't seem like this is used?

adityasa (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideCumulativeMetrics.cs
```diff
@@ -0,0 +1,42 @@
...
+        internal ServerSideCumulativeMetrics(ServerSideMetricsAccumulator accumulator)
+        {
+            this.PartitionedMetrics = accumulator.GetPartitionedServerSideMetrics().Select(metrics => new ServerSidePartitionedMetrics(metrics)).ToList();
+            this.CumulativeMetrics = new ServerSideMetrics(accumulator.GetServerSideMetrics());
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServerSideCumulativeMetrics"/> class.
+        /// </summary>
+        public ServerSideCumulativeMetrics()
```
adityasa, line 28: I would avoid this constructor, especially since this is public and we won't be able to recall it.

adityasa: If we need a public one, I would have and in the constructor, calculate the cumulative metrics using the input partitioned metrics. It may seem wasteful, but all of this is happening on demand and gets cached once calculated so should not be a big issue.

public ServerSideCumulativeMetrics(IEnumerable<ServerSidePartitionedMetrics> partitionedMetrics)
{
 ...
}

Maya-Painter: The public one was introduced at Matias's request so that the class can be mocked

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideCumulativeMetrics.cs
Maya-Painter: updated to take list parameter and accumulate for cumulative value

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideCumulativeMetrics.cs
adityasa: I think the public surface area needs to be controlled further (for e.g. sealed class, no virtual properties, no default constructor) on these. Let us discuss on what the scenario is.

adityasa: I read through the comments. Not my preferred approach for achieving mocking but given that it's an existing practice with much larger part of the codebase (RequestMessage), I am ok with this.

adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsAccumulator.cs
```diff
@@ -0,0 +1,77 @@
...
+            TimeSpan indexLookupTime = TimeSpan.Zero;
+            TimeSpan documentLoadTime = TimeSpan.Zero;
+            TimeSpan runtimeExecutionTime = TimeSpan.Zero;
+            TimeSpan documentWriteTime = TimeSpan.Zero;
+            TimeSpan vMExecutionTime = TimeSpan.Zero;
+
+            foreach (ServerSideMetrics serverSideMetrics in this.serverSideMetricsList)
+            {
+                indexHitRatio = (retrievedDocumentCount + serverSideMetrics.RetrievedDocumentCount) != 0 ?
+                    ((outputDocumentCount * indexHitRatio) + (serverSideMetrics.OutputDocumentCount * serverSideMetrics.IndexHitRatio)) / (retrievedDocumentCount + serverSideMetrics.RetrievedDocumentCount) :
```
neildsh, line 47: >outputDocumentCount [](http://example.com/codeflow?start=22&length=19)

shouldn't this be retrievedDocumentCount ? same for the other term

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsAccumulator.cs
```diff
@@ -0,0 +1,77 @@
...
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Query.Core.Metrics
+{
+    using System;
+    using System.Collections.Generic;
+
+    internal class ServerSideMetricsAccumulator
+    {
+        internal readonly List<ServerSideMetrics> serverSideMetricsList;
```
neildsh, line 12: >serverSideMetricsList [](http://example.com/codeflow?start=50&length=21)

would anyone ever access this list (other than `GetServerSideMetrics`)? If not then we may not need to keep this around

neildsh, line 12: >internal [](http://example.com/codeflow?start=8&length=8)

fields should be private. If internal classes need access create a IReadOnlyList property that is internal

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideCumulativeMetrics.cs
neildsh: Seems like an anti-pattern. I share AdityaSa's reservations. we should not add cruft to the sdk because our customers may have some need to mock some class in the future.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ClientSideMetricsAccumulator.cs
```diff
@@ -0,0 +1,48 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.Query.Core.Metrics
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    internal class ClientSideMetricsAccumulator
```
neildsh, line 10: >class [](http://example.com/codeflow?start=13&length=5)

`sealed` everywhere

Maya-Painter: updated all internal classes to sealed.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsAccumulator.cs
Maya-Painter: You're right, this can be private. Updated now, thanks!

Maya-Painter: Updated to be private.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsAccumulator.cs
Maya-Painter: Good catch, I think you're right.  I've fixed the calculation on this class and the internal accumulator. Looks like this slipped through the cracks in the original metrics class 😅

adityasa (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/IndexUtilizationInfoAccumulator.cs
```diff
@@ -0,0 +1,51 @@
...
+            if (indexUtilizationInfo == null)
+            {
+                throw new ArgumentNullException(nameof(indexUtilizationInfo));
+            }
+
+            this.indexUtilizationInfoList.Add(indexUtilizationInfo);
+        }
+
+        public IndexUtilizationInfo GetIndexUtilizationInfo()
+        {
```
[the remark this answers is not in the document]
Maya-Painter: Good to know.  This provides the same functionality as what existed before these changes, it's just updated to match the pattern used by the other accumulators.  We can explore simplifying this further in a future PR.

kirankumarkolli (commented):
Naming wise, azure central SDK team feedback would be appreciated. 
Currently .NET SDK API changes are not reviewed by central SDK team (not yet Track2). JAVA SDK API changes will trigger the review. 

One path forward is to mark these API's for preview and post JAVA API changes reflect on any changes for GA. 

@adityasa, @Maya-Painter thoughts?

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -49,6 +53,11 @@ public override IReadOnlyList<(string regionName, Uri uri)> GetContactedRegions(
             return this.Value?.Summary?.RegionsContacted;
         }
 
+        public override ServerSideCumulativeMetrics GetQueryMetrics()
```
lmolkova, line 56: what are the reasons to return customly defined metrics instead of reporting them using .NET metrics APIs?

how would such metrics be used? How they would be integrated with OTel?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// Gets the query preparation time in the Azure Cosmos database service.
+        /// </summary>
+        public abstract TimeSpan QueryPreparationTime { get; }
+
+        /// <summary>
+        /// Gets the query index lookup time in the Azure Cosmos database service.
+        /// </summary>
+        public abstract TimeSpan IndexLookupTime { get; }
```
lmolkova, line 47: is it an average time? mean time? Users would be interested in p50, p95, p99.

assuming we don't use .NET Metrics API and report metrics to otel, this should rather be a histogram of some sorts.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        public abstract TimeSpan DocumentWriteTime { get; }
+
+        /// <summary>
+        /// Gets the index hit ratio by query in the Azure Cosmos database service.
+        /// </summary>
+        public abstract double IndexHitRatio { get; }
+
+        /// <summary>
+        /// Gets the VMExecution Time.
```
lmolkova, line 70: what's `VMExecution` ?

lmolkova:
From observability standpoint, I don't really understand how this would help users. 

Metrics is a specific observability signal and .NET supports metrics API which Cosmos could use to report metrics directly to OTel - https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation

Users who would need structured `ServerSideMetrics` would likely need to build integration with their metrics backend to make them useful. Users who just log stuff should be fine with json string (or a dictionary instead of a specific model).

So it does not seem to solve user problem, but introduces confusion and more work for users. 

I'd suggest to:
- keep these metrics unstructured or return them as a dictionary
- Document  that those have nothing to do with .NET metrics (consider using `stats` or another term for it to reduce possible confusion)
- explore adding real metrics that'd report cosmos diagnostics metrics to otel directly.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
adityasa: Goal of this PR is to expose the query metrics as a strongly typed object.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
adityasa: This is absolute value of time it took to perform index lookup. Same for remaining values.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// Gets the output writing/serializing time during query in the Azure Cosmos database service.
+        /// </summary>
+        public abstract TimeSpan DocumentWriteTime { get; }
+
+        /// <summary>
+        /// Gets the index hit ratio by query in the Azure Cosmos database service.
+        /// </summary>
+        public abstract double IndexHitRatio { get; }
```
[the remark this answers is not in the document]
adityasa: As discussed offline, here's a link to query metrics documentation:
https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query-metrics

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetrics.cs
adityasa: As discussed offline, here's a link to query metrics documentation:
https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query-metrics

adityasa:
@lmolkova thank you for walking through your comments offline with me. As agreed in the meeting, I resolved your comments with the link to our customer facing query metrics documentation.

As for the last comment regarding observability below is what we discussed:
.NET's otel/metrics API would be ideal to onboard at some point in future, although we have multiple steps to get there.
https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-collection
This API is more suitable for aggregation (during the application lifetime) and automatic logging to telemetry stores (using otel pipeline).

The current proposed approach is designed for retrieving query metrics per query execution/roundtrip and accumulated in the client application (or output to console for experimentation purposes). This does not necessarily preclude onboarding .NET metrics in future, which is a preferred approach for applicable scenraios.
Kindly comment/correct in case I missed something.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosDiagnostics.cs
```diff
@@ -43,6 +43,16 @@ public virtual int GetFailedRequestCount()
             throw new NotImplementedException($"{nameof(CosmosDiagnostics)}.{nameof(GetFailedRequestCount)}");
         }
 
+        /// <summary>
+        ///  This represents the backend query metrics for the request.
```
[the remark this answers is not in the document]
Maya-Painter: Added as a remark on method doc.

kirankumarkolli (approved):

adityasa (approved):
:shipit:

PR #4003, opened 2023-07-24
[Internal] FFCF: Initial PR for AllVersionsAndDeletes

iainx:
## Description

WIP initial PR for comments

Waiting for a final API before writing full documentation and tests

 - [x] Full Documentation (current documentation is minimal to fix build)
 - [x] Tests

API example

```csharp
var builder = container.GetAllVersionsChangeFeedProcessorBuilder<CartAction>("migrationProcessor", (ChangeFeedProcessorContext context, IReadOnlyCollection <ChangeFeedItemChange<CartAction>> changes, CancellationToken cancellationToken) =>
{
    Console.WriteLine(changes.Count + " Changes Received");


    {
    }
}
```


iainx:
Seems like there's some line endings been messed up somewhere, will fix

iainx:
> Seems like there's some line endings been messed up somewhere, will fix

Fixed now, wasn't line endings or indentation, not sure why Git felt the whole file had changed

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/Observers/ChangeFeedObserverFactoryCore.cs
```diff
@@ -52,6 +52,7 @@ internal sealed class ChangeFeedObserverFactoryCore<T> : ChangeFeedObserverFacto
         private readonly ChangesHandler<T> legacyOnChanges;
         private readonly ChangeFeedHandler<T> onChanges;
         private readonly ChangeFeedHandlerWithManualCheckpoint<T> onChangesWithManualCheckpoint;
+        private readonly AllVersionsAndDeleteChangesHandler<T> onAllVersionsAndDeleteChanges;
```
jcocchi, line 55: If we go with this API it needs to be named `AllVersionsAndDeletesChangesHandler` the missing "s" on Delete is misleading here

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/Observers/ChangeFeedObserverFactoryCore.cs
```diff
@@ -111,6 +125,25 @@ public override ChangeFeedObserver CreateObserver()
             return this.onChangesWithManualCheckpoint(context, changes, context.CheckpointAsync, cancellationToken);
         }
 
+        private Task AllVersionsAndDeleteStreamHandlerAsync(
```
jcocchi, line 128: Delete --> Deletes. Similar feedback on other names that have the singular form of delete

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
```diff
@@ -1309,6 +1309,25 @@ public abstract class Container
...
+        /// <typeparam name="T">Document type</typeparam>
+        /// <param name="processorName">A name that identifies the Processor and the particular work it will do.</param>
+        /// <param name="onChangesHandler">Delegate to receive all changes and deletes</param>
+        /// <returns>An instance of <see cref="ChangeFeedProcessorBuilder"/></returns>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        virtual ChangeFeedProcessorBuilder GetAllVersionsChangeFeedProcessorBuilder<T>(
```
philipthomas-MSFT, line 1324: Can we be consistent here. **GetAllVersionsAndDeletesChangeFeedProcessorBuilder**<T>? And why do we need this if this is only going to throw a **NotImplementedException**?

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -577,7 +577,8 @@ internal abstract partial class ContainerCore : ContainerInternal
             ChangeFeedObserverFactory observerFactory = new CheckpointerObserverFactory(
                 new ChangeFeedObserverFactoryCore<T>(onChangesDelegate, this.ClientContext.SerializerCore),
                 withManualCheckpointing: false);
-            return this.GetChangeFeedProcessorBuilderPrivate(processorName, observerFactory);
+            return this.GetChangeFeedProcessorBuilderPrivate(processorName,
+                observerFactory, ChangeFeedMode.Incremental);
```
philipthomas-MSFT, line 581: Any reason why we are defaulting this to **Incremental**?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
```diff
@@ -54,6 +54,7 @@ internal sealed class ChangeFeedEstimatorIterator : FeedIterator<ChangeFeedProce
                   changeFeedEstimatorRequestOptions,
                   (DocumentServiceLease lease, string continuationToken, bool startFromBeginning) => ChangeFeedPartitionKeyResultSetIteratorCore.Create(
                           lease: lease,
+                          mode: ChangeFeedMode.LatestVersion,
```
philipthomas-MSFT, line 57: Why are we defaulting to **LatestVersion** here?

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -597,7 +598,34 @@ internal abstract partial class ContainerCore : ContainerInternal
...
```
philipthomas-MSFT, line 602: Another decision to default this to **Incremental**? Why not **LatestVersion**?

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -617,7 +645,8 @@ internal abstract partial class ContainerCore : ContainerInternal
             ChangeFeedObserverFactory observerFactory = new CheckpointerObserverFactory(
                 new ChangeFeedObserverFactoryCore<T>(onChangesDelegate, this.ClientContext.SerializerCore),
                 withManualCheckpointing: true);
-            return this.GetChangeFeedProcessorBuilderPrivate(processorName, observerFactory);
+            return this.GetChangeFeedProcessorBuilderPrivate(processorName,
+                observerFactory, ChangeFeedMode.Incremental);
```
philipthomas-MSFT, line 649: Why **Incremental** and not **LatestVersion**?

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -637,7 +666,8 @@ internal abstract partial class ContainerCore : ContainerInternal
             ChangeFeedObserverFactory observerFactory = new CheckpointerObserverFactory(
                 new ChangeFeedObserverFactoryCore(onChangesDelegate),
                 withManualCheckpointing: false);
-            return this.GetChangeFeedProcessorBuilderPrivate(processorName, observerFactory);
+            return this.GetChangeFeedProcessorBuilderPrivate(processorName,
+                observerFactory, ChangeFeedMode.Incremental);
```
philipthomas-MSFT, line 670: Why **Incremental** and not **LatestVersion**?

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -657,7 +687,9 @@ internal abstract partial class ContainerCore : ContainerInternal
             ChangeFeedObserverFactory observerFactory = new CheckpointerObserverFactory(
                 new ChangeFeedObserverFactoryCore(onChangesDelegate),
                 withManualCheckpointing: true);
-            return this.GetChangeFeedProcessorBuilderPrivate(processorName, observerFactory);
+            return this.GetChangeFeedProcessorBuilderPrivate(processorName,
+                observerFactory,
+                ChangeFeedMode.Incremental);
```
philipthomas-MSFT, line 692: Why **Incremental** and not **LatestVersion**?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs
```diff
@@ -80,6 +83,94 @@ public async Task TestWithRunningProcessor()
             Assert.AreEqual("0.1.2.3.4.5.6.7.8.9.", accumulator);
         }
 
+        [TestMethod]
+        public async Task TestWithRunningFFCFProcessor()
```
philipthomas-MSFT, line 87: It is a bit difficult to follow what this test is actually testing for? Maybe too much in one test? Maybe one test can create docs, then verify docs created. Maybe another test to created docs, then updated, then verify docs updated. Maybe another test and verify deletes. Maybe another tests that is past retention and should have no documents.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCoreTests.cs
```diff
@@ -78,6 +78,7 @@ public async Task EtagPassesContinuation()
             
             ChangeFeedPartitionKeyResultSetIteratorCore iterator = ChangeFeedPartitionKeyResultSetIteratorCore.Create(
                 lease: documentServiceLeaseCore,
+                mode: ChangeFeedMode.Incremental,
```
philipthomas-MSFT, line 81: This goes for all of the tests for **Incremental**, **LatestVersion** is going to be replacing this so we need to figure out when we want to transition. We can't get rid of **Incremental** until a major release version change, but we are exposing **LatestVersion** and giving cx access.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/Observers/ChangeFeedObserverFactoryCore.cs
jcocchi: @iainx looks like you marked this comment as resolved but the name still uses a singular "Delete". I think this should be "AllVersionsAndDeletesStreamHandlerAsync"

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
jcocchi: +1, all "Incremental" should use "LatestVersion" instead as this is the new name

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorItem.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Microsoft.Azure.Cosmos
+{
+    using System;
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// An item representing a document change
+    /// </summary>
+    /// <typeparam name="T">The document type</typeparam>
+    public sealed class ChangeFeedProcessorItem<T>
```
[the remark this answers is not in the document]
jcocchi: I've given this feedback before, but I think we should rename this class to "ChangeFeedItem", "ChangeFeedItemChange" is repetitive and confusing IMO

jcocchi:
@iainx can you please add a simple usage example to the description so it's clear how customers would use the new APIs?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCoreTests.cs
bartelink: Can the old names, i.e.  `Incremental` etc not be marked `BrowserVisible.Never` ? Having a noisy set of identifiers in intellisense is a horrible experience - even if obsoletion notes in the xmldoc could temper it a little.

Also I think marking them `Obsolete` would be for the best
- it's not going to inconvenience a lot of people
- it would flush out all usages flagged in the PR reviews here (and the tests that need to cover them can inhibit the warnings)

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
philipthomas-MSFT: Wanted to extend my comment. Are we saying that **ChangeFeedEstimatorIterator** is only used in **LastestVersion** and never **AllVersionsAndDeletes**?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorItem.cs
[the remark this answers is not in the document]
iainx: Yes, we should, but I'd prefer to do that in a separate PR because it affects the whole of the ChangeFeedProcessor and would make this PR more confusing.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
iainx: Because of how the Encryption module is implemented.

This can't be abstract as that would mean it needs to be implemented by the `EncryptionContainer` subclass, but because `Microsoft.Azure.CosmosDB` doesn't expose its internals to `Microsoft.Azure.CosmosDB.Encryption` `EncryptionContainer` can only implement this is `PREVIEW` is set. That means, if `PREVIEW` is not set, a build error happens because an abstract method isn't implemented.

Making this method virtual with an empty body fixes that, which looks and feels weird and wrong but that's a side effect of the way `Microsoft.Azure.CosmosDB.Encryption` is using `Microsoft.Azure.CosmosDB` as a `PackageReference` and not a `ProjectReference`

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
iainx: Well, I used that as the default because that was how it was used before AVAD was added and I didn't want to break the API or the contract of a public API. Adding a new public API that adds a mode parameter makes sense to me.

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
iainx: Because I wasn't sure which was the new version and which was the old. I'll switch them all to `LatestVersion`

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCoreTests.cs
ealsur: Marking Obsolete(which I'd love to) might break a build, which for many customers, is a no-no and considered a breaking change

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
ealsur, line 57: One idea could be: Store the mode the Processor was started with on the `.info` document that is created, then read that document on the estimator to understand what mode we shoud use

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -27,9 +28,10 @@ internal sealed class ChangeFeedProcessorCore : ChangeFeedProcessor
         private DocumentServiceLeaseStoreManager documentServiceLeaseStoreManager;
         private bool initialized = false;
 
-        public ChangeFeedProcessorCore(ChangeFeedObserverFactory observerFactory)
+        public ChangeFeedProcessorCore(ChangeFeedObserverFactory observerFactory, ChangeFeedMode mode)
         {
             this.observerFactory = observerFactory ?? throw new ArgumentNullException(nameof(observerFactory));
+            this.mode = mode;
```
ealsur, line 34: Why not add the mode to the ChangeFeedProcessorOptions that is already passed down?

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
ealsur: Dang `GetAllVersionsAndDeletesChangeFeedProcessorBuilder` is long :D

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCoreTests.cs
bartelink: That's why obsoletion is by default only a warning for C# and F# :P But, fair enough.

IMO hiding it via BrowserVisible is the right thing to do there - the Intellisense experience will be far less noisy.

ealsur: Yeah, but it is common for users to have TreatWarningsAsErrors and that breaks a build :(

I agree that hiding it would be great from Intellisense

ealsur: `[EditorBrowsable(EditorBrowsableState.Never)]` would be the right decorator.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
jcocchi: Should we consider shortening to `GetAllVersionsAndDeletesCFPBuilder`? I'm not sure if there's guidelines against using acronyms

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
```diff
@@ -125,9 +131,8 @@ public override async Task<ResponseMessage> ReadNextAsync(ITrace trace, Cancella
                             this.changeFeedOptions.PageSizeHint.Value.ToString(CultureInfo.InvariantCulture));
                     }
 
-                    requestMessage.Headers.Add(
-                        HttpConstants.HttpHeaders.A_IM,
-                        HttpConstants.A_IMHeaderValues.IncrementalFeed);
+                    // Add the necessary mode headers
```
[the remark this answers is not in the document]
jcocchi: Adding the mode on the `.info` file is a good idea. In Java they added the mode as part of the continuation token for each lease. 

I was under the impression that reusing leases across modes wasn't supported for technical reasons (though customers have already asked how they can use latest version until they "catch up" to all versions and deletes then switch over). Supporting lease compatibility is a good future goal but not current requirement.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCoreTests.cs
bartelink: > Yeah, but it is common for users to have TreatWarningsAsErrors and that breaks a build

I get that, and I've already accepted that it's something that's reasonable to make a subjective call on, for better or for worse.

But now we're still talking, some more devil's advocate points for Obsolete:
- people with TWAE are adults that have opted into something
- using a barely out of preview feature is absolutely the definition of a non-mainline scenario
- they will thank you as a pile of docs are about to make the old names look very stupid
- FCCF is not a mainline scenario
- I don't think it's actually written anywhere that obsoleting something is as cheeky as a breaking change?
- The fact that this repo is working on an evergreen basis without any obsoletions in site is another to consider bending the rule in this instance

Having said all of that, hiding it with BrowserVisible is a fine solution for me in terms of the actual impact on real humans.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
ealsur: Well, it would be a bit inconsistent, some called ChangeFeedProcessor and others CFP :/

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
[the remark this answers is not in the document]
ealsur: `.info` or in the Leases, both solutions work, yeah. As long as we save it to use it, for example, in the Estimator or to validate if the leases were created with one mode and now attempted to be used with another.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
jcocchi: Yeah I think we have to live with the long name or pass the mode as a parameter. BTW, Java exposes [all versions and deletes as a separate method reusing the same builder](https://learn.microsoft.com/en-us/java/api/com.azure.cosmos.changefeedprocessorbuilder?view=azure-java-stable#com-azure-cosmos-changefeedprocessorbuilder-handleallversionsanddeleteschanges(java-util-function-consumer(java-util-list(com-azure-cosmos-models-changefeedprocessoritem)))), not sure if this would apply to .NET/ simplify creation

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
ealsur: After syncing with @jcocchi, using LatestVersion for the Estimator is ok

ealsur: Disregard this comment. For Estimator, using LatestVersion is ok

iainx:
> @iainx can you please add a simple usage example to the description so it's clear how customers would use the new APIs?
> 
> +1

Done. It was initially in the API discussion thread, but I've moved it to here

Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainer.cs
```diff
@@ -592,6 +592,51 @@ internal sealed class EncryptionContainer : Container
                 });
         }
 
+#if SDKPROJECTREF
+        public override ChangeFeedProcessorBuilder GetAllVersionsChangeFeedProcessorBuilder<T>(
```
philipthomas-MSFT, line 596: Where are the unit or integration tests for `GetAllVersionsChangeFeedProcessorBuilder<T>`?

Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainer.cs
```diff
@@ -592,6 +592,51 @@ internal sealed class EncryptionContainer : Container
                 });
         }
 
+#if SDKPROJECTREF
```
philipthomas-MSFT, line 595: What is the significance of `SDKPROJECTREF`?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
philipthomas-MSFT: > After syncing with @jcocchi, using LatestVersion for the Estimator is ok

@ealsur or @jcocchi, just for clarification and from a use case perspective, why do we **NOT** want to support **AllVersionsAndDeletes** for _ChangeFeedEstimatorIterator_?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
```diff
@@ -69,6 +69,17 @@ public ChangeFeedProcessorBuilder WithInstanceName(string instanceName)
             return this;
         }
 
+        /// <summary>
+        /// Sets the mode for the change freed processor.
```
philipthomas-MSFT, line 73: Spelling correction "feed".

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
jcocchi: @philipthomas-MSFT change feed estimator looks at the lag based on the current leases, calculating the lag is independent of change feed mode. The existing estimator works for both modes by design

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs
```diff
@@ -25,8 +26,11 @@ public async Task TestInitialize()
             await base.ChangeFeedTestInit();
 
             string PartitionKey = "/pk";
-            ContainerResponse response = await this.database.CreateContainerAsync(
-                new ContainerProperties(id: Guid.NewGuid().ToString(), partitionKeyPath: PartitionKey),
+            ContainerProperties properties = new ContainerProperties(id: Guid.NewGuid().ToString(), 
+                partitionKeyPath: PartitionKey);
+            properties.ChangeFeedPolicy.FullFidelityRetention = TimeSpan.FromMinutes(5);
```
philipthomas-MSFT, line 31: Need to change. No need to set retention for non-AllVersionsAndDeletes tests.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorItem.cs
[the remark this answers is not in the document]
philipthomas-MSFT: +1 on that changing this in another PR.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
philipthomas-MSFT: Went with `GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes` following other similar method naming conventions.

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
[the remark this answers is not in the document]
philipthomas-MSFT: Confirming this with Java team.

philipthomas-MSFT:
replaced by another PR that is already closed.

PR #4009, opened 2023-07-25
Query: Adds ODE continuation token support for non-ODE pipelines

akotalwar:
# Pull Request Template

## Description

With Optimistic Direct Execution (ODE) being enabled by default in preview mode, a certain concern is brought up. 
What happens if an ODE continuation token goes into a non-ODE pipeline?
This would happen for 2 reasons:
1: Customer has multiple SDKs which are not all of the same version. The newer SDK generates an ODE continuation token and sends it to an older SDK that does not have ODE capabilities. In such a case, a different pipeline would have to handle this ODE token.
2: Customer gets an ODE continuation token, turns off ODE from QueryRequestOptions, and then passes the continuation token back to the SDK. With ODE turned off, a non-ODE pipeline would have to handle this token. 

The first case gets handled by a continuation token check which confirms that the incoming continuation token is of the same version as the SDK. If they are not, then a Malformed Continuation Token exception is thrown. 

This PR looks to add support for the second case. 
In the current behavior, the non-ODE pipeline returns 0 results and ends execution. This behavior is incorrect as it would lead to customers getting wrong results. 
This PR fixes this issue by throwing a Malformed Continuation Token Exception when an ODE continuation token reaches a non-ODE pipeline and lets the customer know that they need to turn on their ODE flag to avoid this issue. 

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
@@ -32,7 +32,6 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
     {
         private const string InternalPartitionKeyDefinitionProperty = "x-ms-query-partitionkey-definition";
         private const string OptimisticDirectExecution = "OptimisticDirectExecution";
-        private const string OptimisticDirectExecutionToken = "OptimisticDirectExecutionToken";
         private const string Passthrough = "Passthrough";
         private const string Specialized = "Specialized";
         private const int PageSizeFactorForTop = 5;
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
@@ -308,19 +308,6 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
             }
             else
             {
-                if (inputParameters.InitialUserContinuationToken != null)
-                {
-                    CosmosObject cosmosObjectContinuationToken = inputParameters.InitialUserContinuationToken as CosmosObject;
-                    if (cosmosObjectContinuationToken.ContainsKey(OptimisticDirectExecutionToken))
-                    {
-                        return TryCatch<IQueryPipelineStage>.FromException(
-                            new MalformedContinuationTokenException(
-                                $"Operation cannot be resumed with the following token, as it requires the Optimistic Direct Execution pipeline which has been disabled on your SDK. " +
-                                $"Enable Optimistic Direct Execution from QueryRequestOptions to avoid this issue." +
-                                $"{inputParameters.InitialUserContinuationToken}."));
-                    }
-                }
-
                 if (createPassthroughQuery)
                 {
                     SetTestInjectionPipelineType(inputParameters, Passthrough);
@@ -766,7 +753,18 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
             ContainerQueryProperties containerQueryProperties,
             ITrace trace)
         {
-            if (!inputParameters.EnableOptimisticDirectExecution) return null;
+            if (!inputParameters.EnableOptimisticDirectExecution)
+            {
+                if (inputParameters.InitialUserContinuationToken != null)
+                {
+                    if (OptimisticDirectExecutionContinuationToken.IsOptimisticDirectExecutionContinuationToken(inputParameters.InitialUserContinuationToken))
+                    {
+                        throw new MalformedContinuationTokenException($"The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. " +
+                                    $"{inputParameters.InitialUserContinuationToken}");
+                    }
+                }
+                
+            }
 
             Debug.Assert(containerQueryProperties.ResourceId != null, "CosmosQueryExecutionContextFactory Assert!", "Container ResourceId cannot be null!");
 
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
@@ -753,17 +753,12 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
             ContainerQueryProperties containerQueryProperties,
             ITrace trace)
         {
-            if (!inputParameters.EnableOptimisticDirectExecution)
+            if (!inputParameters.EnableOptimisticDirectExecution
+                && inputParameters.InitialUserContinuationToken != null
+                && OptimisticDirectExecutionContinuationToken.IsOptimisticDirectExecutionContinuationToken(inputParameters.InitialUserContinuationToken))
             {
-                if (inputParameters.InitialUserContinuationToken != null)
-                {
-                    if (OptimisticDirectExecutionContinuationToken.IsOptimisticDirectExecutionContinuationToken(inputParameters.InitialUserContinuationToken))
-                    {
-                        throw new MalformedContinuationTokenException($"The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. " +
-                                    $"{inputParameters.InitialUserContinuationToken}");
-                    }
-                }
-                
+                throw new MalformedContinuationTokenException($"The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. " +
+                            $"{inputParameters.InitialUserContinuationToken}");
             }
 
             Debug.Assert(containerQueryProperties.ResourceId != null, "CosmosQueryExecutionContextFactory Assert!", "Container ResourceId cannot be null!");
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
@@ -32,6 +32,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
     {
         private const string InternalPartitionKeyDefinitionProperty = "x-ms-query-partitionkey-definition";
         private const string OptimisticDirectExecution = "OptimisticDirectExecution";
+        private const string OptimisticDirectExecutionToken = "OptimisticDirectExecutionToken";
         private const string Passthrough = "Passthrough";
         private const string Specialized = "Specialized";
         private const int PageSizeFactorForTop = 5;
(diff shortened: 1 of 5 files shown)


Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -307,6 +308,19 @@ internal static class CosmosQueryExecutionContextFactory
             }
             else
             {
+                if (inputParameters.InitialUserContinuationToken != null)
+                {
+                    CosmosObject cosmosObjectContinuationToken = inputParameters.InitialUserContinuationToken as CosmosObject;
+                    if (cosmosObjectContinuationToken.ContainsKey(OptimisticDirectExecutionToken))
```
neildsh, line 314: >if (cosmosObjectContinuationToken.ContainsKey(OptimisticDirectExecutionToken)) [](http://example.com/codeflow?start=20&length=78)

we should not leak details of ODE tokens in unrelated classes. This line can also cause a NullReferenceException

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -307,6 +308,19 @@ internal static class CosmosQueryExecutionContextFactory
...
+                    {
+                        return TryCatch<IQueryPipelineStage>.FromException(
+                            new MalformedContinuationTokenException(
+                                $"Operation cannot be resumed with the following token, as it requires the Optimistic Direct Execution pipeline which has been disabled on your SDK. " +
+                                $"Enable Optimistic Direct Execution from QueryRequestOptions to avoid this issue." +
```
adityasa, line 319: The continuation token supplied requires the Optimistic Direct Execution to be enabled in QueryRequestOptions for the query to succeed.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -752,7 +753,18 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
             ContainerQueryProperties containerQueryProperties,
             ITrace trace)
         {
-            if (!inputParameters.EnableOptimisticDirectExecution) return null;
+            if (!inputParameters.EnableOptimisticDirectExecution)
+            {
+                if (inputParameters.InitialUserContinuationToken != null)
```
sourabh1007, line 758: instead of 3 nsted if, change it to single if with && condition

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionContinuationToken.cs
```diff
@@ -30,6 +28,19 @@ public OptimisticDirectExecutionContinuationToken(ParallelContinuationToken toke
 
         public Range<string> Range => this.Token.Range;
 
+        public static bool IsOptimisticDirectExecutionContinuationToken(CosmosElement continuationToken)
+        {
+            CosmosObject cosmosObjectContinuationToken = continuationToken as CosmosObject;
+            if (cosmosObjectContinuationToken == null || !cosmosObjectContinuationToken.ContainsKey(OptimisticDirectExecutionToken))
```
sourabh1007, line 34: return this or negation of this directly  `cosmosObjectContinuationToken == null || !cosmosObjectContinuationToken.ContainsKey(OptimisticDirectExecutionToken)`

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosBasicQueryTests.cs
```diff
@@ -786,6 +785,61 @@ public async Task QueryActivityIdWithContinuationTokenAndTraceTest()
...
         }
 
+        [TestMethod]
+        public async Task TesOdeTokenCompatibilityWithNonOdePipeline()
+        {
+            using (ITrace rootTrace = Trace.GetRootTrace("Root Trace"))
+            {
+                ResponseMessage responseMessage = null;
+                string query = "select top 200 * from c";
+                string expectedErrorMessage = "The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. ";
```
adityasa, line 795: >string expectedErrorMessage = "The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. "; [](http://example.com/codeflow?start=16&length=183)

Please declare right before use.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosBasicQueryTests.cs
```diff
@@ -786,6 +785,61 @@ public async Task QueryActivityIdWithContinuationTokenAndTraceTest()
 
         }
 
+        [TestMethod]
+        public async Task TesOdeTokenCompatibilityWithNonOdePipeline()
+        {
+            using (ITrace rootTrace = Trace.GetRootTrace("Root Trace"))
+            {
+                ResponseMessage responseMessage = null;
```
adityasa, line 793: Please inline the declaration with the initialization.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosBasicQueryTests.cs
```diff
@@ -786,6 +785,61 @@ public async Task QueryActivityIdWithContinuationTokenAndTraceTest()
...
+                    await container.CreateItemAsync<ToDoActivity>(ToDoActivity.CreateRandomToDoActivity());
+                }
+
+                QueryRequestOptions queryRequestOptions = new QueryRequestOptions
+                {
+                    MaxItemCount = 50,
+                    EnableOptimisticDirectExecution = true
+                };
+
+                FeedIteratorInternal feedIterator =
```
adityasa, line 812: >FeedIteratorInternal [](http://example.com/codeflow?start=16&length=20)

Why not just FeedIterator?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionContinuationToken.cs
```diff
@@ -30,6 +28,12 @@ public OptimisticDirectExecutionContinuationToken(ParallelContinuationToken toke
 
         public Range<string> Range => this.Token.Range;
 
+        public static bool IsOptimisticDirectExecutionContinuationToken(CosmosElement continuationToken)
+        {
+            CosmosObject cosmosObjectContinuationToken = continuationToken as CosmosObject;
+            return !(cosmosObjectContinuationToken == null || !cosmosObjectContinuationToken.ContainsKey(OptimisticDirectExecutionToken));
```
adityasa, line 34: Nit, a more compact (single line) version:
return (continuationToken as CosmosObject)?.ContainsKey(OptimisticDirectExecutionToken) == true;

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -752,7 +753,13 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
             ContainerQueryProperties containerQueryProperties,
             ITrace trace)
         {
-            if (!inputParameters.EnableOptimisticDirectExecution) return null;
```
adityasa, line 755: It's not clear to me how the old implementation and new implementation are functionally equivalent.
Before we were returning null but instead now we throw an exception.

More importantly, we would still create ODE pipeline even if the EnableOptimisticDirectExecution flag is off but either null or ODE continuation is supplied.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionContinuationToken.cs
```diff
@@ -42,14 +46,14 @@ public static CosmosElement ToCosmosElement(OptimisticDirectExecutionContinuatio
...
+            if (!IsOptimisticDirectExecutionContinuationToken(cosmosElement))
             {
                 return TryCatch<OptimisticDirectExecutionContinuationToken>.FromException(
-                                    new MalformedContinuationTokenException(
-                                        message: $"Malformed Continuation Token: Expected OptimisticDirectExecutionToken\r\n"));
+                                        new MalformedContinuationTokenException(
+                                            message: $"Malformed Continuation Token: Expected OptimisticDirectExecutionToken\r\n"));
             }
 
+            CosmosObject cosmosObjectContinuationToken = cosmosElement as CosmosObject;
```
adityasa, line 56: I would use hard cast:
(CosmosObject)cosmosElement

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionContinuationToken.cs
akotalwar: I find the existing single line version more readable. Is it ok if we continue using that?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionContinuationToken.cs
```diff
@@ -30,6 +28,12 @@ public OptimisticDirectExecutionContinuationToken(ParallelContinuationToken toke
 
         public Range<string> Range => this.Token.Range;
 
+        public static bool IsOptimisticDirectExecutionContinuationToken(CosmosElement continuationToken)
+        {
+            CosmosObject cosmosObjectContinuationToken = (CosmosObject)continuationToken;
```
adityasa, line 33: This needs to be soft cast (using as keyword like you had before), since hard cast could throw InvalidCastException.

adityasa (approved):
:shipit:

sboshra (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -32,6 +32,7 @@ internal static class CosmosQueryExecutionContextFactory
     {
         private const string InternalPartitionKeyDefinitionProperty = "x-ms-query-partitionkey-definition";
         private const string OptimisticDirectExecution = "OptimisticDirectExecution";
+        private const string OptimisticDirectExecutionToken = "OptimisticDirectExecutionToken";
```
neildsh, line 35: >private const string OptimisticDirectExecutionToken = "OptimisticDirectExecutionToken"; [](http://example.com/codeflow?start=8&length=87)

please remove this

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/OptimisticDirectExecution/OptimisticDirectExecutionContinuationToken.cs
neildsh: >(CosmosObject)continuationToken; [](http://example.com/codeflow?start=57&length=32)

This will throw an `InvalidCastException` please use `as`

akotalwar: I think you're looking at an older iteration of this PR. The newest iteration uses 'as' for the cast.

neildsh (approved):

PR #4020, opened 2023-08-08
Proposed Changes to Fix Concurrency Issues

kundadebdatta:
# Pull Request Template

## Description

1. Validate only the `Unhealthy` replicas, and remove the `Unknown` ones from validation.
2. Add a mechanism to get hold of a connection health state, to detect any established connection before attempting to open new connections.

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -1006,7 +1023,6 @@ private static string ProtocolString(Protocol protocol)
                 .Where(address => address
```
kundadebdatta, line 1023: We might need to invest some time to check the perf of the dictionary creation and maintaining for a large partition. Ideally the size should be constant and proportional to the number of partitions.

PR #4037, opened 2023-08-14
[Internal] Client Telemetry: Refactors code for collectors

sourabh1007:
## Description

This PR is part 1/3 of this https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3816 PR.

As part of PR, refactoring client telemetry code to add Collector abstraction.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -142,7 +142,8 @@ namespace Microsoft.Azure.Cosmos
         private Documents.ConsistencyLevel? desiredConsistencyLevel;
 
         internal CosmosAccountServiceConfiguration accountServiceConfiguration { get; private set; }
-        internal ClientTelemetry clientTelemetry { get; set; }
+
+        internal TelemetryToServiceHelper TelemetryToServiceHelper { get; set; }
 
         private ClientCollectionCache collectionCache;
 
@@ -660,7 +661,7 @@ namespace Microsoft.Azure.Cosmos
                     storeModel: this.GatewayStoreModel, 
                     tokenProvider: this, 
                     retryPolicy: this.retryPolicy,
-                    clientTelemetry: this.clientTelemetry);
+                    telemetryToServiceHelper: this.TelemetryToServiceHelper);
                 this.partitionKeyRangeCache = new PartitionKeyRangeCache(this, this.GatewayStoreModel, this.collectionCache);
 
                 DefaultTrace.TraceWarning("{0} occurred while OpenAsync. Exception Message: {1}", ex.ToString(), ex.Message);
@@ -939,6 +940,16 @@ namespace Microsoft.Azure.Cosmos
             // Loading VM Information (non blocking call and initialization won't fail if this call fails)
             VmMetadataApiHandler.TryInitialize(this.httpClient);
 
+#if !INTERNAL
+            // Starting ClientTelemetry Job
+            this.TelemetryToServiceHelper = TelemetryToServiceHelper.CreateAndInitializeClientConfigAndTelemetryJob(this.clientId,
+                                                                 this.ConnectionPolicy,
+                                                                 this.cosmosAuthorization,
+                                                                 this.httpClient,
+                                                                 this.ServiceEndpoint,
+                                                                 this.GlobalEndpointManager,
+                                                                 this.cancellationTokenSource);
+#endif
             if (sessionContainer != null)
             {
                 this.sessionContainer = sessionContainer;
@@ -960,12 +971,6 @@ namespace Microsoft.Azure.Cosmos
             // For gateway: GatewayProxy.
             // For direct: WFStoreProxy [set in OpenAsync()].
             this.eventSource = DocumentClientEventSource.Instance;
-
-            // Disable system usage for internal builds. Cosmos DB owns the VMs and already logs
-            // the system information so no need to track it.
-#if !INTERNAL
-            this.InitializeClientTelemetry();
-#endif
 
             this.initializeTaskFactory = (_) => TaskHelper.InlineIfPossible<bool>(
                     () => this.GetInitializationTaskAsync(storeClientFactory: storeClientFactory),
@@ -1028,7 +1033,7 @@ namespace Microsoft.Azure.Cosmos
                     storeModel: this.GatewayStoreModel, 
                     tokenProvider: this, 
                     retryPolicy: this.retryPolicy,
-                    clientTelemetry: this.clientTelemetry);
+                    telemetryToServiceHelper: this.TelemetryToServiceHelper);
             this.partitionKeyRangeCache = new PartitionKeyRangeCache(this, this.GatewayStoreModel, this.collectionCache);
             this.ResetSessionTokenRetryPolicy = new ResetSessionTokenRetryPolicyFactory(this.sessionContainer, this.collectionCache, this.retryPolicy);
 
@@ -1044,36 +1049,6 @@ namespace Microsoft.Azure.Cosmos
             }
 
             return true;
-        }
-
-        private void InitializeClientTelemetry()
-        {
-            if (this.ConnectionPolicy.EnableClientTelemetry)
-            {
-                try
-                {
-                    this.clientTelemetry = ClientTelemetry.CreateAndStartBackgroundTelemetry(
-                        clientId: this.clientId,
-                        httpClient: this.httpClient,
-                        userAgent: this.ConnectionPolicy.UserAgentContainer.BaseUserAgent,
-                        connectionMode: this.ConnectionPolicy.ConnectionMode,
-                        authorizationTokenProvider: this.cosmosAuthorization,
-                        diagnosticsHelper: DiagnosticsHandlerHelper.Instance,
-                        preferredRegions: this.ConnectionPolicy.PreferredLocations,
-                        globalEndpointManager: this.GlobalEndpointManager);
-
-                    DefaultTrace.TraceInformation("Client Telemetry Enabled.");
-                }
-                catch (Exception ex)
-                {
-                    DefaultTrace.TraceInformation($"Error While starting Telemetry Job : {ex.Message}. Hence disabling Client Telemetry");
-                    this.ConnectionPolicy.EnableClientTelemetry = false;
-                }
-            }
-            else
-            {
-                DefaultTrace.TraceInformation("Client Telemetry Disabled.");
-            }
         }
 
         private async Task InitializeCachesAsync(string databaseName, DocumentCollection collection, CancellationToken cancellationToken)
@@ -1329,6 +1304,12 @@ namespace Microsoft.Azure.Cosmos
                 this.cosmosAuthorization.Dispose();
             }
 
+            if (this.TelemetryToServiceHelper != null)
+            {
+                this.TelemetryToServiceHelper.Dispose();
+                this.TelemetryToServiceHelper = null;
+            }
+
             if (this.GlobalEndpointManager != null)
             {
                 this.GlobalEndpointManager.Dispose();
@@ -1344,11 +1325,6 @@ namespace Microsoft.Azure.Cosmos
             {
                 this.initTaskCache.Dispose();
                 this.initTaskCache = null;
-            }
-
-            if (this.clientTelemetry != null)
-            {
-                this.clientTelemetry.Dispose();
             }
 
             DefaultTrace.TraceInformation("DocumentClient with id {0} disposed.", this.traceId);
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -565,11 +565,11 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// Internal constructor purely for unit-testing
         /// </summary>
-        internal DocumentClient(Uri serviceEndpoint, string authKey, ConnectionPolicy connectionPolicy = null)
+        internal DocumentClient(Uri serviceEndpoint, string authKey)
         {
             // do nothing 
             this.ServiceEndpoint = serviceEndpoint;
-            this.ConnectionPolicy = connectionPolicy ?? new ConnectionPolicy();
+            this.ConnectionPolicy = new ConnectionPolicy();
         }
 
         internal virtual async Task<ClientCollectionCache> GetCollectionCacheAsync(ITrace trace)
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -565,11 +565,11 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// Internal constructor purely for unit-testing
         /// </summary>
-        internal DocumentClient(Uri serviceEndpoint, string authKey)
+        internal DocumentClient(Uri serviceEndpoint, string authKey, ConnectionPolicy connectionPolicy = null)
         {
             // do nothing 
             this.ServiceEndpoint = serviceEndpoint;
-            this.ConnectionPolicy = new ConnectionPolicy();
+            this.ConnectionPolicy = connectionPolicy ?? new ConnectionPolicy();
         }
 
         internal virtual async Task<ClientCollectionCache> GetCollectionCacheAsync(ITrace trace)
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -1253,6 +1253,12 @@ namespace Microsoft.Azure.Cosmos
                 return;
             }
 
+            if (this.telemetryToServiceHelper != null)
+            {
+                this.telemetryToServiceHelper.Dispose();
+                this.telemetryToServiceHelper = null;
+            }
+
             if (!this.cancellationTokenSource.IsCancellationRequested)
             {
                 this.cancellationTokenSource.Cancel();
@@ -1301,12 +1307,6 @@ namespace Microsoft.Azure.Cosmos
             if (this.cosmosAuthorization != null)
             {
                 this.cosmosAuthorization.Dispose();
-            }
-
-            if (this.telemetryToServiceHelper != null)
-            {
-                this.telemetryToServiceHelper.Dispose();
-                this.telemetryToServiceHelper = null;
             }
 
             if (this.GlobalEndpointManager != null)
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -143,7 +143,7 @@ namespace Microsoft.Azure.Cosmos
 
         internal CosmosAccountServiceConfiguration accountServiceConfiguration { get; private set; }
 
-        internal TelemetryToServiceHelper TelemetryToServiceHelper { get; set; }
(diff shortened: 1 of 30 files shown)


Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1329,6 +1304,12 @@ public void Dispose()
                 this.cosmosAuthorization.Dispose();
             }
 
+            if (this.TelemetryToServiceHelper != null)
```
kirankumarkolli, line 1307: HttpClient is also used by TelemetryServiceHelper and HttpClient should be after from dependency wise right?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -939,6 +940,16 @@ private async Task OpenPrivateAsync(CancellationToken cancellationToken)
             // Loading VM Information (non blocking call and initialization won't fail if this call fails)
             VmMetadataApiHandler.TryInitialize(this.httpClient);
 
+#if !INTERNAL
```
kirankumarkolli, line 943: For compute it will be NULL and results in NullReference for CollectionCaches. Thoughts?

kirankumarkolli, line 943: #if !INTERNAL

TelemetryToServiceHelper.CreateAndInitializeClientConfigAndTelemetryJob implementation it-self is covering for INTERNAL, is above precondition here needed?

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
```diff
@@ -46,7 +46,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-		<None Include="..\..\LICENSE" Pack="true" PackagePath="" Visible="false"/>
+		<None Include="..\..\LICENSE" Pack="true" PackagePath="" Visible="false" />
```
kirankumarkolli, line 49: Spaces only change, not needed, undo.

Microsoft.Azure.Cosmos/src/Routing/ClientCollectionCache.cs
```diff
@@ -214,21 +215,19 @@ internal class ClientCollectionCache : CollectionCache
...
-                                                    containerId: collectionName,
-                                                    operationType: request.OperationType,
-                                                    resourceType: request.ResourceType,
-                                                    subStatusCode: response.SubStatusCode,
-                                                    databaseId: databaseName);
-                                }
+                                
+                                this.telemetryToServiceHelper.GetCollector().CollectCacheInfo(
+                                     ClientCollectionCache.TelemetrySourceName,
+                                     () => new TelemetryInformation
```
kirankumarkolli, line 221: ClientCollectionCache.GetDatabaseAndCollectionName is not used and dead code.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -230,46 +228,42 @@ internal static async Task RunProcessorTaskAsync(string telemetryDate, Task proc
...
-            if (string.IsNullOrEmpty(cacheRefreshSource))
+            if (string.IsNullOrEmpty(cacheName))
             {
-                throw new ArgumentNullException(nameof(cacheRefreshSource));
+                throw new ArgumentNullException(nameof(cacheName));
             }
 
-            DefaultTrace.TraceVerbose($"Collecting cacheRefreshSource {0} data for Telemetry.", cacheRefreshSource);
+            // If latency information is not available. Ignore this datapoint
+            if (!data.RequestLatency.HasValue)
```
kirankumarkolli, line 239: Is latency based filtering only for collectionCache interactions.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,130 @@
...
+                    preferredRegions: this.connectionPolicy.PreferredLocations,
+                    globalEndpointManager: this.globalEndpointManager);
+
+                this.collector = new TelemetryCollector(this.clientTelemetry, this.connectionPolicy);
+
+                DefaultTrace.TraceVerbose("Client Telemetry Enabled.");
+            }
+            catch (Exception ex)
+            {
+                DefaultTrace.TraceInformation($"Error While starting Telemetry Job : {ex.Message}. Hence disabling Client Telemetry");
```
kirankumarkolli, line 112: This needs to be at-least warning level right?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,130 @@
...
+        public void Dispose()
+        {
+            this.StopClientTelemetry();
+        }
+
+        private void StopClientTelemetry()
+        {
+            this.collector = new TelemetryCollectorNoOp();
+
+            this.clientTelemetry?.Dispose();
```
kirankumarkolli, line 126: Explicit Dispose had concurrency issues right? 

Since we only expect one clientTelemetry, dispose can deferred to the Finalizer.

kirankumarkolli (changes requested):
Please check comments

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: removing it. it is already handled inside.

sourabh1007: removed this one and kept inside one. So we don't have to deal with null helper anywhere.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
sourabh1007: missed to push :(..will push now

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: No...if latency info is missing..I am skipping that datapoint for operation telemetry also...

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        public void Dispose()
+        {
+            this.StopClientTelemetry();
+        }
+
+        private void StopClientTelemetry()
+        {
+            this.collector = new TelemetryCollectorNoOp();
```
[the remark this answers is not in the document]
sourabh1007: added documentation. . anyway after other PR's it should become clearer automatically.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: I don't think it matters, `httpclient ` will anyway throw exception if there is any inflight http call and for that, already try catch is there.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: I am not sure about concurrency issue here, as dispose get called only once. It is an attempt to stop everything using cancellation token (which is inside dispose).

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        public static TelemetryToServiceHelper CreateAndInitializeClientConfigAndTelemetryJob(string clientId,
+           ConnectionPolicy connectionPolicy,
+           AuthorizationTokenProvider cosmosAuthorization,
+           CosmosHttpClient httpClient,
+           Uri serviceEndpoint,
+           GlobalEndpointManager globalEndpointManager,
+           CancellationTokenSource cancellationTokenSource)
+        {
+#if INTERNAL
```
kirankumarkolli, line 62: INTERNAL is spilled through. 
Can this be handled at the connectionPolicy level it-self? It will naturally just flow through right?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+            return helper;
+#endif
+        }
+
+        public ITelemetryCollector GetCollector()
+        {
+            return this.collector;
+        }
+
+        public bool IsClientTelemetryJobRunning()
```
kirankumarkolli, line 84: With collectors the telemetry is running or not is implementation details right? 

The only scenarios is when the client is explicitly opting out at the client scope. And that the caller (PipelineBuiler) can base the decision on the ClientOptions right?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
kirankumarkolli: Assuming that only one instance of ClientTelemetry is created and used by only instance of TelemetryCollector then we can delegate the dispose requirement to TelemetryCollector. Which the finalizer will take care of right?

Microsoft.Azure.Cosmos/src/Telemetry/Collector/TelemetryCollector.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                    data.DatabaseId = databaseName;
+                    data.ContainerId = collectionName;
+                }
+
+                this.clientTelemetry?.PushCacheDatapoint(cacheName, data);
+            }
+            catch (Exception ex)
+            {
+                DefaultTrace.TraceError($"Error while collecting cache {cacheName} telemetry. Exception : {ex}");
```
kirankumarkolli, line 45: nit: Move all traces to argument based to avoid unnecessary materialization.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
kirankumarkolli: Is it expected to not have latency? If not how about at-least tracing?

kirankumarkolli: As system telemetry do such data points help us proactively discover the gaps? Thoughts (follow-up)

Microsoft.Azure.Cosmos/src/Handler/ClientPipelineBuilder.cs
```diff
@@ -48,9 +48,9 @@ internal class ClientPipelineBuilder
 #else
             this.diagnosticsHandler = null;
 #endif
-            if (telemetry != null)
+            if (telemetryToServiceHelper.IsClientTelemetryJobRunning())
```
kirankumarkolli, line 51: Eventually the only time handler is unexpected is when client explicitly disables right? 
If so can it be based on the client options?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1329,6 +1303,12 @@ public void Dispose()
                 this.cosmosAuthorization.Dispose();
             }
 
+            if (this.telemetryToServiceHelper != null)
```
kirankumarkolli, line 1306: telemetryToServiceHelper  uses HttpClient. So telemetryToServiceHelper  needs to be disposed before httpclient. 

Please do check the lifeitmes dependencies in all places.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: No it is not spilled..it just used to decide if telemetry helper should be no op or not..that's it.after that it works naturally....it makes sure that pkg doesn't even contain telemetry related code...

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: No once we have API integration.....this will change...opt in/out will be based on API response.

Microsoft.Azure.Cosmos/src/Handler/ClientPipelineBuilder.cs
sourabh1007: And when it is disabled by client config API

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: Exception during httclient dispose is already handled.

If you notice other components like gem, authorisation is also getting disposed later, I believe they also make http calls like client telemetry

Microsoft.Azure.Cosmos/src/Handler/ClientPipelineBuilder.cs
sourabh1007: And along with that, i don' t want to spill connection policy and client option check in the code.... telemetry helper is already deciding, what to do based on the connection policy....

Microsoft.Azure.Cosmos/src/DocumentClient.cs
kirankumarkolli: Agree that other components have this issue. Ideally then we should move down the httpclient right? (Can be a separate PR if needed). 

/cc: @ealsur thoughts?

Microsoft.Azure.Cosmos/src/Handler/ClientPipelineBuilder.cs
kirankumarkolli: What's the necessity of the pipeline conditional composition? 

If collection overhead is almost ignorable in-case of dynamic disable, then the same model be used here as well right? 
i.e. Handler will always be present the collector will ignore it. In that case the check in Pipelinebuiilder is unnecessary right? 

If above assumption is not true, few alternatives are
1. Pipeline is dynamically rebuilt: Today we don't modify the pipeline at-all. Its a much bigger impactful change. 
2. TelemetryHandler always present but will be a simple pass-through (i.e. no overhead) when disabled. 

In any case pipeline conditional composition is not required right?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
kirankumarkolli: If eventually (i.e. after one more PR) pipeline will not be conditional then why to even start with it now?

check: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4037#discussion_r1299228622

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: Let's take this up as separate PR I will create a bug for this.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: Added traces and comment.  It is not expected to have null latency, since it is `nullable`, hence put this null check.

Microsoft.Azure.Cosmos/src/Handler/ClientPipelineBuilder.cs
sourabh1007: removed the condition.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: removed this condition.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: We need to dispose it...as it is cancelling background client telemetry job. I have moved dispose of telemetry at the top..to get it disposed first.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: We need to dispose it...as it is cancelling background client telemetry job. I have moved dispose of telemetry at the top..to get it disposed first.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -215,9 +213,9 @@ internal static async Task RunProcessorTaskAsync(string telemetryDate, Task proc
                 Task resultTask = await Task.WhenAny(processingTask, delayTask);
                 if (resultTask == delayTask)
                 {
-                    DefaultTrace.TraceVerbose($"Processor task with date as {telemetryDate} is canceled as it did not finish in {timeout}");
+                    DefaultTrace.TraceVerbose($"Processor task with date as {0} is canceled as it did not finish in {1}", telemetryDate, timeout);
                     // Operation cancelled
-                    throw new OperationCanceledException($"Processor task with date as {telemetryDate} is canceled as it did not finish in {timeout}");
+                    throw new OperationCanceledException(string.Format($"Processor task with date as {0} is canceled as it did not finish in {1}", telemetryDate, timeout));
```
kirankumarkolli, line 218: Exception inlining is okey, which is expected to materialize always anyways right?

kirankumarkolli (dismissed):

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Performance.Tests/Contracts/BenchmarkResults.json
```diff
@@ -39,7 +39,7 @@
   "MockedItemBenchmark.ReadItemNotExists;[Type=OfT]": 43489.25,
   "MockedItemBenchmark.ReadItemNotExists;[Type=OfTCustom]": 43490,
   "MockedItemBenchmark.ReadItemNotExists;[Type=OfTWithClientTelemetryEnabled]": 43489.25,
-  "MockedItemBenchmark.ReadItemNotExists;[Type=OfTWithDiagnosticsToString]": 58054,
+  "MockedItemBenchmark.ReadItemNotExists;[Type=OfTWithDiagnosticsToString]": 64764.75,
```
kirankumarkolli, line 42: Is it with collection OOF? 

Is it a clickstop to get back to with future PRs?

sourabh1007: yes, it is when collection is OFF. 

I am not sure if it is due to any change in this PR and that too only for `ReadItemNotExists`. These numbers keep fluctuating.

sourabh1007: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4049 created issue to track this. I will look into once all the functional PRs are merged.

kirankumarkolli (approved):

PR #4043, opened 2023-08-16
[Internal] Client Telemetry: Adds condition to make sure metrics is always there

sourabh1007:
## Description
Client Telemetry Service do not accept partial metric information except `SystemPool_IsThreadStarving_True` metrics because service knows this particular metrics doesn't have percentile calculated.

As part of this PR, making sure metrics is added in payload only if there is metrics information available. It means if corresponding histogram has any value then only do any calculations otherwise just skip that metrics.


sourabh1007:
Closing this as this is stale

PR #4047, opened 2023-08-21
[Internal] Automation: Adds logic to tag customer-reported issues

ealsur:
When a user without write/admin access creates an issue, tag the issue with `customer-reported`.

A user that is part of our team would have `write` and in some cases `admin` permissions, but a user that is external (cannot push branches on the repo) has either `none` or `read` permissions.

Based on `none`/`read` we can identify that this is an external users, hence a "customer".

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -39,6 +39,19 @@ configuration:
           label: needs-investigation
       description: 
     - if:
+      - payloadType: Issues
+      - isAction:
+          action: Opened
+      - or:
+        - activitySenderHasPermission:
+            permission: Read
+        - activitySenderHasPermission:
+            permission: None
+      then:
+      - addLabel:
+          label: customer-reported
+      description: Identifies issues created by users without write access as customers
+    - if:
       - payloadType: Pull_Request
       - hasLabel:
           label: auto-merge


.github/policies/resourceManagement.yml
```diff
@@ -38,6 +38,19 @@ configuration:
...
           label: needs-investigation
       description: 
+    - if:
+      - payloadType: Issues
+      - isAction:
+          action: Opened
+      - or:
+        - activitySenderHasPermission:
+            permission: Read
+        - activitySenderHasPermission:
```
kirankumarkolli, line 48: Is this condition supposed to be sendsederHadWritePermission?

ealsur: > Am struggling to understand it.
> 
> Everyone have Read permission right? So first part would always be true and OR condition always evaluates to true right?

This is a good question. What you are asking is, a user that is Admin (for example), will it also have Read? I'll check the docs, my understanding is that this checks the uppermost priviledge.

kirankumarkolli: If it's the MAX permission, then above OR condition makes sense.

jcocchi (dismissed):

.github/policies/resourceManagement.yml
ealsur: Circled back to confirm with the owners of the policies and it seems the condition should be:

```
      - not:
          or:
            - activitySenderHasPermission:
              permission: write
            - activitySenderHasPermission:
              permission: admin
```

PR #4050, opened 2023-08-23
[Internal] Client Telemetry: Adds client config api call to get latest flag status

sourabh1007:
## Description

This PR is part 2/3 of this https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3816 PR.

As part of PR, Removing Environment variable configs and using client config API response to enable or disable client telemetry job.

**What is not covered?** dynamically switching on/off client telemetry flag and start/stop job in sdk accordingly.

**Why so many files are touched?**
As part of this PR, changing the way we enabled telemetry feature. Since Previously, it was only possible from Environment Variable, But now **ClientConfig** API gives this information.

Hence summary of the change is :
1. `ConnectionPolicy.EnableClientTelemetry` and `CosmosClientOptions.EnableClientTelemetry` (Internal Options) is now a feature flag. It just unlocks this whole feature which includes calling client config API and get the latest status of the flag. _(Public contract will be implemented as separate PR)_
2. `TelemetryToServiceHelper.cs` : This file contains the ClientConfig API call and initializes the client telemetry job _(dynamically enable/disable will be covered as separate PR)_
3. **CTL Project:** Previously CTL project was setting environment variables. Now, it will use default SDK functionality i.e. client config API response, to decide if telemetry job should be triggered or not. Feature flag can be passed as an argument.
4. **Benchmark Project:** Previously benchmark project was setting environment variables based on the config. Now, it will use it will use default SDK functionality i.e., client config API response, to decide if telemetry job should be triggered or not. Feature flag can be passed as an argument.
5. `DiagnosticHandlerHelper`: Previously it was checking environment variable to identify if telemetry is enabled or not and adding recorder to the monitor accordingly. Now, added a new API to refresh the handler instance, if this feature is switched. _(It is written to keep next PR i.e. dynamically enable/disable of feature, in mind)_. Added test for the same.
6. **Performance Test:** Previously client telemetry was enabled by default for Performance Test. Now, added mocking of external API calls when instance with client telemetry is requested.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
@@ -4,6 +4,7 @@
 
 namespace Microsoft.Azure.Cosmos.Handlers
 {
+    using System;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Azure.Cosmos.Handler;
@@ -26,6 +27,9 @@ namespace Microsoft.Azure.Cosmos.Handlers
 
             // Record the diagnostics on the response to get the CPU of when the request was executing
             SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance().GetDiagnosticsSystemHistory();
+
+            Console.WriteLine("DiagnosticsHandler: is systemUsageHistory null? {0}", systemUsageHistory == null);
+
             if (systemUsageHistory != null)
             {
                 request.Trace.AddDatum(
--- a/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
@@ -28,8 +28,6 @@ namespace Microsoft.Azure.Cosmos.Handlers
             // Record the diagnostics on the response to get the CPU of when the request was executing
             SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance().GetDiagnosticsSystemHistory();
 
-            Console.WriteLine("DiagnosticsHandler: is systemUsageHistory null? {0}", systemUsageHistory == null);
-
             if (systemUsageHistory != null)
             {
                 request.Trace.AddDatum(
--- a/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.Cosmos.Handlers
             ResponseMessage responseMessage = await base.SendAsync(request, cancellationToken);
 
             // Record the diagnostics on the response to get the CPU of when the request was executing
-            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.Instance.GetDiagnosticsSystemHistory();
+            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance().GetDiagnosticsSystemHistory();
             if (systemUsageHistory != null)
             {
                 request.Trace.AddDatum(
--- a/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.Cosmos.Handlers
             ResponseMessage responseMessage = await base.SendAsync(request, cancellationToken);
 
             // Record the diagnostics on the response to get the CPU of when the request was executing
-            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance()?.GetDiagnosticsSystemHistory();
+            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance().GetDiagnosticsSystemHistory();
 
             if (systemUsageHistory != null)
             {
--- a/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.Cosmos.Handlers
             ResponseMessage responseMessage = await base.SendAsync(request, cancellationToken);
 
             // Record the diagnostics on the response to get the CPU of when the request was executing
-            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance().GetDiagnosticsSystemHistory();
+            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance()?.GetDiagnosticsSystemHistory();
 
             if (systemUsageHistory != null)
             {
--- a/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
@@ -45,8 +45,8 @@ namespace Microsoft.Azure.Cosmos.Handler
             new DiagnosticsHandlerHelper();
 #endif
 
-        private static bool isDiagnosticsMonitoringEnabled = false;
-        private static bool isTelemetryMonitoringEnabled = false;
+        private static bool isDiagnosticsMonitoringEnabled;
+        private static bool isTelemetryMonitoringEnabled;
 
         private readonly SystemUsageMonitor systemUsageMonitor = null;
 
@@ -61,7 +61,6 @@ namespace Microsoft.Azure.Cosmos.Handler
         /// <param name="isClientTelemetryEnabled"></param>
         public static void Refresh(bool isClientTelemetryEnabled)
         {
-            Console.WriteLine("DiagnosticsHandlerHelper.Refresh " + isClientTelemetryEnabled);
             if (isClientTelemetryEnabled != DiagnosticsHandlerHelper.isTelemetryMonitoringEnabled)
             {
                 // Update telemetry flag
@@ -70,7 +69,7 @@ namespace Microsoft.Azure.Cosmos.Handler
                 // Create new instance, it will start a new system monitor job
                 DiagnosticsHandlerHelper newInstance = new DiagnosticsHandlerHelper();
 
-                // Create a new refrence to the old instance
+                // Create a new reference to the old instance
                 DiagnosticsHandlerHelper oldInstance = DiagnosticsHandlerHelper.Instance;
 
                 // Assing new instance to the static refrence,
@@ -79,8 +78,6 @@ namespace Microsoft.Azure.Cosmos.Handler
                 // Stop and dispose old instance of SystemMonitor job
                 oldInstance.StopSystemMonitor();
             }
-
-            Console.WriteLine("DiagnosticsHandlerHelper.Refreshed");
         }
 
         private void StopSystemMonitor()
@@ -99,16 +96,14 @@ namespace Microsoft.Azure.Cosmos.Handler
             // If the CPU monitor fails for some reason don't block the application
             try
             {
-                Console.WriteLine("Added diag recorder");
                 List<SystemUsageRecorder> recorders = new List<SystemUsageRecorder>()
                 {
-                    DiagnosticSystemUsageRecorder,
+                    DiagnosticsHandlerHelper.DiagnosticSystemUsageRecorder,
                 };
 
                 if (DiagnosticsHandlerHelper.isTelemetryMonitoringEnabled)
                 {
-                    Console.WriteLine("Added telemetry recorder");
-                    recorders.Add(TelemetrySystemUsageRecorder);
+                    recorders.Add(DiagnosticsHandlerHelper.TelemetrySystemUsageRecorder);
                 }
 
                 this.systemUsageMonitor = SystemUsageMonitor.CreateAndStart(recorders);
@@ -120,7 +115,6 @@ namespace Microsoft.Azure.Cosmos.Handler
                 DefaultTrace.TraceError(ex.Message);
 
                 DiagnosticsHandlerHelper.isDiagnosticsMonitoringEnabled = false;
-                DiagnosticsHandlerHelper.isTelemetryMonitoringEnabled = false;
             }
         }
 
@@ -137,7 +131,7 @@ namespace Microsoft.Azure.Cosmos.Handler
 
             try
             {
-                return DiagnosticSystemUsageRecorder.Data;
+                return DiagnosticsHandlerHelper.DiagnosticSystemUsageRecorder.Data;
             }
             catch (Exception ex)
             {
@@ -161,12 +155,11 @@ namespace Microsoft.Azure.Cosmos.Handler
 
             try
             {
-                return TelemetrySystemUsageRecorder.Data;
+                return DiagnosticsHandlerHelper.TelemetrySystemUsageRecorder.Data;
             }
             catch (Exception ex)
             {
                 DefaultTrace.TraceError(ex.Message);
-                DiagnosticsHandlerHelper.isTelemetryMonitoringEnabled = false;
                 return null;
             }
         }
--- a/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
@@ -33,22 +33,56 @@ namespace Microsoft.Azure.Cosmos.Handler
             historyLength: 120,
             refreshInterval: DiagnosticsHandlerHelper.ClientTelemetryRefreshInterval);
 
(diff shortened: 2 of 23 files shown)


Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
```diff
@@ -25,7 +26,10 @@ internal class DiagnosticsHandler : RequestHandler
             ResponseMessage responseMessage = await base.SendAsync(request, cancellationToken);
 
             // Record the diagnostics on the response to get the CPU of when the request was executing
-            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.Instance.GetDiagnosticsSystemHistory();
+            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance().GetDiagnosticsSystemHistory();
+
+            Console.WriteLine("DiagnosticsHandler: is systemUsageHistory null? {0}", systemUsageHistory == null);
```
ealsur, line 31: Remove Console.WriteLine

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
```diff
@@ -22,34 +22,69 @@ internal class DiagnosticsHandlerHelper
         private const string Telemetrykey = "telemetry";
 
         public static readonly TimeSpan DiagnosticsRefreshInterval = TimeSpan.FromSeconds(10);
-        private readonly SystemUsageRecorder diagnosticSystemUsageRecorder = new SystemUsageRecorder(
+        private static readonly SystemUsageRecorder DiagnosticSystemUsageRecorder = new SystemUsageRecorder(
```
ealsur, line 25: Why is the static change needed?

sourabh1007: on refresh, I don't want to loose old information. making this static make recorded information available in new recorder also after the refresh. added test for the same.

Microsoft.Azure.Cosmos/src/Resource/Settings/AccountClientConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        [JsonProperty(PropertyName = Constants.Properties.ClientTelemetryConfiguration)]
+        public ClientTelemetryConfiguration ClientTelemetryConfiguration { get; set; }
+
+        /// <summary>
+        /// This contains additional values for scenarios where the SDK is not aware of new fields. 
+        /// This ensures that if resource is read and updated none of the fields will be lost in the process.
+        /// </summary>
+        [JsonExtensionData]
+        internal IDictionary<string, JToken> AdditionalProperties { get; private set; }
```
ealsur, line 22: nit: mixture of internal and public in an internal class, should all be public methods and properties?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -70,12 +75,69 @@ private TelemetryToServiceHelper()
...
-            helper.InitializeClientTelemetry();
+            _ = helper.RetrieveConfigAndInitiateTelemetryAsync(); // Let it run in backgroud
 
             return helper;
 #endif
         }
 
+        private async Task RetrieveConfigAndInitiateTelemetryAsync()
+        {
+            try
```
ealsur, line 86: Shouldn't this be a loop? Otherwise this runs once?

ealsur: How does it detect if it was initially disabled (so `ClientTelemetry.CreateAndStartBackgroundTelemetry` is never called) and then was enabled?

Microsoft.Azure.Cosmos/src/Resource/Settings/AccountClientConfiguration.cs
sourabh1007: I don't think we are following any kind of pattern here and anyway it doesn't matter, if class type is `internal `then all its `public `properties will also be resolved to `internal`.
Along with this, noticed that `AdditionalProperties` is always `internal `and other properties are either `public` or `internal`. For reference, https://github.com/Azure/azure-cosmos-dotnet-v3/blob/83dd4b54348a4b2d172d1bbcfd830134363e212b/Microsoft.Azure.Cosmos/src/Resource/Settings/AccountRegion.cs#L14

Microsoft.Azure.Cosmos/src/Resource/Settings/ClientTelemetryConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        [JsonProperty(PropertyName = Constants.Properties.ClientTelemetryEndpoint)]
+        public string Endpoint { get; set; }
+
+        /// <summary>
+        /// This contains additional values for scenarios where the SDK is not aware of new fields. 
+        /// This ensures that if resource is read and updated none of the fields will be lost in the process.
+        /// </summary>
+        [JsonExtensionData]
+        internal IDictionary<string, JToken> AdditionalProperties { get; private set; }
```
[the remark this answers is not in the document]
sourabh1007: same as above.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
sourabh1007: yes. That will be covered in next PR, here focus is on replacing env variable with client config API call (calling it only once), to get the client telemetry status.

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
```diff
@@ -22,34 +22,71 @@ internal class DiagnosticsHandlerHelper
...
+            try
+            {
+                this.systemUsageMonitor?.Dispose();
+            }
+            catch (ObjectDisposedException ex)
+            {
+                DefaultTrace.TraceError($"Error while stopping system usage monitor. Stacktrace: {0} ", ex.StackTrace);
+            }
+
+            DiagnosticsHandlerHelper.Instance = null; // refrence is lost, GC will collect it
```
ealsur, line 87: Why do we need to set the Instance to `null`? This is supposedly just stopping the system monitor, not clearing the instance, right?

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
```diff
@@ -22,34 +22,71 @@ internal class DiagnosticsHandlerHelper
...
+
+        private void StopSystemMonitor()
+        {
+            try
+            {
+                this.systemUsageMonitor?.Dispose();
+            }
+            catch (ObjectDisposedException ex)
+            {
+                DefaultTrace.TraceError($"Error while stopping system usage monitor. Stacktrace: {0} ", ex.StackTrace);
```
ealsur, line 84:
```suggestion
                DefaultTrace.TraceError($"Error while stopping system usage monitor. {0} ", ex);
```

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
```diff
@@ -25,7 +26,8 @@ internal class DiagnosticsHandler : RequestHandler
             ResponseMessage responseMessage = await base.SendAsync(request, cancellationToken);
 
             // Record the diagnostics on the response to get the CPU of when the request was executing
-            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.Instance.GetDiagnosticsSystemHistory();
+            SystemUsageHistory systemUsageHistory = DiagnosticsHandlerHelper.GetInstance()?.GetDiagnosticsSystemHistory();
```
ealsur, line 29: nit: When can this be null? It starts with an instance and it gets swapped when Refresh is called

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Azure.Cosmos.Handler
     using System.Collections.Generic;
     using Documents.Rntbd;
     using Microsoft.Azure.Cosmos.Core.Trace;
-    using Microsoft.Azure.Cosmos.Telemetry;
+    using Microsoft.Azure.Cosmos.Query.Core.Pipeline;
```
ealsur, line 11: nit: What is using this namespace?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -89,28 +151,34 @@ public bool IsClientTelemetryJobRunning()
...
+                        globalEndpointManager: this.globalEndpointManager,
+                        endpointUrl: clientConfig.ClientTelemetryConfiguration.Endpoint);
+
+                    this.collector = new TelemetryCollector(this.clientTelemetry, this.connectionPolicy);
+
+                    DefaultTrace.TraceVerbose("Client Telemetry Enabled.");
+                }
+                catch (Exception ex)
+                {
+                    DefaultTrace.TraceWarning($"Error While starting Telemetry Job : {0}. Hence disabling Client Telemetry", ex.Message);
```
ealsur, line 179:
```suggestion
                    DefaultTrace.TraceWarning($"Error While starting Telemetry Job : {0}. Hence disabling Client Telemetry", ex);
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTestsBase.cs
```diff
@@ -53,13 +54,7 @@ public abstract class ClientTelemetryTestsBase : BaseCosmosClientHelper
...
         public static void ClassInitialize(TestContext _)
         {
-            Util.EnableClientTelemetryEnvironmentVariables();
-
-            SystemUsageMonitor oldSystemUsageMonitor = (SystemUsageMonitor)typeof(DiagnosticsHandlerHelper)
-                .GetField("systemUsageMonitor", BindingFlags.Instance | BindingFlags.NonPublic).GetValue(DiagnosticsHandlerHelper.Instance);
-            oldSystemUsageMonitor.Stop();
-
-            ClientTelemetryTestsBase.ResetSystemUsageMonitor(true);
+            ClientTelemetryOptions.DefaultIntervalForTelemetryJob = TimeSpan.FromSeconds(1);
```
ealsur, line 57: nit: Should this be undone at the end of the test run? Can this impact other tests (being a static property)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/HttpHandlerHelper.cs
```diff
@@ -52,6 +52,8 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                     throw;
                 }
                 this.ExceptionIntercepter.Invoke(request, ex);
+
+                throw; // Anyway throw this exception
```
ealsur, line 56: What prompted this change?

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
sourabh1007: This class responsibility is to run the system monitor and give stats generated by this monitor. If monitor is stopped, it means, this class is not doing what it needs to do, means it is irrelevant to have this class instance because old stats is of no use for me. 
Hence, I set this instance as null here and along with that, I know I am calling this function only when I refresh the instance, so current instance is of no use for me.

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
sourabh1007: it is mostly a safety check, because in case of internal package when we initialize the helper then it is null and during refresh also, there is a moment when this instance can be `null`.

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
sourabh1007: removed...

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryOptions.cs
```diff
@@ -99,53 +95,11 @@ internal static class ClientTelemetryOptions
         internal static readonly List<int> ExcludedStatusCodes = new List<int> { 404, 409, 412 };
 
         internal static readonly int NetworkTelemetrySampleSize = 200;
+        internal static TimeSpan DefaultIntervalForTelemetryJob = TimeSpan.FromMinutes(10);
```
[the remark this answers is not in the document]
sourabh1007: Might be, I can see it as a candidate to become configurable (not from customer POV but we should be able to configure it for an account)

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
```diff
@@ -419,7 +419,7 @@ public void UpdateSystemUsage()
                 this.systemUsageHistory.Values.Count == 0 ||
                 this.systemUsageHistory.LastTimestamp + DiagnosticsHandlerHelper.DiagnosticsRefreshInterval < DateTime.UtcNow)
             {
-                this.systemUsageHistory = DiagnosticsHandlerHelper.Instance.GetDiagnosticsSystemHistory();
+                this.systemUsageHistory = DiagnosticsHandlerHelper.GetInstance()?.GetDiagnosticsSystemHistory();
```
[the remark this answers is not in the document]
sourabh1007: it is mostly a safety check, because in case of internal package when we initialize the helper then it is null and during refresh also, there is a moment when this instance can be null.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTestsBase.cs
sourabh1007: it should not impact other tests but getting it undone is right thing to do here. Made this change.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/HttpHandlerHelper.cs
sourabh1007: during few debugging sessions, I realized this change. surprisingly it didn't affect anything yet but found it missing hence added.

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
ealsur: But what is the point? You are removing the reference, potentially causing NullRefs that you need to check all over with `DiagnosticsHandlerHelper.GetInstance()?` (the added ?). 

It is weird that a method that is not the Dispose is getting rid of a reference.

If this is only called from `Refresh` then the reference is already replaced by `DiagnosticsHandlerHelper.Instance = new DiagnosticsHandlerHelper();`, doing this makes no difference and just forces you to check for NullRefs

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
ealsur: Before this PR, the `DiagnosticsHandlerHelper.Instance` could also be `null` and it was never checked, because the `DiagnosticsHandler` is not added to the pipeline:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/668d71a49c7109da846dae94e194954b5a2bb718/Microsoft.Azure.Cosmos/src/Handler/ClientPipelineBuilder.cs#L45-L50

The null check is not needed if you remove the `= null` in the `DiagnosticsHandlerHelper`

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandlerHelper.cs
sourabh1007: removed null assignments.

Microsoft.Azure.Cosmos/src/Handler/DiagnosticsHandler.cs
sourabh1007: removed null assignments.

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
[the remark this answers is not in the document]
sourabh1007: removed null assignments.

ealsur (approved):

PR #4054, opened 2023-08-25
[Internal] Query: Fixes escaped string parsing in SqlParser

adityasa:
# Query: [Internal] Fixs escaped string parsing in SqlParser

## Description

This change fixes escaped string parsing in SqlParser. Currently the parsing logic only escapes " character and none of the other characters. This leaves both \, / prepended with escape character \ but for " leading \ is removed (which is correct behavior). When the string literal DOM is serialized (ToString), the leading escape character is escaped again causing inconsistency between original string and new string.
Input : "\\/Solidus" -> Parse -> ToString -> Output : "\\\\/Solidus"
Input : "\\\\ReverseSolidus" -> Parse -> ToString -> Output : "\\\\\\\\ReverseSolidus"

Fix adds support for escaping \ and / and parse them correctly.
## Type of change


Microsoft.Azure.Cosmos/src/Query/Core/Parser/CstToAstVisitor.cs
```diff
@@ -950,8 +951,46 @@ public UnknownSqlObjectException(SqlObject sqlObject, Exception innerException =
...
-            return textWithoutQuotes;
+
+            // 1. Remove leading and trailing (single or double) quotes.
+            // 2. Unescape following characters:
+            //    \" => "
+            //    \\ => \
+            //    \/ => /
+            // Based on the documentation, we should also escape single quote \'.
+            // https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query/constants#remarks
+            // However that's failing in the parser (before reaching this point) and will be fixed separately (after checking server's behavior).
```
leminh98, line 962: I believe this is because the grammar I force on the string. Currently it says: 
String Literal = " string_without_double_quote "   ;
                       = ' string_without_single_quote '

Microsoft.Azure.Cosmos/src/Query/Core/Parser/CstToAstVisitor.cs
```diff
@@ -950,8 +951,46 @@ public UnknownSqlObjectException(SqlObject sqlObject, Exception innerException =
...
+            StringBuilder stringBuilder = new StringBuilder(text.Length);
+            for (int i = 1; i < text.Length - 1; i++)
```
leminh98, line 964: >i [](http://example.com/codeflow?start=21&length=1)

nit: index?

Microsoft.Azure.Cosmos/src/Query/Core/Parser/CstToAstVisitor.cs
```diff
@@ -950,8 +951,46 @@ public UnknownSqlObjectException(SqlObject sqlObject, Exception innerException =
...
+            {
+                switch (text[i])
+                {
+                    case '\\':
+                        if ((i + 1) < (text.Length - 1))
```
leminh98, line 969: >(i + 1) < (text.Length - 1) [](http://example.com/codeflow?start=28&length=27)

less than or equal

otherwise you are not getting the very final character in the string

Microsoft.Azure.Cosmos/src/Query/Core/Parser/CstToAstVisitor.cs
```diff
@@ -950,8 +951,46 @@ public UnknownSqlObjectException(SqlObject sqlObject, Exception innerException =
...
+                        break;
+
+                    default:
+                        stringBuilder.Append(text[i]);
+                        break;
+                }
+            }
+
+            string unescapedString = stringBuilder.ToString();
+            return unescapedString;
```
leminh98, line 993: nit: we can combine these lines.

Perhaps we should rename the function GetUnescapedStringValueFromNode

leminh98:
>         public void ObjectCreate()

please update the baseline for these

---
Refers to: Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Parser/ScalarExpressionSqlParserBaselineTests.cs:400 in e02ace7. [](commit_id = e02ace747788ec20742cfcee492636b9c8e9e400, deletion_comment = False)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Parser/ScalarExpressionSqlParserBaselineTests.cs
```diff
@@ -520,6 +520,24 @@ public void OrderOfOperation()
             this.ExecuteTestSuite(inputs);
         }
 
+        [TestMethod]
+        public void StringLiteral()
+        {
+            List<SqlParserBaselineTestInput> inputs = new List<SqlParserBaselineTestInput>()
+            {
+                // Single quoted string literals do not allow ' even when it's escaped.
+                // Parser currently fails with Antlr4.Runtime.NoViableAltException
```
leminh98, line 529: Ah yes. definitely because the grammar dictates so

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Parser/ScalarExpressionSqlParserBaselineTests.cs
```diff
@@ -520,6 +520,24 @@ public void OrderOfOperation()
...
+                CreateInput(
+                    description: @"Single quoted string literals with escape seqence",
+                    scalarExpression: @"['\""DoubleQuote', '\\ReverseSolidus', '\/solidus', '\bBackspace', '\fSeparatorFeed', '\nLineFeed', '\rCarriageReturn', '\tTab', '\u1234']"),
+                CreateInput(
+                    description: @"Double quoted string literals with escape seqence",
+                    scalarExpression: @"[""'SingleQuote"", ""\""DoubleQuote"", ""\\ReverseSolidus"", ""\/solidus"", ""\bBackspace"", ""\fSeparatorFeed"", ""\nLineFeed"", ""\rCarriageReturn"", ""\tTab"", ""\u1234""]"),
+            };
```
leminh98, line 536: Can you also add test without the verbatim identifier

Microsoft.Azure.Cosmos/src/Query/Core/Parser/CstToAstVisitor.cs
adityasa: We ignore first and last characters since they must be quotes for sting token.
I am using index+1 in next expression, so it's probably more readable this way to indicate that we are going to compare next character. Happy to change it if you feel strongly.

Microsoft.Azure.Cosmos/src/Query/Core/Parser/CstToAstVisitor.cs
adityasa: Combined. As for the rename, I think the name does capture the intent in abstract manner, since more than just unescaping is happening - we are stripping off the quotes as well.

Microsoft.Azure.Cosmos/src/Query/Core/Parser/CstToAstVisitor.cs
leminh98: Yeah I don't have a strong opinion on this so we can keep it as is

leminh98 (approved):

sboshra (approved):
:shipit:

PR #4056, opened 2023-08-28
Client Telemetry: Adds new public APIs

### LINKED ISSUE #3983: [Internal] Client Telemetry Public API Proposal
### Background
As per latest discussion, to get align with Java API and avoid confusion with feature name. We decided to use "Client Telemetry" as a parent feature under which there will be 2 kind of telemetry covered as of now:
**1. Send Client Diagnostic Metrics To Service :** Sending telemetry metrics to Microsoft. 
> Read more about it here: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/docs/observability.md#send-telemetry-from-sdk-to-service-private-preview

**2. Distributed Tracing:** Sending Activities with operation level or network level information to customer's APM tool like AppInsights (May be in future metrics will also be covered). 
> Read more about it here: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/docs/observability.md#distributed-tracing-preview

We are fine if Java API and .Net API doesn't look exactly similar. Here idea is to make sure, it should not confuse customer.

### Java Public API looks like this:

**CosmosClientBuilder** 
image

**CosmosClientTelemetryConfig** 
image

**CosmosDiagnosticsThresholds**
image

### Proposed Public API in .Net SDK to control above features:

By default, both of the above features will be enabled in SDK. 
#### Why?
**1. Send Client Diagnostic Metrics To Service:** It is controlled by portal hence enabling it in SDK, by default, will give seamless experience to the customer where they can just enable it from portal and we can see data flowing to us without any code change. There won't be any perf impact/overhead on sdk, if it is disabled on portal and enabled on SDK as Portal setting has higher priority.
**2. Distributed Tracing**: It works on subscription model, until unless there is no subscriber, it won't generate any activity. Hence no overhead. This feature flag is introduced to support appinsight SDK as customer cannot de-subscribe any activity source there. 

> Want to see distributed tracing in action?
>**OpenTelemetry Way:** 
>https://github.com/Azure/azure-cosmos-dotnet-v3/tree/master/Microsoft.Azure.Cosmos.Samples/Usage/OpenTelemetry
>**AppInsight SDK Way:** 
>https://github.com/Azure/azure-cosmos-dotnet-v3/tree/master/Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights
>

In both the cases, there is a scenario where customer would want to disable individual telemetry from SDK.

#### Public APIs

**CosmosClientBuilder** 
| Class Name | Method Name | Return Type | Comment|
|--------|----------|----------|--------|
| CosmosClientBuilder | WithClientTelemetryOptions(CosmosClientTelemetryOptions options) | Returns the client telemetry config instance for this builder| Single function to control/configure any kind of telemetry |

**CosmosClientOptions** 
| Class Name | Method Name | Return Type |  Default value | Comment|
|--------|----------|----------|--------|--------|
| CosmosClientOptions| CosmosClientTelemetryOptions | Set ClientTelemetry Options | new CosmosClientTelemetryOptions() _(Loaded with default values)_ | Option to control/configure any kind of telemetry |

**CosmosClientTelemetryOptions**
| Class Name | Method Name | Return Type | Default value | Comment|
|--------|----------|----------|--------| --------| 
| CosmosClientTelemetryOptions  | DisableSendingMetricsToService() | `void` | false | Disable sending telemetry data to the service i.e. Microsoft |
| CosmosClientTelemetryOptions | DisableDistributedTracing() `PREVIEW` | `void` |   `PREVIEW` ==> false |Disable Distributed Tracing feature, it means it will stop generating activities even if there are subscribers |
| CosmosClientTelemetryOptions | CosmosThresholdOptions(CosmosThresholdOptions options)  `PREVIEW` | `void` |  new CosmosThresholdOptions() | Options to configure threshold for distributed tracing (Later we can use this same config for opentelemetry metrics, similar to Java) |

**CosmosThresholdOptions**
| Class Name | Method Name | Return Type | Default value | Comment |
|--------|----------|----------|--------|--------|
| CosmosThresholdOptions| NonPointOperationLatencyThreshold(TimeSpan span)  `PREVIEW` | `void` | 3s | Set when `LatencyOverThrehold` event with diagnostic string should be created for non-point operations |
| CosmosThresholdOptions| PointOperationLatencyThreshold(TimeSpan span)  `PREVIEW` | `void` | 1s | Set when `LatencyOverThrehold` event with diagnostic string should be created for point operations |

sourabh1007:
## Description

As part of this PR, exposing different telemetry API in preview package of SDK. All the discussed contracts are mentions here:

ref. https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3983

### Breaking Changes
1. Preview package: `IsDistributedTracingEnabled` is replaced with `CosmosClientTelemetryOptions.DisableDistributedTracing`. 
Impact: Customers needs to change this call if they are disabling explicitly. Impact should be very low.

2. GA package: Not related to this PR but the way client telemetry works is completely changed now. After new release, client telemetry can be enabled only from portal.
 
## Type of change

## Closing issues
To automatically close an issue: closes #3983 and #4094


Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
```diff
@@ -103,8 +103,8 @@ public class BenchmarkConfig
         [Option(Required = false, HelpText = "Disable core SDK logging")]
         public bool DisableCoreSdkLogging { get; set; }
 
-        [Option(Required = false, HelpText = "Enable Distributed Tracing")]
-        public bool EnableDistributedTracing { get; set; }
+        [Option(Required = false, HelpText = "Disable Distributed Tracing feature from source")]
+        public bool DisableDistributedTracing { get; set; } = false;
```
kirankumarkolli, line 107: This will change defaults right? 

=> Impacts existing runner we have for default use-cases. 

One alternative is to add new runs with above combinations and measure them with and with-out. Thoughts?

Microsoft.Azure.Cosmos.Samples/Tools/CTL/CTLConfig.cs
```diff
@@ -103,8 +103,8 @@ public string DiagnosticsThresholdDuration
         [Option("ctl_reservoir_sample_size", Required = false, HelpText = "The reservoir sample size.")]
         public int ReservoirSampleSize { get; set; } = 1028;
 
-        [Option("ctl_enable_client_telemetry", Required = false, HelpText = "Enable Client Telemetry Feature in SDK. Make sure you enable it from the portal also.")]
-        public bool EnableClientTelemetry { get; set; } = true;
+        [Option("ctl_disable_client_telemetry", Required = false, HelpText = "Disable Client Telemetry Feature in SDK. Make sure you enable it from the portal also.")]
```
kirankumarkolli, line 106: For CTL where do telemetry logs go?

Microsoft.Azure.Cosmos.Samples/Tools/CTL/Program.cs
```diff
@@ -52,6 +51,10 @@ public static async Task Main(string[] args)
 
                     logger.LogInformation("Initialization completed.");
 
+                    if(client.ClientOptions.CosmosClientTelemetryOptions.DisableSendingMetricsToService) {
+                        logger.LogInformation("Telemetry Feature flag is disabled for CTL.");
```
kirankumarkolli, line 55: Suggestion: How about tracing all client options as JSON?

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
```diff
@@ -489,6 +480,20 @@ public Func<HttpClient> HttpClientFactory
             set;
         }
 
+        /// <summary>
+        /// Gets or sets Client Telemetry Options like feature flags and corresponding options
+        /// </summary>
+#if PREVIEW
```
kirankumarkolli, line 486: ConnectionPolicy is intenal type right, are preconditions necessary?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -730,9 +730,14 @@ internal Protocol ConnectionProtocol
...
-        /// Flag to enable telemetry
+        /// Gets or sets Client Telemetry Options like feature flags and corresponding options
         /// </summary>
-        internal bool? EnableClientTelemetry { get; set; }
+#if PREVIEW
+        public 
+#else
+        internal
+#endif 
+            CosmosClientTelemetryOptions CosmosClientTelemetryOptions { get; set; }
```
kirankumarkolli, line 740: CosmosPrefix on PropertyName is unnecessary right?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -1013,29 +1019,5 @@ public override bool CanConvert(Type objectType)
...
-        internal DistributedTracingOptions DistributedTracingOptions { get; set; }
-
-        /// <summary>
-        /// Gets or sets the flag to generate operation level <see cref="System.Diagnostics.Activity"/> for methods calls using the Source Name "Azure.Cosmos.Operation".
-        /// </summary>
-        /// <value>
-        /// The default value is true (for preview package).
-        /// </value>
-        /// <remarks>This flag is there to disable it from source. Please Refer https://opentelemetry.io/docs/instrumentation/net/exporters/ to know more about open telemetry exporters</remarks>
-#if PREVIEW
```
kirankumarkolli, line 1030: This is a breaking change, right? 

Also related public documentation etc... needs to be updated.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+#endif 
+        class CosmosThresholdOptions
+    {
+        /// <summary>
+        /// Latency Threshold for non point operations
+        /// </summary>
+        public TimeSpan NonPointOperationLatencyThreshold { get; set; } = TimeSpan.FromMilliseconds(500);
+
+        /// <summary>
+        /// Latency Threshold for point operations
```
kirankumarkolli, line 25: Please expand text with what point operations cover.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+#if PREVIEW
+    public 
+#else
+    internal
+#endif 
+        class CosmosThresholdOptions
+    {
+        /// <summary>
+        /// Latency Threshold for non point operations
```
kirankumarkolli, line 20: Please expand text with what non-point operations cover.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Threshold values for Distributed Tracing
+    /// </summary>
+#if PREVIEW
+    public 
+#else
+    internal
+#endif 
+        class CosmosThresholdOptions
```
kirankumarkolli, line 17: Is this type expected to have any non-latency thresholds? 
If not name reflecting it is better. 

How did JAVA do?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -435,36 +435,6 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
 
         }
 
-        /// <summary>
-        /// Sets whether Distributed Tracing for "Azure.Cosmos.Operation" source is enabled.
-        /// </summary>
-        /// <param name="isEnabled">Whether <see cref="CosmosClientOptions.IsDistributedTracingEnabled"/> is enabled.</param>
-        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
-#if PREVIEW
```
kirankumarkolli, line 443: Breaking change for preview right?  Am unsure on impact. 

One options is to make it non-breaking by internally setting the ClientTelemetryOptions as appropriate, thoughts?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/Filters/DiagnosticsFilterHelper.cs
```diff
@@ -14,21 +14,13 @@ internal static class DiagnosticsFilterHelper
...
-            {
-                latencyThreshold = config.LatencyThresholdForDiagnosticEvent.Value;
-            }
-            else
-            {
-                latencyThreshold = operationType == OperationType.Query ? DistributedTracingOptions.DefaultQueryTimeoutThreshold : DistributedTracingOptions.DefaultCrudLatencyThreshold;
-            }
+            config ??= new CosmosThresholdOptions();
 
+            TimeSpan latencyThreshold = operationType == OperationType.Query ? config.NonPointOperationLatencyThreshold : config.PointOperationLatencyThreshold;
```
kirankumarkolli, line 23: [Un-related to this PR]
There might be other non-point operations than query, ex: Feed.

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
sourabh1007: I will take it as a separate PR where focus will be on making sure benchmarks are running with and without feature. Anyway I checked sometime back, benchmarks are already broken due to some unrelated changes. So, anyway it needs to be revisited.

Microsoft.Azure.Cosmos.Samples/Tools/CTL/CTLConfig.cs
sourabh1007: It will be driven by account and what client config API returns. It is no more client's concern but finally it will be our kusto table.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: it is similar to `CosmosClientOptions`. Anyway, I think we have already closed on the naming conventions.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: We can't really get the exact number of customers who will be impacted but it will impact only those customers who are disabling `distributedTracing ` feature explicitly at source. We need to mention it in the changelog while releasing.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
sourabh1007: I remember we already discussed it while talking about the public API proposal and we decided to go with this to make it consistent with java as Java already has many non latency threshold APIs and for .Net also it is a possibility.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
sourabh1007: IMO, breaking changes in preview pkg is acceptable and that too some API signature changes. We just need to make sure that we mention it in our change log.
That is the reason we release APIs under preview feature, so that we can change it later if required.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/Filters/DiagnosticsFilterHelper.cs
sourabh1007: Initially, we decided to treat Query as non point operation and other operation as point operation. But thanks to point it out. I will take care of it in another PR and will make it similar to java https://github.com/Azure/azure-sdk-for-java/blob/33db08e556262465e56a5ad6b76690147fe55219/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/OperationType.java#L57

Microsoft.Azure.Cosmos.Samples/Tools/CTL/Program.cs
sourabh1007: Client Options is not serializable hence I would not prefer to print all the settings instead printing only those which are overridden.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
sourabh1007: As mentioned above, we already discussed this naming convention in team and with java team also. Anyway it follows, what java is doing.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: Mentioned in PR description also.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
sourabh1007: I have put it as per current logic. It will change it with the logic in separate PR.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/Filters/DiagnosticsFilterHelper.cs
```diff
@@ -20,8 +20,8 @@ internal static class DiagnosticsFilterHelper
             OpenTelemetryAttributes response)
         {
             config ??= new CosmosThresholdOptions();
-
-            TimeSpan latencyThreshold = operationType == OperationType.Query ? config.NonPointOperationLatencyThreshold : config.PointOperationLatencyThreshold;
+          
+            TimeSpan latencyThreshold = DiagnosticsFilterHelper.IsPointOperation(operationType) ? config.NonPointOperationLatencyThreshold : config.PointOperationLatencyThreshold;
```
FabianMeiswinkel, line 24: negating ! missing in condition or if/else need to be switched? should also have been seen in tests - probably missing test coverage?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -771,12 +776,13 @@ internal virtual ConnectionPolicy GetConnectionPolicy(int clientId)
                 EnableTcpConnectionEndpointRediscovery = this.EnableTcpConnectionEndpointRediscovery,
                 EnableAdvancedReplicaSelectionForTcp = this.EnableAdvancedReplicaSelectionForTcp,
                 HttpClientFactory = this.httpClientFactory,
-                ServerCertificateCustomValidationCallback = this.ServerCertificateCustomValidationCallback
+                ServerCertificateCustomValidationCallback = this.ServerCertificateCustomValidationCallback,
+                CosmosClientTelemetryOptions = new CosmosClientTelemetryOptions()
             };
 
-            if (this.EnableClientTelemetry.HasValue)
+            if (this.CosmosClientTelemetryOptions != null)
```
ealsur, line 783: nit: Not really needed, because ConnectionPolicy needs to do a null check anyway to consume it.

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+#if PREVIEW
+        public 
+#else
+        internal
+#endif 
+        class CosmosClientTelemetryOptions
+    {
+        /// <summary>
+        /// Disable sending telemetry to service, <see cref="Microsoft.Azure.Cosmos.CosmosThresholdOptions"/> is not applicable to this as of now. 
+        /// This option will disable sending telemetry to service.even it is opt-in from portal.
```
ealsur, line 19: This sounds like a remark

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+#if PREVIEW
+        public 
+#else
+        internal
+#endif 
+        class CosmosClientTelemetryOptions
+    {
+        /// <summary>
+        /// Disable sending telemetry to service, <see cref="Microsoft.Azure.Cosmos.CosmosThresholdOptions"/> is not applicable to this as of now.
```
ealsur, line 18: nit: What does `is not applicable to this as of now` mean?

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+        false;
+#else
+        true;
+#endif
+
+        /// <summary>
+        /// Threshold values for Distributed Tracing. 
+        /// These values decides whether to generate operation level <see cref="System.Diagnostics.Tracing.EventSource"/> with request diagnostics or not.
+        /// </summary>
+#if PREVIEW
```
ealsur, line 52: All these Ifs are not needed, if you make the class internal, the properties can still be public

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// Latency Threshold for non point operations i.e. Query
```
ealsur, line 20: nit: Please call out the default through `<value>`

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/Filters/DiagnosticsFilterHelper.cs
```diff
@@ -48,11 +35,24 @@ public static bool IsSuccessfulResponse(OpenTelemetryAttributes response)
...
         }
 
+        /// <summary>
+        /// Get default threshold value based on operation type
+        /// </summary>
+        /// <param name="operationType"></param>
+        /// <param name="config"></param>
+        internal static TimeSpan DefaultThreshold(OperationType operationType, CosmosThresholdOptions config)
+        {
+            config ??= new CosmosThresholdOptions();
```
ealsur, line 45: Is this object creation needed for every request?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryRecorderFactory.cs
```diff
@@ -60,7 +60,7 @@ internal static class OpenTelemetryRecorderFactory
                         databaseName: databaseName,
                         operationType: operationType,
                         clientContext: clientContext,
-                        config: requestOptions?.DistributedTracingOptions ?? clientContext.ClientOptions?.DistributedTracingOptions);
+                        config: requestOptions?.CosmosThresholdOptions ?? clientContext.ClientOptions?.CosmosClientTelemetryOptions?.CosmosThresholdOptions);
```
ealsur, line 63: CosmosClientTelemetryOptions is never null, you are initializing it by default?

```
this.CosmosClientTelemetryOptions = new CosmosClientTelemetryOptions();
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTestsBase.cs
```diff
@@ -140,7 +140,10 @@ public virtual void TestInitialize()
             };
 
             this.cosmosClientBuilder = this.GetBuilder()
-                                                .WithTelemetryEnabled()
+                                                 .WithClientTelemetryOptions(new CosmosClientTelemetryOptions()
```
ealsur, line 143: Is this needed? WOuldn't it be enabled by default?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
```diff
@@ -46,7 +45,10 @@ public async Task SourceEnabled_FlagOn_DirectMode_RecordsActivity_AssertLogTrace
 
             await base.TestInit(validateSinglePartitionKeyRangeCacheCall: false, 
                                 customizeClientBuilder: (builder) => builder
-                                                                        .WithDistributedTracing(true)
+                                                                        .WithClientTelemetryOptions(new CosmosClientTelemetryOptions()
```
ealsur, line 48: Same comment as the other file, shouldn't it be enabled by default? Why set the Disabled to false?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/TransportClientHelper.cs
```diff
@@ -83,9 +83,14 @@ internal static class TransportClientHelper
...
                    if (enableDistributingTracing)
                    {
-                       builder.WithDistributedTracingOptions(new DistributedTracingOptions()
+                       builder.WithClientTelemetryOptions(new CosmosClientTelemetryOptions()
                        {
-                           LatencyThresholdForDiagnosticEvent = TimeSpan.FromMilliseconds(.0001)
+                           DisableDistributedTracing = false,
+                           CosmosThresholdOptions = new CosmosThresholdOptions()
+                           {
+                               PointOperationLatencyThreshold = TimeSpan.FromMilliseconds(.0001),
```
ealsur, line 91: nit: Why not use 0? CPUs cannot even measure this time

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: no..in connection policy is never null it is always initialized there. This check will avoid setting it as `null`

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
sourabh1007: Mentioning, _`CosmosThresholdOptions` are not applicable for this feature_. it is present, parallel to this property, hence to avoid confusion, mentioning it explicitly. Tomorrow we might use it for this telemetry also.

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
sourabh1007: I mentioned it like this intentionally, this file contains properties related to 2 features, client telemetry and distributed tracing. Both are in preview right now and will get GAed in future on different dates. 
During a change to make any of the feature GA, I don't have to think about the other feature, I can just focus on the feature which we are making GA.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/Filters/DiagnosticsFilterHelper.cs
sourabh1007: It is `coalesce ` statement, if config is null then it will create an object. It is more of safety check.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTelemetryTestsBase.cs
sourabh1007: no it is not enabled by default.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTracingOTelTests.cs
sourabh1007: `DisableDistributedTracing` is enabled by default only for preview package.

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
ealsur: Both features will use the same `CosmosClientTelemetryOptions` type?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/Filters/DiagnosticsFilterHelper.cs
ealsur: I understand the operator, but if the user is not setting it, the it is creating a new object on each request. Can we avoid this somehow? Maybe setting a default value which is `new()`?

ealsur: Otherwise, if the user is not setting this (which most users won't), you are creating a new object everytime just for the default values.

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Disable sending telemetry to service, <see cref="Microsoft.Azure.Cosmos.CosmosThresholdOptions"/> is not applicable to this as of now. 
+        /// </summary>
+        /// <remarks>This option will disable sending telemetry to service.even it is opt-in from portal.</remarks>
+        /// <value>true</value>
+#if PREVIEW
+        public 
+#else
+        internal
+#endif
+        bool DisableSendingMetricsToService { get; set; } = true;
```
ealsur, line 27: With this change, the feature is turned off by default. Isn't the user experience a bit weird here? To enable it a user needs to set the `Disable` flag to `false` instead of having an `Enable` flag to `true`.

Why did we make this choice? 

Normally a setting that is called "DisableXXXX" is because the feature is enabled by default, like `DisableDistributedTracing`.

sourabh1007: Yes, I agree with you. Here plan was always keeping this feature enable in PREVIEW and even in GA (whenever it will happen). 
But recently compute team suggested to keep `MetricsToService` feature disable by default, before we are done with some perf validations at compute side. Once we are done with this validation. I will enable this feature by default in preview package.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
philipthomas-MSFT, line 783: If `this.CosmosClientTelemetryOptions` is not null, then we set the `connectionPolicy.CosmosClientTelemetryOptions` to `this.CosmosClientTelemetryOptions`. But if `this.CosmosClientTelemetryOptions` is null, we don't set `connectionPolicy.CosmosClientTelemetryOptions` which would still end up being null anyway? So why not skip the null check on `this.CosmosClientTelemetryOptions` and just set `connectionPolicy.CosmosClientTelemetryOptions` to `this.CosmosClientTelemetryOptions`. Are we setting `connectionPolicy.CosmosClientTelemetryOptions` somewhere else and what to hold that value unless `this.CosmosClientTelemetryOptions` is not null. Maybe I am overthinking this.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -687,7 +687,7 @@ private async Task OpenPrivateAsync(CancellationToken cancellationToken)
 
             this.clientId = cosmosClientId;
             this.remoteCertificateValidationCallback = remoteCertificateValidationCallback;
-            this.isDistributedTracingEnabled = isDistributedTracingEnabled;
+            this.cosmosClientTelemetryOptions = cosmosClientTelemetryOptions ?? new CosmosClientTelemetryOptions();
```
philipthomas-MSFT, line 690: In some other code in the PR, we are doing null checks on `this.cosmosClientTelemetryOptions`, but this code is making sure it is initialized. So, are the null checks relevant?

Microsoft.Azure.Cosmos.Samples/Tools/CTL/CTLConfig.cs
```diff
@@ -130,19 +130,27 @@ internal CosmosClient CreateCosmosClient()
...
+                }
             };
 
+            Console.WriteLine("ApplicationName = " + CTLConfig.UserAgentSuffix);
+            Console.WriteLine("DisableSendingMetricsToService = " + this.DisableClientTelemetry);
+
             if (this.UseGatewayMode)
             {
                 clientOptions.ConnectionMode = ConnectionMode.Gateway;
+                Console.WriteLine("ConnectionMode = " + ConnectionMode.Gateway);
```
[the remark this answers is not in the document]
sourabh1007: yes...this is separate where we already have console statements.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: In connection policy, it is never null it is always initialized in the connection policy constructor. This check will avoid setting it as null.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: It is just safety check.

philipthomas-MSFT (approved):

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
FabianMeiswinkel: So, eventually it will have default value of false - in that case the PAI name makes sense to me.

FabianMeiswinkel (dismissed):
One NIT and a comment regarding the default values for the diagnostics thresholds. Otherwise LGTM - thanks!

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// </summary>
+        /// <value>500 ms</value>
+        public TimeSpan NonPointOperationLatencyThreshold { get; set; } = TimeSpan.FromMilliseconds(500);
```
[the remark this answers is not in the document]
sourabh1007: I kept this value unchanged but yes you are right, eventually we want to keep it same for Java and .Net SDK. I am making this change.

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
ealsur: Got it, in that case then makes sense.

FabianMeiswinkel (approved):
LGTM now

ealsur (approved):
Please wait for @kirankumarkolli for public API approval

philipthomas-MSFT (approved):

kirankumarkolli (approved):

Microsoft.Azure.Cosmos.Samples/Tools/CTL/Program.cs
```diff
@@ -14,7 +14,7 @@ namespace CosmosCTL
     using App.Metrics.Gauge;
     using Microsoft.Azure.Cosmos;
     using Microsoft.Extensions.Logging;
-    using Microsoft.Azure.Cosmos.Telemetry;
```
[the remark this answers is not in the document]
sourabh1007: it is outdated change, already reverted.

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
```diff
@@ -0,0 +1,58 @@
...
+        internal
+#endif 
+        class CosmosClientTelemetryOptions
+    {
+        /// <summary>
+        /// Disable sending telemetry to service, <see cref="Microsoft.Azure.Cosmos.CosmosThresholdOptions"/> is not applicable to this as of now. 
+        /// </summary>
+        /// <remarks>This option will disable sending telemetry to service.even it is opt-in from portal.</remarks>
+        /// <value>true</value>
+#if PREVIEW
```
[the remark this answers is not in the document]
sourabh1007: ref https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4056/files/6f8d87929627645240ed0c8b54cca879686b9c16#r1336994592

PR #4073, opened 2023-09-07
Per Partition Automatic Failover: Fixes Gateway 503 Cold Start Issue

### LINKED ISSUE #3981: [Per Partition Automatic Failover] Extend Automatic Failover For Single Master Write Requests
Acceptance Criteria:

- Today, the per partition automatic failover applies to the multi master accounts. The purpose of this task is to extend the automatic failover for the single master accounts. 
- For a single master account, today there is only one write region. Therefore, for failover for the write requests, we will need to mark one of the read regions as a write region and start failing over to that particular region.

kundadebdatta:
# Pull Request Template

## Description

Today, the partition level failovers are applicable for multi master write accounts, for a simple reason. If one of the write region fails with a write forbidden 403 exception, then the SDK has the knowledge (by looking up the `ApplicationPreferredRegions`) of the other regions to failover. With the per partition automatic failover, if a partition is in quorum loss, then the backend automatically marks another region as the write region, based on the account configuration. Therefore, any retry for the write requests, to the next region (configured in the account level) should be successful (eventually).

The "eventual" statement has been used to signify that there may be a possibility that the next region selected by the SDK may not be a write region. Basically, as of today, there are no good ways for the SDK to know what is the next write region. The only way for the SDK to know this is to keep retrying based on the region list. 

This idea extends the SDK's retry logic to retry the **write requests for single master write accounts,** for any service unavailable (status codes 503) errors.

## Current Issues

- [SDK not retrying with next region in case address resolution call to Gateway call fails with 503.](https://msdata.visualstudio.com/CosmosDB/_workitems/edit/2475521/)
- [First client write request after failover is failing with 503(21005)](https://msdata.visualstudio.com/CosmosDB/_workitems/edit/2492475/)
- [SDK is Not Retrying With Account Configured Regions If `ApplicationPreferredRegion` is Set](https://msdata.visualstudio.com/CosmosDB/_workitems/edit/2736120) 

## Scope

This PR fixes all the issues listed above. The scope of the changes applicable only for a single master write accounts.

## Type of change

## Closing issues

To automatically close an issue: closes #3981, #4153


Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -247,8 +247,7 @@ public void OnBeforeSendRequest(DocumentServiceRequest request)
             }
 
             // Received 503 due to client connect timeout or Gateway
-            if (statusCode == HttpStatusCode.ServiceUnavailable
```
philipthomas-MSFT, line 250: How we exhausted any scenarios where this change may affect?

philipthomas-MSFT (approved):

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -884,6 +885,15 @@ private void ValidateLimitToEndpointSettings()
             }
         }
 
+        private void ValidatePartitionLevelFailoverSettings()
+        {
+            if (this.EnablePartitionLevelFailover
+                && (this.ApplicationPreferredRegions == null || this.ApplicationPreferredRegions.Count == 0))
```
ealsur, line 891: Validate also ApplicationRegion

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -247,8 +247,7 @@ public void OnBeforeSendRequest(DocumentServiceRequest request)
...
-                && ClientRetryPolicy.IsRetriableServiceUnavailable(subStatusCode))
+            if (statusCode == HttpStatusCode.ServiceUnavailable)
```
ealsur, line 250: Do we have any existing tests that need updating?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -400,13 +393,6 @@ private ShouldRetryResult ShouldRetryOnServiceUnavailable()
                 return ShouldRetryResult.NoRetry();
             }
 
-            if (!this.canUseMultipleWriteLocations
```
ealsur, line 403: What if Per-Partition Failover is not enabled, can this cause an issue for existing Single Master customers on a Write 503?

docs/PerPartitionAutomaticFailoverDesign.md
```diff
@@ -0,0 +1,51 @@
+# Working Principles for Per Partition Automatic Failover
+
+## Table of Contents
+
+* [Scope.](#scope)
+* [Backgraound.](#backgraound)
```
ealsur, line 6:
```suggestion
* [Background.](#background)
```

docs/PerPartitionAutomaticFailoverDesign.md
```diff
@@ -0,0 +1,51 @@
...
+* [How Does the SDK know Which Region to Fail Over.](#how-does-the-sdk-know-which-region-to-fail-over)
+* [References.](#references)
+
+## Scope
+
+The scope of the per partition automatic failover design document is applicable for the Cosmos .NET SDK configured for both `Gateway` and `Direct` mode.
+
+## Backgraund
+
+Today, the partition level failovers are applicable for multi master write acounts, for a simple reason. If one of the write region fails with a write forbidden 403 exception, then the SDK has the knowledge (by looking up the `ApplicationPreferredRegions`) of the other regions to failover. With the per partition automatic failover, if a partition is in quorum loss, then the backend automatically marks another region as the write region, based on the account configuration. Therefore, any retry for the write requests, to the next preferred region should be successful.
```
ealsur, line 17: > the partition level failovers are applicable for multi master write acounts

How is this enforced?

ealsur, line 17: > Therefore, any retry for the write requests, to the next preferred region should be successful

Is this for certain? Remember Multi Master Writes also pass Session Token, is there any chance that the retry will face 404/ReadSessionNotAvailable?

docs/PerPartitionAutomaticFailoverDesign.md
```diff
@@ -0,0 +1,51 @@
...
+
+This idea extends the SDK's retry logic to retry the write requests for single master write accounts, for any service unavailable (status codes 503) errors.
```
ealsur, line 19: How does the retry work for cases where the 503 is not due to per-partition failover? Imagine a customer is getting Connect Timeouts due to a network problem, which generates a 503 due to connectivity. How will that 503 work? It cannot be retried in another region, right?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kundadebdatta: To answer this question, I did simulated the behavior for a single master write account. If the per partition automatic failover is not enabled, then we do retry on the same region and the write request eventually fails. 

image

IMO, this is expected and aligns with what we discussed before, that SDK should do a retry on any `503`. Also, please note that, if PPAF is not enabled, then the `DocumentClient` sets a `GlobalPartitionEndpointManagerNoOp` which doesn't adds any partition level overrides. Please take a look at [these lines of code](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/DocumentClient.cs#L928-L930) for more details.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kundadebdatta: Yes, I have updated some tests in the [GlobalPartitionEndpointManagerTests.cs](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4073/files#diff-6ba25fc67752c0351bdec5a6db0837748c92a708e7b8c6ef11d467a3ac231107). Please note, I will add more tests to this class to cover as many scenario as possible.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
ealsur: Correct but we are now changing the behavior, right? We went from no retries to 1 retry on the same region. That is what concerns me a bit. That this is not adding any cross-region retries (what the ClientRetryPolicy is supposed to own).

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -134,11 +134,19 @@ private bool CanUsePartitionLevelFailoverLocations(DocumentServiceRequest reques
...
                 partitionKeyRange,
-                (_) => new PartitionKeyRangeFailoverInfo(failedLocation));
-
+                (_) => new PartitionKeyRangeFailoverInfo(failedLocation));
+
+            // For any single master write accounts, the next locations to fail over will be the read regions configured at the account level. Here
+            // the globalEndpointManager.WriteEndpoints are basically initialized from the Location Cache, with the read regions, configured at the
+            // account level. For multi master write accounts, since all the regions are treated as write regions, the next locations to fail over
+            // will be the preferred read regions that are configured in the application preferred regions in the CosmosClientOptions.
+            ReadOnlyCollection<Uri> nextLocations = !this.globalEndpointManager.CanUseMultipleWriteLocations(request)
```
ealsur, line 143: If this is a Read Request on Single Master, it's using the WriteEndpoints. Is that correct?

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -491,26 +494,45 @@ private bool IsEndpointUnavailable(Uri endpoint, OperationType expectedAvailable
...
-                nextLocationInfo.ReadEndpoints = this.GetPreferredAvailableEndpoints(nextLocationInfo.AvailableReadEndpointByLocation, nextLocationInfo.AvailableReadLocations, OperationType.Read, nextLocationInfo.WriteEndpoints[0]);
+                    nextLocationInfo.AvailableWriteLocations = availableWriteLocations;
+                }
+
+                // For any single master write accounts, the write endpoints would be the read regions configured at the account level.
+                // For multi master write accounts, since all the regions are treated as write regions, the write endpoints would be
+                // the preferred write regions that are configured in the application preferred regions in the CosmosClientOptions.
+                bool isSingleMasterAndPartitionLevelFailoverEnabled = !this.CanUseMultipleWriteLocations() && this.partitionLevelFailoverEnabled;
+
+                nextLocationInfo.WriteEndpoints = this.GetPreferredAvailableEndpoints(
```
ealsur, line 518: Why is this needed if you are changing this behavior on the PartitionEndpointManager?

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kundadebdatta: This is not applicable for the read requests and it is covered from line# `108` - `110`.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
kundadebdatta: Discussed offline and I have refactored the code to introduce `AccountReadEndpoints` in the `LocatinCache` for better readability.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
FabianMeiswinkel: +1 on Matias comment - yes, ClientRetryPolicy should do cross-regional retries on all 503 sub status codes. And for reads retrying on 503s might be useful as well (if the COnsistency Reader isn't doing this implicitly) - but for writes it doesn't make sense to retry 503 in same region - would hit the same replica again anyway

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -52,7 +52,7 @@ internal sealed class LocationCache
             this.lockObject = new object();
             this.locationUnavailablityInfoByEndpoint = new ConcurrentDictionary<Uri, LocationUnavailabilityInfo>();
             this.lastCacheUpdateTimestamp = DateTime.MinValue;
-            this.enableMultipleWriteLocations = false;
+            this.enableMultipleWriteLocations = false;
```
ealsur, line 55: Are these whitespace changes?

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -100,6 +100,24 @@ public ReadOnlyCollection<Uri> ReadEndpoints
...
+
+        /// <summary>
+        /// Gets list of account level read endpoints.
+        /// </summary>
+        public ReadOnlyCollection<Uri> AccountReadEndpoints
+        {
+            get
+            {
+                // Hot-path: avoid ConcurrentDictionary methods which acquire locks
+                if (DateTime.UtcNow - this.lastCacheUpdateTimestamp > this.unavailableLocationsExpirationTime
```
ealsur, line 112: Why do you need this check? AccountReadEndpoints is not affected by UnavailableLocationsExpirationTime

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Added the validation for `ApplicationRegion`

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kundadebdatta: Refactored the code to make it similar to how our SDK is behaving today, i.e. no retry for a single master write account, if PPAF is not enabled.

docs/PerPartitionAutomaticFailoverDesign.md
kundadebdatta: Today, there is a check in `ClientRetryPolicy` that covers the above statement:

```
            if (!this.canUseMultipleWriteLocations
                    && !this.isReadRequest)
            {
                // Write requests on single master cannot be retried, no other regions available
                return ShouldRetryResult.NoRetry();
            }
```

docs/PerPartitionAutomaticFailoverDesign.md
kundadebdatta: Fixed type with latest commit.

docs/PerPartitionAutomaticFailoverDesign.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Design Approach
+
+Today, the partition level failover is applicable only for the multi-master write accounts. In order to enable the partition level failover for single master write accounts, below changes are required to be made:
```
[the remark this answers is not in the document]
kundadebdatta: Removed from the doc.

docs/PerPartitionAutomaticFailoverDesign.md
kundadebdatta: I think the idea is - for any 503 errors, the SDK should retry on another region, irrespective of any sub status codes, only for a single master write accounts. Cc: @FabianMeiswinkel for any additional thoughts.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
kundadebdatta: Not sure, why these are appearing as diffs. I validated that there are no whitespace. It may be due to some other character. Will try to fix this.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
kundadebdatta: You are right. This probably doesn't require the `if` condition. I have removed it.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
kundadebdatta: Okay, these were carriage return `CR` characters. I have removed them.

ealsur (dismissed):
LGTM, please get reviews from Fabian

docs/PerPartitionAutomaticFailoverDesign.md
FabianMeiswinkel: All true - but from my perspective the attempt is to explain why a cross-regional retry makes sense for the client. So, rephrasing to "Therefore, any retry for the write requests, to the next preferred region **could** be successful." should be sufficient

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
FabianMeiswinkel: We have decided that 503 should always qualify for cross-regional retry. We made the same change in Java - and it has also been discussed with backend folks.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -888,6 +889,16 @@ private void ValidateLimitToEndpointSettings()
             }
         }
 
+        private void ValidatePartitionLevelFailoverSettings()
+        {
+            if (this.EnablePartitionLevelFailover
+                && (this.ApplicationPreferredRegions == null || this.ApplicationPreferredRegions.Count == 0)
+                && string.IsNullOrEmpty(this.ApplicationRegion))
```
FabianMeiswinkel, line 896: We should still throw even if ApplicationRegion is provided but no ApplicationPreferredRegions if partition level failover is enabled

FabianMeiswinkel (dismissed):
LGTM - but please keep tracking in the backlog a task to also add integration tests against real account when fault injection is available in the CI pipeline

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: ApplicationRegion populates PreferredLocations

ealsur: ApplicationRegion and ApplicationPreferredRegions are 2 different ways to populate Preferred Locations, you want to make sure Preferred Locations is set, so either of both options does it.

kundadebdatta: So @FabianMeiswinkel , the way it works today is based on the application region, the SDK considers the other regions that are closer based on proximity and forming a preferred region list. Therefore, as @ealsur pointed, the application preferred regions are populated in either of the cases. So, validating either should cover this.

kirankumarkolli (approved):

docs/PerPartitionAutomaticFailoverDesign.md
```diff
@@ -0,0 +1,28 @@
...
+## Table of Contents
+
+* [Scope.](#scope)
+* [Background.](#background)
+* [How Does the SDK know Which Region to Fail Over.](#how-does-the-sdk-know-which-region-to-fail-over)
+* [References.](#references)
+
+## Scope
+
+The scope of the per partition automatic failover design document is applicable for the Cosmos .NET SDK configured for both `Gateway` and `Direct` mode.
```
kirankumarkolli, line 12: In `Gateway` mode what are the role/responsibilities of SDK and Gateway?

kirankumarkolli: non-blocker

kirankumarkolli (dismissed):

docs/PerPartitionAutomaticFailoverDesign.md
kundadebdatta: As discussed over the call, I have updated the document to reflect the `Direct` mode as of now.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: As per our offline conversation, we will validate only the `ApplicationPreferredRegions` for now. We can iterate this back in the future as in when needed.

ealsur (approved):

kirankumarkolli (approved):

PR #4081, opened 2023-09-13
Remove Timer from Pool if it is already completed.

### LINKED ISSUE #4027: High memory usage
**Describe the bug**
We have a service running on two nodes. After a load test I see a high rise of memory usage. I took a memory dump of both and see most of memory is consumed by
* `Task+ContingentProperties` (about 23%)
* `Task<Object>` (about 21%)
* `PooledTimer` (about 19%)

If I look at Retained Size I see a usage of 84% on TimerPool (502 instances).

My application is reading and writing to Azure CosmosDB in a highly parallelized way (several Parallel.ForEachAsync).

**To Reproduce**
Sorry, but I can't reproduce this outside of the test environment.

**Expected behavior**
Memory consumption is lower.

**Actual behavior**
Server memory is eaten up somewhere.

**Environment summary**
SDK Version: 3.35.2 (Microsoft.Azure.Cosmos)
OS Version (e.g. Windows, Linux, MacOSX): Windows Server 2019 Datacenter

**Additional context**
I guess this is memory is consumed by the connection pool for cosmos db. I have configured it as
```csharp
MaxTcpConnectionsPerEndpoint = 1000, // Default: 65535
IdleTcpConnectionTimeout = TimeSpan.FromHours(2), // Default: unlimited
```
to limit the memory usage.

Sorry if this is the wrong place for this question but I don't know where to ask elsewhere.

MartinDemberger:
# Pull Request Template

## Description

In #4027 a high memory consumption is described. This only occures when a lot of requests are sent and the TimeOut is high.

With this change timers which are completed are treated as if there time is reached.

## Type of change

## Closing issues

closes #4027

### DIFF
--- a/Microsoft.Azure.Cosmos/src/direct/PooledTimer.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/PooledTimer.cs
@@ -70,6 +70,14 @@ namespace Microsoft.Azure.Documents
         }
 
         /// <summary>
+        /// The timed task is already completed.
+        /// </summary>
+        public bool IsCompleted
+        {
+            get { return this.tcs.Task.IsCompleted; }
+        }
+
+        /// <summary>
         /// Starts the timer for the timeout period specfied in constructor
         /// </summary>
         /// <returns>Returns the Task upon which you can await on until completion</returns>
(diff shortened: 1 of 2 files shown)


MartinDemberger:

PR #4082, opened 2023-09-14
[Internal] Parallel Hedging: Adds Docs

NaluTripician:
# Pull Request Template

## Description

## Type of change

## Closing issues


docs/Parallel Hedging Preview.md
```diff
@@ -0,0 +1,101 @@
...
+client.DisableSpeculativeProcessing(); // Disables speculative processing
+client.EnableSpeculativeProcessing(); // Enables speculative processing
+```
+
+## Dynamic Preferred Regions
+
+If you want to change the preferred regions for a `CosmosClient` you can do so by callinga new method.
+
+```csharp
+client.UpdatePreferredRegions(new List<string>() { "West US", "East US" });
```
kirankumarkolli, line 39: ObservableCollection is also one alternative. https://learn.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.observablecollection-1?view=netstandard-2.0

docs/Parallel Hedging Preview.md
```diff
@@ -0,0 +1,101 @@
...
+    .WithSpeculativeProcessing(speculativeThreshold: TimeSpan.FromMilliseconds(500))
+    .Build();
+```
+
+or
+
+```csharp
+CosmosClientOptions options = new CosmosClientOptions()
+{
+    SpeculativeProcessor = new ThresholdSpeculator(TimeSpan.FromMilliseconds(500))
```
kirankumarkolli, line 18: Is SpeculativeProcessor a public contract CX can add/extend?

docs/Parallel Hedging Preview.md
```diff
@@ -0,0 +1,101 @@
...
+CosmosClient client = new CosmosClient(
+    accountEndpoint: "account endpoint",
+    authKeyOrResourceToken: "auth key or resource token",
+    clientOptions: options);
+```
+
+The example above will create a `CosmosClient` instance with speculative processing enabled with at 500ms threhshold. This means that if a request takes longer than 500ms the SDK will send a new request to the backend in order of the Preferred Regions List. This process will repeat until a request comes back. The SDK will then return the first response that comes back from the backend. The threshold parameter is a required parameter can can be set to any value greater than 0. There are also methods that will allow a user to disable/enabled speculative processing after the client has been created. If you include speculative processing when creating the client it will be enabled by default.
+
+```csharp
+client.DisableSpeculativeProcessing(); // Disables speculative processing
```
kirankumarkolli, line 30: Request level granular is also one ASK. 
Assuming if we have that then are these APIs at Client scope still necessary?

docs/Parallel Hedging Preview.md
```diff
@@ -0,0 +1,101 @@
...
+This will update the preferred regions for the client. The SDK will then use the new preferred regions for all new requests.
+
+## Region Choice at Request Level
+
+To choose what region a request will be routed to at a per request level there is a new option on the `RequestOptions` object.
+
+```csharp
+RequestOptions requestOptions = new RequestOptions()
+{
+    LocationToRouteTo = "West US"
```
kirankumarkolli, line 51: Will it be a single Region or Equivalent to PReferredRegions?

docs/Parallel Hedging Preview.md
NaluTripician: I think it might be better to have it be a private contract here. I will make it so it's only a private contract.

docs/Parallel Hedging Preview.md
```diff
@@ -0,0 +1,101 @@
+# Paralell Hedging Preview
+
+## Paralell Hedging
+
+When Building a new `CosmosClient` there will be an option to include paralell hedging in that client.
+
+```csharp
+CosmosClient client = new CosmosClientBuilder("connection string")
+    .WithSpeculativeProcessing(type: SpeculationType.Threshold, speculativeThreshold: TimeSpan.FromMilliseconds(500))
```
FabianMeiswinkel, line 9: Remove Speculative/Speculation term everywhere

This has burned us already with Walmart - for customers "speculation" sounds too random. In Java we have used the term ThresholdBasedAvailabilityStartegy or something along those lines.

FabianMeiswinkel: Like here
WithAvailabilityStrategy(AvailabilityStartegyType.Threshold) or such

docs/Parallel Hedging Preview.md
FabianMeiswinkel: Making this mutable is risky - in Java we have instead allowed specifying excluded regions (and made that mutable) as well as configurable per request options. Achieves the same - but hasn't the same implications of making preferred regions mutable everywhere

docs/Parallel Hedging Preview.md
FabianMeiswinkel: +1 requestOptions level override definitely is needed - especially for queries (but not just)

docs/Parallel Hedging Preview.md
FabianMeiswinkel: Again - why do the opposite of Java (we used excluded regions here) - region to route to as an override owuld allow routing to a region which is not even in preferred regions?

I think it just opens a can of worms

docs/Parallel Hedging Preview.md
```diff
@@ -0,0 +1,101 @@
...
+        })
+    .Build();
+```
+
+or
+
+```csharp
+CosmosClientOptions options = new CosmosClientOptions()
+{
+    RegionConsistencyLevel = new ObservableCollection<string, ConsistencyLevel>()
```
FabianMeiswinkel, line 88: Wow - why is this needed?

FabianMeiswinkel (changes requested):
As discussed - public API has too many inconsistencies with Java that have been discussed before and are pitfalls.

docs/Parallel Hedging Preview.md
```diff
@@ -0,0 +1,125 @@
...
+
+The example above will create a `CosmosClient` instance with speculative processing enabled with at 500ms threhshold. This means that if a request takes longer than 500ms the SDK will send a new request to the backend in order of the Preferred Regions List. This process will repeat until a request comes back. The SDK will then return the first response that comes back from the backend. The threshold parameter is a required parameter can can be set to any value greater than 0. There will also be options to specify a all options for the `AvailabilityStrategyOptions` object at request level and enable or disable speculative processing at request level.
+
+```csharp
+RequestOptions requestOptions = new RequestOptions()
+{
+    AvailabilityStrategyEnabed = true,
+    AvailabilityStrageyThreshold = TimeSpan.FromMilliseconds(400),
+    AvailabilityStrageyStep = TimeSpan.FromMilliseconds(50)
+};
```
jeet1995, line 42: This is more from a reusability perspective - you already have a user defined type in `AvailabilityStrategyOptions`. Can't we use this type and inject into `RequestOptions`?

docs/Parallel Hedging Preview.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+~~This will set the consistency level for the regions specified in the dictionary. If a region is not specified in the dictionary the SDK will use the default consistency level for the account.~~
+
+### Option 2
+
+```csharp
+RequestOptions requestOptions = new RequestOptions()
+{
+    AvailabilityStrategyFallbackConsistencyLevel = ConsistencyLevel.Eventual
```
jeet1995, line 121: This can be encapsulated by `AvailabilityStrategyOptions` as well right? Also, what are the consistency override semantics here - availability strategy level config overrides request level config overrides client level config?

PR #4097, opened 2023-09-21
Distributed Tracing: Fixes SDK responses compatibility with opentelemetry response

### LINKED ISSUE #4093: NotImplementedException in change feed handler
Using the 3.35.3 nuget package, I'm getting this in my logs;

```
System.NotImplementedException: The method or operation is not implemented. at 
Microsoft.Azure.Cosmos.ChangeFeed.ChangeFeedEstimatorIterator.ChangeFeedEstimatorEmptyFeedResponse.get_Headers() at 
Microsoft.Azure.Cosmos.OpenTelemetryResponse`1..ctor(FeedResponse`1 responseMessage) at 
Microsoft.Azure.Cosmos.ChangeFeed.ChangeFeedEstimatorIterator.<>c.<ReadNextAsync>b__20_1(FeedResponse`1 response) at 
Microsoft.Azure.Cosmos.ClientContextCore.RunWithDiagnosticsHelperAsync[TResult](String containerName, String databaseName, OperationType operationType, ITrace trace, Func`2 task, Func`2 openTelemetry, String operationName, RequestOptions requestOptions)
```

Thrown from here;
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/258d960ae3caa3ad989f60d5e656544c35006d0a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs#L405

sourabh1007:
## Description

There is response type available which is not compatible with _OpenTelemetryResponse_. Hence throwing error. 

As part of this PR, making sure OpenTelemetryResponse is compatible with all the available response types in SDK

## Type of change

## Closing issues

To automatically close an issue: closes #4093

### DIFF
--- a/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
+++ b/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
@@ -103,8 +103,8 @@ namespace CosmosBenchmark
         [Option(Required = false, HelpText = "Disable core SDK logging")]
         public bool DisableCoreSdkLogging { get; set; }
 
-        [Option(Required = false, HelpText = "Disable Distributed Tracing feature from source")]
-        public bool DisableDistributedTracing { get; set; } = false;
+        [Option(Required = false, HelpText = "Enable Distributed Tracing")]
+        public bool EnableDistributedTracing { get; set; } = false;
 
         [Option(Required = false, HelpText = "Client Telemetry Schedule in Seconds")]
         public int  TelemetryScheduleInSec { get; set; }
@@ -138,8 +138,8 @@ namespace CosmosBenchmark
         [Option(Required = false, HelpText = "Application Insights connection string")]
         public string AppInsightsConnectionString { get; set; }
 
-        [Option(Required = false, HelpText = "Disable Client Telemetry Feature in SDK. Make sure you enable it from the portal also.")]
-        public bool DisableClientTelemetry { get; set; } = false;
+        [Option(Required = false, HelpText = "Enable Client Telemetry Feature in SDK. Make sure you enable it from the portal also.")]
+        public bool EnableTelemetry { get; set; } = false;
 
         internal int GetTaskCount(int containerThroughput)
         {
@@ -222,8 +222,8 @@ namespace CosmosBenchmark
                 MaxTcpConnectionsPerEndpoint = this.MaxTcpConnectionsPerEndpoint,
                 CosmosClientTelemetryOptions = new Microsoft.Azure.Cosmos.CosmosClientTelemetryOptions()
                 {
-                    DisableSendingMetricsToService = this.DisableClientTelemetry,
-                    DisableDistributedTracing = this.DisableDistributedTracing
+                    DisableSendingMetricsToService = !this.EnableTelemetry,
+                    DisableDistributedTracing = !this.EnableDistributedTracing
                 }
             };
 
--- a/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Program.cs
+++ b/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Program.cs
@@ -7,10 +7,8 @@ namespace CosmosBenchmark
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
-    using System.Diagnostics.Tracing;
     using System.IO;
     using System.Linq;
-    using System.Net;
     using System.Net.Http;
     using System.Reflection;
     using System.Threading;
@@ -18,7 +16,6 @@ namespace CosmosBenchmark
     using Azure.Monitor.OpenTelemetry.Exporter;
     using CosmosBenchmark.Fx;
     using Microsoft.Azure.Cosmos;
-    using Microsoft.Extensions.Logging;
     using Newtonsoft.Json.Linq;
     using OpenTelemetry;
     using OpenTelemetry.Metrics;
@@ -323,14 +320,7 @@ namespace CosmosBenchmark
         private static async Task<ContainerResponse> CreatePartitionedContainerAsync(BenchmarkConfig options, CosmosClient cosmosClient)
         {
             Microsoft.Azure.Cosmos.Database database = await cosmosClient.CreateDatabaseIfNotExistsAsync(options.Database);
-
-            // Show user cost of running this test
-            double estimatedCostPerMonth = 0.06 * options.Throughput;
-            double estimatedCostPerHour = estimatedCostPerMonth / (24 * 30);
-            Utility.TeeTraceInformation($"The container will cost an estimated ${Math.Round(estimatedCostPerHour, 2)} per hour (${Math.Round(estimatedCostPerMonth, 2)} per month)");
-            Utility.TeeTraceInformation("Press enter to continue ...");
-            Console.ReadLine();
-
+            
             string partitionKeyPath = options.PartitionKeyPath;
             return await database.CreateContainerIfNotExistsAsync(options.Container, partitionKeyPath, options.Throughput);
         }


Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
```diff
@@ -74,5 +75,31 @@ private static string GetPayloadSize(ResponseMessage response)
             }
             return response?.Headers?.ContentLength;
         }
+
+        private static Headers GetHeader(TransactionalBatchResponse responseMessage)
```
ealsur, line 79: What's the value of this method? If the Headers are null `responseMessage?.Headers` already returns null, there is no exception?
And you are calling it like `OpenTelemetryResponse.GetHeader(responseMessage)?.SubStatusCode` so the `?` already checks if Headers is null?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse{T}.cs
```diff
@@ -57,10 +58,35 @@ internal OpenTelemetryResponse(Response<T> responseMessage)
...
+
+        private static Headers GetHeader(Response<T> responseMessage)
+        {
+            try
+            {
+                return responseMessage?.Headers;
+            }
+            catch
+            {
+                Console.WriteLine("Failed to get headers from Response");
```
ealsur, line 87: Remove all Console.WriteLine

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
sourabh1007: yes..this function will return headers and it can be null, either due to dome exception (NoImpmentationException or any other) while fetching it or it is actually null

ealsur: Can't we just fix the places where we throw an exception instead of having this wrapper? For example, in the response that the Issue is talking about, instead of throwing, have an empty Headers object?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/OpenTelemetryRecorderTests.cs
```diff
@@ -44,6 +50,195 @@ public void CheckExceptionsCompatibility()
...
+            };
+
+            IDictionary<string, object> responseInstances = new Dictionary<string, object>()
+            {
+                { "PartitionKeyRangeBatchResponse", await OpenTelemetryRecorderTests.GetPartitionKeyRangeBatchResponse() },
+                { "TransactionalBatchResponse", await OpenTelemetryRecorderTests.GetTransactionalBatchResponse() },
+                { "QueryResponse", new Mock<QueryResponse>().Object },
+                { "QueryResponse`1", OpenTelemetryRecorderTests.GetQueryResponseWithGenerics()},
+                { "ReadFeedResponse`1", ReadFeedResponse<object>.CreateResponse<object> (new ResponseMessage(HttpStatusCode.OK),MockCosmosUtil.Serializer)},
+                { "ClientEncryptionKeyResponse", new Mock<ClientEncryptionKeyResponse>().Object},
```
philipthomas-MSFT, line 71: Is this an exhaustive list of responses? What happens when we add more? Should they all behave the same?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/OpenTelemetryRecorderTests.cs
```diff
@@ -44,6 +50,195 @@ public void CheckExceptionsCompatibility()
             
         }
 
+        [TestMethod]
+        public async Task CheckResponseCompatibility()
+        {
+            IReadOnlyList<string> excludedResponses = new List<string>()
+            {
+                "MediaResponse", // Part of dead code
```
philipthomas-MSFT, line 58: Can we have more details on why these are excluded? And if we get more resdponse types in the future, what determines whether they are excluded or not? Or why they need to be handled also?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/OpenTelemetryRecorderTests.cs
sourabh1007: This test will fail and then that new response needs to be added in either of the dictionary. Depends on , if we are expecting it to be open telemetry supported or not.

sourabh1007: At least, it will be highlighted that it is something which needs attention

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
sourabh1007: I do not want to touch the code which I am not aware of. I don't know why we decided to throw exception there instead of creating it noOp.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/OpenTelemetryRecorderTests.cs
sourabh1007: Makes sense. adding more detail.

sourabh1007: added details

philipthomas-MSFT (approved):

philipthomas-MSFT (dismissed):

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
```diff
@@ -74,5 +75,31 @@ private static string GetPayloadSize(ResponseMessage response)
...
+        {
+            try
+            {
+                return responseMessage?.Headers;
+            }
+            catch
+            {
+                DefaultTrace.TraceWarning("Failed to get headers from TransactionalBatchResponse");
```
ealsur, line 87: Should this be tracing the exception?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
sourabh1007: IMO it should be separate PR to fix this bug. As part of PR, focus is on handling any exception that would have come while accessing any response, to fetch information out of it.
Intention is not hiding bugs, intention is here to make sure telemetry is working with all kind of information. If some expected information is not there, then its a bug in SDK and needs to be addressed separately,

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
sourabh1007: it is good to have tracer where ever we are catching exception.

ealsur: My point is, you are not tracing what the exception is

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
ealsur: But wasn't the point of this PR to fix the bug (NotImplementedException)? How does this pattern (try/catch everything) not hide any potential other bugs in other response types? Why not simply fix the bug (don't throw NotImplementedException) and add the tests so we check if there are any other responses that might be also throwing something, then fix those?

kirankumarkolli: Below are three possibilities right? 
- null check
- NotImplemented by design 
- Implemented but accessing failing

If the goal is to address #3 above then +1 to @ealsur comments above to limit/bound catch.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
kirankumarkolli: Its catch is bound to NotImplemented, then travel level should be lower right? (like verbose at-max info)

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
sourabh1007: @ealsur No, As part of this PR, I am considering "Not Implemented" exception is by design. (I understood that you are saying its a bug, that is separate discussion if should be considered as bug or not)
@kirankumarkolli I didn't limit catch because, I don't want telemetry collection to be failed due to such error or design choices. Telemetry data should tell us, if something is wrong in the code/operation.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
sourabh1007: @ealsur  makes sense, i will add exception Message also.
@kirankumarkolli please refer https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4097#discussion_r1347826162

sourabh1007: @ealsur pushed a code change with logging exception message.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
ealsur: @sourabh1007 If NotImplemented is by design then it cannot be TraceWarning?

```
catch (NotImplementedException ex)
{
	DefaultTrace.TraceVerbose("Failed to get headers from TransactionalBatchResponse. Exception: {0}", ex);
	return null;
}
catch (Exception ex)
{
	DefaultTrace.TraceWarning("Failed to get headers from TransactionalBatchResponse. Exception: {0}", ex);
	return null;
}
```

ealsur: I would personally avoid the `catch (Exception ex)` completely and only handle the expected cases (NotImplementedException), otherwise we are hiding other unknown cases.

Customers do normally not turn Traces on, we cannot rely on them for discovery

kirankumarkolli: > @kirankumarkolli I didn't limit catch because, I don't want telemetry collection to be failed due to such error or design choices. Telemetry data should tell us, if something is wrong in the code/operation

Let's define success? Is it just the code not failing or ding right collection. 
If its later then right data collection guarantee is also a requirement and catch-all goes against that principle right?

sourabh1007: I am agreeing on catching only `NotImplementedException` exception. If tomorrow, any new kind of exception is introduced while fetching the header, this test will fail during development only and then it can be addressed.

@kirankumarkolli here my point is Telemetry data should highlight, if some collection is not right. It should not break the existing SDK functionality. Thats what is my goal. If you notice the current issue, I am addressing as part of this PR, Customer is seeing issue during SDK operation.

ealsur (approved):

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -223,7 +223,7 @@ Documents.OperationType operationType
                     this.scope.AddAttribute(OpenTelemetryAttributeKeys.RequestContentLength, this.response.RequestContentLength);
                     this.scope.AddAttribute(OpenTelemetryAttributeKeys.ResponseContentLength, this.response.ResponseContentLength);
                     this.scope.AddAttribute(OpenTelemetryAttributeKeys.StatusCode, (int)this.response.StatusCode);
-                    this.scope.AddAttribute(OpenTelemetryAttributeKeys.SubStatusCode, (int)this.response.SubStatusCode);
```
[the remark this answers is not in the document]
sourabh1007: prev. typecasting to int was happening in the response class 
image

In case of null substatucode or header, it would fail with nullRef while typecasting.Hence put a null check here

image

and moved the typecasting here in the end of the code flow.

PR #4098, opened 2023-09-21
Distributed Tracing: Fixes dependency failure on appinsights

### LINKED ISSUE #4095: Do not treat 404 for a point read as a dependency failure
Our application does a lot of lookups by id+partitionkey to decide if a document already exists. If it does not, the 404 response is logged to application insights as a dependency failure, which we do not consider it to be, since the 404 just represents the fact that a document is not there.

Do not log 404 on point reads as a failure.

sourabh1007:
## Description

Right now, 404 and other non failure status are marked as failed dependency on Appinsights. Although it is already handled while generating request diagnostics.

As part of this PR, Putting a condition around it to make sure activities are not getting marked as failure for such requests.

Before this PR:
image

After this PR:
image

## Type of change

## Closing issues

To automatically close an issue: closes #4095

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
@@ -178,9 +178,9 @@ namespace Microsoft.Azure.Cosmos.Telemetry
                     this.scope.AddAttribute(OpenTelemetryAttributeKeys.ExceptionMessage, exception.Message);
                 }
 
-                if (exception is CosmosException cosmosException
+                if (exception is not CosmosException || (exception is CosmosException cosmosException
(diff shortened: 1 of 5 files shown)


Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -177,7 +178,12 @@ public void MarkFailed(Exception exception)
                     this.scope.AddAttribute(OpenTelemetryAttributeKeys.ExceptionMessage, exception.Message);
                 }
 
-                this.scope.Failed(exception);
+                if (exception is CosmosException cosmosException
```
ealsur, line 181: What if the exception is NOT a CosmosException? Isn't this change now dropping the log?

aavasthy (approved):

ealsur (approved):

PR #3400, opened 2022-08-22
Query: Adds new system strings in JsonBinaryEncoding, replacing 1-byte user strings

t-krestendue:
## Description

The binary encoding format for JSON documents allows users to represent strings of their choosing with just one or two bytes. However, despite the format permitting it, there are no methods to actually add these user strings. If this feature is eventually supported, the 2048 2-byte user strings we currently support will probably be more than sufficient. It makes sense to convert the 32 1-byte user strings we offer to system strings which we know occur frequently in documents. This PR accomplishes that. The change is non-breaking because currently no collections use user-strings.

## Type of change

- [] New feature (non-breaking change which adds functionality)

PR #4108, opened 2023-10-03
[Query][Internal]: Reverts adding new system strings in JsonBinaryEncoding

Maya-Painter:
## Description

Reverts #3400.  We have decided to go forward with a user-string dictionary instead, which will make use of the encoding space used for new system strings in the reverted change.

## Type of change


adityasa (approved):
:shipit:

sboshra (approved):
:shipit:

PR #4114, opened 2023-10-10
[Internal] Query : Adds test coverage for custom serializers

Maya-Painter:
## Description

Adds baseline tests that establish the current (and non-ideal in some cases) behavior of custom serializers from the `Newtonsoft.Json` and `System.Runtime.Serialization` libraries. It also tests these in combination with `CosmosLinqSerializerOptions`. 

Plus a bit of cleanup of existing test and serialization-related classes. No product changes.

## Type of change


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        [TestCleanup]
+        public async Task TestCleanup()
+        {
+            await TestContainer.DeleteContainerStreamAsync();
+        }
+
+        [TestMethod]
+        public void TestMemberInitializerDotNet()
```
Maya-Painter, line 64: I know there is a gross amount of code duplication in this file but using generics to extract methods here results in less info being captured in the baseline xml file. Up to us if we want to make this decision for the sake of prettier/more maintainable code. @adityasa

adityasa: I would error on the side of avoiding code duplication. Although I am not too sure I fully understand how generics makes baseline xml less informative. Can that be solved by writing specific code for baseline generation?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
```diff
@@ -0,0 +1,262 @@
...
+                inputs.AddRange(camelCaseSettingInputs);
+            }
+
+            this.ExecuteTestSuite(inputs);
+        }
+
+        [TestMethod]
+        public void TestMemberInitializerNewtonsoft()
+        {
+            static DataObjectNewtonsoft createDataObj(Random random)
```
adityasa, line 102: Please use a common function and avoid duplication with test methods.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
```diff
@@ -0,0 +1,262 @@
...
+            public string StringField { get; set; }
+
+            [JsonPropertyName("id")]
+            public string id { get; set; }
+
+            [JsonPropertyName("Pk")]
+            public string Pk { get; set; }
+        }
+
+        internal class DataObjectNewtonsoft : LinqTestObject
```
adityasa, line 228: >DataObjectNewtonsoft [](http://example.com/codeflow?start=23&length=20)

A couple more variations:
1. Newtonsoft - [JsonConverter. . .]
  See:
  internal class DataObject : LinqTestObject
2. Non-newtonsoft - DataMember
  See:
  public class Sport : LinqTestObject

I would also include these in the DataObjectMultiSerializer

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
adityasa: I see the problem now, since the class definitions are different.
I wonder if this can be solved by having a common base class for all that has all properties as abstract (so you can use those here) and overridden in derived classes with appropriate attributes (dotnet, newtonsoft, mixture). That would allow you to write a common method using the base object.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
```diff
@@ -0,0 +1,262 @@
...
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+    using System.Threading.Tasks;
+    using BaselineTest;
+    using global::Azure.Core.Serialization;
+    using Microsoft.Azure.Cosmos.SDK.EmulatorTests;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [SDK.EmulatorTests.TestClass]
+    public class LinqTranslationWithCustomSerializerBaseline : BaselineTests<LinqTestInput, LinqTestOutput>
```
adityasa, line 22: This is good, thanks for putting this together.
I think we should augment this test to show the original problem: the fact that object serialization when we insert/update/delete is different from LINQ query.
Is it possible to add the object payload to the baseline to show the difference?
With a proper fix, that difference should be 0 (maybe in subset of cases) and the test will serve as a spec of the behavior of v3 sdk.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
```diff
@@ -0,0 +1,262 @@
...
+    using System.Linq.Dynamic;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+    using System.Threading.Tasks;
+    using BaselineTest;
+    using global::Azure.Core.Serialization;
+    using Microsoft.Azure.Cosmos.SDK.EmulatorTests;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [SDK.EmulatorTests.TestClass]
```
adityasa, line 21: Is this the right attribute?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
Maya-Painter: 1. seems pretty well tested in https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/CosmosLinqJsonConverterTests.cs , hence why I left it out here.
2. I can add test cases for DataMember, thanks for pointing it out

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
Maya-Painter: All other LinqBaselineTests seem to have this attribute

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
Maya-Painter: image

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
Maya-Painter: Trying to use an abstract class results in unpredictable serializer behavior.  Instead of serializing just one set of properties or the decorated ones, it seems to serialize both, so we end up with 2x the expected number of properties. 

I've updated the test classes to use a templatized data generation func, but the nuances of the serializer might make it difficult for us to use generics for the queries themselves.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
Maya-Painter: Added query payload and results to the baselines.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
Maya-Painter: More discussion of this is in another comment, but I've updated part of the code to use generics and left the linq query generation as is.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
Maya-Painter: Update: DataMember test added

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializerDataMember.xml
```diff
@@ -0,0 +1,194 @@
+﻿<Results>
+  <Result>
+    <Input>
+      <Description><![CDATA[Filter w/ DataObject initializer with constant value, camelcase = True]]></Description>
+      <Expression><![CDATA[query.Where(doc => (doc == new DataObjectDataMember() {NumericField = 12, StringField = "12"}))]]></Expression>
```
adityasa, line 5: >query.Where(doc => (doc == new DataObjectDataMember() {NumericField = 12, StringField = "12"})) [](http://example.com/codeflow?start=27&length=95)

Can you please try this variation:
query.Where(doc => doc.NumericField == 1)

I expect DataMember serialization to be applied in this case to generate following query:
SELECT VALUE root
FROM root
WHERE root.NumericFieldDataMember == 1

adityasa: Here's an example:
<root>\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.EmulatorTests\BaselineTest\TestBaseline\LinqGeneralBaselineTests.ValidateLinqQueries.xml

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqTestsCommon.cs
```diff
@@ -315,6 +325,46 @@ public static DateTime RandomDateTime(Random random, DateTime midDateTime)
...
+        /// <returns>a lambda that takes a boolean which indicate where the query should run against CosmosDB or against original data, and return a query results as IQueryable. Also the serialized payload.</returns>
+        public static (Func<bool, IQueryable<T>>, List<T>) GenerateSerializationTestCosmosData<T>(Func<int, bool, T> func, int count, Container container, bool camelCaseSerialization = false)
+        {
+            List<T> data = new List<T>();
+            for (int i = 0; i < count; i++)
+            {
+                data.Add(func(i, camelCaseSerialization));
+            }
+
+            List<T> insertedData = new List<T>();
```
adityasa, line 345: >T [](http://example.com/codeflow?start=17&length=1)

You will want to keep this as string to ensure you capture the payload as it's stored in the backend (without .NET sdk's intervention/transofrmations).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqTestsCommon.cs
```diff
@@ -508,28 +556,35 @@ Cosmos.Database cosmosDatabase
...
+                // we skip unordered query because the LINQ results vs actual query results are non-deterministic
                 if (!input.skipVerification)
                 {
-                    IQueryable dataResults = compiledQuery(false);
                     LinqTestsCommon.ValidateResults(queryResults, dataResults);
                 }
 
-                return new LinqTestOutput(querySqlStr);
+                string serializedResults = includeResults ?
+                    JsonConvert.SerializeObject(queryResults.Select(item => item is LinqTestObject ? item.ToString() : item), new JsonSerializerSettings { Formatting = Newtonsoft.Json.Formatting.Indented }) :
```
adityasa, line 580: >JsonConvert.SerializeObject(queryResults.Select(item => item is LinqTestObject ? item.ToString() : item), new JsonSerializerSettings { Formatting = Newtonsoft.Json.Formatting.Indented }) : [](http://example.com/codeflow?start=20&length=188)

Please do not serialize using Newtonsoft since it transforms the objects (T) further (and honoring or not corresponding attributes) to create confusing output.

I would write a custom serialize function for the 4 classes you have (and use newtonsoft for everything else).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqTestsCommon.cs
Maya-Painter: The more I look into it the more I think it won't be possible to serialize the payload (at least exactly how the store sees it), since any interaction with the container (inserting, reading, etc.) will happen as type T, never as a string.  The request contents are also inaccessible here.  I'm tempted to just copy an example doc from the emulator directly and leave it as a comment in the test.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializerDataMember.xml
Maya-Painter: Fixed - turns out I was missing the [DataContract] tag on the object

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqTestsCommon.cs
Maya-Painter: Solved by parsing stream from GetItemQueryStreamIterator("SELECT * FROM c")

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqTestsCommon.cs
Maya-Painter: Updated to use overridden ToString()

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializerNewtonsoft.xml
```diff
@@ -0,0 +1,236 @@
...
+  "{\"NumericField\": 1, \"StringField\": \"1\", \"id\": \"1-False\", \"Pk\": \"Test\"}",
+  "{\"NumericField\": 2, \"StringField\": \"2\", \"id\": \"2-False\", \"Pk\": \"Test\"}"
+]]]></InputData>
+      <Results><![CDATA[[]]]></Results>
+    </Output>
+  </Result>
+  <Result>
+    <Input>
+      <Description><![CDATA[Filter w/ DataObject initializer with constant value, camelcase = True]]></Description>
+      <Expression><![CDATA[query.Where(doc => (doc == new DataObjectNewtonsoft() {NumericField = 1, StringField = "1"}))]]></Expression>
```
adityasa, line 26: >NumericField [](http://example.com/codeflow?start=82&length=12)

I would be tempted to document issues we find somewhere (either in code comments or in ongoing email thread). This is one of them, that object construction of T inline with query doesn't honor _any_ serializer logic - neither LINQ nor one at the CosmosClient level.

This is quite likely a class of issues rather than a single issue since essentially we are talking about supporting all surface area of C#.

adityasa: The comment is likely misplaced, since we did apply the transformation here (but not in other place).

adityasa: Well, this is half right. We only seem to apply the translator, but no camelcasing (unlike the .Select case above, where we apply both).

adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
```diff
@@ -0,0 +1,406 @@
...
+            (getQueryCamelCase, getQueryDefault) = this.InsertDataAndGetQueryables<DataObjectDotNet>();
+
+            string insertedData = this.GetInsertedData().Result;
+
+            List<LinqTestInput> inputs = new List<LinqTestInput>();
+            foreach (bool useCamelCaseSerializer in new bool[] { true, false })
+            {
+                Func<bool, IQueryable<DataObjectDotNet>> getQuery = useCamelCaseSerializer ? getQueryCamelCase : getQueryDefault;
+
+                List<LinqTestInput> camelCaseSettingInputs = new List<LinqTestInput>
```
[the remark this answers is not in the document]
Maya-Painter: added

adityasa (approved):

akotalwar (approved):
:shipit:

sboshra (approved):
:shipit:

akotalwar (commented):
/azp run

akotalwar (commented):
/azp run

Maya-Painter:

Maya-Painter:

Maya-Painter:

Maya-Painter:

ealsur:
@microsoft-github-policy-service rerun

ealsur (approved):

PR #4122, opened 2023-10-13
Query: Add Optimistic Direct Execution configuration override support on the Client

akotalwar:
# Pull Request Template

## Description
This PR accepts a new flag from the backend which tells the client whether or not to use the Ode code path. 

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -758,18 +765,30 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
             InputParameters inputParameters,
             PartitionedQueryExecutionInfo partitionedQueryExecutionInfo,
             CosmosQueryContext cosmosQueryContext,
+            CosmosClientContext clientContext,
             ContainerQueryProperties containerQueryProperties,
             ITrace trace)
         {
-            if (!inputParameters.EnableOptimisticDirectExecution)
+            AccountProperties properties = await clientContext.Client.ReadAccountAsync();
+            bool allowOdeGatewayFlag = properties.QueryEngineConfiguration.ContainsKey(AllowOptimisticDirectExecution) && Convert.ToBoolean(properties.QueryEngineConfiguration[AllowOptimisticDirectExecution]);
```
adityasa, line 773: >properties.QueryEngineConfiguration.ContainsKey(AllowOptimisticDirectExecution) && Convert.ToBoolean(properties.QueryEngineConfiguration[AllowOptimisticDirectExecution]) [](http://example.com/codeflow?start=39&length=169)

I have mixed feelings about coding it this way. Technically this is right. However, if this change gets released before the corresponding backend change, ODE gets turned off for accounts where it's ON by default.
I am tempted to leave this the way you have and just ensure that the backend change is released before this is checked in/released.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -758,18 +765,30 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
             InputParameters inputParameters,
             PartitionedQueryExecutionInfo partitionedQueryExecutionInfo,
             CosmosQueryContext cosmosQueryContext,
+            CosmosClientContext clientContext,
             ContainerQueryProperties containerQueryProperties,
             ITrace trace)
         {
-            if (!inputParameters.EnableOptimisticDirectExecution)
+            AccountProperties properties = await clientContext.Client.ReadAccountAsync();
```
adityasa, line 772: >await clientContext.Client.ReadAccountAsync(); [](http://example.com/codeflow?start=43&length=46)

Typically in a user application, clients are cached and used to run multiple (possibly) queries repeatedly.
What is the behavior of ReadAccountAsync call with respect to caching?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
adityasa, line 773: >Convert.ToBoolean [](http://example.com/codeflow?start=122&length=17)

This could throw. What is the behavior in that case?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -758,18 +765,30 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
...
+                                $"{inputParameters.InitialUserContinuationToken}");
+                    }
 
+                    return null;
+                }
+            }
+            else
+            {
                 return null;
             }
```
adityasa, line 793: If ODE is turned off by the backend, we won't provide the useful error message above (for continuing queries using ODE continuation token).

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -758,18 +765,30 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
...
             {
-                if (inputParameters.InitialUserContinuationToken != null 
-                    && OptimisticDirectExecutionContinuationToken.IsOptimisticDirectExecutionContinuationToken(inputParameters.InitialUserContinuationToken))
+                if (!inputParameters.EnableOptimisticDirectExecution)
                 {
-                    throw new MalformedContinuationTokenException($"The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. " +
-                            $"{inputParameters.InitialUserContinuationToken}");
-                }
+                    if (inputParameters.InitialUserContinuationToken != null
+                        && OptimisticDirectExecutionContinuationToken.IsOptimisticDirectExecutionContinuationToken(inputParameters.InitialUserContinuationToken))
```
adityasa, line 781: This is also applicable when ODE is turned off in the service, otherwise the user experience is confusing (Malformed continuation token for continuation of queries).

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -758,18 +765,30 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
...
+                    {
+                        throw new MalformedContinuationTokenException($"The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. " +
```
adityasa, line 783: >The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. [](http://example.com/codeflow?start=72&length=149)

Please augment this message since it's not entirely dependent on the client settings anymore.

adityasa: Also, please add coverage.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
adityasa: Also, please add coverage.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
adityasa: Also, please add coverage.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
ealsur: There is no caching here, this will generate a metadata account call every time.

Consider hooking into the places that already parse and refresh this information:

Initialization:
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/20547bce55d935518803d81f2ad5d235d543fb5d/Microsoft.Azure.Cosmos/src/DocumentClient.cs#L692

Refresh:
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/20547bce55d935518803d81f2ad5d235d543fb5d/Microsoft.Azure.Cosmos/src/DocumentClient.cs#L6453-L6462

ealsur: You can access the QueryPartitionProvider and make sure you have a getter for this new config that you can call.

Basically, do this:
```
if (properties.QueryEngineConfiguration.TryGetValue(AllowOptimisticDirectExecution, out object allowOptimisticDirectExecutionValue))
            {
                bool.TryParse(allowOptimisticDirectExecutionValue.ToString(), out allowOdeGatewayFlag);
            }
```

But within QueryPartitionProvider whenever Update or the constructor is called with the account information.

Then expose the result as a new property

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -41,6 +41,7 @@ internal static class CosmosQueryExecutionContextFactory
         public static IQueryPipelineStage Create(
             DocumentContainer documentContainer,
             CosmosQueryContext cosmosQueryContext,
+            CosmosClientContext clientContext,
```
neildsh, line 44: >CosmosClientContext clientContext, [](http://example.com/codeflow?start=12&length=34)

We should not need to take this. It is a huge interface. Let's chat offline about how this can be changed.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -44,6 +44,7 @@ internal sealed class QueryPartitionProvider : IDisposable
         private IntPtr serviceProvider;
         private bool disposed;
         private string queryengineConfiguration;
+        public IDictionary<string, object> QueryEngineConfigurationValues;
```
neildsh, line 47: >public IDictionary<string, object> QueryEngineConfigurationValues; [](http://example.com/codeflow?start=8&length=66)

Please expose a single immutable bool or enum property for the specific config that you need.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Utils/QueryPartitionProviderTestInstance.cs
```diff
@@ -9,31 +9,49 @@ namespace Microsoft.Azure.Cosmos.Tests
 
     public static class QueryPartitionProviderTestInstance
     {
-        public static readonly IReadOnlyDictionary<string, object> DefaultQueryEngineConfiguration = new Dictionary<string, object>()
+        internal static QueryPartitionProvider Create(string key = null, object value = null)
```
neildsh, line 12: >internal [](http://example.com/codeflow?start=8&length=8)

Please revert these modifications to the common test infrastructure and create a custom object in the specific test that needs it. In the vast majority of the cases, this flag will be absent, so we don't want to lose coverage for that scenario.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Utils/MockDocumentClient.cs
```diff
@@ -125,7 +125,7 @@ internal override Task<QueryPartitionProvider> QueryPartitionProvider
...
+JsonConvert.DeserializeObject<Dictionary<string, object>>("{\"maxSqlQueryInputLength\":262144,\"maxJoinsPerSqlQuery\":5,\"maxLogicalAndPerSqlQuery\":500,\"maxLogicalOrPerSqlQuery\":500,\"maxUdfRefPerSqlQuery\":10,\"maxInExpressionItemsCount\":16000,\"queryMaxInMemorySortDocumentCount\":500,\"maxQueryRequestTimeoutFraction\":0.9,\"sqlAllowNonFiniteNumbers\":false,\"sqlAllowAggregateFunctions\":true,\"sqlAllowSubQuery\":true,\"sqlAllowScalarSubQuery\":true,\"allowNewKeywords\":true,\"sqlAllowLike\":false,\"sqlAllowGroupByClause\":false,\"maxSpatialQueryCells\":12,\"spatialMaxGeometryPointCount\":256,\"sqlAllowTop\":true,\"enableSpatialIndexing\":true,\"clientDisableOptimisticDirectExecution\":false}")));
```
neildsh, line 128: >clientDisableOptimisticDirectExecution\":false [](http://example.com/codeflow?start=659&length=46)

Please removed changes to common test infrastructure code for the same reasons that I quoted in the QueryPartitionProviderTestInstance

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
akotalwar: I did this as follow because by creating a path for future configurations to be passed without the need to create their own immutable bool, we can streamline the process and reduce code in the long run.

neildsh: YAGNI: You Aren't Going to Need It
This is the first time that we needed to expose something like this in all these years. Most likely, we will not need it again soon.

If we do need it in the future, then most likely, we will have to design it properly as a separate component and not as a hack hanging off of QueryPartitionProvider.

In addition, as I noted in my initial comment, the current implementation in this change is buggy with respect to concurrent reads/updates and will throw.

adityasa: >IDictionary [](http://example.com/codeflow?start=15&length=11)

IReadOnlyDictionary

adityasa: >QueryEngineConfigurationValues [](http://example.com/codeflow?start=43&length=30)

QueryEngineConfig

adityasa: >public [](http://example.com/codeflow?start=8&length=6)

private

adityasa: I am actually tempted to
1. create a dedicated class that wraps dictionary and exposes strongly typed settings for e.g. GetValue<T>(string configName)
2. add a method to QueryPartitionProvider GetValue<T>(string configName) instead of exposing the dictionary as is.

adityasa: >object [](http://example.com/codeflow?start=35&length=6)

We want to avoid using object as a value type since it requires casting and in this case is as error prone as it gets.

akotalwar: I went ahead with Neil's comment of adding a boolean property for the QueryPartitionProvider class. Let me know if that is fine.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -45,6 +44,8 @@ internal sealed class QueryPartitionProvider : IDisposable
         private bool disposed;
         private string queryengineConfiguration;
 
+        public bool ClientDisableOptimisticDirectExecution { get; private set; }
```
adityasa, line 47: Can you make a separate follow up change to move this into a config class of its own?

akotalwar: Yes, I can do that. Will add a work item for it

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +143,22 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
...
+                throw new ArgumentNullException(nameof(queryengineConfiguration));
+            }
+
+            if (queryengineConfiguration.TryGetValue("clientDisableOptimisticDirectExecution", out object queryConfigProperty))
+            {
+                bool.TryParse(queryConfigProperty.ToString(), out bool clientDisableOptimisticDirectExecution);
+                return clientDisableOptimisticDirectExecution;
+            }
+
+            throw new KeyNotFoundException($"ClientDisableOptimisticDirectExecution flag was not found in the QueryEngineConfiguration dictionary.");
```
adityasa, line 159: This seems a bit excessive. I think we should return default false in this case.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +143,22 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
...
+        public bool GetClientDisableOptimisticDirectExecution(IDictionary<string, object> queryengineConfiguration)
+        {
+            if (queryengineConfiguration == null)
+            {
+                throw new ArgumentNullException(nameof(queryengineConfiguration));
+            }
+
+            if (queryengineConfiguration.TryGetValue("clientDisableOptimisticDirectExecution", out object queryConfigProperty))
+            {
+                bool.TryParse(queryConfigProperty.ToString(), out bool clientDisableOptimisticDirectExecution);
```
adityasa, line 155: >TryParse [](http://example.com/codeflow?start=21&length=8)

Please check the return value. I would Debug.Assert that TryParse succeeds.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +143,22 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
             return TryCatch<PartitionedQueryExecutionInfo>.FromResult(queryInfo);
         }
 
+        public bool GetClientDisableOptimisticDirectExecution(IDictionary<string, object> queryengineConfiguration)
```
adityasa, line 146: >public [](http://example.com/codeflow?start=8&length=6)

private?

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +143,22 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
...
+        {
+            if (queryengineConfiguration == null)
+            {
+                throw new ArgumentNullException(nameof(queryengineConfiguration));
+            }
```
adityasa, line 151: private methods should not throw since parameter validation should have been performed at the entry points in the class.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +143,22 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
...
+
+            if (queryengineConfiguration.TryGetValue("clientDisableOptimisticDirectExecution", out object queryConfigProperty))
```
adityasa, line 153: >queryConfigProperty [](http://example.com/codeflow?start=106&length=19)

What is the expected type here? String or bool? If it's the later, I wouldn't re-parse, but rather just cast.

akotalwar: I updatec the code with a cast as follows: (bool)queryConfigProperty.
However, this throws the following error "System.InvalidCastException: Unable to cast object of type 'System.String' to type 'System.Boolean'"

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
akotalwar: If we are unable to get the real value for clientDisableOde then we should use the default true right? 
If we can't find this config or get the value from the config, why should customers go down the ODE code path.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
adityasa: Is the type of this object a string or bool? I see in your tests you are using a bool value for mock partition provider.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +144,19 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
             return TryCatch<PartitionedQueryExecutionInfo>.FromResult(queryInfo);
         }
 
+        private bool GetClientDisableOptimisticDirectExecution(IDictionary<string, object> queryengineConfiguration)
+        {
+            if (queryengineConfiguration.TryGetValue("clientDisableOptimisticDirectExecution", out object queryConfigProperty))
+            {
+                bool success = bool.TryParse(queryConfigProperty.ToString(), out bool clientDisableOptimisticDirectExecution);
+                Debug.Assert(success);
```
adityasa, line 152: Kindly add descriptive message:
Debug.Assert(Condition, "<className>", $"Parsing must succeed. Value supplied '{queryConfigProperty}'");

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
akotalwar: Type of the object is string. 
I realize my mistake. I misread your comment. I thought I should cast if its string. 
Since its string, I think I can keep the parsing I have

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +144,19 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
...
+
+                return clientDisableOptimisticDirectExecution;
+            }
+
+            return true;
```
adityasa, line 157: >true [](http://example.com/codeflow?start=19&length=4)

Should we use false as the default?

akotalwar: If we are unable to get the real value for clientDisableOde then we should return true right?
If we can't find this config or get the value from the config, why should customers go down the ODE code path.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -761,13 +761,19 @@ private static Documents.PartitionKeyDefinition GetPartitionKeyDefinition(InputP
...
-                    && OptimisticDirectExecutionContinuationToken.IsOptimisticDirectExecutionContinuationToken(inputParameters.InitialUserContinuationToken))
+                if (inputParameters.InitialUserContinuationToken != null
+                          && OptimisticDirectExecutionContinuationToken.IsOptimisticDirectExecutionContinuationToken(inputParameters.InitialUserContinuationToken))
                 {
-                    throw new MalformedContinuationTokenException($"The continuation token supplied requires the Optimistic Direct Execution flag to be enabled in QueryRequestOptions for the query execution to resume. " +
-                            $"{inputParameters.InitialUserContinuationToken}");
+                    string errorMessage = "This query cannot be executed using the provided continuation token. " +
+                        "Please ensure that the EnableOptimisticDirectExecution flag is enabled in the QueryRequestOptions. " +
+                        "If after enabling this flag, you still see this error, contact the database administrator for assistance or retry the query without the continuation token.";
```
adityasa, line 774: Execution of the query using supplied continuation token requires EnableOptimisticDirectExecution to be set in QueryRequestOptions. If the error persists after that, contact system administrator.

Microsoft.Azure.Cosmos/src/Query/v3Query/QueryIterator.cs
```diff
@@ -146,9 +146,9 @@ internal sealed class QueryIterator : FeedIteratorInternal
                 forcePassthrough: forcePassthrough,
                 enableOptimisticDirectExecution: queryRequestOptions.EnableOptimisticDirectExecution,
                 testInjections: queryRequestOptions.TestSettings);
-
+
             return new QueryIterator(
-                cosmosQueryContext,
+                cosmosQueryContext,
```
adityasa, line 151: Nit, unintended change to this file, kindly revert.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/OptimisticDirectExecutionQueryTests.cs
```diff
@@ -538,6 +539,19 @@ public async Task TestFailingOptimisticDirectExecutionOutput()
...
+        public async Task TestClientDisableOdeFlagInCosmosClient()
+        {
+            string authKey = Utils.ConfigurationManager.AppSettings["MasterKey"];
+            string endpoint = Utils.ConfigurationManager.AppSettings["GatewayEndpoint"];
+
+            CosmosClient client = new CosmosClient($"AccountEndpoint={endpoint};AccountKey={authKey}");
+            AccountProperties properties = await client.ReadAccountAsync();
+
+            Assert.IsTrue(properties.QueryEngineConfigurationString.Contains(ClientDisableOptimisticDirectExecution));
+            Assert.IsFalse(Convert.ToBoolean(properties.QueryEngineConfiguration[ClientDisableOptimisticDirectExecution]));
```
adityasa, line 554: >Convert.ToBoolean [](http://example.com/codeflow?start=27&length=17)

Is it intentional to use Convert.ToBoolean in test as opposed to cast/TryParse in product code?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -436,6 +440,88 @@ public async Task TestPipelineForDistributedQueryAsync()
...
+            {
+                int result = await this.GetPipelineAndDrainAsync(
+                                    input,
+                                    numItems: 100,
+                                    isMultiPartition: false,
+                                    expectedContinuationTokenCount: 10,
+                                    requiresDist: false,
+                                    enableOde,
+                                    clientDisableOde);
+                Assert.Fail();
```
adityasa, line 517: A descriptive message would be nice.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -526,23 +612,28 @@ private static async Task<(MergeTestUtil, IQueryPipelineStage)> CreateFallbackPi
...
             while (await queryPipelineStage.MoveNextAsync(NoOpTrace.Singleton))
             {
-                if (!requiresDist)
+                TryCatch<QueryPage> tryGetPage = queryPipelineStage.Current;
+                tryGetPage.ThrowIfFailed();
+                
+                if (clientDisableOde || !enableOptimisticDirectExecution)
                 {
-                    Assert.AreEqual(TestInjections.PipelineType.OptimisticDirectExecution, queryRequestOptions.TestSettings.Stats.PipelineType.Value);
+                    Assert.AreNotEqual(TestInjections.PipelineType.OptimisticDirectExecution, queryRequestOptions.TestSettings.Stats.PipelineType.Value);
```
adityasa, line 630: Neat.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -667,6 +758,44 @@ private static async Task<IQueryPipelineStage> GetOdePipelineAsync(OptimisticDir
...
+                {"sqlAllowSubQuery", true},
+                {"sqlAllowScalarSubQuery", true},
+                {"allowNewKeywords", true},
+                {"sqlAllowLike", true},
+                {"sqlAllowGroupByClause", true},
+                {"maxSpatialQueryCells", 12},
+                {"spatialMaxGeometryPointCount", 256},
+                {"sqlDisableQueryILOptimization", false},
+                {"sqlDisableFilterPlanOptimization", false},
+                {"clientDisableOptimisticDirectExecution", false}
```
adityasa, line 785: I see this as a bool, but I also recall your other message talking about string. We should match the type of values in mock object with what client will actually get when it runs against production.

adityasa: Same for other config keys.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
adityasa: Please mark a TODO in code as well.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -1043,7 +1181,7 @@ public override async Task<TryCatch<PartitionedQueryExecutionInfo>> TryGetPartit
             using StreamReader streamReader = new(serializerCore.ToStreamSqlQuerySpec(sqlQuerySpec, Documents.ResourceType.Document));
             string sqlQuerySpecJsonString = streamReader.ReadToEnd();
 
-            PartitionedQueryExecutionInfo partitionedQueryExecutionInfo = OptimisticDirectExecutionQueryBaselineTests.GetPartitionedQueryExecutionInfo(sqlQuerySpecJsonString, partitionKeyDefinition);
+            (PartitionedQueryExecutionInfo partitionedQueryExecutionInfo, QueryPartitionProvider queryPartitionProvider) = OptimisticDirectExecutionQueryBaselineTests.GetPartitionedQueryExecutionInfoAndPartitionProvider(sqlQuerySpecJsonString, partitionKeyDefinition);
```
adityasa, line 1184: >OptimisticDirectExecutionQueryBaselineTests [](http://example.com/codeflow?start=123&length=43)

Nit, not required?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -667,6 +758,44 @@ private static async Task<IQueryPipelineStage> GetOdePipelineAsync(OptimisticDir
...
+                {"maxSpatialQueryCells", 12},
+                {"spatialMaxGeometryPointCount", 256},
+                {"sqlDisableQueryILOptimization", false},
+                {"sqlDisableFilterPlanOptimization", false},
+                {"clientDisableOptimisticDirectExecution", false}
+            };
+
+            if (key != null && value != null)
+            {
+                if (queryEngineConfiguration.TryGetValue(key, out _))
```
adityasa, line 790: Not required since you are using setter.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -667,6 +758,44 @@ private static async Task<IQueryPipelineStage> GetOdePipelineAsync(OptimisticDir
...
+                {"sqlAllowLike", true},
+                {"sqlAllowGroupByClause", true},
+                {"maxSpatialQueryCells", 12},
+                {"spatialMaxGeometryPointCount", 256},
+                {"sqlDisableQueryILOptimization", false},
+                {"sqlDisableFilterPlanOptimization", false},
+                {"clientDisableOptimisticDirectExecution", false}
+            };
+
+            if (key != null && value != null)
```
adityasa, line 788: I wouldn't do this check (and let the test fail if wrong values are being passed in). Alternatively, you can fail the test if any of these are false instead of silently continuing.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +145,19 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
             return TryCatch<PartitionedQueryExecutionInfo>.FromResult(queryInfo);
         }
 
+        private bool GetClientDisableOptimisticDirectExecution(IDictionary<string, object> queryengineConfiguration)
+        {
+            if (queryengineConfiguration.TryGetValue("clientDisableOptimisticDirectExecution", out object queryConfigProperty))
```
adityasa, line 150: >clientDisableOptimisticDirectExecution [](http://example.com/codeflow?start=54&length=38)

Please use the const that was declared in the other file.

adityasa: If there's an existing class where the const can live, that's better.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -34,6 +33,7 @@ internal static class CosmosQueryExecutionContextFactory
     {
         private const string InternalPartitionKeyDefinitionProperty = "x-ms-query-partitionkey-definition";
         private const string OptimisticDirectExecution = "OptimisticDirectExecution";
+        private const string ClientDisableOptimisticDirectExecution = "clientDisableOptimisticDirectExecution";
```
adityasa, line 36: >ClientDisableOptimisticDirectExecution [](http://example.com/codeflow?start=29&length=38)

Kindly double check if there's a more suitable class (specific to configs, settings) to hold this constant.

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +145,19 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
...
+            {
+                bool success = bool.TryParse(queryConfigProperty.ToString(), out bool clientDisableOptimisticDirectExecution);
```
adityasa, line 152: >ToString [](http://example.com/codeflow?start=65&length=8)

I would be tempted to cast it as string rather than ToString.
I would also add an assert in the line prior that queryConfigProperty is string

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/OptimisticDirectExecutionQueryTests.cs
```diff
@@ -538,6 +539,20 @@ public async Task TestFailingOptimisticDirectExecutionOutput()
                 documents,
                 (container, documents) => RunFailingTests(container, invalidQueries),
                 "/" + PartitionKeyField);
+        }
+
+        [TestMethod]
+        public async Task TestClientDisableOdeFlagInCosmosClient()
```
adityasa, line 545: >TestClientDisableOdeFlagInCosmosClient [](http://example.com/codeflow?start=26&length=38)

Nit, consider TestClientDisableOdeDefaultValue

adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPartitionProvider.cs
```diff
@@ -140,6 +145,19 @@ public void Update(IDictionary<string, object> queryengineConfiguration)
             return TryCatch<PartitionedQueryExecutionInfo>.FromResult(queryInfo);
         }
 
+        private bool GetClientDisableOptimisticDirectExecution(IDictionary<string, object> queryengineConfiguration)
```
neildsh, line 148: >bool [](http://example.com/codeflow?start=16&length=4)

`static` please

neildsh, line 148: >IDictionary [](http://example.com/codeflow?start=63&length=11)

Please use `IReadOnlyDictionary`

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
```diff
@@ -526,23 +611,28 @@ private static async Task<(MergeTestUtil, IQueryPipelineStage)> CreateFallbackPi
...
-                    Assert.AreEqual(TestInjections.PipelineType.OptimisticDirectExecution, queryRequestOptions.TestSettings.Stats.PipelineType.Value);
+                    Assert.AreNotEqual(TestInjections.PipelineType.OptimisticDirectExecution, queryRequestOptions.TestSettings.Stats.PipelineType.Value);
                 }
 
-                TryCatch<QueryPage> tryGetPage = queryPipelineStage.Current;
-                tryGetPage.ThrowIfFailed();
+                if (!clientDisableOde && enableOptimisticDirectExecution && !requiresDist)
+                {
+                    Assert.AreEqual(TestInjections.PipelineType.OptimisticDirectExecution, queryRequestOptions.TestSettings.Stats.PipelineType.Value);
+                }
```
neildsh, line 635: You should be able to pull these outside the `while` loop

neildsh (approved):
:shipit:

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/OptimisticDirectExecutionQueryBaselineTests.cs
akotalwar: Had already tried this but its not possible because queryRequestOptions.TestSettings.Stats is null before the while loop starts

neildsh (approved):

adityasa (dismissed):

neildsh (approved):
:shipit:

adityasa (approved):
:shipit:

PR #4123, opened 2023-10-13
[Internal] Upgrade Resiliency: Fixes Duplicate Channel and Task Creation.

### LINKED ISSUE #4033: [Cosmos Service Upgrade - .NET] Replica Validation - Fix Duplicate TCP Connections Opening Caused by Concurrent Requests
**Scope:** Applicable when the "Advanced Replica Selection" feature is enabled using the environment variable `AZURE_COSMOS_REPLICA_VALIDATION_ENABLED = True`

**Problem:** One or our recent conversations with the IC3 team has helped us unveiling a potential issue with the open connection async flow. The team migrated to the latest preview sdk version `3.35.2-preview` and experienced a sudden and consistent spike in number of tcp connections since the cosmos client initialization. To understand this in more detail, please take a look at the number of open tcp connections graph below.

image

image

image

image

It appears that during the time of impact, the RPS was around **15000 Reqs/Sec**, and the number of tcp connections has doubled compared to their historical trends.

**Analysis :** Why this is happening, is because it is possible that while validating the **Unknown and Unhealthy** replicas, during the `OpenChannelAsync()` stage, in the `ChannelDictionary` there is no way to validate if there is a connection already made to the BE replica, thus if there are total **K** partitions, and at a given time **N** number of parallel requests land to this method, they all ended up opening **4 x N x K** numbers of tcp connections, since there are **4** replicas per partition. This is what is causing the sudden spike on the number of open tcp connections.

```
        public Task OpenChannelAsync(
            Uri physicalAddress,
            bool localRegionRequest,
            Guid activityId)
        {
            this.ThrowIfDisposed();
            IChannel channel = this.GetChannel(
                physicalAddress,
                localRegionRequest);

            return channel.OpenChannelAsync(activityId);
        }
```

**Solution :** In order to resolve this, two things are needed to be done:

- Firstly, add a thread protected check on the number of channels created per `ServerKey`, and block any additional channel opening, if a new channel is already established.
- A good optimization for such a large workload would be to stop validating the `Unknown` replicas and validate only the `Unhealthy` replicas when a timeout occurs.

**Detailed Sequence Diagram:** Below is the sequence diagram for validating the channel health before opening the connections.

```mermaid
sequenceDiagram
    participant X as GatewayAddressCache <br> [v3 Code]
    participant A as RntbdOpenConnectionHandler <br> [Direct Code]
    participant B as Rntbd.TransportClient <br> [Direct Code]
    participant C as ChannelDictionary <br> [Direct Code]
    participant D as LoadBalancingChannel <br> [Direct Code]
    participant E as LoadBalancingPartition <br> [Direct Code]
    participant F as Channel <br> [Direct Code]
    Note over X: The open connection process is <br> triggered as a background task.    
    X->>A: 1. TryOpenRntbdChannelsAsync <br> (transport addresses)
    Note over A: The incoming requests <br> acquires the semaphore <br> and opens channel.    
    A->>B: 2. OpenConnectionAsync()
    Note over B: Rntbd.TransportClient <br> implements abstract class <br> Microsoft.Azure.Documents.TransportClient
    B->>C: 3. GetChannel <br> (validationRequired: true)
    C-->>B: 4. Gets a <br> LoadBalancingChannel
    B->>B: 5. Is the returned channel healthy <br> channel.Healthy == true ?
    B->>A: 6. Channel is healthy <br> no further validation is required.
    B->>D: 7. OpenChannelAsync() <br> channel is not healthy <br> validation will be required.
    D->>E: 8. OpenChannelAsync()
    E->>F: 9. OpenChannelAsync()
    F->>E: 10. Returns Channel Init Task    
    E-->>D: 11. await on the init task.
    D-->>D: 12. If success, SetHealthy(true)
    D-->>C: Channel Init Task
    D-->>D: 13. if fails, SetHealthy(false)
    D-->>C: Exception is thrown
    C-->>B: 14. Channel Init Task or Exception.
    B-->>A: 15. Channel Init Task or Exception.
    A-->>A: 16. a) If successful, mark transport address to Connected <br> b) If exception is caught, mark transport address to Unhealthy.
```
Alternate Flow:

```mermaid
sequenceDiagram
    participant X as GatewayAddressCache <br> [v3 Code]
    participant A as RntbdOpenConnectionHandler <br> [Direct Code]
    participant B as Rntbd.TransportClient <br> [Direct Code]
    participant C as ChannelDictionary <br> [Direct Code]
    participant D as LoadBalancingChannel <br> [Direct Code]
    participant E as LoadBalancingPartition <br> [Direct Code]
    participant F as Channel <br> [Direct Code]
    Note over X: The open connection process is <br> triggered as a background task.    
    X->>A: 1. TryOpenRntbdChannelsAsync <br> (transport addresses)
    Note over A: The incoming requests <br> acquires the semaphore <br> and opens channel.    
    A->>B: 2. OpenConnectionAsync()
    Note over B: Rntbd.TransportClient <br> implements abstract class <br> Microsoft.Azure.Documents.TransportClient
    B->>C: 3. GetChannel <br> (validationRequired: true)
    C-->>B: 4. Gets a <br> LoadBalancingChannel
    B->>D: 5. OpenChannelAsync() 
    D->>E: 6. OpenChannelAsync()
    E->>E: 7. Are there any healthy LbChannelState <br> present in the openChannels dictionary ?
    E-->>D: 8. Yes - Healthy LbChannelState present. <br> Skip new channel creation.
    E->>F: 9. No - healthy LbChannelState present. <br> Create Channel using OpenChannelAsync().
    F->>E: 10. Returns Channel Init Task.   
    E-->>D: 11. Returns Channel Init Task.
    D-->>C: 12. Returns Channel Init <br> Task Or No-Op.
    C-->>B: 13. Returns Channel Init <br> Task Or No-Op.
    B-->>A: 14. Returns Channel Init <br> Task Or No-Op.
    A-->>A: 15. a) If successful, mark transport address to Connected <br> b) If exception is caught, mark transport address to Unhealthy.
```

**Benchmark Results:** Below are the benchmark results before and after the fix.

**VM Configuration:** _SKU: Standard D16s v3, vCPU: 16, RAM: 64 GB._
**VM Location:** _West US 2._
**Cosmos Test Account Location:** _West US 2._
**Cosmos Test Account RU Count:** _500K._
**Upgrade Duration:** _5 hours._
**Upgrade Domains:** _20_
**Test Run Time:** _6 Hours._
**RPS:** _Approx. 5000._

TCP Connection Count Before the Fix:

image

TCP Connection Count After the Fix:

image

```[tasklist]
### Tasks
- [ ] https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4034
- [ ] https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4035
```

kundadebdatta:
# Pull Request Template

## Description

**Scope:** Applicable when the "Advanced Replica Selection" feature is enabled using the environment variable `AZURE_COSMOS_REPLICA_VALIDATION_ENABLED = True`

**Problem:** One or our recent conversations with the IC3 team has helped us unveiling a potential issue with the open connection async flow. The team migrated to the latest preview sdk version `3.35.2-preview` and experienced a sudden and consistent spike in number of tcp connections since the cosmos client initialization. To understand this in more detail, please take a look at the number of open tcp connections graph below.

image

image

image

image

It appears that during the time of impact, the RPS was around **15000 Reqs/Sec**, and the number of tcp connections has doubled compared to their historical trends.

**Analysis :**  The reason, this is happening, is because it is possible that while validating both the `Unknown` and `Unhealthy` replicas, during the `OpenChannelAsync()` stage, in the `LoadBalancingPartition`, today there is no way to validate if there is a connection already made to the given endpoint of the BE replica, thus if there are N number of parallel requests land to this method, they all ended up opening N tcp connections for the given endpoint. This is what is causing the sudden spike on the number of open tcp connections.

```
        internal Task OpenChannelAsync(Guid activityId)
        {
            IChannel channel = null;
            this.capacityLock.EnterWriteLock();
            try
            {
                if (this.capacity < this.maxCapacity)
                {
                    channel = this.OpenChannelAndIncrementCapacity(
                        activityId: activityId);
                }
                else
                {
                   ....
                }
            }
         }
```

**Solution:** The code changes in this PR is doing the following - 

- It uses the latest version of the `Cosmos.Direct` version `3.31.5` package, that refactors the behavior in the `LoadBalancingPartition`, and adds an extra check to validate if the `openChannels` dictionary in the `LoadBalancingPartiton` contains a healthy `LbChannelState` for a specific endpoint. If there exists a healthy channel established to that endpoint, then we skip the open connection creation for the given endpoint. This basically means, if there is a healthy connection present for a given endpoint, no new connection will required to be opened.

 - Additionally, this PR adds an additional `RefreshAsync()` method in the `AsyncLazyWithRefreshTask<T>` in the `AsyncCacheNonBlocking<K, V>` that maintains a reference to the background task. If the task is already running for the cache key, then any future Refresh requests will be a no-op. This dramatically reduces the background task creation, thus reducing the number of duplicate replica validations for the same partition key range.
 
 - This PR also reduces the replica validation scope, only to the `Unhealthy` replicas by default. The `Unknown` replicas will be validated only when the `CosmosClient` is initialized with `CreateAndInitializeAsync()` flow.

**Detailed Sequence Diagram:** Below is the sequence diagram for validating the channel health before opening the connections.

```mermaid
sequenceDiagram
    participant X as GatewayAddressCache <br> [v3 Code]
    participant A as RntbdOpenConnectionHandler <br> [Direct Code]
    participant B as Rntbd.TransportClient <br> [Direct Code]
    participant C as ChannelDictionary <br> [Direct Code]
    participant D as LoadBalancingChannel <br> [Direct Code]
    participant E as LoadBalancingPartition <br> [Direct Code]
    participant F as Channel <br> [Direct Code]
    Note over X: The open connection process is <br> triggered as a background task.    
    X->>A: 1. TryOpenRntbdChannelsAsync <br> (transport addresses)
    Note over A: The incoming requests <br> acquires the semaphore <br> and opens channel.    
    A->>B: 2. OpenConnectionAsync()
    Note over B: Rntbd.TransportClient <br> implements abstract class <br> Microsoft.Azure.Documents.TransportClient
    B->>C: 3. GetChannel <br> (validationRequired: true)
    C-->>B: 4. Gets a <br> LoadBalancingChannel
    B->>D: 5. OpenChannelAsync() 
    D->>E: 6. OpenChannelAsync()
    E->>E: 7. Are there any healthy LbChannelState <br> present in the openChannels dictionary ?
    E-->>D: 8. Yes - Healthy LbChannelState present. <br> Skip new channel creation.
    E->>F: 9. No - healthy LbChannelState present. <br> Create Channel using OpenChannelAsync().
    F->>E: 10. Returns Channel Init Task.   
    E-->>D: 11. Returns Channel Init Task.
    D-->>C: 12. Returns Channel Init <br> Task Or No-Op.
    C-->>B: 13. Returns Channel Init <br> Task Or No-Op.
    B-->>A: 14. Returns Channel Init <br> Task Or No-Op.
    A-->>A: 15. a) If successful, mark transport address to Connected <br> b) If exception is caught, mark transport address to Unhealthy.
```

**Benchmark Results:** Below are the benchmark results before and after the fix.

**VM Configuration:** _SKU: Standard D16s v3, vCPU: 16, RAM: 64 GB._
**VM Location:** _West US 2._
**Cosmos Test Account Location:** _West US 2._
**Cosmos Test Account RU Count:** _500K._
**Upgrade Duration:** _5 hours._
**Upgrade Domains:** _20_
**Test Run Time:** _6 Hours._
**RPS:** _Approx. 5000._

TCP Connection Count Before the Fix:

image

TCP Connection Count After the Fix:

image

## Type of change

## Closing issues

To automatically close an issue: closes #4033


Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -302,27 +312,24 @@ public void SetOpenConnectionsHandler(IOpenConnectionsHandler openConnectionsHan
                     .ReplicaTransportAddressUris
                     .Any(x => x.ShouldRefreshHealthStatus()))
                 {
-                    Task refreshAddressesInBackgroundTask = Task.Run(async () =>
+                    try
```
kirankumarkolli, line 315: From offline syncup with Deb

```C#
                    /// 1. Async and NonBlocking
                    /// 2. Reducing #Tasks created
                    /// 3. Scoped to a PkRangeId

                    if (Monitor.TryEnter(addresses, 0)) {
                        try
                        {
                            this.serverPartitionAddressCache.GetAsync(
                                key: partitionKeyRangeIdentity,
                                singleValueInitFunc: (currentCachedValue) => this.GetAddressesForRangeIdAsync(
                                        request,
                                        cachedAddresses: currentCachedValue,
                                        partitionKeyRangeIdentity.CollectionRid,
                                        partitionKeyRangeIdentity.PartitionKeyRangeId,
                                        forceRefresh: true));
                        }
                        finally
                        {
                            Monitor.Exit(addresses);
                        }
                    }
```

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -386,14 +419,39 @@ public bool ShouldRemoveFromCacheThreadSafe()
...
+
             private static bool IsTaskRunning(Task t)
             {
                 if (t == null)
                 {
                     return false;
                 }
 
-                return !t.IsCompleted;
+                return !t.IsCompleted && !t.IsFaulted;
```
kirankumarkolli, line 454: This will have wider impact right?

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -183,17 +183,26 @@ public bool TryRemove(TKey key)
...
-                    callbackDelegate: singleValueInitFunc,
-                    operationName: nameof(RefreshAsync));
+                Task backgroundRefreshTask = initialLazyValue.RefreshAsync(
+                    createRefreshTask: singleValueInitFunc);
+
+                Task continuationTask = backgroundRefreshTask.ContinueWith(task =>
+                {
+                    if (task.IsFaulted)
+                    {
+                        this.RemoveKeyForBackgroundExceptions(
```
FabianMeiswinkel, line 199: Please cross check - whether you need to unwrap AggregateExceptions potentially when excpeitons happen aysnc vs. sync in this code path. async await will do this automatically - the way you handle the errors now I beleive requires this manually - if I recall correctly. This needs test coverage - both when backgroundRefreshTask throws a sync exception (before any await) as well as async (after the first await)

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -386,14 +419,39 @@ public bool ShouldRemoveFromCacheThreadSafe()
...
+                            message: "Background refresh task is already in progress, skip creating a new one.");
+
+                        return Task.CompletedTask;
+                    }
+                    else
+                    {
+                        DefaultTrace.TraceInformation(
+                            message: "Started a new background refresh task.");
+
+                        this.backgroundRefreshTask = Task.Run(async () => await this.CreateAndWaitForBackgroundRefreshTaskAsync(
```
FabianMeiswinkel, line 439: Any exception in CreateAndWaitForBackgroundRefreshTaskAsync would result in process crash if validation for unobserved ewxceptions is enabled (like in Compute). Have you double-checked that CreateAndWaitForBackgroundRefreshTaskAsync is guaranteed to never throw an exception?

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
FabianMeiswinkel: Yeah - please narrow down the side effects - add a parameter whether the isFaulted should be treated as running or not - so, that only your new code is treating it differently

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -218,26 +227,46 @@ public bool TryRemove(TKey key)
             }
             catch (Exception ex)
             {
-                if (initialValue.ShouldRemoveFromCacheThreadSafe())
-                {
-                    bool removed = false;
+                this.RemoveKeyForBackgroundExceptions(
```
ealsur, line 230: Why do you need the ContinueWith above if you are doing this in this catch?

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
ealsur: If the failure is AggregatedException then `task?.Exception?.InnerException` might be empty

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
ealsur: Why not simply use GetAsync with the forceRefresh comparing the hash values as we already do above?

ealsur:
```
his.serverPartitionAddressCache.GetAsync(
                        key: partitionKeyRangeIdentity,
                        singleValueInitFunc: (currentCachedValue) => this.GetAddressesForRangeIdAsync(
                                        request,
                                        cachedAddresses: currentCachedValue,
                                        partitionKeyRangeIdentity.CollectionRid,
                                        partitionKeyRangeIdentity.PartitionKeyRangeId,
                                        forceRefresh: true),
                        forceRefresh: (currentCachedValue) =>
                        {
                            int cachedHashCode = request?.RequestContext?.LastPartitionAddressInformationHashCode ?? 0;
                            if (cachedHashCode == 0)
                            {
                                return true;
                            }
                            // The cached value is different then the previous access hash then assume
                            // another request already updated the cache since there is a new value in the cache
                            return currentCachedValue.GetHashCode() == cachedHashCode;
                        })
```

ealsur: Why is reducing the # of Tasks a problem we are trying to solve here but it is not a problem if, for example, 1000 read requests hit a 410 and ALL of them trigger background address refresh? And if the latter never proved a problem, why is this new code path a problem?

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: This piece of code now no longer needed, since I have refactored it.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: Refactored the condition back to only `return !t.IsCompleted`

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: Based on some of the back and forth conversation, we finally agreed to move forward with the global semaphore based approach. Cc: @ealsur , @kirankumarkolli , @FabianMeiswinkel .

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: I do see that there are possibility this method `CreateAndWaitForBackgroundRefreshTaskAsync ` might throw exceptions. However, this is applicable for `GetAsync()` call as well and it's been there since long time. Do you suggest we should swallow any exceptions to avoid process crash ?

kundadebdatta: So @FabianMeiswinkel  - Right now, I do have a `ContinueWith()` token that avoids to throw any unobserved execptions. I think this should cover us with the process crash.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -183,17 +182,28 @@ public bool TryRemove(TKey key)
         /// </summary>
         /// <param name="key">The requested key to be refreshed.</param>
         /// <param name="singleValueInitFunc">A func delegate to be invoked at a later point of time.</param>
-        public async Task RefreshAsync(
+        public void Refresh(
```
ealsur, line 185: Might be worth updating the docs on this method, stating that this is a background refresh and that the concurrency needs to be handled by the upper layer?

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -183,17 +182,28 @@ public bool TryRemove(TKey key)
...
-                    key: key,
-                    initialValue: initialLazyValue,
-                    callbackDelegate: singleValueInitFunc,
-                    operationName: nameof(RefreshAsync));
+                Task backgroundRefreshTask = this.GetAsync(
+                        key: key,
+                        singleValueInitFunc: singleValueInitFunc,
+                        forceRefresh: (_) => true);
+
+                Task continuationTask = backgroundRefreshTask.ContinueWith(task =>
```
ealsur, line 196: Consider replacing with:

```
.ContinueWith(t => DefaultTrace.TraceWarning("initializeTask failed {0}", t.Exception), TaskContinuationOptions.OnlyOnFaulted);
```

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -183,17 +182,28 @@ public bool TryRemove(TKey key)
...
+                {
+                    if (task.IsFaulted)
+                    {
+                        this.RemoveKeyForBackgroundExceptions(
```
ealsur, line 200: Double check if the removal is needed, the other methods seem to already handle it.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -218,26 +228,46 @@ public bool TryRemove(TKey key)
             }
             catch (Exception ex)
             {
-                if (initialValue.ShouldRemoveFromCacheThreadSafe())
-                {
-                    bool removed = false;
+                this.RemoveKeyForBackgroundExceptions(
```
ealsur, line 231: COnsider undoing this area

ealsur (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayAddressCacheTests.cs
```diff
@@ -1293,9 +1293,140 @@ public async Task TryGetAddressesAsync_WhenReplicaVlidationEnabledAndUnhealthyUr
             GatewayAddressCacheTests.AssertOpenConnectionHandlerAttributes(
                 fakeOpenConnectionHandler: fakeOpenConnectionHandler,
                 expectedTotalFailedAddressesToOpenCount: 1,
-                expectedTotalHandlerInvocationCount: 3,
-                expectedTotalReceivedAddressesCount: 7,
-                expectedTotalSuccessAddressesToOpenCount: 6);
+                expectedTotalHandlerInvocationCount: 2,
+                expectedTotalReceivedAddressesCount: 2,
+                expectedTotalSuccessAddressesToOpenCount: 1);
```
ealsur, line 1298: nit: Why did these numbers improve?

kundadebdatta: Well, one of the primary reason is because right now, the direct code opens one connection per endpoint. If there are existing connections present in the `LoadBalancingPartition`, then no new connections will be created/ opened. This is the reason, all the counts are reduced.

FabianMeiswinkel (approved):
LGTM - Thanks. Very good PR description - thanks for setting an example here!

kundadebdatta:
[like]  Debdatta Kunda reacted to your message:
________________________________

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kundadebdatta: This is now refactored.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -315,14 +328,21 @@ public void SetOpenConnectionsHandler(IOpenConnectionsHandler openConnectionsHan
...
-                            DefaultTrace.TraceWarning("Failed to refresh addresses in the background for the collection rid: {0} with exception: {1}. '{2}'",
+                            DefaultTrace.TraceVerbose("Failed to refresh addresses in the background for the collection rid: {0}, partition key range id: {1}, because the semaphore is already acquired. '{2}'",
                                 partitionKeyRangeIdentity.CollectionRid,
-                                ex,
+                                partitionKeyRangeIdentity.PartitionKeyRangeId,
                                 System.Diagnostics.Trace.CorrelationManager.ActivityId);
                         }
-                    });
+                    }
+                    finally
```
kirankumarkolli, line 339: nit: try-finally be moved inside IF (code refractoring and fainally clause then don't need explicit check

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -302,11 +314,12 @@ public void SetOpenConnectionsHandler(IOpenConnectionsHandler openConnectionsHan
                     .ReplicaTransportAddressUris
                     .Any(x => x.ShouldRefreshHealthStatus()))
                 {
-                    Task refreshAddressesInBackgroundTask = Task.Run(async () =>
+                    bool slimAcquired = await this.semaphore.WaitAsync(0);
```
kirankumarkolli, line 317: Clarification: What will happen if this check not exists?

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
```diff
@@ -315,14 +328,21 @@ public void SetOpenConnectionsHandler(IOpenConnectionsHandler openConnectionsHan
...
                                 partitionKeyRangeIdentity.CollectionRid,
-                                ex,
+                                partitionKeyRangeIdentity.PartitionKeyRangeId,
                                 System.Diagnostics.Trace.CorrelationManager.ActivityId);
                         }
-                    });
+                    }
+                    finally
+                    {
+                        if (slimAcquired)
```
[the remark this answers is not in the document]
kundadebdatta: Discussed this offline. The below code should cover this:

```
if (addresses
    .Get(Protocol.Tcp)
    .ReplicaTransportAddressUris
    .Any(x => x.ShouldRefreshHealthStatus()))
```

The semaphore will be released as soon as the task is successfully scheduled. and the above check is good enough to block other parallel threads to create duplicate tasks.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: If the below condition doesn't exist, then there is no gate to detect if other threads has acquired the semaphore in the past and already scheduled a background refresh task. So, without the check, the `try` block will get executed.

```
        if (slimAcquired)
        {
            this.serverPartitionAddressCache.Refresh(
                key: partitionKeyRangeIdentity,
                singleValueInitFunc: (currentCachedValue) => this.GetAddressesForRangeIdAsync(
                    request,
                    cachedAddresses: currentCachedValue,
                    partitionKeyRangeIdentity.CollectionRid,
                    partitionKeyRangeIdentity.PartitionKeyRangeId,
                    forceRefresh: true));
        }
```

Also, the code snippet `bool slimAcquired = await this.semaphore.WaitAsync(0);` checks the semaphore hook is acquired at that point of time. If it's already acquired, then the thread immediately returns a false and continue it's execution. This guarantees that no duplicate tasks were created.

Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs
kundadebdatta: Agreed. I will refactor this with the next PR to master. Good catch!

PR #4158, opened 2023-10-26
[Internal] Query: Fixes optimalPageSize logic for OFFSET LIMIT in ORDER BY queries

sc978345:
## Description

This PR fixes the logic in CosmosQueryExecutionContextFactory where we determine optimal page size for ORDER BY queries that have an OFFSET/LIMIT clause.  Previously, the logic was only being applied to TOP and not OFFSET/LIMIT, resulting in suboptimal query execution (unnecessarily fetching more results than are likely to be needed by the query)

Consider an example query:
SELECT c.pk FROM c ORDER BY c.pk OFFSET 0 LIMIT 500

Without this change, we use a page size of 500 against each target range.  With this change, we will calculate an optimal page size based on the limit on the # of documents being requested (factoring in the number of target ranges) using the same formula that we already use for TOP queries:

```
                    // All partitions should initially fetch about 1/nth of the top value.
                    long pageSizeWithTop = (long)Math.Min(
                        Math.Ceiling(top / (double)targetRanges.Count) * CosmosQueryExecutionContextFactory.PageSizeFactorForTop,
                        top);

                    optimalPageSize = Math.Min(pageSizeWithTop, optimalPageSize);

```
## Type of change


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/TestCosmosQueryClient.cs
```diff
@@ -0,0 +1,107 @@
...
+            throw new NotImplementedException();
+        }
+
+        public override async Task<TryCatch<PartitionedQueryExecutionInfo>> TryGetPartitionedQueryExecutionInfoAsync(SqlQuerySpec sqlQuerySpec, ResourceType resourceType, PartitionKeyDefinition partitionKeyDefinition, bool requireFormattableOrderByQuery, bool isContinuationExpected, bool allowNonValueAggregateQuery, bool hasLogicalPartitionKey, bool allowDCount, bool useSystemPrefix, Cosmos.GeospatialType geospatialType, CancellationToken cancellationToken)
+        {
+            CosmosSerializerCore serializerCore = new();
+            using StreamReader streamReader = new(serializerCore.ToStreamSqlQuerySpec(sqlQuerySpec, Documents.ResourceType.Document));
+            string sqlQuerySpecJsonString = streamReader.ReadToEnd();
+
+            TryCatch<PartitionedQueryExecutionInfo> queryPlan = OptimisticDirectExecutionQueryBaselineTests.TryGetPartitionedQueryExecutionInfo(sqlQuerySpecJsonString, partitionKeyDefinition);
```
neildsh, line 103: >OptimisticDirectExecutionQueryBaselineTests.TryGetPartitionedQueryExecutionInfo [](http://example.com/codeflow?start=64&length=79)

Please avoid ODE specific logic in generic test infrastructure

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/TestCosmosQueryClient.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            List<PartitionKeyRange> partitionKeyRanges = new List<PartitionKeyRange>{new PartitionKeyRange()
+                {
+                    MinInclusive = PartitionKeyInternal.MinimumInclusiveEffectivePartitionKey,
+                    MaxExclusive = PartitionKeyInternal.MaximumExclusiveEffectivePartitionKey
+                }};
+
+            for (int i = 1; i < this.targetRanges; i++)
+            {
+                partitionKeyRanges.Add(new PartitionKeyRange() { MinInclusive = string.Empty, MaxExclusive = string.Empty });
```
neildsh, line 87: >partitionKeyRanges.Add(new PartitionKeyRange() { MinInclusive = string.Empty, MaxExclusive = string.Empty }); [](http://example.com/codeflow?start=16&length=109)

This seems like a bug. we should not have multiple overlapped ranges that are empty like this

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/TestCosmosQueryClient.cs
```diff
@@ -0,0 +1,107 @@
...
+    using Microsoft.Azure.Cosmos.Query.Core.Monads;
+    using Microsoft.Azure.Cosmos.Query.Core.Pipeline.Pagination;
+    using Microsoft.Azure.Cosmos.Query.Core.QueryClient;
+    using Microsoft.Azure.Cosmos.Query.Core.QueryPlan;
+    using Microsoft.Azure.Cosmos.Routing;
+    using Microsoft.Azure.Cosmos.Tracing;
+    using Microsoft.Azure.Documents;
+    using Microsoft.Azure.Documents.Routing;
+
+    internal class TestCosmosQueryClient : CosmosQueryClient
```
neildsh, line 20: >TestCosmosQueryClient [](http://example.com/codeflow?start=19&length=21)

You may be better served by leaving this class inside the ODE tests and just creating a new mock object (using Moq) for your own test... It doesn't seem like you need a lot of functionality from this object.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/FullPipelineTests.cs
```diff
@@ -461,5 +574,80 @@ public async Task<Exception> ShouldReturnFailure()
...
+                this.MonadicDocumentContainer = documentContainer;
+                this.PageSizeSpecified = null;
+            }
+
+            public void Reset()
+            {
+                this.PageSizeSpecified = null;
+            }
+
+            public IMonadicDocumentContainer MonadicDocumentContainer { get; }
```
neildsh, line 593: >public IMonadicDocumentContainer MonadicDocumentContainer { get; } [](http://example.com/codeflow?start=12&length=66)

Please group all properties at the top for readability

neildsh:
Thanks for making this fix!

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/FullPipelineTests.cs
```diff
@@ -249,6 +256,112 @@ public async Task Tracing()
...
+            await this.TestPageSizeAsync("SELECT TOP 15000 c.pk FROM c", 1000, mockInMemoryContainer, documentContainer);
+        }
+
+        private async Task TestPageSizeAsync(string query, int expectedPageSize, MockInMemoryContainer inMemoryContainer, DocumentContainer documentContainer)
+        {
+            (CosmosQueryExecutionContextFactory.InputParameters inputParameters, CosmosQueryContextCore cosmosQueryContextCore) = CreateInputParamsAndQueryContext(
+                query,
+                partitionKeyDefinition,
+                null,
+                64,
```
neildsh, line 301: >64 [](http://example.com/codeflow?start=16&length=2)

[Nit] Please name literal arguments

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/FullPipelineTests.cs
```diff
@@ -249,6 +256,112 @@ public async Task Tracing()
...
+            while (await queryPipelineStage.MoveNextAsync(NoOpTrace.Singleton))
+            {
+                TryCatch<QueryPage> tryGetQueryPage = queryPipelineStage.Current;
+                tryGetQueryPage.ThrowIfFailed();
+
+                elements.AddRange(tryGetQueryPage.Result.Documents);
+            }
+
+            Assert.AreEqual(expected: expectedPageSize, actual: inMemoryContainer.PageSizeSpecified);
+        }
```
neildsh, line 320: >} [](http://example.com/codeflow?start=8&length=1)

Please consider validating the actual results in addition to the page size

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/TestCosmosQueryClient.cs
sc978345: No longer applicable

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/FullPipelineTests.cs
sc978345: Added check on actual results

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/FullPipelineTests.cs
sc978345: Done

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/FullPipelineTests.cs
sc978345: Done

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/TestCosmosQueryClient.cs
sc978345: Done

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/TestCosmosQueryClient.cs
sc978345: Previously there was a single full range and N-1 empty ranges to simulate target ranges.  I'm not sure if empty ranges are considered overlapping (in my head they are not).  But regardless, it's better to try to match real world values so I have added non-overlapping ranges.

neildsh (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
```diff
@@ -550,8 +550,17 @@ internal static class CosmosQueryExecutionContextFactory
             long optimalPageSize = inputParameters.MaxItemCount;
             if (queryInfo.HasOrderBy)
             {
-                int top;
-                if (queryInfo.HasTop && (top = partitionedQueryExecutionInfo.QueryInfo.Top.Value) > 0)
+                int top = 0;
```
sboshra, line 553: > = 0 [](http://example.com/codeflow?start=23&length=4)

Do not initialize local variables; rather, add an else statement that sets it the value to 0.

sboshra (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
sc978345: Done

sboshra (approved):

neildsh (approved):

PR #4163, opened 2023-11-01
[Internal] Query: Adds interface for linq serialization functions

Maya-Painter:
## Description

No behavioral changes. Refactors LINQ translation classes so that all serialization functions are defined on a single interface.

## Type of change


Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -249,7 +244,7 @@ internal static SqlScalarExpression VisitNonSubqueryScalarExpression(Expression
                 case ExpressionType.Conditional:
                     return ExpressionToSql.VisitConditional((ConditionalExpression)inputExpression, context);
                 case ExpressionType.Constant:
-                    return ExpressionToSql.VisitConstant((ConstantExpression)inputExpression, context);
+                    return context.CosmosLinqSerializer.VisitConstant((ConstantExpression)inputExpression, context);
```
leminh98, line 247: >context.CosmosLinqSerializer. [](http://example.com/codeflow?start=27&length=29)

This should stay as ExpressionToSql.VisitConstant, and then inside of it call to context.CosmosLinqSerializer.VisitConstant 
to maintain the visitor design pattern

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -708,71 +610,6 @@ private static SqlScalarExpression VisitTypeIs(TypeBinaryExpression inputExpress
             throw new DocumentQueryException(string.Format(CultureInfo.CurrentCulture, ClientResources.ExpressionTypeIsNotSupported, inputExpression.NodeType));
         }
 
-        public static SqlScalarExpression VisitConstant(ConstantExpression inputExpression, TranslationContext context)
```
leminh98, line 711: >VisitConstant [](http://example.com/codeflow?start=42&length=13)

As mentioned before, please keep this and then call into the customer serializer class in here.

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -798,7 +635,7 @@ private static SqlScalarExpression VisitParameter(ParameterExpression inputExpre
         private static SqlScalarExpression VisitMemberAccess(MemberExpression inputExpression, TranslationContext context)
         {
             SqlScalarExpression memberExpression = ExpressionToSql.VisitScalarExpression(inputExpression.Expression, context);
-            string memberName = inputExpression.Member.GetMemberName(context.linqSerializerOptions);
+            string memberName = inputExpression.Member.GetMemberName(context.CosmosLinqSerializer, context.LinqSerializerOptions);
```
leminh98, line 638: >CosmosLinqSerializer [](http://example.com/codeflow?start=77&length=20)

Should the CosmosLinqSerializer takes the linqSerializerOptions as an input?

leminh98: It seems that everywhere we pass in the CosmosLinqSerializer, we also pass in the LinqSerializerOption, so maybe it would be easier to pass the serializer options in the moment we create the Linqserialier

Maya-Painter: LinqSerializerOption is only passed in for `GetMemberName()`, not the other 2 functions on the interface (for now, this may change when implementing the DotNet serializer).

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqSerializer.cs
```diff
@@ -0,0 +1,305 @@
...
+
+            if (linqSerializerOptions != null)
+            {
+                memberName = CosmosSerializationUtil.GetStringWithPropertyNamingPolicy(linqSerializerOptions, memberName);
+            }
+
+            return memberName;
+        }
+
+        private sealed class CosmosElementToSqlScalarExpressionVisitor : ICosmosElementVisitor<SqlScalarExpression>
```
leminh98, line 227: >CosmosElementToSqlScalarExpressionVisitor [](http://example.com/codeflow?start=29&length=41)

nit: separate file

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqSerializer.cs
```diff
@@ -0,0 +1,305 @@
...
+
+                        return CosmosElement.Parse(serializedValue).Accept(CosmosElementToSqlScalarExpressionVisitor.Singleton);
+                    }
+                }
+            }
+
+            return right;
+        }
+
+        public SqlScalarExpression VisitConstant(ConstantExpression inputExpression, IDictionary<object, string> parameters)
```
leminh98, line 118: >VisitConstant [](http://example.com/codeflow?start=35&length=13)

This function is not a visitor function, so consider a different name. Maybe: ConvertToSqlScalarExpression 

or maybe just Convert

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqSerializer.cs
```diff
@@ -0,0 +1,305 @@
...
+    using System.Reflection;
+    using System.Runtime.Serialization;
+    using Microsoft.Azure.Cosmos.CosmosElements;
+    using Microsoft.Azure.Cosmos.CosmosElements.Numbers;
+    using Microsoft.Azure.Cosmos.Spatial;
+    using Microsoft.Azure.Cosmos.SqlObjects;
+    using Microsoft.Azure.Documents;
+    using Newtonsoft.Json;
+
+    internal class CosmosLinqSerializer : ICosmosLinqSerializer
```
leminh98, line 23: > internal class CosmosLinqSerializer : ICosmosLinqSerializer [](http://example.com/codeflow?start=3&length=60)

Can you add docstring in front of each function to document what each functino does?

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.Linq
+{
+    using System.Collections.Generic;
+    using System.Linq.Expressions;
+    using System.Reflection;
+    using Microsoft.Azure.Cosmos.SqlObjects;
+
+    internal interface ICosmosLinqSerializer
```
leminh98, line 11: >ICosmosLinqSerializer [](http://example.com/codeflow?start=23&length=21)

docstring here also

Microsoft.Azure.Cosmos/src/Linq/TranslationContext.cs
```diff
@@ -29,30 +33,36 @@ internal sealed class TranslationContext
...
-        public QueryUnderConstruction currentQuery;
+        public QueryUnderConstruction CurrentQuery;
 
         /// <summary>
         /// Dictionary for parameter name and value
         /// </summary>
-        public IDictionary<object, string> parameters;
+        public IDictionary<object, string> Parameters;
+
+        public CosmosLinqSerializerOptions LinqSerializerOptions;
```
leminh98, line 43: >public CosmosLinqSerializerOptions LinqSerializerOptions; [](http://example.com/codeflow?start=8&length=57)

move this up next to the CosmosLinqSerializer and add 
/// <summary>

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAttributeContractBaselineTests.cs
```diff
@@ -173,10 +173,11 @@ public Datum2(string jsonProperty, string dataMember, string defaultMember, stri
         [TestMethod]
         public void TestAttributePriority()
         {
-            Assert.AreEqual("jsonProperty", TypeSystem.GetMemberName(typeof(Datum).GetMember("JsonProperty").First()));
-            Assert.AreEqual("dataMember", TypeSystem.GetMemberName(typeof(Datum).GetMember("DataMember").First()));
-            Assert.AreEqual("Default", TypeSystem.GetMemberName(typeof(Datum).GetMember("Default").First()));
-            Assert.AreEqual("jsonPropertyHasHigherPriority", TypeSystem.GetMemberName(typeof(Datum).GetMember("JsonPropertyAndDataMember").First()));
+            ICosmosLinqSerializer cosmosLinqSerializer = new CosmosLinqSerializer();
```
leminh98, line 176: > ICosmosLinqSerialize [](http://example.com/codeflow?start=11&length=21)

perhaps also new tests for the other two functions of the interface

Microsoft.Azure.Cosmos/src/Linq/TranslationContext.cs
```diff
@@ -65,15 +75,14 @@ public TranslationContext(CosmosLinqSerializerOptions linqSerializerOptions, IDi
             this.methodStack = new List<MethodCallExpression>();
             this.lambdaParametersStack = new List<ParameterExpression>();
             this.collectionStack = new List<Collection>();
-            this.currentQuery = new QueryUnderConstruction(this.GetGenFreshParameterFunc());
+            this.CurrentQuery = new QueryUnderConstruction(this.GetGenFreshParameterFunc());
             this.subqueryBindingStack = new Stack<SubqueryBinding>();
-            this.linqSerializerOptions = linqSerializerOptions;
-            this.parameters = parameters;
-            this.memberNames = new MemberNames(linqSerializerOptions);
+            this.LinqSerializerOptions = linqSerializerOptions;
```
leminh98, line 80: >linqSerializerOptions [](http://example.com/codeflow?start=41&length=21)

is this nullable? do we need null checks?

leminh98 (approved):
:shipit:

leminh98 (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Linq/TranslationContext.cs
Maya-Painter: null checks are happening within the CosmosLinqSerializer

Microsoft.Azure.Cosmos/src/Linq/CosmosElementToSqlScalarExpressionVisitor.cs
```diff
@@ -0,0 +1,90 @@
...
+                items.Add(item.Accept(this));
+            }
+
+            return SqlArrayCreateScalarExpression.Create(items.ToImmutableArray());
+        }
+
+        public SqlScalarExpression Visit(CosmosBinary cosmosBinary)
+        {
+            // Can not convert binary to scalar expression without knowing the API type.
+            throw new NotImplementedException();
```
adityasa, line 36: Do you expect this code to be called? If not, I would add a Debug fail as below for readability:
Debug.Fail("CosmosElementToSqlScalarExpressionVisitor Assert", "Unreachable!");

Microsoft.Azure.Cosmos/src/Linq/CosmosElementToSqlScalarExpressionVisitor.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public SqlScalarExpression Visit(CosmosBoolean cosmosBoolean)
+        {
+            return SqlLiteralScalarExpression.Create(SqlBooleanLiteral.Create(cosmosBoolean.Value));
+        }
+
+        public SqlScalarExpression Visit(CosmosGuid cosmosGuid)
+        {
+            // Can not convert guid to scalar expression without knowing the API type.
+            throw new NotImplementedException();
```
adityasa, line 47: Same as above.

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>
+        /// Applies specified custom converters to an expression.
+        /// </summary>
+        SqlScalarExpression ApplyCustomConverters(Expression left, SqlLiteralScalarExpression right);
```
adityasa, line 16: >SqlScalarExpression [](http://example.com/codeflow?start=8&length=19)

The return type doesn't "add up" for a serializer interface. A serializer would typically return string/void (and accept stream for writing) etc.

Maya-Painter: In this case should this class be renamed?

Microsoft.Azure.Cosmos/src/Linq/CosmosElementToSqlScalarExpressionVisitor.cs
Maya-Painter: added Debug.Fail and updated exception type

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
adityasa, line 16: >ApplyCustomConverters [](http://example.com/codeflow?start=28&length=21)

This method signature (name, input, return type) seem too tied to the existing implementation. Is there a way to generalize this better?

adityasa: That would be one way, although not sure if that's the desired approach since we are trying to genuinely abstract the "serialization" concept and have at least one other possible implementation on the way (which is hopefully more straightforward and not as clunky as existing one).

Microsoft.Azure.Cosmos/src/Linq/DefaultCosmosLinqSerializer.cs
```diff
@@ -0,0 +1,226 @@
...
+            {
+                memberExpression = unaryExpression.Operand as MemberExpression;
+            }
+            else
+            {
+                memberExpression = left as MemberExpression;
+            }
+
+            if (memberExpression != null)
+            {
```
adityasa, line 37: It doesn't seem like this should be serializer's responsibility. Default serializer should limit itself to attribute checking and conversion to string. Which of the expressions need to be serialized (right v/s left) should still be decided by the ExpressionToSQL (since it would apply equally for every serializer (System.Text based, Newtonsoft based etc)?

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
Maya-Painter: pulled out serialization aspects of this so that less is happening in the serializer - just retrieving the converter and applying it if needed.

Maya-Painter: Updated

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -538,14 +514,21 @@ private static SqlScalarExpression ApplyCustomConverters(Expression left, SqlLit
                     // Enum
                     if (memberType.IsEnum())
                     {
```
adityasa, line 516: Correct me if I am wrong, but based on the comment above, this seems specific to Newtonsoft based serializer (giving higher priority to property over types)?

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -538,14 +514,21 @@ private static SqlScalarExpression ApplyCustomConverters(Expression left, SqlLit
...
-                            value = Enum.ToObject(memberType, Number64.ToDouble(number64));
+                            Number64 number64 = ((SqlNumberLiteral)right.Literal).Value;
+                            if (number64.IsDouble)
+                            {
+                                value = Enum.ToObject(memberType, Number64.ToDouble(number64));
+                            }
+                            else
+                            {
+                                value = Enum.ToObject(memberType, Number64.ToLong(number64));
+                            }
```
adityasa, line 527: Is this specific to Newtonsoft or would apply to other serializers?

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Gets custom attributes on a member expression. Returns null if none exist.
+        /// </summary>
+        CustomAttributeData GetConverterAttribute(MemberExpression memberExpression, Type memberType);
+
+        /// <summary>
+        /// Serializes object with provided custom converter.
+        /// </summary>
+        string SerializeWithConverter(object value, Type converterType);
```
adityasa, line 22: >SerializeWithConverter [](http://example.com/codeflow?start=15&length=22)

This seems a bit odd/newtonsoft specific that may not be applicable to other serializers.
Can this be rewritten as:

__string Serialize (object value, <<AnythingElseThat allows ICosmosLinqSerializer to internally decide whether to use converter or not>>);__

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
Maya-Painter: All of the serializers (DotNet included) have a CustomAttributeData that can be applied, so I don't believe it is Newtonsoft specific.

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
Maya-Painter: Pretty sure this applies to all serializers.  In testing this code would be hit when enums were decorated with DotNet properties.

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
Maya-Painter: converter type also applies to DotNet

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -504,30 +503,7 @@ private static SqlScalarExpression ApplyCustomConverters(Expression left, SqlLit
...
-                // Newtonsoft gives high precedence to the attribute specified
-                // on a property over on a type (class/enum)
-                // so we check both attributes and apply the same precedence rules
-                // JsonConverterAttribute doesn't allow duplicates so it's safe to
-                // use FirstOrDefault()
-                CustomAttributeData memberAttribute = memberExpression.Member.CustomAttributes.Where(ca => ca.AttributeType == typeof(JsonConverterAttribute)).FirstOrDefault();
-                CustomAttributeData typeAttribute = memberType.GetsCustomAttributes().Where(ca => ca.AttributeType == typeof(JsonConverterAttribute)).FirstOrDefault();
-
-                CustomAttributeData converterAttribute = memberAttribute ?? typeAttribute;
+                CustomAttributeData converterAttribute = context.CosmosLinqSerializer.GetConverterAttribute(memberExpression, memberType);
```
adityasa, line 506: >converterAttribute [](http://example.com/codeflow?start=36&length=18)

Do we need the actual CustomAttribute, or just a boolean indicating whether custom serialization is needed for this expression?

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
adityasa: Possibly, but I am trying to see if we can rewrite GetConverterAttribute as bool IsCustomSerializationRequired which can also help further simplify this?

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Serializes a ConstantExpression.
+        /// </summary>
+        string SerializeScalarExpression(ConstantExpression inputExpression);
+
+        /// <summary>
+        /// Gets a member name with LINQ serializer options applied.
+        /// </summary>
+        string GetMemberName(MemberInfo memberInfo, CosmosLinqSerializerOptions linqSerializerOptions = null);
```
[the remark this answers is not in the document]
adityasa: Ah, I see a caller from TypeSystem.cs

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
Maya-Painter: Ohh I see what you're saying.  Yeah I think that works?  The bool might not be required to be passed in since this function isn't called unless a converter is found

Maya-Painter: On second thought don't think this code will be any simpler.  You'll have to retrieve the converter multiple times (once in GetConverterAttribute() to make sure the converter exists and once in this function).  You will also have to now pass in all the parameters in GetConverterAttribute() to this function to create the converter.

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
Maya-Painter: We can do just a boolean but this will require creating the custom attributes twice.

adityasa: I think retrieving the custom attributes more than once should be fine for the purposes of abstraction. I don't think we need to worry about the perf hit here.

Point with a new interface is that the actual mechanism for custom serialization (attributes v/s methods v/s not supported) is abstracted sufficiently to support different serialization mechanisms.

Maya-Painter: Updated.

Microsoft.Azure.Cosmos/src/Linq/ICosmosLinqSerializer.cs
Maya-Painter: Updated to no longer take a converter as a param.

Microsoft.Azure.Cosmos/src/Linq/DefaultCosmosLinqSerializer.cs
```diff
@@ -0,0 +1,110 @@
...
+            return memberAttribute != null || typeAttribute != null;
+        }
+
+        public string Serialize(object value, MemberExpression memberExpression, Type memberType)
+        {
+            CustomAttributeData memberAttribute = memberExpression.Member.CustomAttributes.FirstOrDefault(ca => ca.AttributeType == typeof(Newtonsoft.Json.JsonConverterAttribute));
+            CustomAttributeData typeAttribute = memberType.GetsCustomAttributes().FirstOrDefault(ca => ca.AttributeType == typeof(Newtonsoft.Json.JsonConverterAttribute));
+            CustomAttributeData converterAttribute = memberAttribute ?? typeAttribute;
+
+            Debug.Assert(converterAttribute.ConstructorArguments.Count > 0);
```
adityasa, line 51: Typically with Debug Assert, it's a good practice to add short and long messages:
Debug.Assert(condition, $"{nameof(DefaultCosmosLinqSerializer)} Assert!", "At least one constructor argument is found!");

adityasa (dismissed):
:shipit:

adityasa (approved):

akotalwar (approved):

PR #4165, opened 2023-11-02
[Internal] Client Telemetry: Adds code to log exception related to it in Request Diagnostics

sourabh1007:
## Description

**Problem Statement:** Today, if anything goes wrong in Client Telemetry Job OR during collection of datapoint, the only way to debug it, is enable tracing and get the logs. which is not very useful because customers don't enable tracing in prod environment (it is not recommended also) and they need lot of support if they try to enable it in non-prod/prod environment.
Hence, we need to find out a better way to get these errors logged by default so that it can be fixed/optimized in future versions.

**Solution:** 
There are 2 types of scenarios:
1. which happens in datapoint collection during a request: 
image

2. which happens in background Job: We decided to log it in traces only.

## Type of change


kirankumarkolli:
Please update the description

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
```diff
@@ -44,7 +44,8 @@ public TelemetryHandler(TelemetryToServiceHelper telemetryToServiceHelper)
                             ConsistencyLevel = request.Headers?[Documents.HttpConstants.HttpHeaders.ConsistencyLevel],
                             RequestCharge = response.Headers.RequestCharge,
                             SubStatusCode = response.Headers.SubStatusCode,
-                            Trace = response.Trace
+                            Trace = response.Trace,
```
kirankumarkolli, line 47: Both are assigned to the same ITrace, when/how are they expected to deviate?

Microsoft.Azure.Cosmos/src/Telemetry/Collector/TelemetryCollector.cs
```diff
@@ -26,45 +26,53 @@ internal class TelemetryCollector : ITelemetryCollector
...
 
                     data.DatabaseId = databaseName;
                     data.ContainerId = collectionName;
                 }
-
                 this.clientTelemetry?.PushCacheDatapoint(cacheName, data);
             }
             catch (Exception ex)
             {
+                data.TraceToLog.AddDatum($"{ClientTelemetryOptions.TelemetryCollectFailedKeyPrefix}-{cacheName}", ex);
```
kirankumarkolli, line 43: Isn't the decision to **not include** any telemetry context in reqeust diagnostics?

sourabh1007:
> onfigure the tracing

It's not about tracing...it's about adding exception, in request diagnostics json, occur during collection of datapoint because collection is part of request lifecycle. I think that's what we discussed (if I am not missing anything)..that collection part we can record in request diagnostics but for exceptions in background job, we need to dependent on trace.

There will be separate PR on "how to use Ilogger with SDK and leverage SDK trace logging?". It will be a sample I am going to add.

Microsoft.Azure.Cosmos/src/Handler/TelemetryHandler.cs
sourabh1007: one is `TraceTolog` which I am using to add log in request diagnostics, so that it can come in the required format (ref. screenshot) and other one is response trace which is used to collect metrics

sourabh1007:
> > onfigure the tracing
> 
> It's not about tracing...it's about adding exception, in request diagnostics json, occur during collection of datapoint because collection is part of request lifecycle. I think that's what we discussed (if I am not missing anything)..that collection part we can record in request diagnostics but for exceptions in background job, we need to dependent on trace.
> 
> There will be separate PR on "how to use Ilogger with SDK and leverage SDK trace logging?". It will be a sample I am going to add.

@kirankk @FabianMeiswinkel What is your thought on a scenario, where customer is already onboarded on client telemetry and due to some (may be network config change or a bug in SDK) reason after some time, it stops sending us telemetry, how would we even know it until unless that customer comes with some issue, hoping that SDK telemetry is going to help. 
IMO, it will be a bad experience for a customer and might discourage him/her to enable it further.
Here, I am talking about a normal customer, not an ideal customer who has all the logger and alerts setup and proactively contact us if they see any issue.

Microsoft.Azure.Cosmos/src/Telemetry/Collector/TelemetryCollector.cs
sourabh1007: I thought collection part, we agreed to include.

sourabh1007:
Never got approved hence closing it.

PR #4176, opened 2023-11-07
[Internal] Query: Adds Baseline test for exceeding header length for index advisor

leminh98:
# Pull Request Template

## Description

This PR adds test coverage to baseline the behavior of the index advisor when the response is too big

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/IndexMetricsParserBaselineTest.cs
```diff
@@ -131,9 +131,46 @@ public void IndexUtilizationParse()
             this.ExecuteTestSuite(inputs);
         }
 
-        public override IndexMetricsParserTestOutput ExecuteTest(IndexMetricsParserTestInput input)
+        [TestMethod]
+        public void IndexUtilizationHeaderLengthTest()
```
adityasa, line 135: >IndexUtilizationHeaderLengthTest [](http://example.com/codeflow?start=20&length=32)

Curious, this is still using old format (base64 encoded string), correct?

adityasa: Can you add coverage for both (old & new) formats?

adityasa (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/IndexMetricsParserBaselineTest.cs
leminh98: In this current version it is base 64 encoded. It's a little challenging to do this because the query response in base64 expect a base64 and parse it right away so we can't add both test for this right now. 
Once we have the flag for the plaintext switch, I will update this test for both scenarios

adityasa (approved):

sboshra (approved):
:shipit:

PR #4183, opened 2023-11-14
[Internal] Query: Adds ability to send plain text response as the default

leminh98:
# Pull Request Template

## Description

This PR enables the V2 request header for plain text for Index Advisor as the default. .NET SDK version from this point on will no longer send V1 requests  (and therefore will no longer receive base64 responses)

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Query/v3Query/QueryResponse.cs
```diff
@@ -189,7 +189,7 @@ class QueryResponse<T> : FeedResponse<T>
             // TODO: Switch the flag to false once V2 is deployed
```
adityasa, line 189: Remove?

adityasa, line 189: Please update the comment since the TODO is being addressed.

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -269,7 +269,7 @@ internal override void PopulateRequestOptions(RequestMessage request)
             if (this.PopulateIndexMetrics.HasValue)
             {
                 // TODO: Switch to V2
```
adityasa, line 271: Nit, please remove.

adityasa (approved):
:shipit:

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -269,7 +269,7 @@ internal override void PopulateRequestOptions(RequestMessage request)
...
-                request.Headers.CosmosMessageHeaders.Add(HttpConstants.HttpHeaders.PopulateIndexMetrics, this.PopulateIndexMetrics.ToString());
+                request.Headers.CosmosMessageHeaders.Add(HttpConstants.HttpHeaders.PopulateIndexMetricsV2, this.PopulateIndexMetrics.ToString());
```
adityasa, line 272: >PopulateIndexMetricsV2 [](http://example.com/codeflow?start=83&length=22)

Is this supported by the emulator (since I thought it's not yet updated)?

adityasa:
Did you not have to update any test baseline since the output is in plaintext now (with a different format)?

leminh98:
> Did you not have to update any test baseline since the **_default_** index advice is in plaintext now?

The baseline wasn't updated because the public emulator used for test run is out of date, so I will double check behavior with the back-end, and then mark the current tests [Ignore] for now

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
leminh98: Yeah the current emulator when it sees this will return empty response.

leminh98: Now that the emulator is updated, this should be supported by emulator

leminh98:
The emulator is updated so I will double check the behavior 

---
In reply to: [1839592523](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4183#issuecomment-1839592523) [](http://example.com/codeflow?ancestors=1839592523)

sboshra (approved):
:shipit:

Maya-Painter (approved):

PR #4213, opened 2023-12-12
[Internal] Automation: Fixes check for issue assignment

ealsur:
Our current automation will automatically close Issues without activity in 2 weeks if they **have the label** with `needs-more-information`.

The goal of the label is to mark that we have responded but now in order to proceed, we need input from the creator of the Issue.

So far the automation would only trigger if the issue has noone assigned, but this was leaving cases where there was someone assigned, which asked for more information, but there was no answer from the creator.

This PR removes the requirement of "Issue not being assigned"

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -20,7 +20,7 @@ configuration:
           label: needs-more-information
       - noActivitySince:
           days: 14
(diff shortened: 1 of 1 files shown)


kirankumarkolli (approved):

FabianMeiswinkel (approved):
LGTM

PR #4218, opened 2023-12-14
Distributed Tracing: Refactors Samples to use latest cosmosdb SDK

sourabh1007:
## Description

As part of this PR, update Samples with latest SDK i.e. 3.37.0 and updated AppInsights SDK and Azure Monitor library to latest stable version.

Added few more activity span to make results more meaningful on AppInsights UI

### DIFF
--- a/Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
+++ b/Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
@@ -51,14 +51,14 @@
                 _telemetryClient = serviceProvider.GetRequiredService<TelemetryClient>();
                 // </SetUpApplicationInsights>
 
-                var infoOperation = _telemetryClient.StartOperation<DependencyTelemetry>(".Net SDK (v3.37.0) : ApplicationInsights SDK (v2.22.0)");
+                var infoOperation = _telemetryClient.StartOperation<DependencyTelemetry>(".Net SDK : ApplicationInsights SDK"); // Application level activity to track the entire execution of the application
 
-                var gops = _telemetryClient.StartOperation<DependencyTelemetry>("GATEWAY MODE");
+                var gops = _telemetryClient.StartOperation<DependencyTelemetry>("GATEWAY MODE"); // Activity to track the execution of the gateway mode
                 await Program.RunCosmosDbOperation(ConnectionMode.Gateway, endpoint, authKey);
                 _telemetryClient.StopOperation(gops);
 
-                var dops = _telemetryClient.StartOperation<DependencyTelemetry>("DIRECT MODE");
-                await Program.RunCosmosDbOperation(ConnectionMode.Direct, endpoint, authKey);
+                var dops = _telemetryClient.StartOperation<DependencyTelemetry>("DIRECT MODE"); // Activity to track the execution of the direct mode
+                await Program.RunCosmosDbOperation(ConnectionMode.Direct, endpoint, authKey); 
                 _telemetryClient.StopOperation(dops);
 
                 _telemetryClient.StopOperation(infoOperation);
@@ -75,6 +75,7 @@
 
         private static async Task RunCosmosDbOperation(ConnectionMode connMode, string endpoint, string authKey)
         {
+            // <EnableDistributedTracing>
             CosmosClientOptions options = new CosmosClientOptions()
             {
                 CosmosClientTelemetryOptions = new CosmosClientTelemetryOptions()
@@ -83,8 +84,8 @@
                 },
                 ConnectionMode = connMode
             };
-
             // </EnableDistributedTracing>
+
             using (CosmosClient client = new CosmosClient(endpoint, authKey, options))
             {
                 Console.WriteLine($"Getting container reference for {containerName}.");
--- a/Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
+++ b/Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
@@ -15,7 +15,7 @@
         private static readonly string databaseName = "samples";
         private static readonly string containerName = "ai-sample";
 
-        private static TelemetryClient _telemetryClient;
+        private static TelemetryClient telemetryClient;
 
         static async Task Main()
         {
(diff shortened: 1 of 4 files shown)


Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
```diff
@@ -52,21 +51,17 @@ static async Task Main()
...
                 // </SetUpApplicationInsights>
 
-                CosmosClientOptions options = new CosmosClientOptions()
-                {
-                    IsDistributedTracingEnabled = true // Defaults to true, set to false to disable
-                };
-                using (CosmosClient client = new CosmosClient(endpoint, authKey, options))
-                {
-                    Console.WriteLine($"Getting container reference for {containerName}.");
+                var infoOperation = _telemetryClient.StartOperation<DependencyTelemetry>(".Net SDK (v3.37.0) : ApplicationInsights SDK (v2.22.0)");
```
ealsur, line 54: nit: I think you can use the Assembly object to extract the versions of the dependencies, otherwise you need to update this text everytime the package is bumped

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
```diff
@@ -52,21 +51,17 @@ static async Task Main()
...
 
-                    ContainerProperties properties = new ContainerProperties(containerName, partitionKeyPath: "/id");
+                var gops = _telemetryClient.StartOperation<DependencyTelemetry>("GATEWAY MODE");
```
ealsur, line 56: Do we need to add comments here explaining why are we running things twice and what each execution means?

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
```diff
@@ -78,6 +73,31 @@ static async Task Main()
...
+            CosmosClientOptions options = new CosmosClientOptions()
+            {
+                CosmosClientTelemetryOptions = new CosmosClientTelemetryOptions()
+                {
+                    DisableDistributedTracing = false
+                },
+                ConnectionMode = connMode
+            };
+
+            // </EnableDistributedTracing>
```
ealsur, line 87: This is breaking documentation. The `// </EnableDistributedTracing>` is supposed to match a `// <EnableDistributedTracing>` and all the code within them is what is rendered in the MsDocs article that consumes this file. 
Maybe the `// <EnableDistributedTracing>` needs to be moved to this method and before the line that starts the options declaration?

Microsoft.Azure.Cosmos.Samples/Usage/OpenTelemetry/Program.cs
```diff
@@ -104,6 +106,31 @@ static async Task Main()
...
+            CosmosClientOptions options = new CosmosClientOptions()
+            {
+                CosmosClientTelemetryOptions = new CosmosClientTelemetryOptions()
+                {
+                    DisableDistributedTracing = false
+                },
+                ConnectionMode = connMode
+            };
+
+            // </EnableDistributedTracing>
```
ealsur, line 120: Same here, this cannot be duplicated in the document, there cannot be 2 of these.

ealsur (changes requested):
Please fix the `// </EnableDistributedTracing>` sections. There can only be one of these and it needs to correctly wrap the piece of code that should pop on the MsDoc

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
kirankumarkolli: Is this how we expect the CX to use as well?

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
kirankumarkolli: Usage section is for application developers to use.

If "GATEWAY MODE" is equivalent to application span then lets please call it out 
Usage sample for common scenarios like WebApps (REST API's) might be one alternative.

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
sourabh1007: @ealsur Thanks for pointing it out, I will remove this version number. We can have this sample without this information also.

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
sourabh1007: Yes, I will put the comment around it. "it is application-level trace and exist in this test just for clarity."

ealsur (approved):

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
```diff
@@ -52,21 +51,17 @@ static async Task Main()
                 _telemetryClient = serviceProvider.GetRequiredService<TelemetryClient>();
```
philipthomas-MSFT, line 51: we are back to putting underscores for local variables? not criticizing, just asking a general question.

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
```diff
@@ -52,21 +51,17 @@ static async Task Main()
...
-                {
-                    IsDistributedTracingEnabled = true // Defaults to true, set to false to disable
-                };
-                using (CosmosClient client = new CosmosClient(endpoint, authKey, options))
-                {
-                    Console.WriteLine($"Getting container reference for {containerName}.");
+                var infoOperation = _telemetryClient.StartOperation<DependencyTelemetry>(".Net SDK : ApplicationInsights SDK"); // Application level activity to track the entire execution of the application
 
-                    ContainerProperties properties = new ContainerProperties(containerName, partitionKeyPath: "/id");
+                var gops = _telemetryClient.StartOperation<DependencyTelemetry>("GATEWAY MODE"); // Activity to track the execution of the gateway mode
```
philipthomas-MSFT, line 56: Should "GATEWAY MODE" and "DIRECT MORE" be enum instead of literal string?

philipthomas-MSFT (approved):
Approving but see my non-blocking comments.

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
sourabh1007: These are just text...nothing to do with enum

Microsoft.Azure.Cosmos.Samples/Usage/ApplicationInsights/Program.cs
sourabh1007: fixed.

philipthomas-MSFT (approved):

ealsur (approved):

PR #4220, opened 2023-12-15
Change Feed Processor: Fixes disposal of unused CancellationTokenSource

### LINKED ISSUE #4208: PartitionSupervisorCore.RunAsync leaks System.Threading.CancellationTokenSource+Linked1CancellationTokenSource

**Describe the bug**
Instances of `System.Threading.CancellationTokenSource+Linked1CancellationTokenSource` are leaked by the `PartitionSupervisorCore.RunAsync` method via calls to `CancellationTokenSource.CreateLinkedTokenSource` which are stored in the `processorCancellation` member.
The caller, `PartitionControllerCore.ProcessPartitionAsync`, has a catch clause that handles `FeedRangeGoneException`, and calls `PartitionControllerCore.HandlePartitionGoneAsync`, which eventually loops back around to `PartitionControllerCore.ProcessPartitionAsync` again, which calls `PartitionSupervisorCore.RunAsync`, when in turn allocates another `CancellationTokenSource`, but _without cleaning up the previous instance_.

**To Reproduce**
I don't know how our application got into this state, but by the time we captured a dump, there were 6.7 million instances of 
 `System.Threading.CancellationTokenSource+Linked1CancellationTokenSource` allocated along the call stack mentioned above.
I suspect the leak was a side effect of some kind of infinite loop from `FeedRangeGoneException` being thrown over and over again, but I have no idea why this happened or how to reproduce it.

**Expected behavior**
The SDK doesn't leak memory.

**Actual behavior**
The SDK leaks `System.Threading.CancellationTokenSource+Linked1CancellationTokenSource` instances in the extreme.

**Environment summary**
SDK Version: 3.32 (I know this isn't the latest but there is no change to memory management that I can see in the most recent code).
Windows Server 2022
net 7.0.304

**Additional context**
image

ealsur:
Once a lease is finished processing, due to split or merge happening or any of the known conditions:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/ea269c99b54a2ac2a4394edbcb50b2ed0cb6c0b0/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionSupervisorCore.cs#L52-L80

The PartitionSupervisor variable is no longer used and goes out of scope to be GCed.

The PartitionSupervisor implements `IDisposable` but there is no path that explicitly calls Dispose when the supervisor is no longer used.

Closes #4208

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionControllerCore.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionControllerCore.cs
@@ -89,8 +89,7 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.FeedManagement
                 throw;
             }
 
-            PartitionSupervisor supervisor = this.partitionSupervisorFactory.Create(lease);
-            this.ProcessPartitionAsync(supervisor, lease).LogException();
+            this.ProcessPartitionAsync(lease).LogException();
         }
 
         public override async Task ShutdownAsync()
@@ -146,8 +145,10 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.FeedManagement
             }
         }
 
-        private async Task ProcessPartitionAsync(PartitionSupervisor partitionSupervisor, DocumentServiceLease lease)
+        private async Task ProcessPartitionAsync(DocumentServiceLease lease)
         {
+            using PartitionSupervisor partitionSupervisor = this.partitionSupervisorFactory.Create(lease);
+
             try
             {
                 await partitionSupervisor.RunAsync(this.shutdownCts.Token).ConfigureAwait(false);
@@ -167,8 +168,6 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.FeedManagement
             }
 
             await this.RemoveLeaseAsync(lease: lease, wasAcquired: true).ConfigureAwait(false);
-
-            partitionSupervisor.Dispose();
         }
 
         private async Task HandlePartitionGoneAsync(DocumentServiceLease lease, string lastContinuationToken)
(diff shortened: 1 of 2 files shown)


kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionControllerCore.cs
```diff
@@ -167,6 +167,8 @@ private async Task ProcessPartitionAsync(PartitionSupervisor partitionSupervisor
             }
 
             await this.RemoveLeaseAsync(lease: lease, wasAcquired: true).ConfigureAwait(false);
+
+            partitionSupervisor.Dispose();
```
Pilchie, line 171: I think it would be clearer to add a `using` to the place where `partitionSupervisor` is created in `AddOrUpdateLeaseAsync`.  That puts the creation/dispose in the same place which makes it easier to understand the lifetime.

ealsur: The problem is the place where the supervisor is created uses a Fire & Forget pattern, if I put the using there, the supervisor would be disposed while the background Task runs.

Pilchie: Ah, fair point.  I'd suggest a comment in each place describing why the separation then.

ealsur: @Pilchie ended up refactoring to move the creation of the instance inside the method itself and that allowed the use of `using`

Pilchie (approved):

philipthomas-MSFT (approved):

FabianMeiswinkel (approved):
LGTM

PR #4226, opened 2023-12-23
[Internal] GlobalEndpointManager: Fixes Memory Leak

NaluTripician:
# Pull Request Template

## Description

This will fix the memory leak in `GlobalEndpointManager` where `CancellationTokenSources` are not disposed. 

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
@@ -351,14 +351,10 @@ namespace Microsoft.Azure.Cosmos.Routing
 
             public void Dispose()
             {
-                lock (this)
+                if (Interlocked.Increment(ref this.disposeCounter) == 1)
                 {
-                    if (this.disposeCounter == 0)
-                    {
-                        Interlocked.Increment(ref this.disposeCounter);
-                        this.CancellationTokenSource?.Cancel();
-                        this.CancellationTokenSource?.Dispose();
-                    }
+                    this.CancellationTokenSource?.Cancel();
+                    this.CancellationTokenSource?.Dispose();
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


FabianMeiswinkel (changes requested):
PLease take a look at the thread-safety comment

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -345,7 +347,20 @@ private static bool IsNonRetriableException(Exception exception)
                 }
 
                 return false;
-            }
+            }
+
+            public void Dispose()
+            {
+                lock (this)
```
FabianMeiswinkel, line 354: You do not need any locks when using Interlocked.Increment - Interlocked.Increment will atomically increment the number and return its updated value. So, you can safely use

```
if (Interlocked.Increment(ref this.disposeCounter) == 1)
{
    this.CancellationTokenSource?.Cancel();
    this.CancellationTokenSource?.Dispose();
}
```

An alternative using locking would usually take advantage of the double-checked-locking pattern (allowing to avoid perf overhead of taking the lock if not necessary) - so, something like...

```
if (this.disposeCounter > 0)
{
    // Exit early - another thread is taking care of clean-up already
    return;
}

lock (this) // see comment below - in c# never lock (this) - or at least be super careful
{
    if (this.disposeCounter > 0)
    {
        // Exit early - another thread is taking care of clean-up already
        return;
    }

    this.disposeCounter += 1;
    this.CancellationTokenSource?.Cancel();
    this.CancellationTokenSource?.Dispose();
}
```

Locks in C# should never ever be made on lock(this) for public /protected classes, because this could easily result in dead-locks (you simply don't know who else also locks (this). For a private sub-class like in this case it is in theory acceptable - I would still always declare a `private readonly object thisLock = new object();` to use it for locking just to avoid copy-and-paste mistakes.

All of that said - for this use-case my preference would be for the Interlocked.Increment pattern above - it avoids locks altogether and we don't need locks for consistency reasons here.

FabianMeiswinkel (approved):
LGTM now - thanks!

kundadebdatta (approved):
LGTM! Thanks!

PR #4229, opened 2023-12-27
GatewayClientStore: Fixes an issue with dealing with invalid JSON HTTP responses

### LINKED ISSUE #4162: Handle invalid JSON HTTP response from Gateway for error scenarios
We have received multiple reports of cases where the SDK throws a JSON serialization exception when processing HTTP responses from the Cosmos DB Gateway, for example:

```
Exception: Newtonsoft.Json.JsonReaderException: Error reading JObject from JsonReader. Path '', line 0, position 0.   
  at Newtonsoft.Json.Linq.JObject.Load(JsonReader reader, JsonLoadSettings settings)    
  at Microsoft.Azure.Documents.JsonSerializable.LoadFrom(JsonReader reader, JsonSerializerSettings serializerSettings)    
  at Microsoft.Azure.Documents.JsonSerializable.LoadFrom[T](JsonTextReader jsonReader, ITypeResolver`1 typeResolver, JsonSerializerSettings settings)    
  at Microsoft.Azure.Cosmos.GatewayStoreClient.<CreateDocumentClientExceptionAsync>d__11.MoveNext() --- End of stack trace from previous location where exception was thrown ---    
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()    
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)    
  at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()    
  at Microsoft.Azure.Cosmos.GatewayStoreClient.<ParseResponseAsync>d__9.MoveNext()
```

`Microsoft.Azure.Cosmos.GatewayStoreClient.<ParseResponseAsync>` hints as this being an HTTP response from Gateway.

`Microsoft.Azure.Cosmos.GatewayStoreClient.<CreateDocumentClientExceptionAsync>` means the statuscode of the HTTP response was a failure (not 200 or any success).

Which means, Gateway returned a failure but the content was invalid.

SDK checks the `Content-Type` header and if it's `application/json` then expects the content to be valid JSON:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/083df87e6ebf8cfc25092e8642846f34273158f1/Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs#L151-L154

In all these cases what happened was that the content was either empty or invalid JSON.

To avoid the exception in the most common case (empty), we should check the Content-Size of the response, and if it's 0, either throw an exception saying that the response is invalid because it's empty or avoid attempting the JSON deserialization of the body and just produce the error response using the statuscode as if it had no content.

## Expected outcome

* If the Content-Size header is `application/json` but the content size is 0, skip attempting to deserialize the content
* If the Content-Size header is `application/json` but the content size > 0, attempt to deserialize the content. If that fails, follow the code path that treats the content as text (non-JSON path).

philipthomas-MSFT:
## Description

Fixes an issue when `Error` string/stream from Gateway response cannot be deserialized.
For more information, see #4162 

## Type of change

## Closing issues

To automatically close an issue: closes #4162


Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -150,6 +151,24 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
             // If service rejects the initial payload like header is to large it will return an HTML error instead of JSON.
             if (string.Equals(responseMessage.Content?.Headers?.ContentType?.MediaType, "application/json", StringComparison.OrdinalIgnoreCase))
             {
+                // For more information, see https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4162.
+                if (await GatewayStoreClient.IsJsonHTTPResponseFromGatewayInvalidAsync(responseMessage))
```
sourabh1007, line 155: Here you are checking if content is not there (means content length is 0) then create your own Error object with hardcoded message.
What about other scenario when content is there but we got a response from gateway which is not de-serializable to `Error` type?

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -150,6 +151,24 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
+                {
+                    return new DocumentClientException(
+                        new Error
+                        {
+                            Code = responseMessage.StatusCode.ToString(),
+                            Message = "Invalid JSON HTTP response from Gateway."
```
sourabh1007, line 161: Message should not be "Invalid" JSON.  It should be something like "No response content from gateway".

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -197,6 +216,19 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
+        /// <summary>
+        /// Checking if exception response (deserializable Error object) is valid based on the content length.
+        /// For more information, see <see href="https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4162."/>.
+        /// </summary>
+        /// <param name="responseMessage"></param>
+        private static async Task<bool> IsJsonHTTPResponseFromGatewayInvalidAsync(HttpResponseMessage responseMessage)
+        {
+            string readString = await responseMessage.Content.ReadAsStringAsync();
+
+            return responseMessage.Content?.Headers?.ContentLength == 0 ||
```
sourabh1007, line 228: I think we can rely on `ContentLength` only to identify if content is there. No need to read the message and then get the length.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: > What about other scenario when content is there but we got a response from gateway which is not de-serializable to `Error` type?

Is this a real case? Or are you suggesting future proofing? If that scenario is valid, would the content type be "application/json" or "text/html"?

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: Sure, thanks for the feedback.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: > I think we can rely on `ContentLength` only to identify if content is there. No need to read the message and then get the length.

if you look at the tests, I have a case where the response could have whitespaces. whitespaces without trimming will not have a content length of zero. so you must read it, trim it, and then check the length.

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/bee917ff59f2536216125280fc8d0ca1884028fe/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs#L30C9-L30C111

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
FabianMeiswinkel: Both are possible - we had bugs in ComputeGateway where the json returned was invalid (not even valid json) but using application/json - and RoutingGateway is using IIS - which can return test/html for every request when some validations (for example headers too large) are hit

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
FabianMeiswinkel: Also Content-Length from service is a SHOULD - not a MUST - so, strictly the header is optional. https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length
So, while header can be used in most cases it is not 100% fool-proof (I would be ok taking the dependency here given we know the endpoints would be compliant with the SHOULD rule) - your choice.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: Then reading the response content is more important in order try parse the Json. Another conversation below was suggesting to not read the content, just the content length, and i did push back on this, but wanted to make sure if this is still a concern, let's put those concerns to bed and I can resolve the other conversation and make this change.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -150,6 +151,24 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
+                        },
+                        responseMessage.Headers,
+                        responseMessage.StatusCode)
+                    {
+                        StatusDescription = responseMessage.ReasonPhrase,
+                        ResourceAddress = resourceIdOrFullName,
+                        RequestStatistics = requestStatistics
+                    };
+                }
+
```
ealsur, line 171: This method does 2 things:

1. If the Content Type is `application/json`, attempts to parse as JSON
2. If not (the `else`) does the `ReadAsStringAsync`.

As Fabian mentioned, there are 2 cases:

1. Content-type is `application/json` but Content Length is 0 (this is the evidence we had from customers, the Content Length header is present and is 0).
2. Content-type is `application/json`, Content Length is greater than 0, but the content is not JSON

In both cases, the `Documents.Resource.LoadFrom` method throws.

If we are introducing some logic that duplicates what the `else` clause does, could we then refactor the code in a way to reduce code duplication?

ealsur: To clarify: The new code that does `return new DocumentClientException` is very similar to what the `else` clause does, could we take advantage of what already exists instead of duplicating it?

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -150,8 +152,23 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
-                Error error = Documents.Resource.LoadFrom<Error>(readStream);
+                // For more information, see https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4162.
+                Error error;
+
+                if (await GatewayStoreClient.IsJsonHTTPResponseFromGatewayInvalidAsync(responseMessage))
+                {
+                    error = new Error
+                    {
+                        Code = responseMessage.StatusCode.ToString(),
+                        Message = "No response content from gateway."
```
ealsur, line 163: How is this approach include the raw non-JSON string in the DocumentClientException?

Assuming there are 2 code blocks, A (what is in the `if`), and B (what is in the `else`), would an approach like this work?

```
if (string.Equals(responseMessage.Content?.Headers?.ContentType?.MediaType, "application/json", StringComparison.OrdinalIgnoreCase))
{
   try
   {
       CodeBlockA
   }
}
// no else

CodeBlock-B
```

Removing the `else` and adding a `try/catch` would simplify and allow to reuse the current code. In the case of an invalid JSON, the exception on the `catch` would flow to CodeBlock-B which treats the body as raw string.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs
```diff
@@ -0,0 +1,106 @@
...
+                startTime: DateTime.UtcNow, 
+                trace: NoOpTrace.Singleton);
+
+            DocumentClientException documentClientException = await GatewayStoreClient.CreateDocumentClientExceptionAsync(
+                responseMessage: responseMessage,
+                requestStatistics: requestStatistics);
+
+            Assert.IsNotNull(value: documentClientException);
+            Assert.AreEqual(expected: HttpStatusCode.NotFound, actual: documentClientException.StatusCode);
+            Assert.IsTrue(condition: documentClientException.Message.Contains("No response content from gateway."));
```
ealsur, line 61: The message should contain the `content`, whatever it is that Gateway returned that is not JSON

philipthomas-MSFT: Ok, thanks for the feedback @ealsur! However, at what point did we decide as a requirement that it is correct to return "whatever it is that Gateway returned that is not JSON"?

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: And also @ealsur , your proposal at refactoring is for the existing code that is not being touched by this fix. I falsely assumed you were suggesting the refactoring of the code that I included to address the problem. I am all about making improvements to parts of the codebase, but what are we gaining? Do we have sufficient enough tests to ensure that the behavior does not change, and that we are not introducing new issues? I mean, I did a search for `CreateDocumentClientExceptionAsync` before I started and did not see one unit test for this method. So, it makes me afraid to touch it until we can understand it more, and put more tests around it. Again, thanks for the feedback. I welcome these discussions.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs
```diff
@@ -0,0 +1,106 @@
...
+    [TestClass]
+    public class GatewayStoreClientTests
+    {
+        /// <summary>
+        /// Testing the exception behavior when a response from the Gateway has no response (deserializable Error object) based on the content length.
+        /// For more information, see <see href="https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4162"/>.
+        /// </summary>
+        /// <returns></returns>
+        [TestMethod]
+        [DataRow(@"")]
```
ealsur, line 28: Add a test case for valid JSON that represents an `Error` type

ealsur: Can be another test method

philipthomas-MSFT: That test already exist @ealsur . `CreateDocumentClientExceptionValidJsonResponseFromGatewayTestAsync`

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs
ealsur: If the Gateway response has content, we should return it in the error. It could be that they erroneously set the Content-Type to json, but if there is content, it can be valuable for debugging purposes. We cannot discard content just because it is not in the format we expected based on the headers, it could make the difference between being able to continue an incident investigation and having no way to know what was returned to the client

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -132,26 +134,66 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
-            // If service rejects the initial payload like header is to large it will return an HTML error instead of JSON.
-            if (string.Equals(responseMessage.Content?.Headers?.ContentType?.MediaType, "application/json", StringComparison.OrdinalIgnoreCase))
+            try
             {
                 Stream readStream = await responseMessage.Content.ReadAsStreamAsync();
                 Error error = Documents.Resource.LoadFrom<Error>(readStream);
+
+                // need to rethink dropping the check for media type "application/json".
+
+                if (responseMessage.Content?.Headers?.ContentLength == 0 ||
```
ealsur, line 187: If the ContentLength is 0, this never hits because the LoadFrom throws.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs
```diff
@@ -0,0 +1,226 @@
...
+            Assert.IsNotNull(value: documentClientException.Error);
+            Assert.AreEqual(expected: HttpStatusCode.NotFound.ToString(), actual: documentClientException.Error.Code);
+            Assert.IsTrue(documentClientException.Error.Message.Contains(contentMessage));
+        }
+
+        /// <summary>
+        /// Testing CreateDocumentClientExceptionAsync when media type is application/json and the header content length is zero.
+        /// </summary>
+        [TestMethod]
+        public async Task TestCreateDocumentClientExceptionWhenMediaTypeIsApplicationJsonAndHeaderContentLengthIsZeroAsync()
```
ealsur, line 64: nit: This is not testing ContentLength 0, it is testing a JSON that has an Error, with property Code = 404 and Message = empty

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs
philipthomas-MSFT: Thanks for the feedback @ealsur . I was planning on doing more work on this today. I will mark this as draft. I have a habit of pushing code frequently.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: This is false, this test [TestCreateDocumentClientExceptionWhenMediaTypeIsApplicationJsonAndContentIsNotValidJsonAndContentLengthIsZeroAsync](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/2f582f648cd240ba73ce6834c1603400f5c7e975/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs#L146C27-L146C141).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs
philipthomas-MSFT: OK. Out of draft mode. Please provide feedback. Thanks.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: OK. Out of draft mode. Please provide feedback. Thanks.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreClientTests.cs
philipthomas-MSFT: OK. Out of draft mode. Please provide feedback. Thanks.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -132,26 +137,196 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
+
+            return error;
+        }
+
+        /// <summary>
+        /// Creating a new DocumentClientException using the Gateway response message.
+        /// </summary>
+        /// <param name="responseMessage"></param>
+        /// <param name="requestStatistics"></param>
+        internal static async Task<DocumentClientException> WorkingCreateDocumentClientExceptionAsync(
```
ealsur, line 271: Is this a test code?

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -132,26 +137,196 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
             }
 
-            // If service rejects the initial payload like header is to large it will return an HTML error instead of JSON.
-            if (string.Equals(responseMessage.Content?.Headers?.ContentType?.MediaType, "application/json", StringComparison.OrdinalIgnoreCase))
+            try
             {
-                Stream readStream = await responseMessage.Content.ReadAsStreamAsync();
-                Error error = Documents.Resource.LoadFrom<Error>(readStream);
+                Stream contentAsStream = await responseMessage.Content.ReadAsStreamAsync();
+                Error error = GatewayStoreClient.GetError(
```
ealsur, line 172: This changes the flow into always attempting to deserialize to JSON regardless of the Content-Type.

Given that we have no need to introduce a new custom text that says `No response content from gateway.` we don't need to specifically detect the case of empty content.

Why not simplify the flow by:

1. Maintain the Content-Type check, add a Content-Length > 0 condition to it.
2. Remove the `else`, treat it as a flow-through
3. Add a `try` with empty `catch` inside the `if` when `LoadFrom` is attempted

```
if (condition)
{
  try
  {
      json parsing
  }
  catch{}
}

string parsing
```

That way we don't add layers of complexity checking and trimming the text.

The goal of this PR should be: Avoid crashing the customer application with a JSON exception while allowing us to capture any non-json content in the Gateway response

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -9,6 +9,9 @@ namespace Microsoft.Azure.Cosmos
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
     using System.IO;
+    using System.Linq;
+    using System.Linq.Expressions;
+    using System.Net;
```
ealsur, line 14: Are all these usings needed?

ealsur (commented):
Great progress! Thanks for doing this

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: That the old one. Removed

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: I understand what your vision for this. However, the problem with this approach is that I envision this to deal with cases where the Error object is valid JSON, but its message is empty, which is why I trimming the message of Error to also deal with potential whitespaces. What you suggest takes this away from what I envisioned it to behave. Just checking the ContentLength only handles cases when Error is empty. Do you think that this is not necessary? Because if I don't include this, the behavior will send the Error back in the DocumentClientException, but the Error will not have a message. If this is perfectly fine and expected from the CX perspective. I can go down your path and call it a day.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: I will remove them on the next push. Thanks.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
ealsur: I understand better now, thanks for detailing what you intended :)

Help me understand a bit more: If the Error object is valid JSON, but the Message property is empty (the ErrorCode is not), then can you expand on why is this a problem? The ErrorCode property is still there and has a value. What is the value we would be adding by removing the Error object content (including the ErrorCode value) for "No response content from Gateway"? What is the latter adding?

From the CX perspective, they are just expecting their app not to crash, the Error object is something that they do not consume specifically, it is for whoever is looking at the DocumentClientException from the Cosmos DB team to figure out what is wrong within Gateway.

philipthomas-MSFT: Thanks for the feedback. I agree with what you are saying. For me it just looks weird to return an Error object with no details on what that error is. Maybe there is no real case that the Error object will have an empty message. Maybe this is just made up in my head. 😂 But we never thought that the Gateway could send an error that is empty. For something like a malformed JSON, like HTML, at least we include that so that the CX can see what is happening and take some action, or no action, but at least they can make a decision on what they see. But acknowledging that there is an Error with no message at all almost seems as worse as failing out ungracefully. 

At the end of the day, I think we all can agree that it would be nice if we didn't have to circumvent this and actually have Gateway explain and fix this themselves. Since that is not going to happen, and this is the best solution, I will allow it without further pushback. 

I already removed it and made the recommended change. Let's just get this pushed through. In the end, I enjoy the conversation around this so let's take it offline.

ealsur: > At the end of the day, I think we all can agree that it would be nice if we didn't have to circumvent this and actually have Gateway explain and fix this themselves.

💯

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -132,69 +131,79 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
+                resourcePath: out _,
+                resourceIdOrFullName: out string resourceIdOrFullName,
+                isNameBased: out _))
             {
                 // if resourceLink is invalid - we will not set resourceAddress in exception.
             }
 
             // If service rejects the initial payload like header is to large it will return an HTML error instead of JSON.
-            if (string.Equals(responseMessage.Content?.Headers?.ContentType?.MediaType, "application/json", StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(responseMessage.Content?.Headers?.ContentType?.MediaType, "application/json", StringComparison.OrdinalIgnoreCase) ||
```
ealsur, line 154: Condition should be `&&`, for attempting JSON parsing the header should be `application/json` AND the content should be > 0

```suggestion
            if (string.Equals(responseMessage.Content?.Headers?.ContentType?.MediaType, "application/json", StringComparison.OrdinalIgnoreCase) &&
```

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -132,69 +131,79 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
+                responseMessage.Content?.Headers.ContentLength != 0)
```
ealsur, line 155: I don't know if the value could be -1 but better be safe :D

```suggestion
                responseMessage.Content?.Headers.ContentLength > 0)
```

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -132,69 +131,79 @@ internal static INameValueCollection ExtractResponseHeaders(HttpResponseMessage
...
                     {
-                        foreach (KeyValuePair<string, IEnumerable<string>> header in requestMessage.Headers)
-                        {
-                            context.AppendLine($"Header: {header.Key} Length: {string.Join(",", header.Value).Length};");
-                        }
+                        contextBuilder.AppendLine($"Header: {header.Key} Length: {string.Join(",", header.Value).Length};");
                     }
                 }
-
-                String message = await responseMessage.Content.ReadAsStringAsync();
```
ealsur, line 185: Nice, you removed a double read we had here!

ealsur (approved):
LGTM, just a few small nits

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
philipthomas-MSFT: Fair point

ealsur (approved):

sourabh1007 (approved):

PR #4251, opened 2024-01-15
IgnoreCertFlag: Adds flag support to ignore SSL check

sourabh1007:
# Pull Request Template

## Description

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -236,6 +242,16 @@ internal virtual CosmosClient Build(DocumentClient documentClient)
             return new CosmosClient(this.accountEndpoint, this.accountKey, this.clientOptions, documentClient);
         }
 
+        /// <summary>
+        /// To ignore SSL certificate validation for the service endpoint.It is NOT recommended to use this flag in production
+        /// </summary>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
+        public CosmosClientBuilder IgnoreCertificateCheck()
+        {
+            this.clientOptions.ServerCertificateCustomValidationCallback = (_, _, _) => true;
```
Pilchie, line 251: Does this work with `Direct` connections?  The doc page says you also need to limit to `Gateway` connections for this to work?

sourabh1007: Today customers can ignore certificates in 2 ways:
1. Just for HTTP calls: create custom httpclient with custom callback
2. For HTTP and TCP calls: Set `ServerCertificateCustomValidationCallback` in `CosmosClientOptions` with custom callback.

My understanding is: Now, we are adding a capability where in connection string itself we would pass a flag which will tell SDK if certificate validation should perform or not. It is equivalent to what we do in 2nd option above.
Is there any scenario where customer would want to ignore cert check only for HTTP call when working with emulator.

cc @kirankumarkolli

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -50,7 +50,8 @@ public class CosmosClientOptions
         private const Protocol DefaultProtocol = Protocol.Tcp;
 
         private const string ConnectionStringAccountEndpoint = "AccountEndpoint";
-        private const string ConnectionStringAccountKey = "AccountKey";
+        private const string ConnectionStringAccountKey = "AccountKey";
+        private const string IgnoreEndpointCertificate = "IgnoreEndpointCertificate";
```
kirankumarkolli, line 54: Let's close this with emulator team on naming.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
kirankumarkolli: It works for both. @aavasthy added this capability.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -236,6 +242,16 @@ internal virtual CosmosClient Build(DocumentClient documentClient)
             return new CosmosClient(this.accountEndpoint, this.accountKey, this.clientOptions, documentClient);
         }
 
+        /// <summary>
+        /// To ignore SSL certificate validation for the service endpoint.It is NOT recommended to use this flag in production
+        /// </summary>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
+        public CosmosClientBuilder IgnoreCertificateCheck()
```
kirankumarkolli, line 249: Is it expected to be public?

sourabh1007: I made it public intentionally to provide a way to ignore certificates if somebody is working with builder pattern and want to use it...got inspired from this https://github.com/dotnet/aspire/pull/1668#event-11508462666
But I just gave another thought, there are high chances that somebody would leave it as it is in production code also which is not recommended. So with this thought, I will make it internal.

sourabh1007: I have removed it.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -845,15 +846,31 @@ internal virtual ConnectionPolicy GetConnectionPolicy(int clientId)
...
+        {
+            return Convert.ToBoolean(CosmosClientOptions.GetValueFromConnectionString<bool>(connectionString, CosmosClientOptions.IgnoreEndpointCertificate, false));
+        }
+
+        internal static CosmosClientOptions GetCosmosClientOptionsWithCertificateFlag(string connectionString, CosmosClientOptions clientOptions)
+        {
+            clientOptions ??= new CosmosClientOptions();
+            if (CosmosClientOptions.IsIgnoreEndpointCertificateFlag(connectionString))
+            {
+                clientOptions.ServerCertificateCustomValidationCallback = (_, _, _) => true;
```
ealsur, line 867: One caveat here: This works but if the customer customizes the HttpClientFactory (CosmosClientOptions.HttpClientFactory), it will not. Maybe we should validate that both are not set? (there are several "ValidateXXXX" methods in this class that are similar). Otherwise if a user uses the custom connection string from the emulator and also sets the HttpClientFactory, they will still have the SSL issue (unless they are disabling SSL validation from the HttpClientFactory (https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-develop-emulator?tabs=windows%2Ccsharp&pivots=api-nosql#connect-to-the-emulator-from-the-sdk)

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: Just for context: goal is to update emulator data-explore to show connection string with certValdiation turned-off.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: 2 things:
1. We should support updating current instance of custom `httpclient` with server callback function. I would need little bit reflection code (below .net6 we cannot get handler instance from `httpclient`). I think that will be better.
2. Currently `ServerCertificateCustomValidationCallback` in `clientoptions`, doesn't have this validation. I think we should have property information about it in the documentation.

sourabh1007: Added validation and documentation

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -62,7 +62,13 @@ internal sealed class CosmosHttpClientCore : CosmosHttpClient
                     throw new InvalidOperationException($"{nameof(connectionPolicy.HttpClientFactory)} can not be set at the same time as {nameof(sendingRequestEventArgs)} or {nameof(ReceivedResponseEventArgs)}");
                 }
 
+                if (connectionPolicy.ServerCertificateCustomValidationCallback != null)
```
ealsur, line 65: Please move this to the other Validate methods in CosmosClientOptions: 

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/431c6f224747e463bed0475083d5e0d7c1afaee2/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs#L753-L757

sourabh1007: I got inspired from above validation check

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
```diff
@@ -885,7 +884,24 @@ public void InvalidApplicationNameCatchTest()
                     ApplicationName = illegal
                 });
             }
-        }
+        }
```
ealsur, line 887: Add a test that validates that if the HttpClientFactory is set along with the ServerValidationCallback, it throws.

Similar to other tests in this file, like:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/431c6f224747e463bed0475083d5e0d7c1afaee2/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs#L589-L595

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -889,8 +919,16 @@ private void ValidateLimitToEndpointSettings()
...
             }
-        }
-
+        }
+
+        private void ValidateIgnoreCertFlagAndHttpFactorySettings()
+        {
+            if (this.ServerCertificateCustomValidationCallback != null && this.HttpClientFactory != null)
+            {
+                throw new ArgumentException($"{nameof(this.ServerCertificateCustomValidationCallback)} or {IgnoreEndpointCertificate} flag can not be set along with {nameof(this.HttpClientFactory)}");
```
ealsur, line 928:
```suggestion
                throw new ArgumentException($"{nameof(this.ServerCertificateCustomValidationCallback)} or IgnoreEndpointCertificate flag in the connection string can not be set along with {nameof(this.HttpClientFactory)}");

```

ealsur (dismissed):
Please verify the Connection String format is ok with the emulator team

Microsoft.Azure.Cosmos/src/CosmosClient.cs
```diff
@@ -184,6 +184,7 @@ protected CosmosClient()
         /// </example>
         /// <remarks>
         /// The returned reference doesn't guarantee credentials or connectivity validations because creation doesn't make any network calls.
+        /// Alternatively, Connection string can have "IgnoreEndpointCertificate" flag to ignore SSL certificate validation for the service endpoint.It cannot be used with custom HttpClientFactory and NOT recommended to use this flag in production.
```
kirankumarkolli, line 187:
```suggestion
        /// Alternatively, Connection string can have "IgnoreSslCertificate" flag to ignore SSL certificate validation for the service endpoint. It cannot be used with custom HttpClientFactory and NOT recommended to use this flag in production.
```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
v1k1: I don't have a major issue with the name `IgnoreEndpointCertificate`. Though I would like to propose `DisableSSLCertificateValidation`.  Did a quick check in google, stackoverflow etc, got an impression `disable ssl` is usually how people search about it, but a soft suggestion still.

sourabh1007: @kirankumarkolli @ealsur @Pilchie What do you think about naming suggestion given by Vikas from Emulator team? I like his suggestion.

Pilchie: I like, it, but we should probably use TLS instead of SSL?

Microsoft.Azure.Cosmos/src/CosmosClient.cs
```diff
@@ -495,6 +496,9 @@ protected CosmosClient()
         /// ]]>
         /// </code>
         /// </example>
+        /// <remarks>
+        /// Alternatively, Connection string can have "IgnoreEndpointCertificate" flag to ignore SSL certificate validation for the service endpoint.It cannot be used with custom HttpClientFactory and NOT recommended to use this flag in production.
```
kirankumarkolli, line 500:
```suggestion
        /// With emulator to ignore sslCertificate please use connectionstring <TBD>
```

kirankumarkolli: Started offline chat to close on it

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -843,34 +844,62 @@ internal virtual ConnectionPolicy GetConnectionPolicy(int clientId)
...
+        {
+            return Convert.ToBoolean(CosmosClientOptions.GetValueFromConnectionString<bool>(connectionString, CosmosClientOptions.ConnectionStringIgnoreEndpointCertificate, false));
+        }
+
+        internal static CosmosClientOptions GetCosmosClientOptionsWithCertificateFlag(string connectionString, CosmosClientOptions clientOptions)
+        {
+            clientOptions ??= new CosmosClientOptions();
+            if (CosmosClientOptions.IsIgnoreEndpointCertificateFlag(connectionString))
+            {
+                clientOptions.ServerCertificateCustomValidationCallback = (_, _, _) => true;
```
kirankumarkolli, line 867: Also thoughts on collecting this info (either in diagnostics/telemetry). Can be a follow-up work item.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -507,6 +507,42 @@ public async Task Verify_CertificateCallBackGetsCalled_ForTCP_HTTP()
...
+            Assert.IsNotNull(cosmosClient.ClientOptions.ServerCertificateCustomValidationCallback);
+
+            Cosmos.Database database = null;
+            try
+            {
+                string databaseName = Guid.NewGuid().ToString();
+                string databaseId = Guid.NewGuid().ToString();
+
+                //HTTP callback
+                database = await cosmosClient.CreateDatabaseAsync(databaseId);
```
kirankumarkolli, line 532: In out test env we don't have any cert issues right, test needs to truly assert that cert is presented and ignored (do a test hook or trace collection)?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli: TLS is right term, unsure if most developers will be able to map to it easily. 

.NET has some similar API without TLS/SSL and used Server, something like [ dangerousacceptanyservercertificatevalidator](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.dangerousacceptanyservercertificatevalidator?view=net-8.0)

sourabh1007: yes thats what I was also looking for, i couldn't find SSL/TLS explicitly mentioned anywhere. I think we should also follow similar approach here...e.g. `DisableServerCertificateValidation`

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: What is the use case? which scenarios, it will help us debugging?

kirankumarkolli: Like if a CX using this unexpectedly for PROD

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: updated.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
sourabh1007: Covered in unit test. emulator test is not required.

Microsoft.Azure.Cosmos/src/CosmosClient.cs
```diff
@@ -184,6 +184,9 @@ protected CosmosClient()
         /// </example>
         /// <remarks>
         /// The returned reference doesn't guarantee credentials or connectivity validations because creation doesn't make any network calls.
+        /// With emulator to ignore SSL Certificate please use connectionstring with "DisableServerCertificateValidation" flag. 
+        /// This flag will be overrided for HTTP calls if custom HttpClientFactory is used.
+        /// It is NOT recommended to use this flag in production.
```
kirankumarkolli, line 189:
```suggestion
        /// 
        /// Emulator: To ignore SSL Certificate please suffix connectionstring with "DisableServerCertificateValidation=True;". 
        /// When CosmosClientOptions.HttpClientFactory is used, SSL certificate needs to be handled appropriately.
        /// NOTE: DO NOT use this flag in production (only for emulator)
```

kirankumarkolli: Code doc needs to be repeated in other places. 

Thoughts on including a public documentation reference and update it with full details?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
```diff
@@ -885,7 +884,24 @@ public void InvalidApplicationNameCatchTest()
...
+        [TestMethod]
+        [DataRow(ConnectionString, false)]
+        [DataRow(ConnectionString + "DisableServerCertificateValidation=true;", true)]
+        public void TestServerCertificatesValidationCallback(string connStr, bool expectedIgnoreCertificateFlag)
+        {
+            CosmosClient cosmosClient = new CosmosClient(connStr);
+
+            if (expectedIgnoreCertificateFlag)
+            {
+                Assert.IsNotNull(cosmosClient.ClientOptions.ServerCertificateCustomValidationCallback);
```
kirankumarkolli, line 898: FWIW the hypothesis is that servercallback is set to ignore. 
If this test is enhanded to valdiate that the `ServerCertificateCustomValidationCallback` indeed is ignoring un-trusted SSL is good enough. 

Test is [CosmosClientResourceUnitTests.cs](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4251/files#diff-f34e244c1b81234a3cd1580903d2667a4b502e71563cb6b39470ce58309691a4) feels too superficial and may be unnecessary.

Microsoft.Azure.Cosmos/src/CosmosClient.cs
sourabh1007: Updating it in public API is must we can have public explaining it with example, that would be better.

kirankumarkolli: Can you please expand with more context.
Is the proposal to inline the code samples inside the remarks?

sourabh1007: proposal is: 
1. Have proper text (without coding samples) in remarks section.
2. Make a public documentation available with code sample.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
sourabh1007: added assertion here itself.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -651,7 +652,9 @@ internal Protocol ConnectionProtocol
         /// </summary>
         /// <remarks>
         /// <para>
-        /// Customizing SSL verification is not recommended in production environments.
+        /// Emulator: To ignore SSL Certificate please suffix connectionstring with "DisableServerCertificateValidation=True;". 
+        /// When CosmosClientOptions.HttpClientFactory is used, SSL certificate needs to be handled appropriately.
```
kirankumarkolli, line 656: This is not applicable here right?

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
sourabh1007: its applicable here also...this new flag is setting this option only.

kirankumarkolli: There are two aspects here right?
- CX overriding `ServerCertificateCustomValidationCallback` and using `DisableServerCertificateValidation`. 
  - Option (1) Chain them so-that its an AND combination
  - Error out that only one of them be used
- HttpClientFactory and `ServerCertificateCustomValidationCallback`: This is an existing combination and already applicable any side-affects right? 
  - It seems this aspect might missing in the doc and needs to be explicitly covered but little differently

sourabh1007: No. 

1. `ServerCertificateCustomValidationCallback` and using `DisableServerCertificateValidation` flag, both are same things. 
If cx configure both the things, `ServerCertificateCustomValidationCallback` will be override by callback set due to the flag.
2. `HttpClientFactory` and `ServerCertificateCustomValidationCallback`: This is an existing combination and already applicable any side-affects . i.e. HTTP calls won't use callback set by the customer.

Both the things should be documented.

ealsur (approved):

PR #4252, opened 2024-01-15
Query : Adds Request Charge to Query Metrics

Maya-Painter:
## Description

Adds per-partition RequestCharge to `ServerSidePartitionedMetrics` and TotalRequestCharge to `ServerSideCumulativeMetrics`.

## Type of change


Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternalAccumulator.cs
```diff
@@ -142,5 +146,66 @@ public static void WalkTraceTreeForQueryMetrics(ITrace currentTrace, ServerSideM
 
             return null;
         }
+
+        private static double? WalkTraceTreeForRequestCharge(ITrace currentTrace)
```
adityasa, line 150: >WalkTraceTreeForRequestCharge [](http://example.com/codeflow?start=31&length=29)

Can we move "Walk" logic in a separate class that walks once and exposes both RequestCharge and PartitionKeyRangeId?

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSidePartitionedMetrics.cs
```diff
@@ -26,5 +26,10 @@ public abstract class ServerSidePartitionedMetrics
         /// Only has a value in direct mode. When using gateway mode, this is null.
         /// </remarks>
         public abstract int? PartitionKeyRangeId { get; }
+
+        /// <summary>
+        /// Gets the request charge for the operation.
```
adityasa, line 31: Nit:
Gets the request charge for the operation on this partition.

adityasa:
>     public abstract class ServerSidePartitionedMetrics

Not an ask from this customer, but should we update the cumulative metrics as well?

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSidePartitionedMetrics.cs:10 in 4aa0a48. [](commit_id = 4aa0a4887cd9a12c681d25ac0ff205c86dcc48c4, deletion_comment = False)

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternalAccumulator.cs
Maya-Painter: The trace datum type where we should look to find the RequestCharge will depend on the result of the PartitionKeyRangeId walk, so I'm not sure if this logic can be supported in a single pass.

Maya-Painter:
> > ```
> > public abstract class ServerSidePartitionedMetrics
> > ```
> 
> Not an ask from this customer, but should we update the cumulative metrics as well?
> 
> Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSidePartitionedMetrics.cs:10 in [4aa0a48](https://github.com/Azure/azure-cosmos-dotnet-v3/commit/4aa0a4887cd9a12c681d25ac0ff205c86dcc48c4). [](commit_id = [4aa0a48](https://github.com/Azure/azure-cosmos-dotnet-v3/commit/4aa0a4887cd9a12c681d25ac0ff205c86dcc48c4), deletion_comment = False)

Updated to add a new ServerSideCumulativeMetrics member TotalRequestCharge. I'm also fine with just leaving this out. Adding this to ServerSideMetrics instead it would make the abstraction less clean since we would have to add the request charge after the class has been constructed by parsing the query metrics diagnostics datum.

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternalAccumulator.cs
adityasa: The walk logic seems common in direct mode. The new class can fish out the partition key range ids and request charge at the same time. In direct mode this should be sufficient. If it couldn't find partition key range ids, it could defer to gateway mode logic.

Purpose of separate class is two-fold:
- it keeps the responsibility of walking the request stats separate
- makes the unit testable if/when we need (for e.g. regression/enhancing the logic for special cases - consider different diagnostic shape)

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsTraceExtractor.cs
```diff
@@ -0,0 +1,72 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Query.Core.Metrics
+{
+    using System;
+    using Microsoft.Azure.Cosmos.Tracing;
+    using Microsoft.Azure.Cosmos.Tracing.TraceData;
+
+    internal static class ServerSideMetricsTraceExtractor
```
Maya-Painter, line 11: @adityasa Open to other names, wasn't sure what an appropriate descriptor of the class would be

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsInternalAccumulator.cs
Maya-Painter: Created new static class for walk logic

Microsoft.Azure.Cosmos/src/Query/Core/Metrics/ServerSideMetricsTraceExtractor.cs
```diff
@@ -0,0 +1,72 @@
...
+                    accumulator.Accumulate(serverSideMetrics);
+                }
+            }
+
+            foreach (ITrace childTrace in currentTrace.Children)
+            {
+                ServerSideMetricsTraceExtractor.WalkTraceTreeForQueryMetrics(childTrace, accumulator);
+            }
+
+            return;
```
adityasa, line 36: >return; [](http://example.com/codeflow?start=12&length=7)

Nit, unnecessary.

kirankumarkolli (approved):

kirankumarkolli:
Signing off on contracts

sboshra (approved):
:shipit:

adityasa (approved):
:shipit:

PR #4262, opened 2024-01-18
PriorityBasedExecution: Adds PriorityLevel in CosmosClientOptions

Achint-Agrawal:
## Description

Adding PriorityLevel to CosmosClientOptions to allow users to set PriorityLevel at the client level

## Type of change


Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -278,6 +278,17 @@ public ConnectionMode ConnectionMode
         /// If this is not set the database account consistency level will be used for all requests.
         /// </summary>
         public ConsistencyLevel? ConsistencyLevel { get; set; }
+
+        /// <summary>
+        /// Sets the PriorityLevel for requests created using cosmos client.
+        /// If PriorityLevel is also specified at request level in RequestOptions, that PriorityLevel takes precedence.
```
ealsur, line 284: This should be within a `<remark>`.

ealsur, line 284: Prefer linking to the class instead of just using the name

```suggestion
        /// If PriorityLevel is also specified at request level in <see cref="RequestOptions.PriorityLevel"/>, that PriorityLevel takes precedence.

```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -278,6 +278,17 @@ public ConnectionMode ConnectionMode
         /// If this is not set the database account consistency level will be used for all requests.
         /// </summary>
         public ConsistencyLevel? ConsistencyLevel { get; set; }
+
+        /// <summary>
+        /// Sets the PriorityLevel for requests created using cosmos client.
```
ealsur, line 283: Should PriorityLevel be 2 words separately?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -435,6 +435,24 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
 
         }
 
+        /// <summary>
+        /// Sets the PriorityLevel for requests created using cosmos client.
+        /// If PriorityLevel is also specified at request level in RequestOptions, that PriorityLevel takes precedence.
+        /// </summary>
+        /// <param name="priorityLevel">The desired consistency level for the client.</param>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
+        /// <seealso href="https://aka.ms/CosmosDB/PriorityBasedExecution"/>
```
ealsur, line 444: This `<seealso` should be added to the other non-builder API?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -435,6 +435,24 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
 
         }
 
+        /// <summary>
+        /// Sets the PriorityLevel for requests created using cosmos client.
+        /// If PriorityLevel is also specified at request level in RequestOptions, that PriorityLevel takes precedence.
+        /// </summary>
+        /// <param name="priorityLevel">The desired consistency level for the client.</param>
```
ealsur, line 442: The parameter text is wrong, it's talking about a consistency level?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -435,6 +435,24 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
 
         }
 
+        /// <summary>
+        /// Sets the PriorityLevel for requests created using cosmos client.
+        /// If PriorityLevel is also specified at request level in RequestOptions, that PriorityLevel takes precedence.
```
ealsur, line 440: Similar comments as the other API. Use `<remarks>` and prefer cref

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -431,6 +436,29 @@ private async Task ValidateAndSetConsistencyLevelAsync(RequestMessage requestMes
...
 
+        /// <summary>
+        /// Set the PriorityLevel in the request headers
+        /// </summary>
+        /// <param name="requestMessage"></param>
+        private void SetPriorityLevel(RequestMessage requestMessage)
+        {
+            Cosmos.PriorityLevel? priorityLevel = null;
+            RequestOptions promotedRequestOptions = requestMessage.RequestOptions;
+            if (promotedRequestOptions != null && promotedRequestOptions.PriorityLevel.HasValue)
```
ealsur, line 447:
```suggestion
            if (promotedRequestOptions?.PriorityLevel.HasValue)
```

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -431,6 +436,29 @@ private async Task ValidateAndSetConsistencyLevelAsync(RequestMessage requestMes
             }
         }
 
+        /// <summary>
+        /// Set the PriorityLevel in the request headers
+        /// </summary>
+        /// <param name="requestMessage"></param>
+        private void SetPriorityLevel(RequestMessage requestMessage)
+        {
+            Cosmos.PriorityLevel? priorityLevel = null;
```
ealsur, line 445: If the variable is `Cosmos.PriorityLevel?` might as well initialize it with:

```suggestion
            Cosmos.PriorityLevel? priorityLevel = this.RequestedClientPriorityLevel;
```

This avoids the `else if` clause below.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -842,6 +853,16 @@ internal virtual ConnectionPolicy GetConnectionPolicy(int clientId)
 
             return (Documents.ConsistencyLevel)this.ConsistencyLevel.Value;
         }
+
+        internal Documents.PriorityLevel? GetDocumentsPriorityLevel()
```
ealsur, line 857: Why is this method needed? The tests can be asserted with the V3 type

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: Used cref

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
Achint-Agrawal: Added

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: Changed to priority level

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
Achint-Agrawal: Corrected to priority level

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
Achint-Agrawal: Done

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
Achint-Agrawal: Changed

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
Achint-Agrawal: Changed as suggested

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: Removed

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/BaselineTest/TestBaseline/PartitionKeyHashBaselineTest.Lists.xml
```diff
@@ -1,32 +1,32 @@
 ﻿<Results>
-	<Result>
```
kirankumarkolli, line 2: Are cahnges to this file required?

Achint-Agrawal: Reverted changes to this file

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -278,6 +278,21 @@ public ConnectionMode ConnectionMode
         /// If this is not set the database account consistency level will be used for all requests.
         /// </summary>
         public ConsistencyLevel? ConsistencyLevel { get; set; }
+
```
kirankumarkolli, line 281: From https://learn.microsoft.com/en-us/azure/cosmos-db/priority-based-execution?tabs=net-v3

`In cases where the number of requests exceeds the capacity that can be processed within the configured Request Units per second (RU/s), then Azure Cosmos DB throttles low priority requests to prioritize the execution of high priority requests`

If the RU's are all used already then any request will be throttled right?
Above statement is confusing. And unclear on how to use this feature.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -278,6 +278,21 @@ public ConnectionMode ConnectionMode
         /// If this is not set the database account consistency level will be used for all requests.
         /// </summary>
         public ConsistencyLevel? ConsistencyLevel { get; set; }
+
+        /// <summary>
+        /// Sets the priority level for requests created using cosmos client.
```
kirankumarkolli, line 283: Do low-priority throttling have different sub-status code to distinguish and troublehsooting on SDK?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -278,6 +278,21 @@ public ConnectionMode ConnectionMode
         /// If this is not set the database account consistency level will be used for all requests.
         /// </summary>
         public ConsistencyLevel? ConsistencyLevel { get; set; }
+
+        /// <summary>
+        /// Sets the priority level for requests created using cosmos client.
+        /// </summary>
+        /// <remarks>
```
kirankumarkolli, line 285: For meta-data requests (ex: Collection read, partitions read etc...) which might goto master parition, can client influce the decision? 

If not are there guard on server/gateway to handle/cover for rogue clients?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: When RU limit is exhausted, we start by throttling low priority. If that is enough to bring down RU usage below provisioned, we only throttle low priority. Other wise we start throttling high priority as well.
Maybe this can be worded better in the documentation to reflect that high priority will also be throttled eventually. Will see if we can do this.
CC: @richagaur

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -278,6 +278,21 @@ public ConnectionMode ConnectionMode
         /// If this is not set the database account consistency level will be used for all requests.
         /// </summary>
         public ConsistencyLevel? ConsistencyLevel { get; set; }
+
+        /// <summary>
+        /// Sets the priority level for requests created using cosmos client.
+        /// </summary>
+        /// <remarks>
+        /// If priority level is also set at request level in <see cref="RequestOptions.PriorityLevel"/>, that priority is used.
```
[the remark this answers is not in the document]
Achint-Agrawal: Global Rate limiter just allows the replica to use unused RUs from other replicas of the account. When this limit is exhausted, PBE will take affect and throttle low priority first to see if it suffices to bring the usage below provisioned throughput. 
In the future, if needed we will put restrictions on how much burst capacity / grl capacity low priority requests can consume.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: If a customer does a, for example ReadContainerAsync call (which is a metadata request), because the code is in RequestInvokerHandler, then we would apply the header. All requests flow through RequestInvokerHandler. 
The question is, does the service ignore the header in those cases or will it also apply priority to metadata for master RU budget?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli: > When RU limit is exhausted, we start by throttling low priority. If that is enough to bring down RU usage below provisioned, we only throttle low priority.

Let's say CX provided 10K RUPS, exhaustion means all 10K are consumed right? 
Even after 10K RUPS usage, high priority requests will be served? 

Apologies but its lot confusing. 

Some version of PriorityLevel discussion had ways to define the boundaries/limits for each priority level. Is that not the case forward at-all?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli: @ealsur is priority-level then needed in diagnostics? 

@sourabh1007 telemetry?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
richagaur: There is some extra budget above 100%, 50% extra (@Achint-Agrawal ) to confirm, which allows cx to swallow some of the peaks in a particular second and throttling happens afterwards in subsequent secs. However, if PBE is enabled, once 100% is reached, all low priority request would be throttled i.e. throttling itself starts at 100% rather than 150% for low priority while high priority would still be allowed which gives them priority of execution. Now in subsequent secs, the throttling will happen only for low priority while high priority can still go through if its budget is still under 100%. In usual scenarios, any request would have been throttled in subsequent secs after more than 100% was utilized previously, this amortized over a period of time provides the impact that when budget crosses over 100% combined (for low and high) only low priority requests are throttled while high still goes through giving the impression of priority of execution. Now, if high itself is over 100% then high priority will also be throttled. 

Coming to defining boundaries for different priority levels, we are defining min and max thresholds which would guaranteed  by system to allow(min) and cap (max) at any time, the implementation is currently in progress.

kirankumarkolli: Thanks for clarifying @richagaur. 

Conceptually right now [100%, 150%] is reserved for high priority and [0, 100%] is available all priorities.
In future we will allow tuning/changing those boundaries right?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli: At-least updating the 429/3200, troubleshooting guide makes sense.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: FYI, here's the TSG for PBE - https://supportability.visualstudio.com/AzureCosmosDB/_wiki/wikis/AzureCosmosDB.wiki/1223924/TSG-Throttling-Priority-based-execution

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: FYI, here's the TSG for PBE - https://supportability.visualstudio.com/AzureCosmosDB/_wiki/wikis/AzureCosmosDB.wiki/1223924/TSG-Throttling-Priority-based-execution

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: Metadata requests which go to Master partition will also get the PriorityLevel set at the client level and it will be obeyed if Priority Based Execution is enabled. Is there an issue with this?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: Clarification: We don't throttle as soon as the RU limit is hit. We allow borrowing from the next second and throttle requests in the next seconds to balance out for the extra RUs. When the borrowed RUs is < 50% of provisioned throughput, we allow both low and high. If it is >50% and <100%, we throttle low priority and allow only high priority. If it is > 100%, we throttle both.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
richagaur: Metadata requests are not in scope of priority-based execution. We don't want to prioritize any metadata request over other given most are system requests anyways. Nor is there any use-case for it.

kirankumarkolli: Ideal is to exclude it in gateways.

kirankumarkolli (dismissed):

kirankumarkolli:
/cc: @kundadebdatta ideal if we can also get this PR for 3.38 release.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -431,6 +436,25 @@ private async Task ValidateAndSetConsistencyLevelAsync(RequestMessage requestMes
             }
         }
 
+        /// <summary>
+        /// Set the PriorityLevel in the request headers
+        /// </summary>
+        /// <param name="requestMessage"></param>
+        private void SetPriorityLevel(RequestMessage requestMessage)
+        {
+            Cosmos.PriorityLevel? priorityLevel = this.RequestedClientPriorityLevel;
```
ealsur, line 445: If we want to exclude metadata requests from Priority Level maybe this helps?

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/06333893e3315cfe54bc733e3363596b6f1974a6/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L215-L220

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
FabianMeiswinkel: Let the service make the call whether to ignore this header or not. That is what is happening anyway - and any enforcement on the client is immutable - while on the service we can evolve as we see fit. And given that there are never any hard guarantees anyway, where is the harm?

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
Achint-Agrawal: As discussed on the group chat, we will pass the priority from the SDK and check whether we need to ignore it in the backend.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -431,6 +436,25 @@ private async Task ValidateAndSetConsistencyLevelAsync(RequestMessage requestMes
...
+            RequestOptions promotedRequestOptions = requestMessage.RequestOptions;
+            if (promotedRequestOptions?.PriorityLevel.HasValue == true)
```
ealsur, line 447:
```suggestion
            if (promotedRequestOptions?.PriorityLevel.HasValue)
```

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -431,6 +436,25 @@ private async Task ValidateAndSetConsistencyLevelAsync(RequestMessage requestMes
...
+            {
+                priorityLevel = promotedRequestOptions.PriorityLevel;
```
ealsur, line 449:
```suggestion
                priorityLevel = promotedRequestOptions.PriorityLevel.Value;
```

ealsur (dismissed):
Left some nits, I think if you use "Add suggestion to batch" from the Files view and you apply them, it would not remove the approval.

If we are pressed on time, feel free to leave it

kirankumarkolli (approved):

kirankumarkolli (dismissed):

kundadebdatta (dismissed):
LGTM.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -442,6 +442,27 @@ public CosmosClientBuilder WithConsistencyLevel(Cosmos.ConsistencyLevel consiste
...
+        /// Sets the priority level for requests created using cosmos client.
+        /// </summary>
+        /// <remarks>
+        /// If priority level is also set at request level in <see cref="RequestOptions.PriorityLevel"/>, that priority is used.
+        /// If <see cref="WithBulkExecution(bool)"/> is set to true, priority level set on the CosmosClient is used.
+        /// </remarks>
+        /// <param name="priorityLevel">The desired priority level for the client.</param>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
+        /// <seealso href="https://aka.ms/CosmosDB/PriorityBasedExecution"/>
+#if PREVIEW
```
kirankumarkolli, line 455: Fix it

kirankumarkolli (approved):

kundadebdatta (approved):
LGTM.

PR #4264, opened 2024-01-18
[Internal] Change Feed Processor: Adds AllVersionsAndDeletes support

### LINKED ISSUE #4007: FFCF API discussion
# FFCF API Discussion

I've made an initial PR for FFCF (aka AllVersionsAndDeletes) support in https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4003 and this is to discuss the potential API

The main API for this feature is the addition of a new `ChangeFeedProcessorBuilder` in `Microsoft.Azure.CosmosDB.Container` which takes a new handler type `AllVersionsAndDeleteChangesHandler<T> (IReadOnlyCollection<ChangeFeedProcessItem<T>> changes, CancellationToken cancellationToken)` which is similar to the existing handlers, but instead of a `IReadOnlyCollection<T>` it wraps `T` in a `ChangeFeedProcessItem` which is a wrapper that contains the current and previous state of `T` and then the metadata.

The code reuses the `ChangeFeedMode` type from the other change feed code, but this necessitates the mode being passed through some other public API.

Currently there are no overloads that pass a default mode, but those could be added if we think they're necessary, I don't know how often they're actually meant to be called by a client.

When thinking about how to do this, I looked at the Java API, but it is a completely different system and I felt it was better to match what we have in the rest of the C# API for simplicity's sake.

API example

```csharp

var builder = container.GetAllVersionsChangeFeedProcessorBuilder<CartAction>("migrationProcessor", (ChangeFeedProcessorContext context, IReadOnlyCollection <ChangeFeedItemChange<CartAction>> changes, CancellationToken cancellationToken) =>
{
    Console.WriteLine(changes.Count + " Changes Received");


    {
    }
}

```

philipthomas-MSFT:
## Description

See [4007](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4007) for issue statement.

The deprecated forked PR exists [here](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4003).

## Type of change

## Closing issues

To automatically close an issue: closes #4007


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs
```diff
@@ -25,8 +26,11 @@ public async Task TestInitialize()
             await base.ChangeFeedTestInit();
 
             string PartitionKey = "/pk";
-            ContainerResponse response = await this.database.CreateContainerAsync(
-                new ContainerProperties(id: Guid.NewGuid().ToString(), partitionKeyPath: PartitionKey),
+            ContainerProperties properties = new ContainerProperties(id: Guid.NewGuid().ToString(), 
+                partitionKeyPath: PartitionKey);
+            properties.ChangeFeedPolicy.FullFidelityRetention = TimeSpan.FromMinutes(5);
```
philipthomas-MSFT, line 31: Need to change this. No need to set retention for non AllVersionsAndDeletes tests.

philipthomas-MSFT (commented):
Blocking myself because I want to include more test. But opened PR so that others can review and give feedback on the code enhancements.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
```diff
@@ -69,6 +69,17 @@ public ChangeFeedProcessorBuilder WithInstanceName(string instanceName)
             return this;
         }
 
+        /// <summary>
+        /// Sets the mode for the change freed processor.
+        /// </summary>
+        /// <param name="changeFeedMode"></param>
+        /// <returns>The instance of <see cref="ChangeFeedProcessorBuilder"/> to use.</returns>
+        internal ChangeFeedProcessorBuilder WithChangeFeedMode(ChangeFeedMode changeFeedMode)
```
ealsur, line 77: Should this be public for preview?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedProcessing/FeedProcessorFactoryCore.cs
```diff
@@ -46,6 +46,7 @@ public override FeedProcessor Create(DocumentServiceLease lease, ChangeFeedObser
             PartitionCheckpointerCore checkpointer = new PartitionCheckpointerCore(this.leaseCheckpointer, lease);
             ChangeFeedPartitionKeyResultSetIteratorCore iterator = ChangeFeedPartitionKeyResultSetIteratorCore.Create(
                 lease: lease,
+                mode: mode,
```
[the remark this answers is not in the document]
ealsur: Doing that simplifies the entire PR by reducing the file count and the Mode can be defaulted to LatestVersion.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
```diff
@@ -1325,6 +1323,25 @@ public abstract class Container
...
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        virtual ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(
+            string processorName,
+            ChangeFeedHandler<ChangeFeedItemChange<T>> onChangesHandler)
+        {
+            throw new NotImplementedException();
```
ealsur, line 1342: Is this throw expected? Or should it wire the Mode?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedProcessing/FeedProcessorFactoryCore.cs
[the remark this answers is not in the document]
ealsur: Looking at the other files, it does look like you added it there, in that case, they are available in `options.Mode`?

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
```diff
@@ -433,7 +433,6 @@ public abstract class Container
         /// * "_ts": Gets the last modified time stamp associated with the item from the Azure Cosmos DB service.
         /// * "_etag": Gets the entity tag associated with the item from the Azure Cosmos DB service.
         /// * "ttl": Gets the time to live in seconds of the item in the Azure Cosmos DB service.
-        /// Note that, this API does not support the usage of <see cref="RequestOptions.IfMatchEtag"/> property at the moment.
```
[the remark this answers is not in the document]
philipthomas-MSFT: I can remove it. It is something Ian added. Not sure the context behind why.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
philipthomas-MSFT: I asked the same question to Ian, but not quite understanding the answer. The original [comment](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4003#discussion_r1344480347).

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedProcessing/FeedProcessorFactoryCore.cs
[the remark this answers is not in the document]
philipthomas-MSFT: Yes, I can remove the _ChangeFeedMode_ from the signature of **Create**. I don't see a need to have that if we have it already in _ChangeFeedProcessorOptions_.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
ealsur: Seems like a placeholder where the wiring is expected

Microsoft.Azure.Cosmos.Encryption/src/Microsoft.Azure.Cosmos.Encryption.csproj
```diff
@@ -26,7 +26,7 @@
   <ItemGroup>
     <AdditionalFiles Include="..\..\Microsoft.Azure.Cosmos\src\stylecop.json" Link="stylecop.json" />
   </ItemGroup>
-	
+
```
ealsur, line 29: Undo this file, it is just changing whitespace

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerInlineCore.cs
```diff
@@ -515,6 +515,89 @@ Task<ResponseMessage> func(ITrace trace)
...
+        /// await this.Container.DeleteItemAsync<dynamic>(id: "1", partitionKey: new PartitionKey("1"));
+        /// 
+        /// allProcessedDocumentsEvent.WaitOne(10 * 1000);
+        /// 
+        /// await changeFeedProcessor.StopAsync();
+        /// ]]>
+        /// </code>
+        /// </example>
+        /// <returns>An instance of <see cref="ChangeFeedProcessorBuilder"/></returns>
+#if PREVIEW
```
ealsur, line 589: This is always `public`, it is an `internal` class.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
```diff
@@ -1325,6 +1324,25 @@ public abstract class Container
...
+        /// <typeparam name="T">Document type</typeparam>
+        /// <param name="processorName">A name that identifies the Processor and the particular work it will do.</param>
+        /// <param name="onChangesHandler">Delegate to receive all changes and deletes</param>
+        /// <returns>An instance of <see cref="ChangeFeedProcessorBuilder"/></returns>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        virtual ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(
```
ealsur, line 1339: The only design-level question I have is: If a customer starts a CFP with AllVersionsAndDeletes and afterwards changes it to LatestVersion (or viceversa) should it pick up the lease containuation? is that something we want to support? What was done in Java? cc @jcocchi

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -0,0 +1,159 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [TestClass]
+    [TestCategory("ChangeFeedProcessor with AllVersionsAndDeletes")]
+    public class GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests : BaseChangeFeedClientHelper
```
ealsur, line 16: For Emulator tests, please add them test suite for CFP on https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs

ealsur (commented):
Great progress! Some polishing comments and an open design question regarding lease compatibility

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerInlineCore.cs
philipthomas-MSFT: Are we not Previewing this? Would have to change the access modifier of `ChangeFeedItemChange<T>` if we started removing #if PREVIEW from these methods and defaulting it to public so we need to keep it like it is unless we are saying we are skipping Preview.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
philipthomas-MSFT: Looking into it.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
ealsur: Thanks for the offline sync, the reason this requires a new test file is the initialization uses Retention

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerInlineCore.cs
ealsur: `ContainerInlineCore.cs` is an internal file.

The public contracts are in `Container.cs`.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
ealsur: Could it be that they are using the mode as part of the lease `id`? Can Abhijeet confirm this?

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerInlineCore.cs
kirankumarkolli: Public/preview is only limited at the contract scope. 
All implementation is shared.

philipthomas-MSFT: So, are we saying that we are good with changing `ChangeFeedItemChange<T>` also and it should not be behind `#if PREVIEW`? Changing the access modifier for `GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>` to public means that `ChangeFeedItemChange<T>` also needs to be public. Same thing applies to `ChangeFeedMetadata`. Get my point?

philipthomas-MSFT: _internal_ ContainerInline.Core (inherits)-> _internal_ ContainerCore (inherits)-> _internal_ ContainerInternal (inherits)-> _public_ Container

Container has
```
#if PREVIEW
        public
#else
        internal
#endif 
        abstract ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(
            string processorName,
            ChangeFeedHandler<ChangeFeedItemChange<T>> onChangesHandler);
```

But `ChangeFeedItemChange<T>` is used by all the types above which is a mix of _internal_ and _public_. This means that either we make `ChangeFeedItemChange<T>` _public_, or make the implementation of `GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>` in all internal types _internal_.

Or leave it as it is.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
philipthomas-MSFT: I am making `GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>` this abstract in _public_ Container.cs because there is no implementation. I am also moving the documentation/example here as well instead of the _internal_ types.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
philipthomas-MSFT: > Could it be that they are using the mode as part of the lease `id`? Can Abhijeet confirm this?

@jeet1995 can you weigh in. Thanks.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionSupervisorFactoryCore.cs
```diff
@@ -28,7 +28,7 @@ internal sealed class PartitionSupervisorFactoryCore : PartitionSupervisorFactor
             this.partitionProcessorFactory = partitionProcessorFactory ?? throw new ArgumentNullException(nameof(partitionProcessorFactory));
         }
 
-        public override PartitionSupervisor Create(DocumentServiceLease lease)
+        public override PartitionSupervisor Create(DocumentServiceLease lease, ChangeFeedMode mode)
```
ealsur, line 31: Where is the mode used here?

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
kirankumarkolli: Otherwise how can that combination/condition will be detected? 

> CFP will not return documents if the mode is switched and also no exceptions 
Combination is very interesting. Won't CX interpret it as no-changes and CRI to us?

ealsur: If the mode is added to the `id` of the leases, then each mode will have different set of leases (similar to processorName/leasePrefix) and they won't mix. But we should be consistent with what Java did

ealsur: Thanks @jeet1995 for the details. I think the current code in this PR aligns with this behavior.

The point, which now might be a question for @jcocchi, is that the Continuation from one mode to the other is in theory valid, because it's just the LSN. The service won't fail or reject the request, but the behavior as described above, might be weird. 

Mainly because when switching inbetween modes, there might be scenarios that don't seem compatible. For example, if running an Incremental CFP, the Continuation might be an LSN older than the Full Fidelity Retention. If that CFP switches to AllVersionsAndDeletes, then the Continuation passed will be outside of retention.

In a nutshell the question is, do we want to support customers changing the mode? Do we want each mode to track different progress (for example, if we add the mode to the `id`, then each mode would create different set of leases and they would not mix) or do we want to throw an exception (the `mode` should be part of the lease and we should verify it matches with the CFP Builder mode).

jcocchi: Few comments:
- `AllVersionsAndDeletes` shouldn't support `startFromBeginning`, if this doesn't error currently it should definitely error. Let's track this as a future update for Java and verify the behavior in .NET.
- In the future, I want to leave the option open to support switching modes with the same leases, so I don't want to add `mode` to the `id`. Let's error on this scenario for now until we can test all edge cases. If the `mode` doesn't match what's stored in the lease then throw an exception.
    - @ealsur Reading an LSN outside of the retention could happen when switching modes as you mentioned, but it can also happen when reading old leases from `AllVersionsAndDeletes` mode itself. Switching modes may make it more likely for changes to be outside of retention, but deciding if a LSN is outside of the retention should happen in the backend and doesn't need to be a blocker for SDK design IMO. 
- @jeet1995 I'm confused about scenario 1. Sounds like you ingested 10 documents, read the 10 documents in `AllVersionsAndDeletes` mode then switched to `LatestVersion` mode. There were no new documents ingested after that so why would `LatestVersion` pick up anything? The `startFromBeginning` key only applies for the first run, if there is existing state in the leases then that takes precedence ([reference doc](https://azure.github.io/azure-cosmosdb-java/3.0.0/com/azure/data/cosmos/ChangeFeedProcessorOptions.html#startFromBeginning--)). Sounds like this scenario is functionally the same as scenario 2 for Java. @philipthomas-MSFT let's verify what happens in .NET.

jeet1995: @jcocchi - my bad - there is no prior test coverage for the lease sharing scenario therefore my assumption was a mode change shouldn't support a lease sharing scenario at least out of the box / the API ref doesn't call it out / `AllVersionAndDeletes` doesn't error out either when `StartFrom` is `BEGINNING` in the lease - so different modes don't have contextual knowledge of the same lease as of yet. Once I ran through the scenario, the lease is indeed shared with continuation being picked up in-line with the doc. I can validate the other scenario as well.

kirankumarkolli: > `startFromBeginning`: Will it never be supported ever? 

SDK once shipped is immutable. So one option is to error out from service => can evolve as service gets enriched more. True experience wise it needs one extra NW call for it to fail, seems a reasonable trade-off. 

> In the future, I want to leave the option open to support switching modes with the same leases, so I don't want to add mode to the id

Switching modes practically won't be atomic right? ex: multi-instance processor and the change needs to roll-through all nodes, otherwise the behavior will be non-deterministic. 

One option/alternative is to treat them as migrations. I.e. create a new lease container with the new mode with current LSN state. 

@jcocchi thoughts?

jcocchi: `startFromBeginning` should error from the service, and I tested that it does on the .NET Pull model. Not sure why CFP doesn't see it as invalid. Maybe CFP is only looking at LSNs, and the LSN was still in the retention window for this case? By design `startFromBeginning` means "beginning of container" `AllVersionsAndDeletes` max retention is 7 days or 30 days, so `startFromBeginning` could only work for new containers in that mode.

@kirankumarkolli agree switching modes needs design work and testing. Let's error for now without closing the opportunity to add support for this in the future. My only comment was adding `mode` to the `id` might make it difficult to support the scenario later.

kirankumarkolli: With CFP, the processor might attempt NW interaction only after its started, i.e. in background. So error will be observed by the application code (not-straight forward for troubleshooting). Can it be treated as fatal failure and stop the processor? @ealsur thoughts? 

On `switching modes`: Even to detect it the lease needs to have the previous mode context right? 
We can include the mode part of the lease and later in logic we can ignore if needed (logic change new CFP/SDK version refresh is needed by application)

ealsur: The only way is if we wired the detection of the mode change through the StartAsync operation. That is the only way you can stop the CFP from starting. Once started, everything is background, the processing cannot be stopped.

If we throw exceptions, if the customer has Notifications API wired, they would see them, but that's all

kirankumarkolli: That's right, even mode validation needs interaction with lead-container => NW interaction. May be StartAsync can be delayed till mode validation. 

`startfrombegining` in non-recoverable/fatal in-theory. May be we can get started with NotificationsAPI for now and later revisit based on CRI support-load?

jeet1995: @jcocchi / @philipthomas-MSFT - stopping `LatestVersion` CFP after an arbitrary no. of documents have been processed and continuing with `AllVersionsAndDeletes` CFP picks up from the continuation of the 'shared' lease in the Java SDK. 

For ref - https://github.com/jeet1995/azure-sdk-for-java/pull/8

philipthomas-MSFT: Hi. I am going to resolve this, but move this conversation since we are not blocking this PR. We will continue this conversation on a new issue once I can gather more information and grow my understanding. Thanks.

Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainer.cs
```diff
@@ -1329,6 +1329,19 @@ private static ItemRequestOptions EncryptionContainerGetClonedItemRequestOptions
             return decryptedItems;
         }
 
+        private async Task<T> DecryptChangeFeedDocumentAsync<T>(
```
ealsur, line 1332: This seems no longer needed?

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
```diff
@@ -1690,7 +1688,6 @@ public abstract class Container
         public abstract Task<IEnumerable<string>> GetPartitionKeyRangesAsync(
             FeedRange feedRange,
             CancellationToken cancellationToken = default);
-
 #endif
```
ealsur, line 1691: Undo this file

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -481,9 +482,31 @@ internal abstract partial class ContainerCore : ContainerInternal
             ChangeFeedObserverFactory observerFactory = new CheckpointerObserverFactory(
                 new ChangeFeedObserverFactoryCore<T>(onChangesDelegate, this.ClientContext.SerializerCore),
                 withManualCheckpointing: false);
-            return this.GetChangeFeedProcessorBuilderPrivate(processorName, observerFactory);
+            return this.GetChangeFeedProcessorBuilderPrivate(processorName,
+                observerFactory, ChangeFeedMode.LatestVersion);
         }
 
+        //////public override ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(
```
ealsur, line 489: Remove commented code

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.cs
```diff
@@ -11,11 +11,9 @@ namespace Microsoft.Azure.Cosmos
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Azure.Cosmos.ChangeFeed;
-    using Microsoft.Azure.Cosmos.ChangeFeed.Pagination;
-    using Microsoft.Azure.Cosmos.ChangeFeed.Utils;
+    using Microsoft.Azure.Cosmos.ChangeFeed.FeedProcessing;
```
ealsur, line 14: What is using this `using`?

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.cs
```diff
@@ -700,5 +698,26 @@ public override async Task<CollectionRoutingMap> GetRoutingMapAsync(Cancellation
                 changeFeedIteratorCore,
                 responseCreator: this.ClientContext.ResponseFactory.CreateChangeFeedUserTypeResponse<T>);
         }
+
+        public override ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(
```
ealsur, line 702: This probably should go to ContainerCore.Items.cs along with all the other CFP methods

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializer.xml
```diff
@@ -0,0 +1,48 @@
+﻿<Results>
+  <Result>
+    <Input>
+      <Description><![CDATA[Filter w/ DataObject initializer with constant value]]></Description>
```
ealsur, line 4: Undo this file

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializerDotNet.xml
```diff
@@ -0,0 +1,236 @@
+﻿<Results>
+  <Result>
+    <Input>
+      <Description><![CDATA[Filter w/ constant value, camelcase = True]]></Description>
```
ealsur, line 4: Undo this file

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializerMultiSerializer.xml
```diff
@@ -0,0 +1,236 @@
+﻿<Results>
+  <Result>
+    <Input>
+      <Description><![CDATA[Filter w/ constant value, camelcase = True]]></Description>
```
ealsur, line 4: undo this file

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/BaselineTest/TestBaseline/PartitionKeyHashBaselineTest.Lists.xml
```diff
@@ -1,32 +1,32 @@
...
-			<PartitionKeyHashV2>B6-3D-B6-C4-3A-4F-1B-01-4C-63-B0-C2-E6-28-E8-12-A6-0C-6C-BE-5A-2D-38-6E-5D-AE-1A-AC-94-21-6B-6C-88-A6-18-5D-2D-D5-1C-96-D0-47-75-B7-2E-FA-BE-08</PartitionKeyHashV2>
-		</Output>
-	</Result>
+  <Result>
+    <Input>
+      <Description>1 Path List</Description>
+      <PartitionKeyValue>["/path1"]</PartitionKeyValue>
+    </Input>
+    <Output>
+      <PartitionKeyHashV1>00-00-00-00-00-00-00-00-00-00-00-00-0A-A1-CC-05</PartitionKeyHashV1>
```
ealsur, line 8: Undo this file

philipthomas-MSFT: Ok, thanks.. So, I think we need to have some conversation about UpdateContracts powershell.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializerMultiSerializer.xml
philipthomas-MSFT: Ok, thanks.. So, I think we need to have some conversation about UpdateContracts powershell.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializerDotNet.xml
philipthomas-MSFT: Ok, thanks.. So, I think we need to have some conversation about UpdateContracts powershell.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqTranslationWithCustomSerializerBaseline.TestMemberInitializer.xml
philipthomas-MSFT: Ok, thanks.. So, I think we need to have some conversation about UpdateContracts powershell.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/BaselineTest/TestBaseline/PartitionKeyHashBaselineTest.Lists.xml
ealsur: Yeah. UpdateContracts.ps1 seems to be affected by the local machine's spacing settings or something like that, I personally dont know

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
jcocchi: Thanks for verifying in Java @jeet1995! 

I think that is the desired long-term behavior, but based on this conversation both .NET and Java should throw when the change feed mode is switched and the leases are the same for the time being. We can cover it offline, and it's non-blocking as Philip mentioned.

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.cs
philipthomas-MSFT: > What is using this `using`?

`CheckpointerObserverFactory` belongs to `using Microsoft.Azure.Cosmos.ChangeFeed.FeedProcessing`

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -1213,17 +1219,38 @@ private string ContcatCachedUriWithId(string resourceId)
                 trace: trace,
                 cancellationToken: cancellationToken);
         }
+        public override ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(
```
kirankumarkolli, line 1222: `ContainerCore.Items.cs` is an internal class not public contract right? 
Isn't it needs to be public API on Container.cs?

philipthomas-MSFT: We already determined that we are not making this public because of the current conversation on switching change feed modes.

ealsur, line 1222: nit: Line space with previous `{`

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.cs
```diff
@@ -11,11 +11,8 @@ namespace Microsoft.Azure.Cosmos
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Azure.Cosmos.ChangeFeed;
-    using Microsoft.Azure.Cosmos.ChangeFeed.Pagination;
-    using Microsoft.Azure.Cosmos.ChangeFeed.Utils;
     using Microsoft.Azure.Cosmos.Diagnostics;
     using Microsoft.Azure.Cosmos.Pagination;
```
ealsur, line 15: undo this file

ealsur (approved):

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
philipthomas-MSFT: Compiling all the feedback in this discussion and transferring it [here](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4308) along with some clarifying questions. Once I have a solid understanding and alignment on the direction, I will finalize, implement and deliver. Thanks. @kirankumarkolli @ealsur @jeet1995 @jcocchi

kirankumarkolli (approved):

PR #2355, opened 2021-03-31
Gateway: Adds parallel retry logic to get account information.

j82w:
# Pull Request Template

## Description

This reduces the time to get the account information from over 70 seconds if the primary region gateway is down to a little over 5 seconds if the user has preferred regions configured. The get account information is needed on SDK initialization and on certain failures which requires getting the updated account information to check for failovers. 

Current SDK logic
Global Endpoint -> 5 sec
Global Endpoint -> 10 sec
Global Endpoint -> 20 sec
Primary Endpoint (which most of the time is the same as the global endpoint) -> 5 sec
Primary Endpoint -> 10 sec
Primary Endpoint -> 20 sec
Secondary Endpoint -> success

Total time to get the success from the secondary region is over 70 seconds + time for the secondary endpoint to respond

New SDK logic
```
        /// This gets the account information
        /// 
        /// Source Task        
        /// Creates Task 1,2 ->                |    Task 1                         |    Task 2          |                 
        ///                                    | Global endpoint -> 10 sec -> fail | Timer wait 5 sec   |
        /// Waits for Any on (Task1, Task2)    | still waiting on response         | Timer is done      |
        /// Creates Task3, Task4                                                                        |     Task 3                                |     Task 4                                 |   
        ///                                                                                             | 1st preferred location -> 10 sec -> fail  | 2nd preferred location -> 1 sec -> success |
        ///                                                                                             | still waiting on response                 | returns success                            |
        /// Waits for Any on (Task1, Task3, Task 4). Task 4 is done return the account information.
        /// Other tasks log the exception or just ignore the response
        /// This gets the account information with multiple preferred region failures
        /// 
        /// Source Task        
        /// Creates Task 1,2 ->                |    Task 1                         |    Task 2          |                 
        ///                                    | Global endpoint -> 10 sec -> fail | Timer wait 5 sec   |
        /// Waits for Any on (Task1, Task2)    | still waiting on response         | Timer is done      |
        /// Creates Task3, Task4                                                                        |     Task 3                                |     Task 4                               |   
        ///                                                                                             | 1st preferred location -> 10 sec -> fail  | 2nd preferred location 3 sec -> fail     |
        ///                                                                                             | 4th preferred location 7 sec -> fail      | 3rd preferred location 1 sec -> fail     |
        ///                                                                                             | still waiting on response                 | 5th preferred location 1 sec -> success  |
        /// Waits for Any on (Task1, Task3, Task 4). Task 4 is done return the account information.
[...]

PR #4265, opened 2024-01-19
CosmosClientOptions: Adds Private Custom Account Endpoints

### LINKED ISSUE #4236: [Per Partition Automatic Failover] Utilize `CosmosClientOptions` to Capture Custom Domain Names to resolve Cx Specified Endpoints
### Problem Statement:

In real world, there are many reasons an endpoint could become non-responsive and stop fullfilling any incoming requests. Some examples for such reasons could be network packet drops, the server node experiencing issues or a larger outage. Today, while initializing, the .NET v3 SDK requires to fetch the account metadata information from the routing gateway, using the global account endpoint. This information is needed to figure out the read, write regions, the resource identifiers, ETag etc, which are needed by the SDK to perform Read/ Write operations. The global account endpoint is passed through the `CosmosClient` constructor (see the below example for more details).

```
{
    ApplicationPreferredRegions = new List<string>()
    {
        Regions.NorthCentralUS,
        Regions.EastAsia,
    },
    EnablePartitionLevelFailover = true,
    ConnectionMode = ConnectionMode.Direct,
};

    "https://testaccount.documents-test.windows-int.net:443/",
    "key==",
    clientOptions
);
```

However, if for some unforeseen reason, the global account endpoint becomes non-responsive, today there is no way to fetch the account metadata information, thus failing the cosmos client initialization. 

### Proposed Solution: ###

The above problem could be solved if the global account metadata information is hosted with-in a private domain name. During an outage, the custom domain names can be used to route the Get Account metadata requests to the custom endpoints, if the primary global account endpoint become non-responsive.

Below is an example of how the SDK will capture the regional endpoints from the end user.

```
    {
        ApplicationPreferredRegions = new List<string>()
        {
            Regions.P1,
            Regions.P2,
            Regions.P3
        },
        RegionalEndpoints = new List<string>()
        {
            { "custom.p-1.documents.azure.com" },
            { "custom.p-2.documents.azure.com" },
        },
        EnablePartitionLevelFailover = true,
    };
```

### Acceptance Criteria: ### 

- Create a **new** and **optional** property `RegionalEndpoints` in the `CosmosClientOptions` so that the user/ customer is able to provide the list of custom account endpoints to resolve the account metadata accurately.

- If the SDK is unable to fetch the account metadata using the global endpoint, it should be able to use the custom regional endpoints to fulfill the operation, and fetch the metadata information successfully.

### Notes: ### 

- The `CosmosClientOptions.RegionalEndpoints` parameter is optional during the SDK initialization. It remains optional regardless of whether the partition level failover is enabled or not.

- The `CosmosClientOptions.RegionalEndpoints` doesn't need to maintain the same order as it appears in the `CosmosClientOptions.ApplicationPreferredRegions`, if provided. The account metadata information is an one time operation during the cosmos client initialization, and doesn't require to fetch the metadata, unless the SDK is re-initialized.

kundadebdatta:
# Pull Request Template

### Problem Statement:

In real world, there are many reasons an endpoint could become non-responsive and stop fullfilling any incoming requests. Some examples for such reasons could be network packet drops, the server node experiencing issues or a larger outage. Today, while initializing, the .NET v3 SDK requires to fetch the account metadata information from the routing gateway, using the global account endpoint. This information is needed to figure out the read, write regions, the resource identifiers, ETag etc, which are needed by the SDK to perform Read/ Write operations. The global account endpoint is passed through the `CosmosClient` constructor (see the below example for more details).

```
{
    ApplicationPreferredRegions = new List<string>()
    {
        Regions.NorthCentralUS,
        Regions.EastAsia,
    },
    EnablePartitionLevelFailover = true,
    ConnectionMode = ConnectionMode.Direct,
};

    "https://testaccount.documents-test.windows-int.net:443/",
    "key==",
    clientOptions
);
```

However, if for some unforeseen reason, the global account endpoint becomes non-responsive, today there is no way to fetch the account metadata information, thus failing the cosmos client initialization. 

### Proposed Solution: ###

The above problem could be solved if the global account metadata information is hosted with-in a private domain name. During an outage, the custom domain names can be used to route the Get Account metadata requests to the custom endpoints, if the primary global account endpoint become non-responsive.

Below is an example of how the SDK will capture the regional endpoints from the end user.

```
    {
        ApplicationPreferredRegions = new List<string>()
        {
            Regions.P1,
            Regions.P2,
            Regions.P3
        },
        AccountInitializationCustomEndpoints= new HashSet<string>()
        {
            { "custom.p-1.documents.azure.com" },
            { "custom.p-2.documents.azure.com" },
        },
        EnablePartitionLevelFailover = true,
    };
```

## Type of change

## Closing issues

To automatically close an issue: closes #4236


Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
```diff
@@ -90,6 +92,27 @@ public void SetPreferredLocations(IReadOnlyList<string> regions)
             }
         }
 
+        /// <summary>
+        /// Sets the regional endpoints required to fetch account information from
+        /// private domain names.
+        /// </summary>
+        /// <param name="regionalEndpoints">An instance of <see cref="ISet{T}"/> containing the regional endpoints
+        /// provided by the customer.</param>
+        public void SetRegionalEndpoints(
```
kirankumarkolli, line 101: Pattern: All existing pattern for get and set is pattern. 
Do RegionalEndpoints really need different pattern?

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
```diff
@@ -90,6 +92,27 @@ public void SetPreferredLocations(IReadOnlyList<string> regions)
             }
         }
 
+        /// <summary>
+        /// Sets the regional endpoints required to fetch account information from
```
kirankumarkolli, line 96: Region endpoint is a new concept and lets cover more context. 
Call out with an example and in which scenario this needs to be used.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -190,6 +190,37 @@ public string ApplicationName
         /// </example>
         /// <seealso href="https://docs.microsoft.com/azure/cosmos-db/high-availability#high-availability-with-cosmos-db-in-the-event-of-regional-outages">High availability on regional outages</seealso>
         public IReadOnlyList<string> ApplicationPreferredRegions { get; set; }
+
+        /// <summary>
+        /// Gets and sets the regional private endpoints for geo-replicated database accounts in the Azure Cosmos DB service. 
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// During the CosmosClient initialization the account information, including the available regions, is obtained from the <see cref="CosmosClient.Endpoint"/>.
```
kirankumarkolli, line 199: Please call out when to use this API? 
Region endpoints is kind of relatively new concept and needs more explanation and when to use it.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -190,6 +190,37 @@ public string ApplicationName
...
+        /// {
+        ///     RegionalEndpoints = new HashSet<string>(){ "custom.p-1.documents.azure.com", "custom.p-2.documents.azure.com" }
+        /// };
+        /// 
+        /// CosmosClient client = new CosmosClient("endpoint", "key", clientOptions);
+        /// ]]>
+        /// </code>
+        /// </example>
+        /// <seealso href="https://docs.microsoft.com/azure/cosmos-db/high-availability#high-availability-with-cosmos-db-in-the-event-of-regional-outages">High availability on regional outages</seealso>
+        public ISet<string> RegionalEndpoints { get; set; }
```
kirankumarkolli, line 223: IEnumerable might be a based denomintor we can use right? Is ISet really needed?

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
```diff
@@ -86,13 +106,31 @@ private async Task<AccountProperties> GetDatabaseAccountAsync(Uri serviceEndpoin
...
-                getDatabaseAccountFn: this.GetDatabaseAccountAsync,
-                cancellationToken: this.cancellationToken);
+            int attemptCounter = 1;
+            List<Exception> exceptionList = new ();
+
+            while (this.serviceEndpointEnumerator.MoveNext())
+            {
+                try
+                {
+                    AccountProperties databaseAccount = await GlobalEndpointManager.GetDatabaseAccountFromAnyLocationsAsync(
```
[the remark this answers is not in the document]
kirankumarkolli: Conceptually, we can imagine that `this.serviceEndpointEnumerator` is either user given or system populated and GlobalEndpointManager.TryGetAccountPropertiesFromAllLocationsAsync should use it.

kirankumarkolli (changes requested):
Conceptually, we can imagine that this.serviceEndpointEnumerator is either user given or system populated and GlobalEndpointManager.TryGetAccountPropertiesFromAllLocationsAsync should use it

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -190,6 +190,37 @@ public string ApplicationName
...
+        /// <summary>
+        /// Gets and sets the regional private endpoints for geo-replicated database accounts in the Azure Cosmos DB service. 
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// During the CosmosClient initialization the account information, including the available regions, is obtained from the <see cref="CosmosClient.Endpoint"/>.
+        /// Should the global endpoint become inaccessible, the CosmosClient will attempt to obtain the account information issuing requests to the regional endpoints provided in <see cref="RegionalEndpoints"/>.
+        /// </para>
+        /// <para>
+        /// Nevertheless, this parameter remains optional and is recommended for implementation when a customer has configured a private endpoint for their Cosmos DB account.
```
FabianMeiswinkel, line 203:
```suggestion
        /// Nevertheless, this parameter remains optional and is recommended for implementation when a customer has configured a private endpoint with a custom DNS hostname (instead of accountname-region.documents.azure.com) etc. for their Cosmos DB account.
```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -301,6 +301,36 @@ public CosmosClientBuilder WithApplicationPreferredRegions(IReadOnlyList<string>
...
 
+        /// <summary>
+        /// Sets the regional private endpoints for geo-replicated database accounts in the Azure Cosmos DB service. 
+        /// During the CosmosClient initialization the account information, including the available regions, is obtained from the <see cref="CosmosClient.Endpoint"/>.
+        /// Should the global endpoint become inaccessible, the CosmosClient will attempt to obtain the account information issuing requests to the regional endpoints
+        /// provided in the regionalEndpoints set.
+        /// </summary>
+        /// <param name="regionalEndpoints">A set of string containing the regional private endpoints for the cosmos db account.</param>
+        /// <remarks>
+        ///  This function is optional and is recommended for implementation when a customer has configured one or more private endpoints for their Cosmos DB account.
```
FabianMeiswinkel, line 312: Dito - note of custom DNS records

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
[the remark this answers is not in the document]
FabianMeiswinkel: I discussed this with Debdatta oflline - I am ok with either using GlobalEndpoint, RegionalEndpoints, GeneratedEndpointsBasedoOnEffectivePreferredRegions - or if (!RegionalEndpoints.IsEmpty) { Global endpoints + regional Endpoints) else { global endpoints + generated Regional Endpoints based on effective preferred regions)

FabianMeiswinkel (changes requested):
LGTM except for the one open issue around picking the right set of endpoints to reach out to, to get account info

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: The original purpose of using `ISet` was avoid confusion on the ordering. However, this can be replaced by the `IEnumerable` as well. Therefore refactored the code to reflect the same.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Updated the verbiage.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
kundadebdatta: Updated the verbiage.

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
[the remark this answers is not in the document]
kundadebdatta: Refactored the code based on the above suggestions. Per my conversation with @FabianMeiswinkel, the SDK should first attempt to fetch the account information from the global and regional endpoints. If the attempts are un-successful, the SDK will then append the preferred regions with the default endpoint and re-attempt to fetch the account information.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -163,30 +185,40 @@ public async Task<AccountProperties> GetAccountPropertiesAsync()
...
 
                 // Start 2 additional tasks to try to get the account information
                 // from the preferred region list since global account has not succeed yet.
-                HashSet<Task> tasksToWaitOn = new HashSet<Task>
-                {
-                    globalEndpointTask,
-                    this.TryGetAccountPropertiesFromAllLocationsAsync(),
-                    this.TryGetAccountPropertiesFromAllLocationsAsync()
-                };
+                tasksToWaitOn.Add(this.TryGetAccountPropertiesFromAllLocationsAsync());
```
FabianMeiswinkel, line 220: Sorry - I do not get this. Why start the same task twice? - I know it was done before - still don't understand why?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -190,6 +190,38 @@ public string ApplicationName
...
+        /// {
+        ///     RegionalEndpoints = new HashSet<string>(){ "custom.p-1.documents.azure.com", "custom.p-2.documents.azure.com" }
+        /// };
+        /// 
+        /// CosmosClient client = new CosmosClient("endpoint", "key", clientOptions);
+        /// ]]>
+        /// </code>
+        /// </example>
+        /// <seealso href="https://docs.microsoft.com/azure/cosmos-db/high-availability#high-availability-with-cosmos-db-in-the-event-of-regional-outages">High availability on regional outages</seealso>
+        public IEnumerable<string> RegionalEndpoints { get; set; }
```
ealsur, line 224: Why is it called RegionalEndpoints? How do we know these are pointing to a Regional endpoint? Or are they "Custom DNS endpoints"? If these were RegionalEndpoints then they should have a region attached, for exmaple:

`RegionalEndpoints = { (WestUS, "http://customdns", EastUS, "http://customdns2")};`

kundadebdatta: As discussed offline, I have updated the API naming to `AccountInitializationCustomEndpoints`.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Agreed, No clue why was it added twice. IMO, keeping any one of these would suffice. If we all agree, I will remove the other duplicate task. Thoughts @ealsur , @kirankumarkolli ?

ealsur: The reasoning I believe is because this uses an Enumerator as input, so it starts 2 parallel loops that drain from the same enumerator https://github.com/Azure/azure-cosmos-dotnet-v3/pull/2355

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
ealsur: Please update the PR title to remove "Regional"

kundadebdatta: Updated the title and the description.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Added relevant description to the API doc.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -191,6 +191,38 @@ public string ApplicationName
         /// </example>
         /// <seealso href="https://docs.microsoft.com/azure/cosmos-db/high-availability#high-availability-with-cosmos-db-in-the-event-of-regional-outages">High availability on regional outages</seealso>
         public IReadOnlyList<string> ApplicationPreferredRegions { get; set; }
+
+        /// <summary>
+        /// Gets and sets the custom private endpoints for geo-replicated database accounts in the Azure Cosmos DB service.
```
kirankumarkolli, line 196:
```suggestion
        /// Gets and sets the custom endpoints to use for account initialization for geo-replicated database accounts in the Azure Cosmos DB service. 
```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -191,6 +191,38 @@ public string ApplicationName
...
+        /// <summary>
+        /// Gets and sets the custom private endpoints for geo-replicated database accounts in the Azure Cosmos DB service. 
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// During the CosmosClient initialization the account information, including the available regions, is obtained from the <see cref="CosmosClient.Endpoint"/>.
+        /// Should the global endpoint become inaccessible, the CosmosClient will attempt to obtain the account information issuing requests to the custom endpoints provided in <see cref="AccountInitializationCustomEndpoints"/>.
+        /// </para>
+        /// <para>
+        /// Nevertheless, this parameter remains optional and is recommended for implementation when a customer has configured a private endpoint with a custom DNS hostname
```
kirankumarkolli, line 204:
```suggestion
        /// Nevertheless, this parameter remains optional and is recommended for implementation when a customer has configured a endpoint with a custom DNS hostname

```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -308,6 +308,37 @@ public CosmosClientBuilder WithApplicationPreferredRegions(IReadOnlyList<string>
             return this;
         }
 
+        /// <summary>
```
kirankumarkolli, line 311: Replicate any documentation changes here too

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
```diff
@@ -89,6 +88,7 @@ public async Task<AccountProperties> InitializeReaderAsync()
             AccountProperties databaseAccount = await GlobalEndpointManager.GetDatabaseAccountFromAnyLocationsAsync(
                 defaultEndpoint: this.serviceEndpoint,
                 locations: this.connectionPolicy.PreferredLocations,
+                customPrivateEndpoints: this.connectionPolicy.CustomPrivateEndpoints,
```
kirankumarkolli, line 91: Naming: consistent across

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -163,30 +185,40 @@ public async Task<AccountProperties> GetAccountPropertiesAsync()
                 {
```
kirankumarkolli, line 185: This needs to also cover for customendpoints right?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -146,13 +166,15 @@ private class GetAccountPropertiesHelper : IDisposable
 
             public GetAccountPropertiesHelper(
                 Uri defaultEndpoint,
-                IEnumerator<string>? locations,
+                IEnumerator<string>? locations,
```
kirankumarkolli, line 169: Conceptually do two concepts needed? 

I.e. even given the locations `LocationHelper.GetLocationEndpoint(this.DefaultEndpoint, location)` can be used to pre-populate the endpoints right? It can even be made lazy like current enumeration. 

This way the code flow gets simpler right?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kirankumarkolli: Yup! faster enumeration was the idea. Jake did it long time back, there were pathologicalses where it ends-up doing all regions and if CX regions is at end.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kirankumarkolli: Am waiting on this.

kirankumarkolli (changes requested):
https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4265/files#r1468064408

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: The `GetAccountPropertiesHelper` constructor is already capturing the custom endpoints within `serviceEndpointEnumerator`. 

The method `GetAccountPropertiesAsync()` will iterate through the `this.ServiceEndpointEnumerator` which already contains the custom endpoints. So, we should be covered.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: @FabianMeiswinkel and I had an offline conversation around this. Ideally the SDK should attempt to reach out to all the endpoints (global + custom private) before appending the location names to the default global endpoint. So, for example please consider the following code snippet:

```
{
    ApplicationPreferredRegions = new List<string>()
    {
        Regions.EastUs,
        Regions.WestUs,
    },
    RegionalEndpoints = new List<string>()
    {
        { "custom.p-1.documents.azure.com" },
        { "custom.p-2.documents.azure.com" },
    },    
    EnablePartitionLevelFailover = true,
    ConnectionMode = ConnectionMode.Direct,
};

    "https://testaccount.documents-test.windows-int.net:443/",
    "key==",
    clientOptions
);
```

The account endpoints should be resolved with the following order:

1. **First Attempt:** https://testaccount.documents-test.windows-int.net:443
2. **Second Attempt:** https://custom.p-1.documents.azure.com:443 
3. **Third Attempt:** https://custom.p-2.documents.azure.com:443
4. **Fourth Attempt:** https://testaccount-EastUs.documents-test.windows-int.net:443
5. **Fifth Attempt:** https://testaccount-WestUs.documents-test.windows-int.net:443

The purpose of the `LocationHelper.GetLocationEndpoint(this.DefaultEndpoint, location)` is to append the region names to the default endpoint (basically see the `4` and `5` steps in the above example) and return the same.

Therefore, I would be more inclined to keep the implementation as is - to define the scope and purposes of the respective methods.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -163,30 +185,40 @@ public async Task<AccountProperties> GetAccountPropertiesAsync()
...
+                // information.
+                HashSet<Task> tasksToWaitOn = new ();
+                while (this.ServiceEndpointEnumerator.MoveNext())
+                {
+                    Task serviceEndpointTask = this.GetAndUpdateAccountPropertiesAsync(
+                        endpoint: this.ServiceEndpointEnumerator.Current);
+
+                    // Start a timer to start secondary requests in parallel.
+                    Task timerTask = Task.Delay(TimeSpan.FromSeconds(5));
+                    await Task.WhenAny(serviceEndpointTask, timerTask);
```
xinlian12, line 202: since we are waiting only on the new tasks, which means we will be unaware of result of previous tasks until one of the two new tasks completes. But I think it will be fine as only max 5s latency will be added here?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -163,30 +185,40 @@ public async Task<AccountProperties> GetAccountPropertiesAsync()
...
+                // as a suffix to the default global endpoint, and try to retrieve the account
+                // information.
+                HashSet<Task> tasksToWaitOn = new ();
+                while (this.ServiceEndpointEnumerator.MoveNext())
+                {
+                    Task serviceEndpointTask = this.GetAndUpdateAccountPropertiesAsync(
+                        endpoint: this.ServiceEndpointEnumerator.Current);
+
+                    // Start a timer to start secondary requests in parallel.
+                    Task timerTask = Task.Delay(TimeSpan.FromSeconds(5));
```
ealsur, line 201: When looking at the original code, the 5 second delay was only for the main endpoint. Only if the main endpoint took longer than 5 seconds we would start the loop. 

Adding the 5 seconds on each iteration, would that be desired?

What if the logic remains similar as before?

1. Use `this.DefaultEndpoint` + 5 seconds
2. If 5 seconds fired, then do a loop on Custom Endpoints + Preferred Locations as a single list unit. The single list can have first the Custom Endpoints and then the Preferred Regions.

That way the logic remains the same, you are just adding Custom Endpoints to the list before Preferred Regions?

kirankumarkolli: Yup that's the intention of omment: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4265#discussion_r1468064408

kundadebdatta: Refactored the code as discussed.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Refactored the code as discussed @kirankumarkolli.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/GatewayAccountReader.cs
kundadebdatta: Refactored across.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Right. However, the code has been refactored recently and the `5s` latency is applicable only for the global endpoint task.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -116,13 +116,15 @@ public Uri GetHubUri()
         /// </summary>
         public static async Task<AccountProperties> GetDatabaseAccountFromAnyLocationsAsync(
             Uri defaultEndpoint,
-            IList<string>? locations,
+            IList<string>? locations,
+            IList<string> accountInitializationCustomEndpoints,
```
ealsur, line 120: Should this be:

```suggestion
            IList<string>? accountInitializationCustomEndpoints,
```

or are we creating a default empty list?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -163,29 +171,29 @@ public async Task<AccountProperties> GetAccountPropertiesAsync()
...
-
-                // Start 2 additional tasks to try to get the account information
-                // from the preferred region list since global account has not succeed yet.
-                HashSet<Task> tasksToWaitOn = new HashSet<Task>
-                {
-                    globalEndpointTask,
-                    this.TryGetAccountPropertiesFromAllLocationsAsync(),
-                    this.TryGetAccountPropertiesFromAllLocationsAsync()
+
+                Task globalEndpointTask = this.GetAndUpdateAccountPropertiesAsync(this.DefaultEndpoint);
```
ealsur, line 175: Is this code changing in spacing? It shows all this changed but really looking at it, it's not

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -254,49 +262,23 @@ private async Task<AccountProperties> GetOnlyGlobalEndpointAsync()
...
+                    return;
+                }
+
+                // We first iterate through all the private endpoints to fetch the account information.
+                // If all the attempt fails to fetch the metadata from the private endpoints, we will
+                // attempt to retrieve the account information from the regional endpoints constructed
+                // using the preferred regions list.
+                while (this.ServiceEndpointEnumerator.MoveNext())
+                {
+                    await this.GetAndUpdateAccountPropertiesAsync(
```
ealsur, line 277: Previous code had a `lock` on the `MoveNext` because this is executed from 2 concurrent Tasks. Should this maintain the thread-safety?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -349,6 +331,49 @@ private static bool IsNonRetriableException(Exception exception)
...
+            /// Returns an enumerator containing the private and regional service endpoints to iterate over.
+            /// </summary>
+            /// <param name="defaultEndpoint">An instance of <see cref="Uri"/> containing the default global endpoint.</param>
+            /// <param name="locations">An instance of <see cref="IList{T}"/> containing the preferred location names.</param>
+            /// <param name="accountInitializationCustomEndpoints">An instance of <see cref="IList{T}"/> containing the custom private endpoints.</param>
+            /// <returns>An instance of <see cref="IEnumerator{T}"/> containing the service endpoints.</returns>
+            private static IEnumerator<Uri> GetServiceEndpointEnumerator(
+                Uri defaultEndpoint,
+                IList<string>? locations,
+                IList<string> accountInitializationCustomEndpoints)
```
ealsur, line 344: If we compare this with `null`, then should this be 
```suggestion
                IList<string>? accountInitializationCustomEndpoints)

```

And use instead `HasValue`?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Yes, the list is created as an empty collection by default. Here is [the code](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/users/kundadebdatta/4236_add_custom_domain_names/Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs#L47) which does the initialization.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -349,6 +331,49 @@ private static bool IsNonRetriableException(Exception exception)
...
+                        // Add all of the custom private endpoints to the service endpoints list.
+                        serviceEndpoints.Add(
+                            new Uri(customEndpoint));
+                    }
+                }
+
+                // The next step is to construct and add all of the regional endpoints to the service endpoints list.
+                // The regional endpoints will be created by appending the preferred region name as a suffix to the
+                // default global endpoint.
+                if (locations != null
```
ealsur, line 363: Since it's nullable, should we use `HasValue` instead?
```suggestion
                if (locations.HasValue && locations.Value.Count > 0

```

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -137,7 +139,8 @@ private class GetAccountPropertiesHelper : IDisposable
         {
             private readonly CancellationTokenSource CancellationTokenSource;
             private readonly Uri DefaultEndpoint;
-            private readonly IEnumerator<string>? Locations;
+            private readonly IEnumerator<string>? Locations;
+            private readonly IEnumerator<Uri> ServiceEndpointEnumerator;
```
kirankumarkolli, line 143:
```suggestion
            private readonly IEnumerator<Uri> ServiceDefaultEndpointEnumerator;
```

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -349,6 +331,49 @@ private static bool IsNonRetriableException(Exception exception)
...
+                // We first iterate through all the private endpoints and add them to the service endpoints list.
+                IList<Uri> serviceEndpoints = new List<Uri>();
+
+                if (accountInitializationCustomEndpoints != null
+                    && accountInitializationCustomEndpoints.Count > 0)
+                {
+                    foreach (string customEndpoint in accountInitializationCustomEndpoints)
+                    {
+                        // Add all of the custom private endpoints to the service endpoints list.
+                        serviceEndpoints.Add(
```
kirankumarkolli, line 355: For public constract, how about taking it as URI? 
That way some kinds of mis-configurations will be covered by the contract it-self?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -349,6 +331,49 @@ private static bool IsNonRetriableException(Exception exception)
...
+            /// <param name="locations">An instance of <see cref="IList{T}"/> containing the preferred location names.</param>
+            /// <param name="accountInitializationCustomEndpoints">An instance of <see cref="IList{T}"/> containing the custom private endpoints.</param>
+            /// <returns>An instance of <see cref="IEnumerator{T}"/> containing the service endpoints.</returns>
+            private static IEnumerator<Uri> GetServiceEndpointEnumerator(
+                Uri defaultEndpoint,
+                IList<string>? locations,
+                IList<string> accountInitializationCustomEndpoints)
+            {
+                // We first iterate through all the private endpoints and add them to the service endpoints list.
+                IList<Uri> serviceEndpoints = new List<Uri>();
```
kirankumarkolli, line 347: Contract is IEnumrable, so we can do `yield return` for every element. 
It avoids the list materialization unnecessarily in most common cases.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -146,14 +149,19 @@ private class GetAccountPropertiesHelper : IDisposable
...
+                this.Locations = locations?.GetEnumerator();
+                this.GetDatabaseAccountFn = getDatabaseAccountFn;
+                this.CancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+                this.ServiceEndpointEnumerator = GetAccountPropertiesHelper.GetServiceEndpointEnumerator(
+                    defaultEndpoint,
+                    locations,
+                    accountInitializationCustomEndpoints);
             }
 
             public async Task<AccountProperties> GetAccountPropertiesAsync()
```
kirankumarkolli, line 167: if (this.Locations == null)
                {
                    return await this.GetOnlyGlobalEndpointAsync();
                }

Also consider he custom endpints?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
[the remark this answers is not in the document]
kirankumarkolli: Also interesting this is cancellation is not wired through inside `getDatabaseAccountFn`, can be an indepdnent follow-up.

kirankumarkolli: Lock is a hacky way to serialize execution and generate a Task for concurrent execution. => Lock is still needed.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -254,49 +262,23 @@ private async Task<AccountProperties> GetOnlyGlobalEndpointAsync()
...
                 }
             }
 
-            private Task TryGetAccountPropertiesFromRegionalEndpointsAsync(string location)
-            {
-                return this.GetAndUpdateAccountPropertiesAsync(
-                    LocationHelper.GetLocationEndpoint(this.DefaultEndpoint, location));
-            }
-
             private async Task GetAndUpdateAccountPropertiesAsync(Uri endpoint)
```
kirankumarkolli, line 282: > this.TransientExceptions.Add(new OperationCanceledException("GlobalEndpointManager: Get account information canceled"));

Include the URI for context

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
ealsur: But could the user assign it to `null`?

kundadebdatta: Yes, the Cx can assign it to `null`. However the place where we set this value has [already has a null check](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/b8788d409c64fa2c9a4d3dd33afc734c7f273fc0/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs#L840C22-L843) which should cover us.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Refactored the code to add the locking mechanism for thread safety.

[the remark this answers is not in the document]
kundadebdatta: Refactored. For the follow up part, will create a separate discussion thread/ task.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Refactored.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: I have refactored the code, to change the logic a little bit. I now introduced a Boolean flag `this.ShouldVisitGlobalEndpointOnly = (locations == null || locations.Count == 0) && (accountInitializationCustomEndpoints == null || accountInitializationCustomEndpoints.Count == 0);` to visit the global endpoint specifically, when the locations and private endpoints are empty.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Updated the public contract to reflect this.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Looks like, even the nullable collections `<IList<string>? list` doesn't have a definition of `HasValue`. 

image

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Looks like, even the nullable collections `<IList<string>? list` doesn't have a definition of `HasValue`. 

image

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -146,24 +149,31 @@ private class GetAccountPropertiesHelper : IDisposable
...
+                IList<string>? locations,
+                IList<Uri>? accountInitializationCustomEndpoints,
                 Func<Uri, Task<AccountProperties>> getDatabaseAccountFn,
                 CancellationToken cancellationToken)
             {
                 this.DefaultEndpoint = defaultEndpoint;
-                this.Locations = locations;
-                this.GetDatabaseAccountFn = getDatabaseAccountFn;
-                this.CancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+                this.ShouldVisitGlobalEndpointOnly = (locations == null || locations.Count == 0) && (accountInitializationCustomEndpoints == null || accountInitializationCustomEndpoints.Count == 0);
```
kirankumarkolli, line 158:
```suggestion
                this.LimitToGlobalEndpointOnly = (locations == null || locations.Count == 0) && (accountInitializationCustomEndpoints == null || accountInitializationCustomEndpoints.Count == 0);
```

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -219,9 +229,9 @@ public async Task<AccountProperties> GetAccountPropertiesAsync()
 
             private async Task<AccountProperties> GetOnlyGlobalEndpointAsync()
             {
-                if (this.Locations != null)
+                if (!this.ShouldVisitGlobalEndpointOnly)
```
kirankumarkolli, line 232: nit: Unsure the value of this check here

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Refactored to reflect the name change.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Agree. But, I feel we should still keep it, because this method might become useful in some other places. With that, the validation might be helpful for any other caller in the future, to understand the purpose of the method.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -250,52 +260,58 @@ private async Task<AccountProperties> GetOnlyGlobalEndpointAsync()
...
+            /// <summary>
+            /// We first iterate through all the private endpoints to fetch the account information.
+            /// If all the attempt fails to fetch the metadata from the private endpoints, we will
+            /// attempt to retrieve the account information from the regional endpoints constructed
+            /// using the preferred regions list.
+            /// </summary>
+            /// <param name="serviceEndpoint">An instance of <see cref="Uri"/> that will contain the service endpoint.</param>
+            /// <returns>A boolean flag indicating if the <see cref="ServiceEndpointEnumerator"/> was advanced in a thread safe manner.</returns>
+            private bool TryMoveNextServiceEndpointhreadSafe(
+                out Uri? serviceEndpoint)
```
kirankumarkolli, line 295: Nullable practically is for value types not reference types. 
I.e. Uri can be already assigned NULL what's the value of marking it nullable?

Gues this is applicable across all contracts.

kundadebdatta: Had a offline sync up with @kirankumarkolli. This is happening because of the way our project is set up. All project templates starting with .NET 6 (C# 10) enable the nullable context for the project. Projects created with earlier templates don't include this element, and these features are off unless you enable them in the project file or use pragmas.

See [this document ](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types)for more details.

image

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
```diff
@@ -90,6 +92,27 @@ public void SetPreferredLocations(IReadOnlyList<string> regions)
...
+        /// <summary>
+        /// Sets the custom private endpoints required to fetch account information from
+        /// private domain names.
+        /// </summary>
+        /// <param name="customEndpoints">An instance of <see cref="IEnumerable{T}"/> containing the custom DNS endpoints
+        /// provided by the customer.</param>
+        public void SetAccountInitializationCustomEndpoints(
+            IEnumerable<Uri> customEndpoints)
+        {
+            if (customEndpoints == null)
```
kirankumarkolli, line 104: Please also validate for the relative URLs?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -250,52 +260,58 @@ private async Task<AccountProperties> GetOnlyGlobalEndpointAsync()
...
+                }
+
+                while (this.TryMoveNextServiceEndpointhreadSafe(
+                        out Uri? serviceEndpoint))
                 {
-                    if (location == null)
+                    if (serviceEndpoint == null)
                     {
-                        DefaultTrace.TraceCritical("GlobalEndpointManager: location is null for TryMoveNextLocationThreadSafe");
+                        DefaultTrace.TraceCritical("GlobalEndpointManager: serviceEndpoint is null for TryMoveNextServiceEndpointhreadSafe.");
```
kirankumarkolli, line 277: Unrelated: How about include the regions attempted or tried.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -250,52 +260,58 @@ private async Task<AccountProperties> GetOnlyGlobalEndpointAsync()
...
-            /// This is done in a thread safe way to allow multiple tasks to iterate over the 
-            /// list of locations.
+            /// This is done in a thread safe way to allow multiple tasks to iterate over the list of service endpoints.
             /// </summary>
             private async Task TryGetAccountPropertiesFromAllLocationsAsync()
-            {
-                while (this.TryMoveNextLocationThreadSafe(
-                        out string? location))
+            {
+                if (this.CancellationTokenSource.IsCancellationRequested)
```
kirankumarkolli, line 267: nit: Redundant check, while condition is already covering it.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
```diff
@@ -348,7 +363,8 @@ public void CosmosClientOptions_WhenPartitionLevelFailoverEnabledAndPreferredReg
                 Assert.IsFalse(clientOptions.AllowBulkExecution);
                 Assert.AreEqual(consistencyLevel, clientOptions.ConsistencyLevel);
                 Assert.IsTrue(clientOptions.EnablePartitionLevelFailover);
```
kirankumarkolli, line 365: Assert that ConnectionPolicy is set?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
```diff
@@ -348,7 +363,8 @@ public void CosmosClientOptions_WhenPartitionLevelFailoverEnabledAndPreferredReg
...
-                Assert.IsNotNull(clientOptions.ApplicationPreferredRegions);
+                Assert.IsNotNull(clientOptions.ApplicationPreferredRegions);
```
kirankumarkolli, line 366: Please check that the endpoints are visited.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Removed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
kundadebdatta: Done.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
kundadebdatta: Updated the test to add explicit validation.

ealsur (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayAccountReaderTests.cs
```diff
@@ -100,6 +101,157 @@ public void DocumentClient_BuildHttpClientFactory_WithFactory()
...
+                httpClient: mockHttpClient.Object);
+
+            if (customEndpointsProvided)
+            {
+                AccountProperties accountProperties = await accountReader.InitializeReaderAsync();
+
+                Assert.IsNotNull(accountProperties);
+                Assert.AreEqual("localhost", accountProperties.Id);
+                Assert.AreEqual("127.0.0.1", accountProperties.ResourceId);
+                Assert.AreEqual(endpointSucceeded, privateEndpoint3);
```
kirankumarkolli, line 187: Moq has inbuilt way to verify the callers. 
Either it can be very explicit at argument level or aggregate count level. 

For example below code is checking for aggregate times its expected to get called.
```
            mockHttpClient.Verify(e => e.GetAsync(It.IsAny(), ...), Times.Exactly(3));
```

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayAccountReaderTests.cs
kundadebdatta: `mockHttpClient.Verify` can validate how many times the `mockHttpClient.GetAsync()` was invoked, and irrespective of the response type, in this case, the method will be invoked `4` times (because we have 4 service endpoints and each endpoint will invoke the method). 

However, our purpose is to validate the `privateEndpoint3` was indeed returned a valid response, which I believe can't be validated with `mockHttpClient.Verify()`.

Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
kundadebdatta: Will follow up as a separate task.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
kundadebdatta: Will follow up as a separate PR.

FabianMeiswinkel (approved):
LGTM - Thanks

PR #3401, opened 2022-08-22
Change Feed Processor: Fixes LeaseLostException leaks on notification APIs

ealsur:
## Description

There were 3 conditions on which the Notification APIs were surfacing `LeaseLostException` to the user:

1. If the lease is not found (the document was deleted, which would indicate an invalid state): https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseUpdaterCosmos.cs#L66-L70
2. If there are constant conflicts beyond the retry scope (basically getting 412, which would indicate a very high concurrency of workers) https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseUpdaterCosmos.cs#L35
3. When a worker processed a batch and was about to checkpoint but the lease was stolen (https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs#L92) which would indicate that a batch will probably be reprocessed by another instance. 

This PR makes it so the notification APIs are called but with the InnerException related to the CosmosException (a NotFound, or a Conflict, or a PreconditionFailed error) instead of the LeaseLostException.

This way we avoid leaking the internal type but we don't miss notifying the problem.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Closing issues

Closes #3379

PR #4276, opened 2024-01-24
Change Feed Processor: Fixes LeaseLostException on Notifications API for Renewer

ealsur:
#3401 fixed all the scenarios where the LeaseLostException is generated based on a logical condition (not network request) to contain a CosmosException that is reportable to the Notifications API.

The PartitionSupervisor was modified in that same PR to only report the inner CosmosException to the Notifications API.

However, there was a missing case:

The PartitionController starts a PartitionSupervisor once acquiring a lease, which spans a PartitionProcessor and Renewer.

The Renewer can hit a LeaseLostException if the PartitionProcessor is not processing (idle partition) but being renewed, in that case, the LeaseLostException goes up, stops the PartitionSupervisor as expected, and gets bubbled up to the PartitionController, which releases the lease.

The problem was that the PartitionController calls the Notification APIs to report Errors without identifying that this can only be done for LeaseLostExceptions when they have a linked inner exception (CosmosException). This handling was previously added to other flows (lease acquiring, lease release) but not on the renewal.

## Symptoms

This was reported in Azure Functions, where the Functions Extension has a switch statement to log CosmosExceptions:

https://github.com/Azure/azure-webjobs-sdk-extensions/blob/cf0fc8022b230aa1e19325638881a56210474189/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerHealthMonitor.cs#L23-L37

But it is receiving the OuterType as LeaseLostException:

image

## Verification

With this fix, the emitted Notifications/Logs, have the proper Severity Level (Information) and the proper Type (CosmosException):

image


FabianMeiswinkel (approved):
LGTM - Thanks

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionControllerCore.cs
```diff
@@ -161,6 +161,17 @@ private async Task ProcessPartitionAsync(DocumentServiceLease lease)
             {
                 DefaultTrace.TraceVerbose("Lease with token {0}: processing canceled", lease.CurrentLeaseToken);
             }
+            catch (LeaseLostException leaseLostException)
+            {
+                // LeaseLostException by itself is not loggable, unless it contains a related inner exception
```
philipthomas-MSFT, line 166: "LeaseLostException by itself is not loggable"

Can you clarify for my growth and understanding. What does that mean?

ealsur: LeaseLostException is a signal that CFP components use to signal each other. The PartitionController signals the PartitionSupervisor that the current lease was lost (acquired by another instance), the upper components catch and handle these signals.

In this case, we were logging the LeaseLostException publicly, which is an internal type, and that is incorrect.

philipthomas-MSFT (approved):

Issue #3819, opened 2023-04-19
Transportation Chaos Simulation - Fault Injection


NaluTripician:
# Fault Injection 

# Background
Currently the SDK does not have a way to simulate network issues. This makes it difficult to test the SDK's behavior when there are network issues. The goal of this feature is to allow customers to configure rules that specify how and when the SDK fails requests. This will allow customers to test the SDK's behavior when there are network issues. Additionally, this will help customers test their own retry policies.

# Design 

## `FaultInjectionRule`

To induce faults, we will introduce a new type `FaultInjectionRule`. This type will allow customers to configure how the SDK fails requests. Once created, `FaultInjectionRule`s can be added to specific containers. 

The `FaultInjectionRule` has three major components, `result`, `condition`, and 

### `result`

The `result` component of the `FaultInjectionRule` specifies what the result of the fault that is to be injected will be. The `result` component can be one of two types, `FaultInjectionServerErrorResult` or `FaultInjectionConnectionErrorResult`.

#### `FaultInjectionServerErrorResult`

Rule will return a server error to the customer. `FaultInjectionServerErrorResult` has three properties: `errorType`, `times`, and `delay`.

| Property | Description |
| -------- | ----------- |
| `errorType` | Required. The type of server error that the SDK will return to the customer. |
| `times` | An optional parameter that limits how many times the same rule can be applied on the operation. If `times` is not specified, the rule will be applied on the operation until the rule is removed or disabled from the container.
| `delay` | Required for `RESPONSE_DELAY` and `CONNECTION_DELAY` error types. The delay is the amount of time the SDK will wait before returning the error to the customer. |

##### **Supported server error types**

```
GONE
RETRY_WITH
INTERNAL_SERVER_ERROR
TOO_MANY_REQUESTS
READ_SESSION_NOT_AVAILABLE
TIMEOUT
PARTITION_IS_MIGRATING
RESPONSE_DELAY // used to simulate transient timeout/broken connections
CONNECTION_DELAY // used to simulate high channel acquisition.
```

#### `FaultInjectionConnectionErrorResult`

Rule will return a connection error to the customer. `FaultInjectionConnectionErrorResult` has three properties: `errorType`, `interval`, and `threshold`.

| Property | Description |
| -------- | ----------- |
| `errorType` | Required. The type of connection error that the SDK will return to the customer. |
| `interval` | Required. The interval is the amount of time the SDK will wait before returning the error to the customer. |
[...]

PR #4281, opened 2024-01-25
FaultInjection: Adds Fault Injection Library

### LINKED ISSUE #4257: Create Fault Injection Library

NaluTripician:
# Pull Request Template

## Description

This PR adds the Fault Injection Library which is a testing library that allows users to simulate network issues with applications using the .NET SDK. To do this, users will create FaultInjection Rules then give those rules to the Cosmos Client upon creation. Fault Injection Rules have two parts, FaultInjectionConditions, which specify when a rule should be applied; and FaultInjectionResults, which tells the SDK what types of errors to inject. This will be released as it's own library and can be used internally to test the SDK, and externally to test CosmosDB applications. For more information see #3819 

How to add a Fault Injection Rule: 

```csharp

//Create Rule With Condition and Result
string ruleId = "rule_id";
FaultInjectionCondition faultInjectionCondition = new FaultInjectionConditionBuilder()
    .WithOperationType(FaultInjectionOperationType.CreateItem)
    .WithConnectionType(FaultInjectionConnectionType.Direct)
    .WithRegion("East US")
    .WithEndpoint(
        new FaultInjectionEndpointBuilder(
            databaseName: "db", 
            containerName: "col",
            feedRange: FeedRange.FromPartitionKey(new PartitionKey("test")))
        .WithReplicaCount(3)
        .WithIncludePrimary(true)
        .Build())
    .Build();
FaultInjectionRule faultInjectionRule = new FaultInjectionRuleBuilder(
    id: ruleId,
    condition: faultInjectionCondition,
    result: FaultInjectionResultBuilder.GetResultBuilder(FaultInjectionServerErrorType.ConnectionDelay)
        .WithDelay(TimeSpan.FromSeconds(6))
        .WithTimes(1)
        .WithSuppressServiceRequest(true)
        .Build()
    )
    .WithDuration(TimeSpan.FromSeconds(10))
    .WithStartDelay(TimeSpan.FromSeconds(5))
    .WithHitLimit(3)
    .Build();

//Create Fault Injector
List<FaultInjectionRule> faultInjectionRules = new List<FaultInjectionRules. { faultInjectionRule};
FaultInjector faultInjector = new FaultInjector(faultInjectionRules);

```

Once you have a `FaultInjector` with all your fault injection rules, you must add it to the client upon creation. 

```csharp
   "endpoint"
   "key"
   FaultInjector.GetFaultInjectionClientOptions(yourCosmosClientOptions));

```

## Type of change

## Closing issues

To automatically close an issue: closes #4257


Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/FaultInjection.csproj
```diff
@@ -0,0 +1,33 @@
...
+		<OutputType>Exe</OutputType>
+		<TargetFramework>net6</TargetFramework>
+		<RootNamespace>Microsoft.Azure.Cosmos.FaultInjection</RootNamespace>
+		<AssemblyName>Microsoft.Azure.Cosmos.FaultInjection</AssemblyName>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>enable</Nullable>
+	</PropertyGroup>
+
+	<ItemGroup>
+		<ProjectReference Include="..\..\..\Microsoft.Azure.Cosmos\src\Microsoft.Azure.Cosmos.csproj" />
```
ealsur, line 19: If we plan to ship this as a Nuget package, it cannot have a project reference. This is the same scenario as the Encryption package?

NaluTripician: Will switch before merge, going to keep this temporarily for testing purposes.

kirankumarkolli: We can also use ProjRef hint like direct packages. Its little incovenience but right thing.
Optimizing for current iteration velocity makes sense.

This change is needed before NuGet package release time.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionCondition.cs
```diff
@@ -0,0 +1,89 @@
...
+            this.connectionType = connectionType;
+            this.region = region;
+            this.endpoint = endpoint;
+        }
+
+        /// <summary>
+        /// The operation type the rule will target.
+        /// </summary>
+        /// <returns>the <see cref="FaultInjectionOperationType"/>.</returns>
+        public FaultInjectionOperationType? GetOperationType()
```
kirankumarkolli, line 43: Constructor getting the value right? Why nullable for return type?

NaluTripician: Fault Injection Rules do not need to specify the Operation Type, will change constructor to reflect this value.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionConnectionErrorResult.cs
```diff
@@ -0,0 +1,73 @@
...
+        public TimeSpan GetTimespan()
+        {
+            return this.interval;
+        }
+
+        /// <summary>
+        /// Returns the percentage of the established connections that will be impacted. By default, the threshold is 1.
+        /// </summary>
+        /// <returns>the threshold represented as a double.</returns>
+        public double GetThreshold()
```
kirankumarkolli, line 55:
```suggestion
        public double GetThresholdPercentage()
```

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjector.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// </summary>
+        /// <param name="activityId"></param>
+        /// <returns>the fault injection rule id</returns>
+        public string? GetFaultInjectionRuleId(Guid activityId)
+        {
+            return this.chaosInterceptorFactory.ChaosInterceptor?.GetFaultInjectionRuleId(activityId);
+        }
+
+        //Get Application Context
+        public FaultInjectionApplicationContext? GetApplicationContext()
```
kirankumarkolli, line 36: Who are the consumers? Doesit needs to be public?

kirankumarkolli: Same for `GetFaultInjectionRuleId`

NaluTripician: This would allow anyone using Fault Injection to see when fault injection rules were applied and to what requests. The `FaultInjectionApplicationContext` tracks when rules were applied and to what request they were applied to.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            this.applicationsByActivityId.TryAdd(activityId, (DateTime.UtcNow, ruleId));
+            this.values.Add((DateTime.UtcNow, ruleId, activityId));
+        }
+
+        /// <summary>
+        /// Gets all application of fault injection rules by DateTime, RuleId, ActivityId
+        /// </summary>
+        /// <returns><see cref="BlockingCollection{T}"/> of Application Time, RuleId, ActivityId</returns>
+        public BlockingCollection<(DateTime, string, Guid)> GetAllApplications()
```
kirankumarkolli, line 39: Application here is Usage right? 

Application name is overloaded and confusing.

kirankumarkolli (approved):

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
NaluTripician: Yes, here application is when a rule is applied to a request. Would `RuleApplication` make things clearer?

kirankumarkolli: Few more alternative
- GetAllRuleExecutions
- GetAllRuleUsage
- GetAllInterceptorExecutions/Usages

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionCondition.cs
NaluTripician: Yes, if operation type is not specified, the rule will be valid for all operation types.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjection.csproj
```diff
@@ -0,0 +1,41 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<Company>Microsoft Corporation</Company>
+		<Product>Microsoft(R) Azure Cosmos</Product>
+		<Description>This client library enables client applications to inject faults into Azure CosmosD DB applications.</Description>
```
philipthomas-MSFT, line 6: CosmosD DB - fix spelling

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2" + AssemblyKeys.MoqPublicKey)]
```
philipthomas-MSFT, line 7: Why do we append "2" to "DynamicProxyGenAssembly2"? I know this is outside the scope of the PR, but for my growth and understanding (and maybe others), I am trying to understand the reasoning.

NaluTripician: From my understanding, this is required if we ever need to do testing with Moq. This is also present in the AssemblyInfo in the [main SDK](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/83958bb3cee2bb7d0533dc526951d26878b7a261/Microsoft.Azure.Cosmos/src/AssemblyInfo.cs#L7).

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionCondition.cs
```diff
@@ -0,0 +1,87 @@
...
+            return this.connectionType;
+        }
+
+        /// <summary>
+        /// The region the rule will target.
+        /// </summary>
+        /// <returns>the region represented as a string.</returns>
+        public string GetRegion()
+        {
+            return this.region ?? string.Empty;
```
philipthomas-MSFT, line 61: What are the scenarios or use cases where we are ok with returning region as `string.Empty`?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionConditionBuilder.cs
```diff
@@ -0,0 +1,87 @@
...
+            this.region = string.IsNullOrEmpty(regionName) 
+                ? throw new ArgumentNullException(nameof(region), "Argument 'region' cannot be null.") 
+                : regionName;
+
+            return this;
+        }
+
+        /// <summary>
+        /// Optional. Specifies which endpoint the rule will target. Once set, the rule will only target requests targeting that endpoint. 
+        /// Only applicable to direct mode.
```
philipthomas-MSFT, line 62: Can you add a few words, and I do mean few, explaining why this is only applicable to direct mode?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionConnectionErrorType.cs
```diff
@@ -0,0 +1,21 @@
...
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.FaultInjection
+{
+    /// <summary>
+    /// Connection Error Type for Fault Injection
+    /// </summary>
+    public enum FaultInjectionConnectionErrorType
+    {
+        /// <summary>
+        /// Emulates a connection close because of a recieved stream close
```
philipthomas-MSFT, line 12: receive

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionConnectionErrorType.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// </summary>
+        RecievedStreamClosed,
+
+        /// <summary>
+        /// Emulates a connection close because of a failure to recieve a response
```
philipthomas-MSFT, line 17: receive

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionConnectionErrorType.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// </summary>
+        RecieveFailed,
```
philipthomas-MSFT, line 19: receive

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionEndpoint.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    public sealed class FaultInjectionEndpoint
+    {
+        private readonly string databaseName;
+        private readonly string containerName;
+        private readonly FeedRange feedRange;
+        private readonly bool includePrimary;
+        private readonly int replicaCount;
+
+        internal static FaultInjectionEndpoint Empty = new FaultInjectionEndpoint(
```
philipthomas-MSFT, line 20: Is it necessary to instantiate the FaultInjectionEndpoint with arguments that default to empty strings?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionEndpoint.cs
```diff
@@ -0,0 +1,93 @@
...
+        { 
+            this.databaseName = databaseName;
+            this.containerName = containerName;
+            this.feedRange = feedRange;
+            this.includePrimary = includePrimary;
+            this.replicaCount = replicaCount;
+        }
+
+        /// <summary>
+        /// Get the FeedRange
```
philipthomas-MSFT, line 48: Consistent ending of summaries with period (.)

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionEndpoint.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <returns>the <see cref="FeedRange"/></returns>
+        public FeedRange GetFeedRange()
+        {
+            return this.feedRange;
+        }
+
+        /// <summary>
+        /// Get the flag indicating if primary replica address can be used.
+        /// </summary>
+        /// <returns>the flag indicating if a primary replica address can be used.</returns>
```
philipthomas-MSFT, line 59: Consistency beginning any comments with capitalization.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionEndpointBuilder.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <summary>
+        /// Set the replica count of the <see cref="FaultInjectionEndpoint"/>.
+        /// </summary>
+        /// <param name="replicaCount">int representing the replica count.</param>
+        /// <returns>the <see cref="FaultInjectionEndpointBuilder"/>.</returns>
+        public FaultInjectionEndpointBuilder WithReplicaCount(int replicaCount)
+        {
+            if (replicaCount < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(replicaCount), "Argument 'replicaCount' cannot be negative.");
```
philipthomas-MSFT, line 41: Why can it not be negative?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionRuleBuilder.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        /// <summary>
+        /// Creates a new <see cref="FaultInjectionRule"/>.
+        /// </summary>
+        /// <returns>the <see cref="FaultInjectionRule"/>.</returns>
+        public FaultInjectionRule Build()
+        {
+            if (this.result == null)
+            {
+                throw new ArgumentNullException(nameof(this.result), "Argument 'result' cannot be null.");
```
philipthomas-MSFT, line 117: Do we think enough about why any argument, not limited to this PR, could possibly be null? Especially when the null check is not a variable that is also an argument for the method that is throwing the exception?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionCondition.cs
NaluTripician: This would be the case where a fault injection rule is not targeting a specific region, so all regions would be effected by the rule. There are checks in place that help make the rules more efficient by checking if this returns `string.Empty`

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionServerErrorType.cs
```diff
@@ -0,0 +1,62 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.FaultInjection
+{
+    /// <summary>
+    /// Types of ServerErrors that can be injected
+    /// </summary>
+    public enum FaultInjectionServerErrorType
```
philipthomas-MSFT, line 9: Can we somehow, and it is totally possible that the answer is no, but is it possible that reuse existing enums for error types instead of creating our own?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionConditionBuilder.cs
NaluTripician: Adding: `Only applicable to direct mode since only direct mode specifies physical endpoints.`

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        public (DateTime, string)? GetRuleExecutionsByActivityId(Guid activityId)
+        {
+            if (this.executionsByActivityId.TryGetValue(activityId, out (DateTime, string) execution))
+            {
+                return execution;
+            }
+
+            return null;
```
philipthomas-MSFT, line 71: Why do we ever want to return a null?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionEndpoint.cs
NaluTripician: `FaultInjectionEndpoint.Empty` will be used as a default when not specifying an endpoint to filter on. This again helps with rule processing efficiency.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionEndpointBuilder.cs
NaluTripician: You cannot have a negative number of replicas in Cosmos so we are making sure that putting a negative amount here does not break other parts of the code. Zero however is a valid number (it will not filter the rules on any of the replicas).

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionServerErrorType.cs
NaluTripician: The reason why we are not reusing existing enums is because some of the cases here are status codes with existing eums for them, but these status codes are not paired with the sub status codes that the server error types correspond with. Additionally, not all the existing status codes are supported/will be used for Fault Injection so reusing them could run into the case where a user tries to add too add a Fault Injection Result that is not supported.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
NaluTripician: This method would return null in the case where we are trying to see of a particular activityId has a fault injection rule executions are associated with it but no fault injection rules were applied to that activityId.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionRuleBuilder.cs
NaluTripician: To clarify this, I will change the constructor to require the two arguments that the null check is being done on.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjection.csproj
```diff
@@ -0,0 +1,41 @@
...
+		<CurrentDate>$([System.DateTime]::Now.ToString(yyyyMMdd))</CurrentDate>
+		<OutputType>Exe</OutputType>
+		<TargetFramework>net6</TargetFramework>
+		<RootNamespace>Microsoft.Azure.Cosmos.FaultInjection</RootNamespace>
+		<AssemblyName>Microsoft.Azure.Cosmos.FaultInjection</AssemblyName>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>enable</Nullable>
+	</PropertyGroup>
+
+	<ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' != 'True' ">
```
ealsur, line 18: Remove IsPreview usage, unless you expect to ship preview FaultInjection packages

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjection.csproj
```diff
@@ -0,0 +1,41 @@
...
+		<PackageReference Include="Microsoft.Azure.Cosmos" Version="$(ClientOfficialVersion)" />
+	</ItemGroup>
+	
+	<ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' == 'True' ">
+		<PackageReference Include="Microsoft.Azure.Cosmos" Version="$(ClientPreviewVersion)-$(ClientPreviewSuffixVersion)" />
+	</ItemGroup>
+
+	<ItemGroup Condition=" '$(SdkProjectRef)' == 'True' ">
```
ealsur, line 26: Might need to set this on the top PropertyGroup

```
<SdkProjectRef>True</SdkProjectRef>
```

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionCondition.cs
philipthomas-MSFT: Would it be better to set it to some constant string so that we understand the intent?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
philipthomas-MSFT: Can we create something to what you described so that we are clear of the intention instead of just setting it to null?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
```diff
@@ -0,0 +1,181 @@
...
+        private readonly FaultInjectionDynamicChannelStore channelStore;
+        private readonly RegionNameMapper regionNameMapper;
+        private readonly Dictionary<string, string> regionSpecialCases;
+
+        public RntbdConnectionErrorInjector(FaultInjectionRuleStore ruleStore, FaultInjectionDynamicChannelStore channelStore)
+        {
+            this.ruleStore = ruleStore ?? throw new ArgumentNullException(nameof(ruleStore));
+            this.channelStore = channelStore ?? throw new ArgumentNullException(nameof(channelStore));
+            this.regionNameMapper = new RegionNameMapper();
+            this.regionSpecialCases = new Dictionary<string, string>
```
philipthomas-MSFT, line 23: Can you explain the significance of a region special cases?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
```diff
@@ -0,0 +1,181 @@
...
+                                        rule.GetId(), 
+                                        addressUri);
+                                    channel.InjectFaultInjectionConnectionError(this.GetTransportException(connectionErrorType, channel));
+                                }
+                            }));
+
+                            return Task.CompletedTask;
+                        }
+
+                        //Case 2: Inject connection error for all endpoins of one region when there is no specific physical address
```
philipthomas-MSFT, line 78: spelling "endpoints"

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
```diff
@@ -0,0 +1,181 @@
...
+                        this.ruleStore.RemoveRule(rule);
+                    }
+                });
+        }
+
+        private TransportException GetTransportException(FaultInjectionConnectionErrorType errorType, Channel channel)
+        {
+            return errorType switch
+            {
+                FaultInjectionConnectionErrorType.ReceiveStreamClosed => new TransportException(
```
philipthomas-MSFT, line 138: So the only difference between `ReceivesStreamClosed` and `ReceivedFailed` this switch is the `errorCode`? Can we reuse the TransportException?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionUnitTests.cs
```diff
@@ -0,0 +1,34 @@
+namespace Microsoft.Azure.Cosmos.FaultInjection.Tests
+{
+    using System;
+
+    [TestClass]
+    public class FaultInjectionUnitTests
+    {
+        [TestMethod]
```
philipthomas-MSFT, line 8: Can you add [Owner] and [Description] to your tests?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
NaluTripician: The error code is a read-only variable so it can only be assigned when creating the `TransportException`.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionServerErrorRule.cs
```diff
@@ -0,0 +1,205 @@
...
+            this.expireTime = duration == TimeSpan.MaxValue ? DateTime.MaxValue : this.startTime + duration;
+            this.hitCount = 0;
+            this.hitCountDetails = new ConcurrentDictionary<string, long>();
+            this.evaluationCount = 0;
+            this.condition = condition ?? throw new ArgumentNullException(nameof(condition));
+            this.result = result ?? throw new ArgumentNullException(nameof(result));
+            this.connectionType = connectionType;
+        }
+
+        public bool IsApplicable(ChannelCallArguments args)
```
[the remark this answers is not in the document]
NaluTripician: Will add more unit tests for all public methods.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
NaluTripician: Refactored to return false if no value exits with an out variable for when it does to make things more clear that there is no value that exists

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionCondition.cs
NaluTripician: I think here the empty string will be sufficient. This would mean that the rule is not targeting any specific region. I also think that changing the default to something like "ALL" would make things messier when checking the output of this method. I will add information describing the empty string output to the summary of the method to make things more clear for users of the library.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionEndpoint.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <param name="feedRange">The <see cref="FeedRange"/>.</param>
+        /// <param name="includePrimary">Indicates wether primary replica can be used</param>
+        /// <param name="replicaCount">Replica count.</param>
+        public FaultInjectionEndpoint(
+            string databaseName,
+            string containerName,
+            FeedRange feedRange,
+            bool includePrimary,
+            int replicaCount)
+
```
xinlian12, line 38: nit: space

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionEndpoint.cs
```diff
@@ -0,0 +1,93 @@
...
+        private readonly string databaseName;
+        private readonly string containerName;
+        private readonly FeedRange feedRange;
+        private readonly bool includePrimary;
+        private readonly int replicaCount;
+
+        internal static FaultInjectionEndpoint Empty = new FaultInjectionEndpoint(
+            string.Empty, 
+            string.Empty, 
+            new FeedRangePartitionKey(new PartitionKey()), false, 0);
```
xinlian12, line 23: nit: add extra line in between

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionOperationType.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        /// <summary>
+        /// Delete items.
+        /// </summary>
+        DeleteItem,
+
+        /// <summary>
+        /// Patch items.
+        /// </summary>  
+        PatchItem,
```
xinlian12, line 44: also worth adding batch?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/ChaosInterceptor.cs
```diff
@@ -0,0 +1,221 @@
...
+        /// </summary>
+        /// <param name="args"></param>
+        /// <returns></returns>
+        public async Task<(bool, StoreResponse?)> OnRequestCallAsync(ChannelCallArguments args)
+        {
+            StoreResponse faultyResponse;
+            FaultInjectionServerErrorRule? serverResponseErrorRule = this.ruleStore?.FindRntbdServerResponseErrorRule(args);
+            if (serverResponseErrorRule != null)
+            {
+                this.applicationContext.AddRuleExecution(serverResponseErrorRule.GetId(), args.CommonArguments.ActivityId);
```
xinlian12, line 80: is the applicatioContext here used to track for all requests?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
```diff
@@ -0,0 +1,74 @@
+﻿namespace Microsoft.Azure.Cosmos.FaultInjection
```
xinlian12, line 1: copyright

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionConditionInternal.cs
```diff
@@ -0,0 +1,235 @@
...
+                }
+            }
+
+            return true;
+        }
+
+        //Used for connection delay
+        public bool IsApplicable(Uri callUri, DocumentServiceRequest request)
+        {
+            foreach (IFaultInjectionConditionValidator validator in this.validators)
```
xinlian12, line 98: for connection delay scenarios, can we also extract the necessary info out into ChannelCallArguments? (the logic of the two isApplicable seems pretty much the same)

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    using System;
+    using System.Collections.Concurrent;
+
+    public class FaultInjectionApplicationContext
+    { 
+
+        private readonly ConcurrentDictionary<string, List<(DateTime, Guid)>> executionsByRuleId;
```
xinlian12, line 9: Another info I found can be very useful is the rules evaluation result -> like why we can not apply the rules

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionConditionInternal.cs
```diff
@@ -0,0 +1,235 @@
...
+            }
+
+            //Used for Connection Delay
+            public bool IsApplicable(DocumentServiceRequest request)
+            {
+                return request.OperationType == this.operationType;
+            }
+        }
+
+        private class ContainerValidator : IFaultInjectionConditionValidator
```
xinlian12, line 190: in the future, when adding the metadata requests support, will need to also include containerName here as for collectionRead request there is no containerRid available yet.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionServerErrorResultInternal.cs
```diff
@@ -0,0 +1,229 @@
...
+                    {
+                        Status = 500,
+                        Headers = args.RequestHeaders,
+                        ResponseBody = new MemoryStream(Encoding.UTF8.GetBytes($"Fault Injection Server Error: Internal Server Error, rule: {ruleId}"))
+                    };
+                    
+                    return storeResponse;
+
+                case FaultInjectionServerErrorType.ReadSessionNotAvailable:
+                    INameValueCollection readSessionHeaders = args.RequestHeaders;
```
xinlian12, line 186: I think we should an invalid sessionToken or a sessionToken with lower lsn etc. The reason being based on the sessionToken, the retry behavior from SDK will be different. 

in the case when we use the same isn as the send request, SDK will retry on 1 replica(random selected) -> sessionTokenRetryPolicy -> 1 replica(random selected) 
when using an invalid sessionToken or a sessionToken with slower progress -> SDK will retry on 4 replica -> sessionTokenRetryPolicy -> 4 replica. This is usually what happens the most for 404/1002

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionServerErrorResultInternal.cs
```diff
@@ -0,0 +1,229 @@
...
+                    storeResponse = new StoreResponse()
+                    {
+                        Status = 408,
+                        Headers = args.RequestHeaders,
+                        ResponseBody = new MemoryStream(Encoding.UTF8.GetBytes($"Fault Injection Server Error: Timeout, rule: {ruleId}"))
+                    };
+
+                    return storeResponse;
+
+                case FaultInjectionServerErrorType.InternalServerEror:
```
xinlian12, line 175: another error case can be 503

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
```diff
@@ -0,0 +1,183 @@
+﻿namespace Microsoft.Azure.Cosmos.FaultInjection
```
xinlian12, line 1: copyright

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
xinlian12: wow I did not know this, we do not have this mapping in JavaSDK, we just use whatever the addresses being returned

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
```diff
@@ -0,0 +1,183 @@
...
+        public Task InjectConnectionErrorTask(FaultInjectionConnectionErrorRule rule)
+        {
+            TimeSpan delay = rule.GetResult().GetTimespan();
+
+            return Task.Delay(delay).ContinueWith(
+                t =>
+                {
+                    //check to see if rule is valid
+                    if (this.IsEffectiveRule(rule))
+
```
xinlian12, line 55: nit: remove extra line in between

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionDirectModeTests.cs
```diff
@@ -0,0 +1,1201 @@
+﻿namespace Microsoft.Azure.Cosmos.FaultInjection.Tests
```
xinlian12, line 1: copyright

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
```diff
@@ -0,0 +1,183 @@
...
+                            addresses.ForEach(addressUri => allChannels.Where(channel => channel.GetServerUri().Equals(addressUri)).ToList().ForEach(channel =>
+                            {
+                                if (random.NextDouble() < rule.GetResult().GetThresholdPercentage())
+                                {
+                                    rule.ApplyRule();
+                                    DefaultTrace.TraceInformation("FaultInjection: Injecting {0} connection error rule: {1}, for address {2}", 
+                                        connectionErrorType, 
+                                        rule.GetId(), 
+                                        addressUri);
+                                    channel.InjectFaultInjectionConnectionError(this.GetTransportException(connectionErrorType, channel));
```
xinlian12, line 73: quick question: this will cause the channel to be closed correct?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionDirectModeTests.cs
```diff
@@ -0,0 +1,1201 @@
...
+
+                serverGoneRule.Disable();
+
+                if (operationType == OperationType.Delete)
+                {
+                    _ = this.container != null
+                        ? await this.container.CreateItemAsync(item) : null;
+                }
+
+                if (operationType == OperationType.Create)
```
xinlian12, line 166: should this be OperationType.Delete?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionDirectModeTests.cs
```diff
@@ -0,0 +1,1201 @@
...
+
+        [TestMethod]
+        [Owner("nalutripician")]
+        [Description("Tests injecting a server error response")]
+        [DataRow(FaultInjectionOperationType.ReadItem, FaultInjectionServerErrorType.Gone, 410, 21005, DisplayName = "Gone")]
+        [DataRow(FaultInjectionOperationType.ReadItem, FaultInjectionServerErrorType.InternalServerEror, 500, 0, DisplayName = "InternalServerError")]
+        [DataRow(FaultInjectionOperationType.ReadItem, FaultInjectionServerErrorType.RetryWith, 449, 0, DisplayName = "RetryWith")]
+        [DataRow(FaultInjectionOperationType.ReadItem, FaultInjectionServerErrorType.TooManyRequests, 429, 0, DisplayName = "TooManyRequests")]
+        [DataRow(FaultInjectionOperationType.ReadItem, FaultInjectionServerErrorType.ReadSessionNotAvailable, 404, 1002, DisplayName = "ReadSessionNotAvailable")]
+        [DataRow(FaultInjectionOperationType.ReadItem, FaultInjectionServerErrorType.Timeout, 410, 20001, DisplayName = "Timeout")]
```
xinlian12, line 785: Timeout maps to 408?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionDirectModeTests.cs
```diff
@@ -0,0 +1,1201 @@
...
+            FaultInjector faultInjector = new FaultInjector(new List<FaultInjectionRule> { connectionErrorRule });
+            await this.Initialize(faultInjector, true);
+            
+            ChaosInterceptor? interceptor = faultInjector.GetChaosInterceptor() as ChaosInterceptor;
+            Assert.IsNotNull(interceptor);
+
+            FaultInjectionDynamicChannelStore channelStore = interceptor.GetChannelStore();
+            Assert.IsTrue(channelStore.GetAllChannels().Count > 0);
+
+            connectionErrorRule.Enable();
```
xinlian12, line 1072: we should also test after this rule, all the channels have been closed

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionDirectModeTests.cs
```diff
@@ -0,0 +1,1201 @@
...
+                || diagnosticsString.Contains("201") 
+                || diagnosticsString.Contains("204"));
+        }
+        private void ValidateFaultInjectionRuleApplication(
+            CosmosDiagnostics diagnostics,
+            int statusCode,
+            int subStatusCode,
+            FaultInjectionRule rule)
+        {
+            string diagnosticsString = diagnostics.ToString();
```
xinlian12, line 1194: do we track the ruleId being applied in the CosmosDiagnostics?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionDirectModeTests.cs
```diff
@@ -0,0 +1,1201 @@
...
+            }
+            finally
+            {
+                hitCountRule.Disable();
+            }
+        }
+
+        [TestMethod]
+        [Owner("nalutripician")]
+        [Description("Tests endpoint filtering with including primary replica")]
```
xinlian12, line 957: I think two more tests worth adding:
1. Test endpoint will be impacted by the operationType. For example if the operationType being write operation ,and  the errorType is Gone, then only primary replica or write region will be chosen
2. Add filter by container tests -> as the rule is configured on client level

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionRuleProcessor.cs
```diff
@@ -0,0 +1,364 @@
...
+                    if (isWriteOnly)
+                    {
+                        TransportAddressUri primary = await this.ResolvePrimaryTransportAddressUriAsync(fauntInjectionAddressRequest, true);
+                        return new List<Uri> { primary.Uri };
+                    }
+
+                    IEnumerable<Uri> resolvedEndpoints = (await this.ResolveAllTransportAddressUriAsync(
+                            fauntInjectionAddressRequest,
+                            addressEndpoints.IsIncludePrimary(),
+                            true)).OrderBy(address => address.Uri.ToString())
```
xinlian12, line 319: we should make sure primary replica is always in the head of the list -> because when includePrimaryReplica being true, we would want to guarantee primary replica will always be selected

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionServerErrorResultInternal.cs
```diff
@@ -0,0 +1,229 @@
...
+                    
+                    return storeResponse;
+
+                case FaultInjectionServerErrorType.TooManyRequests:
+                    INameValueCollection tooManyRequestsHeaders = args.RequestHeaders;
+                    tooManyRequestsHeaders.Add(HttpConstants.HttpHeaders.RetryAfterInMilliseconds, "500");
+
+                    storeResponse = new StoreResponse()
+                    {
+                        Status = 429,
```
xinlian12, line 157: also substatusCode 3200?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
```diff
@@ -0,0 +1,74 @@
...
+        private readonly ConcurrentDictionary<Guid, (DateTime, string)> executionsByActivityId;
```
xinlian12, line 10: Another question -> does .Net SDK use unique activityId for each network request? 
for example, for boundedStaleness, SDK will send two requests to backend concurrently, does each of those request will have a unique activityId?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionServerErrorRule.cs
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        public void Disable()
+        {
+            this.enabled = false;
+        }
+
+        public void Enable()
+        {
+            this.enabled = true;
```
xinlian12, line 192: for enable hmm, I am wondering what should be the ideal behavior. Should we also reset the effectiveStartTime etc when the rule got re-enabled?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionConnectionType.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Client using Direct Mode
+        /// </summary>
+        Direct,
+
+        /// <summary>
+        /// Client using Gateway Mode CURRENTLY NOT SUPPORTED
+        /// </summary>
+        Gateway,
```
xinlian12, line 19: since Gateway is not supported currently, I think we should remove from here?

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/ChaosInterceptor.cs
NaluTripician: applicationContext does track the application of each request that has a rule applied to it. If a rule is not applied, then there is no data stored. I think the memory issue is something to look further into. I have not done any testing to see the memory impact of this so far.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionConditionInternal.cs
NaluTripician: At this point in .NET ChannelCallArguments are not created yet.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
NaluTripician: Will add in a future PR.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/RntbdConnectionErrorInjector.cs
NaluTripician: Yes, this causes the channel to be closed.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionDirectModeTests.cs
NaluTripician: I think here it should be Create. When we are testing create, we first delete the item so that we can verify that when the create operation has the fault injected the item is not actually created.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/tests/FaultInjectionDirectModeTests.cs
NaluTripician: Will add an example in the PR description. Also planning on adding a sample library + documentation at some point. Currently there is no tracking in the diagnostics. Will improve this method to check activityId with the one stored in the Application Context.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
NaluTripician: I think that they might have the same activityId

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionServerErrorRule.cs
NaluTripician: @kirankumarkolli what are your thoughts on this? 

After an offline discussion we believe that there are two trains of thought we can follow with this enable. 
1. Have the expire time of the start/reset when a rule is enabled/reenabled
2. Have the expire time of the rule stay constant based on when the rule is created. 

This method is useful to have during testing. Either of the two ways above could make sense but there would just need to have good documentation added explaining it.

kirankumarkolli: The tool is for us to do choas and fast reasoning. 
So my preference is option to align with fast reasoning. 

Full context on Rule state transition will cover that all context.

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/implementataion/FaultInjectionApplicationContext.cs
NaluTripician: This will become a problem in bounded staleness scenarios where rules are applied twice on the same activityId. Changed to be a list like executionsByRuleId. `TryGetRuleExecutionByActivityId` will now return the first rule Id that is applied.

philipthomas-MSFT (approved):

xinlian12 (approved):
LGTM, thanks

philipthomas-MSFT (dismissed):

xinlian12 (dismissed):
LGTM, thanks

philipthomas-MSFT (approved):

philipthomas-MSFT (approved):

kundadebdatta (approved):
LGTM!

kirankumarkolli (approved):

Issue #4232, opened 2024-01-03
Distributed Tracing: Network level attributes OTel Semantics


sourabh1007:
Right now, in **.NET SDK**, below attributes are getting recorded

| Key | Value| Comment|
|--|--|--|
| tcp.uri | TNTBD replica url ||
| tcp.sub_status_code| Substatuscode returns||
| tcp.status_code| Status Code||

In **Java SDK**, below attributes are getting captured

| Key | Value| Comment|
|--|--|--|
| rntbd.operation_type | e.g. Create ||
| rntbd.activity_id | GUID ||
| rntbd.request_charge| ||
| rntbd.latency_created|||
| rntbd.latency_channelacquisitionstarted|||
| rntbd.latency_pipelined|||
| rntbd.resource_type|||
| rntbd.latency_decodetime|||
| rntbd.latency_transitiontime|||
| rntbd.lsn|||
| rntbd.gclsn|||
| rntbd.backend_latency|||
| rntbd.is_new_channel|||
| rntbd.request_size_bytes|||
| rntbd.url|||
| rntbd.latency|||
| rntbd.region|||
| rntbd.response_size_bytes|||
| rntbd.status_code|||
| rntbd.request_session_token|||
| rntbd.session_token|||
| rntbd.partition_key_range_id|||

Here intention is to make it consistent across SDKs.

### Proposal

Here we are capturing Network calls which are similar to HTTP calls (just different protocol) . ref. Open Telemetry HTTP call semantics https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client

**PREFIX = cosmosdb.tcp** (other suggestions are direct, rntbd, cdbdct)

| Key | Value| Comment|
|--|--|--|
| url.full| replica address||
| url.scheme | rntbd |
| ~~useragent.original~~| ||
| ~~cosmosdb.tcp.request.operation_type~~  PREFIX.operation_type |||
| ~~cosmosdb.tcp.request.activity_id~~ PREFIX.activity_id | GUID ||
| ~~cosmosdb.tcp.request.resource_type~~ PREFIX.resource_type|||
| ~~cosmosdb.tcp.request.is_new_channel~~ PREFIX.is_new_channel |||
| PREFIX.request.size_in_bytes|||
| ~~cosmosdb.tcp.request.region~~ PREFIX.region |||
| PREFIX.request.session_token|||
| PREFIX.response.session_token|| Session token returned from service|
[...]

PR #4282, opened 2024-01-26
[Internal] Direct Version Upgrade: Fixes code to make it compatible with 3.32.1 of direct package

### LINKED ISSUE #3780: [Internal] Logic App Patch Add and Set operations issue with Long numbers being treated as Integer
There is an issue where patching using add or set with Long numbers are yielding exceptions because there is not JTokenType Long and it is defaulted as integer but when cast or converted to Integer, it bombs. This is for Logic Apps standard plan only.

https://www.newtonsoft.com/json/help/html/t_newtonsoft_json_linq_jtokentype.htm

sourabh1007:
## Description

This PR covers: Upgrading Direct package to 3.32.1
1. Added `DiagnosticSource` package of 6.0.1.
2. Upgrade `System.Runtime.CompilerServices.Unsafe` to 6.0.0
3. Fixes Open telemetry code to make it compatible with new diagnostic scope API (which is part of Direct package).
4. As part of direct package upgrade (diagnostic scope file upgrade), 
     - Operation Name attribute is changed. It has prefix "Operation" or "Request" representing if it is operation level activity or network level (_**Impact:** There should not be any impact to the customer, as I didn't see any usage of operation name attribute, in Appinsights site_).
     - Now, `Null` attributes are not getting added in the activity. Previously, everything was getting added to the activity and it was APM tool responsibility to populate it or not. Updated baseline tests to handle it. (_**Impact:** Now APM tools has to do less work, less amount of data will go on wire, No impact to customer_)
     - Network level attribute's prefix is changed to `rntbd` from `tcp`. (_Already discussed and closed ref. https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4232_)
     - Refactored `OpenTelemetryRecorderFactory.cs` to use same instance of `DiagnosticScopeFactory`.
     - In case of exception, `error.type` attribute is also getting added.
     - tags values should be retrieved from `activity.TagObjects` instead of `activity.Tags`.

## Type of change

Package upgrade


NaluTripician (approved):

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
```diff
@@ -112,10 +112,10 @@
 	</ItemGroup>
 
 	<ItemGroup>
-		<PackageReference Include="Azure.Core" Version="1.19.0" />
+		<PackageReference Include="Azure.Core" Version="1.37.0" />
```
kirankumarkolli, line 115: Can you please expand why is this change needed?

Microsoft.Azure.Cosmos/src/CosmosClient.cs
```diff
@@ -21,6 +21,7 @@ namespace Microsoft.Azure.Cosmos
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Cosmos.Tracing.TraceData;
     using Microsoft.Azure.Documents;
+    using ResourceType = Microsoft.Azure.Documents.ResourceType;
```
kirankumarkolli, line 24: What the conflicting type? (i mean why is explicit qualification needed just with this direct package)

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
sourabh1007: I wanted latest version of diagnostic source used by Azure Core. but looks like, I can upgrade only diagnostic source it would be cleaner changes.

Microsoft.Azure.Cosmos/src/CosmosClient.cs
sourabh1007: It was conflicting with Azure core library, looks like in latest version of Azure Core package also has file with exactly same class name. So to handle it, I used this piece of code. But Now I have downgraded the Azure Core and added diagnostic source package directly.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -237,7 +243,7 @@ Documents.OperationType operationType
 
                     if (!DiagnosticsFilterHelper.IsSuccessfulResponse(this.response.StatusCode, this.response.SubStatusCode))
                     {
-                        this.scope.Failed();
+                        this.scope.Failed(new Exception($"Request failed with.{this.response.StatusCode}/{this.response.SubStatusCode}"));
```
kirankumarkolli, line 246: Why new exception is needed? 
One alternative is to add a overload Failed() which can take the required context?

sourabh1007: now, with new version of DiagnosticScope. It is mandatory to pass exception message or error code in the Failed method. Here, it would only go when we have a response which we don't consider as successful...so here context, we have, is `status code` and `substatuscode`, of that response. Thats why I decided to log it here in custom exception message. 
In case of proper exception. anyway it goes here https://github.com/Azure/azure-cosmos-dotnet-v3/blob/b107e0e468dc21b92c2ec5ced1e592d0993da023/Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs#L191 where we would have proper context.
If you think `Exception` type too generic to log, you can suggest me some better exception type, which can indicate this failure.

kirankumarkolli: Exception creation is very expensive. And at scale it will impact performance.

sourabh1007: now setting a string value other than exception object...anyway if request is failing then, already customer would be facing some other issues.

kirankumarkolli: Thanks, Posible idea is to limit impact esp during those time.

kirankumarkolli (approved):

NaluTripician (approved):

PR #4283, opened 2024-01-26
CosmosClientOptions: Adds validation to check DisableServerCertificateValidation and ServerCertificateCustomValidationCallback are set together

sourabh1007:
## Description

As part of this PR, validating `DisableServerCertificateValidation` flag and `ServerCertificateCustomValidationCallback` option are set together. Throws `ArgumentException` if it is true.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -733,6 +733,11 @@ namespace Microsoft.Azure.Cosmos
         internal bool? EnableCpuMonitor { get; set; }
 
         /// <summary>
+        /// Flag indicates the value of DisableServerCertificateValidation flag set at connection string level.Default it is false.
+        /// </summary>
+        internal bool DisableServerCertificateValidation { get; set; }
+
+        /// <summary>
         /// Gets or sets Client Telemetry Options like feature flags and corresponding options
         /// </summary>
         public CosmosClientTelemetryOptions CosmosClientTelemetryOptions { get; set; }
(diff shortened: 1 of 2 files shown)


NaluTripician (changes requested):
Could you please add some context to the PR description on why both cannot be set. Other than that, LGTM.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -929,6 +930,14 @@ private void ValidatePartitionLevelFailoverSettings()
             {
                 throw new ArgumentException($"{nameof(this.ApplicationPreferredRegions)} is required when {nameof(this.EnablePartitionLevelFailover)} is enabled.");
             }
+        }
+
+        private void ValidateServerCallbackSettings()
+        {
+            if (this.HttpClientFactory != null && this.ServerCertificateCustomValidationCallback != null)
```
kirankumarkolli, line 937: Clarification: The combination that we want to fail is DisableServertCert and ServerCertificateCustomValdiatonCallback right?

sourabh1007: Disableservercert flag internally sets this callback in clientoptions only.

sourabh1007: ohh sorry I misread the discussion. Yes, it was flag and callback. Thanks for pointing it out.

kirankumarkolli (approved):

kirankumarkolli:
Lets port this to 3.38 release.

/cc: @kundadebdatta

NaluTripician (approved):

PR #4292, opened 2024-01-30
[DO NOT MERGE]ConsistencyWriter: Tuning HEAD request back-off based on SubStatusCode (LSN gap)

FabianMeiswinkel:
# Pull Request Template

## Description

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
@@ -40,17 +40,16 @@ For globally strong write:
     internal sealed class ConsistencyWriter
     {
         private const int maxNumberOfWriteBarrierReadRetries = 30;
-        private const int delayBetweenWriteBarrierCallsInMs = 30;
         private const int extensiveLsnGapThreshold = 10_000;
-        private const int extensiveLsnGapDelayBetweenWriteBarrierCallsInMs = 5000;
         private const int highLsnGapThreshold = 1_000;
-        private const int highLsnGapDelayBetweenWriteBarrierCallsInMs = 1000;
         private const int mediumLsnGapThreshold = 100;
-        private const int mediumLsnGapDelayBetweenWriteBarrierCallsInMs = 100;
-
         private const int maxShortBarrierRetriesForMultiRegion = 4;
-        private const int shortbarrierRetryIntervalInMsForMultiRegion = 10;
 
+        private static readonly TimeSpan defaultDelayBetweenWriteBarrierCalls = TimeSpan.FromMilliseconds(30);
+        private static readonly TimeSpan shortDelayBetweenWriteBarrierCallsForMultipleRegions = TimeSpan.FromMilliseconds(10);
+        private static readonly TimeSpan extensiveLsnGapDelayBetweenWriteBarrierCalls = TimeSpan.FromMilliseconds(5000);
+        private static readonly TimeSpan highLsnGapDelayBetweenWriteBarrierCalls = TimeSpan.FromMilliseconds(1000);
+        private static readonly TimeSpan mediumLsnGapDelayBetweenWriteBarrierCalls = TimeSpan.FromMilliseconds(100);
         private readonly StoreReader storeReader;
         private readonly TransportClient transportClient;
         private readonly AddressSelector addressSelector;
@@ -300,21 +299,24 @@ For globally strong write:
             return false;
         }
 
-        private async Task delayInitialHeadRequestAsync(long? currentGlobalCommitedLsn, long selectedTargetGlobalCommittedLsn)
+        internal static TimeSpan getInitialHeadRequestDelay(long? currentGlobalCommitedLsn, long selectedTargetGlobalCommittedLsn)
         {
             if (currentGlobalCommitedLsn != null &&
                 currentGlobalCommitedLsn < selectedTargetGlobalCommittedLsn - extensiveLsnGapThreshold)
             {
-                await Task.Delay(highLsnGapDelayBetweenWriteBarrierCallsInMs);
+                return highLsnGapDelayBetweenWriteBarrierCalls;
             }
-            else if (currentGlobalCommitedLsn != null &&
+
+            if (currentGlobalCommitedLsn != null &&
                 currentGlobalCommitedLsn < selectedTargetGlobalCommittedLsn - highLsnGapThreshold)
             {
-                await Task.Delay(mediumLsnGapDelayBetweenWriteBarrierCallsInMs);
+                return mediumLsnGapDelayBetweenWriteBarrierCalls;
             }
+
+            return TimeSpan.Zero;
         }
 
-        private async Task delayBetweenHeadRequestsAsync(int remainingWriteBarrierRetryCount, IList<ReferenceCountedDisposable<StoreResult>> responses)
+        internal static TimeSpan getDelayBetweenHeadRequests(int remainingWriteBarrierRetryCount, IList<ReferenceCountedDisposable<StoreResult>> responses)
         {
             int minLSNGap = extensiveLsnGapThreshold + 1;
             foreach (ReferenceCountedDisposable<StoreResult> response in responses)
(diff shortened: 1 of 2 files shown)


Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
```diff
@@ -41,6 +41,12 @@ internal sealed class ConsistencyWriter
     {
         private const int maxNumberOfWriteBarrierReadRetries = 30;
         private const int delayBetweenWriteBarrierCallsInMs = 30;
+        private const int extensiveLsnGapThreshold = 10_000;
+        private const int extensiveLsnGapDelayBetweenWriteBarrierCallsInMs = 5000;
+        private const int highLsnGapThreshold = 1_000;
+        private const int highLsnGapDelayBetweenWriteBarrierCallsInMs = 1000;
```
ealsur, line 47: How do we know these wait times are what the service expects us to wait and it would not incur in more latency than required?

FabianMeiswinkel: Rough estimate I used was. We know one replica can at most process 10,000 RU/s. We also know each write operationw ill take at least 5 RU. So most optimistic progress would be 2000 Transactions/LSNs per second

FabianMeiswinkel: And then some modifications around that

FabianMeiswinkel: Definitely no exact science - and I will need to align with whatever Abhijit does in backend. But I think the worst case above should be solid?

FabianMeiswinkel: @ealsur - good catch by the way. I ran a couple of queries to see based on Replica table how many LSNs a replica can progress. Mostly my dirty estimate with 5000 is good - but there seem to be few scenarios when replica during copy progress catches up extremely quickly. So, I will probably need to be abit more conservative in the delays.

kirankumarkolli: Forgot what's the name used but a state where it's booted from log but not replica feed. There is will pick-up much faster. This was the contract I was requesting Subbu before. 

Thoughts on possibility of service emit hint for HEAD request's? (Body is not allowed should be max header only)

FabianMeiswinkel: But > 5000 LSNs only happens at a very early stage after restarting the Replica - the last 10 seocnds or so LSN progress was < 1000 - So I think the rough formula above assuming at most 5000 LSN/s makes sense - but will need to be adjusted by backend delay (for this account backend added 250ms delay - via config) and then probably be conservative with delays but aggressive enough to really reduce PCU usage is a reasonable compromise

FabianMeiswinkel: > Forgot what's the name used but a state where it's booted from log but not replica feed. There is will pick-up much faster. This was the contract I was requesting Subbu before.
> 
> Thoughts on possibility of service emit hint for HEAD request's? (Body is not allowed should be max header only)

You mean like with 429 a header telling the delay? I can discuss with Abhijit - the problem is that the backend has better state at rpelication lag but no context on whether it is the first, second or nth retry
I think with the substatus code they provide enough info for the SDK to start making decisions. Even if i reduce the max delay to let's say a second it is still sufficient to make significant progress on high CPU. And when we do head requests we know we don't have to wait for some intermediary LSNs but the latest LSNs  - so IMO the copy from log mode can be ignored IMO - because we don't make any delay linearly to LSN difference -just based on some categories.

Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
```diff
@@ -250,7 +255,7 @@ internal string LastWriteAddress
                     {
                         using (DocumentServiceRequest barrierRequest = await BarrierRequestHelper.CreateAsync(request, this.authorizationTokenProvider, null, request.RequestContext.GlobalCommittedSelectedLSN))
                         {
-                            if (!await this.WaitForWriteBarrierAsync(barrierRequest, request.RequestContext.GlobalCommittedSelectedLSN))
+                            if (!await this.WaitForWriteBarrierAsync(barrierRequest, globalCommittedLsn, request.RequestContext.GlobalCommittedSelectedLSN))
```
ealsur, line 258: I like the approach of the solution and it makes sense. I just wonder if we know which was the original intent of the new substatus codes and if whoever added them intended a different approach in terms of delay management.

Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
```diff
@@ -294,13 +299,85 @@ private bool ShouldPerformWriteBarrierForGlobalStrong(StoreResult storeResult)
             return false;
         }
 
-        private async Task<bool> WaitForWriteBarrierAsync(DocumentServiceRequest barrierRequest, long selectedGlobalCommittedLsn)
+        internal static TimeSpan getInitialHeadRequestDelay(long? currentGlobalCommitedLsn, long selectedTargetGlobalCommittedLsn)
+        {
+            if (currentGlobalCommitedLsn != null &&
```
ealsur, line 304: Since its a nullable you can use `HasValue`.

Consider also bailing out earlier?

```
if (!currentGlobalCommiedLsn.HasValue)
{
  return TimeSpan.Zero;
}
```

Or seeing how this is used, and to avoid the `TimeSpan.Zero` comparisson after, why not:

```
internal static bool ShouldDelayInitialHeadRequestDelay(long? currentGlobalCommitedLsn, long selectedTargetGlobalCommittedLsn, out TimeSpan delay)`
```

That way you can avoid the `TimeSpan.Zero` comparison afterwards.

Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
```diff
@@ -294,13 +299,85 @@ private bool ShouldPerformWriteBarrierForGlobalStrong(StoreResult storeResult)
...
+                return mediumLsnGapDelayBetweenWriteBarrierCalls;
+            }
+
+            return TimeSpan.Zero;
+        }
+
+        internal static TimeSpan getDelayBetweenHeadRequests(int remainingWriteBarrierRetryCount, IList<ReferenceCountedDisposable<StoreResult>> responses)
         {
-            int writeBarrierRetryCount = ConsistencyWriter.maxNumberOfWriteBarrierReadRetries;
+            int minLSNGap = extensiveLsnGapThreshold + 1;
```
ealsur, line 321: Why is the `+ 1` needed?

Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
```diff
@@ -294,13 +299,85 @@ private bool ShouldPerformWriteBarrierForGlobalStrong(StoreResult storeResult)
...
+        private async Task<bool> WaitForWriteBarrierAsync(DocumentServiceRequest barrierRequest, long? currentGlobalCommitedLsn, long selectedTargetGlobalCommittedLsn)
+        {
+            int remainingWriteBarrierRetryCount = ConsistencyWriter.maxNumberOfWriteBarrierReadRetries;
 
             long maxGlobalCommittedLsnReceived = 0;
-            while (writeBarrierRetryCount-- > 0)
+            while (remainingWriteBarrierRetryCount-- > 0)
             {
+                barrierRequest.RequestContext.TimeoutHelper.ThrowTimeoutIfElapsed();
+                TimeSpan initialDelay = getInitialHeadRequestDelay(currentGlobalCommitedLsn, selectedTargetGlobalCommittedLsn);
```
ealsur, line 375: In the PR I see 2 scenarios, one is the initial delay and on that one we are making the calculation of the gap and deciding the delay, in the second one, we are taking the substatuscode from the service, so the size of the gap is the service decision.

Would just adding the second part simplify and make more sense? Or is the initial delay also something that is driving a big impact on the service?

PR #4293, opened 2024-01-31
[Internal] Query: Fixes Debug Assert related to client distribution plan

adityasa:
# Pull Request Template

## Description

With a recent change, client expects coordinator plan to be described using a field named clientDistributionPlan. Previously the field was called coordinatorDistributionPlan. However the change is not deployed in the backend fully and as a result when running against production account it causes assert to fire. Commenting out the code for now since it's not hooked up to any production usage (or tests).

## Type of change

## Closing issues


sc978345 (approved):

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
```diff
@@ -339,18 +339,19 @@ public override void ClearSessionTokenCache(string collectionFullName)
...
+                    //     bool backendPlan = distributionPlan.TryGetValue("backendDistributionPlan", out CosmosElement backendDistributionPlan);
+                    //     bool clientPlan = distributionPlan.TryGetValue("clientDistributionPlan", out CosmosElement clientDistributionPlan);
+
+                    //     Debug.Assert(clientPlan == backendPlan, "Response Body Contract was violated. Out of the backend and client plans, only one  is present in the distribution plan.");
+
+                    //     if (backendPlan && clientPlan)
+                    //     {
+                    //         distributionPlanSpec = new DistributionPlanSpec(backendDistributionPlan.ToString(), clientDistributionPlan.ToString());
+                    //     }
+                    // }
```
sboshra, line 354: It might be better to use #if to comment this code out

sboshra (approved):

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
adityasa: Leaving as is since the change is minor with no product impact.

PR #4299, opened 2024-02-01
[Internal] Query: Adds environment variable for overriding EnableOptimisticDirectExecution default

adityasa:
# Pull Request Template

## Description

This change adds an ability to override the default value of EnableOptimisticDirectExecution setting using an environment variable named "AZURE_COSMOS_OPTIMISTIC_DIRECT_EXECUTION_ENABLED".
The environment variable should have a value that can be parsed as a boolean in order for the override to work.

As before the default value is true and also the backend setting using clientDisableOptimisticDirectExecution acts as a master switch and acts as global account-level override to disable/enable ODE behavior (in clients).

The environment variable can be used while upgrading the sdk package to 3.38 onwards, which turns on ODE by default. As mentioned in the release notes (of 3.38), due to the use of a new type of continuation token, applications that execute queries (and their continuations) in a stateless manner on machines that can be using different versions of sdk, failures can be observed.
This environment variable can be set to true before upgrading the sdk package to avoid such failures during the sdk deployment process in production environment. Once the upgrade is complete the environment variable can be removed or set to false.

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -280,5 +281,29 @@ internal override void PopulateRequestOptions(RequestMessage request)
                 request.Headers.ContinuationToken = continuationToken;
             }
         }
+
+        /// <summary>
+        /// Returns the default value for ODE when QueryRequestOptions are initialized.
+        /// </summary>
+        /// <returns>If an environment variable named EnableOptimisticDirectExecution has as a boolean (true/false) value, it will be honored, otherwise default value of True</returns>
+        private static bool GetDefaultODEValue()
```
FabianMeiswinkel, line 289: Please move this to ConfigurationManager - see https://github.com/Azure/azure-cosmos-dotnet-v3/blob/b933cb8e450df5f1e6d79c648d161a06c46ea65b/Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs

And also use similar Prefix / syntax for environment variable name.

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -280,5 +281,29 @@ internal override void PopulateRequestOptions(RequestMessage request)
...
+        {
+            string enableODEEnvironmentValue;
```
kundadebdatta, line 291: Define it as `null` here.

```
string enableODEEnvironmentValue = null;
```

That way, we can avoid assigning `null` at line `299` and just add a warning message,

adityasa: Thanks, but I would prefer to keep the assignment explicit. This way any branches (now or in future) that don't assign a value are caught by the compiler instead of assuming a (potentially wrong) default.

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
adityasa: @FabianMeiswinkel 
I like the idea of using ConfigurationManager, but it raises an exception (due to how it's implemented at the moment) when a value is supplied that doesn't match the type (see my test variations, many of which I will have to remove).
Given that it's already shipped this way, I am guessing this is not a big deal?

FabianMeiswinkel: I like your way of handling exceptions and using default in that case. IMO this should be added in ConfigurationManager (either as only option or if for compat reasons we need to keep the current one throwing as an alternative). My man point was naming pattern for environment variables and having a single place. In Java we have much wider config options via Java System Properties or in several cases alternatively Environment variables. And knowing there is a single place will all the definitions has helped. See https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/Configs.java

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -280,5 +281,29 @@ internal override void PopulateRequestOptions(RequestMessage request)
...
+            try
+            {
+                enableODEEnvironmentValue = Environment.GetEnvironmentVariable(nameof(EnableOptimisticDirectExecution));
+            }
+            catch (SecurityException)
+            {
+                // Ignore permission failure in trying to retrieve the environment variable.
+                enableODEEnvironmentValue = null;
```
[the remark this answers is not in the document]
adityasa: Not applicable anymore.

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
adityasa: For handling securityexception - I don't think back-compat should be an issue (since the application would have crashed in the past and now we would continue, but log it as a warning (per Debdatta's suggestion below).
If we haven't heard about this case, I am also guessing that not many users run client in low-priv processes.
I am tempted to keep that change separate to keep the hotfix contained.
I don't have strong feelings either ways.

sboshra (approved):
:shipit:

kirankumarkolli (approved):

PR #4304, opened 2024-02-03
[Internal] Query: Adds environment variable for overriding EnableOptimisticDirectExecution default

adityasa:
# Pull Request Template

## Description

Cherry-pick of following PR from master:
https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4299

This change adds an ability to override the default value of EnableOptimisticDirectExecution setting using an environment variable named "AZURE_COSMOS_OPTIMISTIC_DIRECT_EXECUTION_ENABLED".
The environment variable should have a value that can be parsed as a boolean in order for the override to work.

As before the default value is true and also the backend setting using clientDisableOptimisticDirectExecution acts as a master switch and acts as global account-level override to disable/enable ODE behavior (in clients).

The environment variable can be used while upgrading the sdk package to 3.38 onwards, which turns on ODE by default. As mentioned in the release notes (of 3.38), due to the use of a new type of continuation token, applications that execute queries (and their continuations) in a stateless manner on machines that can be using different versions of sdk, failures can be observed.
This environment variable can be set to true before upgrading the sdk package to avoid such failures during the sdk deployment process in production environment. Once the upgrade is complete the environment variable can be removed or set to false.

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Azure.Cosmos
         /// <value>
         /// Direct (optimistic) execution offers improved performance for several kinds of queries such as a single partition streaming query.
         /// </value>
-        public bool EnableOptimisticDirectExecution { get; set; } = true;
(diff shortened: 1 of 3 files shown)


adityasa:
This is abandoned based on offline discussion with @kundadebdatta.

PR #4317, opened 2024-02-09
Client Telemetry: Refactors code to stop throwing an exception.

sourabh1007:
## Description

**What is the change?**
In Client telemetry, we are not throwing any exception to customer instead we just log it in traces but internally (within client telemetry module) we are generating exception which is not really required because in the end we are going to just log it.
So, making changes to just log it at first place.

**Why?**
Recently, one of the client telemetry user, reported app failure during the client telemetry service failure. After analysis, it was realized that Client Telemetry Service failure was leading to high CPU usage which makes their application to misbehave.
As soon as, they disabled this feature, it all came to normal and along with this they are not seeing any impact of client telemetry service if it is returning 200.

## Type of change


Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -52,7 +52,6 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
             catch (Exception ex)
```
kirankumarkolli, line 52: Is try-catch at each level needed? 

The `RunProcessorTask` loop can handle all exception and log it once right?

sourabh1007: I don't want process to fail if anything happens during serialization or while calling service.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -52,7 +52,6 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
...
             {
                 DefaultTrace.TraceError($"Exception while serializing telemetry payload: {ex}");
```
ealsur, line 54:
```suggestion
                DefaultTrace.TraceError("Exception while serializing telemetry payload: {0}", ex);
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/ClientTelemetryTests.cs
```diff
@@ -246,6 +246,9 @@ public async Task CheckIfPayloadIsDividedCorrectlyAsync(int expectedOperationInf
             Assert.AreEqual(expectedRequestInfoSize, actualRequestInfoSize, "Request Info is not correct");
         }
 
+        [Ignore("Keeping this to test the timeout scenario manually in future and " +
```
ealsur, line 249: You could validate that it does **not** throw exceptions to the upper layers though?

sourabh1007: made changes.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
kirankumarkolli: That makes sense but is error handling needed at all levels?
It can be handled at the top level right?

sourabh1007: it is there at only 2 levels: one is at serialization level and other at service call level....if anything fails there and we have traces ON, then we should get what is actually failing. 
Secondly, throwing exception is costly affair. Thats why, I kept try-catch for API call in local function. on the other hand, other try-catch is just safety catch, ideally, I am not expecting serialization logic to fail.

sourabh1007: we are claiming that client telemetry is "exceptionless feature". So I think it wouldn't hurt to have few try-catch to make sure it is truly exceptionless

philipthomas-MSFT (dismissed):

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
sourabh1007: According to offline discussion, keeping the top level try-catch only.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -221,9 +221,7 @@ internal static async Task RunProcessorTaskAsync(string telemetryDate, Task proc
                 Task resultTask = await Task.WhenAny(processingTask, delayTask);
                 if (resultTask == delayTask)
                 {
-                    DefaultTrace.TraceVerbose($"Processor task with date as {0} is canceled as it did not finish in {1}", telemetryDate, timeout);
-                    // Operation cancelled
-                    throw new OperationCanceledException(string.Format($"Processor task with date as {0} is canceled as it did not finish in {1}", telemetryDate, timeout));
+                    DefaultTrace.TraceError($"Processor task with date as {0} is canceled as it did not finish in {1}", telemetryDate, timeout);
```
philipthomas-MSFT, line 224: cancelled

philipthomas-MSFT, line 224: finished in {1} milliseconds/seconds/hours? What is the unit of time?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -75,62 +74,52 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
...
-                        request.Headers.Add(HttpConstants.HttpHeaders.EnvironmentName, envName);
-                    }
-
-                    return request;
-                }
+            using HttpRequestMessage request = new HttpRequestMessage
+            {
+                Method = HttpMethod.Post,
+                RequestUri = new Uri(endpointUrl),
+                Content = new StringContent(jsonPayload, Encoding.UTF8, "application/json")
```
philipthomas-MSFT, line 83: Why don't we ever use `MediaTypeNames.Application.Json` instead of "application/json"?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -75,62 +74,52 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
...
-                                                    HttpTimeoutPolicyNoRetry.Instance,
-                                                    null,
-                                                    cancellationToken);
+            async ValueTask<HttpRequestMessage> CreateRequestMessage()
+            {
+                INameValueCollection headersCollection = new StoreResponseNameValueCollection();
+                await this.tokenProvider.AddAuthorizationHeaderAsync(
+                        headersCollection,
+                        new Uri(endpointUrl),
+                        "POST",
```
philipthomas-MSFT, line 92: Why don't we use `HttpMethod.Post.Method` instead of the literal string "POST"?

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetryProcessor.cs
```diff
@@ -75,62 +74,52 @@ internal ClientTelemetryProcessor(CosmosHttpClient httpClient, AuthorizationToke
...
                 {
-                    DefaultTrace.TraceError("Telemetry Service API response not successful. Status Code : {0},  Message : {1}", response.StatusCode, response.ReasonPhrase);
-                    throw new Exception(string.Format("Telemetry Service API response not successful. Status Code : {0},  Message : {1}", response.StatusCode, response.ReasonPhrase));
+                    request.Headers.Add(key, headersCollection[key]);
                 }
-                else
+
+                request.Headers.Add(HttpConstants.HttpHeaders.DatabaseAccountName, globalDatabaseAccountName);
+                String envName = ClientTelemetryOptions.GetEnvironmentName();
+                if (!String.IsNullOrEmpty(envName))
```
philipthomas-MSFT, line 102: Any reason why you are declaring this as "String" and not the alias "string"?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Telemetry/ClientTelemetryTests.cs
```diff
@@ -249,100 +249,108 @@ public async Task CheckIfPayloadIsDividedCorrectlyAsync(int expectedOperationInf
...
+                for (int i = 0; i < 20; i++)
+                {
+                    OperationInfo opeInfo = new OperationInfo(Regions.WestUS,
+                                                            0,
+                                                            Documents.ConsistencyLevel.Session.ToString(),
+                                                            "databaseName" + i,
+                                                            "containerName",
+                                                            Documents.OperationType.Read,
+                                                            Documents.ResourceType.Document,
+                                                            200,
```
philipthomas-MSFT, line 293: Can we move away primitive obsession? Specifically, the arguments 200 and 0 for parameters **statusCode** and **subStatusCode** respectively. Or use named arguments for now.

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: it will print the `TimeSpan` which will be quite self-explanatory

philipthomas-MSFT: Ah I see now. All I had in my mind was a **long** or **int**. Thanks. Resolving conversation.

philipthomas-MSFT (dismissed):

philipthomas-MSFT (approved):

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
```diff
@@ -221,9 +221,7 @@ internal static async Task RunProcessorTaskAsync(string telemetryDate, Task proc
                 Task resultTask = await Task.WhenAny(processingTask, delayTask);
                 if (resultTask == delayTask)
                 {
-                    DefaultTrace.TraceVerbose($"Processor task with date as {0} is canceled as it did not finish in {1}", telemetryDate, timeout);
-                    // Operation cancelled
-                    throw new OperationCanceledException(string.Format($"Processor task with date as {0} is canceled as it did not finish in {1}", telemetryDate, timeout));
+                    DefaultTrace.TraceError($"Processor task with date as {0} is cancelled as it did not finish in {1} milliseconds", telemetryDate, timeout.TotalMilliseconds);
```
FabianMeiswinkel, line 224: Why does this need to be an error? If the open telemetry endpoint is not reachable it will just clutter traces unnecessarily?

FabianMeiswinkel (approved):
LGTM except for one comment

Microsoft.Azure.Cosmos/src/Telemetry/ClientTelemetry.cs
sourabh1007: you mean client telemetry?

Putting it in verbose is not going to help us in anyway..we would never know if processor task failed or problem is somewhere else, if we don't see telemetry at backend. it won't get called, if client telemetry service call failed, it will get call only if processor initiation is failing.

PR #4138, opened 2023-10-18
Query: Adds support for System.Text LINQ Custom Serializer

Maya-Painter:
## Description

This change allows users to use custom serializers for the translation of LINQ queries.

To use their own custom serializer for LINQ translations users must implement `CosmosLinqSerializer` and set as a custom serializer on the `CosmosClient`. (Sample in CosmosLinqSerializer.cs)

This allows for the following scenarios:
### Property name customization via `JsonPropertyName`

Given an object:

```
public record MyItem
{
      [JsonPropertyName("id")]
      public string Id { get; set; }

      [JsonPropertyName("project_id")]
      public string ProjectId { get; set; }
}
```

With the following query:
`query = testContainer.GetItemLinqQueryable<MyItem>().Where(i => i == new MyItem() { ProjectId = "test", Id = "id1"});`

This will be translated into the following SQL query, taking into account the property attributes on the MyItem class
`{"query":"SELECT VALUE root FROM root WHERE (root = {\"id\": \"id1\", \"project_id\": \"test\"})"} `

### Enum conversion via `StringEnumConverter`
This change also means that the `System.Text.Json.Serialization.StringEnumConverter()` will be honored when translating enums in LINQ queries. If `JsonStringEnumConverter()` is specified as a converter on the custom serializer, enums do not need to be decorated with `[JsonConverter(typeof(JsonStringEnumConverter))]`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

## Closing issues

closes #1936, 
closes #3207, 
closes #2685, 
closes #3250, 
closes #3697, 
closes #1253, 
closes #2386, 
closes #3280, 
closes #3491

PR #4332, opened 2024-02-28
[Internal] SystemTextJsonSerializer: Adds Default STJ Serializer

### LINKED ISSUE #4330: Add a Default `System.Text.JSON` Implementation for `CosmosLinqSerializer`
- Add a Default `System.Text.JSON` Implementation for `CosmosLinqSerializer`. The serializer will be shipped along with the SDK codebase.

kundadebdatta:
# Pull Request Template

## Description

This PR adds a Default `System.Text.JSON` Implementation for `CosmosLinqSerializer`. This default serializer can be used along with the `CosmosClientOptions` to set the runtime serializer. Please take a look at the example below for usage:

```
        {
            Serializer = new CosmosDefaultSystemTextJsonSerializer(
                new System.Text.Json.JsonSerializerOptions())
        };
```

Note that this is just a default implementation which handles the basic scenarios. Any `JsonSerializerOptions` passed in here may not going to be reflected in `SerializeMemberName()`. 

To handle special cases, please create a custom serializer which inherits from the `CosmosDefaultSystemTextJsonSerializer` and overrides the `SerializeMemberName()` method to add any special handling.

Below is an example of a custom serializer, which is inherits the `CosmosDefaultSystemTextJsonSerializer` to cover the special handling for camel case naming property.

```
    public class CustomSystemTextJsonSerializer : CosmosDefaultSystemTextJsonSerializer
    {
        /// <summary>
        /// Initializes a new instance of the <see cref="CustomSystemTextJsonSerializer "/> class.
        /// </summary>
        /// <param name="jsonSerializerOptions"></param>
        public CustomSystemTextJsonSerializer (JsonSerializerOptions jsonSerializerOptions)
            : base(jsonSerializerOptions)
        {
        }

        /// <inheritdoc/>
        public override string SerializeMemberName(MemberInfo memberInfo)
        {
            System.Text.Json.Serialization.JsonExtensionDataAttribute jsonExtensionDataAttribute =
                memberInfo.GetCustomAttribute<System.Text.Json.Serialization.JsonExtensionDataAttribute>(true);
            if (jsonExtensionDataAttribute != null)
            {
                return null;
            }

            JsonPropertyNameAttribute jsonPropertyNameAttribute = memberInfo.GetCustomAttribute<JsonPropertyNameAttribute>(true);
            if (!string.IsNullOrEmpty(jsonPropertyNameAttribute?.Name))
            {
                return jsonPropertyNameAttribute.Name;
            }

            return System.Text.Json.JsonNamingPolicy.CamelCase.ConvertName(memberInfo.Name);
        }
    }
```

## Type of change

## Closing issues

To automatically close an issue: closes #4330


Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,83 @@
...
+                {
+                    return default;
+                }
+
+                if (typeof(Stream).IsAssignableFrom(typeof(T)))
+                {
+                    return (T)(object)stream;
+                }
+
+                return (T)this.systemTextJsonSerializer.Deserialize(stream, typeof(T), default);
```
bartelink, line 49:
```suggestion
                return this.systemTextJsonSerializer.Deserialize<T>(stream, default);
```

kundadebdatta: In order to use the `System.Text.Json` (STJ) serializer, we currently depend on the `Azure.Core` nuget package (specifically for this class `JsonObjectSerializer`). The current version of `Azure.Core` library, referred in our v3 .NET SDK is `1.19.0`, which is quite old. This internally uses `System.Text.Json` v `4.6.0`.

STJ `4.6.0` doesn't support the way you suggested to use the generic Deserializer.

bartelink: OK, that makes sense

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Poco/STJ/Cars.cs
```diff
@@ -0,0 +1,102 @@
...
+            "Toyota 4Runner",
+            "Toyota Camry",
+            "Toyota Prius",
+            "Toyota Rav4",
+            "Toyota Avalon",
+            "Honda Civic",
+            "Honda Accord",
+            "Honda CRV",
+            "Honda MRV",
+            "BMW C18",
```
bartelink, line 21: Hm, need more typesafety in the language

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Poco/STJ/Cars.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            "all wheel drive",
+            "abs",
+            "apple car play",
+            "power window",
+            "heated seating",
+            "steering mounted control",
+        };
+
+        [JsonConstructor]
```
bartelink, line 36: pretty sure this is not required for up to date STJ versions
unless the older version being used dictates this, maybe best to not tag it?

kundadebdatta: Removed and validated the tests are passing.

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    using System;
+    using System.IO;
+    using System.Reflection;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+    using global::Azure.Core.Serialization;
+
+    /// <summary>
+    /// This class provides a default implementation of STJ Cosmos Linq Serializer.
```
ealsur, line 15: Maybe: 
```suggestion
    /// This class provides a default implementation of System.Text.Json for CosmosClient
```

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// </summary>
+        private readonly JsonObjectSerializer systemTextJsonSerializer;
+
+        /// <summary>
+        /// Creates an instance of <see cref="CosmosDefaultSystemTextJsonSerializer"/>
+        /// with the default values for the Cosmos SDK
+        /// </summary>
+        /// <param name="jsonSerializerOptions">An instance of <see cref="JsonSerializerOptions"/> containing the json serialization options.</param>
+        public CosmosDefaultSystemTextJsonSerializer(
+            JsonSerializerOptions jsonSerializerOptions)
```
Maya-Painter, line 30: Just a heads up that any `JsonSerializerOptions` passed in here are not going to be reflected in `SerializeMemberName()` as it is now, which may cause some confusion.

For example, if customers passed in a JsonSerializerOption such as below

```
            JsonSerializerOptions options = new()
            {
                PropertyNamingPolicy = JsonNamingPolicy.CamelCase
            }
```

This would not be honored by SerializeMemberName unless it included special handling for this, e.g.

```
        public override string SerializeMemberName(MemberInfo memberInfo)
        {
            System.Text.Json.Serialization.JsonExtensionDataAttribute jsonExtensionDataAttribute =
                memberInfo.GetCustomAttribute<System.Text.Json.Serialization.JsonExtensionDataAttribute>(true);
            if (jsonExtensionDataAttribute != null)
            {
                return null;
            }

            JsonPropertyNameAttribute jsonPropertyNameAttribute = memberInfo.GetCustomAttribute<JsonPropertyNameAttribute>(true);
            if (!string.IsNullOrEmpty(jsonPropertyNameAttribute?.Name))
            {
                return jsonPropertyNameAttribute.Name;
            }

            return System.Text.Json.JsonNamingPolicy.CamelCase.ConvertName(memberInfo.Name);
        }
```

SerializeMemberName is included on the interface since customers will need to modify it themselves.

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,91 @@
...
+    using System.IO;
+    using System.Reflection;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+    using global::Azure.Core.Serialization;
+
+    /// <summary>
+    /// This class provides a default implementation of STJ Cosmos Linq Serializer.
+    /// </summary>
+    public class CosmosDefaultSystemTextJsonSerializer : CosmosLinqSerializer
```
adityasa, line 17: >CosmosLinqSerializer [](http://example.com/codeflow?start=57&length=20)

Did you mean to implement LinqSerializer or just plain-old CosmosSerializer?

If this is intentional, can you add coverage for LINQ scenarios?

Here's a sample PR that touched relevant query surface area:
[https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4138/files](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4138/files)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs

adityasa (changes requested):
:clock1:

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
kundadebdatta: Discussed with @Maya-Painter offline. One possible solution could be to remove the `JsonSerializerOptions` from the `CosmosDefaultSystemTextJsonSerializer()` constructor. That way, the serializerOptions won't be passed from the end users and we could avoid handling the custom attributes.

The usage will look like the below:

```
        {
            Serializer = new CosmosDefaultSystemTextJsonSerializer()
        };
```

Thoughts @ealsur , @kirankumarkolli ?

ealsur: I'm guessing that it depends on our goal. Is the goal to provide a base implementation to cover the most general cases? If yes, then makes sense to not have the custom options. Or if we know which custom options are invalid, then we could throw if the customer is setting those?

Another idea would be to allow customers to inherit and override the LINQ serialization, so a user with custom options can create a superset class that inherits from this but has a custom SerializeMemberName (I think this is valid right now). But in this case discovery that they need to add their own implementation might be obscure.

kundadebdatta: I agree with the second point. I think to add such custom scenarios, the end users can create their own custom serializer that inherits the default serializer and add an override for the `SerializeMemberName()` method.

Therefore, I have added more documentation on the `SerializeMemberName()` method to cover this. Will update the PR description to make things more clear.

@Maya-Painter : Let me know if you have any concerns/ thoughts.

Cc: @ealsur , @kirankumarkolli .

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
Maya-Painter: I'm fine with it as long as this class and any potential LINQ GitHub issues arising from its use belong to the SDK team :)

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            if (typeof(Stream).IsAssignableFrom(typeof(T)))
+            {
+                return (T)(object)stream;
+            }
+
+            using (stream)
+            {
+                using StreamReader reader = new (stream);
+                return JsonSerializer.Deserialize<T>(reader.ReadToEnd(), this.jsonSerializerOptions);
```
joegoldman2, line 53:
```suggestion
                return JsonSerializer.Deserialize<T>(stream, this.jsonSerializerOptions);
```
STJ is able to deserialize the stream directly without needing to allocate an intermediate string.

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,128 @@
...
+    using System;
+    using System.IO;
+    using System.Reflection;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// This class provides a default implementation of System.Text.Json Cosmos Linq Serializer.
+    /// </summary>
+    public class CosmosDefaultSystemTextJsonSerializer : CosmosLinqSerializer
```
joegoldman2, line 16: Should the name be `CosmosSystemTextJsonSerializer` instead?

kundadebdatta: Refactored the code to reflect the same.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,128 @@
...
+    using System;
+    using System.IO;
+    using System.Reflection;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// This class provides a default implementation of System.Text.Json Cosmos Linq Serializer.
+    /// </summary>
+    public class CosmosSystemTextJsonSerializer : CosmosLinqSerializer
```
kirankumarkolli, line 16: Contract JSON file update is needed right?

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        /// <summary>
+        /// Convert a MemberInfo to a string for use in LINQ query translation.
+        /// Note that this is just a default implementation which handles the basic scenarios. Any <see cref="JsonSerializerOptions"/> passed in
+        /// here are not going to be reflected in SerializeMemberName(). For example, if customers passed in a JsonSerializerOption such as below
+        /// <code language="c#">
+        /// <![CDATA[
+        /// JsonSerializerOptions options = new()
+        /// {
+        ///     PropertyNamingPolicy = JsonNamingPolicy.CamelCase
```
kirankumarkolli, line 80: CTOR taking the serializeroptions but the guidance going into the description. 
Its lots confusing, @ealsur thoughts?

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            if (typeof(Stream).IsAssignableFrom(typeof(T)))
+            {
+                return (T)(object)stream;
+            }
+
+            using (stream)
+            {
+                using StreamReader reader = new (stream);
+                return JsonSerializer.Deserialize<T>(reader.ReadToEnd(), this.jsonSerializerOptions);
```
joegoldman2, line 53:
```suggestion
                return JsonSerializer.Deserialize<T>(stream, this.jsonSerializerOptions);
```
STJ is able to deserialize the stream directly without needing to allocate an intermediate string.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
ealsur: Not sure I understand the comment. From what I read. This is a note for the LINQ serialization for users that may need to adapt this particular method for their custom options if any.
Maybe this needs to go into `<remarks>` instead of description might be the only thing I can think of.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kundadebdatta: Yes, needed to update the contract JSON.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kundadebdatta: Right now, the `System.Text.JSON` package depends on the `Azure.Core` version `1.19.0`, which uses the STJ version `4.6.0`. This version of the STJ serializer doesn't support the `Deserialize()` directly with a `stream`. Hence the conversion is needed at the moment.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
```diff
@@ -55,9 +60,23 @@ public async static Task Initialize(TestContext textContext)
             // to verify the translated LINQ directly as other queries type.
             client.DocumentClient.OnExecuteScalarQueryCallback = q => lastExecutedScalarQuery = q;
 
-            dbName = $"{nameof(LinqAggregateCustomSerializationBaseline)}-{Guid.NewGuid().ToString("N")}";
+            dbName = $"{nameof(LinqAggregateCustomSerializationBaseline)}-{Guid.NewGuid():N}";
             testDb = await client.CreateDatabaseAsync(dbName);
             testContainer = testDb.CreateContainerAsync(new ContainerProperties(id: Guid.NewGuid().ToString(), partitionKeyPath: "/Pk")).Result;
+
+            defaultCosmosSerializer = new CosmosSystemTextJsonSerializer(new JsonSerializerOptions());
```
adityasa, line 67: Is this used? If not, please remove.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
```diff
@@ -55,9 +60,23 @@ public async static Task Initialize(TestContext textContext)
...
+
+            defaultClient = TestCommon.CreateCosmosClient((cosmosClientBuilder)
+                => cosmosClientBuilder.WithCustomSerializer(customCosmosSerializer));
```
adityasa, line 70: >customCosmosSerializer [](http://example.com/codeflow?start=60&length=22)

Should this be defaultCosmosSerializer?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
```diff
@@ -29,6 +29,11 @@ public class LinqAggregateCustomSerializationBaseline : BaselineTests<LinqAggreg
         private static Cosmos.Database testDb;
         private static Container testContainer;
 
+        private static CosmosSerializer defaultCosmosSerializer;
```
adityasa, line 32: >default [](http://example.com/codeflow?start=40&length=7)

Please rename to stjCosmosSerializer

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
```diff
@@ -29,6 +29,11 @@ public class LinqAggregateCustomSerializationBaseline : BaselineTests<LinqAggreg
...
+        private static CosmosClient defaultClient;
```
adityasa, line 33: >default [](http://example.com/codeflow?start=36&length=7)

Please rename to stjClient (if it intends to use new STJ serializer)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
```diff
@@ -29,6 +29,11 @@ public class LinqAggregateCustomSerializationBaseline : BaselineTests<LinqAggreg
...
+        private static Cosmos.Database testDbDefault;
```
adityasa, line 34: >testDbDefault [](http://example.com/codeflow?start=39&length=13)

testDbSTJ

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
```diff
@@ -29,6 +29,11 @@ public class LinqAggregateCustomSerializationBaseline : BaselineTests<LinqAggreg
...
+        private static Container testContainerDefault;
```
adityasa, line 35: testContainerSTJ

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
kundadebdatta: Updated with latest commit.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
kundadebdatta: Updated with latest commit.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
kundadebdatta: Updated with latest commit.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
kundadebdatta: Yes - it should be the `defaultCosmosSerializer`. Thanks! I have renamed it though.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
kundadebdatta: It should be. I have refactored the test to use this serializer.

adityasa:
>                     "Select number -> Min w/ mapping", b => getQuery(b)

Nit - please include serializer name here too.

---
Refers to: Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs:154 in 683ca13. [](commit_id = 683ca137d0f817993e77ffc9162f8b601d5b1c00, deletion_comment = False)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
```diff
@@ -98,29 +117,37 @@ static DataObjectDotNet createDataObj(int index, bool camelCase)
...
 
-            foreach (bool applyCustomSerializer in new List<bool>{ true, false })
+            foreach (KeyValuePair<string, int> entry in serializerIndexes)
             {
-                Func<bool, IQueryable<DataObjectDotNet>> getQuery = getQueryList[applyCustomSerializer ? 0 : 1];
+                Func<bool, IQueryable<DataObjectDotNet>> getQuery = getQueryList[entry.Value];
 
                 inputs.Add(new LinqAggregateInput(
-                    "Avg, Custom serializer: " + applyCustomSerializer, b => getQuery(b)
+                    "Avg, Serializer Name: " + entry.Key, b => getQuery(b)
```
adityasa, line 138: >Serializer Name: [](http://example.com/codeflow?start=26&length=16)

Neat!

adityasa (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Serializer/CosmosDefaultSystemTextJsonSerializer.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
joegoldman2: Ok, clear. Can you add a TODO not to forget to update the implementation once the version of STJ will be updated?

bartelink: snap! https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4332#discussion_r1520153811

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,130 @@
...
+        ///         return jsonPropertyNameAttribute.Name;
+        ///     }
+        ///     return System.Text.Json.JsonNamingPolicy.CamelCase.ConvertName(memberInfo.Name);
+        /// }
+        /// ]]>
+        /// </code>
+        /// To handle such scenarios, please create a custom serializer which inherits from the <see cref="CosmosSystemTextJsonSerializer"/> and overrides the
+        /// SerializeMemberName to add any special handling.
+        /// </remarks>
+        public override string SerializeMemberName(MemberInfo memberInfo)
```
Maya-Painter, line 111: I've updated the STJ sample code and made some improvements to this function in this PR https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4420.  Could you update this function to look like this - https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/SystemTextJson/CosmosSystemTextJsonSerializer.cs ? Thanks!

kundadebdatta: Thanks @Maya-Painter. We are trying to gather some feedback on this from the .NET team on this. I will take your changes as well, and update the PR.

eiriktsarpalis: The `JsonTypeInfo`/`JsonPropertyInfo` types contain authoritative information on the JSON contract used by a particular type/property. See https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/custom-contracts for more details.

kirankumarkolli: Thank you @eiriktsarpalis !

kundadebdatta: Thanks for your suggestion @eiriktsarpalis! Looks like, these ([JsonSerializerOptions.TryGetTypeInfo](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Fdotnet%2Fapi%2Fsystem.text.json.jsonserializeroptions.trygettypeinfo%3Fview%3Dnet-8.0&data=05%7C02%7Cdkunda%40microsoft.com%7Ce3b58671a1f243d198dd08dc632f838f%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638494302946820285%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=dWcECXAzip0CNwP2%2FdqUxZ0ySHvwKiYvhRtBZv07dww%3D&reserved=0), [custom contracts](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/custom-contracts)) are applicable for the latest versions of the `System.Text.JSON` package, and it's not available in `4.6.0`.
 
In order to use the `System.Text.Json` (STJ) serializer, we currently depend on the `Azure.Core` nuget package (specifically for this class `JsonObjectSerializer`). The current version of `Azure.Core` library, referred in our v3 .NET SDK is `1.19.0`, which is quite old. This internally uses `System.Text.Json` version `4.6.0`.
 
I am planning to keep this suggestion as follow-up task [4479](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4479) to upgrade the `Azure.Core` library to the latest. Even the latest version of `Azure.Core` [1.39.0]( https://nuget.info/packages/Azure.Core/1.39.0) library uses the STJ version `4.7.2`.
 
image

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Poco/STJ/Cars.cs
kundadebdatta: I wouldn't worry about it too much since this is an unit test.

adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,107 @@
...
+            this.jsonSerializerOptions = jsonSerializerOptions;
+        }
+
+        /// <inheritdoc/>
+        public override T FromStream<T>(Stream stream)
+        {
+            if (stream == null)
+                throw new ArgumentNullException(nameof(stream));
+
+            if (stream.CanSeek && stream.Length == 0)
```
kirankumarkolli, line 40: Is `canSeek` is not really required right?

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kirankumarkolli: How about only make this type public once is fully addressed?

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        /// <inheritdoc/>
+        public override Stream ToStream<T>(T input)
+        {
+            MemoryStream streamPayload = new ();
+            using Utf8JsonWriter writer = new (streamPayload);
+
+            JsonSerializer.Serialize(
+                writer: writer,
```
bartelink, line 64: its great that I don't need to care as this will be an impl detail deep in a library but this formatting/named params is way overkill and makes it look like some really exciting stuff is going on, which there isnt
is the intention to disambiguate from some adjacent overload?
in general I'd be nice if this code was scannable so people can compare their impls quickly before deleting them in favor of this one

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
bartelink: The fact that Azure.Core 'uses' (gets by with) an old version does not of itself mean that this lib can't take a dependency on a higher version. I'd ask, what version would simplify and/or _significantly_ improve perf here? Then validate the answer by lookign at package download counts at https://www.nuget.org/packages/system.text.json.
For instance 6.0.1 has a net6.0 variant and seems to be in popular use, and is only 18 months newer - it's not like STJ is known to be a pain to upgrade.
But of course, I'm just being a devil's advocate; there's lots of silent votes for the status quo!
Hopefully there will be a security reason along to trigger a forced upgrade in due course ;)
The other benefit of taking a higher dep is that perf in general has been on an upward trajectory, and the perceived experience of the SDK would hence also improve for people that don't explicitly pin latest in their apps

Arcalise08:
Any update on this PR? It's been a few weeks and this is a much-anticipated change.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (stream == null)
+                throw new ArgumentNullException(nameof(stream));
+
+            if (stream.CanSeek && stream.Length == 0)
+            {
+                return default;
+            }
+
+            if (typeof(Stream).IsAssignableFrom(typeof(T)))
```
kirankumarkolli, line 45: Can this check be before above one?

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                using StreamReader reader = new (stream);
+                return JsonSerializer.Deserialize<T>(reader.ReadToEnd(), this.jsonSerializerOptions);
+            }
+        }
+
+        /// <inheritdoc/>
+        public override Stream ToStream<T>(T input)
+        {
+            MemoryStream streamPayload = new ();
```
kirankumarkolli, line 60: Do different payload sizes make any difference on performance?

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kirankumarkolli: @Bartelink is the ASK to inline arguments? (make code compact)

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <param name="memberInfo">Any MemberInfo used in the query.</param>
+        /// <returns>A serialized representation of the member.</returns>
+        /// <remarks>
+        /// Note that this is just a default implementation which handles the basic scenarios.To handle any special cases,
+        /// please create a custom serializer which inherits from the <see cref="CosmosSystemTextJsonSerializer"/> and overrides the
+        /// SerializeMemberName() method.
+        /// </remarks>
+        public override string SerializeMemberName(MemberInfo memberInfo)
+        {
+            JsonExtensionDataAttribute jsonExtensionDataAttribute =
```
kirankumarkolli, line 84: Quick clarification: Is this API called for every propertyname conversion? If so the cardinality for each type conversion is much higher right? 

Assuming the mapping is fixed (one type is loaded), is below code optimal?

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
bartelink: Pretty much... I don't think the explicitly named args add anything in this case (even if they can sometimes be a really good idea). If you do that, it becomes a single line, and conveys "we're just serializing in a perfectly normal fashion here", whereas a multi line thing has me wondering, what subtlety are we trying to emphasize here?

kirankumarkolli:
> Any update on this PR? It's been a few weeks and this is a much-anticipated change.

@Arcalise08 due to some reprioritizations this work was paused. We expect it to resume it after 2 weeks. 
We will try to opportunistically accelerate its close as soon as possible.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kundadebdatta: My understanding is that the `.CanSeek` check is required, otherwise the `.Length` call may fail. [See this](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.canseek?view=net-8.0#remarks) for more details.

image

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kundadebdatta: Refactored the code to address this.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kirankumarkolli: As discussed its a follow-up work item.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kirankumarkolli: Its only called from LINQ and can be followed-up later.

kundadebdatta: This will be addressed by [Task-4479](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4479)

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kundadebdatta: A follow up task [4479](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4479) created to track this.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kundadebdatta: This is now addressed with the latest commit.

kirankumarkolli (approved):

aavasthy (approved):
👍

ealsur (approved):

itsawindup:
Forgive me if I am being a bit slow, but since the change from public to internal in commit 1ff1f06, I can't see a way to use this in my own code.  Have I missed something obvious?

bartelink:
@kirankumarkolli @kundadebdatta Please follow up by adding a quick summary as to the final state of play here.
Thousands of people will hit this issue via search, and this saga had lots of twists and turns along the way!

kirankumarkolli:
@bartelink apologies, we change the usage model towards the end of the PR stage and missed to reflect that in the description.
@kundadebdatta can you please refresh the description.

kundadebdatta:
> @kirankumarkolli @kundadebdatta Please follow up by adding a quick summary as to the final state of play here. Thousands of people will hit this issue via search, and this saga had lots of twists and turns along the way!

Thanks for bringing this up @bartelink ! I have updated the summary to explain how to use this serializer publicly by default. For your reference:

There is a separate [PR-4589](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4589) which explains the steps to use the `CosmosSystemTextJsonSerializer` publicly by setting `UseSystemTextJsonSerializerWithOptions` parameter in the `CosmosClientOpions` for both preview and GA contract. 

I hope this will be helpful! Please feel free to reach out, if you have any questions.

bartelink:
I guess the problem is that over the years there's been lots of PRs and/or snippets. I know this PR has `[Internal]` in the title, but the average person without context won't appreciate what that implies, and will instead focus on the "Adds Default STJ Serializer" part of the title and say "OK, it was merged some time ago and I'm on latest, and I can't see this `CosmosDefaultSystemTextJsonSerializer` anywhere".

----

So the problem is that someoe scanning even the current overview might conclude that there is a `CosmosDefaultSystemTextJsonSerializer`. For very good reasons I agree with, that has has wound up not being `public`. So I guess I was hoping that the last piece of info on this thread (and maybe a prominent note in the overview) could summarise the key points:

- There are two abstract classes in `Microsoft.Azure.Cosmos`: `CosmosSerializer` and `CosmosLinqSerializer` that allow you to customize everything
- Normally you don't want to do that though:
  1. The `CosmosDefaultSystemTextJsonSerializer` in the overview exists, but it ultimately ended up being `internal`
  2. But, that's OK - there's a newer API (see #4589) that provides access by doing
    
         var opts = new CosmosClientOptions {
             UseSystemTextJsonSerializerWithOptions = System.Text.Json.JsonSerializerOptions.Default };

----

Something like this, but written in proper English ;) ?

To apply the `CosmosDefaultSystemTextJsonSerializer` in the overview of this issue on 3.43.0 or later, you set the value of the `CosmosClientOptions.UseSystemTextJsonSerializerWithOptions` property to an instance of `System.Text.Json.JsonSerializerOptions` , e.g.:

     var opts = new CosmosClientOptions { UseSystemTextJsonSerializerWithOptions = System.Text.Json.JsonSerializerOptions.Default };
     var client = new CosmosClient(...., opts);

kirankumarkolli:
Updated the title to remove default and also the updated PR description to reference the PR which made the contracts public.

PR #4333, opened 2024-02-29
Distributed Tracing: Adds configuration to disable network level tracing in sdk permanently

sourabh1007:
## Description
It was decided that we are going to stop emitting network level traces to avoid confusion in customer. 
Because it has information related to RNTBD calls which is an internal protocol, without much information documented publicly.

1. Updated Sample Project also: to get it reflected here https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/sdk-observability?tabs=dotnet
2. Removed all the logic related to network level traces and corresponding tests.

## Type of change


Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6654,9 +6654,16 @@ private void InitializeDirectConnectivity(IStoreClientFactory storeClientFactory
             }
             else
             {
+                // It is decided to switch this feature off completely for external users but keep it uchanged for internal users,
+                // due to the nature of information, we are collecting here. RNTBD is internal protocol and we do not expose it to the customers.
                 Documents.Telemetry.DistributedTracingOptions distributedTracingOptions = new ()
                 {
+#if INTERNAL
                     IsDistributedTracingEnabled = !this.cosmosClientTelemetryOptions.DisableDistributedTracing
```
kirankumarkolli, line 6662: nit: Same name for both ClientOptions and down-stream but using it differently is little confusing!

Microsoft.Azure.Cosmos.Samples/Usage/OpenTelemetry/Program.cs
```diff
@@ -71,7 +71,6 @@ static async Task Main()
                 AppContext.SetSwitch("Azure.Experimental.EnableActivitySource", true);
                 _traceProvider = Sdk.CreateTracerProviderBuilder()
                     .AddSource("Azure.Cosmos.Operation", // Cosmos DB source for operation level telemetry
-                               "Azure.Cosmos.Request", // Cosmos DB source for DIRECT Mode network request level telemetry
```
kirankumarkolli, line 74: Test coverage for scenario where `Azure.Cosmos.Request` is specified and expected behavior.

kirankumarkolli (dismissed):

Microsoft.Azure.Cosmos.Samples/Usage/OpenTelemetry/Program.cs
sourabh1007: This feature is disabled from public SDK, hence for customer, this feature doesn't exist anymore and there is no expected behavior

Microsoft.Azure.Cosmos/src/DocumentClient.cs
sourabh1007: Yes, it is little confusing, I believe that's what we decided because `IsDistributedTracingEnabled` is internal flag we wanted to keep it straightforward but `DisableDistributedTracing` is customer exposing flag and we decided this name to keep it user friendly(and align with the future decision where we want to keep it enable always)

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6654,9 +6654,16 @@ private void InitializeDirectConnectivity(IStoreClientFactory storeClientFactory
             }
             else
             {
+                // It is decided to switch this feature off completely for external users but keep it uchanged for internal users,
```
philipthomas-MSFT, line 6657: spelling: "unchanged"

philipthomas-MSFT (approved):

Microsoft.Azure.Cosmos/src/DocumentClient.cs
FabianMeiswinkel, line 6662: Why keep it for internal usage when you remove all tests? Either remove it enitrely or keep the tests (for internal usage)?

sourabh1007: Compute is going to use this feature that's why it is available for internal usage.
From Tests it is all removed to make sure this feature is not available for public package. I tried to write a test but looks like there is no sraightforward way to access this option in tests :(

philipthomas-MSFT (approved):

kirankumarkolli (approved):

PR #4334, opened 2024-02-29
ChangeFeedProcessor: Fixes when ChangeFeedMode is switched should throw an exception.

### LINKED ISSUE #4308: Change Feed Processor: Switching ChangeFeedMode should throw an exception
From the previous PR [4264](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4264#discussion_r1469819930), this discussion was centered around the .NET SDK functional capabilities of switching from **AllVersionsAndDeletes** to **LatestVersion** with the same lease and how was the Java SDK is handling this type of scenario. Also, what is the expected behavior if this capability does not exist currently in the .NET SDK.

***Scenario***
- Ingest 10 documents.
- Read 5 documents from the beginning with **LatestVersion**.
- Stop the processor.
- Switch modes to **AllVersionsAndDeletes**.
- Start the processor.
- I want to know if the remaining 5 documents will be picked up or if it will hang in this scenario. If 10 and 5 are too small to reliably stop processing in the middle of the set, you can test with higher document counts.

- **Ends with unlocking the CFP AllVersionsAndDeletes capability for public preview. The last [4264](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4264) added the engineering, but kept it internalized until we are aligned on the understanding of the existing and expected behavior.**
- **Establish, recommendations, and provide samples and best practices for using this type of scenarios.**
- **Ends with exceptions being thrown when `StartFromBeginning` is set.**
- **Ends with exceptions being thrown if the lease is the same, or changes, after the switch.**

- Exception
- Support

- Write unit tests to demonstrate the existing and expected behavior.
- Design and document any changes if needed to reflect the existing and expected behavior.

**Additional context**

**Clarification questions**
- Exception route
  - What type of exception should be thrown? New custom exception.
    - What message?
  - Should we allow switching change feed mode from **AllVersionsAndDeletes** to **LatestVersion**?
    - If so, why? What CX scenarios and use cases support this type of behavior?
  - Exception is thrown when **LatestVersion** has **StartFromBeginning** set?
  - Exception thrown when using **same lease only**?
  - How do we identify that the mode has switched?
    - Save the mode?
      - .info file?
      - stored in the continuation token?
- Support route (After GA)
  - **AllVersionsAndDeletes** use `ChangeFeedHandler<T>` as part of its response contract of the feed. **LatestVersion** uses `ChangesHandler<T>`. How will switching modes affect this?
  - Since there is no mode exposed to the CX and it is predetermined based on whether the CX invokes `GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes` or `GetChangeFeedProcessorBuilder`, are we expecting the CX to invoke the **LatestVersion** method in order to switch modes, or do we need to expose a way for the customer to switch modes without this?
  - If `WithStartFromBeginning()` is set, then this is not allowed and an exception should be thrown. But why?
  - Should we be supporting a different lease on the switch?
    - Why or why not?

philipthomas-MSFT:
# Pull Request Template
See #4308 for problem statement, repro, scenarios, expected behavior, etc.

## Description

## Type of change

## Closing issues

To automatically close an issue: closes #4308


Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorRunner.cs
```diff
@@ -139,11 +139,14 @@ private async Task InitializeLeaseStoreAsync()
...
-                    leaseContainer: this.leaseContainer,
-                    leaseContainerPrefix: leasePrefix,
-                    instanceName: ChangeFeedEstimatorRunner.EstimatorDefaultHostName);
+                DocumentServiceLeaseStoreManager documentServiceLeaseStoreManager = await DocumentServiceLeaseStoreManagerBuilder
+                    .InitializeAsync(
+                        monitoredContainer: this.monitoredContainer,
+                        leaseContainer: this.leaseContainer,
+                        leaseContainerPrefix: leasePrefix,
+                        instanceName: ChangeFeedEstimatorRunner.EstimatorDefaultHostName,
+                        changeFeedMode: this.changeFeedLeaseOptions.Mode)
```
ealsur, line 148: This is for the Estimator, Estimator based on the previous PR, would use LatestVersion?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
```diff
@@ -70,13 +70,20 @@ public ChangeFeedProcessorBuilder WithInstanceName(string instanceName)
         }
 
         /// <summary>
-        /// Sets the mode for the change freed processor.
+        /// Sets the <see cref="ChangeFeedMode"/> on <see cref="ChangeFeedProcessorOptions"/> for a monitored
```
ealsur, line 73: ChangeFeedProcessorOptions and ChangeFeedLeaseOptions are all internals, do not include them in public API documentation, nor link gitHub issues on public documentation.

The `<summary>` should be short and simple. Any other comments like "You cannot change the mode once started" should be in `<remarks>`

ealsur: (I know this API is `internal` right now but eventually will be `public`)

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +80,110 @@ public override async Task StopAsync()
...
+        /// <summary>
+        /// If the lease<see cref="Container"/>'s lease document is found, this method checks for lease 
+        /// document's <see cref="ChangeFeedMode"/> and if the new <see cref="ChangeFeedMode"/> is different
+        /// from the current <see cref="ChangeFeedMode"/>, a <see cref="CosmosException"/> is thrown.
+        /// This is based on an issue located at <see href="https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4308"/>.
+        /// </summary>
+        /// <param name="key"></param>
+        private async Task ChangeFeedModeSwitchingCheckAsync(string key)
+        {
+            FeedIterator<dynamic> feedIterator = this.leaseContainer.GetItemQueryIterator<dynamic>(queryText: "SELECT * FROM c");
```
ealsur, line 123: This is querying all the documents in the lease container, the lease container can have other unrelated documents, documents from other change feed processors that are sharing the same lease container (https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/change-feed-processor?tabs=dotnet#share-the-lease-container).

These should ideally be done by the `DocumentServiceLeaseStoreManager`. It has access to `DocumentServiceLeaseContainer` which has the method to list all leases `GetAllLeasesAsync`, that will return only the leases for the current processor configuration (monitored container + processorname)

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
philipthomas-MSFT: I reverted back to the `<summary>` on the last commit without the extra information to keep it simple.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +80,110 @@ public override async Task StopAsync()
...
             string containerRid = await this.monitoredContainer.GetCachedRIDAsync(
-                forceRefresh: false, 
-                NoOpTrace.Singleton, 
+                forceRefresh: false,
+                NoOpTrace.Singleton,
                 default);
+
             string monitoredDatabaseAndContainerRid = await this.monitoredContainer.GetMonitoredDatabaseAndContainerRidAsync();
+
+#if PREVIEW
```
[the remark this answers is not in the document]
philipthomas-MSFT: Yes, that was removed already. I think you checked this before I pushed it.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorRunner.cs
ealsur: No. There are 2 ways to use the Estimator https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-use-change-feed-estimator?tabs=dotnet

One is a pull model, the other is a push model. This class is for the Push model. It is not part of the Processor, but a way to run estimator as a notifications service.

In a nutshell, this should also have the Mode as LatestVersion.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
ealsur: Did you push the commit? I don't see any new commits

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: I will look into this today. Thanks for the feedback and recommendation.

philipthomas-MSFT: So, @ealsur I am discovering that your suggestion to call `GetAllLeasesAsync` does not work, even after `InitializeAsync` is executed. `GetAllLeasesAsync` does not return anything.

ealsur: `GetAllLeasesAsync` is used during the initialization and load balancing:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/deab135c9b8709abf3e62f7be08e6c022513e4bf/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionLoadBalancerCore.cs#L80

Either the LeaseContainer has not been initialized (this might be related to ordering of execution) or the lease container is empty.

Whatever solution we decide though, we cannot be executing a blank query, that would not work because of the nature of the lease container. Even if the lease container is used by a single CFP, there is a `.info` document and then the leases (the `.info` document is not a lease).

philipthomas-MSFT: And it works because I am always getting all the lease documents, looking for the matching id using the key, then checking for the mode.

philipthomas-MSFT: So, I am not opposed to creating a new `GetAllLeasesAsync` variation without the filter and putting that on the `DocumentServiceLeaseContainer ` and moving the logic. But using the current `GetAllLeasesAsync` will not work with all scenarios.

philipthomas-MSFT: Now the query looks like `SELECT * FROM c WHERE STARTSWITH(c.id, 'LatestVersion127.0.0.1_IjdaAA==_IjdaAKPUuS0=..')` or `SELECT * FROM c WHERE STARTSWITH(c.id, 'AllVersionsAndDeletes127.0.0.1_IjdaAA==_IjdaAKPUuS0=..')`. If we understandingly don't want to `SELECT * FROM c`, I can make to `
SELECT * FROM c WHERE CONTAINS(c.id, '127.0.0.1_+rF1AA==_+rF1ANlHcJQ=..'))` which leaves out the processor name.

ealsur: > My solution works both ways, when the lease document id is the same, but also different when a new processor is created

That would be an incorrect behavior.

A different `id` pattern means a different Deployment Unit: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/change-feed-processor?tabs=dotnet#deployment-unit

You can only enforce the mode check for the same `id` pattern. What you need to validate is that, the user, for the **same Deployment Unit**, did not alter the mode.

A different `processorName` means a different configuration and that is fine, because if you start a processor with name `X` and then create another processor with name `Y`, `Y` will not pickup the leases from `X`, they are **different processors**.

The scenario that we need to validate is if the user uses the **same** `processorName` + Monitored Container + Lease Container, which yields the same `id` pattern, that they don't switch modes. Because it's when these 3 properties are the same that we resume using the same leases.

Just to clarify, we need to block this:

```
.GetChangeFeedProcessorBuilder(processorName: $"myProcessor", onChangesDelegate: (ChangeFeedProcessorContext context, IReadOnlyCollection<dynamic> documents, CancellationToken token) => 
```

Stop processor.

```
.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes(processorName: $"myProcessor", onChangesDelegate: (ChangeFeedProcessorContext context, IReadOnlyCollection<dynamic> documents, CancellationToken token) => 
```

Start processor.

Should throw.

philipthomas-MSFT: Understood. So, we really should only be testing this when the `processorName` is the same. Anything else is outside the scope of this incident? Got it. We can talk more about it offline. In the meanwhile, I will change up to support same `processorName` using `GetAllLeasesAsync`.

ealsur: Correct. A different `processorName` naturally generates different leases. It is by design.

Customer could certainly create 2 processors, one for LastVersion, and another for AllVersions, different names, and they would be independent.

What we need to validate is that the same processor is not reused switching modes.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +80,98 @@ public override async Task StopAsync()
...
+
+            bool shouldThrowException = false;
+
+            // No lease documents. Return.
+            if (documentServiceLeases.Count == 0)
+            {
+                return;
+            }
+
+            DocumentServiceLease documentServiceLease = documentServiceLeases.FirstOrDefault(lease => lease.Id.Contains(monitoredDatabaseAndContainerRid));
```
ealsur, line 132: I don't think this is needed. GetAllLeasesAsync already only returns the leases for this particular Rid, procesorName combination.

ealsur: At this point, we know the Count is > 0, we probably don't need the LINQ query

```suggestion
            DocumentServiceLease documentServiceLease = documentServiceLeases[0];
```

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +80,98 @@ public override async Task StopAsync()
...
+            // AllVersionsAndDeletes. If the ChangeFeedProcessor's mode is not legacy, a CosmosException should thrown.
+            if (string.IsNullOrEmpty(documentServiceLease.Mode))
+            {
+                if (this.changeFeedLeaseOptions.Mode != ChangeFeedMode.LatestVersion)
+                {
+                    shouldThrowException = true;
+                }
+            }
+
+            string changeFeedProcessorMode = this.NormalizeChangeFeedProcessorMode(this.changeFeedLeaseOptions.Mode);
```
ealsur, line 151: Should this be on an else block?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: Then I can change Line 123 to `documentServiceLeases.FirstOrDefault()`, because I only need one.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +80,98 @@ public override async Task StopAsync()
...
+
+            // If the ChangeFeedProcessor mode is not the mode in the lease document, a CosmosException should be thrown.
+            if (string.Compare(documentServiceLease.Mode, changeFeedProcessorMode, StringComparison.OrdinalIgnoreCase) != 0)
```
ealsur, line 154: We could simplify this by doing the string.Compare within the helper method? To avoid doing 2 things (normalizing and then comparing) on the caller.

```
private bool VerifyChangeFeedProcessorMode(ChangeFeedMode changeFeedMode, DocumentServiceLease lease)
        {
            // obtain the comparable string
            return // the string.Compare with the resolution of the comparable string and the lease
        }
```

That simplifies the caller's code responsibility

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
```diff
@@ -70,5 +70,10 @@ internal abstract class DocumentServiceLease
         /// Gets or sets custom lease properties which can be managed from <see cref="LoadBalancingStrategy"/>.
         /// </summary>
         public abstract Dictionary<string, string> Properties { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="ChangeFeedMode"/>.
```
ealsur, line 75: Which type does this cref point to?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: Sure. I will allow it.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseStoreManagerBuilder.cs
```diff
@@ -22,7 +22,8 @@ internal class DocumentServiceLeaseStoreManagerBuilder
             ContainerInternal monitoredContainer,
             ContainerInternal leaseContainer,
             string leaseContainerPrefix,
-            string instanceName)
+            string instanceName,
+            ChangeFeedMode changeFeedMode = default)
```
ealsur, line 26: When should we expect this to be `default` and what is the `default` of the enum?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
philipthomas-MSFT: Meant to remove this cref. Originally, I was using the type but changed to string.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseStoreManagerBuilder.cs
philipthomas-MSFT: `ChangeFeedEstimatorIterator`. I will set the default on that to **LatestVersion** and remove the default optional parameter.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: Yes, I can see about changing this.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +80,99 @@ public override async Task StopAsync()
...
+                .ConfigureAwait(false);
+
+            // No lease documents. Return.
+
+            if (documentServiceLeases.Count == 0)
+            {
+                return;
+            }
+
+            DocumentServiceLease documentServiceLease = documentServiceLeases.FirstOrDefault();
```
ealsur, line 130: nit: At this point you already know the Count is > 0 (otherwise it would return on the previous `if`. So the `OrDefault` part is not adding much value.

To remove LINQ usage (> perf), this would yield the same (and you can remove the `null` check below)

```suggestion
            DocumentServiceLease documentServiceLease = documentServiceLeases[0];
```

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +80,99 @@ public override async Task StopAsync()
...
+            {
+                return;
+            }
+
+            // Mode attribute exists on lease document, but it is not set. legacy is always LatestVersion because
+            // AllVersionsAndDeletes does not exist. There should not be any legacy lease documents that are
+            // AllVersionsAndDeletes. If the ChangeFeedProcessor's mode is not legacy, a CosmosException should thrown.
+            // If the ChangeFeedProcessor mode is not the mode in the lease document, a CosmosException should be thrown.
+
+            bool shouldThrowException = string.IsNullOrEmpty(documentServiceLease.Mode)
```
ealsur, line 144: nit: This could be a single line:

```suggestion
            bool shouldThrowException = this.VerifyChangeFeedProcessorMode(
                    changeFeedMode: string.IsNullOrEmpty(documentServiceLease.Mode) ? ChangeFeedMode.LatestVersion : documentServiceLease.Mode,
                    leaseChangeFeedMode: documentServiceLease.Mode,
                    normalizedProcessorChangeFeedMode: out string normalizedProcessorChangeFeedMode);
```

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -5,15 +5,20 @@
 namespace Microsoft.Azure.Cosmos.ChangeFeed
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
     using System.Threading.Tasks;
+    using global::Azure;
```
ealsur, line 11: Please verify that the usings that are being added are actually being used and it's not Visual Studio adding them automatically.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +80,99 @@ public override async Task StopAsync()
...
+                : this.VerifyChangeFeedProcessorMode(
+                    changeFeedMode: this.changeFeedLeaseOptions.Mode,
+                    leaseChangeFeedMode: documentServiceLease.Mode,
+                    normalizedProcessorChangeFeedMode: out normalizedProcessorChangeFeedMode);
+
+            // If shouldThrowException is true, throw the CosmosException.
+
+            if (shouldThrowException)
+            {
+                throw CosmosExceptionFactory.CreateBadRequestException(
```
ealsur, line 158: Should we throw a CosmosException with Status Code HTTP 400 or simply an InvalidArgumentException or InvalidOperationException? 

CosmosException might be useful mostly in cases where either we represent a service response or the customer expects a service response from the operation. But in this case, this is CFP.StartAsync, it has no return type, and customers do not expect any service response.

Thoughts @kirankumarkolli ?

ealsur (commented):
Great progress! Just some nit comments and one question for the broader crew in terms of the type of exception we want to throw. Open to discussion on it.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: sure. have no qualms.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: since it is a nit, i will resolve, unless you feel strongly about it, then remove nit and we can talk offline?

ealsur: No strong feelings, I was just mentioning it to avoid the repeated lines for the other parameters

ealsur: Repeating the lines leads to things like `out string normalizedProcessorChangeFeedMode` and `out normalizedProcessorChangeFeedMode` on the other because you cannot add `string` to both

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: I could rock with the `InvalidArgumentException` personally. The use of the `CosmosException` seemed a bit much I suppose. And defaulting things like header because we don't have it was not ideal. So, I vote for `InvalidArgumentException`. But we can wait to see what @kirankumarkolli thinks if we must. Also, need another reviewer if Kiran does not mind.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: What I just committed will hopefully be a good compromise. @ealsur

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: If I don't get a weigh in, I will make the decision to change to `InvalidArgumentException` and we can always open this up for discussion later.

philipthomas-MSFT: And by `InvalidArgumentException` we really mean `ArgumentException` just to clarify.

ealsur (dismissed):
Let's just make sure the exception type is something the rest agrees on :) Great work

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorRunner.cs
```diff
@@ -139,11 +139,14 @@ private async Task InitializeLeaseStoreAsync()
...
-                    leaseContainerPrefix: leasePrefix,
-                    instanceName: ChangeFeedEstimatorRunner.EstimatorDefaultHostName);
+                DocumentServiceLeaseStoreManager documentServiceLeaseStoreManager = await DocumentServiceLeaseStoreManagerBuilder
+                    .InitializeAsync(
+                        monitoredContainer: this.monitoredContainer,
+                        leaseContainer: this.leaseContainer,
+                        leaseContainerPrefix: leasePrefix,
+                        instanceName: ChangeFeedEstimatorRunner.EstimatorDefaultHostName,
+                        changeFeedMode: ChangeFeedMode.LatestVersion)
+                    .ConfigureAwait(false);
```
kirankumarkolli, line 149: Is ConfigureAwait(false) needed? 
This was not followed before.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -75,23 +77,88 @@ public override async Task StopAsync()
...
+                    .InitializeAsync(
+                        this.monitoredContainer,
+                        this.leaseContainer,
+                        leaseContainerPrefix,
+                        this.instanceName,
+                        changeFeedMode: this.changeFeedProcessorOptions.Mode)
+                    .ConfigureAwait(false);
             }
 
+            await this.ChangeFeedModeSwitchingCheckAsync().ConfigureAwait(false);
```
kirankumarkolli, line 99: In theory this can be views as Argument validation for the LeaseManager right? 
Thoughts on including it part of it LeaseManager?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/Configuration/ChangeFeedLeaseOptions.cs
```diff
@@ -44,5 +44,10 @@ public ChangeFeedLeaseOptions()
         /// instances pointing at the same feed while using the same auxiliary collection.
         /// </summary>
         public string LeasePrefix { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="ChangeFeedMode"/>.
+        /// </summary>
+        public ChangeFeedMode Mode { get; set; }
```
kirankumarkolli, line 51: How about default to incremental? (Avoids defaults related code in callers)

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
```diff
@@ -70,5 +70,10 @@ internal abstract class DocumentServiceLease
         /// Gets or sets custom lease properties which can be managed from <see cref="LoadBalancingStrategy"/>.
         /// </summary>
         public abstract Dictionary<string, string> Properties { get; set; }
+
+        /// <summary>
+        /// Gets or sets the ChangeFeedMode.
+        /// </summary>
+        public abstract string Mode { get; set; }
```
kirankumarkolli, line 77: How about default to incremental? (Avoids defaults related code in callers)

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCoreEpk.cs
```diff
@@ -66,6 +66,9 @@ public override DateTime Timestamp
         [JsonProperty("properties")]
         public override Dictionary<string, string> Properties { get; set; } = new Dictionary<string, string>();
 
+        [JsonProperty("Mode", NullValueHandling = NullValueHandling.Ignore)]
```
kirankumarkolli, line 69: nit: small case (similar to others)

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorRunner.cs
philipthomas-MSFT: Thanks @kirankumarkolli for the review and feedback. It is much appreciated. Not sure why it was not called before I added my change to include **changeFeedMode** as a parameter, but since I was adding that, I included this as well. In short, there are some general performance improvements, and deadlock avoidance, and overall best practice to including `.ConfigureAwait(false)`, much like how it is done everywhere else in the codebase. We don't have to have a back and forth on it and since it is not a **nit:**, maybe you feel strongly about it. I am willing to remove it and we can talk offline so that I am not holding this PR hostage and in limbo.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: Thanks @kirankumarkolli for the review and feedback. It is much appreciated.

Adding it to **LeaseManager** since this is a check against the lease documents for the lease container makes sense to me. I can timebox an hour to see if I can refactor this and test it effectively.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCoreEpk.cs
philipthomas-MSFT: Thanks @kirankumarkolli for the review and feedback. It is much appreciated.

I think I was trying to be consistent with Java SDK in this regard in the overall format of the lease document itself.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
philipthomas-MSFT: Thanks @kirankumarkolli for the review and feedback. It is much appreciated.

Just acknowledging, but I have no thoughts yet on this yet. I'll assess the situation and respond accordingly.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
ealsur: The only caveat here is: The check should only happen when the Processor is starting, not the Estimator

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorRunner.cs
ealsur: For CFP codebase, we migrated most of it from the original repo. The original repo had it.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
philipthomas-MSFT: I moved it, please check it out and give me your thoughts. At this point, I think I can go either way. Thanks.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCoreEpk.cs
philipthomas-MSFT: If you feel strongly about the **nit:**, I can change it. Otherwise, I am going to resolve conversation.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
philipthomas-MSFT: This is property is not **defaulted** in related code in callers. Resolving conversation, but please feel free to reopen to extend this conversation.

kirankumarkolli: IMO its better be consistent with other existing property names on convention.
@ealsur thoughts?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCoreEpk.cs
kirankumarkolli: IMO its better be consistent with other existing property names on convention.
@ealsur thoughts?

kirankumarkolli (approved):
Please resolve the JSON casing part with Matias.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCoreEpk.cs
ealsur: The other properties, for example:

```
[JsonProperty("LeaseToken")]
public string LeaseToken { get; set; }

[JsonProperty("FeedRange", NullValueHandling = NullValueHandling.Ignore)]
public override FeedRangeInternal FeedRange { get; set; }
```

Seem to mostly use upper camel-case. Not really an issue for me.

ealsur (approved):

Issue #903, opened 2019-10-12
Add support for IAsyncEnumerable<T>


thomaslevesque:
**Is your feature request related to a problem? Please describe.**

The API for enumerating query results isn't great, because it forces the user to be aware of the fact that the results are fetched page by page:
```csharp
var iterator = ...
while (iterator.HasMoreResults)
{
    var page = await iterator.ReadNextAsync();
    foreach (var item in page)
    {
        ...
    }
}
```

**Describe the solution you'd like**

The `FeedIterator<T>` API is useful, because it gives you more information (e.g. request charge), but in most cases the user just wants the results, so it would be nice to have a higher level abstraction. The `IAsyncEnumerable<T>` interface seems perfect for this.

There could be an `AsAsyncEnumerable` extension method like this:

```csharp
public static async IAsyncEnumerable<T> AsAsyncEnumerable<T>(
    this FeedIterator<T> iterator,
    [EnumeratorCancellation] CancellationToken cancellationToken = default)
{
    while (iterator.HasMoreResults)
    {
        var page = await iterator.ReadNextAsync(cancellationToken);
        foreach (var item in page)
        {
            cancellationToken.ThrowIfCancellationRequested();
            yield return item;
        }
    }
}
```
The consuming code would be simplified to this:

```csharp
var iterator = ...
await foreach (var item in iterator.AsAsyncEnumerable())
{
    ...
}
```

**Describe alternatives you've considered**
N/A

**Additional context**

The `await foreach` feature was introduced in C# 8, so this would only be useful to people already using C# 8.
However, the `Microsoft.Bcl.AsyncInterfaces` package is compatible with netstandard2.0 and net461, so it wouldn't require running .NET Core 3.

ealsur:
We are working on this at the moment 😄

thomaslevesque:
Awesome!

ealsur:
@thomaslevesque See the V4 (preview) package https://docs.microsoft.com/en-us/azure/cosmos-db/create-sql-api-dotnet-v4#query-the-items

thomaslevesque:
Cool, thanks!

cfindlayvcs:
@ealsur , I just tried this with the latest version of the Cosmos DB API (3.15.0) on .NET Core 3.1 and `GetItemQueryIterator<>` does not return `IAsyncEnumerable`. I get this:

[...]

Issue #692, opened 2019-08-14
How could we handle ReadItemAsync if item not existed in collection


kevinding0218:
For the `ReadItemAsync()` method, if an item does not exist in the collection, will throw a CosmosException of status code "NotFound" (e.g, `catch(CosmosException ex) when (ex.StatusCode == System.Net.HttpStatusCode.NotFound)`), so instead of implementing the business logic in the `catch(...)` statement, could we just handle the _item not existed_ case outside the catch statement but in the try statment? For example, if the `ItemResponse<T>` could contain some flag to identity if the item existed or not? or is there a way just by check ` if (itemResponse<T>.Resource == null)` ?

Thank you very much for your help!

j82w:
Hi @kevinding0218 , 

Do you just need to verify the object exists or do you need the object for the business logic after the read?

j82w:
You could do something like this:

```
ItemResponse<ToDoActivity> response = null;
            try
            {
                response = await this.Container.ReadItemAsync<ToDoActivity>();
            }
            catch (CosmosException ex) when (ex.StatusCode == System.Net.HttpStatusCode.NotFound) { }

            if(response == null)
            {

            }
```

if(

kevinding0218:
@j82w , yes that is exactly what I'm currently doing, by using the status code with exception filter to handle the non-existed case. I am just wondering why would non-existed document consider as an exception rather than just returning a null object? So that we don't need to handle it in the exception filter case.

j82w:
This is a very long running debate. We initially started v3 out with it returning null. We got feedback from central .NET team and central SDK team that it should throw. The reason why is 404 is considered an exception according to the [http standard](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) and most developers expect it throw. We had several users report it thinking it was a bug.

If you want you can use the Stream APIs. They only throw argument exceptions, and return all other exceptions as a status code.

T item = null;
using(ResponseMessage responseMessage = await this.Container.ReadItemStreamAsync()){
if(responseMessage.StatusCode == System.Net.HttpStatusCode.NotFound){
   // Do other logic
}

 item = jsonSerializer.Deserialize<T>(responseMessage.Content);
}

bartelink:
To save me thinking/looking it up a) what does v2 do b) can/does the xmldoc for `ReadItemAsync` mention this as being the advised approach (I agree with the stance)

j82w:
[...]

PR #4355, opened 2024-03-18
Implement IAsyncEnumerable on CosmosLinqQuery

onionhammer:
# Pull Request Template

## Description

This updates CosmosLinqQuery to support IAsyncEnumerable, a new 'AsAsyncEnumerable' extension method, and a test method to utilize the new extension method

## Type of change


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
```diff
@@ -141,6 +141,31 @@ public void LinqQueryToIteratorBlockTest(bool isStreamIterator)
...
+                .AsAsyncEnumerable();
+
+            int found = 0;
+            await foreach (ToDoActivity item in query)
+            {
+                Assert.IsNotNull(item);
+                ++found;
+            }
+
+            Assert.IsTrue(found > 0);
```
adityasa, line 166: >found > 0 [](http://example.com/codeflow?start=26&length=9)

Nit, Assert.AreEqual may be a bit stronger?

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqQuery.cs
```diff
@@ -283,5 +283,23 @@ private FeedIteratorInlineCore<T> CreateFeedIterator(bool isContinuationExpected
                 this.responseFactory.CreateQueryFeedUserTypeResponse<T>),
                 this.container.ClientContext);
         }
+
+        public async IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
adityasa: Closer to GetEnumerator would be good.

adityasa:
>         public IEnumerator<T> GetEnumerator()

This should now delegate to the Async implementation (after the L106-L110 check).

---
Refers to: Microsoft.Azure.Cosmos/src/Linq/CosmosLinqQuery.cs:104 in 9b513c7. [](commit_id = 9b513c70d4d224ae43beeac3888c9e86974c93b4, deletion_comment = False)

bartelink:
How about if it yielded an `IAsyncEnumerable<FeedResponse<T>>` instead ? My problem with an API like this is that it makes it impossible to dig down and get relevant context like e.g. the request charges incurred.

For instance, I'd be able to use such a slightly lower level mechanism as a building block in wrappers that I find critical in successful usage:

```
    let private taskEnum (iterator: FeedIterator<'T>) = taskSeq {
        while iterator.HasMoreResults do
            let! response = iterator.ReadNextAsync()
            yield response.Diagnostics.GetClientElapsedTime(), response.RequestCharge, response.Resource }

    let fetch<'T> (desc: string) (container: Container) (queryDefinition: QueryDefinition) = taskSeq {
        if Log.IsEnabled Serilog.Events.LogEventLevel.Debug then Log.Debug("CosmosQuery.fetch {desc} {query}", desc, queryDefinition.QueryText)
        let sw = System.Diagnostics.Stopwatch.StartNew()
        let iterator = container.GetItemQueryIterator<'T>(queryDefinition)
        let mutable responses, items, totalRtt, totalRu = 0, 0, TimeSpan.Zero, 0.
        try for rtt, rc, response in taskEnum iterator do
                responses <- responses + 1
                totalRu <- totalRu + rc
                totalRtt <- totalRtt + rtt
                for item in response do
                    items <- items + 1
                    yield item
        finally Log.Information("CosmosQuery.fetch {desc} found {count} ({trips}r, {totalRtt:N0}ms) {rc}RU {latency}ms",
                                desc, items, responses, totalRtt.TotalMilliseconds, totalRu, sw.ElapsedMilliseconds) }
```

Same for `GetItemStreamQueryIterator` ?

onionhammer:
> How about if it yielded an `IAsyncEnumerable<FeedResponse<T>>` instead ? My problem with an API like this is that it makes it impossible to dig down and get relevant context like e.g. the request charges incurred.

You're not forced to use it? But this opens up capabilities for improved integration with other libraries, such as HotChocolate. Enumerating a "FeedResponse" would make observers have to know about the Cosmos SDK in a way that the current implementation does not.

bartelink:
I am not saying this isnt useful. But exposing `IAsyncEnumerable<Response<T>` instead of `IAsyncEnumerable<T>` would be a very useful counterpart as it means you have a way to access all that context info as you inevitably wind up wanting to do, without having to go and program things in an entirely different way.

Not sure if implementing two IAsyncEnumerable interfaces will make a mess; if it did, maybe that could be `await foreach(var r in query.Responses<T>())`

My point is simply to consider what happens the minute you need to add instrumentation - it's not disssimilar to https://github.com/Azure/azure-cosmos-dotnet-v3/issues/692#issuecomment-2014681616 - thinking about what happens after someone hits the limits of what is exposed - do they fall off a cliff and have to work out lots of things that are a big jump.

None of this is to invalidate the usefulness of what you're adding; having this in the box is absolutely transformative and a no brainer to add.

onionhammer:
> I am not saying this isnt useful. But exposing `IAsyncEnumerable<Response<T>` instead of `IAsyncEnumerable<T>` would be a very useful counterpart as it means you have a way to access all that context info as you inevitably wind up wanting to do, without having to go and program things in an entirely different way.

I think this would be useful as an additional extension method to elide some of the iterator ugliness

dersia:
@onionhammer

thanks for this! And I appreciate your work here, but this PR would not Close https://github.com/Azure/azure-cosmos-dotnet-v3/issues/903 because it is adding `IAsyncEnumerable` only to the Linq provider. so, please change that in your description or the github bot might accidentally close that issue.

Also I would like to point out that I would love for `IAsyncEnumerable` to work all the way to the db. what I mean by that is that it would be great if the db would make use of http-streaming and we could use `IAsyncEnumerable` on that. this would really improve the performance and make this a top api.

having said that, I don't know if the server supports http-streaming yet.

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqExtensions.cs
```diff
@@ -284,6 +284,32 @@ public static QueryDefinition ToQueryDefinition<T>(this IQueryable<T> query)
...
+        /// <code language="c#">
+        /// <![CDATA[
+        /// IOrderedQueryable<ToDoActivity> linqQueryable = this.Container.GetItemLinqQueryable<ToDoActivity>();
+        /// IAsyncEnumerable<ToDoActivity> asyncEnumerable = linqQueryable.Where(item => (item.taskNum < 100)).AsAsyncEnumerable();
+        /// ]]>
+        /// </code>
+        /// </example>
+        public static IAsyncEnumerable<T> AsAsyncEnumerable<T>(this IQueryable<T> query)
+        {
+            if (query is IAsyncEnumerable<T> asyncEnumerable)
```
adityasa, line 305: >IAsyncEnumerable [](http://example.com/codeflow?start=25&length=16)

This should be CosmosLinqQuery, no?

adityasa:
>         public IEnumerator<T> GetEnumerator()

Is this being addressed?

---
In reply to: [2012904380](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4355#issuecomment-2012904380) [](http://example.com/codeflow?ancestors=2012904380)

---
Refers to: Microsoft.Azure.Cosmos/src/Linq/CosmosLinqQuery.cs:104 in 9b513c7. [](commit_id = 9b513c70d4d224ae43beeac3888c9e86974c93b4, deletion_comment = False)

onionhammer:
Hi @adityasa I must have missed that one - apologies. I'm not sure if I agree that this is necessary/good; I assume to be DRY?

The code is different enough where it would be annoying to call synchronously, unless you have a better way

```C#
            IAsyncEnumerator<T> asyncEnumerator = this.GetAsyncEnumerator();
            bool hasMoreResults = true;

            while (hasMoreResults)
            {
                ValueTask<bool> moveNextTask = asyncEnumerator.MoveNextAsync();

                if (!moveNextTask.IsCompletedSuccessfully)
                {
                    #pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
                    if (!TaskHelper.InlineIfPossible(() => moveNextTask.AsTask(), null).GetAwaiter().GetResult())
                    {
                        break;
                    }
                    #pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
                }
                else if (!moveNextTask.Result)
                {
                    break;
                }

                yield return asyncEnumerator.Current;
            }

```

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqExtensions.cs
onionhammer: Sure, that should work; we cant return it because it's internal though

adityasa: What I meant is we should check CosmosLinqQuery type and only then return successfully (base type), otherwise throw an exception since we don't support any other AsyncEnumerable.

onionhammer: @adityasa, Ok, done.

adityasa:
No worries. I have a simple sample below. Is something like this possible?

public IEnumerator<int> GetEnumerator()
{
    IAsyncEnumerator<int> asyncEnumerator = this.GetAsyncEnumerator();
    while(asyncEnumerator.MoveNextAsync().Result)
    {
        yield return asyncEnumerator.Current;
    }
}
public async IAsyncEnumerator<int> GetAsyncEnumerator(CancellationToken cancellationToken = default)
{
    yield return 1;
    yield return 2;
    yield return 3;
}

---
In reply to: [2066546588](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4355#issuecomment-2066546588) [](http://example.com/codeflow?ancestors=2066546588)

onionhammer:
> No worries. I have a simple sample below. Is something like this possible?
> 
> public IEnumerator GetEnumerator() { IAsyncEnumerator asyncEnumerator = this.GetAsyncEnumerator(); while(asyncEnumerator.MoveNextAsync().Result) { yield return asyncEnumerator.Current; } } public async IAsyncEnumerator GetAsyncEnumerator(CancellationToken cancellationToken = default) { yield return 1; yield return 2; yield return 3; }
> 
> In reply to: [2066546588](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4355#issuecomment-2066546588)

This code does not dispose the enumerator (which is IAsyncDisposable)

adityasa:
The example above does not achieve disposal, so it may not be sufficient.

Maybe use [https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskasyncenumerableextensions.toblockingenumerable?view=net-8.0](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskasyncenumerableextensions.toblockingenumerable?view=net-8.0)


---
In reply to: [2077739958](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4355#issuecomment-2077739958) [](http://example.com/codeflow?ancestors=2077739958,2066546588)

onionhammer:
@adityasa are you just trying to make it DRY? I think the current implementation (keeping GetEnumerator as it is today) is much simpler

adityasa:
Yes. In general, I would like to avoid inheritance (and favor composition) and for cases like this where there isn't a suitable option (since we already implemented IEnumerable and IAsyncEnumerable is more general), at least avoid code duplication.

---
In reply to: [2077756442](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4355#issuecomment-2077756442) [](http://example.com/codeflow?ancestors=2077756442)

onionhammer:
> Yes. In general, I would like to avoid inheritance (and favor composition) and for cases like this where there isn't a suitable option (since we already implemented IEnumerable and IAsyncEnumerable is more general), at least avoid code duplication.
> 
> In reply to: [2077756442](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4355#issuecomment-2077756442)

Okay; if you still want me to move ahead with that change, let me know - I think it increases the code-size and complexity to be "DRY"

adityasa (approved):
:shipit:

adityasa:
I agree that the full implementation increases the complexity and compared to that current implementation is sufficient.
So unless we can replace sync implementation by delegating to BlockingEnumerable over async method (or something simpler), I am ok with leaving it as is.

---
In reply to: [2108024057](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4355#issuecomment-2108024057) [](http://example.com/codeflow?ancestors=2108024057)

onionhammer:
@adityasa any status update on this?

kirankumarkolli:
Closing due to in-activity, feel free to re-open as needed

onionhammer:
> Closing due to in-activity, feel free to re-open as needed

And whose fault is that? 🙄

PR #4359, opened 2024-03-20
[Internal] Query: Fixes occasional hang while querying using partial partition key against a sub-partitioned container

### LINKED ISSUE #4326: Feed iterator enters an infinite loop after a physical partition split has occurred
**Describe the bug**

When I run a targeted multi-partition query with a partition key that includes the first level of subpartitioning, the `FeedIterator<T>.ReadNextAsync` method never completes after a physical partition split has occurred. The iterator enters an infinite loop reading partition key ranges and the application eventually runs out of memory.

**To Reproduce**

I wrote [a console application](https://github.com/Azure/azure-cosmos-dotnet-v3/files/14375118/PartitionSplitIssue.zip) to reproduce the issue. Please download the solution, open the `Program.cs` file and specify a connection string to an Azure Cosmos DB for NoSQL account. The application will create a new database with one container and one physical partition. It will then upsert about 20 GB of data and attempt to read the first 10 documents from two logical partitions. At this point there should be two physical partitions. You can run the application multiple times. It will not upsert data if the container already exists.

The data consists of 40,000 documents (articles) with the following properties:

- Id : string
- CustomerId : Guid
- SectionId : Guid
- Content : string

There are two customers. Each customer has 20,000 articles organized in 2,000 sections (10 articles per section).

The container uses a hierarchical partition key `/customerId, /sectionId`, so there are 4,000 hierarchical logical partitions with 10 documents.

Finally, the application will execute a query `SELECT * FROM c OFFSET 0 LIMIT 10`, using the customer ID as the partition key in the request options. When a response is returned, it displays the number of documents and the total request charge.

**Expected behavior**

The `FeedIterator<T>.ReadNextAsync` method returns a response with the first 10 documents.

**Actual behavior**

The `FeedIterator<T>.ReadNextAsync` method never completes and the application eventually runs out of memory.

**Environment summary**

SDK Version: 3.38.1 (version 3.37.1 has the same issue)
OS Version: Windows 11 Enterprise (10.0.22631 Build 22631)

**Additional context**

You can examine [the trace and the diagnostic string](https://github.com/Azure/azure-cosmos-dotnet-v3/files/14375134/diagnostics.zip) from another incident. It happens in both `Direct` and `Gateway` mode.

adityasa:
# Pull Request Template

## Description

This change contains the fix for occasional hang that can occur while querying using partition partition key against a sub-partitioned container. Let's assume a container with partition key definition "/id", "/value1", "/value2". Unlike single-key partition definition, for hierarchical partitions, partition splits occurs such that for a certain value of "/id" (say 2), more than one partition contains the data for this (partial) partition value. When a query is issued with partial partition key ("/id" = 2 in this case), query enumerator (QueryPartitionRangePageAsyncEnumerator L50) attempts to query a feedrange based on this partial value that isn't fully contained inside a single physical partition in the backend. Since such a range cannot be resolved to any single physical partition, enumerator receives a GoneException from the underlying stack. The GoneException is caught by upstream (CrossPartitionRangePageAsyncEnumerator L114) which attempts to resolve the physical partition ranges by consulting the routing map. This is typically a no-op since the GoneException isn't received because of a change in state of the backend.
This doesn't actually help since CrossPartitionRangePageAsyncEnumerator attempts to locate FeedRangeEpk (without the partial partition key, which honors partition boundary) inside a physical partition range - which always succeeds and further downstream (RequestInvokerHandler.ResolveFeedRangeBasedOnPrefixContainerAsync and RequestInvokerHandler L244) ignores this FeedRangeEpk and instead queries using partial partition key (FeedRangePartitionKey) which spans across the partition boundary. The later continues to cause another GoneException and cycle repeats.

This fix limits the epk range for QueryPartitionRangePageAsyncEnumerator such that it does not span across the single physical partition range. It also uses FeedRangeEpk to represent this new partition range, which causes downstream to no longer perform extraneous operations (such as ResolveFeedRangeBasedOnPrefixContainerAsync).

Several parts of the code are suspect here and a more comprehensive fix may be needed to handle all cases generally, _outside this change_.
1. RequestInvokerHandler.ResolveFeedRangeBasedOnPrefixContainerAsync should not happen. Ignoring the FeedRange supplied by the upstream blindly is plain wrong. Instead upstream enumerators etc. should provide explicit range.
2. At the very least using type of FeedRangeInternal (Epk v/s Partition) to determine extra processing in RequestInvokerHandler should not exist. This should be an explicit action.
3. It's not clear how filtering occurs in the backend when partial partition key is specified and request lands on a physical partition where data for other logical partitions also exists. We need to replicate this behavior in the test infrastructure in the SDK so that such scenarios can be tested broadly.
4. Query enumerators should not take a dependency on partition key. Instead partition key -> physical partition resolution should take place upstream before they are instantiated.

## Type of change

## Closing issues

To automatically close an issue: closes #4326

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/Parallel/QueryPartitionRangePageAsyncEnumerator.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/Parallel/QueryPartitionRangePageAsyncEnumerator.cs
@@ -68,66 +68,77 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.CrossPartition.Parallel
             // We sadly need to check the partition key, since a user can set a partition key in the request options with a different continuation token.
             // In the future the partition filtering and continuation information needs to be a tightly bounded contract (like cross feed range state).
             FeedRangeInternal feedRange = this.FeedRangeState.FeedRange;
-
-            if (feedRange is FeedRangeEpk feedRangeEpk && this.partitionKey.HasValue)
+            if (this.partitionKey.HasValue)
             {
-                if (this.containerQueryProperties.EffectiveRangesForPartitionKey == null ||
-                    this.containerQueryProperties.EffectiveRangesForPartitionKey.Count == 0)
+                // ISSUE-HACK-adityasa-3/25/2024 - We should not update the original feed range inside this class.
+                // Instead we should guarantee that when enumerator is instantiated it is limited to a single physical partition.
+                // Ultimately we should remove enumerator's dependency on PartitionKey.
+                if ((this.containerQueryProperties.PartitionKeyDefinition.Paths.Count > 1) &&
+                    (this.partitionKey.Value.InternalKey.Components.Count != this.containerQueryProperties.PartitionKeyDefinition.Paths.Count) &&
+                    (feedRange is FeedRangeEpk feedRangeEpk))
                 {
-                    throw new InvalidOperationException(
-                        "EffectiveRangesForPartitionKey should be populated when PK is specified in request options.");
-                }
-
-                foreach (Documents.Routing.Range<String> epkForPartitionKey in
-                    this.containerQueryProperties.EffectiveRangesForPartitionKey)
-                {
-                    if (Documents.Routing.Range<String>.CheckOverlapping(
-                            feedRangeEpk.Range,
-                            epkForPartitionKey))
+                    if (this.containerQueryProperties.EffectiveRangesForPartitionKey == null ||
+                        this.containerQueryProperties.EffectiveRangesForPartitionKey.Count == 0)
                     {
-                        if (!feedRangeEpk.Range.Equals(epkForPartitionKey))
-                        {
-                            String overlappingMin;
-                            bool minInclusive;
-                            String overlappingMax;
-                            bool maxInclusive;
-
-                            if (Documents.Routing.Range<String>.MinComparer.Instance.Compare(
-                                    epkForPartitionKey,
-                                    feedRangeEpk.Range) < 0)
-                            {
-                                overlappingMin = feedRangeEpk.Range.Min;
-                                minInclusive = feedRangeEpk.Range.IsMinInclusive;
-                            }
-                            else
-                            {
-                                overlappingMin = epkForPartitionKey.Min;
-                                minInclusive = epkForPartitionKey.IsMinInclusive;
-                            }
-
-                            if (Documents.Routing.Range<String>.MaxComparer.Instance.Compare(
-                                    epkForPartitionKey,
-                                    feedRangeEpk.Range) > 0)
-                            {
-                                overlappingMax = feedRangeEpk.Range.Max;
-                                maxInclusive = feedRangeEpk.Range.IsMaxInclusive;
-                            }
-                            else
-                            {
-                                overlappingMax = epkForPartitionKey.Max;
-                                maxInclusive = epkForPartitionKey.IsMaxInclusive;
-                            }
-
-                            feedRange = new FeedRangeEpk(
-                                new Documents.Routing.Range<String>(
-                                    overlappingMin,
-                                    overlappingMax,
-                                    minInclusive,
-                                    maxInclusive));
-                        }
-
-                        break;
+                        throw new InvalidOperationException(
+                            "EffectiveRangesForPartitionKey should be populated when PK is specified in request options.");
                     }
+
+                    foreach (Documents.Routing.Range<String> epkForPartitionKey in
+                        this.containerQueryProperties.EffectiveRangesForPartitionKey)
+                    {
+                        if (Documents.Routing.Range<String>.CheckOverlapping(
+                                feedRangeEpk.Range,
+                                epkForPartitionKey))
+                        {
+                            if (!feedRangeEpk.Range.Equals(epkForPartitionKey))
+                            {
+                                String overlappingMin;
+                                bool minInclusive;
+                                String overlappingMax;
+                                bool maxInclusive;
+
+                                if (Documents.Routing.Range<String>.MinComparer.Instance.Compare(
+                                        epkForPartitionKey,
+                                        feedRangeEpk.Range) < 0)
+                                {
+                                    overlappingMin = feedRangeEpk.Range.Min;
+                                    minInclusive = feedRangeEpk.Range.IsMinInclusive;
+                                }
+                                else
+                                {
+                                    overlappingMin = epkForPartitionKey.Min;
+                                    minInclusive = epkForPartitionKey.IsMinInclusive;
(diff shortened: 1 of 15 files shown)


sboshra (approved):
:shipit:

Maya-Painter (approved):

neildsh:
>                     tryCreatePipelineStage = TryCreateSpecializedDocumentQueryExecutionContext(documentContainer, cosmosQueryContext, inputParameters, targetRanges, containerQueryProperties, partitionedQueryExecutionInfo);

[Nit] Please indent for readability

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs:301 in d23b9b7. [](commit_id = d23b9b722e5c290628d3e546efa248a13e7fa183, deletion_comment = False)

neildsh:
Regarding (3) in the description, the EPK range headers are used to indicate EPK range filtering to the backend. That, or the query itself can have a filter predicate.
Regarding (4), currently, I dont think query enumerators have a dependency on the partition key

adityasa:
Neil - yes, I learnt about the startEpkRange and endEpkRange headers while working on this change. Do you know if the DocumentResourceContentEnumerator honors this range in the backend for Query (or something else) ?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/Parallel/QueryPartitionRangePageAsyncEnumerator.cs
```diff
@@ -42,15 +46,103 @@ protected override Task<TryCatch<QueryPage>> GetNextPageAsync(ITrace trace, Canc
...
+
+        /// <summary>
+        /// Updates the FeedRange to limit the scope of this enumerator to single physical partition.
+        /// Generally speaking, a subpartitioned container can experience split partition at any level of hierarchical partition key.
+        /// This could cause a situation where more than one physical partition contains the data for a partial partition key.
+        /// Currently, enumerator instantiation does not honor physical partition boundary and allocates entire epk range which could spans across multiple physical partitions to the enumerator.
+        /// Since such an epk range does not exist at the container level, Service generates a GoneException.
+        /// This method restrics the range of each container by shrinking the ends of the range so that they do not span across physical partition.
+        /// </summary>
+        private FeedRangeInternal LimitFeedRangeToSinglePartition()
```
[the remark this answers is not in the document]
adityasa: There is a non-trivial amount of logic with a number of places doing this calculation in CosmosQueryExecutionContextFactory (and other classes). I would rather keep that change separate in the interest of unblocking the cx in time.

sboshra (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/Parallel/QueryPartitionRangePageAsyncEnumerator.cs
```diff
@@ -42,15 +46,103 @@ protected override Task<TryCatch<QueryPage>> GetNextPageAsync(ITrace trace, Canc
...
+            // In the future the partition filtering and continuation information needs to be a tightly bounded contract (like cross feed range state).
+            FeedRangeInternal feedRange = this.FeedRangeState.FeedRange;
+            if (this.partitionKey.HasValue)
+            {
+                // ISSUE-HACK-adityasa-3/25/2024 - We should not update the original feed range inside this class.
+                // Instead we should guarantee that when enumerator is instantiated it is limited to a single physical partition.
+                // Ultimately we should remove enumerator's dependency on PartitionKey.
+                if ((this.containerQueryProperties.PartitionKeyDefinition.Paths.Count > 1) &&
+                    (this.partitionKey.Value.InternalKey.Components.Count != this.containerQueryProperties.PartitionKeyDefinition.Paths.Count) &&
+                    (feedRange is FeedRangeEpk feedRangeEpk))
```
neildsh, line 78: >feedRange is FeedRangeEpk feedRangeEpk [](http://example.com/codeflow?start=21&length=38)

We would run into the same issue if someone gave us a `FeedRangePartitionKey` that is partially specified

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/Parallel/QueryPartitionRangePageAsyncEnumerator.cs
```diff
@@ -42,15 +46,103 @@ protected override Task<TryCatch<QueryPage>> GetNextPageAsync(ITrace trace, Canc
...
+                {
+                    if (this.containerQueryProperties.EffectiveRangesForPartitionKey == null ||
+                        this.containerQueryProperties.EffectiveRangesForPartitionKey.Count == 0)
+                    {
+                        throw new InvalidOperationException(
+                            "EffectiveRangesForPartitionKey should be populated when PK is specified in request options.");
+                    }
+
+                    foreach (Documents.Routing.Range<String> epkForPartitionKey in
```
neildsh, line 87: >String [](http://example.com/codeflow?start=53&length=6)

[Nit][Coding convention] Please use `string` everywhere

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/Parallel/QueryPartitionRangePageAsyncEnumerator.cs
```diff
@@ -42,15 +46,103 @@ protected override Task<TryCatch<QueryPage>> GetNextPageAsync(ITrace trace, Canc
...
+                                    overlappingMax = epkForPartitionKey.Max;
+                                    maxInclusive = epkForPartitionKey.IsMaxInclusive;
+                                }
+
+                                feedRange = new FeedRangeEpk(
+                                    new Documents.Routing.Range<String>(
+                                        overlappingMin,
+                                        overlappingMax,
+                                        minInclusive,
+                                        maxInclusive));
```
neildsh, line 132: Instead of doing this, why not just throw ? The upper layers expect the whole FeedRange to be queried. If we are silently overwriting this, it means that we are dropping results from the physical range that was truncated. It would be better to error than return incorrect results

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/Parallel/QueryPartitionRangePageAsyncEnumerator.cs
neildsh: Apparently not, because of `RequestInvokerHandler.ResolveFeedRangeBasedOnPrefixContainerAsync` which changes this to a `FeedRangeEpk` and causes a `GoneException`

PR #4361, opened 2024-03-20
Performance: Refactors query prefetch mechanism

kevin-montrose:
## Description

Reworks `ParallelPrefetch.PrefetchInParallelAsync` to reduce allocations.

This came out of profiling an application, and discovering that this method is allocating approximately as many bytes worth of `Task[]` as the whole application is creating `byte[]` for IO.  This is because `Task.WhenAny(...)` is a) used in a loop b) makes a defensive copy of the passed `Task`s.

This version is substantially more complicated, and accordingly there are a lot of tests in this PR (code coverage is 100% of lines and blocks).  Special attention was paid to exception and cancellation cases.

#### Greatly Reduced Allocations

In my benchmarking, anywhere from 30% to 99% depending on the total number of `IPrefetcher`s used.

More benchmarking discussion is at the bottom of this PR.

#### Special Casing For `maxConcurrency`

When `== 0` we do no work, more efficiently than current code.
When `== 1` we devolve to a foreach, which is just about ideal.

#### Special Casing When Only 1 `IPrefetcher`

We accept an `IEnumerable<IPrefetcher>`, but when that is only going to yield one `IPrefetcher` a lot of work (even with the old code) is pointless.  New code detects this case (generically, it doesn't look for specific types) and devolves into a single `await`.

#### Prompter Starting Of Next Task

Old code starts at most one task per pass through the while loop, so if multiple `Task`s are sitting there completed there's a fair amount a work done before they are all replaced with active `Task`s.

New code has the completed `Task` start its replacement, which should keep us closer to `maxConcurrency` active `Task`s.

#### `IEnumerator<IPrefetcher>` Disposed

Small nit, but the old code doesn't dispose the `IEnumerator<IPrefetcher>`.  While unlikely, this can put more load on the finalizer thread or potentially leak resources.

### Outline

 - There are 4 paths through the method now
   * `maxConcurrency == 0` just returns
   * `maxConcurrency == 1` is just a `foreach`
   * `maxConcurrency <= BatchSize` is more complicated
     + Up to `BatchSize` `IPrefetcher`s are loaded into a rented array
     + `Task`s are then started for each of those `IPrefetcher`s
        - These `Task`s grab and start the next `IPrefetcher` of the `IEnumerator<IPrefetcher>` when they finish with their last one
     + Every started `Task` is then awaited in order
   * `maxConcurrency > BatchSize` reuses a lot of the above case, but is still more complicated
      + Up to `BatchSize` `IPrefetchers` are loaded and started as above
         - Also as above, the `Task`s grab and start the next `IPrefetcher` when they finish with one
      + But we continue trying to start new batches of `IPrefetchers` (up to `maxConcurrency`) while there are active `Task`s
      + All of this is tracked in a pseudo-linked-list of rented `object[]`, which is awaited in turn once `maxConcurrency` is reached (or the `IEnumerator<T>` finishes)

We distinguish between the two `maxConcurrency > 1` cases to avoid allocating very large arrays, and to make sure we start some prefetches fairly promptly even when `maxConcurrency` is very large.  `BatchSize` is, somewhat arbitrarily, `512` - any value `> 1` and `< 8,192` would be valid.

## Type of change

Sort of a bug I guess?  Current code allocates a lot more than you'd expect.

## Benchmarking

Standard caveats about micro-benchmarking apply, but I did [some benchmarking](https://gist.github.com/kevin-montrose/c2ffd6b3a5bdd5f3d80ee3fc3eaa3571) to see how this stacks up versus the old code.

**TL;DR** - across the board improvements in allocations, with no wall clock regressions in what I believe is the common case.  There are some narrow, less common cases, where small wall clock regressions are observed.

I consider the primary case here when the `IPrefetcher` actually goes async, and takes some non-trivial time to do its work.  My expectation is that the two versions of the code should have about the same wall-clock time when `# tasks > maxConcurrency`, with the new code edging out old as `# tasks` increases.

That said, I did also test the synchronous completion case, and the "goes async, but then completes immediately" cases to make sure performance wasn't terrible.

In all cases I expect the new code to perform fewer allocations than the old.

Summarizing some results (the full data is in the previous link)...

Here's old vs new on .NET 6 where the `IPrefetcher` is just an `await Task.Delay(1)` (`< 1` is an improvement):
image
<sub>As expected, wall clock time is basically unaffected (the delay dominates) but allocations are improved across the board.  The benefits of improved replacement `Task` starting logic are visible at the very extreme ends of max concurrency and prefetcher counts.</sub>

Again, but this time `IPrefetcher` just `return default;`s so everything completes synchronously:
image
<sub>We see here that between 2 and 8 tasks there are configurations with _wall clock_ regressions.  I could try and improve that, but I believe "all synchronous completions" is fantastically rare, so it's not worth the extra code complications.</sub>

And finally, `IPrefetcher` is just `await Task.Yield();` so everything completes almost immediately but forces all the async completion machinery to run:
image
<sub>Similarly, between 4 and 8 tasks there are some _wall clock_ regressions.  While more realistic than the "all synchronous"-case, I think this would still be pretty rare - most `IPrefetcher`s should be doing real work after some asynchronous operation.</sub>

Since we target netstandard, I also benchmarked under a Framework version (4.6.2) and the results are basically the same:
image


kevin-montrose:

Microsoft.Azure.Cosmos/src/Pagination/ParallelPrefetch.cs
```diff
@@ -5,73 +5,850 @@
...
+        /// </summary>
+        private sealed class CommonPrefetchState
+        {
+            // we also use this to signal if we're finished enumerating, to save space
+            private IEnumerator<IPrefetcher> enumerator;
+
+            /// <summary>
+            /// If this is true, it's a signal that new work should not be queued up.
+            /// </summary>
+            internal bool FinishedEnumerating
```
neildsh, line 31: >internal [](http://example.com/codeflow?start=12&length=8)

[Nit][Query SDK Coding convention] members of private classes are either `private` or `public`, never `internal`

Microsoft.Azure.Cosmos/src/Pagination/ParallelPrefetch.cs
```diff
@@ -5,73 +5,850 @@
...
+        }
+
+        /// <summary>
+        /// For testing purposes, provides ways to instrument <see cref="PrefetchInParallelAsync(IEnumerable{IPrefetcher}, int, ITrace, CancellationToken)"/>.
+        /// 
+        /// You shouldn't be using this outside of test projects.
+        /// </summary>
+        internal sealed class ParallelPrefetchTestConfig : ITrace
+        {
+            private ITrace innerTrace;
```
neildsh, line 106: >innerTrace [](http://example.com/codeflow?start=27&length=10)

Please consider making this `readonly`

Microsoft.Azure.Cosmos/src/Pagination/ParallelPrefetch.cs
```diff
@@ -5,73 +5,850 @@
...
+
+        /// <summary>
+        /// Number of tasks started at one time, maximum, when working through prefetchers.
+        /// 
+        /// Also used as a the limit between Low and High concurrency implementations.
+        /// 
+        /// This number should be reasonable large, but less than the point where a 
+        /// Task[BatchLimit] ends up on the LOH (which will be around 8,192).
+        /// </summary>
+        private const int BatchLimit = 512;
```
neildsh, line 210: >BatchLimit [](http://example.com/codeflow?start=26&length=10)

Can this be moved to the top of the file so it's easier to find?

Microsoft.Azure.Cosmos/src/Pagination/ParallelPrefetch.cs
```diff
@@ -5,73 +5,850 @@
...
+        {
+            return trace.StartChild(name: "Prefetching", TraceComponent.Pagination, TraceLevel.Info);
+        }
+
+        /// <summary>
+        /// Helper for grabbing a reusable array.
+        /// </summary>
+        private static T[] RentArray<T>(ParallelPrefetchTestConfig config, int minSize, bool clear)
+        {
+            T[] ret;
```
neildsh, line 268: >ret [](http://example.com/codeflow?start=16&length=3)

[Nit][Query SDK Coding Convention] Avoid abbreviations like `ret`. Please consider using `result` or `array`

Microsoft.Azure.Cosmos/src/Pagination/ParallelPrefetch.cs
```diff
@@ -5,73 +5,850 @@
...
+                                    // proceed to the next item
+                                    //
+                                    // we need this lock because at this point there
+                                    // are other Tasks potentially also looking to call
+                                    // e.MoveNext()
+                                    lock (innerCommonState)
+                                    {
+                                        // this can have transitioned to null since we last checked
+                                        // so this is basically double-check locking
+                                        IEnumerator<IPrefetcher> e = innerCommonState.Enumerator;
```
neildsh, line 400: >e [](http://example.com/codeflow?start=65&length=1)

[Nit][Query SDK Coding Convention] `enumerator` please

Microsoft.Azure.Cosmos/src/Pagination/ParallelPrefetch.cs
```diff
@@ -5,73 +5,850 @@
...
+
+                        // if we encounter any error, we remember it
+                        // but as soon as we start a single task we've got
+                        // to see most of this code through so we observe them
+                        ExceptionDispatchInfo capturedException = null;
+
+                        while (true)
+                        {
+                            // start and store the last set of Tasks we got from FillPrefetcherBuffer
+                            for (int toStartIx = 0; toStartIx < bufferedPrefetchers; toStartIx++)
```
neildsh, line 696: >toStartIx [](http://example.com/codeflow?start=37&length=9)

[Nit][Query SDk Coding Convention] Please consider `toStartIndex` instead of `toStartIx`

Microsoft.Azure.Cosmos/src/Pagination/ParallelPrefetch.cs
```diff
@@ -5,73 +5,850 @@
...
+
+                            if (toAwaitIndex == runningTasks.Length - 1)
+                            {
+                                object[] oldChunk = runningTasks;
+
+                                runningTasks = (object[])runningTasks[runningTasks.Length - 1];
+                                toAwaitIndex = 0;
+
+                                // we're done with this, let some other caller reuse it immediately
+                                ReturnRentedArray(config, oldChunk, oldChunk.Length);
```
neildsh, line 825: >ReturnRentedArray [](http://example.com/codeflow?start=32&length=17)

Would there be any value in wrapping these calls in an `IDisposable` ? So when we rent an array we just get an `IDisposable` that knows how to clean itself up.

If you think that the current approach is more readable than the above, please keep the code as is.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Pagination/ParallelPrefetchTests.cs
```diff
@@ -0,0 +1,1312 @@
...
+            internal void AssertAllReturned()
+            {
+                Assert.IsTrue(this.rented.IsEmpty);
+            }
+        }
+
+        /// <summary>
+        /// ITrace which does nothing.
+        /// </summary>
+        private sealed class NullTrace : ITrace
```
neildsh, line 441: >NullTrace [](http://example.com/codeflow?start=29&length=9)

Does this implementation offer an advantage over `NoOpTrace.Singleton` ?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Pagination/ParallelPrefetchTests.cs
```diff
@@ -0,0 +1,1312 @@
...
+    using System;
+    using System.Buffers;
+    using System.Collections;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Antlr4.Runtime.Tree;
```
neildsh, line 12: >Antlr4.Runtime.Tree [](http://example.com/codeflow?start=10&length=19)

[Nit] Do we need this one ? Hard to figure out in a code flow window :)

neildsh (dismissed):

kevin-montrose:
> You have nice performance charts for this implementation in the description. If you wrote some code for benchmarking this, it may be good to add it to the Performance Tests project

The benchmarks are off [in a gist](https://gist.github.com/kevin-montrose/c2ffd6b3a5bdd5f3d80ee3fc3eaa3571) (also linked in the description), but they take a loooong time to run (I just ran 'em overnight during development) so I don't think it'd make much sense to check them into anything that is regularly run.  Since I didn't intend to run them regularly, I assembled the charting by hand - so there's nothing to save there.

neildsh (approved):

ealsur (dismissed):

neildsh (approved):

sboshra (approved):
:shipit:

PR #4365, opened 2024-03-22
Make `RegionNameMapper` static

joegoldman2:
## Description

Make class `RegionNameMapper` static to save some memory allocations.
Several trailing spaces and non-normalized carriage returns have been fixed, so I advise you to review this PR by hiding the whitespaces.

## Type of change


ealsur:
The class is only instantiated once during the lifetime of a client. Is there a substantial benefit to making it static? Are you seeing this type use up memory in your case and if so, are you also seeing multiple client instances (which really that would be the problem)?

joegoldman2:
I think I was looking at the wrong type. I checked again with dotnet-dump and indeed this type seems to be allocated only once.

PR #4389, opened 2024-04-04
[Internal] FaultInjection: Fixes Bug with Headers for TooManyRequests

NaluTripician:
# Pull Request Template

## Description

Changes FaultInjection responses so headers are set with `Set` rather than `Add`. `Add` works for the original request however, when the SDK retries the 429s it subsequently throws because the header is already present. 

## Type of change

## Closing issues


ealsur:
This shouldn't be included in the changelog, can we add an `[Internal]` to the PR title? Or some way for the person creating the changelog can know that its not related to the SDK?

NaluTripician:
> This shouldn't be included in the changelog, can we add an `[Internal]` to the PR title? Or some way for the person creating the changelog can know that its not related to the SDK?

I will put the internal tag on for now. We should maybe consider updating the release documentation to include that FaultInjection changes should not be included in the changelog as the FaultInjection package will presumably have its own changelog once it is released.

kirankumarkolli (approved):

ealsur (approved):

PR #4419, opened 2024-04-15
Query: Fixes bug in OrderByCrossPartitionQueryPipelineStage to ensure that errors in inner pipeline creation are bubbled up

neildsh:
## Description

Fixes bug in OrderByCrossPartitionQueryPipelineStage to ensure that errors in inner pipeline creation are bubbled up to the higher stages

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/OrderBy/OrderByCrossPartitionQueryPipelineStage.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/OrderBy/OrderByCrossPartitionQueryPipelineStage.cs
@@ -160,7 +160,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.CrossPartition.OrderBy
             }
 
             TryCatch<bool> hasNext = await this.inner.TryAsync(pipelineStage => pipelineStage.MoveNextAsync(trace, cancellationToken));
-            return hasNext.Succeeded && hasNext.Result;
+            return hasNext.Failed || hasNext.Result;
         }
 
         public ValueTask DisposeAsync()
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/OrderByQueryTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/OrderByQueryTests.cs
@@ -1183,6 +1183,58 @@
                 indexV2Policy);
         }
 
+        [TestMethod]
+        public async Task TestBadOrderByQueriesAsync()
+        {
+            const string NoCompositeIndex = "The order by query does not have a corresponding composite index that it can be served from.";
+            const string OrderByOverCorrelatedPath = "Order-by over correlated collections is not supported.";
+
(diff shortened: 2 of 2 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/OrderByQueryTests.cs
```diff
@@ -1181,6 +1181,58 @@ public async Task TestMixedTypeOrderByAsync()
...
+                foreach (bool enableOptmisticDirectExecution in new []{ false, true })
+                {
+                    QueryRequestOptions options = new QueryRequestOptions()
+                    {
+                        EnableOptimisticDirectExecution = enableOptmisticDirectExecution,
+                    };
+
+                    foreach ((string query, string error) in queries)
+                    {
+                        using FeedIterator feedIterator = container.GetItemQueryStreamIterator(query, continuationToken: null, options);
```
adityasa, line 1222: >GetItemQueryStreamIterator [](http://example.com/codeflow?start=68&length=26)

I am not 100% sure whether ODE kicks in for Stream API. We had a pending item for this. Although pipeline construction seems agnostic of top level api so may be that item is a no-ip.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/OrderBy/OrderByCrossPartitionQueryPipelineStage.cs
```diff
@@ -160,7 +160,7 @@ public async ValueTask<bool> MoveNextAsync(ITrace trace, CancellationToken cance
             }
 
             TryCatch<bool> hasNext = await this.inner.TryAsync(pipelineStage => pipelineStage.MoveNextAsync(trace, cancellationToken));
-            return hasNext.Succeeded && hasNext.Result;
```
adityasa, line 163: I stared at this line the other day for good 10 minutes and didn't find anything wrong. I find the TryCatch implementation coupled with pipeline's MoveNextAsync return value contract non-intuitive.

adityasa (approved):
:shipit:

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/OrderByQueryTests.cs
adityasa: Maybe this is the reason why an error is being produced in ODE case as well.

neildsh: I verified that the ODE pipeline is kicking in for these test cases

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/OrderBy/OrderByCrossPartitionQueryPipelineStage.cs
neildsh: Agreed. This is really easy to miss.

Maya-Painter (approved):

PR #4433, opened 2024-04-18
Distributed Tracing: Adds Request charge and Payload size Threshold options

### LINKED ISSUE #4385: Distributed Tracing: Add RequestChargeThreshold to CosmosThresholdOptions
Customers should be able to add a filter for traces based on the RU charge of a request. This option already exists in Java https://learn.microsoft.com/en-us/java/api/com.azure.cosmos.cosmosdiagnosticsthresholds?view=azure-java-stable

sourabh1007:
## Description

As part of this PR, providing Request Charge threshold and Payload load size threshold options to customers.It would generate request diagnostics if these thresholds were crossed,

## Type of change

## Closing issues

To automatically close an issue: closes #4385

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
@@ -10,17 +10,30 @@ namespace Microsoft.Azure.Cosmos
     public class CosmosClientTelemetryOptions
     {
         /// <summary>
-        /// Disable sending telemetry to service, <see cref="Microsoft.Azure.Cosmos.CosmosThresholdOptions"/> is not applicable to this as of now. 
+        /// Disable sending telemetry data to Microsoft, <see cref="Microsoft.Azure.Cosmos.CosmosThresholdOptions"/> is not applicable for this. 
         /// </summary>
-        /// <remarks>This option will disable sending telemetry to service.even it is opt-in from portal.</remarks>
+        /// <remarks>This feature has to be enabled at 2 places:
+        /// <list type="bullet">
+        /// <item>Opt-in from portal to subscribe for this feature.</item>
+        /// <item>Setting this property to false, to enable it for a particular client instance.</item>
+        /// </list>
+        /// </remarks>
         /// <value>true</value>
         public bool DisableSendingMetricsToService { get; set; } = true;
 
         /// <summary>
-        /// This method enable/disable generation of operation level <see cref="System.Diagnostics.Activity"/> if listener is subscribed to the Source Name "Azure.Cosmos.Operation".
+        /// This method enable/disable generation of operation level <see cref="System.Diagnostics.Activity"/> if listener is subscribed to the Source Name <i>"Azure.Cosmos.Operation"</i>(to capture operation level traces) 
+        /// and <i>"Azure-Cosmos-Operation-Request-Diagnostics"</i>(to capture events with request diagnostics JSON)
         /// </summary>
         /// <value>false</value>
-        /// <remarks> Please Refer https://opentelemetry.io/docs/instrumentation/net/exporters/ to know more about open telemetry exporters</remarks>
+        /// <remarks>
+        /// You can set different thresholds values by setting <see cref="Microsoft.Azure.Cosmos.CosmosThresholdOptions"/>. 
+        /// It would generate events with Request Diagnostics JSON, if any of the configured threshold is crossed, otherwise it would always generate events with Request Diagnostics JSON for failed requests.
+        /// There is some overhead of emitting the more detailed diagnostics - so recommendation is to choose these thresholds that reduce the noise level
+        /// and only emit detailed diagnostics when there is really business impact seen.<br></br>
+        /// Refer <a href="https://opentelemetry.io/docs/instrumentation/net/exporters/"></a> to know more about open telemetry exporters available. <br></br>
+        /// Refer <a href="https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/sdk-observability?tabs=dotnet"></a> to know more about this feature.
+        /// </remarks>
         public bool DisableDistributedTracing { get; set; } =
 #if PREVIEW
         false;
@@ -30,9 +43,8 @@ namespace Microsoft.Azure.Cosmos
 
         /// <summary>
         /// Threshold values for Distributed Tracing. 
-        /// These values decides whether to generate operation level <see cref="System.Diagnostics.Tracing.EventSource"/> with request diagnostics or not.
+        /// These values decides whether to generate an <see cref="System.Diagnostics.Tracing.EventSource"/> with request diagnostics or not.
         /// </summary>
         public CosmosThresholdOptions CosmosThresholdOptions { get; set; } = new CosmosThresholdOptions();
-
     }
 }
--- a/Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
@@ -38,8 +38,7 @@ namespace Microsoft.Azure.Cosmos
         /// Can be used to define a custom RU (request charge) threshold. When the threshold is exceeded more detailed
         /// diagnostics will be emitted (including the request diagnostics). There is some overhead of emitting the
         /// more detailed diagnostics - so recommendation is to choose a request charge threshold that reduces the noise
-        /// level and only emits detailed diagnostics when the request charge is significantly  higher thane expected.
-        /// The default value for the request charge threshold is 1000
+        /// level and only emits detailed diagnostics when the request charge is significantly higher than expected.
         /// </summary>
         public double? RequestChargeThreshold { get; set; } = null;
 
@@ -48,7 +47,8 @@ namespace Microsoft.Azure.Cosmos
         /// response payloads more detailed diagnostics will be emitted (including the request diagnostics).
         /// There is some overhead of emitting the more detailed diagnostics - so recommendation is to choose a
         /// payload size threshold that reduces the noise level and only emits detailed diagnostics when the payload size
-        /// is significantly higher than expected.        /// </summary>
+        /// is significantly higher than expected.
+        /// </summary>
         public int? PayloadSizeThresholdInBytes { get; set; } = null;
     }
 }
--- a/Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Azure.Cosmos
         /// is significantly higher than expected.
         /// The default value for the payload size threshold is Int32.MaxValue 
         /// </summary>
-        /// <value> 2147483647 Bytes </value>
+        /// <value>2147483647 Bytes</value>
         public int PayloadSizeThresholdInBytes { get; set; } = Int32.MaxValue;
     }
 }
--- a/Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Azure.Cosmos
     using System;
 
     /// <summary>
-    /// This class describes the thresholds when more details diagnostics are emitted for an operation due to high latency,
+    /// This class describes the thresholds when more details diagnostics events are emitted, if subscribed, for an operation due to high latency,
     /// high RU consumption or high payload sizes.
     /// </summary>
     public class CosmosThresholdOptions
--- a/Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
@@ -41,18 +41,14 @@ namespace Microsoft.Azure.Cosmos
         /// level and only emits detailed diagnostics when the request charge is significantly  higher thane expected.
         /// The default value for the request charge threshold is 1000
         /// </summary>
-        /// <value>1000 RU</value>
-        public double RequestChargeThreshold { get; set; } = 1000;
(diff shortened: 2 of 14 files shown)


FabianMeiswinkel (approved):
LGTM

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -7,20 +7,52 @@ namespace Microsoft.Azure.Cosmos
     using System;
 
     /// <summary>
-    /// Threshold values for Distributed Tracing
+    /// This class describes the thresholds when more details diagnostics are emitted for an operation due to high latency,
```
kirankumarkolli, line 10: Text needs to be explicit that its only about otel diagnostics, right? 
Otherwise it might get mis-interpreted right?

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -7,20 +7,52 @@ namespace Microsoft.Azure.Cosmos
...
+    /// high RU consumption or high payload sizes.
     /// </summary>
     public class CosmosThresholdOptions
     {
         /// <summary>
-        /// Latency Threshold for non point operations i.e. Query
+        /// Can be used to define custom latency thresholds. When the latency threshold is exceeded more detailed
+        /// diagnostics will be emitted (including the request diagnostics). There is some overhead of emitting the
+        /// more detailed diagnostics - so recommendation is to choose latency thresholds that reduce the noise level
```
kirankumarkolli, line 18: > There is some overhead of emitting the
        /// more detailed diagnostics - so recommendation is to choose latency thresholds that reduce the noise level
        /// and only emit detailed diagnostics when there is really business impact seen

This part is repetitive and can be moved up to the class definition?

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -7,20 +7,52 @@ namespace Microsoft.Azure.Cosmos
     using System;
 
     /// <summary>
-    /// Threshold values for Distributed Tracing
+    /// This class describes the thresholds when more details diagnostics are emitted for an operation due to high latency,
+    /// high RU consumption or high payload sizes.
     /// </summary>
```
kirankumarkolli, line 12: Please call out that all the conditions/thresholds below need to be met.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -7,20 +7,52 @@ namespace Microsoft.Azure.Cosmos
...
+
+        /// <summary>
+        /// Can be used to define a custom RU (request charge) threshold. When the threshold is exceeded more detailed
+        /// diagnostics will be emitted (including the request diagnostics). There is some overhead of emitting the
+        /// more detailed diagnostics - so recommendation is to choose a request charge threshold that reduces the noise
+        /// level and only emits detailed diagnostics when the request charge is significantly  higher thane expected.
+        /// The default value for the request charge threshold is 1000
+        /// </summary>
+        /// <value>1000 RU</value>
+        public double RequestChargeThreshold { get; set; } = 1000;
```
kirankumarkolli, line 45: Default '1k', what's the rational for it? 

What's the default in JAVA? 
One alternative is to have payload/RU by-default OFF, thoughts?

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -7,20 +7,52 @@ namespace Microsoft.Azure.Cosmos
     using System;
 
     /// <summary>
-    /// Threshold values for Distributed Tracing
+    /// This class describes the thresholds when more details diagnostics are emitted for an operation due to high latency,
+    /// high RU consumption or high payload sizes.
     /// </summary>
     public class CosmosThresholdOptions
```
kirankumarkolli, line 13: Its only applicable for successful operations, right? 
How about calling it out explicitly?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/CosmosDbEventSource.cs
```diff
@@ -35,17 +36,30 @@ public static bool IsEnabled(EventLevel level)
...
-                    operationType: operationType,
-                    response: response) && CosmosDbEventSource.IsEnabled(EventLevel.Warning))
-            {
-                CosmosDbEventSource.Singleton.LatencyOverThreshold(response.Diagnostics.ToString());
+                if (!DiagnosticsFilterHelper.IsSuccessfulResponse(
+                                        response.StatusCode, response.SubStatusCode))
+                {
+                    CosmosDbEventSource.Singleton.FailedRequest(response.Diagnostics.ToString());
+                }
+                else if (DiagnosticsFilterHelper.IsLatencyThresholdCrossed(
```
kirankumarkolli, line 46: Are thresholds be AND / OR ?

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
sourabh1007: It has nothing to do with "Otel" specifically, SDK will generate Event, now listener can be otel listener or any custom listener.
But yes, I added specifically, it is about generating events, if subscribed.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
sourabh1007: IMO, we should keep it here, generally customer/developer check the documentation of the configuration first, if that is not convincing then it goes for class level documentation. This is important information, we should keep it here. (this is how, it is there java sdk also).

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
sourabh1007: No, any one condition has to be true. I don't think, we need to explain it.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
sourabh1007: It is taken from Java SDK only. Setting it to 1000 is keeping it OFF only (in most of the cases).

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/CosmosDbEventSource.cs
sourabh1007: OR

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
sourabh1007: Failed operation, will have an event emitted by default.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
kirankumarkolli: OR seems right, but its confusing.

Microsoft.Azure.Cosmos/src/CosmosThresholdOptions.cs
```diff
@@ -7,20 +7,52 @@ namespace Microsoft.Azure.Cosmos
     using System;
 
     /// <summary>
-    /// Threshold values for Distributed Tracing
```
kirankumarkolli, line 10: Caller has little context on when this is applicable.

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/400abfd5184822f1f7f86f59cb285ef1ee3ee19c/Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs#L13

sourabh1007: Added more information

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
```diff
@@ -10,17 +10,30 @@ namespace Microsoft.Azure.Cosmos
...
         /// </summary>
-        /// <remarks>This option will disable sending telemetry to service.even it is opt-in from portal.</remarks>
+        /// <remarks>This feature has to be enabled at 2 places:
+        /// <list type="bullet">
+        /// <item>Opt-in from portal to subscribe for this feature.</item>
+        /// <item>Setting this property to false, to enable it for a particular client instance.</item>
+        /// </list>
+        /// </remarks>
         /// <value>true</value>
         public bool DisableSendingMetricsToService { get; set; } = true;
```
FabianMeiswinkel, line 22: I don't get this - why is the default true? Wasn't the whole idea ofallowing setting in the portal to make whatever the portal has be the default? And maybe also allow cx to opt-out in SDK - but by default use whatever portal is saying?

FabianMeiswinkel (dismissed):
LGTM - only comment left is not regarding a change in this PR - but please let's discuss that one next week. Seem very odd to me.

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
sourabh1007: We will do it once we are on public preview, thats why flag name is "Disablexxxx" so that by default it would be `false` (same as portal) and cx can set it to true when they want to disable it.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/CosmosDbEventSource.cs
```diff
@@ -75,5 +91,17 @@ private void FailedRequest(string message)
         {
             this.WriteEvent(3, message);
         }
+
+        [Event(4, Level = EventLevel.Warning)]
```
kirankumarkolli, line 95: Separate events becomes expectation or contract. Is it intentional? 
Its also unclear on when which event is emitted when multiple criteria met (ex: RU and latency both met only one event is emitted)

One alternative is to emit all of them under same eventId.

sourabh1007: AFAIR, we discussed this design to have separate event names for different thresholds. cc @jcocchi 

If it was discussed and changed, we can take it as separate PR. Focusing on that change only.

Microsoft.Azure.Cosmos/src/CosmosClientTelemetryOptions.cs
jcocchi: In principle Portal is source of truth and the customer should be able to opt-out per application

Portal `true`, SDK `true` == true
Portal `true`, SDK `false` == false
Portal `false`, SDK `true` == false
Portal `false`, SDK `false` == false

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/CosmosDbEventSource.cs
jcocchi: I know we discussed naming the event based on threshold, but that was when there was only one threshold option. If two thresholds are passed (ex. latency and RU) I think we should only emit one event. Maybe the thresholds that were passed become an attribute on the event instead of the event name?

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/CosmosDbEventSource.cs
kirankumarkolli: That's a great point, @sourabh1007 is it possible to pass extra context/attribute?

sourabh1007: Looks like, we can pass more information in an event, https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource.writeevent?view=net-8.0#system-diagnostics-tracing-eventsource-writeevent(system-int32-system-string-system-string)
Need to check, how it would appear on AppInsights UI.

sourabh1007: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4461 to track this.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/CosmosDbEventSource.cs
```diff
@@ -81,7 +75,7 @@ private void Exception(string message)
         }
 
         [Event(2, Level = EventLevel.Warning)]
-        private void LatencyOverThreshold(string message)
+        private void ThresholdViolation(string message)
```
FabianMeiswinkel, line 77: why not add an attribute for the trigger(s) of this event?

FabianMeiswinkel: At least that is the model I would prefer in Java - use the same event id for any threshold violation (only failure would be different) - but add an attribute with info which trigger(s) caused the threshold violation

FabianMeiswinkel: If we agree that we handle this differently on Java or .Net I wouldn't block on this - but I don't want to have to change this and remove the trigger in Java again

sourabh1007: I will put that information as a separate PR, I need to check if appinsight is populating that information or not or how is the experience.
Created this bug, to track this change https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4461

FabianMeiswinkel (approved):
LGTM

philipthomas-MSFT (approved):

PR #4446, opened 2024-04-22
Query: Adds a new capability for non streaming order by in QueryFeatures

neildsh:
Adds:
- [x] a new capability for non streaming order by in QueryFeatures
- [x] a new flag to QueryInfo that will be used by ServiceInterop to indicate non streaming order by queries

These will be used by the Compute gateway to fail queries that require non streaming order by but do not pass this capability

## Type of change


adityasa (approved):
:shipit:

PR #4447, opened 2024-04-23
FaultInjection: Adds SendDelay ServerErrorType

NaluTripician:
# Pull Request Template

## Description

This PR adds the `SendDelay` ServerErrorType. This error will add a delay before the request is sent to the backend. This is similar to the existing `ResponseDelay` ServerErrorType, who's behavior is also fixed in this PR. 

`SendDelay` will add a delay to the request before sending. 
`ResponseDelay` functions the same however the delay is after sending the request but before sending the response. 

When the delay is set to above the timeout time and users are trying to mimic timeout scenarios, these two rules will allow users to decide whether they want the request to reach the backend or not. 

## Type of change

## Closing issues


kirankumarkolli (dismissed):

Microsoft.Azure.Cosmos.Samples/Tools/FaultInjection/src/FaultInjectionServerErrorResultBuilder.cs
```diff
@@ -42,15 +42,17 @@ public FaultInjectionServerErrorResultBuilder WithTimes(int times)
         /// 
         /// Only used RESPONSE_DELAY and CONNECTION_DELAY.
         /// 
-        /// For RESPONSE_DELAY, it is the delay added before the response.
+        /// For SEND_Delay, it is the delay added before the request is sent.
```
ealsur, line 45: If these are public enums then prefer

```suggestion
        /// For <see cref="FaultInjectionServerErrorType.SendDelay"/>, it is the delay added before the request is sent.
```

The same with the other comments. `RESPONSE_DELAY` for example does not really represent any of the enum values, it does not match the casing nor name (if we wanted to at least match the name it would be ResponseDelay)

ealsur (approved):

kirankumarkolli (approved):

PR #4459, opened 2024-04-29
Query: Fixes non streaming order by to use flag from query plan

neildsh:
## Description

The contract with the backend for non streaming order by has changed. The backend response will no longer have _streaming. Instead there is a new flag on the query plan to indicate the presence of a non streaming order by. This PR switches non streaming order by to use flag from query plan instead of _streaming in the backend response.

This PR does not do any code cleanup, since the SDK is required for integration testing of the backend. The cleanup PR will follow this change.

## Type of change


adityasa (approved):
:shipit:

sboshra (approved):
:shipit:

PR #4470, opened 2024-05-07
NonBlockingAsyncCache: Fixes lambda func capturing the outer context (memory optimization)

### LINKED ISSUE #4345: Memory usage for AddressCache - DocumentServiceRequest capture
During operation flow, whenever Replica Addresses are required to be refreshed or obtained for the first time, the `DocumentServiceRequest` of the current operation is captured in the lambda context:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/d07f9cabad7d67ee1e3d3cb5e709a5d7207c0318/Microsoft.Azure.Cosmos/src/Routing/GatewayAddressCache.cs#L282-L290

Once the `request` is captured, it remains "tied" to that cache entry. The `request` contains also the `ClientSideRequestStatisticsDatum` which bloats the required memory footprint of that cache entry.

This effect gets multiplied by the number of clients and the number of containers these clients interact with.

## Feature request

Decouple the `DocumentServiceRequest` from the cache entry or potentially reduce the memory footprint by removing the capturing of `ClientSideRequestStatisticsDatum`.

guiyong:
# Pull Request Template

## Description

Set createValueFunc to null once the Func delegate is not needed. Without breaking the reference, the lambda function will keep capturing of any outer variables, in #4345, DocumentServiceRequest was captured, which prevent GC from recycling the request object, cause memory leak. 

Below code, the request object will be captured by singleValueInitFunc delegate.

```
addresses = await this.serverPartitionAddressCache.GetAsync( 
                         key: partitionKeyRangeIdentity, 
                         singleValueInitFunc: (_) => this.GetAddressesForRangeIdAsync( 
                             request, 
                             cachedAddresses: null, 
                             partitionKeyRangeIdentity.CollectionRid, 
                             partitionKeyRangeIdentity.PartitionKeyRangeId, 
                             forceRefresh: false), 
                         forceRefresh: (_) => false); 
```

## Type of change

## Closing issues

To automatically close an issue: closes #4345


Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -253,11 +253,11 @@ public bool TryRemove(TKey key)
         private sealed class AsyncLazyWithRefreshTask<T>
         {
             private readonly CancellationToken cancellationToken;
-            private readonly Func<T, Task<T>> createValueFunc;
             private readonly object valueLock = new ();
             private readonly object removedFromCacheLock = new ();
 
-            private bool removedFromCache = false;
+            private bool removedFromCache = false;
+            private Func<T, Task<T>> createValueFunc;
```
kirankumarkolli, line 260: As member the lifetime is interpreted differently. 
How about remove it as member and always force the caller to provide on GetValueAsync(Func<T, Task<T>>)?

kirankumarkolli:
Closing due to inactivity, feel free to re-open as needed.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
guiyong: There might be a problem, if the async lazy was inserted by another thread, we could not provide Func<T, Task> belong to current thread to GetValueAsync(Func<T, Task>), thoughts?
Here we know we just need to evaluate the Func once no matter which thread did that, then we don't need it anymore.

guiyong:
@kirankumarkolli,  could you please let me know where I can re-open it. Thanks!

FabianMeiswinkel (dismissed):
LGTM

Pilchie (approved):

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
kirankumarkolli: Calback is not thread specific and applicable to all the callers on that scope.

guiyong: In that way, AsyncLazyWithRefreshTask will lose the 'Lazy' part, which is fine, second thread might get its callback run first with associated request, which is also fine in the context. I could convert this to always force caller to provide callback.

kirankumarkolli (approved):

guiyong:

PR #4471, opened 2024-05-08
[Inernal] Ignore: Some thoughts on hedging

kirankumarkolli:
# Pull Request Template

## Description

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
```diff
@@ -0,0 +1,233 @@
...
+        {
+            if (!this.ShouldHedge(request))
+            {
+                return await sender(request, cancellationToken);
+            }
+
+            using (CancellationTokenSource cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
+            {
+                // Get effective order of regions to route to (static once populated)
+                IReadOnlyCollection<Uri> availableRegions = client.DocumentClient.GlobalEndpointManager.GetApplicableEndpoints(request.RequestOptions.ExcludeRegions, isReadRequest: true);
```
NaluTripician, line 115: I think this is better than getting the current available regions for the scenario where an offline region becomes available again. Cavoite is that if a region is not available, a request will still be sent to it but since it will hedge on other regions too this is not much of a problem.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
```diff
@@ -0,0 +1,233 @@
...
+                List<Task> requestTasks = new List<Task>(availableRegions.Count + 1);
```
NaluTripician, line 116: the +1 is a good catch here

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
```diff
@@ -0,0 +1,233 @@
...
+
+                ResponseMessage responseMessage = null;
+                
+                //available regions or hedge regions?
+                //Send out hedged requests
+                for (int requestNumber = 0; requestNumber < availableRegions.Count; requestNumber++)
+                {
+                    TimeSpan awaitTime = this.Threshold + TimeSpan.FromMilliseconds(requestNumber * this.ThresholdStep.Milliseconds);
```
NaluTripician, line 124: For this I do not think this will give the right timespans except for the first await. 
Request 0: await threshold before sending next request -- correct 
Request 1: await threshold + step before sending next request -- is waiting too much time, since here the threshold amount of time should have passed then it should only be waiting the threshold step amount of time. 

Does this make sense? So the time should really be: `Timespan awaitTime = requestNumber == 0 ? this.Threshold : this.ThresholdStep;`

This is because the `WhenAny` call has the await which will complete when the Task.Delay is done (or when a request completes).

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
```diff
@@ -0,0 +1,233 @@
...
+                    Task thresholdDelayTask = Task.Delay(awaitTime, cancellationToken);
+
+                    using (RequestMessage clonedRequest = (requestNumber == 0) ? request : request.Clone(request.Trace.Parent))
+                    {
+                        clonedRequest.RequestOptions ??= new RequestOptions();
+
+                        clonedRequest.RequestOptions.ExcludeRegions = null;
```
NaluTripician, line 131: Do we want to ignore the original exclude regions list here if it is provided? Also, by using location endpoint to route rather than exclude regions we would be allowing cross regional retries on the hedged requests. Is this something we want to do; this behavior is different than Java.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
```diff
@@ -0,0 +1,233 @@
...
+                }
+
+                //Wait for a good response from the hedged requests/primary request
+                while (requestTasks.Any())
+                {
+                    Task completedTask = await Task.WhenAny(requestTasks);
+                    requestTasks.Remove(completedTask);
+
+                    (bool isNonTransient, responseMessage) = await (Task<(bool, ResponseMessage)>)completedTask;
+                    if (isNonTransient)
```
NaluTripician, line 169: Here I think we can do `if(isNonTransient || requestTasks.Count == 1)`

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
```diff
@@ -0,0 +1,233 @@
...
+                ResponseMessage responseMessage = null;
+                
+                //available regions or hedge regions?
+                //Send out hedged requests
+                for (int requestNumber = 0; requestNumber < availableRegions.Count; requestNumber++)
+                {
+                    TimeSpan awaitTime = this.Threshold + TimeSpan.FromMilliseconds(requestNumber * this.ThresholdStep.Milliseconds);
+                    Task thresholdDelayTask = Task.Delay(awaitTime, cancellationToken);
+
+                    using (RequestMessage clonedRequest = (requestNumber == 0) ? request : request.Clone(request.Trace.Parent))
```
NaluTripician, line 127: I think creation of the cloned message has to be in a different helper method or else when moving outside the for loop the message will be disposed

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
kirankumarkolli: There are two side affects
- In-case a future region is made available that might get excluded
- Possible higher latency in-case of future un-avilale was included

Both issues were present even with earlier model (may be first full request might cover it?)

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
kirankumarkolli: I just used existing logic from your PR, that's a clarification I have too, feel free to update as needed

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
kirankumarkolli: Exclude regions are alreay considered in initial list population already. 

here we are using single region targeting `LocationEndpointToRoute`

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
kirankumarkolli: Great point. This needs attention

PR #4490, opened 2024-05-14
Query: Adds Distribution for MakeList and MakeSet

ezrahaleva-msft:
## Description

Adds MakeList and MakeSet aggregate operators and implements MakeListAggregator and MakeSetAggregator.

## Type of change


Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            CosmosElement[] cosmosElementArray = new CosmosElement[this.globalList.Count];
+            this.globalList.CopyTo(cosmosElementArray);
+            return CosmosArray.Create(cosmosElementArray);
+        }
+
+        public string GetContinuationToken()
+        {
+            return this.globalList.ToString();
+        }
```
khdang, line 52: as discussed, this would not work for large lists. we should consider if it makes sense to support small lists or not

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public CosmosElement GetResult()
+        {
+            CosmosElement[] cosmosElementArray = new CosmosElement[this.globalList.Count];
+            this.globalList.CopyTo(cosmosElementArray);
+            return CosmosArray.Create(cosmosElementArray);
+        }
+
+        public string GetContinuationToken()
+        {
```
khdang, line 50: as discussed, let's include tests for the following

- single partition, without distribution needed
- with continuation token
- large lists that exceed the continuation token size limit

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -126,6 +128,34 @@ public async Task TestAggregateFunctionsAsync()
...
+                                else if (argument.AggregateOperator.Equals("MAKELIST") || argument.AggregateOperator.Equals("MAKESET"))
+                                {
+                                    if ((expected is CosmosArray expectedArray) && (actual is CosmosArray actualArray))
+                                    {
+                                        CosmosElement[] normalizedExpected = expectedArray.ToArray();
+                                        Array.Sort(normalizedExpected);
+                                        CosmosElement[] normalizedActual = actualArray.ToArray();
+                                        Array.Sort(normalizedActual);
+
+                                        Assert.IsTrue(normalizedExpected.Length == normalizedActual.Length, message);
```
khdang, line 140: >Assert.IsTrue [](http://example.com/codeflow?start=40&length=13)

use Assert.AreEqual

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -126,6 +128,34 @@ public async Task TestAggregateFunctionsAsync()
...
+                                        for( int i = 0; i < normalizedExpected.Length; i++ )
+                                        {
+                                            if (normalizedActual[i].CompareTo(normalizedExpected[i]) != 0)
+                                            {
+                                                expectedEqualsActual = false;
+                                                break;
+                                            }
+                                        }
+
+                                        Assert.IsTrue(expectedEqualsActual, message);
```
khdang, line 152: can't you just use CollectionAssert.AreEqual here?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -126,6 +128,34 @@ public async Task TestAggregateFunctionsAsync()
                                 {
                                     Assert.AreEqual(Number64.ToDouble(expectedNumber.Value), Number64.ToDouble(actualNumber.Value), .01);
                                 }
+                                else if (argument.AggregateOperator.Equals("MAKELIST") || argument.AggregateOperator.Equals("MAKESET"))
+                                {
+                                    if ((expected is CosmosArray expectedArray) && (actual is CosmosArray actualArray))
```
khdang, line 133: can we combine these two conditions to avoid the extra else?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -126,6 +128,34 @@ public async Task TestAggregateFunctionsAsync()
                                 {
                                     Assert.AreEqual(Number64.ToDouble(expectedNumber.Value), Number64.ToDouble(actualNumber.Value), .01);
                                 }
+                                else if (argument.AggregateOperator.Equals("MAKELIST") || argument.AggregateOperator.Equals("MAKESET"))
+                                {
```
khdang, line 132: please add a comment on the reason we need to ignore the order of the items

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -179,6 +209,38 @@ public async Task TestAggregateFunctionsAsync()
...
+                                        }
+                                    }
+
+                                    Assert.IsTrue(expectedEqualsActual, message);
+                                }
+                                else
+                                {
+                                    Assert.AreEqual(expected, actual, message);
+                                }
+                            }
```
khdang, line 243: same comments with the similar code in NonOdeImplementationAsync

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -237,6 +314,14 @@ public async Task TestAggregateFunctionsAsync()
                         aggregateOperator: "COUNT",
                         expectedValue: CosmosNumber64.Create(inputDocuments.Count()),
                         predicate: "true"),
+                    new AggregateQueryArguments(
+                        aggregateOperator: "MAKELIST",
+                        expectedValue: CosmosArray.Create(makeListResult), // construct cosmos array from linq result
```
khdang, line 319: >, // construct cosmos array from linq result [](http://example.com/codeflow?start=73&length=44)

no need for this comment, and the same comment below because it doesn't add extra info to the code itself

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -15,6 +16,7 @@
     using Microsoft.Azure.Documents;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using Newtonsoft.Json.Linq;
+    using static System.Net.Mime.MediaTypeNames;
```
khdang, line 20: we don't need this dependency right?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -609,10 +720,19 @@ FROM c
                     {
                         Assert.AreEqual(1, items.Count);
                         CosmosElement aggregateResult = items.First();
-
                         if(aggregateResult is not CosmosUndefined)
                         {
-                            writer.WriteCData(items.Single().ToString());
+                            if ((formattedQuery.Contains("MAKELIST") || formattedQuery.Contains("MAKESET"))
+                                && (aggregateResult is CosmosArray aggregateResultArray))
```
khdang, line 726: We shouldn't do this check because it's fragile. We can determine if we need to sort the result in advance right? we can just have something like a separate array to indicate if the result needs to be sorted or not.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Removed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Comment Added.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Comment also added in NonOdeImplementationAsynch

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: removed comment.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -286,16 +449,18 @@ public async Task TestAggregateFunctionsAsync()
...
                 this.AggregateOperator = aggregateOperator;
                 this.ExpectedValue = expectedValue;
                 this.Predicate = predicate;
+                this.SortResult = sortResult;
             }
 
             public string AggregateOperator { get; }
             public CosmosElement ExpectedValue { get; }
             public string Predicate { get; }
+            public bool SortResult { get; }
```
khdang, line 463: >SortResult [](http://example.com/codeflow?start=24&length=10)

How about calling it IgnoreOrder to be clear on the intention here?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -527,7 +718,7 @@ private struct AggregateQueryMixedTypes
                 args.UndefinedKey
             };
 
-            string[] aggregateOperators = new string[] { "AVG", "MIN", "MAX", "SUM", "COUNT" };
+            string[] aggregateOperators = new string[] { "AVG", "MIN", "MAKELIST", "MAKESET", "MAX", "SUM", "COUNT" };
```
khdang, line 721: >"MAKELIST", "MAKESET", [](http://example.com/codeflow?start=71&length=22)

this is list is not sorted, so either we sort it, or we should put them at the end since they're different kind

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,79 @@
...
+                partialList = cosmosPartialList;
+            }
+            else
+            {
+                partialList = CosmosArray.Empty;
+            }
+
+            return TryCatch<IAggregator>.FromResult(
+                new MakeListAggregator(initialList: partialList));
+        }
```
khdang, line 70: we should just have this in one line

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (!(localList is CosmosArray cosmosArray))
+            {
+                throw new ArgumentException($"{nameof(localList)} must be an array.");
+            }
+
+            foreach (CosmosElement listItem in cosmosArray)
+            {
+                this.globalList.Add(listItem);
+            }
+        }
```
khdang, line 41: I think there is a way to add all items in one call

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeSetAggregator.cs
```diff
@@ -0,0 +1,84 @@
...
+                partialSet = cosmosPartialSet;
+            }
+            else
+            {
+                partialSet = CosmosArray.Empty;
+            }
+
+            return TryCatch<IAggregator>.FromResult(
+                new MakeSetAggregator(initialSet: partialSet));
+        }
```
khdang, line 75: one line

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Changed to use CollectionAssert

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Changed to use CollectionAssert

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Combined to avoid third else

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Removed contains check, using tuple instead.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
ezrahaleva-msft: Replaced with single call to AddRange, also for MakeSet with UnionWith()

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
ezrahaleva-msft: Done.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeSetAggregator.cs
ezrahaleva-msft: Done.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Changed name to IgnoreResultOrder

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Sorted list, updated baseline.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,79 @@
...
+            return TryCatch<IAggregator>.FromResult(
+                new MakeListAggregator(initialList: partialList));
+        }
+
+        public CosmosElement GetCosmosElementContinuationToken()
+        {
+            CosmosElement[] cosmosElementArray = new CosmosElement[this.globalList.Count];
+            this.globalList.CopyTo(cosmosElementArray);
+            return CosmosArray.Create(cosmosElementArray);
+        }
```
khdang, line 74: as discussed, we need to confirm this is conceptually correct

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
...
+                (
+                    "SELECT c.name AS NameGroup, MAKELIST(c.age) AS list_age FROM c GROUP BY c.name",
+                    documents
+                        .GroupBy(document => document["name"])
+                        .Select(grouping => CosmosObject.Create(
+                            new Dictionary<string, CosmosElement>()
+                            {
+                                { "NameGroup", grouping.Key },
+                                { "list_age", CosmosArray.Create(
+                                    grouping.Select(document => document["age"])
```
khdang, line 552: we should this in one line

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
...
+                    "SELECT c.name AS NameGroup, MAKESET(c.age) AS list_age FROM c GROUP BY c.name",
+                    documents
+                        .GroupBy(document => document["name"])
+                        .Select(grouping => CosmosObject.Create(
+                            new Dictionary<string, CosmosElement>()
+                            {
+                                { "NameGroup", grouping.Key },
+                                { "list_age", CosmosArray.Create(
+                                    grouping.Select(document => document["age"]).Distinct()
+                                )}
```
khdang, line 567: update to one line

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
                 }
             }
 
-            // Test that continuation token is blocked
+            List<(string, List<CosmosElement>)> queryAndExpectedResultsListArrayAggregates = new List<(string, List<CosmosElement>)>()
+            {
+                (
+                    "SELECT c.name AS NameGroup, MAKELIST(c.age) AS list_age FROM c GROUP BY c.name",
```
khdang, line 544: we should just call it 'Name', no need the word 'Group'. Also, the aliases naming convention should be consistent. We shouldn't have 'NameGroup' and 'list_age' which are in different styles

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
...
+                            {
+                                { "NameGroup", grouping.Key },
+                                { "list_age", CosmosArray.Create(
+                                    grouping.Select(document => document["age"])
+                                )}
+                            }))
+                        .ToList<CosmosElement>()
+                ),
+                (
+                    "SELECT c.name AS NameGroup, MAKESET(c.age) AS list_age FROM c GROUP BY c.name",
```
khdang, line 558: this should be 'set_age' right?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
...
-                    List<JToken> actual = await QueryTestsBase.QueryWithContinuationTokensAsync<JToken>(
+                    List<CosmosElement> actualWithoutContinuationTokens = await QueryTestsBase.QueryWithoutContinuationTokensAsync<CosmosElement>(
+                        container,
+                        query,
+                        new QueryRequestOptions()
+                        {
+                            MaxConcurrency = 2,
+                            MaxItemCount = maxItemCount,
+                            MaxBufferedItemCount = 100,
+                        });
```
khdang, line 585: for readability, can we create one QueryRequestOptions and use them in both cases?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
...
+                            myDict.Add(kvp.Key, CosmosArray.Create(normalizedArray));
+                        }
+                        else
+                        {
+                            myDict.Add(kvp.Key, kvp.Value);
+                        }
+                    }
+
+                    IReadOnlyDictionary<string, CosmosElement> newDict = new ReadOnlyDictionary<string, CosmosElement>(myDict);
+                    results[i] = CosmosObject.Create(newDict);
```
khdang, line 663: this is not neccessary right? doesn't Dictionary implement IReadOnlyDictionary?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
...
+            }
+        }
+
+        private void NormalizeGroupByArrayAggregateResults(
+            List<CosmosElement> results)
+        {
+            for (int i = 0; i < results.Count; i++)
+            {
+                if (results[i] is CosmosObject cosmosObject)
                 {
```
khdang, line 645: is there any case you expect not to be a CosmosObject?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
...
+                    IDictionary<string, CosmosElement> myDict = new Dictionary<string, CosmosElement>();
```
khdang, line 646: for readability, please avoid using the word 'my' in variable names as it doesn't convey any info. Consider using 'old', 'new', 'existing', 'current', etc. instead

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -537,22 +538,129 @@ FROM c
...
                             MaxConcurrency = 2,
-                            MaxItemCount = 1
+                            MaxItemCount = maxItemCount,
+                            MaxBufferedItemCount = 100,
                         });
-                    Assert.Fail("Expected an error when trying to drain a GROUP BY query with continuation tokens.");
+
+                    this.NormalizeGroupByArrayAggregateResults(actualWithTryGetContinuationTokens);
+                    HashSet<CosmosElement> actualWithTryGetContinuationTokensSet = new HashSet<CosmosElement>(actualWithTryGetContinuationTokens);
+
```
khdang, line 602: why is a HashSet needed? don't we just normalize the result the line right above this?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
```diff
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Collections.ObjectModel;
```
khdang, line 5: what is this needed for?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -14,12 +14,105 @@
...
     using Newtonsoft.Json.Linq;
 
     [TestClass]
     [TestCategory("Query")]
     public sealed class AggregateCrossPartitionQueryTests : QueryTestsBase
     {
+
+        [TestMethod]
+        public async Task TestArrayAggregatesContinuationToken()
+        {
```
khdang, line 27: can you move this after the main test as it is a more speficic aspect of the aggregate function

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -14,12 +14,105 @@
...
+                documents,
+                ImplementationAsync,
+                "/id");
+
+            async static Task ImplementationAsync(Container container, IReadOnlyList<CosmosObject> documents)
+            {
+                foreach (string[] queriesToCompare in new string[][]
+                {
+                //new string[]{ "SELECT VALUE c.age FROM c ORDER BY c.age", "SELECT VALUE MakeList(c.age) FROM c GROUP BY c.name" },
+                new string[]{ "SELECT VALUE c.age FROM c ORDER BY c.age", "SELECT VALUE MakeList(c.age) FROM c" },
```
khdang, line 65: can we please have Group By queries here as well?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -623,6 +827,17 @@ FROM c
...
+            //string filenameOutput = $"AggregateQueryTests.AggregateMixedTypes_OUTPUT{randomStr}.xml";
+
+            //// Set a variable to the Documents path.
+            //string docPath = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
+
+            //// Write the text to a new file named "WriteFile.txt".
+            //File.WriteAllText(Path.Combine(docPath, filenameOutput), builder.ToString());
+
+
+
```
khdang, line 840: please remove commented debugging code

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -392,6 +557,32 @@ public async Task TestAggregateFunctionsWithEmptyPartitionsAsync()
...
+                        List<CosmosElement> items = await QueryTestsBase.RunQueryAsync(
+                            container,
+                            query,
+                            new QueryRequestOptions()
+                            {
+                                MaxConcurrency = 10,
+                            });
+
+                        Assert.IsTrue((items.Single() is CosmosArray result) && result.Equals(CosmosArray.Create(CosmosNumber64.Create(valueOfInterest))));
+                    }
```
khdang, line 580: we should also assert the items has exactly 1 element

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/AggregateOperator.cs
```diff
@@ -9,6 +9,8 @@ internal enum AggregateOperator
         Average,
         Count,
         Max,
+        MakeList,
+        MakeSet,
         Min,
         Sum,
```
khdang, line 15: please sort

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/SingleGroupAggregator.cs
```diff
@@ -381,7 +389,6 @@ public override CosmosElement GetCosmosElementContinuationToken()
                         case AggregateOperator.Sum:
                             tryCreateAggregator = SumAggregator.TryCreate(continuationToken);
                             break;
-
```
khdang, line 384: please add the new line back

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Done.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Done.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Renamed to 'Name', 'AgeList', and 'AgeSet'

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Yes, renamed to "AgeSet'

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Done. For original cases and ArrayAggregate cases.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Regular 'Dictionary' does not, but I've changed it to 'ReadOnlyDictionary' instead of 'IReadOnlyDictionary'.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: No, but we need to downcast results[i] from CosmosElement to a CosmosObject anyway so this line works: 

"foreach (KeyValuePair<string, CosmosElement> kvp in cosmosObject)"

We could do the casting outside the if statement though if you think it's cleaner, then just return or assert(false).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Ah, yes, this should've been cleaned out. Was just a testing name. Renamed to 'normalizedResult'.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: That's because the order of the groups also changes depending on the order in which we access the partitions. The original GroupBy cases without MakeList/MakeSet also use the HashSet.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Was not needed, these usings keep getting added automatically when I'm testing how to do different things, I just need to clean them out at the end.

ezrahaleva-msft: Needed for 'ReadOnlyDictionary'

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: I don't think Group By would work in this test. To validate the results we'd have to do the same thing we're doing in GroupByQueryTests.cs, so might as well put those queries in there. We could replicate the test we're doing here in the GroupByQueryTests though.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Removed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Added.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/AggregateOperator.cs
ezrahaleva-msft: Fixed.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/SingleGroupAggregator.cs
ezrahaleva-msft: Fixed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
ezrahaleva-msft: Added coverage for this case to GroupByQueryTests.cs

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/GroupByQueryTests.cs
ezrahaleva-msft: Changed to explicit cast, and if results[i] is not a CosmosObject we assert false.

adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.AggregateMixedTypes_baseline.xml
```diff
@@ -102,106 +102,208 @@ WHERE c.key = "undefinedKey"]]></Query>
     <Aggregation />
   </Result>
   <Result>
-    <Query><![CDATA[SELECT VALUE MIN (c.field)
+    <Query><![CDATA[SELECT VALUE MAKELIST (c.field)
```
neildsh, line 105: >[SELECT VALUE MAKELIST (c.field) [](http://example.com/codeflow?start=19&length=32)

Can you put these tests at the end so that we dont see any diffs for existing tests ?

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeSetAggregator.cs
```diff
@@ -0,0 +1,77 @@
...
+    using Microsoft.Azure.Cosmos.CosmosElements;
+    using Microsoft.Azure.Cosmos.CosmosElements.Numbers;
+    using Microsoft.Azure.Cosmos.Query.Core.Exceptions;
+    using Microsoft.Azure.Cosmos.Query.Core.Monads;
+
+    internal sealed class MakeSetAggregator : IAggregator
+    {
+        private readonly HashSet<CosmosElement> globalSet;
+
+        private MakeSetAggregator(CosmosArray initialSet)
```
neildsh, line 22: >initialSet [](http://example.com/codeflow?start=46&length=10)

Please add a null check

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        private readonly List<CosmosElement> globalList;
+
+        private MakeListAggregator(CosmosArray initialList)
+        {
+            this.globalList = new List<CosmosElement>();
+            foreach (CosmosElement listItem in initialList)
+            {
+                this.globalList.Add(listItem);
+            }
```
neildsh, line 28: Please use AddRange instead of loop

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeSetAggregator.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            this.globalSet = new HashSet<CosmosElement>();
+            foreach (CosmosElement setItem in initialSet)
+            {
+                this.globalSet.Add(setItem);
+            }
```
neildsh, line 28: Please use UnionWith instead of loop

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.AggregateMixedTypes_baseline.xml
```diff
@@ -102,106 +102,208 @@ WHERE c.key = "undefinedKey"]]></Query>
...
-    <Aggregation><![CDATA[8204]]></Aggregation>
+    <Aggregation><![CDATA[[8204,8204,14799,14799,19377,19377,21344,21344,27832,27832,28384,28384,31256,31256,39908,39908,50445,50445,59354,59354,62645,62645,63899,63899,68915,68915,70086,70086,75246,75246,80799,80799,88308,88308,92569,92569,94673,94673,94726,94726]]]></Aggregation>
   </Result>
   <Result>
-    <Query><![CDATA[SELECT VALUE MIN (c.field)
+    <Query><![CDATA[SELECT VALUE MAKELIST (c.field)
 FROM c
 WHERE IS_OBJECT(c.field)]]></Query>
-    <Aggregation />
+    <Aggregation><![CDATA[[{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{}]]]></Aggregation>
```
neildsh, line 132: >{} [](http://example.com/codeflow?start=27&length=2)

Can you check why we are seeing these undefined results in the output? Ideally, an undefined should not be present in the output

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -996,5 +1108,103 @@ public async Task TestNonValueAggregatesAsync()
...
+        [TestMethod]
+        public async Task TestArrayAggregatesWithContinuationTokenAsync()
+        {
+            await this.TestArrayAggregatesWithContinuationToken(100);
+
+            // using 2048 + 1 documents here to ensure list size hits continuation token limit of 16KB
+            // We aggregates c.age (integers) which has 8 bytes, 16KB / 8B = 2048
+            await this.TestArrayAggregatesWithContinuationToken(2049);
+        }
+        private async Task TestArrayAggregatesWithContinuationToken(int numDocuments)
```
neildsh, line 1121: [Nit] newline please

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.cs
```diff
@@ -996,5 +1108,103 @@ public async Task TestNonValueAggregatesAsync()
...
+        {
+            int seed = (int)(DateTime.UtcNow - new DateTime(1970, 1, 1)).TotalSeconds;
```
neildsh, line 1123: >seed [](http://example.com/codeflow?start=16&length=4)

Would you mind using a fixed seed so that the test is deterministic?

ezrahaleva-msft: Done.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.AggregateMixedTypes_baseline.xml
ezrahaleva-msft: These empty objects are by design of the test, not actually undefineds. This is from the document generation code of that test, can see we're inserting a "new object {}" as the value of c.field

Document objectDoc = new Document();
objectDoc.SetPropertyValue(args.PartitionKey, Guid.NewGuid());
objectDoc.SetPropertyValue(args.Field, new object { });
documents.Add(objectDoc.ToString());
objectDoc.SetPropertyValue(args.PartitionKey, args.ObjectOnlyKey);
documents.Add(objectDoc.ToString());

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/AggregateQueryTests.AggregateMixedTypes_baseline.xml
ezrahaleva-msft: Done.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeSetAggregator.cs
ezrahaleva-msft: Fixed.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeSetAggregator.cs
ezrahaleva-msft: Added.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
ezrahaleva-msft: Fixed.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,72 @@
...
+    using Microsoft.Azure.Cosmos.CosmosElements;
+    using Microsoft.Azure.Cosmos.CosmosElements.Numbers;
+    using Microsoft.Azure.Cosmos.Query.Core.Exceptions;
+    using Microsoft.Azure.Cosmos.Query.Core.Monads;
+
+    internal sealed class MakeListAggregator : IAggregator
+    {
+        private readonly List<CosmosElement> globalList;
+
+        private MakeListAggregator(CosmosArray initialList)
```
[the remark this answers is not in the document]
ezrahaleva-msft: Added.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+            else
+            {
+                partialList = CosmosArray.Empty;
+            }
+
+            return TryCatch<IAggregator>.FromResult(new MakeListAggregator(initialList: partialList));
+        }
+
+        public CosmosElement GetCosmosElementContinuationToken()
```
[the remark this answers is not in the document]
neildsh: This is not being used in the SDk client code path, so we will keep this for now. I need to do a clean up of these code paths later.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
ezrahaleva-msft: Confirmed with Neil. This method won't be called in the sdk, only in the compute path, which he will fix himself.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
ezrahaleva-msft: Cases covered in AggregateQueryTests.cs.

ezrahaleva-msft: Added cases in AggregateQueryTests.cs

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private MakeListAggregator(CosmosArray initialList)
+        {
+            this.globalList = new List<CosmosElement>();
+
+            // InitialList should never be null, but if it is we just keep global list as an empty list. 
+            if (initialList == null)
+            {
+                return;
+            }
```
sboshra, line 30: Assert(initialList != null)

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
```diff
@@ -0,0 +1,76 @@
...
+    using Microsoft.Azure.Cosmos.CosmosElements;
+    using Microsoft.Azure.Cosmos.CosmosElements.Numbers;
+    using Microsoft.Azure.Cosmos.Query.Core.Exceptions;
+    using Microsoft.Azure.Cosmos.Query.Core.Monads;
+
+    internal sealed class MakeListAggregator : IAggregator
+    {
+        private readonly List<CosmosElement> globalList;
+
+        private MakeListAggregator(CosmosArray initialList)
```
sboshra, line 22: >CosmosArray initialLis [](http://example.com/codeflow?start=35&length=22)

perhaps we remove this and follow the object create with a call to Aggregate

sboshra (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
ezrahaleva-msft: Done.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/Aggregators/MakeListAggregator.cs
ezrahaleva-msft: Replaced with a call to Aggregate() which is already performing the null check, so this is no longer needed.

khdang (approved):

sboshra (approved):
:shipit:

PR #4507, opened 2024-05-19
Query : Fixes ORDER BY query issue when partial partition key is specified with hierarchical partition

### LINKED ISSUE #4472: Incorrect order of items on "Order by" query on multi-partition container
**Describe the bug**
When I run a targeted multi-partition query with a partition key that includes the first level of subpartitioning and "Order by" keyword, the response contains items with incorrect order. This happens only if QueryRequestOptions.PartitionKey is set.  

**To Reproduce**
I created [a console application](https://github.com/Azure/azure-cosmos-dotnet-v3/files/15248311/GetItemsOrderBy.zip) to reproduce the issue. It is based on our test DB. Please download the solution, open the Program.cs and update key (can be supplied separately).
Comment out the line 50 (requestOptions: requestOptions) and all works as expected.

**Expected behavior**
The output should be sorted. 

**Actual behavior**
The output does not sorted 

**Environment summary**
SDK Version: 3.39.1
OS Version: Windows 11 Enterprise

adityasa:
# Pull Request Template

## Description

If partial partition key is specified during query execution and the container partition is split at the top level of the partition key, ORDER BY query results may be wrong. This happens because query is incorrectly treated as a streaming ORDER BY query, even though it spans across multiple physical partitions. The fix is to ensure cross-partition ORDER BY pipeline gets used, which applies global sort on results obtained from each backend partition (using k-way merge). Since the physical partitions are already resolved by the time this determination is made (they just happen to get ignored at the moment), the fix is straightforward and bulk of the change is test related.

## Type of change

## Closing issues

To automatically close an issue: closes #4472

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
@@ -248,6 +248,8 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
                    inputParameters.InitialFeedRange,
                    trace);
 
+            Debug.Assert(targetRanges != null, $"{nameof(CosmosQueryExecutionContextFactory)} Assert!", "targetRanges != null");
+
             TryCatch<IQueryPipelineStage> tryCreatePipelineStage;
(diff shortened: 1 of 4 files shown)


sboshra (approved):
:shipit:

Maya-Painter (approved):

kirankumarkolli (approved):

PR #4515, opened 2024-05-24
[Internal] Query : Fixes response parsing error for ReadFeed by limiting binary format to query operations

adityasa:
# Pull Request Template

## Description

This change restricts use of SupportedSerializationFormats header to query scenarios only. Technically, this header is only used by QueryRequestOptions, however, the QueryRequestOptions (and therefore this header) also get invoked for ReadFeed. There is no direct API for ReadFeed, instead any of the query APIs (GetItemLinqQueryable, GetItemQueryIterator<T>, GetItemStreamQueryIterator) invoke ReadFeed operation (instead of Query operation) if supplied query (or QueryDefinition) is null.

ReadFeed API in the backend currently doesn't support this header and returns the response in text. However as it turns out, if this API is invoked with a specific partition key in the request options, backend honors the header. This is an issue, since SDK doesn't support handling of binary response in ReadFeed codepath.

As a result, when query APIs are invoked with query = null and a specific partition key, response cannot be parsed successfully. The issue is recently discovered after new emulator is published, which enabled binary encoding by default (when /EnablePreview flag is supplied). The issue was discovered in the form of test failures in the pipeline (which uses latest emulator with EnablePreview flag).
Currently following tests are failing:
EncryptionCreateItemAndQuery
ValidateSupportedSerializationFormatsRntbd
ValidateSupportedSerializationFormatsGateway

This change fixes following tests:
ValidateSupportedSerializationFormatsRntbd
ValidateSupportedSerializationFormatsGateway

The encryption project does not use the SDK code from master (by referencing project), but rather released SDK v3.35.4. The failing portion of the EncryptionCreateItemAndQuery test is made conditional, by only allowing to run when SDK project is referenced.

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos.Encryption/tests/EmulatorTests/Microsoft.Azure.Cosmos.Encryption.EmulatorTests.csproj
+++ b/Microsoft.Azure.Cosmos.Encryption/tests/EmulatorTests/Microsoft.Azure.Cosmos.Encryption.EmulatorTests.csproj
@@ -68,4 +68,8 @@
   <PropertyGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' == 'True' ">
     <DefineConstants>$(DefineConstants);ENCRYPTIONTESTPREVIEW</DefineConstants>
   </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(SdkProjectRef)' == 'True'">
+    <DefineConstants>$(DefineConstants);SDKPROJECTREF</DefineConstants>
+  </PropertyGroup>
 </Project>
--- a/Microsoft.Azure.Cosmos/src/Query/v2Query/DocumentQueryExecutionContextBase.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/v2Query/DocumentQueryExecutionContextBase.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Azure.Cosmos.Query
     internal abstract class DocumentQueryExecutionContextBase : IDocumentQueryExecutionContext
     {
         // ISSUE-TODO-adityasa-2024/5/27 - Re-enable binary format. Currently encryption codepath is not able to handle binary response.
-        public static readonly string DefaultSupportedSerializationFormats = string.Join(",", SupportedSerializationFormats.JsonText);
+        public static readonly string DefaultSupportedSerializationFormats = SupportedSerializationFormats.JsonText.ToString();
 
         public readonly struct InitParams
         {
--- a/Microsoft.Azure.Cosmos/src/Query/v2Query/DocumentQueryExecutionContextBase.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/v2Query/DocumentQueryExecutionContextBase.cs
@@ -32,8 +32,9 @@ namespace Microsoft.Azure.Cosmos.Query
 
     internal abstract class DocumentQueryExecutionContextBase : IDocumentQueryExecutionContext
     {
-        public static readonly string DefaultSupportedSerializationFormats = string.Join(",", SupportedSerializationFormats.JsonText, SupportedSerializationFormats.CosmosBinary);
-        
+        // ISSUE-TODO-adityasa-2024/5/27 - Re-enable binary format. Currently encryption codepath is not able to handle binary response.
+        public static readonly string DefaultSupportedSerializationFormats = string.Join(",", SupportedSerializationFormats.JsonText);
+
         public readonly struct InitParams
         {
             public IDocumentQueryClient Client { get; }
--- a/Microsoft.Azure.Cosmos/src/Query/v2Query/DocumentQueryExecutionContextBase.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/v2Query/DocumentQueryExecutionContextBase.cs
@@ -32,9 +32,8 @@ namespace Microsoft.Azure.Cosmos.Query
 
     internal abstract class DocumentQueryExecutionContextBase : IDocumentQueryExecutionContext
     {
-        // ISSUE-TODO-adityasa-2024/5/27 - Re-enable binary format. Currently encryption codepath is not able to handle binary response.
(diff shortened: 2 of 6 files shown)


Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -237,8 +237,11 @@ internal override void PopulateRequestOptions(RequestMessage request)
             {
                 request.Headers.Add(HttpConstants.HttpHeaders.ResponseContinuationTokenLimitInKB, this.ResponseContinuationTokenLimitInKb.ToString());
             }
-            
-            request.Headers.CosmosMessageHeaders.SupportedSerializationFormats = this.SupportedSerializationFormats?.ToString() ?? DocumentQueryExecutionContextBase.DefaultSupportedSerializationFormats;
+
+            if (request.OperationType == OperationType.Query)
```
Pilchie, line 241: Is it worth a comment explaining when you'd be in this codepath where OperationType is not Query?

FabianMeiswinkel: Why not just change this in service code? To always return text for ReadFeed? Then we can introduce a new SDK capability - BinaryReadFeed and only new SDKs with that capability backend would ever send binary? We have to avoid any regressions on old SDKs if we want to automatically enable binary everywhere (and we can't guarantee that the number of customers impacted is so small, that we can manually work with them to upgrade before enabling binary)

adityasa: I am not an expert on ReadFeed and question has been asked of appropriate folks to investigate this further. It appears that ReadFeed returns a response in Text, except when partition key is specified.

Microsoft.Azure.Cosmos/src/Query/v2Query/DocumentQueryExecutionContextBase.cs
```diff
@@ -32,8 +32,9 @@ namespace Microsoft.Azure.Cosmos.Query
 
     internal abstract class DocumentQueryExecutionContextBase : IDocumentQueryExecutionContext
     {
-        public static readonly string DefaultSupportedSerializationFormats = string.Join(",", SupportedSerializationFormats.JsonText, SupportedSerializationFormats.CosmosBinary);
-        
+        // ISSUE-TODO-adityasa-2024/5/27 - Re-enable binary format. Currently encryption codepath is not able to handle binary response.
+        public static readonly string DefaultSupportedSerializationFormats = SupportedSerializationFormats.JsonText.ToString();
```
ealsur, line 36: Is the comment to re-enable temporary?

ealsur: This change would be breaking this: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/504c2dfd8d6dacb77789a5b48c09897899363b55/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/FeedOptionTests.cs#L75

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
Maya-Painter: In the BE ReadFeed does not honor the SupportedSerializationFormats header and only returns text format currently.  It seems likely that ReadFeed with partition key becomes query in the BE.

(code ptr - https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB?path=/Product/Backend/native/StoreProvider/ResourceManagers/BaseResourceManager.PrepareForReadFeedContextBase.cpp&version=GBmaster&line=640&lineEnd=644&lineStartColumn=1&lineEndColumn=10&lineStyle=plain&_a=contents)

Microsoft.Azure.Cosmos/src/Query/v2Query/DocumentQueryExecutionContextBase.cs
adityasa: Only a temporary change (for validating if we could ever pass the encryption test with SDK project reference). Once the gates for you change finish, I will undo this change.

ealsur: Reverting this in a commit to verify all tests are passing now

Microsoft.Azure.Cosmos.Encryption/tests/EmulatorTests/Microsoft.Azure.Cosmos.Encryption.EmulatorTests.csproj
```diff
@@ -68,4 +68,8 @@
   <PropertyGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' == 'True' ">
     <DefineConstants>$(DefineConstants);ENCRYPTIONTESTPREVIEW</DefineConstants>
   </PropertyGroup>
+
```
adityasa, line 71: @ealsur I am guessing this needs to be reverted? Actually I am not too sure who made the change (it could have been me) since I also shared the branch with others too.

ealsur: I made this change. This is the way to communicate if the `SdkProjectRef` variable is being used on the build/test command, so on the test code, we can conditionally execute the ReadFeed test (in this case).

So when the build/test uses `/p:SdkProjectRef=true`, it runs the ReadFeed part and when it doesn't (Nuget reference), it doesn't

adityasa: Ah, yes that makes sense. I rushed in with my comment, but thanks for double-checking anyways.

ealsur (approved):

Maya-Painter (approved):

PR #4516, opened 2024-05-29
[Internal] Query: Adds support for binary encoded ReadFeed and Read responses

Maya-Painter:
## Description

This PR adds support for binary encoded ReadFeed responses.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Azure.Cosmos.Handlers
     using System.Net.Http;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos.Query;
     using Microsoft.Azure.Cosmos.Routing;
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Documents;
@@ -292,6 +293,11 @@ namespace Microsoft.Azure.Cosmos.Handlers
                     else if (operationType == OperationType.Patch)
                     {
                         request.Headers.ContentType = RuntimeConstants.MediaTypes.JsonPatch;
+                    }
+                    else if (operationType == OperationType.Read) 
+                    {
+                        request.Headers.CosmosMessageHeaders.SupportedSerializationFormats = DocumentQueryExecutionContextBase.DefaultSupportedSerializationFormats;
+                        request.Headers.CosmosMessageHeaders.ContentSerializationFormat = ContentSerializationFormat.CosmosBinary.ToString();
                     }
 
                     if (ChangeFeedHelper.IsChangeFeedWithQueryRequest(operationType, streamPayload != null))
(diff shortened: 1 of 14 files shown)


Microsoft.Azure.Cosmos/src/Resource/RewriteResponseUtils.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace Microsoft.Azure.Cosmos.Resource
+{
+    using System;
+    using System.IO;
+    using System.Runtime.InteropServices;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos.Json;
+    using Microsoft.Azure.Cosmos.Tracing;
+
+    internal class RewriteResponseUtils
```
adityasa, line 14: >RewriteResponseUtils [](http://example.com/codeflow?start=19&length=20)

Curious, just ResponseUtils?

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -241,7 +241,7 @@ internal override void PopulateRequestOptions(RequestMessage request)
             // All query APIs (GetItemQueryIterator, GetItemLinqQueryable and GetItemQueryStreamIterator) turn into ReadFeed operation if query text is null.
             // In such a case, query pipelines are still involved (including QueryRequestOptions). In general backend only honors SupportedSerializationFormats
             //  for OperationType Query but has a bug where it returns a binary response for ReadFeed API when partition key is also specified in the request.
-            if (request.OperationType == OperationType.Query)
+            if (request.OperationType == OperationType.Query || request.OperationType == OperationType.ReadFeed)
```
adityasa, line 244: This would require a few test updates. Kindly confirm (otherwise I am guessing we would have test gaps).

Maya-Painter: It is passing all gates. Which tests are you thinking about in particular?  Before the rewrite there were multiple readfeed failures (same as we saw when the emulator was first upgraded).

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -293,6 +294,11 @@ internal class RequestInvokerHandler : RequestHandler
                     {
                         request.Headers.ContentType = RuntimeConstants.MediaTypes.JsonPatch;
                     }
+                    else if (operationType == OperationType.Read) 
+                    {
+                        request.Headers.CosmosMessageHeaders.SupportedSerializationFormats = DocumentQueryExecutionContextBase.DefaultSupportedSerializationFormats;
```
Maya-Painter, line 299: Apparently SupportedSerializationFormats header isn't honored by backend for read operations, even though it is expected to. Using ContentSerializationFormat for testing for now.

FabianMeiswinkel: @Maya-Painter  - Shivam merged a PR for this a while back (several weeks)  - did you test this against a prod account or emulator - and if emulator with /EnablePreview or without?

Maya-Painter: Tested on both a prod account and the newest emulator with /EnablePreview and got the same result. I've looked at Shivam's change and I see no reason why it shouldn't work, so this is surprising.

PR #4523, opened 2024-06-03
Change Feed / Processor AVAD: Fixes timeToLiveExpired missing from metadata

philipthomas-MSFT:
# Pull Request Template

## Description

Bug recognize when onboarding a customer that needed to set TTL on the item. _timeToLiveExpired_ was not being deserialized and always returning default Boolean value of false. The TTL deletes where recognized in the feed, just wrong value for _timeToLiveExpired_.

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -68,6 +68,6 @@ class ChangeFeedMetadata
         /// Used to distinquish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
         /// </summary>
         [JsonProperty(PropertyName = "timeToLiveExpired", NullValueHandling= NullValueHandling.Ignore)]
-        public bool IsTimeToLiveExpired { get; }
+        public bool IsTimeToLiveExpired { get; set; }
```
kirankumarkolli, line 71: Is it supposed to be readonly? 
In what scenarios customers needs to set it?

philipthomas-MSFT: Deserialization does not work without this. This is the reason for the bug fix. Without, it never gets deserialized from the backend response.

kirankumarkolli: As public contract public setter is not desired.

Deserialization is implementation aspect and can be decoupled from public API contract. ex: internal set or wrapped property?

philipthomas-MSFT: Thanks for the review. Much appreciated. As I mentioned in chat, I changed that over 3 hours ago, which is why I resolved the conversation.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +31,128 @@ public async Task Cleanup()
...
+            Exception exception = default;
+            int ttlInSeconds = 5;
+            Stopwatch stopwatch = new();
+
+            ChangeFeedProcessor processor = monitoredContainer
+                .GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes(processorName: "processor", onChangesDelegate: (ChangeFeedProcessorContext context, IReadOnlyCollection<ChangeFeedItem<dynamic>> docs, CancellationToken token) =>
+                {
+                    // NOTE(philipthomas-MSFT): Please allow these Debug lines because TTL on items will purge at random times so I am using this to test when ran locally using emulator.
+
+                    Debug.WriteLine($"@ {DateTime.Now}, {nameof(stopwatch)} -> CFP AVAD took '{stopwatch.ElapsedMilliseconds}' to read document CRUD in feed.");
```
kirankumarkolli, line 51: This only traces in DEBUG mode, is that intentional? 

For tests we can be verbose as much as possible.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +31,128 @@ public async Task Cleanup()
...
+        /// </summary>
+        /// <param name="stopwatch"></param>
+        /// <param name="processor"></param>
+        /// <returns></returns>
+        private static async Task CheckIfAllDocumentsAreProcessed(
+            Stopwatch stopwatch, 
+            ChangeFeedProcessor processor)
+        {
+            do
+            {
```
kirankumarkolli, line 146: This will just keep spinning CPU right? 

Better alternative is to use eventing mechanism, ref: https://learn.microsoft.com/en-us/dotnet/api/system.threading.manualresetevent?view=net-8.0

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.json
```diff
@@ -89,7 +89,7 @@
           "Attributes": [
             "JsonPropertyAttribute"
           ],
-          "MethodInfo": "Boolean IsTimeToLiveExpired;CanRead:True;CanWrite:False;Boolean get_IsTimeToLiveExpired();IsAbstract:False;IsStatic:False;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+          "MethodInfo": "Boolean IsTimeToLiveExpired;CanRead:True;CanWrite:True;Boolean get_IsTimeToLiveExpired();IsAbstract:False;IsStatic:False;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
```
kirankumarkolli, line 92: Contract might needs to refresh.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: Yes, this is intentional. I am running debug viewer to see the results in real time.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.json
philipthomas-MSFT: checking on this, unless **UpdateContracts.ps1** treats _internal_ as if they are _public_.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: As explained in the method summary and comments, I am moving away from the norm like the other tests, which is **ManualResetEvent**, because I do not have a 'x' number of minutes to run this type of tests for it to be successful. The purging based on ttl is so random that sometimes the purged items come back instantly and sometimes it comes back several minutes later. I also marked this **TestCategory** as **LongRunning** so that it does not get ran like out tests.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.json
philipthomas-MSFT: Also, **private set;** may work for this as well. I think the absence of a setter is what is throwing the serializer off.

philipthomas-MSFT: Tested this with **private set;** yields the desired result of **"timeToLiveExpired":true** for the deleted item. So I will update to use _private_ instead of **internal**.

```
Debug Trace:
@ 6/4/2024 8:32:46 AM, CFProcessor starting...
@ 6/4/2024 8:32:47 AM, Document created.
@ 6/4/2024 8:32:51 AM, stopwatch -> CFP AVAD took '5503' to read document CRUD in feed.
@ 6/4/2024 8:32:51 AM, docs -> [{"current":{"id":"1","pk":"1","description":"Testing TTL on CFP.","ttl":5,"_rid":"YCAKAI71+jUBAAAAAAAAAA==","_self":"dbs/YCAKAA==/colls/YCAKAI71+jU=/docs/YCAKAI71+jUBAAAAAAAAAA==/","_etag":"\"00000000-0000-0000-b67b-4ea69a1201da\"","_attachments":"attachments/","_ts":1717504367},"metadata":{"crts":1717504367,"lsn":8,"operationType":"create","previousImageLSN":0,"timeToLiveExpired":false**}}]
@ 6/4/2024 8:36:17 AM, stopwatch -> CFP AVAD took '211081' to read document CRUD in feed.
@ 6/4/2024 8:36:17 AM, docs -> [{"current":{},"metadata":{"crts":1717504574,"lsn":9,"operationType":"delete","previousImageLSN":8,"timeToLiveExpired":true**},"previous":{"id":"1","pk":"1","description":"Testing TTL on CFP.","ttl":5,"_rid":"YCAKAI71+jUBAAAAAAAAAA==","_self":"dbs/YCAKAA==/colls/YCAKAI71+jU=/docs/YCAKAI71+jUBAAAAAAAAAA==/","_etag":"\"00000000-0000-0000-b67b-4ea69a1201da\"","_attachments":"attachments/","_ts":1717504367}}]
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: Notice how it takes 4 minutes for the purger on the emulator to recognize the ttl on the item. And the ttl on the item is set at 5 seconds. This is the randomness of the purger.

```
Debug Trace:
@ 6/4/2024 8:32:46 AM, CFProcessor starting...
@ 6/4/2024 8:32:47 AM, Document created.
@ 6/4/2024 8:32:51 AM, stopwatch -> CFP AVAD took '5503' to read document CRUD in feed.
@ 6/4/2024 8:32:51 AM, docs -> [{"current":{"id":"1","pk":"1","description":"Testing TTL on CFP.","ttl":5,"_rid":"YCAKAI71+jUBAAAAAAAAAA==","_self":"dbs/YCAKAA==/colls/YCAKAI71+jU=/docs/YCAKAI71+jUBAAAAAAAAAA==/","_etag":"\"00000000-0000-0000-b67b-4ea69a1201da\"","_attachments":"attachments/","_ts":1717504367},"metadata":{"crts":1717504367,"lsn":8,"operationType":"create","previousImageLSN":0,"timeToLiveExpired":false**}}]
@ 6/4/2024 8:36:17 AM, stopwatch -> CFP AVAD took '211081' to read document CRUD in feed.
@ 6/4/2024 8:36:17 AM, docs -> [{"current":{},"metadata":{"crts":1717504574,"lsn":9,"operationType":"delete","previousImageLSN":8,"timeToLiveExpired":true**},"previous":{"id":"1","pk":"1","description":"Testing TTL on CFP.","ttl":5,"_rid":"YCAKAI71+jUBAAAAAAAAAA==","_self":"dbs/YCAKAA==/colls/YCAKAI71+jU=/docs/YCAKAI71+jUBAAAAAAAAAA==/","_etag":"\"00000000-0000-0000-b67b-4ea69a1201da\"","_attachments":"attachments/","_ts":1717504367}}]
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.json
philipthomas-MSFT: **Discovery**: Changing it to **internal set;** or **private set;** and running **UpdateContracts.ps1** will still change **CanWrite:True**. Not sure this is an issue or not.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
ealsur, line 51: For traces un tests, there is a utility that will print it to console and also on the VS UI: `Logger.LogLine`

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +31,128 @@ public async Task Cleanup()
...
+                            // current
+                            Assert.AreEqual(expected: "1", actual: change.Current.id.ToString());
+                            Assert.AreEqual(expected: "1", actual: change.Current.pk.ToString());
+                            Assert.AreEqual(expected: "Testing TTL on CFP.", actual: change.Current.description.ToString());
+                            Assert.AreEqual(expected: ttlInSeconds, actual: change.Current.ttl.ToObject<int>());
+
+                            // metadata
+                            Assert.IsTrue(DateTime.TryParse(s: change.Metadata.ConflictResolutionTimestamp.ToString(), out _), message: "Invalid csrt must be a datetime value.");
+                            Assert.IsTrue(long.TryParse(change.Metadata.Lsn.ToString(), out _), message: "Invalid lsn must be a long value.");
+                            Assert.AreEqual(expected: false, actual: change.Metadata.IsTimeToLiveExpired);
```
ealsur, line 67:
```suggestion
                            Assert.IsFalse(change.Metadata.IsTimeToLiveExpired);
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +31,128 @@ public async Task Cleanup()
...
+
+                            // previous
+                            Assert.IsNull(change.Previous);
```
ealsur, line 70: Should we validate the other metadata properties also?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +31,128 @@ public async Task Cleanup()
...
+
+            // NOTE(philipthomas-MSFT): Please allow these Debug lines because TTL on items will purge at random times so I am using this to test when ran locally using emulator.
+
+            Debug.WriteLine($"@ {DateTime.Now}, Document created.");
+
+            await GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.CheckIfAllDocumentsAreProcessed(
+                stopwatch: stopwatch,
+                processor: processor);
+
+            if (exception != default)
```
ealsur, line 127: Validate that the stopwatch is stopped? That way you can validate that the delete was detected? Otherwise, how do you know the `else if (change.Metadata.OperationType == ChangeFeedOperationType.Delete)` entered once?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +31,128 @@ public async Task Cleanup()
...
+                        {
+                            // current
+                            Assert.AreEqual(expected: "1", actual: change.Current.id.ToString());
+                            Assert.AreEqual(expected: "1", actual: change.Current.pk.ToString());
+                            Assert.AreEqual(expected: "Testing TTL on CFP.", actual: change.Current.description.ToString());
+                            Assert.AreEqual(expected: ttlInSeconds, actual: change.Current.ttl.ToObject<int>());
+
+                            // metadata
+                            Assert.IsTrue(DateTime.TryParse(s: change.Metadata.ConflictResolutionTimestamp.ToString(), out _), message: "Invalid csrt must be a datetime value.");
+                            Assert.IsTrue(long.TryParse(change.Metadata.Lsn.ToString(), out _), message: "Invalid lsn must be a long value.");
```
ealsur, line 66: This Assert will never be false. `change.Metadata.Lsn` is a `long`, so the `ToString()` will always be a long. Ideally the validation should be checking if the value is > 0:

```suggestion
                            Assert.IsTrue(change.Metadata.Lsn > 0, message: "Invalid lsn must be a long value.");
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +31,128 @@ public async Task Cleanup()
...
+                            // stop after reading delete since it is the last document in feed.
+                            stopwatch.Stop();
+
+                            // current
+                            Assert.IsNull(change.Current.id);
+
+                            // metadata
+                            Assert.IsTrue(DateTime.TryParse(s: change.Metadata.ConflictResolutionTimestamp.ToString(), out _), message: "Invalid csrt must be a datetime value.");
+                            Assert.IsTrue(long.TryParse(change.Metadata.Lsn.ToString(), out _), message: "Invalid lsn must be a long value.");
+                            Assert.AreEqual(expected: true, actual: change.Metadata.IsTimeToLiveExpired);
```
ealsur, line 83:
```suggestion
                            Assert.IsTrue(change.Metadata.IsTimeToLiveExpired);
```

ealsur (commented):
Please update the title in a way that makes sense for a customer when reading it in the changelog. Maybe `Change Feed / Processor AVAD: Fixes timeToLiveExpired missing from metadata` or something that when an external user without much context reads it, can understand what it fixed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: because the stopwatch.Stop() is being called inside of that **else-if condition**. It's going to keep going until that **else-if condition** is met, then it immediately calls stopwatch.Stop() which breaks the loop.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: committed your suggestions on other conversations.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: Sure, I can switch over to that and try that.

philipthomas-MSFT: using Logger.LogLine as recommended.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.json
philipthomas-MSFT: Please unresolved if we need more clarification.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: And here is one that took less than a minute. Very random

```
@ 6/4/2024 3:40:45 PM, CFProcessor starting...
@ 6/4/2024 3:40:46 PM, Document created.
@ 6/4/2024 3:40:50 PM, stopwatch -> CFP AVAD took '5579' to read document CRUD in feed.
@ 6/4/2024 3:40:50 PM, docs -> [{"current":{"id":"1","pk":"1","description":"Testing TTL on CFP.","ttl":5,"_rid":"ACIkALl2AFkBAAAAAAAAAA==","_self":"dbs/ACIkAA==/colls/ACIkALl2AFk=/docs/ACIkALl2AFkBAAAAAAAAAA==/","_etag":"\"00000000-0000-0000-b6b7-188fb86f01da\"","_attachments":"attachments/","_ts":1717530046},"metadata":{"crts":1717530046,"lsn":1030,"operationType":"create","previousImageLSN":0,"timeToLiveExpired":false}}]
@ 6/4/2024 3:41:16 PM, stopwatch -> CFP AVAD took '30724' to read document CRUD in feed.
@ 6/4/2024 3:41:16 PM, docs -> [{"current":{},"metadata":{"crts":1717530075,"lsn":1031,"operationType":"delete","previousImageLSN":1030,"timeToLiveExpired":true},"previous":{"id":"1","pk":"1","description":"Testing TTL on CFP.","ttl":5,"_rid":"ACIkALl2AFkBAAAAAAAAAA==","_self":"dbs/ACIkAA==/colls/ACIkALl2AFk=/docs/ACIkALl2AFkBAAAAAAAAAA==/","_etag":"\"00000000-0000-0000-b6b7-188fb86f01da\"","_attachments":"attachments/","_ts":1717530046}}]
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,128 @@ public async Task Cleanup()
...
+        {
+            do
+            {
+                if (!stopwatch.IsRunning)
+                {
+                    await processor.StopAsync();
+
+                    break;
+                }
+            } while (stopwatch.IsRunning);
```
ealsur, line 154: To avoid the CPU block you can use ManualResetEvent: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/5a28704392dfa3c74058c41566e9222694cab66e/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs#L45

You can make the code wait until the flag is set and have the handler code set the flag where today you are stopping the watchdog.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,128 @@ public async Task Cleanup()
             await base.TestCleanup();
         }
 
+        [TestMethod]
```
ealsur, line 35:
```suggestion
        [TestMethod]
        [Timeout(30000)]
```

To avoid a bug causing an infinite wait, add a timeout to the test

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
kirankumarkolli: ManualResetEvent can be used to set and wait semantics. Please check https://learn.microsoft.com/en-us/dotnet/api/system.threading.eventwaithandle.set?view=net-8.0#system-threading-eventwaithandle-set

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: 30 second (30000) timeout is not going to be enough as I explained before. The purger sometimes returns the purged document back in a matter of seconds, and other times, minutes. So, I will use it and maybe cap it a 5 minutes (300000 ms).

See
https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4523#discussion_r1625939998

and

https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4523#discussion_r1626535810

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: fixed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: fixed

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: fixed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,106 @@ public async Task Cleanup()
...
+
+            await processor.StartAsync();
+            await Task.Delay(GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.ChangeFeedSetupTime);
+            await monitoredContainer.CreateItemAsync<dynamic>(new { id = "1", pk = "1", description = "Testing TTL on CFP.", ttl = ttlInSeconds }, partitionKey: new PartitionKey("1"));
+
+            // NOTE(philipthomas-MSFT): Please allow these Logger.LogLine because TTL on items will purge at random times so I am using this to test when ran locally using emulator.
+
+            Logger.LogLine($"@ {DateTime.Now}, Document created.");
+
+            allDocsProcessed.WaitOne();
```
kirankumarkolli, line 127: ManualResetEvent doesn't have async waiting, for the tests control-flow wise okey. 
One issue with blocking call might be that in-case of test taking longer than expected then test execution timeout (I am not sure what it is) might kill the test and loose the required traces. 

Graceful handling (timeout) of such to collect required diagnostics would be nice.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,106 @@ public async Task Cleanup()
...
+
+                    foreach (ChangeFeedItem<dynamic> change in docs)
+                    {
+                        if (change.Metadata.OperationType == ChangeFeedOperationType.Create)
+                        {
+                            // current
+                            Assert.AreEqual(expected: "1", actual: change.Current.id.ToString());
+                            Assert.AreEqual(expected: "1", actual: change.Current.pk.ToString());
+                            Assert.AreEqual(expected: "Testing TTL on CFP.", actual: change.Current.description.ToString());
+                            Assert.AreEqual(expected: ttlInSeconds, actual: change.Current.ttl.ToObject<int>());
```
kirankumarkolli, line 65: > there should be 1 create and 1 delete that will appear for that " +
            "document when using ChangeFeedProcessor with AllVersionsAndDeletes set as the ChangeFeedMode

from test description that's the expectation and body validation is unrelated right? 
If application has TTL value part of the document they will get it (similar to any other content in the document). 

Metadata has TTLExpired, and may be worth asserting that, thoughts?

kirankumarkolli: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/time-to-live#time-to-live-for-containers-and-items

If the scenario goal is to cover, detecting TTLed items in changeFeed then `IsTimeToLiveExpired ` suffice. 

If the goal of including TTL in document payload is to have less time than container TTL time?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,106 @@ public async Task Cleanup()
...
+        [TestCategory("LongRunning")]
+        [Owner("philipthomas-MSFT")]
+        [Description("Scenario: When a document is created with ttl set, there should be 1 create and 1 delete that will appear for that " +
+            "document when using ChangeFeedProcessor with AllVersionsAndDeletes set as the ChangeFeedMode.")]
+        public async Task WhenADocumentIsCreatedWithTtlSetThenTheDocumentIsDeletedTestsAsync()
+        {
+            ContainerInternal monitoredContainer = await this.CreateMonitoredContainer(ChangeFeedMode.AllVersionsAndDeletes);
+            Exception exception = default;
+            int ttlInSeconds = 5;
+            Stopwatch stopwatch = new();
```
kirankumarkolli, line 46: Is StopWatch still needed? (its used earlier for synchronization)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
ealsur, line 127:
```suggestion
            bool receivedDelete = allDocsProcessed.WaitOne(60000);
            Assert.IsTrue(receivedDelete, "Timed out waiting for docs to process");
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,106 @@ public async Task Cleanup()
...
+                .Build();
+
+            stopwatch.Start();
+
+            // NOTE(philipthomas-MSFT): Please allow these Logger.LogLine because TTL on items will purge at random times so I am using this to test when ran locally using emulator.
+
+            Logger.LogLine($"@ {DateTime.Now}, CFProcessor starting...");
+
+            await processor.StartAsync();
+            await Task.Delay(GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.ChangeFeedSetupTime);
```
ealsur, line 120: Please add a try/finally on anything after StartAsync, where processor.StopAsync() is called in the finally block. Leaving zombie CFPs will consume resources

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: Guys. Stopwatch is used for logging as I mentioned before. So I can see how long these purges are taking.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: I agree with Matias previous comment on testing the other properties. How does this hurt anything?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: Already handled now.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
philipthomas-MSFT: You sure this is necessary? And if so, then maybe we need to PR it later to fix all the other CFP tests instead of blocking this PR? Just a thought.

ealsur: What other CFP tests are not calling StopAsync at the end? And in any case, why introduce more debt?

Yes, it is necessary. If you don't stop the CFPz it will continue running in the background for the whole duration of the entire run.

philipthomas-MSFT: Done.

ealsur (dismissed):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,115 @@ public async Task Cleanup()
...
 
+        [TestMethod]
+        [Timeout(300000)]
+        [TestCategory("LongRunning")]
+        [Owner("philipthomas-MSFT")]
+        [Description("Scenario: When a document is created with ttl set, there should be 1 create and 1 delete that will appear for that " +
+            "document when using ChangeFeedProcessor with AllVersionsAndDeletes set as the ChangeFeedMode.")]
+        public async Task WhenADocumentIsCreatedWithTtlSetThenTheDocumentIsDeletedTestsAsync()
+        {
+            ContainerInternal monitoredContainer = await this.CreateMonitoredContainer(ChangeFeedMode.AllVersionsAndDeletes);
```
kirankumarkolli, line 43: Cleanup `monitoredContainer`?

Base test clean-up only doing database delete not container?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,115 @@ public async Task Cleanup()
...
+            Exception exception = default;
+            int ttlInSeconds = 5;
+            Stopwatch stopwatch = new();
+            ManualResetEvent allDocsProcessed = new ManualResetEvent(false);
```
kirankumarkolli, line 47: IDisposable, prefer using

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
kirankumarkolli: One possibility might Database delete will recursively delete all the containers inside? (Can you please check if that's the case).

kirankumarkolli: `leaseContaienr` is also needs clean-up (if Database delete don't already). It's an existing gap for other existing tests using leasecontainer. 

Other tests gaps is not a blocker for this PR.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
```diff
@@ -29,6 +32,115 @@ public async Task Cleanup()
...
+            try
+            {
+                await Task.Delay(GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.ChangeFeedSetupTime);
+                await monitoredContainer.CreateItemAsync<dynamic>(new { id = "1", pk = "1", description = "Testing TTL on CFP.", ttl = ttlInSeconds }, partitionKey: new PartitionKey("1"));
+
+                // NOTE(philipthomas-MSFT): Please allow these Logger.LogLine because TTL on items will purge at random times so I am using this to test when ran locally using emulator.
+
+                Logger.LogLine($"@ {DateTime.Now}, Document created.");
+
+                bool receivedDelete = allDocsProcessed.WaitOne(300000);
```
kirankumarkolli, line 130: Timeout value needs to be lower than the test timeout, so-that execution gets some buffer otherwise the TestExecutor might cancel it.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.cs
ealsur: All containers are cleaned up through `await base.TestCleanup();` on the `[TestCleanup]` phase.

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/20d678b60d8a76ae9f551bf590b1a798c62dd72a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/BaseCosmosClientHelper.cs#L50-L63

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -68,6 +68,6 @@ class ChangeFeedMetadata
         /// Used to distinquish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
         /// </summary>
         [JsonProperty(PropertyName = "timeToLiveExpired", NullValueHandling= NullValueHandling.Ignore)]
-        public bool IsTimeToLiveExpired { get; }
+        public bool IsTimeToLiveExpired { get; internal set; }
```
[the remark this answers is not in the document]
kirankumarkolli: @ealsur that's a great point.

Looking at STJ documentation: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/immutability

Microsoft.Azure.Cosmos.Samples/Usage/ReEncryption/ReEncryption.csproj
```diff
@@ -8,7 +8,7 @@
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.11.4" />
+    <PackageReference Include="Azure.Identity" Version="1.11.0" />
```
ealsur, line 11: Please undo this change. 1.11.4 was bumped up for a security vulnerability. Reference: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4544

ealsur (dismissed):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.json
kirankumarkolli, line 92: Is this file upto date?

philipthomas-MSFT: No changes. The file is up to date. All of them have CanWrite:True. This is normal behavior.

kirankumarkolli: Curious the internal set is all applied to other properties right, why in contract only one entry is showing up?

ealsur (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.json
philipthomas-MSFT: They all are internal set and have CanWrite:True

PR #2352, opened 2021-03-30
ReadMany: Adds ReadManyItems Api using Query under the covers

asketagarwal:
## Sample

Stream:
```
  IReadOnlyList<(string, PartitionKey)> itemList = new List<(string, PartitionKey)>
  {
       ("Id1", new PartitionKey("pkValue1")),
       ("Id2", new PartitionKey("pkValue2")),
       ("Id3", new PartitionKey("pkValue3"))
  };
   
   using (ResponseMessage responseMessage = await this.Container.ReadManyItemsStreamAsync(itemList))
   {
       using (Stream stream = response.ReadBodyAsync())
       {
           //Read or do other operations with the stream
           using (StreamReader streamReader = new StreamReader(stream))
           {
               string content = streamReader.ReadToEndAsync();
           }
       }
   }
```

Typed:

```
   public class ToDoActivity{
       public string id {get; set;}
       public string status {get; set;}
   }
  
   IReadOnlyList<(string, PartitionKey)> itemList = new List<(string, PartitionKey)>
   {
       ("Id1", new PartitionKey("pkValue1")),
       ("Id2", new PartitionKey("pkValue2")),
       ("Id3", new PartitionKey("pkValue3"))
   };
  
  FeedResponse<ToDoActivity> feedResponse = this.Container.ReadManyItemsAsync<ToDoActivity>(itemList);
```
## Closing issues

To automatically close an issue: closes #IssueNumber

PR #4533, opened 2024-06-06
ReadManyItemsAsync -> ReadManyTaskHelperAsync wraps forked IO task into Task.Run

### LINKED ISSUE #4527: ReadManyItemsAsync -> ReadManyTaskHelperAsync wraps forked IO task into Task.Run
**Describe the bug**
**`ReadManyTaskHelperAsync`** code has [suspicious Task.Run call](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/5a28704392dfa3c74058c41566e9222694cab66e/Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs#L99) that looks as a potential issue: `Task.Run` schedules task on thread pool's thread and potentially can cause thread pool being exhausted.

This `Task.Run` call looks unnecessary and instead of wrapping async code into `Task.Run` (that can put higher pressure on thread pool) `ReadManyTaskHelperAsync -> for loop -> tasks.Add(Task.Run(async () =>` why code doesn't do this instead:
`ReadManyTaskHelperAsync -> for loop -> tasks.Add(DoStuffAsync(params, semaphore))`

```
private Task<List<ResponseMessage>> DoStuffAsync(.. otherParams, SemaphoreSlim semaphore)
{
    try
    {
       QueryDefinition queryDefinition = ((this.partitionKeySelectors.Count == 1) && (this.partitionKeySelectors[0] == "[\"id\"]"))
            ? this.CreateReadManyQueryDefinitionForId(entry.Value, indexCopy)
            : this.CreateReadManyQueryDefinitionForOther(entry.Value, indexCopy);

        return await this.GenerateStreamResponsesForPartitionAsync(queryDefinition,
            entry.Key,
            readManyRequestOptions,
            childTrace,
            cancellationToken);
    }
    finally
    {
        semaphore.Release();
        childTrace.Dispose();
    }
}
```

**To Reproduce**
Run ReadManyTaskHelperAsync 

**Expected behavior**
`Task.Run` is not called in `ReadManyItemsAsync` execution. Thread pool is not used.

**Actual behavior**
`Task.Run` is called in `ReadManyItemsAsync` execution. Thread pool is used.

**Environment summary**
SDK Version: 3.39.1
OS Version: Windows

**Additional context**
There was potential issue reported [here](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3665) but it was not really proven that issue came out of this method call and reporter eventually mitigated with other solution.

cc @ealsur

yar-shukan:
# Pull Request Template

## Description
`ReadManyTaskHelperAsync` uses `Task.Run` due to usage of `SemaphoreSlim` and rate limiting amount of parallel calls. `Task.Run` uses thread pool and schedules execution to background thread. Optimize solution not to use the thread pool.

## Type of change

## Closing issues

To automatically close an issue: closes #4527

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
+++ b/Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
@@ -7,7 +7,6 @@ namespace Microsoft.Azure.Cosmos
     using System;
     using System.Collections;
     using System.Collections.Generic;
-    using System.Linq;
     using System.Net;
     using System.Text;
     using System.Threading;
@@ -85,46 +84,42 @@ namespace Microsoft.Azure.Cosmos
                                   CancellationToken cancellationToken)
         {
             SemaphoreSlim semaphore = new SemaphoreSlim(this.maxConcurrency, this.maxConcurrency);
-
-            List<(KeyValuePair<PartitionKeyRange, List<(string, PartitionKey)>> entry, int startIndex)> rangeEntries =
-                new List<(KeyValuePair<PartitionKeyRange, List<(string, PartitionKey)>> entry, int startIndex)>();
+            List<Task<List<ResponseMessage>>> tasks = new List<Task<List<ResponseMessage>>>();
 
             foreach (KeyValuePair<PartitionKeyRange, List<(string, PartitionKey)>> entry in partitionKeyRangeItemMap)
             {
                 // Fit MaxItemsPerQuery items in a single query to BE
                 for (int startIndex = 0; startIndex < entry.Value.Count; startIndex += this.maxItemsPerQuery)
                 {
-                    rangeEntries.Add((entry, startIndex));
+                    // Only allow 'maxConcurrency' number of queries at a time
+                    await semaphore.WaitAsync();
+
+                    ITrace childTrace = trace.StartChild("Execute query for a partitionkeyrange", TraceComponent.Query, TraceLevel.Info);
+                    int indexCopy = startIndex;
+                    tasks.Add(Task.Run(async () =>
+                    {
+                        try
+                        {
+                            QueryDefinition queryDefinition = ((this.partitionKeySelectors.Count == 1) && (this.partitionKeySelectors[0] == "[\"id\"]")) ?
+                                               this.CreateReadManyQueryDefinitionForId(entry.Value, indexCopy) :
+                                               this.CreateReadManyQueryDefinitionForOther(entry.Value, indexCopy);
+
+                            return await this.GenerateStreamResponsesForPartitionAsync(queryDefinition,
+                                                                       entry.Key,
+                                                                       readManyRequestOptions,
+                                                                       childTrace,
+                                                                       cancellationToken);
+                        }
+                        finally
+                        {
+                            semaphore.Release();
+                            childTrace.Dispose();
+                        }
+                    }));
                 }
             }
-
-            IEnumerable<Task<List<ResponseMessage>>> tasks = rangeEntries.Select(async (entry, startIndex) =>
-            {
-                // Only allow 'maxConcurrency' number of queries at a time
-                await semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
-
-                ITrace childTrace = trace.StartChild("Execute query for a partitionkeyrange", TraceComponent.Query, TraceLevel.Info);
-
-                try
-                {
-                    QueryDefinition queryDefinition = ((this.partitionKeySelectors.Count == 1) && (this.partitionKeySelectors[0] == "[\"id\"]")) ?
-                        this.CreateReadManyQueryDefinitionForId(entry.entry.Value, startIndex) :
-                        this.CreateReadManyQueryDefinitionForOther(entry.entry.Value, startIndex);
-
-                    return await this.GenerateStreamResponsesForPartitionAsync(queryDefinition,
-                        entry.entry.Key,
-                        readManyRequestOptions,
-                        childTrace,
-                        cancellationToken).ConfigureAwait(false);
-                }
-                finally
-                {
-                    semaphore.Release();
-                    childTrace.Dispose();
-                }
-            });
-
-            return await Task.WhenAll(tasks).ConfigureAwait(false);
+            
+            return await Task.WhenAll(tasks);
         }
 
         private async Task<IDictionary<PartitionKeyRange, List<(string, PartitionKey)>>> CreatePartitionKeyRangeItemListMapAsync(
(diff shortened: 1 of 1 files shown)


Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -84,42 +85,46 @@ internal sealed class ReadManyQueryHelper : ReadManyHelper
...
-                            childTrace.Dispose();
-                        }
-                    }));
+                    rangeEntries.Add((entry, startIndex));
                 }
             }
-            
-            return await Task.WhenAll(tasks);
+
+            IEnumerable<Task<List<ResponseMessage>>> tasks = rangeEntries.Select(async (entry, startIndex) =>
```
ealsur, line 101: This is creating all the Tasks in advance and then suspending them with the semaphore inside, versus the current approach which creates the Tasks only when there is room in the concurrency. Wouldn't the increase in concurrent Task creation cause increased CPU consumption?

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -84,42 +85,46 @@ internal sealed class ReadManyQueryHelper : ReadManyHelper
...
-                                                                       childTrace,
-                                                                       cancellationToken);
-                        }
-                        finally
-                        {
-                            semaphore.Release();
-                            childTrace.Dispose();
-                        }
-                    }));
+                    rangeEntries.Add((entry, startIndex));
```
ealsur, line 97: If the problem is the Task.Run, wouldn't creating an async method and calling that (passing the semaphore and other required references) as input to the `tasks.Add` call achieve the same result and remove the Task.Run?

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
yar-shukan: No, this doesn't create Tasks in advance: IEnumerable nature and `await semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);` inside of `.Select( )` lambda produces tasks lazily.

yar-shukan: Yes, this will create task objects, but no IO is made here:
image

Why tasks creation in advance would cause increased CPU consumption compared to problems of usage of thread pool in e.g. ASP.NET app?

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
yar-shukan: Yes, good point. Changed to that as was initially proposed in issue comment and removed usage of IEnumerable

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
yar-shukan: @ealsur would creating some syntatic Benchmark.NET microbenchmark of both versions (Task.Run) vs add all async tasks at once help in making judgement?

ealsur: I think the main point here is, understanding what is the current impact or the current problem. There must have been a reason for the semaphore preventing the unbounded creation of Tasks.

Thoughts @Pilchie , @kirankumarkolli, @FabianMeiswinkel ?

Pilchie: > There must have been a reason for the semaphore preventing the unbounded creation of Tasks.

I expect the real desire was to avoid unbounded parallelism of network calls, not the Tasks themselves.  The threadpool does a reasonable job of managing how many tasks actually execute in parallel, so I expect them main difference would be a little bit more memory for the tasks.

yar-shukan: At the end both options create list of tasks and final memory allocated is same, isn't it? It's just existing solution can spread allocation over time, but I guess the main point was to limit the concurrency.

BTW about concurrency limit: existing solution uses `Environment.ProcessorCount * 10` which is far away from being ok solution considering containerized environments like docker where container can have small amount of virtual cores e.g 1-4, but `Environment.ProcessorCount` is can be 32 or 64. But this should be addressed in scope of other work

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -91,37 +91,46 @@ internal async Task<List<ResponseMessage>[]> ReadManyTaskHelperAsync(IDictionary
...
-                                                                       childTrace,
-                                                                       cancellationToken);
-                        }
-                        finally
-                        {
-                            semaphore.Release();
-                            childTrace.Dispose();
-                        }
-                    }));
+                    tasks.Add(this.GenerateStreamResponsesForPartitionAsync(entry, startIndex, readManyRequestOptions, trace, semaphore, cancellationToken));
```
kirankumarkolli, line 94: Isn't just removing Task.Run from existing code suffice? (and equivalent of this refracturing?)

yar-shukan: Task.Run was using lamba before, so I had to move it to method now because I can't leave it inlined. or am I missing some point?

kirankumarkolli:
From the original PR related comment.
https://github.com/Azure/azure-cosmos-dotnet-v3/pull/2352/files#r608601340

With-out test coverage it might easily regress, @yar-shukan thoughts?

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
kirankumarkolli: My understanding of typical async stack is to lazy yield. 

Withe current scheduling on threadpool each task given chance to progress while the main on is creating tasks. Where as with change the main Task it-self bottleneck? 

One solution might be forced Yiled() thoughts?

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
kirankumarkolli, line 94: Can you please add code comments (just to make sure in future at-leas any changes will have local context)

kirankumarkolli: Also please update description with details (it will get referred from changelog)

kirankumarkolli:
Inactivity closing PR, please re-open when there are updates

PR #4538, opened 2024-06-10
Query: Fixes plumbing VectorEmbeddingPolicy to ServiceInterop to choose correct default distance function

neildsh:
## Description

This change plumbs the vector embedding policy to the service interop so that the query plan is generated with the correct default distance function when non is specified in the query by the user

## Type of change


adityasa (approved):
:shipit:

Maya-Painter (approved):

sboshra (approved):
:shipit:

PR #4550, opened 2024-06-14
Query: Fixes call to ServiceInterop in case of DISP_E_BUFFERTOOSMALL to account for vector embedding policy

neildsh:
# Pull Request Template

## Description

Fixes call to ServiceInterop in case of DISP_E_BUFFERTOOSMALL to account for vector embedding policy, and adds test cases to cover this path

## Type of change


adityasa (approved):
:shipit:

Maya-Painter (approved):

PR #4558, opened 2024-06-25
Query: Removes compute specific logic from query pipelines that is no longer required

neildsh:
## Description

This change removes compute specific logic from query pipelines that is no longer required


Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/AggregateQueryPipelineStage.Client.cs
```diff
@@ -28,7 +28,7 @@ private sealed class ClientAggregateQueryPipelineStage : AggregateQueryPipelineS
                 // all the work is done in the base constructor.
             }
 
-            public static TryCatch<IQueryPipelineStage> MonadicCreate(
+            public static new TryCatch<IQueryPipelineStage> MonadicCreate(
```
sboshra, line 31: >new [](http://example.com/codeflow?start=26&length=3)

Hmmm

sboshra (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Aggregate/AggregateQueryPipelineStage.Client.cs
neildsh: I will remove these in my next PR. I did not consolidate these classes since that would make the diffs unreadable.

khdang (approved):

PR #4589, opened 2024-07-16
SystemTextJsonSerializer: Refactors Code to Introduce `SystemTextJsonSerializerOptions` to Set the Default STJ Serializer

kundadebdatta:
# Pull Request Template

## Description

This PR marks the Default `CosmosSystemTextJsonSerializer` as `public` for both `preview` and `GA` release. To read more about the default serializer, please [refer to this PR](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4332).

## Type of change

## Closing issues


kirankumarkolli (dismissed):

NaluTripician (dismissed):

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Azure.Cosmos
     /// <summary>
     /// This class provides a default implementation of System.Text.Json Cosmos Linq Serializer.
     /// </summary>
-    internal class CosmosSystemTextJsonSerializer : CosmosLinqSerializer
+    public class CosmosSystemTextJsonSerializer : CosmosLinqSerializer
```
Pilchie, line 16: An alternate proposal would be to leave this private, and instead provide a `CosmosClientOption` of `UseSystemTextJsonForSerialization`, and then instantiate this privately.

Pilchie (approved):

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kundadebdatta: Thinking more on this, I feel, since we already have a public API `Serializer` in the `CosmosClientOptions` to set the custom serializer, we could simply use it and pass a new instance. 

We can add this method `UseSystemTextJsonForSerialization()` too in the cosmos client options, but my worry is that it will open the door for similar implementations for any other serializers in the future. Therefore, I am inclined towards keeping the implementation as is.

Thoughts on this @kirankumarkolli

kirankumarkolli: Some thoughts
- ClientOptions integration: Do allow discovery (code documentation as alternative is not equal)
- ClientOptions: For completeness any CTOR parameters also needs to be promoted into the ClientOptions (nit picking)
- Public type: Idea was that CX can also extend it to customize in any scenarios when needed (why/when is unknown but a possibility)
- Eventually (v4) we will only have default serializer and customized first-class options at-least not-anticipated right now (big assumption/statement).

kirankumarkolli: Two choices 
- ClientOptions and Serializer private: 
- Serializer public and include in ClientOptions

By V4 STJ will be default implementation, any feedback around if it needs to be public will help us by then. 
We can start with ClientOptions and hear the feedback as an option

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -833,8 +847,11 @@ internal Protocol ConnectionProtocol
         internal void SetSerializerIfNotConfigured(CosmosSerializer serializer)
         {
             if (this.serializerInternal == null)
-            {
-                this.serializerInternal = serializer ?? throw new ArgumentNullException(nameof(serializer));
+            {
+                this.serializerInternal = this.UseSystemTextJsonForSerialization
+                    ? new CosmosSystemTextJsonSerializer(
+                        new System.Text.Json.JsonSerializerOptions())
+                    : serializer ?? throw new ArgumentNullException(nameof(serializer));
```
Pilchie, line 854: I don't think this is enough.  If you look at where there is called, it comes from `CosmosSerializerCore.GetCustomOrDefaultSerializer()`, but the same `CosmosSerializerCore` is used in a couple of places.  Don't we need to get the STJSerializer into the SerializerCore so that it is used in other places too?

kundadebdatta: Thanks @Pilchie! The recent LINQ test failures did already hinted the same problem as you highlighted. The easiest fix would be to set the default serializer in the `ClientContextCore`, right before creating the `CosmosSerializerCore`.

```
            if (clientOptions.UseSystemTextJsonForSerialization)
            {
                clientOptions.Serializer = new CosmosSystemTextJsonSerializer(
                    new System.Text.Json.JsonSerializerOptions());
            }
```

This should address the failures and set it correctly.

Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs
kundadebdatta: Per our offline conversation, the first choice is probably the best fit at the moment. Made the code changes to address this.

Pilchie (dismissed):

Pilchie:
Maybe change the title of the PR?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -392,6 +392,19 @@ public ConnectionMode ConnectionMode
...
+        /// used for JSON serialization.
+        /// </summary>
+        /// <remarks>
+        /// Note that, if this flag is set to true, then any custom seriliazer provided will be disregarded and overridden with the
+        /// default STJ serializer <see cref="CosmosSystemTextJsonSerializer"/>.
+        /// </remarks>
+        /// <value>
+        /// The default value is false
+        /// </value>
+        public bool UseSystemTextJsonForSerialization { get; set; } = false;
```
kirankumarkolli, line 407: Customization is reasonably important when LINQ contract serialization is non-trivial. 
For Builder the contract can take SerializerSetting as argument. 

In ClientOptions: 
- So-far we only have properties => SystemTextJsonSerializerSettings as property? 

/cc: @jcocchi , @Pilchie thughts

kirankumarkolli, line 407: There are three serilization contracts
- CosmosSerializationOptions
- CosmosSerialzier contract
- STJSerializer

All three needs to be exclusive (i.e. ensure that only one is configured). 
So the setter for them needs to check for exclusivity and error out appropriately.

jcocchi: Not sure I understand, would setting this to true override any custom LINQ serialization settings?

kirankumarkolli: Update (Offline sync-up): Pivot to align with V4 API contract which is STJ and SETJ SerializerSettings.

kundadebdatta: This is now addressed and related tests to validate the same has been added to the PR.

kundadebdatta: Updated the contract as per discussion.

kirankumarkolli: Both properties I.e. use and properties seems lots more overloaded. Can it be simplified to one property?

kirankumarkolli: Just properties might be fine.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -392,6 +392,30 @@ public ConnectionMode ConnectionMode
...
+        ///     {
+        ///         DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull,
+        ///     }
+        /// };
+        /// 
+        /// CosmosClient client = new CosmosClient("endpoint", "key", clientOptions);
+        /// ]]>
+        /// </code>
+        /// </example>
+        /// <remarks>Note that if this option is provided, then the SDK will use the <see cref="CosmosSystemTextJsonSerializer"/> as the default serializer
```
Pilchie, line 415: Remarks doesn't show up in Visual Studio, so we might want to move this directly into `<summary>`.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Refactored the code to reflect the same.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -392,6 +392,29 @@ public ConnectionMode ConnectionMode
...
+        ///     {
+        ///         DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull,
+        ///     }
+        /// };
+        /// 
+        /// CosmosClient client = new CosmosClient("endpoint", "key", clientOptions);
+        /// ]]>
+        /// </code>
+        /// </example>
+        public System.Text.Json.JsonSerializerOptions SystemTextJsonSerializerOptions { get; set; }
```
kirankumarkolli, line 417: Setter needs to be conditional on others not set right? 

Same for other Serializer property too.

kundadebdatta: This should already be taken care in the `ClientContextCore` .

```
            if (clientOptions.SystemTextJsonSerializerOptions != null)
            {
                if (clientOptions.Serializer != null || clientOptions.SerializerOptions != null)
                {
                    throw new ArgumentException(
                        $"{nameof(clientOptions.SystemTextJsonSerializerOptions)} is not compatible with {nameof(clientOptions.Serializer)} or {nameof(clientOptions.SerializerOptions)}. Only one can be set.  ");
                }

                clientOptions.Serializer = new CosmosSystemTextJsonSerializer(
clientOptions.SystemTextJsonSerializerOptions);
            }
```

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
```diff
@@ -128,6 +128,18 @@ private static RemoteCertificateValidationCallback SslCustomValidationCallBack(F
                 requestInvokerHandler = clientPipelineBuilder.Build();
             }
 
+            if (clientOptions.SystemTextJsonSerializerOptions != null)
+            {
+                if (clientOptions.Serializer != null || clientOptions.SerializerOptions != null)
+                {
+                    throw new ArgumentException(
```
kirankumarkolli, line 135: Only one choice can be validated at the options settings it-self avoids real initialization for detection.

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
```diff
@@ -128,6 +128,18 @@ private static RemoteCertificateValidationCallback SslCustomValidationCallBack(F
...
+                        $"{nameof(clientOptions.SystemTextJsonSerializerOptions)} is not compatible with {nameof(clientOptions.Serializer)} or {nameof(clientOptions.SerializerOptions)}. Only one can be set.  ");
+                }
+
+                clientOptions.Serializer = new CosmosSystemTextJsonSerializer(
```
kirankumarkolli, line 139: Do client see a Serailzier thought they didn't initialize it?

kundadebdatta: Correct. And this is applicable even with the existing code as well. I did double checked (with the master branch) that the serializer, by behavior gets initialized to a default JSON wrapper serializer, if none provided. This is set by the below piece of code in `ClientContextCore`.

```
            // This sets the serializer on client options which gives users access to it if a custom one is not configured.
            clientOptions.SetSerializerIfNotConfigured(serializerCore.GetCustomOrDefaultSerializer());
```

{52B901FB-F063-43FE-A0BA-1F9F4356004C}

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli: That's the existing pattern right?

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
kundadebdatta: The validation logic is moved to the `CosmosClientOptions`. However, the initialization remains in the `ClientContextCore`.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -543,10 +580,10 @@ public CosmosSerializationOptions SerializerOptions
             get => this.serializerOptions;
             set
             {
-                if (this.Serializer != null)
+                if (this.Serializer != null || this.SystemTextJsonSerializerOptions != null)
                 {
                     throw new ArgumentException(
-                        $"{nameof(this.SerializerOptions)} is not compatible with {nameof(this.Serializer)}. Only one can be set.  ");
+                        $"{nameof(this.SerializerOptions)} is not compatible with {nameof(this.Serializer)} or {nameof(this.SystemTextJsonSerializerOptions)}. Only one can be set.  ");
```
kirankumarkolli, line 586: Line 618: Also needs check right?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Azure.Cosmos
     using System.Net;
     using System.Net.Http;
     using System.Net.Security;
-    using System.Security.Cryptography.X509Certificates;
+    using System.Security.Cryptography.X509Certificates;
```
kirankumarkolli, line 15: Is this related?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -392,6 +393,42 @@ public ConnectionMode ConnectionMode
         /// <seealso cref="ItemRequestOptions.EnableContentResponseOnWrite"/>
         /// <seealso cref="TransactionalBatchItemRequestOptions.EnableContentResponseOnWrite"/>
         public bool? EnableContentResponseOnWrite { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="System.Text.Json.JsonSerializerOptions"/> for the default STJ serializer <see cref="CosmosSystemTextJsonSerializer"/>.
```
kirankumarkolli, line 398:
```suggestion
        /// Gets or sets the <see cref="System.Text.Json.JsonSerializerOptions"/> for the System.Text.Json serializer.

```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -392,6 +393,42 @@ public ConnectionMode ConnectionMode
         /// <seealso cref="ItemRequestOptions.EnableContentResponseOnWrite"/>
         /// <seealso cref="TransactionalBatchItemRequestOptions.EnableContentResponseOnWrite"/>
         public bool? EnableContentResponseOnWrite { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="System.Text.Json.JsonSerializerOptions"/> for the default STJ serializer <see cref="CosmosSystemTextJsonSerializer"/>.
+        /// Note that if this option is provided, then the SDK will use the <see cref="CosmosSystemTextJsonSerializer"/> as the default serializer and set
```
kirankumarkolli, line 399:
```suggestion
        /// Note that if this option is provided, then the SDK will use the System.Text.Json as the default serializer and set

```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -392,6 +393,42 @@ public ConnectionMode ConnectionMode
...
+        ///     {
+        ///         DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull,
+        ///     }
+        /// };
+        /// 
+        /// CosmosClient client = new CosmosClient("endpoint", "key", clientOptions);
+        /// ]]>
+        /// </code>
+        /// </example>
+        public System.Text.Json.JsonSerializerOptions SystemTextJsonSerializerOptions
```
kirankumarkolli, line 418: Naming: UseSystemTextJsonSerializerWithOptions ?

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -647,6 +647,20 @@ public CosmosClientBuilder WithContentResponseOnWrite(bool contentResponseOnWrit
             this.clientOptions.EnableContentResponseOnWrite = contentResponseOnWrite;
             return this;
         }
+
+        /// <summary>
+        /// Sets the serializer options for <see cref="CosmosSystemTextJsonSerializer"/>. If none provided
```
kirankumarkolli, line 652:
```suggestion
        /// Sets the serializer options for System.Text.Json serializer. If none provided

```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -647,6 +647,20 @@ public CosmosClientBuilder WithContentResponseOnWrite(bool contentResponseOnWrit
...
+
+        /// <summary>
+        /// Sets the serializer options for <see cref="CosmosSystemTextJsonSerializer"/>. If none provided
+        /// the default system text json serializer will not be used for serialization.
+        /// </summary>
+        /// <param name="stjSerializerOptions">An instance of <see cref="System.Text.Json.JsonSerializerOptions"/>
+        /// containing the system text json serializer options.</param>
+        /// <returns>The <see cref="CosmosClientBuilder"/> object</returns>
+        public CosmosClientBuilder WithSystemTextJsonSerializerOptions(
+            System.Text.Json.JsonSerializerOptions stjSerializerOptions)
```
kirankumarkolli, line 659:
```suggestion
            System.Text.Json.JsonSerializerOptions serializerOptions)
```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: If the check is added, it will break the STJ serializer setter flow in the [client context core here](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/users/kundadebdatta/make_stj_serializer_public/Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs#L131-L135).

Ideally, if the `SystemTextJsonSerializerOptions` is provided, then it will override any serializer, the user provided. This is the reason I originally had the validation in the `ClientContextCore`.

Ref: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4589#discussion_r1689027514

kirankumarkolli: That's still implementation details right? As contract its dis-joint set.

We can move the serializer as internal property which will be set by both to work-around?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Pilchie:
```suggestion
        /// Note that if this option is provided, then the SDK will use the System.Text.Json as the default serializer and set

```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -392,6 +393,42 @@ public ConnectionMode ConnectionMode
...
         /// <seealso cref="TransactionalBatchItemRequestOptions.EnableContentResponseOnWrite"/>
         public bool? EnableContentResponseOnWrite { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="System.Text.Json.JsonSerializerOptions"/> for the default STJ serializer <see cref="CosmosSystemTextJsonSerializer"/>.
+        /// Note that if this option is provided, then the SDK will use the <see cref="CosmosSystemTextJsonSerializer"/> as the default serializer and set
+        /// the serializer options as the constructor args.
+        /// </summary>
+        /// <example>
+        /// An example on how to configure the STJ serialization option to ignore null values
```
Pilchie, line 403:
```suggestion
        /// An example on how to configure the System.Text.Json serialization option to ignore null values
```

We should probably just avoid saying STJ anywhere in the public surface area.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -392,6 +393,42 @@ public ConnectionMode ConnectionMode
...
+        /// <code language="c#">
+        /// <![CDATA[
+        /// CosmosClientOptions clientOptions = new CosmosClientOptions()
+        /// {
+        ///     SystemTextJsonSerializerOptions = new System.Text.Json.JsonSerializerOptions()
```
Pilchie, line 408: Maybe mention that just using `JsonSerializerOptions.Default` works if people just want to use STJ with no extra options.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Pilchie: I like that name.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Pilchie: Looks like it to me.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -647,6 +647,20 @@ public CosmosClientBuilder WithContentResponseOnWrite(bool contentResponseOnWrit
             this.clientOptions.EnableContentResponseOnWrite = contentResponseOnWrite;
             return this;
         }
+
+        /// <summary>
+        /// Sets the serializer options for <see cref="CosmosSystemTextJsonSerializer"/>. If none provided
+        /// the default system text json serializer will not be used for serialization.
```
Pilchie, line 653:
```suggestion
        /// Configures the <see cref="CosmosClientBuilder"/> to use System.Text.Json for serialization. Use <see cref="System.Text.Json.JsonSerializerOptions.Default" /> to use System.Text.Json with a default configuration. If no options are specified, Newtonsoft.Json will be used for serialization instead.
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
```diff
@@ -64,10 +63,10 @@ public async static Task Initialize(TestContext textContext)
             testDb = await client.CreateDatabaseAsync(dbName);
             testContainer = testDb.CreateContainerAsync(new ContainerProperties(id: Guid.NewGuid().ToString(), partitionKeyPath: "/Pk")).Result;
 
-            stjCosmosSerializer = new CosmosSystemTextJsonSerializer(new JsonSerializerOptions());
-
             stjClient = TestCommon.CreateCosmosClient((cosmosClientBuilder)
-                => cosmosClientBuilder.WithCustomSerializer(stjCosmosSerializer));
+                => cosmosClientBuilder.WithSystemTextJsonSerializerOptions(
+                    new JsonSerializerOptions()),
```
Pilchie, line 68:
```suggestion
                    JsonSerializerOptions.Default),
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
```diff
@@ -57,7 +57,10 @@ public async static Task Initialize(TestContext textContext)
             TestDb = await CosmosClient.CreateDatabaseAsync(dbName);
 
             CosmosDefaultSTJClient = TestCommon.CreateCosmosClient((cosmosClientBuilder)
-                => cosmosClientBuilder.WithCustomSerializer(new CosmosSystemTextJsonSerializer(new JsonSerializerOptions())));
+                => cosmosClientBuilder
+                    .WithSystemTextJsonSerializerOptions(
+                        new JsonSerializerOptions()),
```
Pilchie, line 62:
```suggestion
                        JsonSerializerOptions.Default),
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
```diff
@@ -500,7 +521,41 @@ public void ThrowOnCustomSerializerWithSerializerOptions()
...
+                SerializerOptions = new CosmosSerializationOptions(),
+                SystemTextJsonSerializerOptions = new System.Text.Json.JsonSerializerOptions(),
+            };
+
+            CosmosClient client = new(
+                "https://fake-account.documents.azure.com:443/",
+                Convert.ToBase64String(Encoding.UTF8.GetBytes(Guid.NewGuid().ToString())),
+                options
+            );
+        }
```
Pilchie, line 558: Need some tests in the other orders, right?

Pilchie (commented):
Looks pretty good.  A few suggestions, and agree with the name proposal in @kirankumarkolli's review.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
jcocchi: I like this suggestion, can we give one example of using the default options and another example of customizations like you have now?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
jcocchi: I like it too- it's long but very clear on what serializer will be used as well as the options

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
kundadebdatta: Updated the tests.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqAggregateCustomSerializationBaseline.cs
kundadebdatta: Our `System.Text.Json` package version referenced in the project is `4.6.0`, which is introduced by the `Azure.Core` library version `1.19.0`. It appears that the `Default` option is not available in that version. 

Our [public documentation](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.default?view=netstandard-2.0#system-text-json-jsonserializeroptions-default) for .netstandard 2.0 has the Default mentioned, however it seems it's coming from [System.Text.Json v9.0.0](https://www.nuget.org/packages/System.Text.Json/9.0.0-preview.1.24080.9).

When we finally update the `Azure.Core` library version to the latest ([1.41.0](https://www.nuget.org/packages/Azure.Core/1.41.0)), that has a dependency to the `System.Text.Json` version `4.7.2`, we can check if this option can be explored.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
kundadebdatta: Updated the comment.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Our `System.Text.Json` package version referenced in the project is `4.6.0`, which is introduced by the `Azure.Core` library version `1.19.0`. It appears that the `Default` option is not available in that version. 

Our [public documentation](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.default?view=netstandard-2.0#system-text-json-jsonserializeroptions-default) for .netstandard 2.0 has the Default mentioned, however it seems it's coming from [System.Text.Json v9.0.0](https://www.nuget.org/packages/System.Text.Json/9.0.0-preview.1.24080.9).

When we finally update the `Azure.Core` library version to the latest ([1.41.0](https://www.nuget.org/packages/Azure.Core/1.41.0)), that has a dependency to the `System.Text.Json` version `4.7.2`, we can check if this option can be explored.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Updated.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/LinqTranslationWithCustomSerializerBaseline.cs
Pilchie: Ugh.  We really need to update that sometime.

kundadebdatta: Yes correct. I will try to find out the earliest version that introduced the `Default` option, and act on it accordingly. I just created a task to update the system text json package, as well as the azure core package to the latest version.

Pilchie (dismissed):

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: It's not. I am not sure why it appeared as a change.

Pilchie (approved):

kirankumarkolli (approved):

kirankumarkolli (approved):

Pilchie:
Hooray! 🎆 Looking forward to deleting the copies of that code that I've been littering in my projects ;)

CSharpFiasco:
Any hope this will make 3.43.0? I very much appreciate this PR by the way, I'm looking forward to it!

PR #4598, opened 2024-07-22
Hedging: Adds Read hedging PREVIEW contracts

NaluTripician:
# Pull Request Template

## Description

Adds PREVIEW contracts for request hedging

## Parallel Hedging APIs + Samples

When Building a new `CosmosClient` there will be an option to include Parallel hedging in that client.

```csharp
CosmosClient client = new CosmosClientBuilder("connection string")
    .WithApplicationPreferredRegions(
        new List<string> { "East US", "Central US", "West US" } )
    .WithAvailabilityStrategy(
        AvailabilityStrategy.CrossRegionHedgingStrategy(
        threshold: TimeSpan.FromMilliseconds(500),
        thresholdStep: TimeSpan.FromMilliseconds(100)
     ))
    .Build();
```

or

```csharp
CosmosClientOptions options = new CosmosClientOptions()
{
    AvailabilityStrategy
     = AvailabilityStrategy.CrossRegionHedgingStrategy(
        threshold: TimeSpan.FromMilliseconds(500),
        thresholdStep: TimeSpan.FromMilliseconds(100)
     )
      ApplicationPreferredRegions = new List<string>() { "East US", "West US", "Central US"},
};

    accountEndpoint: "account endpoint",
    authKeyOrResourceToken: "auth key or resource token",
    clientOptions: options);
```

The example above will create a `CosmosClient` instance with AvailabilityStrategy enabled with at 500ms threhshold. This means that if a request takes longer than 500ms the SDK will send a new request to the backend in order of the Preferred Regions List. If the `ApplicationRegion` or `ApplicationPreferredRegions` list is not set, then an AvailabilityStrategy will not be able to be set. If still no response comes back from the first hedge or the primary request after the step time, another parallel request will be made to the next region.  The SDK will then return the first response that comes back from the backend. The threshold parameter is a required parameter can can be set to any value greater than 0. There is also an option to the `AvailabilityStrategy` at request level and override the client level `AvailabilityStrategy` by setting the `AvailabilityStrategy` on the `RequestOptions` object.

> Note: `ApplicationRegion`  or `ApplicationPreferredRegions` MUST be set to use Hedging

Override `AvailabilityStrategy`:

```csharp
//Send one request out with a more aggressive threshold
ItemRequestOptions requestOptions = new ItemRequestOptions()
{
    AvailabilityStrategyOptions =AvailabilityStrategy.CrossRegionHedgingStrategy(
        threshold: TimeSpan.FromMilliseconds(100),
        thresholdStep: TimeSpan.FromMilliseconds(50)
     ))
};
```

Hedging can be enabled for all read requests: ReadItem, Queries (single and cross partition), ReadMany, and ChangeFeed. It is not enabled for write requests.

### Diagnostics

In the diagnostics data there are two new areas of note `Response Region` and `Hedge Context` that will appear when using this feature. `Response Region` shows the region that the request is ultimately served out of. `Hedge Context` shows all the regions requests were sent to. 

### Design

The SDK will send the first request to the primary region. If there is no response from the backend before the threshold time, then the SDK will begin sending hedged requests to the regions in order of the `ApplicationPreferredRegions` list. After the first hedged request is sent out, the hedged requests will continue to be fired off one by one after waiting the time specified in the threshold step. Once a response is received from one of the requests, the availability strategy will check to see if the result is considered final. If the result is final, then it is returned. If not, the SDK will skip the remaining threshold/threshold step time and send out the next hedged request. If all hedged requests are sent out and no final response is received, the SDK will return the last response it received. The AvaiabilityStrategy operates on the RequestInvokerHandler level meaning that each hedged request will go through its own [handler pipeline](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/docs/SdkDesign.md#handler-pipeline), including the ClientRetryPolicy. This means that the hedged requests will be retried independently of each other. Note that the hedged requests are restricted to the region they are sent out in so no cross region retries will be made, only local retries. The primary request will be retried as normal.

#### Status Codes SDK Will Consider Final

| Status Code | Description |
| --- | --- |
| 1xx | 1xx Status Codes are considered Final |
| 2xx | 2xx Status Codes are considered Final |
| 3XX | 3xx Status Codes are considered Final |
| 400 | Bad Request |
| 401 | Unauthorized |
| 404/0 | Not Found, 404/0 responses are final results as the document was not yet available after enforcing the consistency model |
| 409 | Conflict |
| 405 | Method Not Allowed |
| 412 | Precondition Failed |
| 413 | Request Entity Too Large |

All other status codes are treated as possible transient errors and will be retried with hedging.

#### Example Flow For Cross Region Hedging With 3 Regions

```mermaid
graph TD
    A[RequestMessage] <--> B[RequestInvokerHandler]
    B <--> C[CrossRegionHedgingStrategy]
    C --> E(PrimaryRequest)
    E --> F{time spent < threshold}

    F -- No --> I
    F -- Yes --> G[[Wait for response]]
    G -- Response --> H{Is Response Final}
    H -- Yes --> C
    H -- No --> I(Hedge Request 1)
    
    I --> J{time spent < threshold step}

    J -- No --> K(Hedge Request 2)
    J -- Yes --> M[[Wait for response]]
    M -- Response --> N{Is Response Final}
    N -- Yes --> C
    N -- No --> K

    K --> O[[Wait for response]]
    O -- Response --> P{Is Response Final}
    P -- Yes --> C
    P -- No, But this is the final hedge request --> C

    
```


Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -11,7 +11,12 @@ namespace Microsoft.Azure.Cosmos
     /// <summary>
     /// Types of availability strategies supported
     /// </summary>
-    internal abstract class AvailabilityStrategy
```
kirankumarkolli, line 14: ClientOptions also need to have this contract public right?

kirankumarkolli: And builder

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -27,6 +32,10 @@ internal abstract class AvailabilityStrategy
...
             CancellationToken cancellationToken);
 
-        internal abstract bool Enabled();
+        /// <summary>
+        /// Checks to see if the strategy is enabled
+        /// </summary>
+        /// <returns>a bool representing if the strategy is enabled</returns>
+        public abstract bool Enabled();
     }
 }
```
kundadebdatta, line 41: Can we remove the carriage return ?

kundadebdatta (approved):
LGTM.

kirankumarkolli:
Update PR description with 
- API contracts.
- Sample usage
- Any troubleshooting guides/references

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -27,6 +32,10 @@ internal abstract class AvailabilityStrategy
             RequestMessage requestMessage,
             CancellationToken cancellationToken);
 
-        internal abstract bool Enabled();
+        /// <summary>
+        /// Checks to see if the strategy is enabled
+        /// </summary>
+        /// <returns>a bool representing if the strategy is enabled</returns>
+        public abstract bool Enabled();
```
ealsur, line 39: methods should remain internal. customers won't use them, right?

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
```diff
@@ -59,7 +64,11 @@ internal class CrossRegionParallelHedgingAvailabilityStrategy : AvailabilityStra
             this.ThresholdStep = thresholdStep ?? TimeSpan.FromMilliseconds(-1);
         }
 
-        internal override bool Enabled()
+        /// <summary>
+        /// Checks to see if the strategy is enabled
+        /// </summary>
+        /// <returns>a bool representing if the strategy is enabled</returns>
+        public override bool Enabled()
```
ealsur, line 71: internal. The other method tood.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/DisabledAvailabilityStrategy.cs
```diff
@@ -10,9 +10,18 @@ namespace Microsoft.Azure.Cosmos
...
+    internal
+#endif
+    class DisabledAvailabilityStrategy : AvailabilityStrategy
     {
-        internal override bool Enabled()
+        /// <summary>
+        /// Checks to see if the strategy is enabled
+        /// </summary>
+        /// <returns>a bool representing if the strategy is enabled, will always be false</returns>
+        public override bool Enabled()
```
ealsur, line 24: Same. Methods private

ealsur (changes requested):
Please check accessors on methods

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionParallelHedgingAvailabilityStrategy.cs
```diff
@@ -59,6 +64,10 @@ internal class CrossRegionParallelHedgingAvailabilityStrategy : AvailabilityStra
             this.ThresholdStep = thresholdStep ?? TimeSpan.FromMilliseconds(-1);
         }
 
+        /// <summary>
+        /// Checks to see if the strategy is enabled
+        /// </summary>
+        /// <returns>a bool representing if the strategy is enabled</returns>
```
ealsur, line 70: nit: Not a blocker, just for the future.

When you inherit and the base class already has a comment for the method, just add inheritdoc

Like:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/c2043ae85c2cb11c1c2c7a46f88c0473f73c43a2/Microsoft.Azure.Cosmos/src/Json/JsonReader.cs#L105

ealsur (approved):

kundadebdatta (dismissed):

kirankumarkolli:
> CosmosClient client = new CosmosClientBuilder("connection string")
    .WithAvailabilityStrategy(
        new CrossRegionParallelHedgingAvailabilityStrategy(
        threshold: TimeSpan.FromMilliseconds(500),
        thresholdStep: TimeSpan.FromMilliseconds(100)
     ))
    .Build();

With-out preferedRegions/ApplicationRegion Heading is no-op right?
Can we please call this out in the code doc.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -21,12 +26,16 @@ internal abstract class AvailabilityStrategy
         /// <param name="requestMessage"></param>
         /// <param name="cancellationToken"></param>
         /// <returns>The response from the service after the availability strategy is executed</returns>
-        public abstract Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
+        internal abstract Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
```
kirankumarkolli, line 29: Feels weird contract, that customers can enable but can't implement strategy.

kirankumarkolli: The end result with hedging contract is getting the two user inputs (hedge time, step time) right?
Thoughts on alternatives?

ealsur: From what I understand the customers need to choose from one of 2 strategies (can be more in the future). So the customer does not have a custom implementation.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -19,23 +16,26 @@ namespace Microsoft.Azure.Cosmos
     abstract class AvailabilityStrategy
     {
         /// <summary>
-        /// Execute the availability strategy
+        ///  stuff
```
ealsur, line 19: Please write a meaningful description of what this strategy does

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -19,23 +16,26 @@ namespace Microsoft.Azure.Cosmos
...
-            CancellationToken cancellationToken);
+        /// <returns>something</returns>
+        public static AvailabilityStrategy DisabledAvailabilityStrategy()
+        {
+            return new DisabledAvailabilityStrategy();
+        }
 
         /// <summary>
-        /// Checks to see if the strategy is enabled
+        /// stuff
```
ealsur, line 28: Please write a meaningful description of what this strategy does

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -19,23 +16,26 @@ namespace Microsoft.Azure.Cosmos
...
         /// </summary>
-        /// <returns>a bool representing if the strategy is enabled</returns>
-        internal abstract bool Enabled();
+        /// <param name="threshold"></param>
```
ealsur, line 30: Add parameter documentation

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -19,23 +16,26 @@ namespace Microsoft.Azure.Cosmos
...
-        /// <param name="requestMessage"></param>
-        /// <param name="cancellationToken"></param>
-        /// <returns>The response from the service after the availability strategy is executed</returns>
-        internal abstract Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
-            Func<RequestMessage, CancellationToken, Task<ResponseMessage>> sender,
-            CosmosClient client,
-            RequestMessage requestMessage,
-            CancellationToken cancellationToken);
+        /// <returns>something</returns>
+        public static AvailabilityStrategy DisabledAvailabilityStrategy()
```
ealsur, line 22: Regarding method name, repeating AvailabilityStrategy when the class is already AvailabilityStrategy might be repetitive?

What about
```suggestion
        public static AvailabilityStrategy DisabledStrategy()
```

@kirankumarkolli thoughts?

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -4,29 +4,44 @@
...
+        /// <summary>
+        /// After a request's duration passes a threshold, this strategy will send out
+        /// hedged request to other regions. The first hedge request will be sent after the threshold. 
+        /// After that, the strategy will send out a request every thresholdStep
+        /// until the request is completed or regions are exausted
+        /// </summary>
+        /// <param name="threshold"> how long before SDK begins hedging</param>
+        /// <param name="thresholdStep">Period of time between first hedge and next hedging attempts</param>
+        /// <returns>something</returns>
+        public static AvailabilityStrategy CrossRegionHedgingAvailabilityStrategy(TimeSpan threshold,
```
ealsur, line 36: Thoughts on this name? Should we also remove the AvailabilityStrategy and leave Strategy?

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -4,29 +4,44 @@
...
+            TimeSpan? thresholdStep)
+        {
+            return new CrossRegionHedgingAvailabilityStrategy(threshold, thresholdStep);
+        }
+
+        /// <summary>
+        /// Name of Availability Strategy
+        /// </summary>
+        public abstract string StrategyName { get; }
```
kirankumarkolli, line 45: Name is a public contract, is it needed?

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -4,29 +4,44 @@
...
-        /// <param name="requestMessage"></param>
-        /// <param name="cancellationToken"></param>
-        /// <returns>The response from the service after the availability strategy is executed</returns>
-        public abstract Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
-            Func<RequestMessage, CancellationToken, Task<ResponseMessage>> sender,
-            CosmosClient client,
-            RequestMessage requestMessage,
-            CancellationToken cancellationToken);
+        /// <returns>something</returns>
+        public static AvailabilityStrategy DisabledStrategy()
```
kirankumarkolli, line 22: Is `null` availability strategy is equivalent to Disable?

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -4,29 +4,44 @@
...
     /// <summary>
     /// Types of availability strategies supported
     /// </summary>
-    internal abstract class AvailabilityStrategy
+#if PREVIEW
+    public
+#else
+    internal
+#endif
+    abstract class AvailabilityStrategy
```
kirankumarkolli, line 16: It still enables customer to nothing implement and pass right?

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
ealsur: I think the point of disable is: if you set the hedging one at the client level, you can disable it at the request level

kirankumarkolli: Got it, wondering if we can add in per-request override based on understanding the scenarios?

NaluTripician: Right now, request level strategies override the client level strategies

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
```diff
@@ -76,7 +76,12 @@ public class RequestOptions
         /// reduce latency and increase availability. Currently there is one type of availability strategy, parallel request hedging.
         /// If there is a globally enabled availability strategy, setting one in the request options will override the global one.
         /// </summary>
-        internal AvailabilityStrategy AvailabilityStrategy { get; set; }
+#if PREVIEW
```
kirankumarkolli, line 79: This API surface allows either setting at `clientOptions` or in `RequestOptions`. 
Override is not possible and based on user feedback we will add it alter right?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -694,7 +694,12 @@ public Func<HttpClient> HttpClientFactory
         /// <summary>
         /// Availability Strategy to be used for periods of high latency
         /// </summary>
-        internal AvailabilityStrategy AvailabilityStrategy { get; set; }
+#if PREVIEW
```
kirankumarkolli, line 697: Argument check that AvailabilityStrategy can only be set when ApplicationRegion/PreferredRegions are set?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -694,7 +694,12 @@ public Func<HttpClient> HttpClientFactory
         /// <summary>
         /// Availability Strategy to be used for periods of high latency
         /// </summary>
-        internal AvailabilityStrategy AvailabilityStrategy { get; set; }
+#if PREVIEW
+        public
```
kirankumarkolli, line 698: With single region and availability strategy selected few scenarios for test coverage
- Request to that region failed before hedgeTime

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
NaluTripician: Based on offline discussion, including both.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -694,7 +695,25 @@ public Func<HttpClient> HttpClientFactory
         /// <summary>
         /// Availability Strategy to be used for periods of high latency
```
kirankumarkolli, line 696: Apologies should have mentioned it before, please update code documentation with more context and usage (reference from description)

kirankumarkolli:
PR description builder pattern needs to include ApplicationRegion or PreferredRegions

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -694,7 +695,25 @@ public Func<HttpClient> HttpClientFactory
...
+        public
+#else
+        internal
+#endif
+        AvailabilityStrategy AvailabilityStrategy 
+        {
+            get => this.availabilityStrategy;
+            set
+            {
+                if (this.ApplicationPreferredRegions == null
```
ealsur, line 708: nit: This forces an order (the user should have set these first before the AvailabilityStrategy). If the goal is to check if they are set together, regardless of order, then there is a Method that is used for that:

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/2fbc27835fde0837d4b3efea0f722c5f5c64119e/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs#L886-L890

Just add a new check there (like the PartitionLevelFailover check), that way you are not forcing the user to set an order, just that at the moment when the client is created (when all options should be set) both are set.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -4,29 +4,46 @@
...
+        internal AvailabilityStrategy()
+        {
+        }
 
-        internal abstract bool Enabled();
+        /// <summary>
+        ///  Used on a per request level to disable a client level AvailabilityStrategy
+        /// </summary>
+        /// <returns>something</returns>
+        internal static AvailabilityStrategy DisabledStrategy()
```
ealsur, line 29: If this one is internal, imagine this scenario:

1. User sets CrossRegionHedgingStrategy at the client levle
2. User wants to, for a specific operation, have no hedging

How would they do it? Is the above a valid scenario?

NaluTripician: This could be a valid scenario. @kirankumarkolli thoughts about exposing this API in the preview and not waiting like previously discussed.

kirankumarkolli: We discussed this before. In theory its possible to override, one example is different thresholds which is a valid possibility. Idea is to hear the feedback and we can add it back.

kirankumarkolli (dismissed):

kirankumarkolli (approved):

philipthomas-MSFT (approved):
LGTM. No major qualms.

PR #4618, opened 2024-08-01
ChangeFeed/Processor: Fixes Deserialization of ChangeFeedItem and ChangeFeedMetadata to support System.Text.Json and Newtonsoft.Json

### LINKED ISSUE #4616: CF/P AVAD: Issues with Serialization/Deserialization with JsonSerializerOptions on Custom serializer and UseSystemTextJsonSerializerWithOptions.

**Stakeholder**
- Any teams that is using CFP-AVAD and CF-AVAD in Preview.

**PR**: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4618

**Describe the bug**
Using [System.Text.Json](https://learn.microsoft.com/en-us/dotnet/api/system.text.json) custom serializer with CFP-AVAD, [GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs#L1223), is resulting with deserialization issues for customer when type [ChangeFeedItem](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs#L56) and [ChangeFeedMetadata](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs#L20) properties uses **internal set properties** instead of the **default public parameterless constructor**. 

References [Blocking Issue: System.Text.Json and non-public members](https://github.com/dotnet/runtime/issues/31511), [Consider extending JsonIncludeAttribute and JsonConstructorAttribute support to internal and private members.](https://github.com/dotnet/runtime/issues/87431), and [Use immutable types and properties](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/immutability).

We also lack unit test coverage, so I will be improving the SDK by including those in this PR. There were recent discussions on fixing this, but it needs to happen sooner than later, as we have a stakeholder who is blocked.

[System.Text.Json](https://learn.microsoft.com/en-us/dotnet/api/system.text.json) does not deserialize the same as [Newtonsoft.Json](https://www.newtonsoft.com/json) so some changes have to be made to accommodate both.

Current code state prior to this issue and attach branch and PR is the annotation of type [ChangeFeedItem](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs#L56) and [ChangeFeedMetadata](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs#L20) properties with [JsonProperty](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Serialization_JsonProperty.htm) obviously only works with [Newtonsoft.Json](https://www.newtonsoft.com/json) so names must match instead of using alias property names or I would have to dual annotate with both serializers.

~This was noticeably a problem with _ConflictResolutionTimestamp_, _IsTimeToLiveExpired_, and _PreviousLsn_ with [System.Text.Json](https://learn.microsoft.com/en-us/dotnet/api/system.text.json).~ The response payload pre-deserialization returns _crts_, _timeToLiveExpired_, and _previousImageLSN_ elements respectively as shown below:

```json
[
	{
		"current": {},
		"metadata": {
			"lsn": 17,
			"crts": 1722511591,
			"operationType": "delete",
			"timeToLiveExpired": true,
			"previousImageLSN": 16
		},
		"previous": {
			"id": "1",
			"pk": "1",
			"description": "Testing TTL on CFP.",
			"ttl": 5,
			"_rid": "SnxPAOM2VfMBAAAAAAAAAA==",
			"_self": "dbs/SnxPAA==/colls/SnxPAOM2VfM=/docs/SnxPAOM2VfMBAAAAAAAAAA==/",
			"_etag": "\"00000000-0000-0000-e405-5632b83c01da\"",
			"_attachments": "attachments/",
			"_ts": 1722511453
		}
	}
]
```
~**THIS WILL BE A BREAKING CHANGE** for those that are using the CFP-AVAD and CF-AVAD in Preview. I am also proposing to remove [JsonProperty](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Serialization_JsonProperty.htm) altogether to avoid the conflict between the two serializers.In addition, _ConflictResolutionTimestamp_, now _crts_ has changed from DateTime to long because the value pre-deserialization is a long. We can leave it to the user to convert it using their serializer of choice.~

There is also some restructuring in the the `tests\Microsoft.Azure.Cosmos.EmulatorTests` to separate [GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs#L1223) tests without custom serialization from [GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs#L1223) tests with custom serialization.

NOTE: [GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs#L1223) is still in Preview.

**To Reproduce**
Use a custom serializer with [JsonSerializerOptions](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions?view=net-8.0).

This could be reproduced via `CosmosClientOptions.Serializer` property

```cs
CosmosClientOptions clientOptions = new CosmosClientOptions
{
    Serializer = new Microsoft.Azure.Cosmos.CosmosSystemTextJsonSerializer(new JsonSerializerOptions()
    {
        PropertyNameCaseInsensitive = true,
        Converters = { new JsonStringEnumConverter() }
    })
};
```
or

`CosmosClientOptions.UseSystemTextJsonSerializerWithOptions`

```cs
CosmosClientOptions clientOptions = new CosmosClientOptions
{
    UseSystemTextJsonSerializerWithOptions = new JsonSerializerOptions()
    {
        PropertyNameCaseInsensitive = true,
        Converters = { new JsonStringEnumConverter() }
    }
};
```
There is a new UnixDateTimeConverter for  [System.Text.Json](https://learn.microsoft.com/en-us/dotnet/api/system.text.json).

**Expected behavior**
Deserialization of  [ChangeFeedItem](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs#L56) and [ChangeFeedMetadata](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs#L20) when using [Newtonsoft.Json](https://www.newtonsoft.com/json)  or [System.Text.Json](https://learn.microsoft.com/en-us/dotnet/api/system.text.json).

**Actual behavior**
Deserialzation of  [ChangeFeedItem](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs#L56) and [ChangeFeedMetadata](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/aba2b2ce4c81914426484b870d88a76bb7413bbe/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs#L20) is setting all properties to default values when using [System.Text.Json](https://learn.microsoft.com/en-us/dotnet/api/system.text.json).

**Environment summary**
SDK Version: Discovered by customer with version [3.43.0-preview.0](https://www.nuget.org/packages/Microsoft.Azure.Cosmos/3.43.0-preview.0)

**Additional context**

philipthomas-MSFT:
# Pull Request Template

## Description

Details in #4616 

## Type of change

## Closing issues

To automatically close an issue: closes #4616

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
+++ b/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
@@ -4,8 +4,6 @@
 
 namespace Microsoft.Azure.Cosmos
 {
-    using Newtonsoft.Json;
-
     /// <summary>
     /// The typed response that contains the current, previous, and metadata change feed resource when <see cref="ChangeFeedMode"/> is initialized to <see cref="ChangeFeedMode.AllVersionsAndDeletes"/>.
     /// </summary>
@@ -58,19 +56,16 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// The full fidelity change feed current item.
         /// </summary>
-        [JsonProperty(PropertyName = "current")]
         public T Current { get; set; }
 
         /// <summary>
         /// The full fidelity change feed metadata.
         /// </summary>
-        [JsonProperty(PropertyName = "metadata", NullValueHandling = NullValueHandling.Ignore)]
         public ChangeFeedMetadata Metadata { get; set; }
 
         /// <summary>
         /// For delete operations, previous image is always going to be provided. The previous image on replace operations is not going to be exposed by default and requires account-level or container-level opt-in.
         /// </summary>
-        [JsonProperty(PropertyName = "previous", NullValueHandling = NullValueHandling.Ignore)]
         public T Previous { get; set; }
     }
 }
--- a/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
+++ b/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
@@ -4,11 +4,6 @@
 
 namespace Microsoft.Azure.Cosmos
 {
-    using System;
-    using Microsoft.Azure.Documents;
-    using Newtonsoft.Json;
-    using Newtonsoft.Json.Converters;
-
     /// <summary>
     /// The metadata of a change feed resource with <see cref="ChangeFeedMode"/> is initialized to <see cref="ChangeFeedMode.AllVersionsAndDeletes"/>.
     /// </summary>
@@ -16,39 +11,54 @@ namespace Microsoft.Azure.Cosmos
     public
 #else
     internal
-#endif 
+#endif
         class ChangeFeedMetadata
     {
         /// <summary>
+        /// New instance of meta data for <see cref="ChangeFeedItem{T}"/> created.
+        /// </summary>
+        /// <param name="crts"></param>
+        /// <param name="lsn"></param>
+        /// <param name="operationType"></param>
+        /// <param name="previousImageLSN"></param>
+        /// <param name="timeToLiveExpired"></param>
+        public ChangeFeedMetadata(
+            long crts,
+            long lsn,
+            ChangeFeedOperationType operationType,
+            long previousImageLSN,
+            bool timeToLiveExpired)
+        {
+            this.Crts = crts;
+            this.Lsn = lsn;
+            this.OperationType = operationType;
+            this.PreviousImageLSN = previousImageLSN;
+            this.TimeToLiveExpired = timeToLiveExpired;
+        }
+
+        /// <summary>
         /// The conflict resolution timestamp.
         /// </summary>
-        [JsonProperty(PropertyName = "crts", NullValueHandling = NullValueHandling.Ignore)]
-        [JsonConverter(typeof(UnixDateTimeConverter))]
-        public DateTime ConflictResolutionTimestamp { get; internal set; }
+        public long Crts { get; }
 
         /// <summary>
         /// The current logical sequence number.
         /// </summary>
-        [JsonProperty(PropertyName = "lsn", NullValueHandling = NullValueHandling.Ignore)]
-        public long Lsn { get; internal set;  }
+        public long Lsn { get; }
 
         /// <summary>
         /// The change feed operation type.
         /// </summary>
-        [JsonProperty(PropertyName = "operationType")]
-        [JsonConverter(typeof(StringEnumConverter))]
-        public ChangeFeedOperationType OperationType { get; internal set; }
+        public ChangeFeedOperationType OperationType { get; }
 
         /// <summary>
         /// The previous logical sequence number.
         /// </summary>
-        [JsonProperty(PropertyName = "previousImageLSN", NullValueHandling = NullValueHandling.Ignore)]
-        public long PreviousLsn { get; internal set; }
+        public long PreviousImageLSN { get; }
 
         /// <summary>
         /// Used to distinquish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
         /// </summary>
-        [JsonProperty(PropertyName = "timeToLiveExpired", NullValueHandling= NullValueHandling.Ignore)]
-        public bool IsTimeToLiveExpired { get; internal set; }
+        public bool TimeToLiveExpired { get; }
     }
 }
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderTests.cs
@@ -2,7 +2,7 @@
 // Copyright (c) Microsoft Corporation.  All rights reserved.
 //------------------------------------------------------------
 
-namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
+namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.CFP.AllVersionsAndDeletes
 {
     using System;
     using System.Collections.Generic;
@@ -10,26 +10,28 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos;
     using Microsoft.Azure.Cosmos.ChangeFeed.Utils;
+    using Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed;
     using Microsoft.Azure.Cosmos.Services.Management.Tests;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using Newtonsoft.Json;
     using Newtonsoft.Json.Linq;
 
     [TestClass]
-    [TestCategory("ChangeFeedProcessor")]
-    public class GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests : BaseChangeFeedClientHelper
+    [TestCategory("ChangeFeedProcessor.AllVersionsAndDeletes")]
+    public class BuilderTests : BaseChangeFeedClientHelper
     {
         [TestInitialize]
         public async Task TestInitialize()
         {
-            await base.ChangeFeedTestInit();
+            await this.ChangeFeedTestInit();
         }
 
         [TestCleanup]
         public async Task Cleanup()
         {
-            await base.TestCleanup();
+            await this.TestCleanup();
         }
 
         [TestMethod]
@@ -47,14 +49,13 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
             ManualResetEvent allDocsProcessed = new ManualResetEvent(false);
 
             ChangeFeedProcessor processor = monitoredContainer
-                .GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes(processorName: "processor", onChangesDelegate: (ChangeFeedProcessorContext context, IReadOnlyCollection<ChangeFeedItem<dynamic>> docs, CancellationToken token) =>
+                .GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes(processorName: "processor", onChangesDelegate: (ChangeFeedProcessorContext context, IReadOnlyCollection<ChangeFeedItem<ToDoActivity>> docs, CancellationToken token) =>
                 {
                     // NOTE(philipthomas-MSFT): Please allow these Logger.LogLine because TTL on items will purge at random times so I am using this to test when ran locally using emulator.
 
                     Logger.LogLine($"@ {DateTime.Now}, {nameof(stopwatch)} -> CFP AVAD took '{stopwatch.ElapsedMilliseconds}' to read document CRUD in feed.");
-                    Logger.LogLine($"@ {DateTime.Now}, {nameof(docs)} -> {JsonConvert.SerializeObject(docs)}");
 
-                    foreach (ChangeFeedItem<dynamic> change in docs)
+                    foreach (ChangeFeedItem<ToDoActivity> change in docs)
                     {
                         if (change.Metadata.OperationType == ChangeFeedOperationType.Create)
                         {
@@ -62,12 +63,12 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
                             Assert.AreEqual(expected: "1", actual: change.Current.id.ToString());
                             Assert.AreEqual(expected: "1", actual: change.Current.pk.ToString());
                             Assert.AreEqual(expected: "Testing TTL on CFP.", actual: change.Current.description.ToString());
-                            Assert.AreEqual(expected: ttlInSeconds, actual: change.Current.ttl.ToObject<int>());
+                            Assert.AreEqual(expected: ttlInSeconds, actual: change.Current.ttl);
 
                             // metadata
-                            Assert.IsTrue(DateTime.TryParse(s: change.Metadata.ConflictResolutionTimestamp.ToString(), out _), message: "Invalid csrt must be a datetime value.");
+                            Assert.IsTrue(DateTime.TryParse(s: change.Metadata.Crts.ToString(), out _), message: "Invalid csrt must be a datetime value.");
                             Assert.IsTrue(change.Metadata.Lsn > 0, message: "Invalid lsn must be a long value.");
-                            Assert.IsFalse(change.Metadata.IsTimeToLiveExpired);
+                            Assert.IsFalse(change.Metadata.TimeToLiveExpired);
 
                             // previous
                             Assert.IsNull(change.Previous);
@@ -78,15 +79,15 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
                             Assert.IsNull(change.Current.id);
 
                             // metadata
-                            Assert.IsTrue(DateTime.TryParse(s: change.Metadata.ConflictResolutionTimestamp.ToString(), out _), message: "Invalid csrt must be a datetime value.");
+                            Assert.IsTrue(DateTime.TryParse(s: change.Metadata.Crts.ToString(), out _), message: "Invalid csrt must be a datetime value.");
                             Assert.IsTrue(change.Metadata.Lsn > 0, message: "Invalid lsn must be a long value.");
-                            Assert.IsTrue(change.Metadata.IsTimeToLiveExpired);
+                            Assert.IsTrue(change.Metadata.TimeToLiveExpired);
 
                             // previous
                             Assert.AreEqual(expected: "1", actual: change.Previous.id.ToString());
                             Assert.AreEqual(expected: "1", actual: change.Previous.pk.ToString());
                             Assert.AreEqual(expected: "Testing TTL on CFP.", actual: change.Previous.description.ToString());
-                            Assert.AreEqual(expected: ttlInSeconds, actual: change.Previous.ttl.ToObject<int>());
+                            Assert.AreEqual(expected: ttlInSeconds, actual: change.Previous.ttl);
 
                             // stop after reading delete since it is the last document in feed.
                             stopwatch.Stop();
@@ -120,8 +121,8 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
 
             try
             {
-                await Task.Delay(GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests.ChangeFeedSetupTime);
-                await monitoredContainer.CreateItemAsync<dynamic>(new { id = "1", pk = "1", description = "Testing TTL on CFP.", ttl = ttlInSeconds }, partitionKey: new PartitionKey("1"));
+                await Task.Delay(ChangeFeedSetupTime);
+                await monitoredContainer.CreateItemAsync<ToDoActivity>(new ToDoActivity  { id = "1", pk = "1", description = "Testing TTL on CFP.", ttl = ttlInSeconds }, partitionKey: new PartitionKey("1"));
 
                 // NOTE(philipthomas-MSFT): Please allow these Logger.LogLine because TTL on items will purge at random times so I am using this to test when ran locally using emulator.
 
@@ -158,6 +159,8 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
                     string pk = default;
                     string description = default;
 
+                    Console.WriteLine(JsonConvert.SerializeObject(docs));
+
                     foreach (ChangeFeedItem<dynamic> change in docs)
                     {
                         if (change.Metadata.OperationType != ChangeFeedOperationType.Delete)
@@ -174,10 +177,10 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
                         }
 
                         ChangeFeedOperationType operationType = change.Metadata.OperationType;
-                        long previousLsn = change.Metadata.PreviousLsn;
-                        DateTime m = change.Metadata.ConflictResolutionTimestamp;
+                        long previousLsn = change.Metadata.PreviousImageLSN;
+                        long m = change.Metadata.Crts;
                         long lsn = change.Metadata.Lsn;
-                        bool isTimeToLiveExpired = change.Metadata.IsTimeToLiveExpired;
+                        bool isTimeToLiveExpired = change.Metadata.TimeToLiveExpired;
                     }
 
                     Assert.IsNotNull(context.LeaseToken);
@@ -194,7 +197,7 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
                     Assert.AreEqual(expected: "1", actual: createChange.Current.pk.ToString());
                     Assert.AreEqual(expected: "original test", actual: createChange.Current.description.ToString());
                     Assert.AreEqual(expected: createChange.Metadata.OperationType, actual: ChangeFeedOperationType.Create);
-                    Assert.AreEqual(expected: createChange.Metadata.PreviousLsn, actual: 0);
+                    Assert.AreEqual(expected: createChange.Metadata.PreviousImageLSN, actual: 0);
                     Assert.IsNull(createChange.Previous);
 
                     ChangeFeedItem<dynamic> replaceChange = docs.ElementAt(1);
@@ -203,20 +206,20 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
                     Assert.AreEqual(expected: "1", actual: replaceChange.Current.pk.ToString());
                     Assert.AreEqual(expected: "test after replace", actual: replaceChange.Current.description.ToString());
                     Assert.AreEqual(expected: replaceChange.Metadata.OperationType, actual: ChangeFeedOperationType.Replace);
-                    Assert.AreEqual(expected: createChange.Metadata.Lsn, actual: replaceChange.Metadata.PreviousLsn);
+                    Assert.AreEqual(expected: createChange.Metadata.Lsn, actual: replaceChange.Metadata.PreviousImageLSN);
                     Assert.IsNull(replaceChange.Previous);
 
                     ChangeFeedItem<dynamic> deleteChange = docs.ElementAt(2);
                     Assert.IsNull(deleteChange.Current.id);
                     Assert.AreEqual(expected: deleteChange.Metadata.OperationType, actual: ChangeFeedOperationType.Delete);
-                    Assert.AreEqual(expected: replaceChange.Metadata.Lsn, actual: deleteChange.Metadata.PreviousLsn);
+                    Assert.AreEqual(expected: replaceChange.Metadata.Lsn, actual: deleteChange.Metadata.PreviousImageLSN);
                     Assert.IsNotNull(deleteChange.Previous);
                     Assert.AreEqual(expected: "1", actual: deleteChange.Previous.id.ToString());
                     Assert.AreEqual(expected: "1", actual: deleteChange.Previous.pk.ToString());
                     Assert.AreEqual(expected: "test after replace", actual: deleteChange.Previous.description.ToString());
 
-                    Assert.IsTrue(condition: createChange.Metadata.ConflictResolutionTimestamp < replaceChange.Metadata.ConflictResolutionTimestamp, message: "The create operation must happen before the replace operation.");
-                    Assert.IsTrue(condition: replaceChange.Metadata.ConflictResolutionTimestamp < deleteChange.Metadata.ConflictResolutionTimestamp, message: "The replace operation must happen before the delete operation.");
+                    Assert.IsTrue(condition: createChange.Metadata.Crts < replaceChange.Metadata.Crts, message: "The create operation must happen before the replace operation.");
+                    Assert.IsTrue(condition: replaceChange.Metadata.Crts < deleteChange.Metadata.Crts, message: "The replace operation must happen before the delete operation.");
                     Assert.IsTrue(condition: createChange.Metadata.Lsn < replaceChange.Metadata.Lsn, message: "The create operation must happen before the replace operation.");
                     Assert.IsTrue(condition: createChange.Metadata.Lsn < replaceChange.Metadata.Lsn, message: "The replace operation must happen before the delete operation.");
 
@@ -236,7 +239,7 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
             // 1 second delay between operations to get different timestamps.
 
             await processor.StartAsync();
-            await Task.Delay(BaseChangeFeedClientHelper.ChangeFeedSetupTime);
+            await Task.Delay(ChangeFeedSetupTime);
 
             await monitoredContainer.CreateItemAsync<dynamic>(new { id = "1", pk = "1", description = "original test" }, partitionKey: new PartitionKey("1"));
             await Task.Delay(1000);
@@ -246,7 +249,7 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
 
             await monitoredContainer.DeleteItemAsync<dynamic>(id: "1", partitionKey: new PartitionKey("1"));
 
-            bool isStartOk = allDocsProcessed.WaitOne(10 * BaseChangeFeedClientHelper.ChangeFeedSetupTime);
+            bool isStartOk = allDocsProcessed.WaitOne(10 * ChangeFeedSetupTime);
 
             await processor.StopAsync();
 
@@ -270,16 +273,16 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
             ContainerInternal monitoredContainer = await this.CreateMonitoredContainer(ChangeFeedMode.LatestVersion);
             ManualResetEvent allDocsProcessed = new(false);
 
-            await GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests
-                .BuildChangeFeedProcessorWithLatestVersionAsync(
+            await
+                BuildChangeFeedProcessorWithLatestVersionAsync(
                     monitoredContainer: monitoredContainer,
                     leaseContainer: this.LeaseContainer,
                     allDocsProcessed: allDocsProcessed,
                     withStartFromBeginning: withStartFromBeginning);
 
             ArgumentException exception = await Assert.ThrowsExceptionAsync<ArgumentException>(
-                () => GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests
-                    .BuildChangeFeedProcessorWithAllVersionsAndDeletesAsync(
+                () =>
+                    BuildChangeFeedProcessorWithAllVersionsAndDeletesAsync(
                         monitoredContainer: monitoredContainer,
                         leaseContainer: this.LeaseContainer,
                         allDocsProcessed: allDocsProcessed));
@@ -302,8 +305,8 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
             ContainerInternal monitoredContainer = await this.CreateMonitoredContainer(ChangeFeedMode.LatestVersion);
             ManualResetEvent allDocsProcessed = new(false);
 
-            await GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests
-                .BuildChangeFeedProcessorWithLatestVersionAsync(
+            await
+                BuildChangeFeedProcessorWithLatestVersionAsync(
                     monitoredContainer: monitoredContainer,
                     leaseContainer: this.LeaseContainer,
                     allDocsProcessed: allDocsProcessed,
@@ -311,14 +314,14 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
 
             // Read lease documents, remove the Mode, and update the lease documents, so that it mimics a legacy lease document.
 
-            await GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests
-                .RevertLeaseDocumentsToLegacyWithNoMode(
+            await
+                RevertLeaseDocumentsToLegacyWithNoMode(
                     leaseContainer: this.LeaseContainer,
                     leaseDocumentCount: 2);
 
             ArgumentException exception = await Assert.ThrowsExceptionAsync<ArgumentException>(
-                () => GetChangeFeedProcessorBuilderWithAllVersionsAndDeletesTests
-                    .BuildChangeFeedProcessorWithAllVersionsAndDeletesAsync(
+                () =>
+                    BuildChangeFeedProcessorWithAllVersionsAndDeletesAsync(
                         monitoredContainer: monitoredContainer,
                         leaseContainer: this.LeaseContainer,
                         allDocsProcessed: allDocsProcessed));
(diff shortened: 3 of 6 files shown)


Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -4,51 +4,61 @@
...
+            this.Crts = crts;
+            this.Lsn = lsn;
+            this.OperationType = operationType;
+            this.PreviousImageLSN = previousImageLSN;
+            this.TimeToLiveExpired = timeToLiveExpired;
+        }
+
         /// <summary>
         /// The conflict resolution timestamp.
         /// </summary>
```
kirankumarkolli, line 41: Metadata is internal type and can use any default serializer. 

Using System.Text.Json types will make us future proofed.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -4,51 +4,61 @@
...
-        [JsonProperty(PropertyName = "crts", NullValueHandling = NullValueHandling.Ignore)]
-        [JsonConverter(typeof(UnixDateTimeConverter))]
```
kirankumarkolli, line 26: This change is changing the return type as well.

Also in-consistent with in some other places (ex: DataEncryptionProperties).

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
```diff
@@ -58,19 +56,16 @@ class ChangeFeedItem<T>
         /// <summary>
         /// The full fidelity change feed current item.
         /// </summary>
-        [JsonProperty(PropertyName = "current")]
```
kirankumarkolli, line 61: Do serializer case sensitive settings impact it? 
ex: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/character-casing

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -4,51 +4,61 @@
...
-        public DateTime ConflictResolutionTimestamp { get; internal set; }
```
ealsur, line 27: This is removing this property from the type, is this expected? Is this property not going to be available/used by customers?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderTests.cs
```diff
@@ -47,27 +49,26 @@ public async Task WhenADocumentIsCreatedWithTtlSetThenTheDocumentIsDeletedTestsA
...
-                .GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes(processorName: "processor", onChangesDelegate: (ChangeFeedProcessorContext context, IReadOnlyCollection<ChangeFeedItem<dynamic>> docs, CancellationToken token) =>
+                .GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes(processorName: "processor", onChangesDelegate: (ChangeFeedProcessorContext context, IReadOnlyCollection<ChangeFeedItem<ToDoActivity>> docs, CancellationToken token) =>
                 {
                     // NOTE(philipthomas-MSFT): Please allow these Logger.LogLine because TTL on items will purge at random times so I am using this to test when ran locally using emulator.
 
                     Logger.LogLine($"@ {DateTime.Now}, {nameof(stopwatch)} -> CFP AVAD took '{stopwatch.ElapsedMilliseconds}' to read document CRUD in feed.");
-                    Logger.LogLine($"@ {DateTime.Now}, {nameof(docs)} -> {JsonConvert.SerializeObject(docs)}");
 
-                    foreach (ChangeFeedItem<dynamic> change in docs)
+                    foreach (ChangeFeedItem<ToDoActivity> change in docs)
```
ealsur, line 58: Why do we need to change from `dynamic` to `ToDoActivity`?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
```diff
@@ -0,0 +1,690 @@
...
+    using Microsoft.Azure.Cosmos;
+    using Microsoft.Azure.Cosmos.ChangeFeed.Utils;
+    using Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed;
+    using Microsoft.Azure.Cosmos.Services.Management.Tests;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+
+    [TestClass]
+    [TestCategory("ChangeFeedProcessor.AllVersionsAndDeletes")]
```
ealsur, line 24: Our YAML pipelines are filtering tests by categories, where is this category supposed to fall into?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
```diff
@@ -0,0 +1,690 @@
...
+            ValidateNewtonsoftJsonDeserialization(json);
+
+            static void ValidateSystemTextJsonDeserialization(string json)
+            {
+                ValidateDeserialization(
+                    System.Text.Json.JsonSerializer.Deserialize<List<ChangeFeedItem<ToDoActivity>>>(
+                        json: json,
+                        options: new JsonSerializerOptions()
+                        {
+                            PropertyNameCaseInsensitive = true,
```
ealsur, line 91: These are custom options, does it not work with the default ones? @kundadebdatta added the new System.Text.Json serializer to the SDK, does this one work?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
philipthomas-MSFT: Apologies, was there a question here?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
philipthomas-MSFT: Yes, I explained it on the link in description.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
philipthomas-MSFT: In the test, i am using that very setting and so is the ARM Core team which is why this is a problem. There exact implementation is

```cs
private static readonly JsonSerializerOptions s_canonicalJsonSerializerSettings = new JsonSerializerOptions()
        {
            PropertyNameCaseInsensitive = true,  // as long as ARM is case insensitive, should keep this setting
            MaxDepth = JsonExtensions.JsonSerializationMaxDepth,
            Converters =
            {
               new JsonStringEnumConverter(),
               new Iso8601TimeSpanConverter(),
               new CosmosIso8601DateTimeConverter(),
               // TODO: Decide whether we still need JTOKEN support.
               // new JTokenConverter()
            }
        };
```

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
philipthomas-MSFT: As I explained in the link in description, ConflictResolutionTimestamp needs to be renamed.

https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4616

I am having to rename three of these properties to match the response if I am to strip Newtonsoft's explicit JsonProperty annotations.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderTests.cs
philipthomas-MSFT: Because setting up a type was better for testing the deserialization of both serializer better for me to get an accurate repro of the problem that ARM Core was having.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
philipthomas-MSFT: Understood. I will change this back to ChangeFeedProcessor. I forgot about the YAML pipelines.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
philipthomas-MSFT: For the problem that the customer is having, this is how they set up there serializer. And I am using Deb's WithSystemTextJsonSerializerOptions and JsonSerializerOptions, but for the customer, they have all the additonal settings and converters.

https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4616

```cs
private static readonly JsonSerializerOptions s_canonicalJsonSerializerSettings = new JsonSerializerOptions()
        {
            PropertyNameCaseInsensitive = true,  // as long as ARM is case insensitive, should keep this setting
            MaxDepth = JsonExtensions.JsonSerializationMaxDepth,
            Converters =
            {
               new JsonStringEnumConverter(),
               new Iso8601TimeSpanConverter(),
               new CosmosIso8601DateTimeConverter(),
               // TODO: Decide whether we still need JTOKEN support.
               // new JTokenConverter()
            }
        };
```

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
kirankumarkolli: `ChangeFeedMetadata` is a metadata context but its inlined with payload. 
With current implementation, user serializer is applicable.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
kirankumarkolli: > PropertyNameCaseInsensitive = true,

=> Its NOT case sensitive so it works
Default is false, in those cases it will fail right? 

Currently we only allow serializer at client scope (not request scope). 
In-case application needs CASE SENSITIVE as goal then it will break right?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
ealsur: The problem is that it cannot be not-inlined, it's part of the Change Feed (user data) response? If the metadata would not have been part of the document, it would have been different

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
ealsur: Ahh completely missed the fact that CRTS was ConflictResolutionTimestamp :D

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
ealsur: If it works with the default System.Text.Json serializer (without having to customize it), then great. The point is, users should not need to add PropertyNameCaseInsensitive for it to work.

philipthomas-MSFT: But you miss the problem @ealsur . It is not just our type, this is a wrapper for their types as well, the actual document. There is no way to separate the two since they use the same serialization. ChangeFeedItem[T].

philipthomas-MSFT: Added Support for System.Text.Json. Please review.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
philipthomas-MSFT: Added Support for System.Text.Json. Please review.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
philipthomas-MSFT: Added Support for System.Text.Json. Please review.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -16,39 +17,66 @@ namespace Microsoft.Azure.Cosmos
...
     {
+        /// <summary>
+        /// New instance of meta data for <see cref="ChangeFeedItem{T}"/> created.
+        /// </summary>
+        /// <param name="conflictResolutionTimestamp"></param>
+        /// <param name="lsn"></param>
+        /// <param name="operationType"></param>
+        /// <param name="previousLSN"></param>
+        /// <param name="isTimeToLiveExpired"></param>
+        public ChangeFeedMetadata(
```
kirankumarkolli, line 31: During last PR (about deserialization failing), our closure if to use only properties across for consistency. 
And have only the default constructor right? 

Why is a constructor needed?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
```diff
@@ -59,18 +60,21 @@ class ChangeFeedItem<T>
         /// The full fidelity change feed current item.
         /// </summary>
         [JsonProperty(PropertyName = "current")]
+        [JsonPropertyName("current")]
```
kirankumarkolli, line 63: For better reaadability, how about fully qualify the attributes? 

Names of types in both packages are very similar. (Just for places where both are inlined)

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -16,39 +17,66 @@ namespace Microsoft.Azure.Cosmos
...
+        public ChangeFeedOperationType OperationType { get; }
 
         /// <summary>
         /// The previous logical sequence number.
         /// </summary>
-        [JsonProperty(PropertyName = "previousImageLSN", NullValueHandling = NullValueHandling.Ignore)]
-        public long PreviousLsn { get; internal set; }
+        [Newtonsoft.Json.JsonProperty(PropertyName = "previousImageLSN", NullValueHandling = NullValueHandling.Ignore)]
+        [JsonPropertyName("previousImageLSN")]
+        public long PreviousLSN { get; }
```
kirankumarkolli, line 73: `PrevisouLsn` -> `PreviousLSN`: Is the suffix change intentional?

Java API's went through API review and we can align with conventions in-case.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
philipthomas-MSFT: I would argue that it reduces readability from code clutter. I went down that path frist, but changed it because it was just ugly to look at seemed repetitive.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
philipthomas-MSFT: i have no qualms. changing and retesting and rebuilding contracts.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
philipthomas-MSFT: @kirankumarkolli still waiting on a counter for this.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
kirankumarkolli: Ideal is to future proofing i.e. even when new properties are added (which is possible) contract should not be breaking. (ref: Even breaking APIs on preview SDK's central SDK team is not happy with)

Changing an existing constructor is a breaking change. 

Possible solutions
- Shadow internal types: These types exact match wire contracts and SDK will deserialize to & later convert them to public types (clean, but every iterator on top needs to be proofed/validated)
- Public setter: Possible future CRI's
- TypeConverter: Its definitely more extra work but future proof's it and the unit of work fits into V4 STJ as default.

kirankumarkolli: Hence the suggestion of typeconverters

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
kirankumarkolli: Your call, not a blocker for this PR.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
philipthomas-MSFT: Added Type converter

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs
philipthomas-MSFT: Not an issue since I am using Type converter now.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
```diff
@@ -0,0 +1,80 @@
...
+                    metadata.PreviousLsn = property.Value.GetInt64();
+                }
+            }
+
+            return metadata;
+        }
+
+        public override void Write(Utf8JsonWriter writer, ChangeFeedMetadata value, JsonSerializerOptions options)
+        {
+            if (value == null)
```
kirankumarkolli, line 64: Write: Is this for scenario where CX tries to serialize the object?

philipthomas-MSFT: The customer should never try to write, but I do not like leaving methods with NotImplementedException and implemented JsonConverter<T> forces it.

philipthomas-MSFT: And I actually had a test fail where I needed it.

kirankumarkolli: > The customer should never try to write, but I do not like leaving methods with NotImplementedException and implemented JsonConverter forces it.

Is it not a scenario for customers to serialize the returned object?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -5,50 +5,52 @@
...
-        public long Lsn { get; internal set;  }
+        [JsonProperty(PropertyName = ChangeFeedMetadataFields.Lsn, NullValueHandling = NullValueHandling.Ignore)]
+        public long Lsn { get; internal set; }
 
         /// <summary>
-        /// The change feed operation type.
+        /// The change's feed operation type <see cref="ChangeFeedOperationType"/>.
         /// </summary>
-        [JsonProperty(PropertyName = "operationType")]
-        [JsonConverter(typeof(StringEnumConverter))]
```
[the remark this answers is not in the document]
philipthomas-MSFT: Got carried away but oddly, my tests did not fail so wondering if I even needed it in the first place.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
philipthomas-MSFT: And persisting metadata? No, that is not a scenario. They are only reading it. They can't write it back. Their data, if they want to update that, just goes back through Create, Upsert, Delete, etc.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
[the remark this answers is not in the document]
philipthomas-MSFT: added it back. resolving

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
kirankumarkolli: Some of the types are for consumption only and can't be written back to store. 
But as a developer serializing the type is one common pattern for logging or transferring the state between services (Its in theory possible they can use our type of that).

As public type when SDK exposing it, it needs to be complete.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                return;
+            }
+
+            writer.WriteStartObject();
+
+            writer.WriteNumber(ChangeFeedMetadataFields.ConflictResolutionTimestamp, ((DateTimeOffset)value.ConflictResolutionTimestamp).ToUnixTimeSeconds());
```
ealsur, line 71: The property is a DateTime, not DateTimeOffset, why is the cast needed?

philipthomas-MSFT: @ealsur  Because I need to get ToUnixTimeSeconds which exist on DateTimeOffset

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
philipthomas-MSFT: Sure. Included tests.

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
ealsur: The casting works only because the variable was created as with `.UtcDateTime`, otherwise the cast would be dangerous because the DateTime could have any other timezone that is not Utc.

Just keep in mind that this is not something that can be done always and a more secure way would be:

`(yourDatetime - DateTime.UnixEpoch).TotalSeconds` because there is no assumption on the `yourDateTime` time zone. Not needed in this case, but for future reference.

ealsur (approved):

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
philipthomas-MSFT: As mentioned in Teams, wrote my own functions instead of using DateTimeOffset as it was failing on the gated pipeline so as you mentioned, it was best to use and DateTime.UnixEpoc is not available for us to use in our project. If tests pass we can use this. If not, back to the drawing board.

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/b6fde92a6db6ea78ec77efd4d0d0128749b076cc/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs#L80

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/b6fde92a6db6ea78ec77efd4d0d0128749b076cc/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs#L86

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
```diff
@@ -0,0 +1,92 @@
...
+            writer.WriteNumber(ChangeFeedMetadataFields.Lsn, value.Lsn);
+            writer.WriteString(ChangeFeedMetadataFields.OperationType, value.OperationType.ToString());
+            writer.WriteNumber(ChangeFeedMetadataFields.PreviousImageLSN, value.PreviousLsn);
+
+            writer.WriteEndObject();
+        }
+
+        private static long ToUnixTimeInSecondsFromDateTime(DateTime date)
+        {
+            DateTime unixEpoch = new (1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
```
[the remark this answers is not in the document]
ealsur: Like 

```
// up top on the file
private static DateTime UnixEpoch = new (1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);

// down below
private static long ToUnixTimeInSecondsFromDateTime(DateTime date)
{
    return (long)(date - ChangeFeedMetadataConverter.UnixEpoch).TotalSeconds;
}

private static DateTime ToDateTimeFromUnixTimeInSeconds(long unixTimeInSeconds)
{
    return unixEpoch.AddSeconds(unixTimeInSeconds);
}
```

ealsur (approved):

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -5,50 +5,54 @@
...
-#endif 
+#endif
         class ChangeFeedMetadata
     {
         /// <summary>
-        /// The conflict resolution timestamp.
+        /// The change's conflict resolution timestamp.
         /// </summary>
-        [JsonProperty(PropertyName = "crts", NullValueHandling = NullValueHandling.Ignore)]
+        [JsonProperty(PropertyName = ChangeFeedMetadataFields.ConflictResolutionTimestamp, NullValueHandling = NullValueHandling.Ignore)]
```
kirankumarkolli, line 29: With converter is [JsonProperty] attributes still needed?

kirankumarkolli: Oh its Newtonsoft JsonProperty?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderTests.cs
```diff
@@ -236,7 +239,7 @@ public async Task WhenADocumentIsCreatedThenUpdatedThenDeletedTestsAsync()
             // 1 second delay between operations to get different timestamps.
 
             await processor.StartAsync();
-            await Task.Delay(BaseChangeFeedClientHelper.ChangeFeedSetupTime);
+            await Task.Delay(ChangeFeedSetupTime);
```
kirankumarkolli, line 242: Current convention is to use <TypeName>.<property> for static or const. 
Is this change intentional? (not a functional blocker but better to be consistent with existing patterns)

kirankumarkolli: Applicable for similar changes.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
```diff
@@ -0,0 +1,734 @@
...
+			            ""_rid"": ""HpxDAL+dzLQBAAAAAAAAAA=="",
+			            ""_self"": ""dbs/HpxDAA==/colls/HpxDAL+dzLQ=/docs/HpxDAL+dzLQBAAAAAAAAAA==/"",
+			            ""_etag"": ""\""00000000-0000-0000-e384-28a5abdd01da\"""",
+			            ""_attachments"": ""attachments/"",
+			            ""_ts"": 1722455971
+		            }
+	            }
+            ]";
+
+            ValidateSystemTextJsonDeserialization(json);
```
kirankumarkolli, line 255: For static API's qualify with the type name.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderTests.cs
philipthomas-MSFT: i'll fix

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
philipthomas-MSFT: i'll fix

philipthomas-MSFT: Correction: This is a local method so no need for qualifiers.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
```diff
@@ -0,0 +1,734 @@
...
+        private ContainerInternal MonitoredContainer = null;
+
+        [TestInitialize]
+        public async Task TestInitialize()
+        {
+            CosmosClient cosmosClient = TestCommon.CreateCosmosClient((cosmosClientBuilder) =>
+                cosmosClientBuilder.WithSystemTextJsonSerializerOptions(
+                    new JsonSerializerOptions()
+                    {
+                        PropertyNameCaseInsensitive = true,
```
kirankumarkolli, line 36: How about test coverage with case-sensitive which will cover for in-sensitive with full coverage?

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
kirankumarkolli: There are still few places with `caseInsensitive = true`

kirankumarkolli: There are two sets of tests one with true and one with false.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerPropertyNameCaseInsensitiveFalseTests.cs
```diff
@@ -0,0 +1,722 @@
...
+    using Microsoft.Azure.Cosmos.ChangeFeed.Utils;
+    using Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed;
+    using Microsoft.Azure.Cosmos.Services.Management.Tests;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+
+    [TestClass]
+    [TestCategory("ChangeFeedProcessor")]
+    public partial class BuilderWithCustomSerializerPropertyNameCaseInsensitiveFalseTests : BaseChangeFeedClientHelper
```
kirankumarkolli, line 25: It's a pure copy with setting changes to `false` right? => Code duplication. 
Thoughts on avoiding code duplication?

philipthomas-MSFT: have at it.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
```diff
@@ -0,0 +1,755 @@
...
+               ""_self"": ""dbs/SnxPAA==/colls/SnxPAOM2VfM=/docs/SnxPAOM2VfMBAAAAAAAAAA==/"",
+               ""_etag"": ""\""00000000-0000-0000-e405-5632b83c01da\"""",
+               ""_attachments"": ""attachments/"",
+               ""_ts"": 1722511453
+              }
+             }
+            ]";
+
+            ValidateSystemTextJsonDeserialization(json, propertyNameCaseInsensitive);
+            ValidateNewtonsoftJsonDeserialization(json);
```
kirankumarkolli, line 59: casesensitive also needs to flow into Newtonsoft validations right?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
```diff
@@ -0,0 +1,755 @@
...
+
+            static void ValidateSystemTextJsonDeserialization(string json, bool propertyNameCaseInsensitive)
+            {
+                ValidateDeserialization(System.Text.Json.JsonSerializer.Deserialize<List<ChangeFeedItem<ToDoActivity>>>(
+                    json: json,
+                    options: new JsonSerializerOptions()
+                    {
+                        PropertyNameCaseInsensitive = propertyNameCaseInsensitive,
+                        Converters = { new JsonStringEnumConverter(), }
```
kirankumarkolli, line 139: Is `JsonStringEnumConverter` still needed? 

We are using `ChangeFeedMetadataConverter` right?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CFP/AllVersionsAndDeletes/BuilderWithCustomSerializerTests.cs
philipthomas-MSFT: It was never meant to be tested on NSJ. We changed nothing about NSJ and I just added support for STJ. I added the _PropertyNameCaseInsensitive_ to STJ only because the customer was using it and I wanted to make sure that was tested for them.

kirankumarkolli (approved):

NaluTripician (approved):
LGTM

FabianMeiswinkel (approved):
LGTM

PR #4632, opened 2024-08-06
PatchItemAsync: Adds documentation on behavior of Replace patch operation when replacing properties that do not exist.

### LINKED ISSUE #4613: PatchItemAsync failing with HTTP400 when using CosmosPropertyNamingPolicy.CamelCase instead of [JsonProperty("id")
My documents all have the Id property named `Id` (upper case) as is the convention in C#. To get things to work with CosmosDb, I'm using the `CosmosClientOptions.SerializerOptions.PropertyNamingPolicy` which I'm setting to `CosmosPropertyNamingPolicy.CamelCase` to have the CosmosDb SDK write my `Id` as `id`:

```
var cosmosOptions = new CosmosClientOptions()
{
    AllowBulkExecution = true,
    SerializerOptions = new CosmosSerializationOptions()
};
cosmosOptions.SerializerOptions.PropertyNamingPolicy = CosmosPropertyNamingPolicy.CamelCase;
var cosmosClient = new CosmosClient(endpoint, key, cosmosOptions);
```

This has worked fine for me so far - inserting, upserting, deleting, searching all work just fine. It's on Patch where I ran into an issue. Here's a sample to illustrate:

```
using Microsoft.Azure.Cosmos;

namespace ConsoleApp1;

internal class CosmosDbTest
{
    string endpoint = "endpoint_url";
    string key = "key";

    internal async Task RunTest()
    {
        try
        {
            var cosmosOptions = new CosmosClientOptions()
            {
                AllowBulkExecution = true,
                SerializerOptions = new CosmosSerializationOptions()
            };
            cosmosOptions.SerializerOptions.PropertyNamingPolicy = CosmosPropertyNamingPolicy.CamelCase;
            var cosmosClient = new CosmosClient(endpoint, key, cosmosOptions);
            var createDb = await cosmosClient.CreateDatabaseIfNotExistsAsync("TestDatabase");
            var db = createDb.Database;
            var createContainer = await db.CreateContainerIfNotExistsAsync(new ContainerProperties("TestContainer", "/id"));
            var container = createContainer.Container;
            var testItem = new PhoneBookCategory
            {
                Name = "test object", Id = Guid.NewGuid().ToString(), SubProp = new TestObject { StringProp = "hello world", IntProp = 10 }
            };
            var createRes = await container.CreateItemAsync(testItem, new PartitionKey(testItem.Id));
            if (createRes.StatusCode == System.Net.HttpStatusCode.Created)
            {
                testItem.Name = "renamed by upsert";
                testItem.SubProp.IntProp = 9;
                var upsertRes = await container.UpsertItemAsync(testItem, new PartitionKey(testItem.Id));
                if (upsertRes.StatusCode == System.Net.HttpStatusCode.OK)
                    Console.WriteLine($"Successfully upserted item {upsertRes.Resource.Id}");
                else
                    Console.WriteLine($"Upsert failed - status: {upsertRes.StatusCode}");

                List<PatchOperation> patchOperations =
                [
                     PatchOperation.Add($"/{nameof(PhoneBookCategory.Name)}", "updated name"),
                     PatchOperation.Replace($"/{nameof(PhoneBookCategory.SubProp)}/{nameof(TestObject.IntProp)}", 5)
                ];
                var patchRes = await container.PatchItemAsync<PhoneBookCategory>(testItem.Id, new PartitionKey(testItem.Id), patchOperations);
                Console.WriteLine($"Item modified with id {patchRes.Resource.Id}");
                var deleteRes = await container.DeleteItemAsync<PhoneBookCategory>(testItem.Id, new PartitionKey(testItem.Id));
                if (deleteRes.StatusCode == System.Net.HttpStatusCode.NoContent)
                    Console.WriteLine($"Successfully cleaned up test object");
                else
                    Console.Write($"Unable to clean up test object - status: {deleteRes.StatusCode}");
            }
            else
                Console.WriteLine($"Unable to create test object - status: {createRes.StatusCode}");
        }
        catch (CosmosException ex)
        {
            Console.WriteLine($"Cosmos DB Exception with Status {ex.StatusCode} Message: {ex.Message}");
        }
    }
}

public class PhoneBookCategory
{
    //[JsonProperty(PropertyName = "id")]
    public string Id { get; set; }
    public string Name { get; set; }
    public TestObject SubProp { get; set; }
}
public class TestObject
{
    public string StringProp { get; set; }
    public int? IntProp { get; set; }
}
```

This will cause the HTTP 400 error on this line

`var patchRes = await container.PatchItemAsync<PhoneBookCategory>(testItem.Id, new PartitionKey(testItem.Id), patchOperations);`

The exception message being:

> Response status code does not indicate success: BadRequest (400); Substatus: 0; ActivityId: bfc4ab22-b363-4cad-970b-63900522677c; Reason: ();

Now uncomment the line

`//[JsonProperty(PropertyName = "id")]`

and comment out the line

`cosmosOptions.SerializerOptions.PropertyNamingPolicy = CosmosPropertyNamingPolicy.CamelCase;`

And it'll work. So it seems we have an issue with Patch and `CosmosPropertyNamingPolicy.CamelCase`. I'd expect the original sample to work and patch the document (it works on all other operations in this sample after all - the insert, the upsert, even the delete). You can even leave the property naming override (`[JsonProperty(PropertyName = "id")]`) active, as soon as the `PropertyNamingPolicy`  is set, you get the issue.

Environment: 
Microsoft.Azure.Cosmos 3.42.0
.NET 8.0.7
VS2022 17.10.5

philipthomas-MSFT:
# Pull Request Template

## Description

Adding documentation to Replace patch operation to describe the following scenario:

```txt
Scenario: Replace operation on a non-existent property
   Given a Cosmos container with an item
   And the item is created with an "id", "name", and "description"
   When a Replace PatchOperation is attempted on a non-existent property "/fake/item"
   Then the operation should throw a CosmosException
   And the HTTP status code should be 400 BadRequest
```

## Type of change

## Closing issues

To automatically close an issue: closes #4613


Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
```diff
@@ -733,6 +733,11 @@ public abstract Task<FeedResponse<T>> ReadManyItemsAsync<T>(
         /// The patch operations are atomic and are executed sequentially.
         /// By default, resource body will be returned as part of the response. User can request no content by setting <see cref="ItemRequestOptions.EnableContentResponseOnWrite"/> flag to false.
         /// Note that, this API does not support the usage of <see cref="RequestOptions.IfMatchEtag"/> property at the moment.
+        ///
+        /// <para>The Replace operation is similar to Set in that it updates a field's value, but differs in one crucial way: Replace
```
kirankumarkolli, line 737: Since these are about respective operations, how about move them to PatchOperation enum and inline with respective operations?

kirankumarkolli: May be in both places?

kirankumarkolli (dismissed):

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
philipthomas-MSFT: I see your point. I think it would be better https://github.com/Azure/azure-cosmos-dotnet-v3/blob/c6d907ca0e4230710317ef860422a060a0db9389/Microsoft.Azure.Cosmos/src/Patch/PatchOperation.cs#L85. Not sure we need it at both places, but it would not hurt.

philipthomas-MSFT: moved it to PatchOperation.Replace. I think it looks better there instead of crowding PatchItemAsync. If we ever need to include more documentation for any other operations, it may just be overwhelming for PatchItemAsycn. @kirankumarkolli

kirankumarkolli (approved):

Issue #4553, opened 2024-06-20
Distributed Tracing/Open Telemetry: Feature requests


sourabh1007:
### Query sanitization (https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4664)
  1. Parameterized queries shouldn't be sanitized (because the user self-sanitizes through parameters), and collecting parameter values will be an opt in config
  2. For all other queries, text should only be captured if it's sanitized. Literal values should be replaced by ?
### Batch ( https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4622 )
  1. New attribute to capture batch size
  2. Operation name should be prepended with BATCH for homogenous operations, or be just BATCH for heterogenous operations. If the DB system has a native name that makes sense we can use that (ex. our API names for ReadMany, TransactionalBatch etc.)
### Bulk
  1. Operation name should be prepended with Bulk. : Covered as part of https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4643
  2. Verify bulk user experience. : Done in Scrum Call
  
### Span name ( https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4553#issuecomment-2264312696)
 1. ~~Change span name to `<operation name> <target>` where target is the collection name if available~~
 2. ~~This removes the `<namespace>` or account name in our case from the span name~~

### Operation Name `db.operation.name` (https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4643)

Batch/bulk operations:
 
- `execute_batch`
- `execute_bulk` 
- `batch_< operation name >`
- `bulk_< operation name >` 
 
Change feed operations:
 
- `query_change_feed`
 
Conflicts operations:
 
- `delete_conflict`
- `query_conflicts`
- `read_all_conflicts`
- `read_conflict`
 
Container operations:
 
- `create_container`
- `create_container_if_not_exists`
- `delete_container`
- `query_containers`
- `read_all_containers`
- `read_container`
- `replace_container`
 
Database operations:
 
- `create_database`
- `create_database_if_not_exists`
- `delete_database`
- `query_databases`
- `read_all_databases`
 
Encryption key operations:
 
- `create_client_encryption_key`
- `query_encryption_keys`
- `read_all_encryption_keys`
- `read_client_encryption_key`
- `replace_client_encryption_key`
 
Item operations:
 
- `create_item`
- `delete_all_items_by_partition_key`
- `delete_item`
- `patch_item`
- `query_items`
[...]

PR #4643, opened 2024-08-16
Open Telemetry: Fixes Operation name to follow Otel convention

sourabh1007:
## Description

ref #4553 

1. **Operation Name List `db.operation.name`**

**Batch/bulk operations:**
execute_batch
execute_bulk
batch_< operation name >
bulk_< operation name > **_(not covered yet)_**

**Change feed operations:**
query_change_feed

**Conflicts operations:**
delete_conflict
query_conflicts
~~read_all_conflicts~~
read_conflict

**Container operations:**
create_container
create_container_if_not_exists
delete_container
~~query_containers~~
~~read_all_containers~~
read_container
replace_container

**Database operations:**
create_database
create_database_if_not_exists
delete_database
~~query_databases~~
~~read_all_databases~~

**Encryption key operations:**
create_client_encryption_key
~~query_encryption_keys~~
~~read_all_encryption_keys~~
read_client_encryption_key
replace_client_encryption_key

**Item operations:**
create_item
delete_all_items_by_partition_key
delete_item
patch_item
query_items
~~read_all_items~~
~~read_all_items_of_logical_partition~~
read_many_items
read_item
replace_item
upsert_item

**Permission operations:**
create_permission
delete_permission
~~query_permissions~~
~~read_all_permissions~~
read_permission
replace_permission
upsert_permission

**Stored procedure operations:**
create_stored_procedure
delete_stored_procedure
execute_stored_procedure
~~query_stored_procedures~~
~~read_all_stored_procedures~~
read_stored_procedure
replace_stored_procedure

**Throughput operations:**
read_throughput
replace_throughput

**Trigger operations:**
create_trigger
delete_trigger
~~query_triggers~~
~~read_all_triggers~~
read_trigger
replace_trigger

**User operations:**
create_user
delete_user
~~query_users~~
~~read_all_users~~
read_user
replace_user
upsert_user

**User-defined function operations:**
create_user_defined_function
delete_user_defined_function
~~query_user_defined_functions~~
read_all_user_defined_functions
read_user_defined_function

2. Fixed operation name generation for `Batch`.
3. Appended `bulk_` as prefix for operations
4. Make sure diagnostic string and opentelemetry operation names are different


Microsoft.Azure.Cosmos.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29123.88
+# Visual Studio Version 17
```
kirankumarkolli, line 3: Undo changes to this this file

sourabh1007: always it creates a PR with "non-draft" status :(

Microsoft.Azure.Cosmos/src/Batch/BatchCore.cs
```diff
@@ -215,8 +224,14 @@ public override Task<TransactionalBatchResponse> ExecuteAsync(
             TransactionalBatchRequestOptions requestOptions,
             CancellationToken cancellationToken = default)
         {
+            string operationName = OpenTelemetryConstants.Operations.ExecuteBatch;
+            if (this.isHomogenousOperations)
```
FabianMeiswinkel, line 228: Is that really necessary? IMO it creates more harm/confusion than benefit - transactional batch should always be executeBatch - it doesn't really matter what operations you do anyway - and in many cases it will be a mx anyway.

Microsoft.Azure.Cosmos/src/Batch/BatchCore.cs
```diff
@@ -251,6 +264,7 @@ public virtual TransactionalBatch PatchItemStream(
             TransactionalBatchPatchItemRequestOptions requestOptions = null)
         {
             this.AddOperation(new ItemBatchOperation(
+                operationName: OpenTelemetryConstants.Operations.PatchItemStream,
```
FabianMeiswinkel, line 267: From Luidmila's proposal I did not get the feeling it would make sense to distinguish xxxItem from xxxItemStream - either we want to normalize operation names between SDKs - then distinguishing how serialization is done doesn't matter and causes just confusion - or we keep is separate for each SDK/language.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
```diff
@@ -113,7 +114,7 @@ private ChangeFeedEstimatorIterator(
         public override Task<FeedResponse<ChangeFeedProcessorState>> ReadNextAsync(CancellationToken cancellationToken = default)
         {
             return this.monitoredContainer.ClientContext.OperationHelperAsync(
-                operationName: "Change Feed Estimator Read Next Async",
+                operationName: OpenTelemetryConstants.Operations.QueryChangeFeedEstimator,
```
FabianMeiswinkel, line 117: dito - no need to know whether it is estimator or not. if that is used - put it in an attribute - like the filter type - but not a dimension of operation name

Microsoft.Azure.Cosmos/src/Resource/FeedIterators/FeedIteratorInlineCore{T}.cs
```diff
@@ -47,7 +48,7 @@ internal FeedIteratorInlineCore(
         public override Task<FeedResponse<T>> ReadNextAsync(CancellationToken cancellationToken = default)
         {
             return this.clientContext.OperationHelperAsync(
-                        operationName: "Typed FeedIterator ReadNextAsync",
+                        operationName: OpenTelemetryConstants.Operations.TypedQueryItems,
```
FabianMeiswinkel, line 51: dito - no need to use operation name to distinguish types from stream

FabianMeiswinkel (changes requested):
Left few comments

Microsoft.Azure.Cosmos/src/Batch/BatchCore.cs
sourabh1007: it was decided to get more information about the batch. @jcocchi @kirankumarkolli what you think about it?

jcocchi: We did call out homogeneous vs. heterogeneous in the proposal, but my understanding of the implementation might have been flawed. This only came up in the database semantic conventions because of how homogenous vs. heterogeneous operations effect SQL traces, and I was trying to equate this to our own APIs.

Transactional Batch - I was thinking represents heterogeneous operations, regardless of what a user is actually doing in the transaction. We don't need to inspect the batch to determine if all operations are actually the same type or not, we should just tell them there was a batch. I'm aligned `executeBatch` covers this case well

Bulk - I thought this was always homogenous and if there was a group of creates and a group of reads, they would get separate network requests for each type for each partition key. After chatting with @sourabh1007 seems like my understanding was wrong. In that case, I don't think we need to call out heterogenous vs homogeneous here either- we can just prepend whatever the operation type was with `bulk_`

Microsoft.Azure.Cosmos/src/Batch/BatchCore.cs
jcocchi: Do other SDKs have stream APIs? Maybe we can add an extra attribute to indicate stream in .NET instead and keep operation name uniform?

Microsoft.Azure.Cosmos/src/Batch/TransactionalBatchInternal.cs
```diff
@@ -52,8 +52,8 @@ protected void AddOperation(ItemBatchOperation itemBatchOperation)
         {
             if (this.isHomogenousOperations && this.operations.Count > 0)
             {
-                this.isHomogenousOperations = this.operations.First().OperationType == itemBatchOperation.OperationType;
-                this.homogenousOperation = this.isHomogenousOperations ? itemBatchOperation.OperationType : null;
+                this.isHomogenousOperations = this.operations.First()?.OperationName == itemBatchOperation.OperationName;
+                this.homogenousOperation = this.isHomogenousOperations ? itemBatchOperation.OperationName : null;
```
jcocchi, line 56: I'm not sure I understand this line, if there's a heterogeneous operation in the transactional batch, would the operation name be null?  `executeBatch` as the operation name is a better alternative in this case

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
jcocchi: I think it's useful to capture estimator vs. change feed reads as part of the operation name because the intent of the request is totally different, agree the type of change feed read can be attributes

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
```diff
@@ -93,7 +94,7 @@ private ChangeFeedPartitionKeyResultSetIteratorCore(
         public override Task<ResponseMessage> ReadNextAsync(CancellationToken cancellationToken = default)
         {
             return this.clientContext.OperationHelperAsync(
-                                operationName: "Change Feed Processor Read Next Async",
+                                operationName: OpenTelemetryConstants.Operations.QueryChangeFeedForPartitionKeyRange,
```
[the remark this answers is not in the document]
jcocchi: Just to clarify- this is tracking requests for all of the existing PK ranges for a given container right? Not requests for actual change feed data?

Querying for partition key ranges feels different than querying change feed for data, it's more of a metadata call. I agree querying change feed for data based on FeedRange vs. based on a PK (or query etc. in the future) can be captured as an attribute though

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryConstants.cs
```diff
@@ -0,0 +1,119 @@
...
+namespace Microsoft.Azure.Cosmos.Telemetry.OpenTelemetry
+{
+    internal class OpenTelemetryConstants
+    {
+        public static class Operations
+        {
+            // Batch/Bulk Operations
+            public const string ExecuteBatch = "execute_batch";
+            public const string ExecuteBatchPrefix = "batch_";
+            public const string ExecuteBulk = "execute_bulk";
```
jcocchi, line 14: Do we have `execute_bulk` in .NET? When would this be used?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryConstants.cs
```diff
@@ -0,0 +1,119 @@
...
+
+namespace Microsoft.Azure.Cosmos.Telemetry.OpenTelemetry
+{
+    internal class OpenTelemetryConstants
+    {
+        public static class Operations
+        {
+            // Batch/Bulk Operations
+            public const string ExecuteBatch = "execute_batch";
+            public const string ExecuteBatchPrefix = "batch_";
```
jcocchi, line 13: Assuming this is for transactional batch, do we need `batch_`? Would there be a separate trace for each item in the transactional batch or just one top level for the whole operation (represented by `execute_batch`?

Microsoft.Azure.Cosmos/src/Batch/BatchCore.cs
sourabh1007: Stream APIs: Removing stream from operation name and later we can add a new attribute to indicate that, it is a stream API as a non breaking change.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
sourabh1007: We are going ahead with, what we have right now.

jcocchi: To clarify, operation names would be:
- query_change_feed
- query_change_feed_for_partition_key_range
- query_change_feed_estimator

Attributes we can add later if we decide to include if the `query_change_feed` was for a particular pk or not

Microsoft.Azure.Cosmos/src/Batch/TransactionalBatchInternal.cs
sourabh1007: it is an implementation of `batch_< operation name >` and `execute_batch` operation name for homogenous and heterogenous operation. but anyway, it doen't matter now. we are going to remove this implementation and I am going to remove this code.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryConstants.cs
sourabh1007: as per our last discussion, there won't be `execute_bulk` operation name. There is always going to be `bulk_< operation name >` operation name for each operational traces.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryConstants.cs
sourabh1007: For transactional batch, there is going to be only one top level trace for whole operation. and batch_size will tell the number of items in the batch.

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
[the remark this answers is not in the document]
FabianMeiswinkel: I really don't think this makes sense - one of the initial decisions was that operation level activities/traces should map to operations a customer triggered. QueryPKRanges is never something a customer directly triggers. How do you want to represent the call hierarchy here - we have transport level traces - that is where these kind of traces can show up. PKRanges queryor change feed is simply a concept that customers can't digest.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
FabianMeiswinkel: what about query_change_feed_for_feed_range then?

FabianMeiswinkel: I still think it explodes operation names too much. If you feel strong about it, go ahead with this in .Net - but in Java I will use attributes for the filter type (and we don't have an explicit estimator API - so, that one is void anyway)

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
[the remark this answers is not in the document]
jcocchi: Maybe I misunderstood what this call is from. Is this the same as calls to `GetFeedRangesAsync()`? This is used as part of the change feed pull model and is the documented way to parallelize reading change feed in the pull model across multiple partitions (ref: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/change-feed-pull-model?tabs=dotnet#use-feedrange-for-parallelization)

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
jcocchi: Doesn't java have `getCurrentState()` and `getEstimatedLag()` ? (ref https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-use-change-feed-estimator?tabs=java#implement-the-change-feed-estimator)

The goal of having these names is to normalize them across SDKs where possible so I think we should land on a mutually agreeable set but I'm definitely open to more discussion!

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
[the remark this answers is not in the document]
FabianMeiswinkel: That would be just query_feed_ranges - nothing with change feed - and at least in Java it also is handled by the cache most of the time (we allow customers to side-step the cache). Anyway, if there would be an operation for COntainer.GetFeedRangesAsync that makes sense - but it should be query_feed_range or read_all_feed_ranges - nothing related to change feed. The changedfeed for pkranges is purely background task that we use to maintain our internal caches for partitions

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs
FabianMeiswinkel: Yes - we have those APIs in the SDK but mapping query_change_feed_estimator to ChangeFeedProcessor.getCurrentState isn't necessarily intuitive

FabianMeiswinkel: So, we can either leave those as query_change_feed (because that is what is happening after all against the backend)
Or if we want to distinguish them I could live with change_feed_processor_context change_feed_processor_estimate or somethign along those lines

FabianMeiswinkel: But query_change_feed  with some attribute that provides more context IMO better draws the big picture. I am pretty flexible here - I just think the _estimator suffix is really confusing in Java because the is no explicit ChangeFeedEstimator class/API like in .Net - and the getEstimate API is kind of deprecated - anything new will only be added to getCurrentState

sourabh1007: As we discussed, rename it to `get_change_feed_processor_estimate`

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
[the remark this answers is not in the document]
sourabh1007: As we discussed, need to remove the activity create for this call, as it is a metadata call.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse{T}.cs
```diff
@@ -37,10 +38,21 @@ internal OpenTelemetryResponse(Response<T> responseMessage)
...
                   correlatedActivityId: OpenTelemetryResponse<T>.GetHeader(responseMessage)?.CorrelatedActivityId,
-                  operationType: responseMessage is QueryResponse ? Documents.OperationType.Query : Documents.OperationType.Invalid)
+                  operationType: GetOperationType(responseMessage),
+                  resourceType: GetResourceType(responseMessage))
         {
         }
 
+        private static OperationType GetOperationType(Response<T> responseMessage)
+        {
+            return responseMessage is QueryResponse ? Documents.OperationType.Query : Documents.OperationType.Invalid;
```
FabianMeiswinkel, line 48: This looks pretty hacky to me - why not have Response<T> return ResourceType and OperationType?

Microsoft.Azure.Cosmos/src/CosmosClient.cs
```diff
@@ -1303,7 +1304,7 @@ internal virtual Task<ResponseMessage> CreateDatabaseStreamAsync(
                         trace,
                         cancellationToken);
                 },
-                openTelemetry: (response) => new OpenTelemetryResponse(response));
+                openTelemetry: new (OpenTelemetryConstants.Operations.CreateDatabaseIfNotExists, (response) => new OpenTelemetryResponse(response)));
```
kirankumarkolli, line 1307: CreateDatabaseStreamAsync

Microsoft.Azure.Cosmos/src/Resource/FeedIterators/FeedIteratorInlineCore{T}.cs
```diff
@@ -53,7 +54,7 @@ public override Task<FeedResponse<T>> ReadNextAsync(CancellationToken cancellati
                         operationType: Documents.OperationType.ReadFeed,
                         requestOptions: null,
                         task: trace => this.feedIteratorInternal.ReadNextAsync(trace, cancellationToken),
-                        openTelemetry: (response) => new OpenTelemetryResponse<T>(responseMessage: response));
+                        openTelemetry: new (OpenTelemetryConstants.Operations.QueryItems, (response) => new OpenTelemetryResponse<T>(responseMessage: response)));
```
kirankumarkolli, line 57: FeedIterator.ReadNextAsync != QueryItems

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryConstants.cs
```diff
@@ -0,0 +1,99 @@
...
+    internal class OpenTelemetryConstants
+    {
+        public static class Operations
+        {
+            // Batch/Bulk Operations
+            public const string ExecuteBatch = "execute_batch";
+            public const string ExecuteBulkPrefix = "bulk_";
+
+            // Change feed operations
+            public const string QueryChangeFeed = "query_change_feed";
```
kirankumarkolli, line 16: Is "_" expected in `change_feed`?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryConstants.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            // Database Operations
+            public const string CreateDatabase = "create_database";
+            public const string CreateDatabaseIfNotExists = "create_database_if_not_exists";
+            public const string DeleteDatabase = "delete_database";
+            public const string ReadDatabase = "read_database";
+
+            // Item Operations
+            public const string CreateItem = "create_item";
+            public const string DeleteAllItemsByPartitionKeyStream = "delete_all_items_by_partition_key_stream";
```
jcocchi, line 44: Are we removing "stream" from the name for consistency across languages?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryConstants.cs
jcocchi: these are two words so "_" makes sense

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryConstants.cs
sourabh1007: yes, I missed it somehow.

kirankumarkolli (approved):

kirankumarkolli (approved):

FabianMeiswinkel (approved):
LGTM

jcocchi (approved):

kirankumarkolli (approved):

PR #4649, opened 2024-08-20
Query: Fixes prefetching to be disabled when MaxConcurrency is less than or equal to one

neildsh:
## Description

Disables parallel prefetch when MaxConcurrency is less than or equal to one, since that would entail sequential fetches from the backend resulting in increased latency for no gain.

## Type of change


kirankumarkolli (approved):

FabianMeiswinkel (approved):
LGTM

kirankumarkolli:
@neildsh can you please take a look at the failing query test?

kirankumarkolli (approved):

PR #4669, opened 2024-09-09
Use account-level regions when preferred regions is not set on client.

jeet1995:


Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -358,81 +378,81 @@ private static bool IsNonRetriableException(Exception exception)
...
-                if (Interlocked.Increment(ref this.disposeCounter) == 1)
-                {
-                    this.CancellationTokenSource?.Cancel();
-                    this.CancellationTokenSource?.Dispose();
-                }
-            }
+            }
+
+            /// <summary>
+            /// Returns an instance of <see cref="IEnumerable{Uri}"/> containing the private and regional service endpoints to iterate over.
```
kundadebdatta, line 384: Doesn't look like there are any changes here. Could you please fix this (and the other places) ? This really helps to focus on the actual diffs.

Issue #4390, opened 2024-04-04
[FEATURE REQ]Immediately trigger cross-regional fail-over on `410/LeaseNotFound(1022)` and `429/SystemResourceUnavailable(3092)`


FabianMeiswinkel:
**Is your feature request related to a problem? Please describe.**
Today when a read region loses the read lease it returns `410/LeaseNotFound` - the client will then retry locally for up-to 30 seconds (catch-all 410 retry like for 410/0 or connectivity related 410) .

**Describe the solution you'd like**

- Instead the `410/LeaseNotFound(1022)` should be mapped immediately to a `503` to allow the cross-regional retry to happen as quickly as possible (it should also result in per partition mark-down when available) - retrying in the next preferred region like usual for reads is acceptable.

- For `429/3092 (Request Throttled/ SystemResourceUnavailable)`, we would only want to short-circuit this to retry cross-region quickly when we can actually retry cross region (not a write in single master) - and for reads only when it is returned from all replica in a region. The latter is extremely unlikely - so, I would probably compromise on a pragmatic approach and simply short-circuit by mapping `429/3092` to a 503 for write operations when multi-master is enabled (and there is >1 write regions).
That way we get the cross-region retry in the scenario where this is most critical. 
 

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Description Added
- [x] Expected solution specified

PR #4677, opened 2024-09-11
`ClientRetryPolicy`: Adds Cross Regional Retry Logic on `429/3092` and `410/1022`

### LINKED ISSUE #4656: ICM-537938324 - Analyze the 410-Lease Not Found and Attempt to Fail Faster
- ICM-537938324 - Analyze the 410-Lease Not Found and Attempt to Fail Faster

kundadebdatta:
# Pull Request Template

## Description

This PR is covering two CRI scenarios in general:

- Gone Exception (Status Code: **410**) with Lease Not Found (Sub Status Code: **1022**).
- Request Throttled (Status Code: **429**) with System Resource Unavailable (Sub Status Code: **3092**).

The behavior changes are described below in detail:

**Background:** In general `410/1022` is used as a signal to indicate that a region is not part of the dynamic quorum - meaning in global strong with dynamic quorum all reads from the secondary region kicked out of the quorum will return `410/1022`.

- Today when a read region loses the read lease it returns `410/LeaseNotFound` - the client will then retry locally for up-to `60` seconds (for global strong) and up-to `30` seconds (for others) (catch-all 410 retry like for 410/0 or connectivity related 410) for any.
Instead the `410/LeaseNotFound(1022)` should be mapped immediately to a `503` to allow the cross-regional retry to happen as quickly as possible. Retrying in the next preferred region like usual for reads is acceptable. This has been taken care in the `Cosmos.Direct` release [`3.36.0`](https://www.nuget.org/packages/Microsoft.Azure.Cosmos.Direct/3.36.0) and this PR is upgrading the `Cosmos.Direct` version to `3.36.0` to reflect the change.

- For `429/3092 (Request Throttled/ SystemResourceUnavailable)`, we would only want to short-circuit this to retry cross-region quickly when we can actually retry cross region (not a write in single master) - and for reads only when it is returned from all replica in a region. The latter is extremely unlikely - so, I would probably compromise on a pragmatic approach and simply short-circuit by mapping `429/3092` to a 503 for write operations when multi-master is enabled (and there is >1 write regions).
That way we get the cross-region retry in the scenario where this is most critical. 

## Type of change

## Closing issues

To automatically close an issue: closes #4656, #4390

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,7 +3,7 @@
 		<ClientOfficialVersion>3.43.0</ClientOfficialVersion>
 		<ClientPreviewVersion>3.44.0</ClientPreviewVersion>
 		<ClientPreviewSuffixVersion>preview.0</ClientPreviewSuffixVersion>
-		<DirectVersion>3.35.0</DirectVersion>
+		<DirectVersion>3.36.0</DirectVersion>
 		<EncryptionOfficialVersion>2.0.4</EncryptionOfficialVersion>
 		<EncryptionPreviewVersion>2.1.0</EncryptionPreviewVersion>
 		<EncryptionPreviewSuffixVersion>preview4</EncryptionPreviewSuffixVersion>
--- a/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
+++ b/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
@@ -48,6 +48,7 @@ namespace Microsoft.Azure.Cosmos
         {
             this.throttlingRetry = new ResourceThrottleRetryPolicy(
                 retryOptions.MaxRetryAttemptsOnThrottledRequests,
+                endpointManager: globalEndpointManager,
                 retryOptions.MaxRetryWaitTimeInSeconds);
 
             this.globalEndpointManager = globalEndpointManager;
--- a/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
+++ b/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
@@ -123,19 +123,13 @@ namespace Microsoft.Azure.Cosmos
 
             ShouldRetryResult throttleRetryResult = await this.throttlingRetry.ShouldRetryAsync(exception, cancellationToken);
 
-            // Received 503 due to client connect timeout or Gateway
+            // Today, the only scenario where we would receive a ServiceUnavailableException from the Throttling Retry Policy
+            // is when we get 410 (Gone) with sub status code 3092 (System Resource Not Available). Note that this is applicable
+            // for write requests targeted to a multiple master account. In such case, the 410/3092 will get converted into 503.
             if (throttleRetryResult.ExceptionToThrow is ServiceUnavailableException)
             {
-                DefaultTrace.TraceWarning("ClientRetryPolicy: ServiceUnavailable. Refresh cache and retry. Failed Location: {0}; ResourceAddress: {1}",
-                    this.documentServiceRequest?.RequestContext?.LocationEndpointToRoute?.ToString() ?? string.Empty,
-                    this.documentServiceRequest?.ResourceAddress ?? string.Empty);
-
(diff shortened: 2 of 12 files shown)


NaluTripician:
Since you are updating the direct package in this PR it might be good to include a summary of the changes in the PR description as well.

Directory.Build.props
```diff
@@ -3,7 +3,7 @@
 		<ClientOfficialVersion>3.43.0</ClientOfficialVersion>
 		<ClientPreviewVersion>3.44.0</ClientPreviewVersion>
 		<ClientPreviewSuffixVersion>preview.0</ClientPreviewSuffixVersion>
-		<DirectVersion>3.35.0</DirectVersion>
```
kirankumarkolli, line 6: 410/1022 is directly coming from direct package right? No changes are needed. 
How about a separate PR just with that scope? 

In-general that's the pattern we are following where direct revision with features it brings.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -120,7 +121,18 @@ public async Task<ShouldRetryResult> ShouldRetryAsync(
                 }
             }
 
-            return await this.throttlingRetry.ShouldRetryAsync(exception, cancellationToken);
+            ShouldRetryResult throttleRetryResult = await this.throttlingRetry.ShouldRetryAsync(exception, cancellationToken);
+
+            // Today, the only scenario where we would receive a ServiceUnavailableException from the Throttling Retry Policy
+            // is when we get 410 (Gone) with sub status code 3092 (System Resource Not Available). Note that this is applicable
+            // for write requests targeted to a multiple master account. In such case, the 410/3092 will get converted into 503.
+            if (throttleRetryResult.ExceptionToThrow is ServiceUnavailableException)
```
kirankumarkolli, line 129: Curious: Alternative is to do 429/3092 check before throttlingPolicy here explicitly, that seems direct and simpler right? 
Also exception creation is expensive and throttle do last longer and overall impact will be higher.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -529,11 +529,12 @@ public bool ShouldRefreshEndpoints(out bool canRefreshInBackground)
 
         public bool CanUseMultipleWriteLocations(DocumentServiceRequest request)
```
kirankumarkolli, line 530: Its used in other scenario like evaluating for sessionToken sending or not. 
SDK region count might be stale and it should not be used to make decision of not passing sesisonToken.

These are two different usecases and should be distinguished.

Directory.Build.props
kundadebdatta: Make sense. However, if we are splitting this PR into two (410/1022 and 429/3092), then the Direct PR needed to go first since this work (329/3092) has dependency on the sub status codes, which is originated from the Direct package.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
kundadebdatta: Ideally the background account refresh (`InitializeAccountPropertiesAndStartBackgroundRefresh()`) should take care of the staleness isn't it ? Any way, will refactor the code to create another method just to safeguard the existing code path.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kundadebdatta: The error combination we are trying to handle here is `429/3092` which itself is a scenario of a throttled response. Therefore, IMO ideal place to handle such cases would be the `ResourceThrottleRetryPolicy`. This will avoid any unnecessary status combination check that should not belong to the `ClientRetryPolocy`.

On creating and throwing the exception part, I will double check if this can be optimized, but from the scenario handling perspective, I feel `ResourceThrottleRetryPolicy` is the ideal place.

@FabianMeiswinkel : Any thoughts on this ?

PR #4709, opened 2024-09-19
Availability: Adds account-level read regions as effective preferred regions when preferred regions is not set on client. 

### LINKED ISSUE #4665: [SDK Cold Start] - Retry Metadata Requests in Other Regions When Preferred Regions is Empty and Hub Region is Offlined
**Acceptance Criteria:**

- Today, when the .NET sdk is in clod start mode and the primary hub region is experiencing an outage, given that the user didn't specify the preferred regions list, the SDK won't do a cross regional retry to get the metadata information.

- This behavior causes the SDK to act as a single point of failure in the above situation, and the goal of this task is to fix this behavior and do cross regional retry attempts, even when the user didn't specify the preferred regions list. That ideally means to populate the preferred regions internally, when the list is provided empty.

jeet1995:
# Pull Request Template

## Problem Statement

As of today, customers who do not configure `ApplicationPreferredRegions` or `ApplicationRegion` are pinned to either the hub region (in multi-write accounts) or the primary region (in single-write accounts). In outage scenarios, availability is scoped to just a single region and is not ideal (both reads/writes in multi-write accounts and reads in single-write multi-region accounts have their availability curbed to just the one region). Setting `ApplicationPreferredRegions` or `ApplicationRegion` as empty is not an opt-out of availability decision from the customer's perspective unless of course on the client a regional endpoint has been set. This PR aims to fix this issue when preferred regions is not set, and a global endpoint is set on the client.

## Approach taken in this PR

The idea is to construct an _effective preferred region list_ and to rely on account-level read and account-level write regions for it. If there is client-perceived unavailability or account-level topology changes - then this _effective preferred regions list_ is reordered accordingly or reflects the account-level regions post a cached `DatabaseAccount` refresh in the SDK.

There are also changes made to the flow which decides when `DatabaseAccount` refresh is triggered. The decision depends on a check whether the SDK has a different effective first preferred read / write region from the first account-level read / write region.

## Closing issues
closes #4665


Microsoft.Azure.Cosmos/src/Resource/Container/ContainerInlineCore.cs
```diff
@@ -442,7 +442,7 @@ public override Task<ResponseMessage> ReadManyItemsStreamAsync(
                 containerName: this.Id,
                 databaseName: this.Database.Id,
                 operationType: Documents.OperationType.Read,
-                requestOptions: null,
+                requestOptions: readManyRequestOptions,
```
kirankumarkolli, line 445: Thank you, this might have impacted the Telemetry collector. 
To get visibility (changelog) how about roll this change as separate PR?

jeet1995: Agreed - will follow up with a separate PR.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -689,6 +730,25 @@ private bool SkipRefresh(bool forceRefresh)
...
+            if (this.connectionPolicy.PreferredLocations != null && this.connectionPolicy.PreferredLocations.Count > 0)
+            {
+                return this.connectionPolicy.PreferredLocations;
+            }
+
+            this.locationCacheDatabaseAccountReadWriteLock.EnterReadLock();
+
+            try
+            {
+                return new Collection<string>(this.locationCache.EffectivePreferredLocations);
```
kirankumarkolli, line 746: Is copy needed? EffectivePreferredLocations is Readonly anyways.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -689,6 +730,25 @@ private bool SkipRefresh(bool forceRefresh)
             TimeSpan timeSinceLastRefresh = DateTime.UtcNow - this.LastBackgroundRefreshUtc;
             return (this.isAccountRefreshInProgress || this.MinTimeBetweenAccountRefresh > timeSinceLastRefresh)
                 && !forceRefresh;
+        }
+
+        public Collection<string> GetEffectivePreferredLocations()
```
kirankumarkolli, line 735: Return type ReadOnlyCollection?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -689,6 +730,25 @@ private bool SkipRefresh(bool forceRefresh)
...
+        }
+
+        public Collection<string> GetEffectivePreferredLocations()
+        {
+            if (this.connectionPolicy.PreferredLocations != null && this.connectionPolicy.PreferredLocations.Count > 0)
+            {
+                return this.connectionPolicy.PreferredLocations;
+            }
+
+            this.locationCacheDatabaseAccountReadWriteLock.EnterReadLock();
```
kirankumarkolli, line 742: Locking: Is it to server for scenarios where the account topology changes? 
With-out locking/synchronization the  in-consistency will not last long right? I.e. probably during update/refresh older version is served which is acceptable?

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
```diff
@@ -206,7 +206,7 @@ private void InitializeClientTelemetry(AccountClientConfiguration clientConfig)
                         connectionMode: this.connectionPolicy.ConnectionMode,
                         authorizationTokenProvider: this.cosmosAuthorization,
                         diagnosticsHelper: DiagnosticsHandlerHelper.GetInstance(),
-                        preferredRegions: this.connectionPolicy.PreferredLocations,
+                        preferredRegions: this.globalEndpointManager.GetEffectivePreferredLocations(),
```
kirankumarkolli, line 209: This service is getting deprecated very soon. 

On new public contract on GEM vs empty preferredRegion contract: Later is a reasonable compromise.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -119,7 +128,8 @@ public static async Task<AccountProperties> GetDatabaseAccountFromAnyLocationsAs
             IList<string>? locations,
             IList<Uri>? accountInitializationCustomEndpoints,
             Func<Uri, Task<AccountProperties>> getDatabaseAccountFn,
-            CancellationToken cancellationToken)
+            CancellationToken cancellationToken,
```
kirankumarkolli, line 131: CancellationToken always last argument.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -325,110 +333,148 @@ public Uri ResolveServiceEndpoint(DocumentServiceRequest request)
...
+                return isReadRequest ? this.ReadEndpoints : this.WriteEndpoints;
+            }
+
+            DatabaseAccountLocationsInfo databaseAccountLocationsInfoSnapshot = this.locationInfo;
+
+            ReadOnlyCollection<string> effectivePreferredLocations = databaseAccountLocationsInfoSnapshot.PreferredLocations;
+
+            if (effectivePreferredLocations == null || effectivePreferredLocations.Count == 0)
+            {
+                effectivePreferredLocations = databaseAccountLocationsInfoSnapshot.EffectivePreferredLocations;
```
kirankumarkolli, line 347: if EffectivePreferredLocation is modeled always as super-set then it will cover both cases and callers don't need this special casing right?

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -325,110 +333,148 @@ public Uri ResolveServiceEndpoint(DocumentServiceRequest request)
...
 
         /// <summary>
         /// Gets applicable endpoints for a request, if there are no applicable endpoints, returns the fallback endpoint
         /// </summary>
         /// <param name="regionNameByEndpoint"></param>
+        /// <param name="effectivePreferredLocations"></param>
         /// <param name="fallbackEndpoint"></param>
         /// <param name="excludeRegions"></param>
         /// <returns>a list of applicable endpoints for a request</returns>
         private ReadOnlyCollection<Uri> GetApplicableEndpoints(
```
kirankumarkolli, line 402: static method?

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -325,110 +333,148 @@ public Uri ResolveServiceEndpoint(DocumentServiceRequest request)
...
         /// <summary>
         /// Gets applicable endpoints for a request, if there are no applicable endpoints, returns the fallback endpoint
         /// </summary>
-        /// <param name="regionNameByEndpoint"></param>
+        /// <param name="availableLocations"></param>
+        /// <param name="effectivePreferredLocations"></param>
         /// <param name="fallbackRegion"></param>
         /// <param name="excludeRegions"></param>
         /// <returns>a list of applicable endpoints for a request</returns>
         private ReadOnlyCollection<string> GetApplicableRegions(
```
kirankumarkolli, line 449: static method?

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -675,6 +730,23 @@ private void UpdateLocationCache(
...
 
+                if (nextLocationInfo.PreferredLocations == null || nextLocationInfo.PreferredLocations.Count == 0)
+                {
+                    if (!nextLocationInfo.AvailableReadLocationByEndpoint.TryGetValue(this.defaultEndpoint, out string regionForDefaultEndpoint))
+                    {
+                        nextLocationInfo.EffectivePreferredLocations = nextLocationInfo.AvailableReadLocations;
+                    }
+                    else
+                    {
+                        List<string> locations = new ()
```
kirankumarkolli, line 741: What's the scenarios for else case?

kirankumarkolli: Is it when the configured endpoint not available?

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -702,17 +774,43 @@ private ReadOnlyCollection<Uri> GetPreferredAvailableEndpoints(ReadOnlyDictionar
...
+                    else
+                    {
+                        foreach (string location in orderedLocations)
+                        {
+                            if (endpointsByLocation.TryGetValue(location, out Uri endpoint))
                             {
-                                endpoints.Add(endpoint);
+                                if (this.defaultEndpoint.Equals(endpoint))
+                                {
+                                    endpoints = new List<Uri>();
```
kirankumarkolli, line 802: Is it for case where regional endpoint is configured?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
jeet1995: No not really - returning a more abstract `IList<string>` in next iteration.

Microsoft.Azure.Cosmos/src/Telemetry/TelemetryToServiceHelper.cs
jeet1995: Sure - have reverted in latest iteration.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
jeet1995: Modified to abstract `IList<string>` to align type with account-resolution calls.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
jeet1995: Fixed in next iteration.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
jeet1995: Fixed in next iteration.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
jeet1995: Fixed in next iteration.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
jeet1995: Correct - idea was to do a equality check for `defaultEndpoint` with a regional endpoint and to short circuit populating account-level regions if the equality check is true. Same idea for Line 740-742

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
jeet1995: Correct

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
jeet1995: How much is long enough along with how many requests might see the inconsistency is the non-deterministic part. We can sync up on this point and reach a consensus on next steps.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
jeet1995: Good point - I've made some refactoring changes to this - `EffectivePreferredLocations` now gets set to `PreferredLocations` first - so special casing in any preferred locations dependent methods is not needed.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
jeet1995: Synced offline - inconsistency is scoped to CPU cycles b/w `AccountProperties` being set and `LocationCache` or of nanosecond range. The lock doesn't solve if the SDK is out of band if the service-side topology change hasn't been reflected in the SDK - so reverting the lock change in `GlobalEndpointManager`.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -257,16 +259,22 @@ public Uri GetHubUri()
...
         }
-
-        public ReadOnlyCollection<string> GetAvailableWriteLocations()
-        {
-            return this.locationInfo.AvailableWriteLocations;
+
+        /// <summary>
+        /// Gets account-level read locations.
+        /// </summary>
+        public ReadOnlyCollection<string> GetAvailableReadLocations()
```
kirankumarkolli, line 267: Feel free to rename the API to reflect it (can be follow-up PR also)

kirankumarkolli (dismissed):

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
jeet1995: Updated the API to `GetAccountLevelAvailableReadLocations` and `GetAccountLevelAvailableWriteLocations`.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -689,6 +699,16 @@ private bool SkipRefresh(bool forceRefresh)
             TimeSpan timeSinceLastRefresh = DateTime.UtcNow - this.LastBackgroundRefreshUtc;
             return (this.isAccountRefreshInProgress || this.MinTimeBetweenAccountRefresh > timeSinceLastRefresh)
                 && !forceRefresh;
+        }
+
+        public IList<string> GetEffectivePreferredLocations()
+        {
+            if (this.connectionPolicy.PreferredLocations != null && this.connectionPolicy.PreferredLocations.Count > 0)
```
philipthomas-MSFT, line 706: nit: consider simplifying with both Null-conditional and Null-coalescing operators. But not going to block.

return this.connectionPolicy.PreferredLocations?.Count > 0
        ? this.connectionPolicy.PreferredLocations
        : this.locationCache.EffectivePreferredLocations;

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -325,110 +333,119 @@ public Uri ResolveServiceEndpoint(DocumentServiceRequest request)
...
                 request.RequestContext.ExcludeRegions);
-        }
-
+        }
+
         public ReadOnlyCollection<string> GetApplicableRegions(IEnumerable<string> excludeRegions, bool isReadRequest)
         {
-            return this.GetApplicableRegions(
-                isReadRequest ? this.locationInfo.AvailableReadLocations : this.locationInfo.AvailableWriteLocations,
-                this.locationInfo.PreferredLocations[0],
```
philipthomas-MSFT, line 348: Are `PreferredLocations` and `effectivePreferredLocations` always guaranteed to have at list 1 item? A non-pure function that is a public/internal method, where the results are not based on it's inputs.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -325,110 +333,119 @@ public Uri ResolveServiceEndpoint(DocumentServiceRequest request)
...
-            {
-                applicableEndpoints.Add(fallbackEndpoint);
-            }
-
-            return new ReadOnlyCollection<Uri>(applicableEndpoints);
-        }
-
+        {
+            List<Uri> applicableEndpoints = new List<Uri>(regionNameByEndpoint.Count);
+            HashSet<string> excludeRegionsHash = excludeRegions == null ? null : new HashSet<string>(excludeRegions);
```
philipthomas-MSFT, line 380: by changing to `HashSet<string> excludeRegionsHash = excludeRegions != null ? new HashSet<string>(excludeRegions) : new HashSet<string>();` you can get rid of the null check and multiple foreach loops.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
jeet1995: Fixed in next iteration.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
jeet1995: Yes, `EffectivePreferredLocations` is guaranteed to be non-empty since it is either populated by account-level read regions or the region corresponding to the default endpoint if it also a regional endpoint. In any case, I have added an `ArgumentException` if it is null / empty prior to index-based access of `EffectivePreferredLocations` (as discussed offline).

PR #4724, opened 2024-09-26
Query: Adds deserialization of HybridSearchQueryInfo to the query plan

neildsh:
## Description

Adds deserialization of HybridSearchQueryInfo to the query plan.

## Type of change


adityasa (dismissed):
:shipit:

adityasa (approved):

Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryFeature.cs
```diff
@@ -30,6 +30,7 @@ enum QueryFeatures : ulong
         DCount = 1 << 10,
         NonStreamingOrderBy = 1 << 11,
         ListAndSetAggregate = 1 << 12,
-        HybridSearch = 1 << 13,
+        CountIf = 1 << 13,
+        HybridSearch = 1 << 14,
```
changanhan, line 34: If we change the enum value of HybridSearch, does it cause compatibility issue and deserialization problem if client SDK version is old?

PR #4738, opened 2024-09-30
[Internal] Query: Adds an aggregate operator for CountIf. Removes debug assert from QueryPartitionProvider

neildsh:
## Description

Adds an aggregate operator for CountIf. Removes debug assert from QueryPartitionProvider that is no longer relevant

## Type of change


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/QueryPartitionProviderTests.cs
```diff
@@ -104,10 +113,22 @@ public void TestHybridSearchQueryInfoDeserialization()
...
-                string actual = JsonConvert.SerializeObject(
-                    queryInfo.HybridSearchQueryInfo,
-                    new JsonSerializerSettings { Formatting = Formatting.None });
+                Assert.IsTrue(queryInfo.QueryInfo != null || queryInfo.HybridSearchQueryInfo != null);
+
+                string actual;
+                if (queryInfo.QueryInfo != null)
+                {
+                    actual = JsonConvert.SerializeObject(
+                        queryInfo.QueryInfo,
```
adityasa, line 122: >queryInfo.QueryInfo [](http://example.com/codeflow?start=24&length=19)

Nit, if can be replaced by ternary on this parameter for brevity.

actual = JsonConvert.SerializeObject(
queryInfo.QueryInfo != null, QueryInfo : HybridSearchQueryInfo,

adityasa (approved):
:shipit:

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/QueryPartitionProviderTests.cs
neildsh: Agreed. Though I prefer to do it this way, since it allows one to easily put a breakpoints while running the test

lemonlion:
Has `CountIf` been documented in the official docs anywhere?  I can't see it here:

https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query/countif
or here
https://learn.microsoft.com/en-us/cosmos-db/query/countif

or anywhere else

Issue #4678, opened 2024-09-12
[Client encryption]: Optimize encryption/decryption flows


juraj-blazek:
**Is your feature request related to a problem? Please describe.**
When using the `Microsoft.Azure.Cosmos.Encryption.Custom` encryption package, there is a big jump in CPU and memory consumption, leading to higher compute cost, higher GC pressure, and lower throughput. This is especially sensitive in the large-scale services.

See sample benchmark results of the `EncryptionProcessor` (PR with the baseline benchmark projects will be linked):

|  Method | DocumentSizeInKb |        Mean |      Error |     StdDev |     Gen0 |     Gen1 |     Gen2 |  Allocated |
|-------- |----------------- |------------:|-----------:|-----------:|---------:|---------:|---------:|-----------:|
| **Encrypt** |                **1** |    **47.90 μs** |   **1.284 μs** |   **1.842 μs** |   **4.5776** |   **1.1597** |        **-** |   **56.22 KB** |
| Decrypt |                1 |    59.67 μs |   1.041 μs |   1.558 μs |   5.2490 |   1.3428 |        - |   64.79 KB |
| **Encrypt** |               **10** |   **154.57 μs** |   **2.728 μs** |   **3.998 μs** |  **20.7520** |   **4.1504** |        **-** |  **255.95 KB** |
| Decrypt |               10 |   220.03 μs |   6.124 μs |   8.585 μs |  29.0527 |   5.8594 |        - |  357.41 KB |
| **Encrypt** |              **100** | **2,761.51 μs** | **213.677 μs** | **319.822 μs** | **218.7500** | **173.8281** | **142.5781** | **2459.89 KB** |
| Decrypt |              100 | 2,445.99 μs | 136.839 μs | 200.577 μs | 347.6563 | 300.7813 | 253.9063 | 3406.33 KB |

**Describe the solution you'd like**
As a first step, we need to bring the allocations down as much as possible. We already started updating the core `Microsoft.Data.Encryption.Cryptography` dependency to add allocation-free APIs. We'll follow with contributions to Cosmos DB SDK to have them utilized.

**Describe alternatives you've considered**
Not using encryption. This is not a viable alternative.

**Additional context**
This is a placeholder issue where we'll link any optimizations PR to.

PR #4739, opened 2024-10-01
[Client encryption]: Fixes Faster validations and string comparisons

JanHyka:
# Pull Request Template

## Description

1. Switch frequent constants string comparison mode to `ordinal`
2. Reduce allocations and validations complexity

## Type of change

## Closing issues

Contributes to #4678


Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
```diff
@@ -55,26 +55,26 @@ public static async Task<Stream> EncryptAsync(
                 return input;
             }
 
-            if (!encryptionOptions.PathsToEncrypt.Distinct().SequenceEqual(encryptionOptions.PathsToEncrypt))
+            if (encryptionOptions.PathsToEncrypt.Distinct().Count() != encryptionOptions.PathsToEncrypt.Count())
```
kirankumarkolli, line 58: SequenceEquals covering values right? where as new change is just the count not values?

JanHyka: We are comparing encryptionOptions.PathsToEncrypt on both sides... that means values are always identical and is useless to compare them. We only need to validate there are no repetitions - ie. Distinct().Count() vs Count()

kirankumarkolli: You are right its the same argument. I mis read it before. Thanks for clarifying it.

kirankumarkolli (approved):

PR #4740, opened 2024-10-01
[Client encryption]: Drop allocations on MDE Encryption path

JanHyka:
# Pull Request Template

## Description

On Encryption path we are switching to MDE API with offsets. This allows reducing of allocations.
To sustain compatibility, enhanced API is backward implemented to obsoleted local copy of encryption algorithm. No performance optimizations on obsoleted code paths.

## Type of change

## Closing issues

Contributes to #4678


Microsoft.Azure.Cosmos.Encryption.Custom/src/AeadAes256CbcHmac256Algorithm.cs
```diff
@@ -137,6 +147,28 @@ public override byte[] EncryptData(byte[] plainText)
...
+        public override int EncryptData(byte[] plainText, int plainTextOffset, int plainTextLength, byte[] output, int outputOffset)
+        {
+            byte[] buffer = this.EncryptData(plainText.AsSpan(plainTextOffset, plainTextLength).ToArray());
+
+            if (buffer.Length > output.Length - outputOffset)
+            {
+                throw new ArgumentOutOfRangeException($"Output buffer is shorter than required {buffer.Length} bytes.");
+            }
+
+            buffer.CopyTo(output, outputOffset);
```
kirankumarkolli, line 168: Curiosity: EncryptData can be refractured to use the incoming output buffer instead of allocating inside and then copy right?

Microsoft.Azure.Cosmos.Encryption.Custom/src/AeadAes256CbcHmac256Algorithm.cs
```diff
@@ -418,5 +450,40 @@ private byte[] PrepareAuthenticationTag(byte[] iv, byte[] cipherText, int offset
...
+            return buffer.Length;
+        }
+
+        public override int GetEncryptByteCount(int plainTextLength)
+        {
+            // Output buffer size = size of VersionByte + Authentication Tag + IV + cipher Text blocks.
+            return sizeof(byte) + AuthenticationTagSizeInBytes + IvSizeInBytes + GetCipherTextLength(plainTextLength);
+        }
+
+        public override int GetDecryptByteCount(int cipherTextLength)
```
kirankumarkolli, line 473: EncryptData has similar logic, thoughts on re-using it?

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosEncryptor.cs
```diff
@@ -45,27 +44,94 @@ public override async Task<byte[]> DecryptAsync(
...
+            string encryptionAlgorithm,
+            CancellationToken cancellationToken = default)
+        {
+            DataEncryptionKey dek = await this.GetEncryptionKeyAsync(dataEncryptionKeyId, encryptionAlgorithm, cancellationToken);
+
             return dek.DecryptData(cipherText);
         }
 
         /// <inheritdoc/>
+        [Obsolete("It is suggested to use GetEncryptionKeyAsync + key.DecryptData to reduce overhead.")]
```
kirankumarkolli, line 64: `Obsolete` immediately feels very aggressive. Can it de-coupled and late bound in future iteration?
The usage of encryption library is low so may be its fine. @kr-santosh thoughts?

Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests/Readme.md
```diff
@@ -2,19 +2,18 @@
...
-| Decrypt |                1 |    44.51 μs |   0.900 μs |   1.291 μs |    43.83 μs |   4.8828 |   1.2207 |        - |    60.1 KB |
-| **Encrypt** |               **10** |   **114.63 μs** |   **3.234 μs** |   **4.841 μs** |   **113.80 μs** |  **16.2354** |   **3.2959** |        **-** |  **199.75 KB** |
-| Decrypt |               10 |   146.56 μs |   1.205 μs |   1.766 μs |   146.08 μs |  24.4141 |   4.6387 |        - |  301.94 KB |
-| **Encrypt** |              **100** | **1,784.59 μs** | **180.928 μs** | **270.805 μs** | **1,743.86 μs** | **158.2031** |  **95.7031** |  **82.0313** | **1890.27 KB** |
-| Decrypt |              100 | 2,772.68 μs | 261.923 μs | 392.035 μs | 2,648.81 μs | 236.3281 | 158.2031 | 142.5781 | 3064.91 KB |
-
+|  Method | DocumentSizeInKb |        Mean |      Error |     StdDev |     Gen0 |     Gen1 |    Gen2 |  Allocated |
+|-------- |----------------- |------------:|-----------:|-----------:|---------:|---------:|--------:|-----------:|
+| **Encrypt** |                **1** |    **37.15 μs** |   **0.683 μs** |   **1.002 μs** |   **3.8452** |   **0.9766** |       **-** |   **47.28 KB** |
```
kirankumarkolli, line 14: For documents < 10KB slight CPU trade-off for memory, is my understanding right? (less future GC)

Microsoft.Azure.Cosmos.Encryption.Custom/src/AeadAes256CbcHmac256Algorithm.cs
JanHyka: We could refactor the class to have this method as a main one and the one without offset as a helper, however I am not sure we want to put effort/optimize algorithm/implementation that was obsoleted years ago. It's very likely same time spent on MDE paths would have way higher impact (both on code efficiency as well as motivation factor to migrate to MDE one if somebody still uses this class at all).  These methods are pretty much duct tape to comply with updated DEK interface.
Up to your preference.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosEncryptor.cs
JanHyka: Yeah, I am not too happy around this either (see https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4740/files/611b3ac485adc698df5b3a7fe062cf8c4e42af0e#r1782492951)... 
There are few points:
old way without offsets can stay as is... for newer way with offsets we don't really want to call GetEncryptionKeyAsync more than absolutely necessary, it's quite expensive... and then the question is whether we want to bloat CosmosEncryptor API at all (since we'll be calling once retrieved DEK directly). That way we could only add GetEncryptionKeyAsync here and drop all the remaining methods.
But again, it's mainly question of what kind of API we want to expose (and where).

Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests/Readme.md
JanHyka: CPU time fluctuates a bit, I am running it on laptop workstation and all various processes compete for resources, machine throttles out of my control etc., so numbers with that small difference are not to be trusted absolutely. 

However there might be some small static cost increase that gets balanced out with bigger than very small payloads getting processed faster. 

We'll be able to reduce below 30us on this small load with next PR thanks to changes introduced here so this small difference should be bearable for now.

kirankumarkolli (approved):

PR #4742, opened 2024-10-02
[INTERNAL] lessmsi: Fixes the chocolatey unavailability issues

kirankumarkolli:
[INTERNAL] lessmsi: Fixes the chocolatey unavailability issues

Lessmsi is pre-loaded into repo and used at rutnime to avoid availablity issues with Chocolatey.
ref: https://github.com/activescott/lessmsi/releases/tag/v2.1.1

17% of CI pipelines and 35% of daily pipelines are impacted by his unreliability
image

```
By installing, you accept licenses for the packages.
Downloading package from source '[https://community.chocolatey.org/api/v2/'](https://community.chocolatey.org/api/v2/%27)
[NuGet] Error downloading 'lessmsi.2.1.1' from '[https://community.chocolatey.org/api/v2/package/lessmsi/2.1.1'.](https://community.chocolatey.org/api/v2/package/lessmsi/2.1.1%27.)
[NuGet] Response status code does not indicate success: 503 (Service Unavailable: Back-end server is at capacity).
[NuGet] Error downloading 'lessmsi.2.1.1' from '[https://community.chocolatey.org/api/v2/package/lessmsi/2.1.1'.](https://community.chocolatey.org/api/v2/package/lessmsi/2.1.1%27.)
[NuGet] Response status code does not indicate success: 503 (Service Unavailable: Back-end server is at capacity).
lessmsi not installed. An error occurred during installation:
Error downloading 'lessmsi.2.1.1' from '[https://community.chocolatey.org/api/v2/package/lessmsi/2.1.1'.](https://community.chocolatey.org/api/v2/package/lessmsi/2.1.1%27.)
lessmsi package files install failed with exit code 1. Performing other installation steps.
```


templates/emulator-setup.yml
```diff
@@ -6,10 +6,16 @@ steps:
...
       dir "$env:temp" 
-      choco install lessmsi
-      choco upgrade lessmsi
+
       mkdir "$env:temp\Azure Cosmos DB Emulator"
-      lessmsi x "$env:temp\azure-cosmosdb-emulator.msi" "$env:temp\Azure Cosmos DB Emulator\"
+      $rootLocation=(Get-Location).Path
+
+      mkdir "$env:temp\lessmsi"
+      Expand-Archive -LiteralPath 'tools\lessmsi-v2.1.1.zip' -DestinationPath "$env:temp\lessmsi"
```
Pilchie, line 14: What is the behavior if this folder already exists? Do we ever re-use machines such that this might already exist and cause errors?

Pilchie, line 14: How big is this zip?  Once you add this binary, it will be in the history forever, even if we later stop using it, and every fresh clone of the repo will require folks to download it.

kirankumarkolli: It seems not, even for emulator the folder is created just above unconditional so-far no issues. 
But lets me address it anyways.

Pilchie: I'm not worried about 650KB

Pilchie: You've just moved the problem down one line. What does `Expand-Archive` do if the directory exists?  Does it overwrite the files? Error?  Skip the files?

kirankumarkolli: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/expand-archive?view=powershell-7.4#-destinationpath
image

kirankumarkolli: This is even true for lessmsi also the target direct if not exists it will create it. 

Removing the directory creation part to simplify.

kirankumarkolli: I see your comment is about if the files are already extracted, got it.

kirankumarkolli: lessmsi is overwriting it by-default
Expand-archive is not and an explicit '-Force' option needed. 

Will update script to clean folder if-exists (if containers are re-used then process use might fail in both cases anyways)

templates/emulator-setup.yml
```diff
@@ -6,11 +6,28 @@ steps:
...
-      choco upgrade lessmsi
-      mkdir "$env:temp\Azure Cosmos DB Emulator"
-      lessmsi x "$env:temp\azure-cosmosdb-emulator.msi" "$env:temp\Azure Cosmos DB Emulator\"
-      Add-MpPreference -ExclusionPath "$env:temp\Azure Cosmos DB Emulator\SourceDir\Azure Cosmos DB Emulator"
+
+      function Remove-DirectoryIfExists {
+          param ([string]$Path)
+
+          if (Test-Path -Path $Path -PathType Container) {
+            Remove-Item -Path $Path -Recurse -Force
```
Pilchie, line 14: Could conceivably still hit issues here with locked files, etc (Filesystems are hard), but I'm fine with merging this and seeing if we run into any issues in practice.

Pilchie (approved):

PR #4745, opened 2024-10-03
[INTERNAL] CI: Fixes adding auto-retries for tests

kirankumarkolli:
[INTERNAL] CI: Fixes adding auto-retries for tests

Only the failed tests are re-run. Its still possible that that re-run tests might see the state of the container at that time. 
Test needs to be proofed in future.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ActivityScope.cs
+++ b/Microsoft.Azure.Cosmos/src/ActivityScope.cs
@@ -13,6 +13,7 @@ namespace Microsoft.Azure.Cosmos
         public ActivityScope(Guid activityId)
         {
             this.ambientActivityId = Trace.CorrelationManager.ActivityId;
+
             Trace.CorrelationManager.ActivityId = activityId;
         }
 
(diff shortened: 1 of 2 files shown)


Microsoft.Azure.Cosmos/src/ActivityScope.cs
```diff
@@ -13,6 +13,7 @@ internal sealed class ActivityScope : IDisposable
         public ActivityScope(Guid activityId)
         {
             this.ambientActivityId = Trace.CorrelationManager.ActivityId;
+
```
sourabh1007, line 16: nit:  revert this file.

templates/build-test.yml
```diff
@@ -35,6 +35,7 @@ jobs:
   - task: DotNetCoreCLI@2
     displayName: Microsoft.Azure.Cosmos.Tests
     condition: succeeded()
+    retryCountOnTaskFailure: 2
```
sourabh1007, line 38: it will increase the feedback cycle on an on going PR because, any genuine failure will be reported after 2 iteration of whole task run (not even test run).

sourabh1007: at test class level also we can define retrycounts https://github.com/Azure/azure-cosmos-dotnet-v3/blob/1e150b6fc89ee5776afa79bac8bf88d0bc773c1f/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/RetriableTestClass.cs#L26, we might explore this and target only flaky tests.

philipthomas-MSFT (approved):
LGTM

Pilchie (approved):

PR #4752, opened 2024-10-05
[INTERNAL] CI: Fixes Moving flaky tests into separate groups

kirankumarkolli:
[INTERNAL] CI: Moving flaky tests into separate groups

### DIFF
--- a/templates/build-preview.yml
+++ b/templates/build-preview.yml
@@ -139,7 +139,7 @@ jobs:
     inputs:
       command: test
       projects: 'Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/*.csproj'
-      arguments: ${{ parameters.Arguments }} --filter "TestCategory!=Flaky" --configuration ${{ parameters.BuildConfiguration }} /p:IsPreview=true /p:OS=${{ parameters.OS }}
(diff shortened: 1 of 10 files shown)


templates/build-preview.yml
```diff
@@ -142,4 +142,32 @@ jobs:
...
+      version: '6.x'
+
+  - task: DotNetCoreCLI@2
+    displayName: PREVIEW Microsoft.Azure.Cosmos.Tests
+    retryCountOnTaskFailure: 4
+    condition: succeeded()
+    inputs:
+      command: test
+      projects: 'Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/*.csproj'
+      arguments: --filter "TestCategory=Flaky" --verbosity normal --configuration ${{ parameters.BuildConfiguration }} /p:IsPreview=true /p:OS=${{ parameters.OS }}
```
Pilchie, line 170: You miss out on the other parts of `parameters.Arguments` here - looks like that is `TestCategory!=Quarantine & TestCategory!=Functional`.  Is that ok?

Pilchie (approved):

templates/build-preview.yml
kirankumarkolli: Yes, it’s a pure whitelisting model.
Idea is to retry them more times and also manual retry will be quicker.

PR #4753, opened 2024-10-06
[Client encryption]: Switch to non-allocating MDE 2.0 api

JanHyka:
# Pull Request Template

## Description

Switch both MDE encryption and encryption paths to use MDE2.0 api

## Type of change

## Closing issues

Contributes to #4678


Microsoft.Azure.Cosmos.Encryption.Custom/src/ArrayPoolManager.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                if (disposing)
+                {
+                    foreach (T[] buffer in this.rentedBuffers)
+                    {
+                        ArrayPool<T>.Shared.Return(buffer, clearArray: true);
+                    }
+                }
+
+                this.rentedBuffers = null;
```
kirankumarkolli, line 37: nit: Move inside if (disposing) section

Microsoft.Azure.Cosmos.Encryption.Custom/src/Encryptor.cs
```diff
@@ -13,6 +13,15 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom
     /// </summary>
     public abstract class Encryptor
     {
+        /// <summary>
```
kirankumarkolli, line 16: What's the change here? (Is it moving line from below to up)?

Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
```diff
@@ -37,7 +37,7 @@
   <ItemGroup>
      <PackageReference Include="Azure.Core" Version="1.38.0" />
      <PackageReference Include="Azure.Identity" Version="1.11.4" />
-     <PackageReference Include="Microsoft.Data.Encryption.Cryptography" Version="1.2.0" />
+     <PackageReference Include="Microsoft.Data.Encryption.Cryptography" Version="2.0.0-pre*" />
```
kirankumarkolli, line 40: Custom package is in preview so preview pckage dependency is fine.
Wild card as dependency for public package is not a good practice. Application can upgrade as needed. 
For now can we please limit to pre007?

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
```diff
@@ -100,19 +105,23 @@ public static async Task<Stream> EncryptAsync(
                             continue;
                         }
 
-                        (typeMarker, plainText) = EncryptionProcessor.Serialize(propertyValue);
```
kirankumarkolli, line 103: nit: CosmosDB convention for static/const is to explicitly use the class-cope in naming, will help improve readability. 
I am fine either way just letting you know FYI.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
```diff
@@ -121,16 +130,16 @@ public static async Task<Stream> EncryptAsync(
                             throw new InvalidOperationException($"{nameof(Encryptor)} returned null cipherText from {nameof(EncryptAsync)}.");
                         }
 
-                        itemJObj[propertyName] = cipherTextWithTypeMarker;
+                        itemJObj[propertyName] = cipherTextWithTypeMarker.AsSpan(0, encryptedBytesCount + 1).ToArray();
```
kirankumarkolli, line 133: New allocation, any possibilities to avoid it?

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
```diff
@@ -283,7 +292,18 @@ private static async Task<DecryptionContext> MdeEncAlgoDecryptObjectAsync(
             CosmosDiagnosticsContext diagnosticsContext,
             CancellationToken cancellationToken)
         {
-            JObject plainTextJObj = new ();
+            _ = diagnosticsContext;
+
+            if (encryptionProperties.EncryptionFormatVersion != 3)
```
kirankumarkolli, line 297: If the data is already encrypted with older format version then this will be BIG breaking change right?
And the workaround is to re-encrypt all data, feels like a heavy hammer

Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Tests/AeadAes256CbcHmac256AlgorithmTests.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Cosmos.Encryption.Tests
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using System;
     using System.Linq;
+    using System.Text;
```
kirankumarkolli, line 11: nit: Is this change needed?

Microsoft.Azure.Cosmos.Encryption.Custom/src/Encryptor.cs
JanHyka: Yup, I have synchronized the order with CosmosEncryptor.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
JanHyka: We should get full 2.0.0 soon, I'll check with Juraj for conditions. putting pre007 for now.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
JanHyka: There is no change actually, this test was running for each Property in previous version, now it runs only once at the beginning of decryption (original file ln 357).

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
JanHyka: Sadly JValue and serializers can't work with offset/length so there is no way to use rented arrays. But we can just directly create fixed size array instead.

PR #4756, opened 2024-10-06
[INTERNAL] Build: Fix warnings in encryption project

kirankumarkolli:
[INTERNAL] Build: Fix warnings in encryption project

### DIFF
--- a/Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainerExtensions.cs
+++ b/Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainerExtensions.cs
@@ -13,6 +13,7 @@ namespace Microsoft.Azure.Cosmos.Encryption
     /// <summary>
     /// Extension methods for <see cref="Container"/> to support client-side encryption.
     /// </summary>
+    [CLSCompliant(false)]
     public static class EncryptionContainerExtensions
     {
         /// <summary>
--- a/Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
+++ b/Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
@@ -6,6 +6,7 @@
     <LangVersion>$(LangVersion)</LangVersion>
     <CurrentDate>$([System.DateTime]::Now.ToString(yyyyMMdd))</CurrentDate>
     <Version>$(CustomEncryptionVersion)</Version>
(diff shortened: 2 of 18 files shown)


JanHyka:
nit: seeing 'mdesrc' it looks like quite old copy of mde library... it might be worth considering just refencing at least 1.2.0 and getting rid of this copy, now that we have release versions.

kirankumarkolli:
> nit: seeing 'mdesrc' it looks like quite old copy of mde library... it might be worth considering just refencing at least 1.2.0 and getting rid of this copy, now that we have release versions.

Part of our PR its upgraded to 2.0.0-pre007 right?

JanHyka:
> > nit: seeing 'mdesrc' it looks like quite old copy of mde library... it might be worth considering just refencing at least 1.2.0 and getting rid of this copy, now that we have release versions.
> 
> Part of our PR its upgraded to 2.0.0-pre007 right?

With `1.2.0 stable` release you should have out of box improvements both on encryption algorithm as well as sql serializers, with no code changes on `Cosmos.Encryption` itself. However while switching from this mdesrc copy to real mde, I'd suggest benchmark before/after.

`2.0.0` is on -pre until we can deploy it together with new `Cosmos.Encryption.Custom -preview` and validate everything works as intended. Even though we went hardcore with testing (100% line/branch/mutation coverage) having partner service validation is still worth it. Also except few small cases 2.0.0 brings improvements purely via non-allocation api, there are almost no direct compute improvements, so upgrading to that compared to 1.2.0 without additional code changes in Cosmos.Encryption won't bring you much extra.

Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainerExtensions.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.Azure.Cosmos.Encryption
     /// <summary>
     /// Extension methods for <see cref="Container"/> to support client-side encryption.
     /// </summary>
+    [CLSCompliant(false)]
```
Pilchie, line 16: Nit: You could probably suppress this diagnostic. I'm not sure anyone cares about CLSCompliance anymore.

Pilchie (approved):

philipthomas-MSFT (approved):
LGTM

Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainerExtensions.cs
kirankumarkolli: Will circle back on this wholistically across all projects

PR #4760, opened 2024-10-08
[Internal] Flaky Test: Fixes Client Telemetry Flaky tests

### LINKED ISSUE #4731: [ClientTelemetry] Client telemetry tests are flaky
`ClientTelemetryTestsBase` has logic that is timing dependent. This causes the tests to be flaky.
The test also appears to be making real network requests, presumably to a server outside the node that tests run on.
Ideally, the `WaitAndAssert` method should be rewritten to correct these deficiencies. 

**To Reproduce**
This reproduces in the PR builds.

**Expected behavior**
Tests should not use calls to `Task.Delay` or make network requests outside the node that they run on. The outcome of a test should be deterministic.

sourabh1007:
## Description

Removing dependency on `Task.Delay()
`
## Type of change

## Closing issues

To automatically close an issue: closes #4731


philipthomas-MSFT (commented):
nit: more context on how this stabilizes the tests and are there things we need to avoid with future tests.

kirankumarkolli (approved):

PR #4770, opened 2024-10-09
[INTERNA] CI: Fixes disabling Binskim as nuget repo failing

kirankumarkolli:
[INTERNA] CI: Fixes disabling Binskim as nuget repo failing

### DIFF
--- a/templates/static-tools.yml
+++ b/templates/static-tools.yml
@@ -25,6 +25,7 @@ jobs:
 
   - task: securedevelopmentteam.vss-secure-development-tools.build-task-binskim.BinSkim@4
     displayName: 'BinSkim'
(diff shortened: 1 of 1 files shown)


templates/static-tools.yml
```diff
@@ -25,6 +25,7 @@ jobs:
 
   - task: securedevelopmentteam.vss-secure-development-tools.build-task-binskim.BinSkim@4
     displayName: 'BinSkim'
+    condition: eq(1,2) #disablng as nuget repo failing
```
Pilchie, line 28: Can we file an issue on this?  Also, why not just `false` here?

I suspect this will get us in SFI trouble at some point, so we should work on tracking it down.

PR #4776, opened 2024-10-09
[INTERNAL] CI: Adds .NET8 SDK support

kirankumarkolli:
[INTERNAL] CI: Adds .NET8 SDK support

RISK: not all pipelines are exercised part of CI gates and might be surprises later.


azure-pipelines-encryption-custom.yml
```diff
@@ -21,7 +21,13 @@ stages:
         displayName: Use .NET 6.0
         inputs:
           packageType: 'sdk'
```
Pilchie, line 23: Could potentially change this to a .NET 6 runtime instead.

Pilchie (commented):
Do we want to actually run the tests with both, instead of just building with the .NET 8 SDK?

kirankumarkolli:
> Do we want to actually run the tests with both, instead of just building with the .NET 8 SDK?
That's next follow-up PR.

NaluTripician (approved):
LGTM

Pilchie (approved):

PR #4779, opened 2024-10-09
[Client encryption]: Add JsonNodeSqlSerializer

JanHyka:
# Pull Request Template

## Description

Initial commit for JsonNode on Encryption path. This is currently not executed on any production/preview path. Depends on JsonNode features available from System.Text.Json 8.0+.

- Adds JsonNodeSqlSerializer
- JObjectSqlSerializer now doesn't format inner serialized JArrays/JObjects (with line breaks/indentations)

To be processed after #4766 

## Type of change

Please delete options that are not relevant.

- [] New feature (non-breaking change which adds functionality)

## Closing issues

Contributes to #4678

PR #4780, opened 2024-10-09
[Client encryption]: Add JsonNode Encryption processor

JanHyka:
# Pull Request Template

## Description

- Added support for System.Text.JsonNode DOM on encryption path
- Configuration option to select Json Processor (defaults to Newtonsoft)
- Custom byte[] coverter supporting offset+length for JsonNode
- Perf tests expanded to cover both json processors (JsonNode decryption still uses Newtonsoft, this will be upgraded with further PRs)

To be processed after #4779 

## Type of change

## Closing issues

Contributes to #4678


Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests/Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests.csproj
```diff
@@ -6,6 +6,7 @@
     <TargetFrameworks>net8.0</TargetFrameworks>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+	<ServerGarbageCollection>true</ServerGarbageCollection>
```
kirankumarkolli, line 9: nit: tabs to spaces

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
```diff
@@ -6,6 +6,25 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom
 {
     using System.Collections.Generic;
 
+    /// <summary>
+    /// API for JSON processing
+    /// </summary>
+    public enum JsonProcessor
```
kirankumarkolli, line 12: Processor is for encryption processing only. 
Apart from performance any functional aspects developers to be aware of to make this choice? (including for migration)

JanHyka: In this PR I have only done encryption path to reduce PR size. Next PR is adding also decryption path and cross testing - encryption, decryption and validation in all combinations of Newtonsoft x STJ x validateBy (Newtonsoft/STJ). So, there should be full compatibility. Different Json engines though could have different level of tolerance towards jsons that don't completely follow specification. So for example if there was a service that takes unvalidated request body of unspecified quality and sends its stream to EncryptionProcessor, it's possible the behavior won't be identical, though same could happen with different versions of Newtonsoft itself too.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProperties.cs
```diff
@@ -5,23 +5,29 @@
...
 
         [JsonProperty(PropertyName = Constants.EncryptedData)]
+        [JsonPropertyName(Constants.EncryptedData)]
         public byte[] EncryptedData { get; }
 
         [JsonProperty(PropertyName = Constants.EncryptedPaths)]
+        [JsonPropertyName(Constants.EncryptedPaths)]
         public IEnumerable<string> EncryptedPaths { get; }
 
         [JsonProperty(PropertyName = Constants.CompressionAlgorithm)]
```
kirankumarkolli, line 33: JsonPropertyName for compression properties?

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
kirankumarkolli: On decryption the choice of Newstonsoft vs STJ is it based on Options or what's used during encryption?
(Just curiosity to understand, anyways decryption is not part of this PR)

kirankumarkolli (approved):

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
JanHyka: If you call generic  decryption path method (Stream in/out scenario), you have to hint if you prefer STJ, method without extra param defaults on Newtonsoft for now, to stay on a safe side. If you call methods that take specific Json engine class, it's determined by that (ie. JObject param -> Newtonsoft x JsonNode -> STJ).

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProperties.cs
JanHyka: fixed, thanks!

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
kirankumarkolli: Point I am trying to understand is its possible that on encryption path JObject is used. 
And during decryption time later in future its possible that the code choice might be JsonNode, the developer needs to be aware of it. Is that right?

JanHyka: Encryption and decryption paths are entirely decoupled, you can use one processor for encryption and other for decryption. 
On Encryption path, you have
Stream -> Stream [via JObject, JsonNode]

On Decryption path
Stream -> Stream [via JObject, JsonNode]

Without extra parameters, JObject is used by default. That's what happens if consumers do upgrade by version bump only. If you specify processor via extra parameter in overloaded method, processor of your choice is used. You can use any combination on both encryption and decryption path without change of result.

We have also added Decryption for
JObject -> JObject
JsonNode -> JsonNode
however I am rather inclining towards removing those before release (be it -preview or full one). While it might seem comfortable for users to send in their JObject/JsonNode directly, the overhead compared to Utf8Reader/Writer approach I am implementing now is so big that relevance of any DOM based approach is very poor.

PR #4781, opened 2024-10-09
AppInsights: Adds classic attribute back to cosmos db to support appinsights sdk

sourabh1007:
## Description

As part of this PR, adding back attributes required by appinsights sdk. It broke the customer experience with appinsight as Appinsight SDK supports very specific set of attributes.

This implementation will change in future release, where open telemetry attributes will be controlled by Env Variable as mentioned here.
https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#semantic-conventions-for-database-client-calls

**4.42.3**
image

**v4.43.0**
image

**v4.43.1**
image

After this PR:
image

After this change, customer has to set `OTEL_SEMCONV_STABILITY_OPT_IN` to `database/dup`, in order to see latest attributes. otherwise SDK will emit only classic attributes which would be compatible with appinsights sdk also.

## Type of change

## Closing issues


philipthomas-MSFT (commented):
LGTM. I will re-review after Draft version

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -149,19 +152,33 @@ public void Record(
         {
             if (this.IsEnabled)
             {
-                this.scope.AddAttribute(OpenTelemetryAttributeKeys.DbOperation, operationName);
-                this.scope.AddAttribute(OpenTelemetryAttributeKeys.DbName, databaseName);
-                this.scope.AddAttribute(OpenTelemetryAttributeKeys.ContainerName, containerName);
-                
+                if (otelStabilityMode == OpenTelemetryStablityModes.DatabaseDupe)
+                {
+                    this.scope.AddAttribute(OpenTelemetryAttributeKeys.DbOperation, operationName);
```
kirankumarkolli, line 157: Choice is about naming right? 

How about have fields/property/methods which can return the right operation name?

kirankumarkolli (approved):

philipthomas-MSFT (approved):

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
sourabh1007: yes, it needs to be redesign, I will take it as next task.

PR #4783, opened 2024-10-09
[INTERNAL] CI: Adds multi targeting for tests [DO-NOT-COMMIT]

kirankumarkolli:
[INTERNAL] CI: Adds multi targeting for tests


Microsoft.Azure.Cosmos.Encryption.Custom/tests/EmulatorTests/Microsoft.Azure.Cosmos.Encryption.Custom.EmulatorTests.csproj
```diff
@@ -3,7 +3,7 @@
     <IsTestProject>true</IsTestProject>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Platform>AnyCPU</Platform>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net6.0;net8.0</TargetFramework>
```
Pilchie, line 6: I don't think we need to do this, and I don't think it's sufficient to actually test on .NET 8.

Instead, we should multi-target to `net8.0` only if there are APIs we want to use there.

We should change our invocations of `dotnet test` to specify what runtimes to run on and add steps there in the pipelines, instead of here.

kirankumarkolli: This PR is not meant for commit (full round of validation of .NET8 environment validation). 
Yml files changes are little bigger then 4Porj files hence shortcut for this validation.

PR #4799, opened 2024-10-14
Cherry-Pick Appinsights: Adds classic attribute back to cosmos db to support appinsights sdk (#4781)

sourabh1007:
## Description

As part of this PR, adding back attributes required by appinsights sdk. It broke the customer experience with appinsight as Appinsight SDK supports very specific set of attributes.

This implementation will change in future release, where open telemetry attributes will be controlled by Env Variable as mentioned here.

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#semantic-conventions-for-database-client-calls

**4.42.3**

image

**v4.43.0**

image

**v4.43.1**

image

After this PR:

image

After this change, customer has to set `OTEL_SEMCONV_STABILITY_OPT_IN` to `database/dup`, in order to see the otel (new and old) attributes. otherwise SDK will emit only classic attributes which would be compatible with appinsights sdk also.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
+++ b/Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.Cosmos.Telemetry
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using Antlr4.Runtime.Misc;
     using global::Azure.Core;
     using Microsoft.Azure.Cosmos.Telemetry.Diagnostics;
     using Microsoft.Azure.Documents;
@@ -18,6 +19,8 @@ namespace Microsoft.Azure.Cosmos.Telemetry
     internal struct OpenTelemetryCoreRecorder : IDisposable
     {
         private const string CosmosDb = "cosmosdb";
+
+        private static readonly string otelStabilityMode = Environment.GetEnvironmentVariable("OTEL_SEMCONV_STABILITY_OPT_IN");
 
         private readonly DiagnosticScope scope = default;
         private readonly CosmosThresholdOptions config = null;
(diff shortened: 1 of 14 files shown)


philipthomas-MSFT (approved):

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
```diff
@@ -149,19 +152,33 @@ public void Record(
         {
             if (this.IsEnabled)
             {
-                this.scope.AddAttribute(OpenTelemetryAttributeKeys.DbOperation, operationName);
-                this.scope.AddAttribute(OpenTelemetryAttributeKeys.DbName, databaseName);
-                this.scope.AddAttribute(OpenTelemetryAttributeKeys.ContainerName, containerName);
-                
+                if (otelStabilityMode == OpenTelemetryStablityModes.DatabaseDupe)
+                {
+                    this.scope.AddAttribute(OpenTelemetryAttributeKeys.DbOperation, operationName);
```
kirankumarkolli, line 157: Dupe needs to emit both old and news as described in comments?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryStablityModes.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// emit the new, stable database conventions, and stop emitting the old experimental database conventions that the instrumentation emitted previously.
+        /// </summary>
+        public const string Database = "database";
+
+        /// <summary>
+        ///  emit both the old and the stable database conventions, allowing for a seamless transition.
+        /// </summary>
+        public const string DatabaseDupe = "database/dup";
```
kirankumarkolli, line 20: The value is set by the application right? 

Is "database/dup" aligned with our versioning strategy forward?
AFAIU we will emit only one version aligned right?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryCoreRecorder.cs
sourabh1007: Here default behavior is to emit, _non-otel_ version of attributes by default (to mitigate the cx issue) and if OTEL_SEMCONV_STABILITY_OPT_IN is set to dupe. it will emit all _otel_ attributes.  There is no point to emit duplicate information for Otel customers. It will confuse them

Versioning using `database` and `database/dupe` will be implemented as a separate PR.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryStablityModes.cs
sourabh1007: Versioning using database and database/dupe will be implemented as a separate PR.

kirankumarkolli (approved):

PR #4806, opened 2024-10-15
[Client encryption]: Adds Stream based processor

JanHyka:
# Pull Request Template

## Description

- JsonProcessor.Stream was added to drop DOM overhead.
- EncryptionProcessor has added overloads allowing consumer provision of output stream
- tests were extended to new processor
- performance tests were extended by provided stream overloads. RecyclableMemoryStream 3.0.1 is being used.

## Type of change

- [] This change might require a documentation update

## Closing issues

Contributes to #4678


Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
```diff
@@ -6,6 +6,31 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom
 {
     using System.Collections.Generic;
 
+    /// <summary>
+    /// API for JSON processing
+    /// </summary>
+    public enum JsonProcessor
```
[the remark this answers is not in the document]
JanHyka: Yes, if you check MdeEncryptionProcessorTests, we are testing every combination of Processor x Compressor. However bear in mind Newtonsoft vs STJ can behave differently when parsing out-of-spec json. JsonNode and Stream will behave identical, JsonNode is built up on Utf8JsonReader.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
```diff
@@ -120,6 +183,152 @@ public static async Task<Stream> EncryptAsync(
...
+            if (properties?.EncryptionProperties == null)
+            {
+                await input.CopyToAsync(output, cancellationToken: cancellationToken);
+                return null;
+            }
+
+            if (properties.EncryptionProperties.EncryptionAlgorithm != CosmosEncryptionAlgorithm.MdeAeadAes256CbcHmac256Randomized)
+            {
+                input.Position = 0;
+                throw new NotSupportedException($"Streaming mode is only allowed for {nameof(CosmosEncryptionAlgorithm.MdeAeadAes256CbcHmac256Randomized)}");
```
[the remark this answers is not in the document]
JanHyka: I have added that fallback but don't think it's a good idea.

Old encryption algorithm was obsoleted 3.5yrs ago. If partners still have data that weren't refreshed for so long, they should accommodate by fallback on their side if they want to use new api. In fact I'd go even further and entirely remove old algorithm option on encryption path.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
```diff
@@ -90,6 +91,60 @@ public static async Task<Stream> EncryptAsync(
...
+                if (string.IsNullOrWhiteSpace(path) || path[0] != '/' || path.IndexOf('/', 1) != -1)
+                {
+                    throw new InvalidOperationException($"Invalid path {path ?? string.Empty}, {nameof(encryptionOptions.PathsToEncrypt)}");
+                }
+
+                if (path.AsSpan(1).Equals("id".AsSpan(), StringComparison.Ordinal))
+                {
+                    throw new InvalidOperationException($"{nameof(encryptionOptions.PathsToEncrypt)} includes a invalid path: '{path}'.");
+                }
+            }
```
juraj-blazek, line 132: Let's move these validations to `EncryptionOptionsExtensions.Validate`

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
```diff
@@ -140,11 +195,76 @@ public static async Task<Stream> EncryptAsync(
...
+            else if (properties.EncryptionProperties.EncryptionAlgorithm == CosmosEncryptionAlgorithm.AEAes256CbcHmacSha256Randomized)
+            {
+                (Stream stream, context) = await DecryptAsync(input, encryptor, diagnosticsContext, cancellationToken);
+                await stream.CopyToAsync(output, cancellationToken);
+                output.Position = 0;
+            }
+            else
+            {
+                input.Position = 0;
+                throw new NotSupportedException($"Streaming mode is not supported for encryption algorithm {properties.EncryptionProperties.EncryptionAlgorithm}");
```
juraj-blazek, line 253: Let's use the same exception format as below, because this is not about streaming, but unknown stored algo in general:
```c#
throw new NotSupportedException($"Encryption Algorithm : {encryptionProperties.EncryptionAlgorithm} is not supported.");
```

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
```diff
@@ -140,11 +195,76 @@ public static async Task<Stream> EncryptAsync(
...
+            CosmosDiagnosticsContext diagnosticsContext,
+            JsonProcessor jsonProcessor,
+            CancellationToken cancellationToken)
+        {
+            if (input == null)
+            {
+                return null;
+            }
+
+            if (jsonProcessor != JsonProcessor.Stream)
```
juraj-blazek, line 218: I think this check is redundant, the caller should make the decision if to use streaming API.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -0,0 +1,281 @@
...
+            }
+
+            state = reader.CurrentState;
+            return reader.BytesConsumed;
+        }
+
+        private void TransformDecryptProperty(ref Utf8JsonReader reader, Utf8JsonWriter writer, string decryptPropertyName, EncryptionProperties properties, DataEncryptionKey encryptionKey, bool containsCompressed, ArrayPoolManager arrayPoolManager)
+        {
+            BrotliCompressor decompressor = null;
+            if (properties.EncryptionFormatVersion == 4)
```
juraj-blazek, line 219: We really should turn these into constants or `static readonly`'s everywhere

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
JanHyka: That's a lot of code to add as in that case we'd have to start supporting it in all processors. Given complexity of this PR and fact we want to entirely drop JsonNode processor type next, I don't think it's a good idea to put Stream api for encryption and decryption in MdeEncryptionProcessor, MdeJObjectEncryptionProcessor and MdeJsonNodeEncryptionProcessor for now. I'd prefer to drop JsonNode first.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -0,0 +1,281 @@
...
+                }
+            }
+
+            writer.Flush();
+            outputStream.Position = 0;
+
+            return EncryptionProcessor.CreateDecryptionContext(pathsDecrypted, properties.DataEncryptionKeyId);
+        }
+
+        private long TransformDecryptBuffer(ReadOnlySpan<byte> buffer, bool isFinalBlock, Utf8JsonWriter writer, ref JsonReaderState state, ref bool isIgnoredBlock, ref string decryptPropertyName, HashSet<string> encryptedPaths, List<string> pathsDecrypted, EncryptionProperties properties, bool containsCompressed, ArrayPoolManager arrayPoolManager, DataEncryptionKey encryptionKey)
```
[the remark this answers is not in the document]
JanHyka: In general I avoid local methods as it makes code way harder to read (at least for me). Both options are possible based on preferences though, there are no differences on benchmark. I've updated the code to use local methods instead of methods that should suffer header compression for comparison.
@kirankumarkolli what are your preferences?

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
JanHyka: EncryptionFormatVersion.AeAed,Mde and MdeWithCompression were added

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
JanHyka: updated

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
JanHyka: updated

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
```diff
@@ -16,12 +16,18 @@ public enum JsonProcessor
...
         /// System.Text.Json
         /// </summary>
         /// <remarks>Available with .NET8.0 package only.</remarks>
         SystemTextJson,
+
+        /// <summary>
+        /// Ut8JsonReader/Writer
+        /// </summary>
+        /// <remarks>Available with .NET8.0 package only.</remarks>
+        Stream,
```
kirankumarkolli, line 30: Do both STJ and Utf8Reader are needed as options?

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionPropertiesWrapper.cs
```diff
@@ -0,0 +1,19 @@
+﻿// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Encryption.Custom
+{
+    using System.Text.Json.Serialization;
+
+    internal class EncryptionPropertiesWrapper
```
kirankumarkolli, line 9: Who are the consumers?

Microsoft.Azure.Cosmos.Encryption.Custom/src/RentArrayBufferWriter.cs
```diff
@@ -0,0 +1,211 @@
...
+
+using System;
+using System.Buffers;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+/// <summary>
+/// https://gist.github.com/ahsonkhan/c76a1cc4dc7107537c3fdc0079a68b35
```
kirankumarkolli, line 17: @Pilchie its a copy for code, author is MS, is that fine?

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
JanHyka: I have followup PR prepared that drops JsonNode approach, we'd only keep Newtonsoft for 'just in case' compatibility and STJ's Utf8JsonReader.
Since non-DOM approach easily wins over JsonNode by a significant margin JsonNode is not really needed.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionPropertiesWrapper.cs
JanHyka: That's a holder to deserialize EncryptionProperties from serialized json. Since it's nested in the json document under "_ei" object, we map it using this wrapper and then in first pass deserialization use STJ's strongly typed deserializer to get it.

Microsoft.Azure.Cosmos.Encryption.Custom/src/RentArrayBufferWriter.cs
JanHyka: Also, we'll drop this in one of followup PRs entirely.

sourabh1007:
Can you please add the perf numbers also, it would help us to understand, that it is really contributing to the attached issue.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -0,0 +1,269 @@
...
+                // we need to scale out buffer
+                if (leftOver == dataSize)
+                {
+                    byte[] newBuffer = arrayPoolManager.Rent(buffer.Length * 2);
+                    buffer.AsSpan().CopyTo(newBuffer);
+                    buffer = newBuffer;
+                }
+                else if (leftOver != 0)
+                {
+                    buffer.AsSpan(dataSize - leftOver, leftOver).CopyTo(buffer);
```
kirankumarkolli, line 91: Curious: Are there any un-safe aspects of CopyTo using the same buffer? (Its kind of left shift so might be okey)

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -0,0 +1,269 @@
...
+                            break;
+                    }
+                }
+
+                state = reader.CurrentState;
+                return reader.BytesConsumed;
+            }
+        }
+
+        private void TransformDecryptProperty(ref Utf8JsonReader reader, Utf8JsonWriter writer, string decryptPropertyName, EncryptionProperties properties, DataEncryptionKey encryptionKey, bool containsCompressed, ArrayPoolManager arrayPoolManager)
```
kirankumarkolli, line 204: nit: argument per line please.

kirankumarkolli (approved):

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionPropertiesWrapper.cs
kirankumarkolli: This type is not referenced outside of this file, am I missing something?

JanHyka: It's used in EncryptionProcessor (for example ln. 194).

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
JanHyka: It should be safe, this part is pretty much same as example by .NET Core team:
https://github.com/dotnet/samples/blob/main/core/json/Program.cs
Internally CopyTo goes down to C++ memmove explicitly allowing object overlap:
https://en.cppreference.com/w/cpp/string/byte/memmove

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
JanHyka: changed to local method

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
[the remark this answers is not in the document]
juraj-blazek: > In fact I'd go even further and entirely remove old algorithm option on encryption path.

Yes, that can be on of the next steps. But we need to make sure we can read stored data.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
juraj-blazek: Nope, I didn't want to extend anything, just asking if we need this check here. As discussed, we have these on other places, so let's keep that unified. After the other planned cleanup changes we'll review the API once more.

juraj-blazek:
> Can you please add the perf numbers also, it would help us to understand, that it is really contributing to the attached issue.

The benchmark results can be found in the [Readme.md](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4806/files?short_path=a7e02b6#diff-a7e02b68ae790f24be2191125149b694198098fa81e913dc5391c70dc43f4a77)

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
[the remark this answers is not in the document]
JanHyka: It's added, partially here, partially in followup where we actually have means to test it.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
JanHyka: There should be cleanup opportunities, I suspect there might be leftover methods/classes with zero references as well that slipped under radar too.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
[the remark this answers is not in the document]
JanHyka: Switched to local methods where paramcount >= 7

kirankumarkolli (approved):

PR #4813, opened 2024-10-16
[INTERNAL] CI: Fixes emulator set-up to leverage central SDK teams scripts

kirankumarkolli:
[INTERNAL] CI: Fixes emulator set-up to leverage central SDK teams scripts

Observations
- Lessmsi: extraction taking longer (1M) then msiexec based model
- Starting emulator from ProgramFolder path is faster 
- Unnecessary exponential retry logic (current one)


templates/Cosmos-Emulator.ps1
```diff
@@ -0,0 +1,151 @@
...
+      return
+    }
+  }
+
+  $argumentList = ""
+  if (-not [string]::IsNullOrEmpty($StartParameters)) {
+      $argumentList += , $StartParameters
+  } else {
+    # Use the default params if none provided
+    $argumentList = "/noexplorer /noui /enablepreview /EnableSqlComputeEndpoint /SqlComputePort=9999 /disableratelimiting /partitioncount=10 /consistency=Strong"
```
Pilchie, line 107: Looks like we weren't passing `/SqlComputePort=9999` before.  Is that a problem?

Pilchie (dismissed):

templates/Cosmos-Emulator.ps1
kirankumarkolli: There is no real problem, some how this script is more efficient. 

As called out in description
- Lessmsi is an expensive process and unnecessary
- emulator startup retries are not helping it-sems better to let it complete 
- The only fundamental change I see is the place where its installed. Other than that there no logical changes on start-up paths. Unclear where the real difference is (will let the rolling build for a week and then we can re-access)

kirankumarkolli: Reverted compute port changes

PR #4814, opened 2024-10-16
DeleteAllItemsByPartitionKeyStreamAsync: Adds DeleteAllItemsByPartitionKeyStreamAsync API to GA

kirankumarkolli:
DeleteAllItemsByPartitionKeyStreamAsync: Adds DeleteAllItemsByPartitionKeyStreamAsync API to GA

Changes
- DeleteAllItemsByPartitionKeyStreamAsync: Marked virtual with default NotSupportedException (Dependent libraries will continue to work)
- Encryption package related changes will be in a follow-up PR after the package is published


Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
```diff
@@ -1691,11 +1690,15 @@ public abstract ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithManu
...
         /// A <see cref="Task"/> containing a <see cref="ResponseMessage"/>.
         /// </returns>
-        public abstract Task<ResponseMessage> DeleteAllItemsByPartitionKeyStreamAsync(
+        public virtual Task<ResponseMessage> DeleteAllItemsByPartitionKeyStreamAsync(
                Cosmos.PartitionKey partitionKey,
                RequestOptions requestOptions = null,
-               CancellationToken cancellationToken = default);
+               CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
```
Pilchie, line 1698: Nit: I usually think of `NotImplementedException` as a marker for something that just hasn't been done yet, but that is intended to be done.  For something like this where we never intend it to be implemented, I would pick `NotSuportedException` instead.

Pilchie (approved):

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
kirankumarkolli: This API is implemented in core package. 

NotImplemented is to not break Encryption/dependency packages till we publish newer version and fix encryption package. In theory it will be supproted.

kundadebdatta (approved):

PR #4819, opened 2024-10-18
Azurecore: Fixes upgrading azure core dependency to latest

kirankumarkolli:
Azure.Core:  Fixed upgrading azure core dependency to 1.44.1

Changes
- ResourceType: Conflict between Azure.Core and Microsoft.Azure.Cosmos.Documents
- Microsoft.Bcl.AsyncInterfaces: Azure.Core needs at-least 6.0.0 (as Azure core upgraded part of minor version, we are good as well)
- Microsoft.Azure.Cosmos.Encryption.Custom: Direct dependency on Azure.Core removed (now its transitive)
- Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests: `NU1903` added Newtonsoft.Json dependency
- Performance project: `NU1903` Newtonsoft dependency upgraded


Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
```diff
@@ -112,11 +112,11 @@
 	</ItemGroup>
 
 	<ItemGroup>
-		<PackageReference Include="Azure.Core" Version="1.19.0" />
+		<PackageReference Include="Azure.Core" Version="1.44.1" />
 		<PackageReference Include="System.Collections.Immutable" Version="1.7.0" />
 		<PackageReference Include="System.Diagnostics.DiagnosticSource" Version="8.0.1" />
 		<PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
-		<PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.0.0" />
+		<PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
```
Pilchie, line 119: Is this necessary, or do we get it transitively through Azure.Core anyway?

Pilchie (approved):

NaluTripician (approved):

sourabh1007:
Just curious: What are benefits we are getting with this upgrade? and did we evaluate "breaking changes" because here we are making big jump (from 19 to 44 version)?

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
kirankumarkolli: Its coming transitively from two packages Azure.core and Direct packages. 
Since its an existing dependency raising it (which will match the end state in any case)

kirankumarkolli:
> Just curious: What are benefits we are getting with this upgrade? and did we evaluate "breaking changes" because here we are making big jump (from 19 to 44 version)?

It's a minor version upgrade (not major). Benefits
- Transient dependencies vulnerabilities (we are patching them in our code which is unnecessary)

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
Pilchie: Removing it here shouldn't be observably different from a consumer AFAIK, and means we don't have to be in the business of maintaining it in the future.  But it's not a big deal to have it.

PR #4839, opened 2024-10-22
[Dependencies]: Removes direct reference for Newtonfoft package and add a target build.

### LINKED ISSUE #4674: Microsoft.Azure.Cosmos references many out of support and vulnerable package versions.
**Describe the bug**
Microsoft.Azure.Cosmos references many packages which are out of support and vulnerable.

**To Reproduce**
Create a new project with the latest .NET 9.0 preview SDK which includes NuGet audit for security vulnerabilities.  Add a reference to `<PackageReference Include="Microsoft.Azure.Cosmos" Version="3.43.0"/>` and restore the project.

**Expected behavior**
No warnings when restoring.

**Actual behavior**
The following warnings occur:
```
    C:\scratch\azureCosmos\azureCosmos.csproj : warning NU1903: Package 'Newtonsoft.Json' 10.0.2 has a known high severity vulnerability, https://github.com/advisories/GHSA-5crp-9r3c-p9vr
    C:\scratch\azureCosmos\azureCosmos.csproj : warning NU1903: Package 'System.Net.Http' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-7jgj-8wvc-jh57
    C:\scratch\azureCosmos\azureCosmos.csproj : warning NU1903: Package 'System.Text.RegularExpressions' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-cmhx-cq75-c4mj
```

**Environment summary**
SDK Version: 3.43.0
.NET SDK: 9.0.0-preview.7.24405.7
OS Version: Windows 11 23H2

**Additional context**
See https://learn.microsoft.com/en-us/nuget/concepts/auditing-packages
https://devblogs.microsoft.com/nuget/nugetaudit-2-0-elevating-security-and-trust-in-package-management/

```[tasklist]
### Tasks
- [ ] https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4790
- [ ] https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4775
- [ ] https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4756
- [ ] https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4693
- [ ] https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4694
- [ ] https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4676
```

aavasthy:
# Pull Request Template

## Description
The intent of this change is to ensure that the `Newtonsoft.Json` dependency, which is currently used within the SDK, does not propagate to downstream consumers as Package dependency. The SDK references `Newtonsoft.Json` for its internal functionality, but this package should not be exposed to users of the SDK. This measure also avoids the risk of consumers relying on a specific version of `Newtonsoft.Json` that might have vulnerabilities or other limitations.

## Type of change
- **Dependency Management Update**: 
  - Make `Newtonsoft.Json` package reference as private asset.
  - Updated `Microsoft.Azure.Cosmos.targets` to ensure the Newtonsoft.Json dependency is not passed to consumer apps and there is a build failure when the consumer app do not have Newtonsoft package installed.
  - Provide an option to consumer app to disable this newtonsoft package build failure by setting DisableNewtonsoftJsonCheck to true.
  
```
  	<PropertyGroup>
		<DisableNewtonsoftJsonCheck>true</DisableNewtonsoftJsonCheck>
	</PropertyGroup>
```

**Testing the change on a consumer console app with updated dotnet SDK package without installing newtonsoft package  in consumer app:**
- **Installing the Microsoft.Azure.Cosmos package in the console app:**
   - No Newtonsoft package dependency getting added in consumer app.
   - Build failure when no newtonsoft package is available. Failure message "The Newtonsoft.Json package must be explicitly referenced with version >= 10.0.2. Please add a reference to Newtonsoft.Json, or set the 'AzureCosmosDisableNewtonsoftJsonCheck' property to 'true' to bypass this check."
   
image
image
image

**Once the Newtonsoft nuget package is included the build succeeds:**
image
 
image
image

**Testing the change on a consumer console app with updated dotnet SDK package when consumer app installs any other version of newtonsoft package on their end:**
image

**Testing the change on a consumer console app with updated dotnet SDK package when consumer app uses packages.config:**
image
image
image
image

**Testing the change on a consumer console app with updated dotnet SDK package and Newtonsoft available as transitive dependency:**
image
image

**Testing the change on a consumer console app with updated dotnet SDK package and build failure and allowing consumer app the option to opt out this build failure due to missing newtonsoft:**
image
image

**Testing the change on a consumer console app by installing updated Microsoft.Azure.Cosmos.Encryption which has dependency on updated  Microsoft.Azure.Cosmos package. The build failure happens if Newtonsoft is missing:**
image
image

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Closing issues

To automatically close an issue: closes #4674


bartelink:
Really appreciate the time you took to write the overview (and the fact that it's also very well written!)

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -53,5 +53,35 @@ Copyright (C) Microsoft Corporation. All rights reserved.
           <Visible>False</Visible>
       </ContentWithTargetPath>
   </ItemGroup>
-    
+
+
+  <!-- Run validation before compilation starts -->
+  <Target Name="VerifyNewtonsoftJsonReference"
+    BeforeTargets="PrepareForBuild;CoreCompile;GetCompileOutputs"
+    Condition="'$(BuildingProject)' != 'true'">
```
ericstj, line 61: Why did you add this condition?

aavasthy: Updated the code and added this condition instead Condition="'$(ExcludeRestorePackageImports)' != 'true'".
This ensures that the target only runs during the actual build process and not during package restore.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -53,5 +53,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+  <!-- Add check for Newtonsoft.Json package in cosnumer app.  -->
+  <Target Name="CheckNewtonsoftJsonVersion" BeforeTargets="CoreCompile" Condition="'$(ExcludeRestorePackageImports)' != 'true'">
+
+    <ItemGroup>
+        <NewtonsoftJsonPackageReference Include="@(PackageReference)" Condition="'%(PackageReference.Identity)' == 'Newtonsoft.Json'" />
+    </ItemGroup>
+    <!-- If Newtonsoft.Json is not referenced, fail the build -->
+    <Error Text="The Newtonsoft.Json package must be explicitly referenced with version >= 13.x.x."
+        Condition="'@(NewtonsoftJsonPackageReference)' == ''" />
```
Pilchie, line 66: We actually depend on v10.0.2, so we should probably list that as the minimum we would take. We could suggest 13.x, but we know we work with v10.0.2, and for customers that can't take the breaking changes but know they aren't vulnerable, they may want to stay on v10.0.2.

Pilchie (commented):
Slight nit in the overview - you should probably check "Breaking Change" instead of non-breaking bug fix.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: This build error message with 13.x suggestion comes when there isn't a newtonsoft package installed in consumer application. But if client is already on 10.0.2 then they won't receive this build error.
But yes makes sense to let them know newtonsoft versions >=10.0.2 will work with our package.

seesharprun: Per <https://github.com/advisories/GHSA-5crp-9r3c-p9vr>, .NET 9+ will issue a warning to the user is they have a version <13.0.1

Pilchie: Yes, we understand that, but given that there are breaking changes between v10 and v13, there may be customers that need to disable the warning after using other methods to avoid the vulnerability (as we did internally).

@aavasthy - a related test to run is to ensure that someone who adds a ref to v10 does get the warning, and that the "NoWarn" in our private assets ref doesn't suppress it for them.  (If it did, I would guess that's a NuGet bug, but we should verify).

Pilchie (dismissed):

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: Yes I did the test and added screenshots above for the same. So if consumer app installs a newtonsoft package which has vulnerability then it will give a warning during build.

Pilchie: Great!

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -53,5 +53,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+  <!-- Add check for Newtonsoft.Json package in cosnumer app.  -->
+  <Target Name="CheckNewtonsoftJsonVersion" BeforeTargets="CoreCompile" Condition="'$(ExcludeRestorePackageImports)' != 'true'">
+
+    <ItemGroup>
+        <NewtonsoftJsonPackageReference Include="@(PackageReference)" Condition="'%(PackageReference.Identity)' == 'Newtonsoft.Json'" />
+    </ItemGroup>
+    <!-- If Newtonsoft.Json is not referenced, fail the build -->
+    <Error Text="The Newtonsoft.Json package must be explicitly referenced with version >= 10.0.2."
+        Condition="'@(NewtonsoftJsonPackageReference)' == ''" />
```
ericstj, line 66: Did you test this in a project that uses `packages.config`?  I think that would import these targets, but not define any `PackageReference`.

Also I see you don't have a version check (any version will do) nor a property to opt-out.  Not having the version check might be ok since the version you depend on is very old, but make sure you know how this will fail in that case and document it.  

I wouldn't recommend shipping this without some opt-out property.   You cannot anticipate the wacky build systems that folks have - best to give a way to disable the error.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -53,5 +53,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
       </ContentWithTargetPath>
   </ItemGroup>
-    
+
+
+  <!-- Add check for Newtonsoft.Json package in cosnumer app.  -->
+  <Target Name="CheckNewtonsoftJsonVersion" BeforeTargets="CoreCompile" Condition="'$(ExcludeRestorePackageImports)' != 'true'">
+
+    <ItemGroup>
+        <NewtonsoftJsonPackageReference Include="@(PackageReference)" Condition="'%(PackageReference.Identity)' == 'Newtonsoft.Json'" />
```
ericstj, line 62: What if some other package brought a sufficient version of this transitively?  Would you still have a PackageReference?

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
eerhardt: > I wouldn't recommend shipping this without some opt-out property.

+1

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: I added a check for transitive dependency and also added test results for it in the description. 
If Newtonsoft package gets added as transitive dependency then the build will pass.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: Added a check for packages.config and also added test results for it in the description.
Haven't added a version check, if the required version is below expected (10.0.2) version which SDK needs then it fails at runtime.

Also added a `DisableNewtonsoftJsonCheck`, if client sets this property as true then their solution will pass the build but in case they are missing newtonsoft even at runtime then they will get a runtime error.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -53,5 +53,46 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
   </ItemGroup>
+
+
+  <!-- Add check for Newtonsoft.Json package in consumer app.  -->
+  <Target Name="CheckNewtonsoftJsonPresence" BeforeTargets="CoreCompile" Condition="'$(ExcludeRestorePackageImports)' != 'true'">
+  
+    <!-- Allow consumers to disable this check -->
+    <PropertyGroup>
+        <DisableNewtonsoftJsonCheck Condition="'$(DisableNewtonsoftJsonCheck)' == ''">false</DisableNewtonsoftJsonCheck>
+    </PropertyGroup>
```
eerhardt, line 64: 1. You don't need to define this property to be able to check it.
2. You can skip this whole Target if the property is set to `true`. So if you change the above 6 lines of code to just:

```xml
<Target Name="CheckNewtonsoftJsonPresence" BeforeTargets="CoreCompile" Condition="'$(ExcludeRestorePackageImports)' != 'true' AND '$(DisableNewtonsoftJsonCheck)' != 'true'">
```

You can make this code simpler and it will have better performance for builds that disable the check.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -53,5 +53,46 @@ Copyright (C) Microsoft Corporation. All rights reserved.
           <Visible>False</Visible>
       </ContentWithTargetPath>
   </ItemGroup>
+
+
+  <!-- Add check for Newtonsoft.Json package in consumer app.  -->
+  <Target Name="CheckNewtonsoftJsonPresence" BeforeTargets="CoreCompile" Condition="'$(ExcludeRestorePackageImports)' != 'true'">
```
Pilchie, line 59: Is `BeforeTargets="CoreCompile"` sufficient, or do we need to ensure that it is after the resolve references targets?

Pilchie (commented):
Agree about putting the condition directly on the whole <Target>.

Also, have we verified that this works properly transitively?  Specifically, if a library like the Aspire integration package references the Cosmos package, but disables the check to pass the buck on to the consumer, does the consumer correctly get the error?

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -53,5 +53,46 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <ItemGroup>
+        <NewtonsoftJsonPackageReference Include="@(PackageReference)" Condition="'%(PackageReference.Identity)' == 'Newtonsoft.Json'" />
+    </ItemGroup>
+    
+    <!-- Check for assembly references to Newtonsoft.Json (including transitive dependencies) -->
+    <ItemGroup>
+        <NewtonsoftJsonAssemblyReference Include="@(ReferencePath)" Condition="'%(FileName)' == 'Newtonsoft.Json'" />
+    </ItemGroup>
     
+    <!-- Check for Newtonsoft.Json in packages.config -->
```
ericstj, line 76: I think your check of `ReferencePath` is sufficient here.  You shouldn't need to parse `packages.config`  - that's just used by NuGet to download the package.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
ericstj: > Haven't added a version check, if the required version is below expected (10.0.2) version which SDK needs then it fails at runtime.

Fair enough, sounds like you've scope this out.  It is possible to do that check - using the version on the package reference, or the `FusionName` metadata on `ReferencePath` as I shared in the sample - but its challenging to get it right.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
```diff
@@ -49,6 +49,7 @@
      <PackageReference Include="Azure.Identity" Version="1.11.4" />
      <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
      <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
+     <PackageReference Include="Newtonsoft.Json" Version="10.0.2" NoWarn="NU1903" PrivateAssets="All" />
```
kirankumarkolli, line 52: Is build action triggered even in-case of transitive dependency?

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
kirankumarkolli: Also please move the property to outside Target?

Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
aavasthy: Encryption.Custom project directly uses Newtonsoft. Initially newtonsoft was being provided by Azure.Cosmos project but after setting private assets to true in Azure.Cosmos project, Encryption.Custom project gets a build failure if we don't directly refer it.

eerhardt:
> Also, have we verified that this works properly transitively?

That's a great point. The .targets file needs to be included/imported from the `buildTransitive` folder in the nuget package. See https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: Yes, using BeforeTargets="CoreCompile" should be sufficient as by this stage in the MSBuild process, the ResolveAssemblyReferences target should have already run, so @(ReferencePath) is fully populated with all resolved assembly references—including transitive dependencies from both PackageReference and packages.config.

Pilchie: If it runs *immediately* before `CoreCompile` true, but I don't know exactly how MSBuild's ordering works.  Is it free to run it way earlier as long as it is sometime before `CoreCompile`?

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: updated the code with suggested changes.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: Cleaned up the code with suggested changes.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
kirankumarkolli: For customers of this package with `PrivateAssets` Newtonsoft won't be part of the package dependency. 
So if they don't explicitly include Newtonsoft will it be build issue OR runtime issue?

aavasthy: Updated the code and added package path as buildTransitive. Did tests with Microsoft.Azure.Cosmos.Encryption package and added test results.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
```diff
@@ -144,7 +145,7 @@
 		<None Include="$(NugetPackageRoot)\Microsoft.HybridRow\$(HybridRowVersion)\lib\netstandard2.0\Microsoft.Azure.Cosmos.Core.dll" Pack="true" IsAssembly="true" PackagePath="lib\netstandard2.0" />
 		<None Include="$(NugetPackageRoot)\Microsoft.HybridRow\$(HybridRowVersion)\lib\netstandard2.0\Microsoft.Azure.Cosmos.Serialization.HybridRow.dll" Pack="true" IsAssembly="true" PackagePath="lib\netstandard2.0" />
 		<None Include="$(NugetPackageRoot)\Microsoft.Azure.Cosmos.Direct\$(DirectVersion)\lib\netstandard2.0\Microsoft.Azure.Cosmos.Direct.dll" Pack="true" IsAssembly="true" PackagePath="lib\netstandard2.0" />
-		<None Include="$(MSBuildThisFileDirectory)\Microsoft.Azure.Cosmos.targets" Pack="true" PackagePath="build\netstandard2.0">
+		<None Include="$(MSBuildThisFileDirectory)\Microsoft.Azure.Cosmos.targets" Pack="true" PackagePath="buildTransitive\">
```
eerhardt, line 148: Note that not including the `netstandard2.0` TFM in the buildTransitive directory means that NuGet will no longer warn/disallow when installing this package on a project that doesn't support `netstandard2.0` anymore. This is probably less of a concern as in the past, but still something to be aware of. We may want to keep the `netstandard2.0` directory under `buildTransitive`.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -15,43 +15,63 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      <ContentWithTargetPath Include="$(MSBuildThisFileDirectory)..\runtimes\win-x64\native\vcruntime140_1.dll">
           <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
           <TargetPath>vcruntime140_1.dll</TargetPath>
           <Visible>False</Visible>
       </ContentWithTargetPath>
   </ItemGroup>
+
+
+  <!-- Add check for Newtonsoft.Json package in consumer app.  -->
+    <Target Name="CheckNewtonsoftJsonPresence" BeforeTargets="CoreCompile" Condition="'$(ExcludeRestorePackageImports)' != 'true' AND '$(DisableNewtonsoftJsonCheck)' != 'true'">
```
eerhardt, line 59: Since MSBuild properties are global, you may want to consider adding a prefix like "AzureCosmos" to the `DisableNewtonsoftJsonCheck` property name.

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
```diff
@@ -15,43 +15,63 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     
   <ItemGroup>
-    <ContentWithTargetPath Include="$(MSBuildThisFileDirectory)..\..\runtimes\win-x64\native\Microsoft.Azure.Cosmos.ServiceInterop.dll">
+    <ContentWithTargetPath Include="$(MSBuildThisFileDirectory)..\runtimes\win-x64\native\Microsoft.Azure.Cosmos.ServiceInterop.dll">
```
eerhardt, line 18: Should these be reverted now that the targets file is back under `buildTranstitive\netstandard2.0`?

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: Updated the property name to AzureCosmosDisableNewtonsoftJsonCheck.

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj
aavasthy: Ended up including two PackagePath
PackagePath="buildTransitive\netstandard2.0" : For transitive dependency in projects Microsoft.Azure.Cosmos.Encryption package which use package

PackagePath="build\netstandard2.0" : For dependencies in package.config

Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.targets
aavasthy: Using AfterTargets="ResolveAssemblyReferences", it ensures the target runs immediately after all assembly references are resolved and item groups like @(ReferencePath) are fully populated. This timing of this will help minimize the risk of any other targets file modifying the references before CoreCompil.

kirankumarkolli (approved):

Pilchie (approved):

roji:
What would be the guidance here for a layer on top of the Cosmos SDK, e.g. the EF provider for Cosmos? Following the logic in this issue, the EF provider shouldn't reference a specific version of Newtonsoft.JSON for the exact same reason as Microsoft.Azure.Cosmos shouldn't. But if I understand correctly, only the project directly referencing Microsoft.Azure.Cosmos gets the build error about the lack of Newtonsoft.JSON, so that wouldn't work...

In other words, it seems like the technique here doesn't work well for transitive scenarios where Microsoft.Azure.Cosmos isn't referenced directly by the user's end application, but rather via another layer.

eerhardt:
> What would be the guidance here for a layer on top of the Cosmos SDK, e.g. the EF provider for Cosmos? 

The .NET Aspire CosmosDB integration fits into this. My plan for it is to set `AzureCosmosDisableNewtonsoftJsonCheck=true` in Aspire.Microsoft.Azure.CosmosDB's .csproj.

> But if I understand correctly, only the project directly referencing Microsoft.Azure.Cosmos gets the build error about the lack of Newtonsoft.JSON, so that wouldn't work...

This isn't correct. Since the MSBuild logic is going into the `buildTransitive` folder, it will apply for any project that references Microsoft.Azure.Cosmos (directly or transitively).

roji:
> This isn't correct. Since the MSBuild logic is going into the buildTransitive folder, it will apply for any project that references Microsoft.Azure.Cosmos (directly or transitively).

I see, thanks - I missed that part. So the EF provider will be able to do the same then.

ElisaPijnAvanade:
I'm having my own internal library using Microsoft.Azure.Cosmos and a different project relying on my internal library.
In my project I'm not using Newtonsoft.Json but using System.Text.Json. So I don't want to install Newtonsoft.Json on my own library as I don't need it.
Do I understand it correctly that on my internal library I can set the AzureCosmosDisableNewtonsoftJsonCheck=true and that way I can still use the Microsoft.Azure.Cosmos, as Microsoft.Azure.Cosmos will handle the dependency internally?
Is it true that I need to set this AzureCosmosDisableNewtonsoftJsonCheck=true on all the projects that are relying on my internal library?

Pilchie:
> Do I understand it correctly that on my internal library I can set the AzureCosmosDisableNewtonsoftJsonCheck=true and that way I can still use the Microsoft.Azure.Cosmos, as Microsoft.Azure.Cosmos will handle the dependency internally?

No, this is not correct.  Someone, somewhere still needs to include a reference to Newtonsoft.Json, and it needs to be deployed with your applications.  Microsoft.Azure.Cosmos still depends on it internally, but we want to get out of the business of being the package that dictates the version. So, you have two choices:

1. Either your library adds the reference.
2. You add the `AzureCosmosDisableNewtonsoftJsonCheck=true` and pass that responsibility on to consumers of your library.

PR #4847, opened 2024-10-24
[Client encryption]: Drops RentArrayBufferWriter

JanHyka:
# Pull Request Template

## Description

This PR drops RentArrayBufferWriter and uses RecyclableMemoryStream 3.0.1 instead (it has both Stream and IBufferWriter<byte> functionality so it's pretty much ideal for our use case). 

To be processed after #4846 

## Type of change

## Closing issues

Contributes to #4678


Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
```diff
@@ -43,6 +43,7 @@
 
   <ItemGroup Condition=" '$(IsPreview)' == 'True' ">
     <PackageReference Include="Microsoft.Data.Encryption.Cryptography" Version="2.0.0-pre007" />
+    <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.1" />
```
kirankumarkolli, line 46: RecyclableMemoryStream is outside of Azure core has allowed list of dependencies. 
One options is to inline the code as Jan has before.

https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-dependencies

JanHyka: I can see where that requirement comes from... to avoid dependency hell. The question is whether avoiding one evil doesn't invite much greater one instead:
1. Copypasted code leads to bloating of code base and rises maintenance costs
2. Nobody maintains it or keeps it up to date (or rather, nobody can afford to)
3. Security issues within original code will be entirely missed in such copies. While affected library gets marked by security tools (or .NET9 analyzers that catch those already at compile), hidden copy will go unnoticed... until it gets exploited.

I am not saying referencing some random 3rd party library would be a good idea, otoh entirely blacklisting pretty much everything including Microsoft.* doesn't seem neither safe nor future proof either. Might be worth consideration.

PR #4854, opened 2024-10-25
Open Telemetry: Adds open telemetry based versioning

sourabh1007:
## Description

Introducing versioning for open telemetry attributes, to make we are always backward compatible.

`database` => Have latest stable/experimental attributes
`database/dup` => have latest and old stable/experimental attributes
default => old attributes. in our case, it is Appinsights Compatible attributes
`appinsightssdk` => reserved for future, when we GA open telemetry attributes, then by default open telemetry compatible attributes will be emitted but customer would be able to get backward compatibility, if they set this env variable.

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/TracesStabilityFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+namespace Microsoft.Azure.Cosmos.Telemetry.OpenTelemetry
+{
+    using System;
+    using global::Azure.Core;
+
+    /// <summary>
+    /// Factory for handling telemetry trace stability modes, allowing attribute settings
+    /// based on environment-specified stability mode configurations.
+    /// </summary>
+    internal class TracesStabilityFactory
```
FabianMeiswinkel, line 14: I get what you want to do here - but why not use subclasses. Like have a CosmosTracer base class (withonly private ctor) and a static factory methode that will create the right subclass (each implementing an abstract populateXXX method). That avoids the need to change signatures when a third version is added and seems more readbale/intuitive than passing around actions?

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryStablityModes.cs
```diff
@@ -18,5 +18,10 @@ internal sealed class OpenTelemetryStablityModes
         ///  emit both the old and the stable database conventions, allowing for a seamless transition.
         /// </summary>
         public const string DatabaseDupe = "database/dup";
+
```
FabianMeiswinkel, line 21: I still can't say i understasnd/like the version identifiers. That put aside - the docs should refer to a place where the difference (schema of events) is clearly spelled out for the different versions. Can be a separate md file just referenced here.

FabianMeiswinkel (changes requested):
LGTM overall but left two comments.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryStablityModes.cs
FabianMeiswinkel: Yes - given the version identifier is completely unintuitive and nothing a customer could reason about, we should at least refer to a document showing what the difference in schema would be for each version (ideally also perf difference if overhead is significantly different) - that would allow customer to reason about this configuration. You know how often customers do not look at public documentation - even fewer customers will look at any Open Telemetry semantic profile.

FabianMeiswinkel (approved):
LGTM - Thanks!

kirankumarkolli (approved):
Pending perf validation

sourabh1007:
> Pending perf validation

I will cover it is next PR

PR #4855, opened 2024-10-25
Query: Adds asynchronous enumeration

### LINKED ISSUE #903: Add support for IAsyncEnumerable<T>

seesharprun:
## Description

I've always wanted to use the new `await foreach` and this PR is an attempt to implement the `IAsyncEnumerable<>` interface in two places:

- As an interface for `FeedIterator<>` to give consumers another option for querying entities

    ```csharp
    QueryDefinition query = new QueryDefinition(
        query: "SELECT * FROM items i WHERE i.pk = @pk ORDER BY i.cost"
    ).WithParameter("@pk", "<partition-key>");    

    FeedIterator<Item> feedIterator = this.Container.GetItemQueryIterator<Item>(query);

    IAsyncEnumerable<FeedResponse<Item>> pages = feedIterator.AsAyncEnumerable();
    ```  

- As an extension method `AsAsyncEnumerable()` for `CosmosLinqQuery<>` queries

    ```csharp
    IAsyncEnumerable<FeedResponse<Item>> pages = this.Container.GetItemLinqQueryable<Item>()
        .Where(i => i.pk == "<partition-key>")
        .OrderBy(i => i.cost)
        .AsAsyncEnumerable();
    ```

This is a pretty major lift, so I anticipate some major changes and opinionated reviews. This is based on @onionhammer work in #4355. I've implemented a few of the items of feedback from the engineering team:

- First, I made sure you can get an asynchronous iterator from either the "LINQ" or "NoSQL query" options for a `Container` instance.
- Second, I added unit and emulator tests for both options. I was unable to mock the `CosmosLinqQuery<>` type because it was sealed, so I would appreciate guidance on unit testing for that scenario.
- Finally, I returned an `IAsyncEnumerable<FeedResponse<>>`. This balances between the convenience of `await foreach` and the depth of reading response metadata/headers. Here's an example query:

    ```csharp
    double requestCharge = 0.0;
    List<Item> results = new();
    await foreach(var page in pages)
    {
        requestCharge += page.RequestCharge;
        foreach (var item in page)
        {
            items.Add(item);
        }
    }
    ```

I also tried to be as verbose as possible with the XML comments so they can generate into the documentation with examples.

## Type of change

## Closing issues

- Closes #903


seesharprun:
It looks like these two tests are failing in the preview run:

- `EnumIsPreservedAsINTest`
- `PreviewContractChanges`

seesharprun:
I'm not sure why the preview tests are failing

seesharprun:
Is it possible to get a review on this PR? @jcocchi

onionhammer:
Hmm, what happened with this?

PR #4866, opened 2024-11-01
[Internal] JSON Binary Encoding: Adds support for encoding uniform arrays

sboshra:
## Description

Added full end-to-end support for writing and reading binary-encoded uniform number arrays, as well as nested arrays of uniform number arrays.

**Uniform Number Arrays**
A uniform number array is a JSON array where all items share the same numeric type. The encoding supports the following numeric types:
- **Int8**: Signed 1-byte integer (-128 to 127)
- **UInt8**: Unsigned 1-byte integer (0 to 255)
- **Int16**: Signed 2-byte integer
- **Int32**: Signed 4-byte integer
- **Int64**: Signed 8-byte integer
- **Float16**: 2-byte floating-point value (currently unsupported)
- **Float32**: 4-byte floating-point value
- **Float64**: 8-byte floating-point value

Uniform number arrays are represented by these new type markers:
- **ArrNumC1**: Uniform number array with a 1-byte item count
- **ArrNumC2**: Uniform number array with a 2-byte item count

Both type markers are encoded as follows:
`| Type marker | Item type marker | Item count |`

To maintain backward compatibility, writing uniform number arrays is controlled via the `EnableNumberArrays `write option. When enabled, at the end of writing an array, the writer checks if all values are numeric. It identifies the smallest numeric type that fits all values and compares the length of the uniform number array to the regular array. If the new length is less than or equal to the old one, the array is converted to a uniform number array.

**Arrays of Uniform Number Arrays**
This encoding enhancement allows for encoding multiple uniform number arrays with the same underlying numeric type and item count into a single contiguous array of numbers. The items in all arrays are preceded by a prefix indicating the common array encoding and the number of encoded arrays.

Arrays of uniform number arrays are supported by these two new type-markers:

- **ArrArrNumC1C1**: Array of 1-byte item count of common uniform number arrays with 1-byte item count.
- **ArrArrNumC2C2**: Array of 2-byte item count of common uniform number arrays with 2-byte item count.

Both new values are encoded as follows:
`| Type-marker | Array type-marker | Number type-marker | Number item count | Array item count |`

Similar to uniform number arrays, the writing of arrays of uniform number arrays is conditional on the `EnableNumberArrays` write option being specified. This ensures backward compatibility with readers and navigators that do not yet support this encoding.

**JSON Serialization Testing**

- Introduced a new set of tests for both uniform number arrays and nested arrays of uniform number arrays.
- Enhanced the `JsonToken` class to support representation of uniform number array tokens.
- Updated `JsonWriterTest` to include additional validation. This now not only checks the expected output but also verifies round-trip consistency across different formats and write options for all three rewrite scenarios: JSON Navigator, JSON Reader - Write All, and JSON Reader - Write Current Token.

## Type of change


Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.Enumerator.cs
```diff
@@ -13,57 +13,72 @@ internal static partial class JsonBinaryEncoding
...
-                // Seek to the first array item
-                buffer = buffer.Slice(firstArrayItemOffset);
-
-                while (buffer.Length != 0)
+                if (uniformArrayInfo != null)
                 {
-                    int arrayItemLength = JsonBinaryEncoding.GetValueLength(buffer.Span);
-                    if (arrayItemLength > buffer.Length)
+                    int itemStartOffset = arrayOffset + uniformArrayInfo.PrefixSize;
+                    for (int i = 0; i < uniformArrayInfo.ItemCount; i++)
```
neildsh, line 37: >i [](http://example.com/codeflow?start=29&length=1)

[Nit][Coding Convention] please consider renaming to index.

Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.Enumerator.cs
```diff
@@ -13,57 +13,72 @@ internal static partial class JsonBinaryEncoding
...
-                    }
+                        int arrayItemLength = JsonBinaryEncoding.GetValueLength(buffer.Span);
+                        if (arrayItemLength > buffer.Length)
+                        {
+                            // Array Item got cut off.
+                            throw new JsonInvalidTokenException();
+                        }
 
-                    yield return segment;
+                        yield return new ArrayItem(arrayOffset + (arrayLength - buffer.Length), default);
```
neildsh, line 67: >default [](http://example.com/codeflow?start=96&length=7)

[Nit] Should this be null for readability? I think we tend to use null for classes and default for structs.

Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.Enumerator.cs
sboshra: Switched to iterating over item offsets and renamed it to offset

Microsoft.Azure.Cosmos/src/Json/JsonWriter.JsonBinaryWriter.cs
```diff
@@ -443,6 +464,237 @@ public override void WriteBinaryValue(ReadOnlySpan<byte> value)
                 this.bufferedContexts.Peek().Count++;
             }
 
+            /// <inheritdoc />
+            public override void WriteNumberArray(IReadOnlyList<byte> values)
```
neildsh, line 468: >public override void WriteNumberArray(IReadOnlyList<byte> values) [](http://example.com/codeflow?start=12&length=65)

perhaps there is value in consolidating all of these WriteNumberArray methods into a single method:
`void WriteNumberArray<T>(IReadOnlyList<T> values, Typemarker typeMarker)`

Microsoft.Azure.Cosmos/src/Json/JsonWriter.cs
```diff
@@ -143,6 +145,104 @@ public virtual void WriteStringValue(string value)
         /// <inheritdoc />
         public abstract void WriteBinaryValue(ReadOnlySpan<byte> value);
 
+        /// <inheritdoc />
+        public virtual void WriteNumberArray(IReadOnlyList<byte> values)
```
neildsh, line 149: > void WriteNumberArray(IReadOnlyList<byte> values) [](http://example.com/codeflow?start=22&length=50)

Please consider consolidating all of these into a single method `void WriteNumberArray<T>(IReadOnlyList<T> values)`

Microsoft.Azure.Cosmos/src/Json/JsonReader.JsonBinaryReader.cs
```diff
@@ -272,31 +275,60 @@ public override bool Read()
...
+                            tokenType = JsonTokenType.FieldName;
+                        }
+                    }
+                    else if ((tokenType == JsonTokenType.BeginArray) || (tokenType == JsonTokenType.BeginObject))
+                    {
+                        // If we are currently within a nested array, then the BeginArray token must be for
+                        // a uniform number array that is within a uniform array of number arrays.
+                        if (this.arrayAndObjectEndStack.IsWithinUniformArray())
+                        {
+                            // ASSERT(tokenType == JsonTokenType.BeginArray);
```
neildsh, line 315: >ASSERT [](http://example.com/codeflow?start=31&length=6)

Should this be an explicit `Debug.Assert` ?

Microsoft.Azure.Cosmos/src/Json/JsonWriter.JsonBinaryWriter.cs
```diff
@@ -715,6 +991,277 @@ private void WriteArrayOrObjectEnd(bool isArray)
...
+                    }
+                    else
+                    {
+                        throw new JsonUnexpectedTokenException();
+                    }
+
+                    arrayBuffer = arrayBuffer.Slice(itemLength);
+                }
+
+                // Assert(numberValues.Count == itemCount);
```
neildsh, line 1047: >Assert [](http://example.com/codeflow?start=19&length=6)

Should these be explicit `Debug.Assert`s ?

neildsh (approved):

neildsh:
In a future change, it may be good to consider adding some emulator equivalence tests that utilize these code paths. We can use supported serialization formats to get both binary and text responses and then transcribe the binary to text using the sdk and compare with the backend text response as a baseline.

Microsoft.Azure.Cosmos/src/Json/JsonWriter.JsonBinaryWriter.cs
sboshra: Unlike C++, I don't think this would work here given the limited types supported by binaryWriter.Write functions

Microsoft.Azure.Cosmos/src/Json/JsonWriter.cs
sboshra: This would require that to be able to cast any type T to Number64, which is not the case.

This would only work if we're able to constrain T to a specific set of types - is this possible?

Microsoft.Azure.Cosmos/src/Json/JsonReader.JsonBinaryReader.cs
sboshra: We don't use any asserts in this code. I remember this was decided to avoid popups while executing tests which halt the execution.

Microsoft.Azure.Cosmos/src/Json/JsonReader.JsonBinaryReader.cs
```diff
@@ -313,7 +345,8 @@ public override Number64 GetNumberValue()
                 }
 
                 return JsonBinaryEncoding.GetNumberValue(
-                    this.jsonBinaryBuffer.GetBufferedRawJsonToken(this.currentTokenPosition).Span);
+                    this.jsonBinaryBuffer.GetBufferedRawJsonToken(this.currentTokenPosition).Span,
+                    this.arrayAndObjectEndStack.GetUniformArrayInfo());
```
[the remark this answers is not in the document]
sboshra: I believe this is the correct indentation (the second argument to GetNumberValue function.

sboshra:
I'm not too sure if we need to rely on the emulator for that. All the new tests I’ve added are direct ports from the backend, which include the expected binary output.

In addition, each implementation should be self-sufficient when it comes to verification as long as it's able to verify round-tripping between different formats. This is true for both SDK and backend tests. 

---
In reply to: [2453253955](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4866#issuecomment-2453253955) [](http://example.com/codeflow?ancestors=2453253955)

neildsh (approved):

adityasa:
Does it make sense to put these changes behind a flag to minimize issues (since binary is on by default)?

Microsoft.Azure.Cosmos/src/Json/JsonWriter.JsonBinaryWriter.cs
sboshra: This code does not make use of debug asserts to avoid popups when running the tests

sboshra:
For the JsonBinaryWriter, the EnableUniformNumberArray option must be specified to enable writing uniform number arrays; without this, the behavior remains unchanged. For JsonReader and JsonNavigator, there is no option to disable the reading of uniform number arrays, and I don't believe such a feature is needed. I have significantly expanded the test coverage, including comprehensive round-trip testing, to ensure confidence in implementing these changes.

---
In reply to: [2455776367](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4866#issuecomment-2455776367) [](http://example.com/codeflow?ancestors=2455776367)

adityasa (dismissed):
:shipit:

adityasa (dismissed):
:shipit:

adityasa (approved):

neildsh (approved):

PR #4867, opened 2024-11-01
FaultInjection: Adds method to add FaultInjection using CosmosClientBuilder

NaluTripician:
# Pull Request Template

## Description

Adds a method to use the cosmos client builder to inject faults into the Cosmos DB client operations.

## Type of change

- 
## Closing issues


FabianMeiswinkel (approved):
LGTM - Thanks!

Microsoft.Azure.Cosmos/FaultInjection/src/FaultInjector.cs
```diff
@@ -48,5 +48,10 @@ internal IChaosInterceptorFactory GetChaosInterceptorFactory()
         {
             return this.chaosInterceptorFactory;
         }
+
+        IChaosInterceptorFactory IFaultInjector.GetChaosInterceptorFactory()
```
ananth7592, line 52: Wouldn't it be better for the project to use an `IFaultInjector` interface to access the `IChaosInterceptorFactory` instance? In what situations would directly accessing the `FaultInjector` class internally (without the `IFaultInjector` interface) be more advantageous?

NaluTripician: In the main SDK we define the `IFaultInjector` interface so that the client can accept the needed classes. We then will implement this class in the FaultInjection Library so we can keep the two separate. The `FaultInjector` is an object that the CX interacts with and creates that serves multiple purposes (like getting Fault Injecton Rule application context). Internally it will give the chaos interceptor factory to the client which will give it the information and internal classes that are needed to create the ChaosInterceptor which the CX does not see and it passed all the way through the RNTBD layer

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -898,6 +900,22 @@ internal Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> GetServerCerti
         /// </summary>
         public CosmosClientTelemetryOptions CosmosClientTelemetryOptions { get; set; }
 
+        /// <summary>
+        /// Create a client with Fault Injection capabilities.
```
kirankumarkolli, line 904: Please add a code sample on how to use this API.

Microsoft.Azure.Cosmos/src/FaultInjection/IFaultInjector.cs
```diff
@@ -0,0 +1,16 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.FaultInjection
+{
+    /// <summary>  
+    /// Interface for injecting faults into the Cosmos DB client operations.  
+    /// </summary>  
+    public interface IFaultInjector
```
kirankumarkolli, line 9: In theory its more ChaosInjector, any alternative naming suggestions? 
Am fine as long as we are consistent with JAVA.

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/FaultInjection/IFaultInjector.cs
NaluTripician: Java uses something a bit different: `CosmosFaultInjectionHelper`, but that class functions a bit differently from ours.

Microsoft.Azure.Cosmos/FaultInjection/src/FaultInjector.cs
ananth7592: Thanks for the detailed explanation. Approving it.

ananth7592 (approved):

PR #4875, opened 2024-11-08
[Internal] Query:  Fixes and enables hybrid search tests and configures pipeline to use internal emulator.

sc978345:
## Description
This PR modified the emulator setup to use the internal emulator build, and sets queryEnableFullText to true so that we can test full text search

It also enables some full-text-search-related tests


kundadebdatta (dismissed):
LGTM!

adityasa:
It's not clear why following tests were marked as ignore in Debdatta's earlier pr (I don't see Full-Text dependency but I may be missing something). Can you please double check? @kundadebdatta in case he knows:
CosmosContainerSettingsTests.DefaultIndexingPolicySameAsDocumentCollection
SettingsContractTests.ContainerV2CompatTest

adityasa (dismissed):
:shipit:

neildsh (approved):

templates/emulator-setup.yml
```diff
@@ -2,8 +2,8 @@
 
 steps:
   - pwsh: |
-      Write-Host "Downloading Cosmos Emulator - $env:EMULATORMSIURL" -ForegroundColor green
```
sourabh1007, line 5: is it temporary change or permanent? and why are we not configuring it, using environment variable?

kirankumarkolli: Temporary and will be reverted in future. 
This change was moved to a different PR which is merged

PR #4879, opened 2024-11-09
[INTERNAL] Emulator: Fixes upgrades to internal emulator

kirankumarkolli:
Emulator: Fixes upgrades to internal emulator

Observations: 
- FFCF/AVAD: Emulator defaults resulting in 400/back request => Unblocking with `enablePreviousImageForDeleteInFFCF` override
- ReadFeed with Binary now service returning in binary format (related tests upated)
- Misc escaping changes
- Emulator: Startup times seems regressed (will follow-up later)

On FFCF part: There are two tests
Updates -> Drains changes -> Deletes -> ChangeFeed => 400
Update + Deletes -> Changefeed -> Succeeded with 200 but no previous image
 
Included `includeIdAndPKForDeletesInFFCFResponse` override working for both scenarios.


Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -188,7 +188,7 @@ internal static async Task<DocumentClientException> CreateDocumentClientExceptio
                 {
                     foreach (KeyValuePair<string, IEnumerable<string>> header in requestMessage.Headers)
                     {
-                        contextBuilder.AppendLine($"Header: {header.Key} Length: {string.Join(",", header.Value).Length};");
```
kirankumarkolli, line 191: Undo before commit.

sc978345 (approved):

PR #4883, opened 2024-11-11
[Internal] JSON Binary Encoding: Adds support for writing unsigned 64-bit integer values

sboshra:
## Description

This PR adds support for serializing unsigned Int64 values in Binary encoding, ensuring that values exceeding the maximum signed Int64 limit are preserved without conversion to floating-point doubles. This enhancement enables full-fidelity roundtrips between Text and Binary encoding formats. For values outside the signed Int64 range, retrieval from either the JSON Reader or Navigator (across all encoding formats) will return them as floating-point doubles.

## Type of change

## Closing issues
None


adityasa:
Does this change contain any bugfixes (either by themselves or for previous commit)? Asking to see if it should be included in the release that's in progress.

sboshra:
It mainly adds support of reading/writing UInt64 values, which is soon to be needed once writing such values is enabled in the backend.

---
In reply to: [2468664283](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4883#issuecomment-2468664283) [](http://example.com/codeflow?ancestors=2468664283)

adityasa (approved):
:shipit:

PR #4884, opened 2024-11-11
[Internal] Binary Encoding: Fixes Performance Regression

kundadebdatta:
# Pull Request Template

## Description

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Serializer/CosmosBufferedStreamWrapper.cs
+++ b/Microsoft.Azure.Cosmos/src/Serializer/CosmosBufferedStreamWrapper.cs
@@ -138,7 +138,6 @@ namespace Microsoft.Azure.Cosmos.Serializer
                 this.Flush();
                 if (this.shouldDisposeInnerStream)
                 {
-                    Console.WriteLine($"Dispose inner stream called: {this.innerStream.GetType()}");
                     this.innerStream.Dispose();
                 }
                 else
--- a/Microsoft.Azure.Cosmos/src/Serializer/CosmosBufferedStreamWrapper.cs
+++ b/Microsoft.Azure.Cosmos/src/Serializer/CosmosBufferedStreamWrapper.cs
@@ -5,8 +5,10 @@
 namespace Microsoft.Azure.Cosmos.Serializer
 {
     using System;
+    using System.Diagnostics;
     using System.IO;
-    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
     using Microsoft.Azure.Cosmos.Json;
     using Microsoft.Azure.Documents;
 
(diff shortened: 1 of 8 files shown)


Microsoft.Azure.Cosmos/src/Serializer/CosmosBufferedStreamWrapper.cs
```diff
@@ -129,11 +157,48 @@ protected override void Dispose(bool disposing)
         /// </returns>
         public byte[] ReadAll()
         {
-            ArraySegment<byte> byteSegment = this.innerStream.GetBuffer();
+            int count, totalBytes = 0, offset = (int)this.Position, length = (int)this.Length;
```
FabianMeiswinkel, line 160: NIT - I admit personal opinion - but I find it unreadable when multiple variables are declared in one line.

Microsoft.Azure.Cosmos/src/Serializer/CosmosBufferedStreamWrapper.cs
```diff
@@ -129,11 +157,48 @@ protected override void Dispose(bool disposing)
...
+            byte[] bytes = new byte[length];
+
+            while ((count = this.innerStream.Read(bytes, offset, length - offset)) > 0)
```
FabianMeiswinkel, line 163: The implementation here is functionally incorrect when you first use Stream.Read and subsequently Stream.ReadAll - for example you would copy the first byte again in the buffer for ReadAll - while it has already been processed in that case.

Microsoft.Azure.Cosmos/src/Serializer/CosmosBufferedStreamWrapper.cs
FabianMeiswinkel, line 160: this.Length is not save when the inner stream is not buffered. And the API contract does not ensure that - it would thorw an exception if the inner stream is not buffered.

Microsoft.Azure.Cosmos/src/Serializer/CosmosBufferedStreamWrapper.cs
```diff
@@ -129,11 +157,48 @@ protected override void Dispose(bool disposing)
...
+
+        /// <summary>
+        /// Asynchronously reads all bytes from the current position to the end of the stream.
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous read operation. The value of the TResult parameter contains a byte array with all the bytes read from the stream, or <c>null</c> if no bytes were read.
+        /// </returns>
+        public async Task<byte[]> ReadAllAsync(CancellationToken cancellationToken = default)
+        {
+            int count, totalBytes = 0, offset = (int)this.Position, length = (int)this.Length;
```
FabianMeiswinkel, line 186: See comments in sync PAI. In general I think Stream implementations should implement the async variation and just await it in the sync overload. The perf overhead of doing it even if the asnyc overload completes synchronously is relatively small.

kundadebdatta:
[Follow Up PR](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4901) created to track this work.

PR #4885, opened 2024-11-12
Query: Fixes (workaround) for query plan issue where placeholder index does not start at zero

neildsh:
## Description

During query plan generation, we were using the loop index index over order by expressions as the index for placeholders within component queries. However, VectorDistance expressions do not require placeholders. As a result the output placeholder indexes may not always start at 0. The fix for this is currently in PR on the gateway.

We are putting this workaround in the SDK until the gateway fix is deployed.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
@@ -366,7 +366,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.CrossPartition.HybridSearch
             List<QueryInfo> rewrittenQueryInfos = new List<QueryInfo>(queryInfos.Count);
             foreach (QueryInfo queryInfo in queryInfos)
             {
-                QueryInfo rewrittenQueryInfo = RewriteOrderByQueryInfo(queryInfo, statistics);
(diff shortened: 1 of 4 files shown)


Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
```diff
@@ -665,6 +667,33 @@ private static string FormatComponentQueryText(string format, GlobalFullTextSear
...
+
+            int statisticsIndex = 0;
+            for (int componentIndex = 0; componentIndex < componentCount; ++componentIndex)
+            {
+                string totalWordCountPlaceholder = string.Format(Placeholders.FormattableTotalWordCount, componentIndex);
+                string hitCountsArrayPlaceholder = string.Format(Placeholders.FormattableHitCountsArray, componentIndex);
+
+                if (query.IndexOf(totalWordCountPlaceholder) == -1)
+                {
+                    continue;
```
adityasa, line 682: Consider asserting 
query.IndexOf(hitCountsArrayPlaceHolder) == -1
for readability/code consistency.

adityasa (approved):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
neildsh: This is the workaround for the query plan placeholder index issue, so we need this here.

sc978345 (approved):

kundadebdatta (approved):

PR #4896, opened 2024-11-18
[Client encryption]: Fixes missing files errors at time of nuget publishing.

aavasthy:
# Pull Request Template

## Description

Updated the Microsoft.Azure.Cosmos.Encryption package and  Microsoft.Azure.Cosmos.Encryption.Custom package so as to fix missing icon.png and license.txt errors when publishing these packages locally.

## Type of change
Below the are the errors the project throws while publishing the nuget package.
image
image

After the fix is implemented the nuget package publishes successfully.
image
image

## Closing issues


Microsoft.Azure.Cosmos.Encryption.Custom/src/LICENSE.txt
```diff
@@ -0,0 +1,42 @@
...
+For users of this code who wish to consume it under the "BSD" license
+rather than under the public domain or CC0 contribution text mentioned
+above, the code found under this directory is *also* provided under the
+following license (commonly referred to as the BSD 2-Clause License). This
+license does not detract from the above stated release of the code into
+the public domain, and simply represents an additional license granted by
+the Author.
+
+-----------------------------------------------------------------------------
+** Beginning of "BSD 2-Clause License" text. **
```
Pilchie, line 15: Most Microsoft Open Source code uses the MIT license, as does the main repo?  Why are we using this hybrid of the BSD-2, and CC0?

Also, assuming we can move to MIT, I think you can use a [`PackageLicenseExpression`](https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-a-license-expression-or-a-license-file) instead of file.

Pilchie (changes requested):

Microsoft.Azure.Cosmos.Encryption.Custom/src/LICENSE.txt
aavasthy: I have added the same License.txt that we have in main repo https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/OSS/HdrHistogram/LICENSE.txt
But I can update it to using PackageLicenseExpression if that is recommended.
<PropertyGroup>
    <PackageLicenseExpression>MIT</PackageLicenseExpression>
</PropertyGroup>

Pilchie: That is specific to that subdirectory where we pulled in the source from an external OSS library.  The main license is [MIT](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/LICENSE).

I think we should use PackageLicenseExpression with MIT as the value.

aavasthy: Updated the file reference and using the main license file now and referencing it in the way it is being referenced in the main project (using  PackageLicenseFile).

Microsoft.Azure.Cosmos.Encryption/src/Microsoft.Azure.Cosmos.Encryption.csproj
```diff
@@ -28,11 +27,19 @@
...
     <AdditionalFiles Include="..\..\Microsoft.Azure.Cosmos\src\stylecop.json" Link="stylecop.json" />
   </ItemGroup>
-	
+
   <ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' != 'True' ">
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="[3.35.4,3.37.0]" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Include="icon.png" Pack="true" PackagePath="\" />
```
Pilchie, line 36: Do we have a shared one of these that we can use instead of adding two identical versions?

aavasthy: I updated and kept the file outside in the common so that all three projects can refer it.

NaluTripician (approved):

ananth7592 (approved):
nit: Does the icon.png at line number 30 also needs to be changed to this new path https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj#L30 ?

aavasthy:
> nit: Does the icon.png at line number 30 also needs to be changed to this new path https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj#L30 ?

PackagePath="" places the Icon.png file at the root of the package so that is why the <PackageIcon> value does not need to change in this case.

PR #4899, opened 2024-11-20
Remove unnecessary package dependencies.

teo-tsirpanis:
# Pull Request Template

## Description

This PR removes three package dependencies from the `Microsoft.Azure.Cosmos` project. These are very old packages and not needed when targeting .NET Standard 2.0+; in fact they bring in dozens of unnecessary transitive dependencies and significantly increase the total download size.

## Type of change

## Closing issues


Pilchie:
Note that removing these is a (very small) source breaking change. If some downstream consumer relied on the fact that we brought them in, they will have to update their own projects to reference these instead.

Not saying we shouldn't do it, just that we should document it clearly in release notes, and not do it in a patch release.

NaluTripician:
Hi @teo-tsirpanis, thank you for the contribution! This PR has been idle for ~413 days. Is this change still needed on your side? Flagging for SDK-team triage as part of an open-PR cleanup pass — we'd like to give you a clear yes/no rather than let it sit.

teo-tsirpanis:
Looks like it's already done in #5386. Thanks!

PR #4652, opened 2024-08-23
[Internal] Binary Encoding: Adds Binary Encoding Support for Point Operations

kundadebdatta:
# Pull Request Template

# Description

This PR introduces binary encoding support on request and responses for different Point operations.

## What is Binary Encoding?

As the name suggests, binary encoding is a encoding mechanism through which the request payload will be encoded to binary first and sent to backend for processing. Decoding to Text will happen on the response path. The biggest benefit of binary encoding is to reduce cost on backend storage which helps to reduce the overall COGS.

## Scope

The point operations that are currently in and out of scope for binary encoding are given below in tabular format:

| Operations Currently in Scope| Operations Currently Out of Scope| Reason for Out of Scope|
| --- | --- | --- |
| `CreateItemAsync()` | `PatchItemAsync()` | Operation Currently not Supported in BE
| `CreateItemStreamAsync()` | `PatchItemStreamAsync()` | Operation Currently not Supported in BE
| `ReadItemAsync()` | `TransactionalBatches` | Operation Currently not Supported in BE
| `ReadItemStreamAsync()` | `Bulk APIs` | Operation Currently not Supported in BE
| `UpsertItemAsync()` |  | |
| `UpsertItemStreamAsync()` |  | |
| `RepalceItemAsync()` |  | |
| `ReplaceItemStreamAsync()` |  | |
| `DeleteItemAsync()` |  | |
| `DeleteItemStreamAsync()` |  | |

## How to Enable Binary Encoding?

This PR introduces a new environment variable `AZURE_COSMOS_BINARY_ENCODING_ENABLED` to opt-in or opt-out the binary encoding feature on demand. Setting this environment variable to `True` will enable Binary encoding support.

## How Binary Encoding has been Achieved?

The binary encoding in the .NET SDK has been divided into two parts which are applicable differently for `ItemAsync()` and `ItemStreamAsync()` apis. The details are given below:

- **`ItemAsync()` APIs:** Currently the `CosmosJsonDotNetSerializer` has been refactored to read and write the binary bits directly into the stream. This reduces any conversion of the text stream to binary and vice versa and makes the serialization and de-serialization process even faster.

[...]

PR #4901, opened 2024-11-22
[Internal] Binary Encoding: Fixes Performance Regression

kundadebdatta:
# Pull Request Template

## Description

Recently during our v3 sdk CI rolling runs, we observed some performance regressions on the `ItemStreamAsync()` APIs. They regressed beyond 5%.

image

Upon doing further investigation, we figured out that during the non-binary flow, we end up converting the incoming stream into `CloneableStream` which might be the reason for this regression. Please note that the reason this was not caught during the [original version of the binary encoding PR](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4652) was that the performance test used to capture the benchmark for the original PR, was targeted a real cosmos container, where for the CI runs, we use our mocked containers.

This PR skips `CloneableStream` conversation for non-binary encoding flow.

With the above change in place, our CI builds started passing:

image

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -924,9 +924,14 @@ private async Task<ResponseMessage> ProcessItemStreamAsync(
             string resourceUri = this.GetResourceUri(requestOptions, operationType, itemId);
 
             // Convert Text to Binary Stream.
-            streamPayload = CosmosSerializationUtil.TrySerializeStreamToTargetFormat(
-                targetSerializationFormat: ContainerCore.GetTargetRequestSerializationFormat(),
-                inputStream: streamPayload == null ? null : await StreamExtension.AsClonableStreamAsync(streamPayload));
+            if (ConfigurationManager.IsBinaryEncodingEnabled())
```
Pilchie, line 927: Below (on line 950) it seems to use a different condition to indicate that Binary Serialization should be used.  Should those conditions line up?

kundadebdatta: Thanks for pointing this out @Pilchie! The idea was, in the response path, when we try to convert binary back to text, we should convert unconditionally (expect either - **a)** explicitly requested as binary OR **b)** the response is meant for `ItemAsync()` APIs). That would imply that when we produce a response, the content format is always in **Text**.

This is the reason we did not wanted to wrap that in an `if` condition. However, it is highly unlikely that a binary response would be returned from backend, when binary encoding is not enabled in SDK.

kirankumarkolli (approved):

aavasthy (approved):

PR #4922, opened 2024-12-04
Change feed: Adds id and pk to ChangeFeedMetadata for delete operations

jcocchi:
# Pull Request Template

## Description

Adds id and partition key to ChangeFeedMetadata. This is populated for delete operations using all versions and deletes change feed mode.

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
+++ b/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
@@ -59,13 +59,13 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// The id of the previous item version. Used for delete operations only.
         /// </summary>
-        [JsonProperty(PropertyName = ChangeFeedMetadataFields.DeletedItemId, NullValueHandling = NullValueHandling.Ignore)]
-        public string DeletedItemId { get; internal set; }
+        [JsonProperty(PropertyName = ChangeFeedMetadataFields.Id, NullValueHandling = NullValueHandling.Ignore)]
+        public string Id { get; internal set; }
 
         /// <summary>
         /// The partition key of the previous item version. Dictionary Key is the partition key property name and Dictionary Value is the partition key property value. Used for delete operations only.
         /// </summary>
-        [JsonProperty(PropertyName = ChangeFeedMetadataFields.DeletedItemPartitionKey, NullValueHandling = NullValueHandling.Ignore)]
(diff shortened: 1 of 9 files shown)


Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -50,9 +51,21 @@ class ChangeFeedMetadata
...
+        /// Used to distinguish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
         /// </summary>
         [JsonProperty(PropertyName = ChangeFeedMetadataFields.TimeToLiveExpired, NullValueHandling = NullValueHandling.Ignore)]
         public bool IsTimeToLiveExpired { get; internal set; }
+
+        /// <summary>
+        /// The id of the previous item version. Used for delete operations only.
+        /// </summary>
+        [JsonProperty(PropertyName = ChangeFeedMetadataFields.DeletedItemId, NullValueHandling = NullValueHandling.Ignore)]
+        public string DeletedItemId { get; internal set; }
```
dibahlfi, line 63: I understand that the backend only returns Deleted Item information for now but in the future it will also return update information and so on. Do you think it would be better to name this field as ItemId instead of DeletedItemId? We can update our documentation to reflect this fact.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -50,9 +51,21 @@ class ChangeFeedMetadata
...
+
+        /// <summary>
+        /// The partition key of the previous item version. Dictionary Key is the partition key property name and Dictionary Value is the partition key property value. Used for delete operations only.
+        /// </summary>
+        [JsonProperty(PropertyName = ChangeFeedMetadataFields.DeletedItemPartitionKey, NullValueHandling = NullValueHandling.Ignore)]
+        public Dictionary<string, string> DeletedItemPartitionKey { get; internal set; }
```
dibahlfi, line 69: I would recommend checking with Kiran/Fabien on how the Hierarchical keys are represented with in our codebase so that we are consistent but otherwise I would mark it as Dictionary<string, object> as the value could of any type.

NaluTripician:
Closing — superseded by #5191 (same feature picked up by @dibahlfi). Closing as part of an open-PR cleanup pass.

PR #4925, opened 2024-12-09
Open Telemetry: Adds option to opt-in optional dimensions

sourabh1007:
## Description

Added a new `ClientTelemetryOptions` to configure Operation and Network Level Metrics.

- `OperationMetricsOptions`
    - `IsRegionIncluded` - to include Region Contacted List
    - `CustomDimensions` - Provide flexibility to customer to include app specific dimensions in the metrics

- `NetworkMetricsOptions`
    - `IsRoutingIdIncluded` - to include Routing id (i.e. pkrange or partion/replica info)
    - `CustomDimensions` - Provide flexibility to customer to include app specific dimensions in the metrics

## Type of change


Microsoft.Azure.Cosmos/src/NetworkMetricsOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// should be included in the network metrics. The default value is <c>false</c>.
+        /// </summary>
+        public bool ShouldIncludeRoutingId { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets a collection of custom dimensions to include in the network metrics.
+        /// Each dimension is defined as a key-value pair, where the key is the dimension name and the value is a function that returns the dimension value.
+        /// </summary>
+        [JsonIgnore]
+        public IDictionary<string, Func<string>> CustomDimensions { get; set; }
```
kirankumarkolli, line 27: Why is the value modeled as Fun<>? 
At every operation level application had no control right => practically only the const/static makes sense.

Microsoft.Azure.Cosmos/src/NetworkMetricsOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Represents the configuration options for collecting metrics related to Cosmos DB network operations.
+    /// </summary>
+    public class NetworkMetricsOptions
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether the routing ID (e.g., PK Range ID for Gateway Mode or Partition/Replica information for Direct Mode, if available) 
+        /// should be included in the network metrics. The default value is <c>false</c>.
+        /// </summary>
+        public bool ShouldIncludeRoutingId { get; set; } = false;
```
kirankumarkolli, line 20:
```suggestion
        public bool IncludeRoutingId { get; set; } = false;
```

Microsoft.Azure.Cosmos/src/NetworkMetricsOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Represents the configuration options for collecting metrics related to Cosmos DB network operations.
+    /// </summary>
+    public class NetworkMetricsOptions
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether the routing ID (e.g., PK Range ID for Gateway Mode or Partition/Replica information for Direct Mode, if available) 
+        /// should be included in the network metrics. The default value is <c>false</c>.
```
kirankumarkolli, line 18: Let's please expand on why the default is `false` Or what's the trade-off?

Microsoft.Azure.Cosmos/src/OperationMetricsOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <summary>
+    /// Represents the configuration options for collecting metrics related to Cosmos DB operations.
+    /// </summary>
+    public class OperationMetricsOptions
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether the region information should be included in the operation metrics.
+        /// Default value is <c>false</c>.
+        /// </summary>
+        public bool ShouldIncludeRegion { get; set; } = false;
```
kirankumarkolli, line 20:
```suggestion
        public bool IncludeRegion { get; set; } = false;
```

Microsoft.Azure.Cosmos/src/NetworkMetricsOptions.cs
kirankumarkolli: Let's also please extend on when custom dimensions are needed with an example?

kirankumarkolli, line 27: Are dimensions Operation/Network specific? 
If not do they needs be moved up?

FabianMeiswinkel: I assume the Func is meant to allow customers to change some dimension value over time without restart. This is fine in theory - but a slippery slope because cardinality of values for a dimension have to be small - otherwise any metric system will be overwhelmed by the sheer number of time-series it has to handle. So, I would also start with constant only unless there is a clear scenario already. And if so, the comment here needs to spell out the reasonable constraints on cardinality for dimensions.

sourabh1007: Provided separate settings for Network-level dimensions and Operation-level dimensions. This approach offers customers the flexibility to add custom dimensions specific to each level of metrics.

sourabh1007: Added some examples to justify, `Func<string>` dimension. I have added a documentation for cardinality/cost/complexity issue.

Microsoft.Azure.Cosmos/src/NetworkMetricsOptions.cs
```diff
@@ -6,6 +6,7 @@ namespace Microsoft.Azure.Cosmos
 {
     using System;
     using System.Collections.Generic;
+    using Antlr4.Runtime;
```
FabianMeiswinkel, line 9: Revert?

Microsoft.Azure.Cosmos/src/NetworkMetricsOptions.cs
FabianMeiswinkel: I think teh samples show why using a Func is a bad idea - let's please move this to a IDictionary<string, string>

Microsoft.Azure.Cosmos/src/OperationMetricsOptions.cs
```diff
@@ -17,12 +18,63 @@ public class OperationMetricsOptions
...
+        /// However, it may increase the cardinality of metrics, potentially leading to higher storage costs and generating a large volume of low-sample-count metrics, 
+        /// which can complicate analysis.
+        /// </para>
+        /// <para>
+        /// So, Carefully evaluate whether the additional granularity is necessary 
+        /// for your use case, as it may lead to increased resource consumption and complexity.
+        /// </para>
+        /// </remarks>
         [JsonIgnore]
         public IDictionary<string, Func<string>> CustomDimensions { get; set; }
```
FabianMeiswinkel, line 79: IDictionary<string, string>

FabianMeiswinkel (changes requested):
Please move to static values for custom dimensions

Microsoft.Azure.Cosmos/src/NetworkMetricsOptions.cs
```diff
@@ -16,7 +13,7 @@ public class NetworkMetricsOptions
     {
         /// <summary>
         /// Gets or sets a value indicating whether the routing ID (e.g., PK Range ID for Gateway Mode or Partition/Replica information for Direct Mode, if available)
```
FabianMeiswinkel, line 15: NIT: Leaving comment about default behaviuor (if not specified) is still useful

Microsoft.Azure.Cosmos/src/OperationMetricsOptions.cs
```diff
@@ -31,51 +27,24 @@ public class OperationMetricsOptions
         /// for your use case, as it may lead to increased resource consumption and complexity.
         /// </para>
         /// </remarks>
-        public bool IncludeRegion { get; set; } = false;
+        public bool? IncludeRegion { get; set; }
```
FabianMeiswinkel, line 30: NIT: Brief comment about default behavior if not specified would be still useful.

FabianMeiswinkel (dismissed):
LGTM - Thanks!

Microsoft.Azure.Cosmos/src/OperationMetricsOptions.cs
sourabh1007: Although, current doc should be able to explain the default behavior but as you mentioned, I have included it specifically to make it clearer.

FabianMeiswinkel (approved):
LGTM Thanks

kirankumarkolli (approved):

PR #4939, opened 2024-12-26
Query : Fixes OFFSET, LIMIT, TOP data types to match those coming from query plan

adityasa:
## Query : Fixes OFFSET, LIMIT, TOP data types to match those coming from query plan

This change ensures that the data types of OFFSET, LIMIT and TOP fields on QueryInfo and HybridSearchQueryInfo objects match those coming from the query plan. Currently the queryplan/service uses ULONG which has a range of 0x00000000 to 0xFFFFFFFF. Since QueryInfo fields use int?, this causes an overflow for a value that's larger than 0x7FFFFFFF for any of these fields.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos.sln
+++ b/Microsoft.Azure.Cosmos.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29123.88
+# Visual Studio Version 17
+VisualStudioVersion = 17.11.35431.28
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Cosmos", "Microsoft.Azure.Cosmos\src\Microsoft.Azure.Cosmos.csproj", "{36F6F6A8-CEC8-4261-9948-903495BC3C25}"
 EndProject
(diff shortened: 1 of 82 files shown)


Microsoft.Azure.Cosmos.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29123.88
+# Visual Studio Version 17
```
adityasa, line 3: Revert.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Take/TakeQueryPipelineStage.Client.cs
```diff
@@ -255,11 +255,11 @@ private sealed class LimitContinuationToken : TakeContinuationToken
...
                 /// <param name="limit">The limit to the number of document drained for the remainder of the query.</param>
                 /// <param name="sourceToken">The continuation token for the source component of the query.</param>
-                public LimitContinuationToken(int limit, string sourceToken)
+                public LimitContinuationToken(uint limit, string sourceToken)
                 {
-                    if (limit < 0)
+                    if (limit > int.MaxValue)
                     {
-                        throw new ArgumentException($"{nameof(limit)} must be a non negative number.");
+                        throw new ArgumentNullException(nameof(limit));
```
adityasa, line 262: >ArgumentNullException [](http://example.com/codeflow?start=34&length=21)

ArgumentOutOfRangeException

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/Take/TakeQueryPipelineStage.cs
```diff
@@ -17,14 +17,19 @@ internal abstract partial class TakeQueryPipelineStage : QueryPipelineStageBase
...
             : base(source)
-        {
-            this.takeCount = takeCount;
+        {
+            if (takeCount > int.MaxValue)
+            {
+                throw new ArgumentOutOfRangeException(nameof(takeCount));
+            }
+
+            this.takeCount = (int)takeCount;
```
adityasa, line 28: >takeCount [](http://example.com/codeflow?start=17&length=9)

Maybe store this as uint and remove cast in pipelinestages.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/PipelineFactory.cs
```diff
@@ -108,19 +108,33 @@ public static TryCatch<IQueryPipelineStage> MonadicCreate(
                     maxConcurrency: maxConcurrency);
 
                 if (hybridSearchQueryInfo.Skip != null)
-                {
+                {
+                    if (hybridSearchQueryInfo.Skip.Value > int.MaxValue)
+                    {
+                        throw new ArgumentOutOfRangeException("skipCount");
+                    }
+
```
sboshra, line 116: What guarantees that this would never happen?

adityasa: If offset/limit/top values are out of range, we do throw this exception.
I have a test that captures this:
NegativeQueryTests.TestTopOffsetLimitClientRanges

sboshra (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/PipelineFactory.cs
```diff
@@ -108,19 +108,33 @@ public static TryCatch<IQueryPipelineStage> MonadicCreate(
                     maxConcurrency: maxConcurrency);
 
                 if (hybridSearchQueryInfo.Skip != null)
-                {
+                {
+                    if (hybridSearchQueryInfo.Skip.Value > int.MaxValue)
+                    {
+                        throw new ArgumentOutOfRangeException("skipCount");
+                    }
```
neildsh, line 115: Please consider moving all of these checks to QueryPartitionProvider. That way, we do not need to sprinkle them in all the pipeline stage classes. If need be, we can create clones of the QueryInfo types that can be consumed by the pipelines

neildsh (approved):

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/PipelineFactory.cs
adityasa: Thanks for catching this. Moved the checks inside the QueryPartitionProvider.

neildsh (approved):

sboshra (approved):
:shipit:

PR #4940, opened 2024-12-27
ThroughputBucketing: Adds Throughput Bucket to RequestOptions and ClientOptions

Achint-Agrawal:
## Description

This PR add ThroughputBucket in RequestOptions and ClientOptions to support Throughput Bucketing

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos.sln
+++ b/Microsoft.Azure.Cosmos.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29123.88
+# Visual Studio Version 17
+VisualStudioVersion = 17.12.35514.174 d17.12
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Cosmos", "Microsoft.Azure.Cosmos\src\Microsoft.Azure.Cosmos.csproj", "{36F6F6A8-CEC8-4261-9948-903495BC3C25}"
 EndProject
@@ -180,6 +180,18 @@ Global
 		{021DDC27-02EF-42C4-9A9E-AA600833C2EE}.Release|Any CPU.Build.0 = Release|Any CPU
 		{021DDC27-02EF-42C4-9A9E-AA600833C2EE}.Release|x64.ActiveCfg = Release|Any CPU
 		{021DDC27-02EF-42C4-9A9E-AA600833C2EE}.Release|x64.Build.0 = Release|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Cover|Any CPU.ActiveCfg = Debug|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Cover|Any CPU.Build.0 = Debug|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Cover|x64.ActiveCfg = Debug|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Cover|x64.Build.0 = Debug|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Debug|x64.Build.0 = Debug|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Release|Any CPU.Build.0 = Release|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Release|x64.ActiveCfg = Release|Any CPU
+		{D744906A-1091-403F-B0B6-794DE045169A}.Release|x64.Build.0 = Release|Any CPU
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Cover|Any CPU.ActiveCfg = Debug|Any CPU
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Cover|Any CPU.Build.0 = Debug|Any CPU
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Cover|x64.ActiveCfg = Debug|Any CPU
--- a/Microsoft.Azure.Cosmos.sln
+++ b/Microsoft.Azure.Cosmos.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.12.35514.174 d17.12
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29123.88
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Cosmos", "Microsoft.Azure.Cosmos\src\Microsoft.Azure.Cosmos.csproj", "{36F6F6A8-CEC8-4261-9948-903495BC3C25}"
 EndProject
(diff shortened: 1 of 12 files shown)


Microsoft.Azure.Cosmos.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29123.88
+# Visual Studio Version 17
+VisualStudioVersion = 17.12.35514.174 d17.12
```
Achint-Agrawal, line 4: Revert file

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -940,6 +940,16 @@ public IFaultInjector FaultInjector
...
 
+        /// <summary>
+        /// Sets the throughput bucket for requests created using cosmos client.
+        /// </summary>
+        /// <remarks>
+        /// If throughput bucket is also set at request level in <see cref="RequestOptions.ThroughputBucket"/>, that throughput bucket is used.
+        /// If <see cref="AllowBulkExecution"/> is set to true in CosmosClientOptions, throughput bucket set on the CosmosClient is used.
+        /// </remarks>
+        /// <seealso href="https://aka.ms/cosmsodb-bucketing"/>
+        public int? ThroughputBucket { get; set; }
```
FabianMeiswinkel, line 951: Is an int the only possible way to identify a throughput bucket or can names/identifiers be also used? The int/number seems like it will cause confusion for customers easily.

FabianMeiswinkel: How is this feature rolled-out? If there is any opt-in necessary on the service side in .Net SDK this should only be released in the preview nuget package first - and at least the PR description should have clear instructions for onboarding. If the feature is just enabled automatically for all accounts we can add this as public API assuming we have confidence in the public API in the normal nuget packaage (assuming all deployements including non-public clouds have been finished to enable the backend feature)

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -802,5 +802,21 @@ public CosmosClientBuilder WithClientTelemetryOptions(CosmosClientTelemetryOptio
             this.clientOptions.CosmosClientTelemetryOptions = options;
             return this;
         }
+
+        /// <summary>
+        /// Sets the throughput bucket for requests created using cosmos client.
+        /// </summary>
+        /// <remarks>
+        /// If throughput bucket is also set at request level in <see cref="RequestOptions.ThroughputBucket"/>, that throughput bucket is used.
+        /// If <see cref="WithBulkExecution(bool)"/> is set to true, throughput bucket set on the CosmosClient is used.
```
FabianMeiswinkel, line 811: Why this constraint? That seems super confusing. I get what you are saying - if bulk mode is enabled and request options have different values this is a challenge in .Net - but  why not throw an ArgumentException if request options has throughput bucket defined and bulk mode is enabled - that way the contract is enforced cleanly

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -940,6 +940,16 @@ public IFaultInjector FaultInjector
             }
         }
 
+        /// <summary>
```
FabianMeiswinkel, line 943: Please also add an aka.ms link to explain the concept and constraints of throughput buckets

FabianMeiswinkel (changes requested):
Thanks - overall changes loog good. Mainly blocking on clarification for versioning/rollout/onboarding.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -7093,6 +7093,12 @@ private INameValueCollection GetRequestHeaders(
             {
                 headers.Set(HttpConstants.HttpHeaders.PreserveFullContent, bool.TrueString);
             }
+
+            if (options.ThroughputBucket.HasValue)
+            {
+                headers.Set(HttpConstants.HttpHeaders.ThroughputBucket, options.ThroughputBucket.ToString());
```
Pilchie, line 7099: I know we only support a small number of buckets, but we should really specify a format here to not have something like thousands separators vary by client locale settings.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
kirankumarkolli: +1 to throw and fail.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -513,6 +517,25 @@ private void SetPriorityLevel(RequestMessage requestMessage)
             }
         }
 
+        /// <summary>
+        /// Set the ThroughputBucket in the request headers
+        /// </summary>
+        /// <param name="requestMessage"></param>
+        private void SetThroughputBucket(RequestMessage requestMessage)
```
kirankumarkolli, line 524: [Non blocker] 
Our RequestOptions are stateless but pattern of overrides implementation in RequestInvokerHandler is odd.
One alternative way is to pass the CosmosClientOptions along with ReqeustMessge for RequestOptions.RequestOptions?

/cc: @FabianMeiswinkel thoughts?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: The link added in seealso will contain this information. We are still working on publishing the documentation

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: int is the only way to identify buckets right now. Bucket ids can be between 1 and 5. We might increase it to more buckets if needed.

Achint-Agrawal: The feature is behind a config which is currently off. There is no opt in from the customer end. In the beginning, we will enable the config for a few customers as we receive requests for the feature. Eventually, the config will be turned on by default for all accounts. 

In the past, with Priority Based Execution, we saw a lot of push back from customers for using the preview SDK which impacts the adoption of the feature. If it is possible to do so, it would be great if we could keep the request option in the public SDK. Setting the throughput bucket does not have any impact - if the feature is disabled, the bucket is ignored. No error is thrown.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
Achint-Agrawal: Sure will do this.
Same handling is done for PriorityLevel. Would it be ok to throw an Argument Exception for that as well? There might be existing customers who have not handled this and would start getting an error on upgrading the SDK version.

Achint-Agrawal: Made changes for ThroughputBucket. Keeping open to discuss any changes for PriorityLevel

Microsoft.Azure.Cosmos/src/DocumentClient.cs
Achint-Agrawal: used CultureInfo.InvariantCulture

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
Achint-Agrawal: Changed to only expose this option for preview SDKs

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
Achint-Agrawal: I am assuming this is not specifically related to this PR but how we generally apply request options in RequestInvokerHandler and would be taken up separately if needed. Please let me know if this is not the case. Thanks

FabianMeiswinkel (dismissed):
LGTM - thanks

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -802,5 +802,26 @@ public CosmosClientBuilder WithClientTelemetryOptions(CosmosClientTelemetryOptio
...
+        /// </remarks>
+        /// <param name="throughputBucket">The desired throughput bucket for the client.</param>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
+        /// <seealso href="https://aka.ms/cosmsodb-bucketing"/>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        CosmosClientBuilder WithThroughputBucket(int? throughputBucket)
```
kirankumarkolli, line 821: Is argument needs to be nullable? 
Application can avoid calling this API instead, right?

Achint-Agrawal: Made it non nullable. Thanks

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -513,6 +517,32 @@ private void SetPriorityLevel(RequestMessage requestMessage)
...
+            int? throughputBucket = this.RequestedClientThroughputBucket;
+            RequestOptions promotedRequestOptions = requestMessage.RequestOptions;
+
+            if (promotedRequestOptions?.ThroughputBucket.HasValue == true)
+            {
+                if (this.client.ClientOptions.AllowBulkExecution)
+                {
+                    throw new ArgumentException($"{nameof(requestMessage.RequestOptions.ThroughputBucket)} cannot be set in " +
+                        $"{nameof(requestMessage.RequestOptions)} with {nameof(this.client.ClientOptions.AllowBulkExecution)} set to true. " +
+                        $"Set {nameof(this.client.ClientOptions.ThroughputBucket)} in {nameof(this.client.ClientOptions)} instead.");
```
sourabh1007, line 535: nit: What if throughputbucket is already set at client options also, along with request option?  Rephrase this message to "Set {nameof(this.client.ClientOptions.ThroughputBucket)} in {nameof(this.client.ClientOptions)} instead only" or something in similar lines.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -513,6 +517,32 @@ private void SetPriorityLevel(RequestMessage requestMessage)
             }
         }
 
+        /// <summary>
+        /// Set the ThroughputBucket in the request headers
+        /// </summary>
+        /// <param name="requestMessage"></param>
+        private void SetThroughputBucket(RequestMessage requestMessage)
```
sourabh1007, line 524: nit: rename to "ValidateAndSetThroughputBucket"

sourabh1007 (approved):
LGTM

kirankumarkolli (approved):

PR #4944, opened 2024-12-31
[INTERNAL] Open Telemetry: Adds documentation with details

sourabh1007:
## Description

1. Remove Client Telemetry Documentation
2. Added Open Telemetry Traces/Metrics Documentation
3. Added Traces for recording telemetry

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
+++ b/Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
@@ -532,17 +532,21 @@ namespace Microsoft.Azure.Cosmos
                     // Checks if OpenTelemetry is configured for this operation and either Trace or Metrics are enabled by customer
                     if (isOtelCompatibleOperation)
                     {
-                        // Extracts and records telemetry data from the result of the operation.
-                        OpenTelemetryAttributes otelAttributes = openTelemetry?.GetAttributes(result);
+                        using (ITrace telemetryTrace = trace.StartChild("Recording Telemetry"))
+                        {
+                            // Extracts and records telemetry data from the result of the operation.
+                            OpenTelemetryAttributes otelAttributes = openTelemetry?.GetAttributes(result);
 
-                        // Records the telemetry attributes for Distributed Tracing (if enabled) and Metrics
-                        recorder.Record(otelAttributes);
-                        RecordMetrics(getOperationName,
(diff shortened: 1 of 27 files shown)


Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
```diff
@@ -532,17 +532,21 @@ private async Task<TResult> RunWithDiagnosticsHelperAsync<TResult>(
...
-
-                        // Records the telemetry attributes for Distributed Tracing (if enabled) and Metrics
-                        recorder.Record(otelAttributes);
-                        RecordMetrics(getOperationName,
-                            this.client.Endpoint,
-                            containerName,
-                            databaseName,
-                            requestOptions,
-                            attributes: otelAttributes);
+                        using (ITrace telemetryTrace = trace.StartChild("Recording Telemetry"))
```
FabianMeiswinkel, line 535: isn't this trace way too noisy for the benefit it provides?

FabianMeiswinkel: Saw in the tests that this is consistent - .Net is just emitting extremely noisy traces all over the place. So, not blocking on this.

FabianMeiswinkel (dismissed):
LGTM - thanks

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
kirankumarkolli: Is there value in including otel details in the diagnostics?

sourabh1007: 1. Request diagnostics should include information about telemetry recording to provide a single source of truth for identifying issues or exceptions that occur during the process.
2. If telemetry recording takes longer than expected and causes high latency for the customer, it is crucial to identify this easily. Without this information, we would have to manually track and analyze operations to determine the cause of the delay. With this feature, however, it would be immediately evident that the high latency is due to this specific step.

While I understand that request diagnostics is already verbose—which is a separate issue—it must continue to effectively serve its core purpose.

sourabh1007: As per our offline discussion, .now traces will be added, only if there is any exception

PR #4949, opened 2025-01-06
RequestOptions: Fixes Etag code documentation

kirankumarkolli:
RequestOptions: Fixes Etag code documentation

### DIFF
--- a/Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.Cosmos
         /// <remarks>
         /// Most commonly used with the Delete* and Replace* methods of <see cref="Container"/> such as <see cref="Container.ReplaceItemAsync{T}(T, string, PartitionKey?, ItemRequestOptions, System.Threading.CancellationToken)"/>.
         /// <see cref="Container.CreateItemAsync{T}(T, PartitionKey?, ItemRequestOptions, System.Threading.CancellationToken)"/> will ignore <see cref="IfMatchEtag"/> if specificed. 
+        /// <see cref="Container.UpsertItemAsync{T}(T, PartitionKey?, ItemRequestOptions, System.Threading.CancellationToken)"/> will ignore <see cref="IfMatchEtag"/> when materialized as Create, otherwise for Replace Etag constraint will be applied.
         /// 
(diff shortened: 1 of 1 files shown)


kushagraThapar (commented):
One suggestion: Can we reference this documentation in the .NET API doc section?
https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#implementing-optimistic-concurrency-control-using-etag-and-http-headers

FabianMeiswinkel (dismissed):
LGTM - Thx!

kirankumarkolli:
> One suggestion: Can we reference this documentation in the .NET API doc section? https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#implementing-optimistic-concurrency-control-using-etag-and-http-headers

Thanks @kushagraThapar, updated PR.

kushagraThapar (dismissed):
LGTM, thanks @kirankumarkolli

ananth7592 (approved):

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
```diff
@@ -14,18 +14,28 @@ namespace Microsoft.Azure.Cosmos
     public class RequestOptions
     {
         /// <summary>
-        /// Gets or sets the If-Match (ETag) associated with the request in the Azure Cosmos DB service.
+        /// Gets or sets the If-Match (ETag) associated with the request in the Azure Cosmos DB service.
         /// </summary>
         /// <remarks>
-        /// Most commonly used with the Delete* and Replace* methods of <see cref="Container"/> such as <see cref="Container.ReplaceItemAsync{T}(T, string, PartitionKey?, ItemRequestOptions, System.Threading.CancellationToken)"/>.
+        /// Most commonly used with the Delete* and Replace* methods of <see cref="Container"/> such as <see cref="Container.ReplaceItemAsync{T}(T, string, PartitionKey?, ItemRequestOptions, System.Threading.CancellationToken)"/>.
```
jeet1995, line 20: Do we want to be more explicit for `Upsert` behavior and when `If-Match` header is set? From the offline discussion, looks like `Upsert` has `Replace` semantics when document exists and `Create` semantics when doesn't.

kirankumarkolli: Worth calling it out. Will update PR.

PR #4977, opened 2025-01-22
Heiarchical Partition Keys: Fixes bug for ReadMany where `None` Partition does not return results

### LINKED ISSUE #4946: [ReadMany] does not return results for partial hierarchical partiton keys
**Describe the bug**
Having a container with partition keys `/tenant_id`, `/user_id`
Having an object
``` JSON
{
  "$type": "role",
  "id": "x",
  "tenant_id": "t"
}
```
executing `ReadManyAsync` with partition key `new PartitionKeyBuilder().Add("t").AddNoneType().Build()` reads nothing.
No other combination works too.

**To Reproduce**
Define objects as described above

**Expected behavior**
Read happens successfully and returns objects with partial hierarchical partition keys

**Actual behavior**
Nothing returned.

**Environment summary**
SDK Version: 3.47.0-preview.0

**Additional context**
Both Azure and Emulator works the same

NaluTripician:
# Pull Request Template

## Description

There is currently a bug in the `ReadMany` method where it does not return results for partial hierarchical partition keys when a partition key with a `None` Type is set. 

```
new PartitionKeyBuilder().Add("part1").AddNoneType().Build()
```

## Type of change

## Closing issues

To automatically close an issue: closes #4946


Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -304,6 +304,15 @@ private QueryDefinition CreateReadManyQueryDefinitionForOther(List<(string, Part
                     }
                     for (int j = 0; j < this.partitionKeySelectors.Count; j++)
                     {
+                        if (pkValues[j] == Undefined.Value)
```
kirankumarkolli, line 307: In-general ReadMany is for many point operation reads => each partition-key should be complete/full.

Hierarchical partitions: ReadMany as name might get mis-mapped for HierarchicalParittioning, isn't the ideal solution to use regular query instead?

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -304,6 +304,15 @@ private QueryDefinition CreateReadManyQueryDefinitionForOther(List<(string, Part
...
+                        {
+                            queryStringBuilder.Append(" AND ");
+                            queryStringBuilder.Append("IS_DEFINED(c");
```
kirankumarkolli, line 310: Clarification: Create will not work if any of the partition-keys are not defined right? 
Can we please check this with query team.

kirankumarkolli: Can this clause it-self be excluded?

NaluTripician: Create will be able to work with `null` as the secondary partition key value. The problem here is that in order to do the ReadMany you will need to have a complete partition key. To do this you need to set one of the values as `NoneType` currently, when you do this the resulting query that the `ReadMany` call does results in no results. Doing a normal query where the PK is `NoneType` is ok, however this change was not carried over to when HPK were added

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosMultiHashTest.cs
```diff
@@ -510,5 +510,28 @@ public async Task MultiHashQueryItemTest()
...
+
+        [TestMethod]
+        public async Task ReadManyNullPkValueTest()
+        {
+            Document doc = new Document { Id = "readMany" };
+            doc.SetValue("ZipCode", "10000");
+
+            await this.container.CreateItemAsync<Document>(doc);
+
+            Cosmos.PartitionKey pk = new PartitionKeyBuilder()
```
kirankumarkolli, line 522: All component None supported. 
If so then the query composition need to account for it.

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
NaluTripician: Discussed ofline

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosMultiHashTest.cs
NaluTripician: Added Test coverage

kirankumarkolli (approved):

kundadebdatta (approved):
LGTM! Thanks!

PR #4992, opened 2025-01-31
Query: Adds ArrayContains to CosmosLinqExtensions to allow partial matching on array fields in queries

### LINKED ISSUE #4991: Add `Array_Contains` partial match support to Linq

The `Array_Contains` function CosmosDB Sql has a 3rd parameter which allows it to do a partial match on the given item.

https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query/array-contains

The current Linq `array.Contains(item)` don't have a way to set this 3rd parameter to do the partial match.

A Linq extension method that allows me to set all the CosmosSql Array_Contains function parameters, something like this:

```csharp
documents.Where(document => document.ObjectArray.ArrayContains(new { Name = "abc" }, true))
```

We can wrap this `Array_Contains` call in a Cosmos SQL user defined function, but we ran into issues with quindex

**Additional context**

I had a quick play around to see if I could add this to the client lib, if this is a valid request I can create a PR for this new extension method

https://github.com/bcrobinson/azure-cosmos-dotnet-v3/tree/feature/array_contains_builtin_function

bcrobinson:
# Pull Request Template

## Description

The `Array_Contains` function CosmosDB Sql has a 3rd parameter which allows it to do a partial match on the given item. This is unable to be called with the built in Linq `array.Contains(item)` extension methods.

This adds this adds an explicit mapping to this function to allow it to be called in Linq like this: 

```
documents.Where(document => document.ObjectArray.ArrayContains(new { Name = "abc" }, true))
```

#### New extension method called `ArrayContains` instead of `Contains`

I namef the new `CosmosLinqExtensions` extension method `ArrayContains` instead of `Contains` as the other extensions methods matched thier Cosmos Sql counterparts and because this is for a direct invokation of this function.

Can change the name if you require.

##### `ArrayContains` Signature

```csharp
bool ArrayContains(this IEnumerable obj, object itemToMatch, bool partialMatch)
```

The `itemToMatch` paramater is an `object` as it will need to take anonymous objects (and c# doesn't have a similar concept to typescript's `Partial<T>`). If was generic callers would be force to pass in an object of the same type, and the default serialiser creates object with every property that exists on the class. So the created query object with default values (e.g. Array_Contains(root["Items"],  `{ "I" = 0, "s" = null })` )which would the cause Cosmos to try and match on all properties.

Because it's not generic the `obj` is then `IEnumerable`, so it would at a minimum only work on iterable values.

```csharp
bool ArrayContains(this IEnumerable obj, object itemToMatch, bool partialMatch)
```

If this was `ArrayContains<T>(this IEnumerable<T> obj, T itemToMatch, bool partialMatch)` then we'd need to always pass the same `T` type to this method, and the serialiser would then serialise all the properties on the type `T`, which would cause the partial matching in cosmos to fail. The downside is you loose some type checking (the anonymous object could have a property name typed wrong),

### Modification to `ArrayContainsVisitor`

The `ArrayContainsVisitor` visitor already exists for the `Contains` Linq mappings, the I modified this to work with the different extension version for Array Contains.

## Type of change

## Closing issues

closes #4991


adityasa:
@bcrobinson technically adding this functionality makes sense. However, ARRAY_CONTAINS with partial match is a very expensive function (both RU consumption and latency wise) and use of it is generally discouraged in favor of another pattern.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqTranslationBaselineTests.cs
```diff
@@ -346,6 +347,31 @@ public void TestRegexMatchFunction()
...
+                new LinqTestInput("ArrayContains in Filter clause with object value and match partial true", b => getQuery(b).Where(doc => doc.ObjectArrayField.ArrayContains(new { Field = "abc" }, true))),
+                
+                new LinqTestInput("ArrayContains in Select clause with int value and match partial false", b => getQuery(b).Select(doc => doc.ArrayField.ArrayContains(1, false))),
+                new LinqTestInput("ArrayContains in Filter clause with int value and match partial false", b => getQuery(b).Where(doc => doc.ArrayField.ArrayContains(1, false))),
+                new LinqTestInput("ArrayContains in Select clause with object value and match partial false", b => getQuery(b).Select(doc => doc.ObjectArrayField.ArrayContains(new { Field = "abc" }, false))),
+                new LinqTestInput("ArrayContains in Filter clause with object value and match partial false", b => getQuery(b).Where(doc => doc.ObjectArrayField.ArrayContains(new { Field = "abc" }, false))),
```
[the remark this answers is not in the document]
bcrobinson: Yes, added an overload of the `ArrayContains` extension method with only 2 parameters for this as well. Can't use default values in expressions so had to add the overload.

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqExtensions.cs
```diff
@@ -237,6 +238,62 @@ public static bool RegexMatch(this object obj, string regularExpression, string
...
+        /// ]]>
+        /// </code>
+        /// </example>
+        public static bool ArrayContains(this IEnumerable obj, object itemToMatch)
+        {
+            // The signature for this is not generic so the user can pass in anonymous type for the item to match
+            // e.g documents.Where(document => document.FooItems.ArrayContains(new { Name = "Bar" }, true)
+            throw new NotImplementedException(ClientResources.TypeCheckExtensionFunctionsNotImplemented);
+        }
+
```
leminh98, line 266: I think this Array_contains is functionally the same as IEnumerable.Contains(), so please remove this and rename the other extension method to Contains instead

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqExtensions.cs
```diff
@@ -237,6 +238,62 @@ public static bool RegexMatch(this object obj, string regularExpression, string
...
+        /// <example>
+        /// <code>
+        /// <![CDATA[
+        /// var matched = documents.Where(document => document.Namess.ArrayContains(<itemToMatch>, <partialMatch>));
+        /// // To do a partial match on an array of objects, pass in an anonymous object set partialMatch to true
+        /// var matched = documents.Where(document => document.ObjectArray.ArrayContains(new { Name = <name> }, true));
+        /// ]]>
+        /// </code>
+        /// </example>
+        public static bool ArrayContains(this IEnumerable obj, object itemToMatch, bool partialMatch)
```
leminh98, line 287: >ArrayContains [](http://example.com/codeflow?start=27&length=13)

renamed to Contains

Microsoft.Azure.Cosmos/src/Linq/CosmosLinqExtensions.cs
```diff
@@ -237,6 +238,62 @@ public static bool RegexMatch(this object obj, string regularExpression, string
...
+        /// There's no implementation provided in the client library.
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <param name="itemToMatch">The value to search within the array.</param>
+        /// <param name="partialMatch">Indicating whether the search should check for a partial match (true) or a full match (false).</param>
+        /// <returns>Returns true if the array array contains the specified value; otherwise, false.</returns>
+        /// <example>
+        /// <code>
+        /// <![CDATA[
+        /// var matched = documents.Where(document => document.Namess.ArrayContains(<itemToMatch>, <partialMatch>));
```
leminh98, line 281: >Namess [](http://example.com/codeflow?start=63&length=6)

typo

NaluTripician:
Hi @bcrobinson, thank you for the contribution! This PR has been idle for ~432 days. Is this change still needed on your side? Flagging for SDK-team triage as part of an open-PR cleanup pass — we'd like to give you a clear yes/no rather than let it sit.

Issue #4795, opened 2024-10-10
Fault Injection for Metadata requests


NaluTripician:

NaluTripician:
Relevant ask for metadata requests:

https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4710

NaluTripician:
Closed by #4989

PR #5017, opened 2025-02-13
AddressRefresh: Fixes behavior of AddressRefresh calls so they preform cross regional retries

### LINKED ISSUE #4979: Address Refresh Cross Regional Retries
# Address Refresh Cross Regional Retries

Following the 12/26 outage of SCUS, during the analyis of the impact of several customers there was a gap in the retry logic of the SDK, particuallry with the logic of cross regional retries for address refresh calls. This document outlines the impact of the gap and the proposed solution.

## Impact

In the case of a regional outage, currently if the SDK attempts Address Refresh calls and the primary region is down, the SDK will not attempt to retry the call in the secondary region. This is due to the fact that when the address refresh times out, the SDK will thow a task cancelled exception. Currently the SDK does not have logic to catch this exception and will treat this as a timeout. With other timeouts, the SDK will wrap the exception in a 503 and upon reaching the RetryLayer, the `ClientRetryPolicy` will attempt to retry the call in the secondary region if available.

## Proposed Solution

The proposed solution would be to catch the OperationCanceledException as well as any 410s (Timeouts) and wrap them in a 503. This will allow the `ClientRetryPolicy` to attempt to retry the call in the secondary region if available.

## Impact of the Solution

There might be an impact with the use of this with the compute gateway. Further investigation will be needed to determine the impact of this change. A possible way to mitigate the impact would be to have a flag that would allow the user to enable this feature. This flag would be internal and not accessible to external customers. 

## Testing

The testing will be done with the FaultInjeciton Library, which will need to have metadata request support added before this fix can be tested. See #4795 for more information.

NaluTripician:
# Pull Request Template

## Address Refresh Cross Regional Retries

Following the 12/26 outage of SCUS, during the analyis of the impact of several customers there was a gap in the retry logic of the SDK, particuallry with the logic of cross regional retries for address refresh calls. This document outlines the impact of the gap and the proposed solution.

## Impact

In the case of a regional outage, currently if the SDK attempts Address Refresh calls and the primary region is down, the SDK will not attempt to retry the call in the secondary region. This is due to the fact that when the address refresh times out, the SDK will throw a task cancelled exception. Currently the SDK does not have logic to catch this exception and will treat this as a timeout. With other timeouts, the SDK will wrap the exception in a 503 and upon reaching the RetryLayer, the `ClientRetryPolicy` will attempt to retry the call in the secondary region if available.

## Proposed Solution

The proposed solution would be to catch the OperationCanceledException as well as any 410s (Timeouts) and wrap them in a 503. This will allow the `ClientRetryPolicy` to attempt to retry the call in the secondary region if available.

## Impact of the Solution

There might be an impact with the use of this with the compute gateway. Further investigation will be needed to determine the impact of this change. A possible way to mitigate the impact would be to have a flag that would allow the user to enable this feature. This flag would be internal and not accessible to external customers. 

### Example Diagnostics 

[AddressRefreshFixDiag.md](https://github.com/user-attachments/files/19257257/AddressRefreshFixDiag.md)

## Testing

The testing will be done with the FaultInjection Library, which will need to have metadata request support added before this fix can be tested. See #4795 for more information.

## Type of change

## Closing issues

To automatically close an issue: closes #4979


ananth7592 (dismissed):

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -39,7 +39,7 @@ public static HttpTimeoutPolicy GetTimeoutPolicy(
             //Get Addresses Requests
             if (documentServiceRequest.ResourceType == ResourceType.Address)
             {
-                return HttpTimeoutPolicyControlPlaneRetriableHotPath.Instance;
+                return HttpTimeoutPolicyControlPlaneRetriableHotPath.InstanceShouldThrow503OnTimeout;
```
FabianMeiswinkel, line 42: What is the behavior here? 503 is only thrown when the last retry also times out? or after the first timeout already?

FabianMeiswinkel: Also we have seen issues in a CRI where the aggressive timeouts on the first attempt are too small - we should add an option to override the default value (of 500ms i think) via env variable - Ananth has more context

kirankumarkolli: Good question, expectation is after all the retries. Otherwise its a regression.

NaluTripician: Cross region retries happen after 3 local retries (500 ms timeout, 5 second timeout, 65 second timeout)
[AddressRefreshFixDiag.md](https://github.com/user-attachments/files/19257257/AddressRefreshFixDiag.md)

kirankumarkolli:
```
            "(0, 0)": 3
```

Not a blocker but worth fixing it for future.

kirankumarkolli: How about include the diagnostics in PR description?

kirankumarkolli (approved):

kundadebdatta (approved):
LGTM.

PR #5023, opened 2025-02-15
[Internal] Circuit Breaker: Adds Code to Implement Per Partition Circuit Breaker

### LINKED ISSUE #4981: [Per Partition Automatic Failover] - Add Circuit Breaker to Improve Read Availability.
**Acceptance Criteria:**

- Reads should be able to trigger partition failover.
- The failover logic should be threshold based.
- There has to be a optimized way of failing back the read requests.

kundadebdatta:
# Pull Request Template

## Description

The idea of having a per partition circuit breaker (aka PPCB) is to optimize **a)** read availability , in a single master account and **b)** read + write availability in a multi master account, during a time when a specific partition in one of the regions is experiencing an outage/ quorum loss. This feature is independent of the partition level failover triggered by the backend. The per partition circuit breaker is developed behind a feature flag `AZURE_COSMOS_CIRCUIT_BREAKER_ENABLED`.

However, when the partition level failover is enabled, we will enable the PPCB by default so that the reads can benefits from it.

## Scope

- For single master, only the read requests will use the circuit breaker to add the pk-range to region override mapping, and use this mapping as a source of truth to route the read requests. 

- For multi master, both the read and write requests will use the circuit breaker to add the pk-range to region override mapping and use this mapping as a source of truth to route both read and write requests.

## Understanding the Configurations exposed by the environment variables:

- `AZURE_COSMOS_CIRCUIT_BREAKER_ENABLED`: This environment variable is used to enable/ disable the partition level circuit breaker feature. The default value is `false`.

- `AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECONDS`: This environment variable is used to set the background periodic address refresh task interval. The default value for this interval is `60 seconds`.

- `AZURE_COSMOS_PPCB_ALLOWED_PARTITION_UNAVAILABILITY_DURATION_IN_SECONDS`: This environment variable is used to set the partition unavailability time duration in seconds. The unavailability time indicates how long a partition can remain unhealthy, before it can re-validate it's connection status. The default value for this property is `5 seconds`.

- `AZURE_COSMOS_PPCB_CONSECUTIVE_FAILURE_COUNT_FOR_READS`: This environment variable is used to set the consecutive failure count for reads, before triggering per partition level circuit breaker flow. The default value for this flag is `10` consecutive failures within 1 min window.

- `AZURE_COSMOS_PPCB_CONSECUTIVE_FAILURE_COUNT_FOR_WRITES`: This environment variable is used to set the consecutive failure count for writes, before triggering per partition level circuit breaker flow. The default value for this flag is `5` consecutive failures within 1 min window.

## Understanding the Working Principle:

On a high level, there are three parts of the circuit breaker implementation:

- **Short Circuit and Failover detection:** The failover detection logic will reside in the SDK ClientRetryPolicy, just like we have for PPAF. Ideally the detection logic is based on the below two principles: 

    - **Status Codes:** The status codes that are indicative of partition level circuit breaker would be the following: a) `503` Service Unavailable, b) `408` Request Timeout, c) cancellation token expired.

    - **Threshold:** Once the failover condition is met, the SDK will look for some consecutive failures, until it hits a particular threshold. Once this threshold is met, the SDK will fail over the read requests to the next preferred region for that offending partition. For example, if the threshold value for read requests is `10`, then the SDK will look for `10` consecutive failures. If the threshold is met/ exceeded, the SDK will add the region failover information for that partition.

- **Failover a faulty partition to the next preferred region:** Once the failover conditions are met, the `ClientRetryPolicy` will trigger a partition level override using `GlobalPartitionEndpointManagerCore.TryMarkEndpointUnavailableForPartitionKeyRange` to the next region in the preferred region list. This failover information will help the current, as well as the subsequent requests (reads in single master and both reads and writes in multi master) to route the request to the next region.

- **Failback the faulty partition to it's original first preferred region:** With PPAF enabled, ideally the write requests will rely on 403.3 (Write Forbidden) signal to fail the partition back to the primary write region. However, this is not true for reads. That means SDK doesn’t have a definitive signal to identify when to initiate a failback for read requests.   
Hence, the idea is to create a background task during the time of read failover, which will keep track of the pk-range and region mapping. The task will periodically fetch the address from the gateway address cache for those pk ranges in the faulty region, and it will try to initiate Rntbd connection to all 4 replicas of that partition. The RNTBD open connection attempt will be made similar to that of the replica validation flow. The life cycle of the background task will get initiated during a failover and will remain until the SDK is disposed.
If the attempt to make the connection to all 4 replicas  is successful, then the task will remove/ override the entry with the primary region, resulting the SDK to failback the read requests.

## Type of change

## Closing issues

To automatically close an issue: closes #4981


Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+        /// </summary>
+        private readonly Lazy<ConcurrentDictionary<PartitionKeyRange, PartitionKeyRangeFailoverInfo>> PartitionKeyRangeToLocationForWrite = new (
             () => new ConcurrentDictionary<PartitionKeyRange, PartitionKeyRangeFailoverInfo>());
-
+
+        /// <summary>
+        /// A <see cref="Lazy{T}"/> instance of <see cref="ConcurrentDictionary{K,V}"/> containing the partition key range to failover info mapping.
+        /// This mapping is primarily used for reads in a single master account, and both reads and writes in a multi master account.
+        /// </summary>
+        private readonly Lazy<ConcurrentDictionary<PartitionKeyRange, PartitionKeyRangeFailoverInfo>> PartitionKeyRangeToLocationForReadAndWrite = new (
```
NaluTripician, line 72: Adding a method to access information from these dictionaries could be beneficial for work done in the logs in the future. Not blocking for this PR just a note for the future.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemIntegrationTests.cs
```diff
@@ -145,5 +145,450 @@ public async Task ReadMany2UnreachablePartitionsTest()
...
+            };
+
+            List<CosmosIntegrationTestObject> itemsList = new ()
+            {
+                new() { Id = "smTestId1", Pk = "smpk1" },
+            };
+
+            try
+            {
+                CosmosClient cosmosClient = new(connectionString: this.connectionString, clientOptions: cosmosClientOptions);
```
NaluTripician, line 197: We should probably put this client in a using statement to ensure it is disposed at the end of the test or incase of a test failure.

kundadebdatta: Thanks @NaluTripician! I have added the `using` statement.

Microsoft.Azure.Cosmos/src/RetryPolicy.cs
```diff
@@ -14,6 +14,7 @@ internal sealed class RetryPolicy : IRetryPolicyFactory
         private readonly GlobalEndpointManager globalEndpointManager;
         private readonly bool enableEndpointDiscovery;
         private readonly bool isPertitionLevelFailoverEnabled;
```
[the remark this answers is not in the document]
kundadebdatta: Fixed with latest commit.

Microsoft.Azure.Cosmos/src/RetryPolicy.cs
```diff
@@ -14,6 +14,7 @@ internal sealed class RetryPolicy : IRetryPolicyFactory
...
+        private readonly bool isPertitionLevelCircuitBreakerEnabled;
```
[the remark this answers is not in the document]
kundadebdatta: Fixed with latest commit.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -45,7 +46,8 @@ public ClientRetryPolicy(
             GlobalPartitionEndpointManager partitionKeyRangeLocationCache,
             RetryOptions retryOptions,
             bool enableEndpointDiscovery,
-            bool isPertitionLevelFailoverEnabled)
+            bool isPertitionLevelFailoverEnabled,
+            bool isPertitionLevelCircuitBreakerEnabled)
```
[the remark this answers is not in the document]
kundadebdatta: Fixed with latest commit.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -29,6 +29,7 @@ internal sealed class ClientRetryPolicy : IDocumentClientRetryPolicy
         private readonly GlobalPartitionEndpointManager partitionKeyRangeLocationCache;
         private readonly bool enableEndpointDiscovery;
         private readonly bool isPertitionLevelFailoverEnabled;
+        private readonly bool isPertitionLevelCircuitBreakerEnabled;
```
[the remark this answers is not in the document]
kundadebdatta: Fixed with latest commit.

jeet1995:
> AZURE_COSMOS_PPCB_CONSECUTIVE_FAILURE_COUNT: This environment variable is used to set the consecutive failure count for reads, before triggering per partition level circuit breaker flow. The default value for this flag is 10 consecutive requests within 1 min window.

So according to this, our circuit breaking tolerance is same for both reads and writes. A write workload has higher chance of faulting than reads due to 1 primary v/s 3 secondaries - so I'd prefer having separate thresholds.

jeet1995:
> Threshold: Once the failover condition is met, the SDK will look for some consecutive failures, until it hits a particular threshold. Once this threshold is met, the SDK will fail over the read requests to the next preferred region for that offending partition. For example, if the threshold value for read requests is 10, then the SDK will look for 10 consecutive failures. If the threshold is met/ exceeded, the SDK will add the region failover information for that partition.

Just scoped to the description but writes will failover to for a multi-write account.

jeet1995:
> Status Codes: The status codes that are indicative of partition level circuit breaker would be the following: a) 503 Service Unavailable, b) 408 Request Timeout, c) cancellation token expired.

A minor point - in the Java SDK, we track 500s too.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+            if (!this.IsRequestEligibleForPartitionFailover(
+                request,
+                shouldValidateFailedLocation: false,
+                out PartitionKeyRange? partitionKeyRange,
+                out Uri? _))
+            {
+                return false;
+            }
+
+            if (partitionKeyRange == null)
```
jeet1995, line 127: nit: This logic can be front loaded or be within `IsRequestEligibleForPartitionFailover`?

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+            if (!this.IsRequestEligibleForPartitionFailover(
+                request,
+                shouldValidateFailedLocation: true,
+                out PartitionKeyRange? partitionKeyRange,
+                out Uri? failedLocation))
+            {
+                return false;
+            }
+
+            if (partitionKeyRange == null || failedLocation == null)
```
jeet1995, line 164: nit: Dito comment on moving this within `IsRequestEligibleForPartitionFailover`.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+
+        /// <summary>
+        /// Determines if partition level failover locations can be used for the given request.
+        /// </summary>
+        /// <param name="request">An instance of the <see cref="DocumentServiceRequest"/>.</param>
+        /// <returns>True if partition level failover locations can be used, otherwise false.</returns>
+        private bool CanUsePartitionLevelFailoverLocations(
+            DocumentServiceRequest request)
+        {
+            if (this.globalEndpointManager.ReadEndpoints.Count <= 1)
```
jeet1995, line 302: In PPCB scenario, when determining whether to failover, we also check post user-enforced region exclusion if the resultant region count is <= 1.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
-            if (partionFailover.TryMoveNextLocation(
-                    locations: nextLocations,
-                    failedLocation: failedLocation))
+        /// <summary>
+        /// This is a recursive method that will be called every minute to refresh the connection to the failed backend replicas.
+        /// Note that the refresh interval can configured by the end user using the environment variable:
+        /// AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECONDS.
+        /// </summary>
+#pragma warning disable VSTHRD100 // Avoid async void methods
+        private async void InitiateCircuitBreakerFailbackLoop()
```
jeet1995, line 360: Any memory-based pitfalls with using recursion here?

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+            {
+                if (this.cancellationTokenSource.IsCancellationRequested && (ex is OperationCanceledException || ex is ObjectDisposedException))
+                {
+                    return;
+                }
+
+                DefaultTrace.TraceCritical("GlobalPartitionEndpointManagerCore: InitiateCircuitBreakerFailbackLoop() - Unable to get address and open connections. Exception: {0}", ex.ToString());
+            }
+
+            // Call itself to create a loop to continuously do background refresh every 5 minutes
```
jeet1995, line 393: Regarding this comment about every 5 minutes - it is reliant on this `AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECONDS` right?

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -207,19 +628,70 @@ public bool TryMoveNextLocation(
...
+            }
+
+            public void IncrementRequestFailureCounts(
+                DateTime currentTime)
+            {
+                this.SnapshotPartitionFailoverTimestamps(
+                    out DateTime _,
+                    out DateTime lastRequestFailureTime);
+
+                if (currentTime - lastRequestFailureTime > this.TimeoutCounterResetWindowInMinutes)
```
jeet1995, line 660: nit / non-blocking: I wonder if `TimeoutCounterResetWindowInMinutes` can also be exposed as a system property.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
jeet1995: In the Java SDK, what we do is we snapshot `PartitionKeyRangeFailoverInfo` in the request. This snapshotting helps identify what the exact state was when a failover decision was made.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemIntegrationTests.cs
```diff
@@ -145,5 +145,450 @@ public async Task ReadMany2UnreachablePartitionsTest()
...
         }
+
+        [TestMethod]
+        [TestCategory("MultiRegion")]
+        [DataRow("10", DisplayName = "Scenario whtn the circuit breaker consecutive failure threshold is set to 10.")]
+        [DataRow("20", DisplayName = "Scenario whtn the circuit breaker consecutive failure threshold is set to 20.")]
+        [DataRow("30", DisplayName = "Scenario whtn the circuit breaker consecutive failure threshold is set to 30.")]
+        [Owner("dkunda")]
+        [Timeout(70000)]
+        public async Task ReadItemAsync_WithCircuitBreakerEnabledAndSingleMasterAccountAndServiceUnavailableReceived_ShouldApplyPartitionLevelOverride(
```
jeet1995, line 156: It would be good to extend this to non-point operations such as queries and batch.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -207,19 +628,70 @@ public bool TryMoveNextLocation(
...
+            /// Helper method to snapshot the connection timestamps.
+            /// </summary>
+            /// <param name="firstRequestFailureTime">A <see cref="DateTime"/> field containing the last send attempt time.</param>
+            /// <param name="lastRequestFailureTime">A <see cref="DateTime"/> field containing th e last send attempt time.</param>
+            public void SnapshotPartitionFailoverTimestamps(
+                out DateTime firstRequestFailureTime,
+                out DateTime lastRequestFailureTime)
+            {
+                Debug.Assert(!Monitor.IsEntered(this.timestampLock));
+                lock (this.timestampLock)
```
jeet1995, line 679: On another note, if we don't want to expose `TimeoutCounterResetWindowInMinutes`, some staleness here would be okay I guess (no locking).

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+        /// <summary>
+        /// Attempts to route the request to a partition level override location if available.
+        /// This method checks if there is a failover location for the given partition key range
+        /// and updates the request context to route to that location.
+        /// </summary>
+        /// <param name="partitionKeyRange">The partition key range for which the request is being routed.</param>
+        /// <param name="request">The document service request to be routed.</param>
+        /// <param name="partitionKeyRangeToLocationMapping">The mapping of partition key ranges to their failover locations.</param>
+        /// <returns>True if the request was successfully routed to a partition level override location, otherwise false.</returns>
+        private bool TryRouteRequestForPartitionLevelOverride(
```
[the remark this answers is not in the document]
kundadebdatta: `Say, I have NCUS, WUS and CUS single-master setup and is also the preferred order - if NCUS is marked down, WUS would be the override. So, in a particular window, if reads sees a 503 for WUS, it will maybe retry again on WUS - how do you make it retry on CUS?`

=> I have optimized the code with [the latest commit](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5023/commits/bea2bad9e4bb3c4e164bd6a6a8bd24f9f5bfc28c) to retry even on `CUS`, when the SDK receives a `503` from `WUS`. To understand better, please [have a look at this test.](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/bea2bad9e4bb3c4e164bd6a6a8bd24f9f5bfc28c/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemIntegrationTests.cs#L262-L420)

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kundadebdatta: I feel there are already 4 environment variables we are exposing to control the failover detection and failback. IMO, we should keep the `TimeoutCounterResetWindowInMinutes` as internal only (which is `1` minute at the moment), and iterate over this, if needed.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kundadebdatta: `partitionKeyRange` and `failedLocation` are `out` parameters, which can be `null` as well. Therefore, a `null` check is required.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kundadebdatta: `partitionKeyRange` and `failedLocation` are `out` parameters, which can be `null` as well. Therefore, a `null` check is required.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kundadebdatta: That is correct. The interval is driven by `AZURE_COSMOS_PPCB_STALE_PARTITION_UNAVAILABILITY_REFRESH_INTERVAL_IN_SECOND`.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kundadebdatta: We use both [recursive ](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs#L563-L611)and [iterative approach](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs#L289-L364) in our v3 codebase for running indefinite loop to do periodic background refresh. Given that, my preference would be to use the iterative approach to reduce additional memory due to recursion call stack. 

Updated the code to use indefinite `while` loop for the background refresh.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kundadebdatta: Given that exclude region will be honored, irrespective of circuit breaker initiated failovers, I feel updating the logic here would not make much difference. We can go with an example offline, to get to a closure.

kundadebdatta:
> > AZURE_COSMOS_PPCB_CONSECUTIVE_FAILURE_COUNT: This environment variable is used to set the consecutive failure count for reads, before triggering per partition level circuit breaker flow. The default value for this flag is 10 consecutive requests within 1 min window.
> 
> So according to this, the circuit breaking threshold is same for both reads and writes. A write workload has higher chance of faulting than reads due to 1 primary v/s 3 secondaries (SF automation for primary change might be very sensitive with primary down regardless) - so I'd prefer having separate thresholds. This will allow failover trigger through writes more aggressively if needed even if the workload itself is read heavy.

Good Point. I think it's a good idea to keep separate threshold for both reads and writes. Exposed two separate environment variables for capturing read and write thresholds.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -939,8 +939,11 @@ internal virtual void Initialize(Uri serviceEndpoint,
 #endif
 
             this.GlobalEndpointManager = new GlobalEndpointManager(this, this.ConnectionPolicy);
-            this.PartitionKeyRangeLocation = this.ConnectionPolicy.EnablePartitionLevelFailover
-                ? new GlobalPartitionEndpointManagerCore(this.GlobalEndpointManager)
+            this.PartitionKeyRangeLocation = this.ConnectionPolicy.EnablePartitionLevelFailover || this.ConnectionPolicy.EnablePartitionLevelCircuitBreaker
```
kirankumarkolli, line 942: PPCB depdnent on PPAF, some PPAF check might suffice?

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
```diff
@@ -22,6 +22,36 @@ internal static class ConfigurationManager
         /// </summary>
         internal static readonly string PartitionLevelFailoverEnabled = "AZURE_COSMOS_PARTITION_LEVEL_FAILOVER_ENABLED";
 
+        /// <summary>
+        /// A read-only string containing the environment variable name for enabling per partition circuit breaker. The default value
+        /// for this flag is false.
+        /// </summary>
+        internal static readonly string PartitionLevelCircuitBreakerEnabled = "AZURE_COSMOS_PARTITION_LEVEL_CIRCUIT_BREAKER_ENABLED";
```
kirankumarkolli, line 29: nit: Just for simplicity how about use PPCB? (Below configs do use it anyways)

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -727,7 +727,12 @@ public Func<HttpClient> HttpClientFactory
         /// <summary>
         /// Enable partition key level failover
         /// </summary>
-        internal bool EnablePartitionLevelFailover { get; set; } = ConfigurationManager.IsPartitionLevelFailoverEnabled(defaultValue: false);
+        internal bool EnablePartitionLevelFailover { get; set; } = ConfigurationManager.IsPartitionLevelFailoverEnabled(defaultValue: false);
+
+        /// <summary>
+        /// Enable partition level circuit breaker
+        /// </summary>
+        internal bool EnablePartitionLevelCircuitBreaker { get; set; } = ConfigurationManager.IsPartitionLevelCircuitBreakerEnabled(defaultValue: false);
```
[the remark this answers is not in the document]
kirankumarkolli: May not be always true right?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -524,6 +531,35 @@ private bool ShouldMarkEndpointUnavailableOnSystemResourceUnavailableForWrite(
...
+        /// </returns>
+        private bool IsRequestEligibleForPerPartitionAutomaticFailover()
+        {
+            return !this.documentServiceRequest.IsReadOnlyRequest
+                && !this.isMultiMasterWriteRequest
+                && this.isPartitionLevelFailoverEnabled;
+        }
+
+        /// <summary>
+        /// Determines if a read request is eligible for partition-level circuit breaker.
```
kirankumarkolli, line 548: nit: Read request used in the doc, even writes in MM are also used.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -454,10 +457,14 @@ private ShouldRetryResult TryMarkEndpointUnavailableForPkRangeAndRetryOnServiceU
 
             if (shouldMarkEndpointUnavailableForPkRange)
```
kirankumarkolli, line 458: Some failures might be fatal and might need to by-pass circuit-breaker? (ex: 429/3092)

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli, line 735: What's the guidance for compute gateway?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -524,6 +531,35 @@ private bool ShouldMarkEndpointUnavailableOnSystemResourceUnavailableForWrite(
...
+        /// Determines if a read request is eligible for partition-level circuit breaker.
+        /// This method checks if the request is a read-only request, if partition-level circuit breaker is enabled,
+        /// and if the partition key range location cache indicates that the partition can fail over based on the number of request failures.
+        /// </summary>
+        /// <returns>
+        /// True if the read request is eligible for partition-level circuit breaker, otherwise false.
+        /// </returns>
+        private bool IsRequestEligibleForPartitionLevelCircuitBreaker()
+        {
+            return (this.documentServiceRequest.IsReadOnlyRequest || this.isMultiMasterWriteRequest)
```
kirankumarkolli, line 557: Is MM + PPAF in scope in near term?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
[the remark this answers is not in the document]
kundadebdatta: The real validation is happening on line# 988:

image

Let me know if it makes sense.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kundadebdatta: Updated the verbiage.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -524,6 +531,35 @@ private bool ShouldMarkEndpointUnavailableOnSystemResourceUnavailableForWrite(
                 && subStatusCode == SubStatusCodes.SystemResourceUnavailable;
         }
 
+        /// <summary>
+        /// Determines if the current request is a write request.
+        /// </summary>
+        /// <returns>
+        /// True if the request is a write request; otherwise, false.
+        /// </returns>
+        private bool IsRequestEligibleForPerPartitionAutomaticFailover()
```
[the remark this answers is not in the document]
kundadebdatta: Thanks for the suggestion. I feel keeping the logic in one single place is always a good idea. Refactored the code to keep the logic in `partitionKeyRangeLocationCache`

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: For compute gateway, by default the PPAF will be disabled, so does the PPCB. If compute choses to enable PPAF, then we will enable PPCB by default.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kundadebdatta: Yes. Thanks for pointing this out. `429/3092` should trigger the partition failover irrespective of whether the PPCB counter reaches it's threshold. Updated the code to reflect the same.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kirankumarkolli: Can we please include this in the property codedoc for reference?

kundadebdatta: Updated the code doc.

jeet1995 (approved):
LGTM - thanks! Looks like the testing coverage can be extended to non-point operations too. 

One other thing and this is for PPAF with <Strong + PPCB - reads could go to a non-write region and fail with a 404:1002, we need to ensure 404:1002 cross-region retry happens on the partition-set level write region (cross-region retry semantics for reads hitting 404:1002 is to use the account-level write region).

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/DocumentClient.cs
kundadebdatta: The real validation is happening on line# 988:

image

Let me know if it makes sense.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+                    PartitionKeyRangeFailoverInfo partionFailover = this.PartitionKeyRangeToLocationForReadAndWrite.Value[pkRange];
+
+                    partionFailover.SnapshotPartitionFailoverTimestamps(
+                        out DateTime firstRequestFailureTime,
+                        out DateTime _);
+
+                    if (DateTime.UtcNow - firstRequestFailureTime > TimeSpan.FromSeconds(this.partitionUnavailabilityDurationInSeconds))
+                    {
+                        // The first failed location would always be the first preferred location.
+                        Uri originalFailedLocation = partionFailover.FirstFailedLocation;
```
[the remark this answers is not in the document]
kundadebdatta: True, updated the verbiage accordingly. Exclude regions will always be honored.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -7,181 +7,602 @@ namespace Microsoft.Azure.Cosmos.Routing
...
+        }
+
+        /// <summary>
+        /// Attempts to open connections to unhealthy endpoints and initiates failback if the connections are successful.
+        /// This method checks the partition key ranges that have failed locations and tries to re-establish connections
+        /// to those locations. If a connection is successfully re-established, it initiates a failback to the original
+        /// location for the partition key range.
+        /// </summary>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        private async Task TryOpenConnectionToUnhealthyEndpointsAndInitiateFailbackAsync()
```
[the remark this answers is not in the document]
kundadebdatta: This is rather a indefinite `while` loop running on the same thread. And the async task `TryOpenConnectionToUnhealthyEndpointsAndInitiateFailbackAsync()` will be called by the while loop itself.

PR #5024, opened 2025-02-15
Query: Fixes logic to determine whether to use distributed query by adding a check for gateway connection mode

sc978345:
## Description

This changes the logic for determining whether to use the distributed query dedicated gateway.

With this PR, we check both the **AZURE_COSMOS_DISTRIBUTED_QUERY_GATEWAY_ENABLED** environment variable **and** the connection mode to determine whether to use the distributed query dedicated gateway. If the environment variable is set to **True** **and** the connection mode is **gateway**, then we **will** use the dedicated gateway.  If the connection mode is **direct**, we should not use the distributed query dedicated gateway.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/v3Query/QueryIterator.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/v3Query/QueryIterator.cs
@@ -150,7 +150,7 @@ namespace Microsoft.Azure.Cosmos.Query
                 returnResultsInDeterministicOrder: queryRequestOptions.ReturnResultsInDeterministicOrder,
                 enableOptimisticDirectExecution: queryRequestOptions.EnableOptimisticDirectExecution,
                 isNonStreamingOrderByQueryFeatureDisabled: queryRequestOptions.IsNonStreamingOrderByQueryFeatureDisabled,
-                enableDistributedQueryGatewayMode: queryRequestOptions.EnableDistributedQueryGatewayMode,
+                enableDistributedQueryGatewayMode: queryRequestOptions.EnableDistributedQueryGatewayMode && (clientContext.ClientOptions?.ConnectionMode == ConnectionMode.Gateway),
                 testInjections: queryRequestOptions.TestSettings);
 
             return new QueryIterator(
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/DistributedQueryClientTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/DistributedQueryClientTests.cs
@@ -162,6 +162,33 @@ namespace Microsoft.Azure.Cosmos.EmulatorTests.Query
                 ImplementationAsync);
         }
 
+        [TestMethod]
+        public async Task DirectModeTestsAsync()
+        {
+            using EnvironmentVariableOverride enableDistributedQueryOverride = new EnvironmentVariableOverride(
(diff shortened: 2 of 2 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/DistributedQueryClientTests.cs
```diff
@@ -162,6 +162,33 @@ await this.CreateIngestQueryDeleteAsync(
...
+                        "SELECT VALUE COUNT(1) FROM c",
+                        PageSizes,
+                        Expectations.OneValueWithDocumentCountIsPresent),
+                };
+
+                return RunStreamIteratorTestsAsync(container, testCases);
+            }
+
+            await this.CreateIngestQueryDeleteAsync(
+                ConnectionModes.Direct,
```
neildsh, line 186: >ConnectionModes.Direct, [](http://example.com/codeflow?start=16&length=23)

Based on your current change, this would not run the Distributed Query pipeline, correct ?

Microsoft.Azure.Cosmos/src/Query/v3Query/QueryIterator.cs
```diff
@@ -150,7 +150,7 @@ public static QueryIterator Create(
                 returnResultsInDeterministicOrder: queryRequestOptions.ReturnResultsInDeterministicOrder,
                 enableOptimisticDirectExecution: queryRequestOptions.EnableOptimisticDirectExecution,
                 isNonStreamingOrderByQueryFeatureDisabled: queryRequestOptions.IsNonStreamingOrderByQueryFeatureDisabled,
-                enableDistributedQueryGatewayMode: queryRequestOptions.EnableDistributedQueryGatewayMode,
+                enableDistributedQueryGatewayMode: queryRequestOptions.EnableDistributedQueryGatewayMode && (clientContext.ClientOptions?.ConnectionMode == ConnectionMode.Gateway),
```
neildsh, line 153: >clientContext.ClientOptions [](http://example.com/codeflow?start=109&length=27)

Are ClientOptions allowed to be null here ?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/DistributedQueryClientTests.cs
sc978345: **With** my change:
enableDistributedQueryGatewayMode gets set to false
DistributedQueryPipelineStage is **not** created
Does not error out, because the non-DQ pipeline is used

**Without** my change:
enableDistributedQueryGatewayMode gets set to true
DistributedQueryPipelineStage **is** created
Errors out with BadRequest, 'The provided cross partition query can not be directly served by the gateway.'

Microsoft.Azure.Cosmos/src/Query/v3Query/QueryIterator.cs
sc978345: I did not see anything preventing it from being null

neildsh (approved):

Microsoft.Azure.Cosmos/src/Query/v3Query/QueryIterator.cs
sc978345: I have pushed a change to remove the unneeded check

sboshra (approved):
:shipit:

PR #5050, opened 2025-03-05
Enhancement: Configurable First Retry Timeout for HTTP Calls

ananth7592:
- Added support for configuring the first retry timeout in the HTTP retry stack via an environment variable.
- Default timeout remains 0.5 seconds, ensuring backward compatibility.
- This change enables customers to customize retry behavior based on their needs.

# Pull Request Template

## Description

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.Cosmos
     using System.Collections.Generic;
     using System.Net.Http;
     using System.Text;
+    using Microsoft.Azure.Documents;
 
     internal sealed class HttpTimeoutPolicyControlPlaneRetriableHotPath : HttpTimeoutPolicy
     {
@@ -15,6 +16,12 @@ namespace Microsoft.Azure.Cosmos
         public bool shouldThrow503OnTimeout;
         private static readonly string Name = nameof(HttpTimeoutPolicyControlPlaneRetriableHotPath);
 
+        public const string HttpFirstRetryTimeoutValue = "AZURE_COSMOS_SDK_HTTP_FIRST_RETRY_TIMEOUT_VALUE_SECONDS";
+
+        /// <summary>
+        /// A read-only <see cref="TimeSpan"/> containing the default value of the first request timeout after which the request is retried.
+        /// </summary>
+        private const double firstRetryTimeoutDefault = 0.5;
         private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeout)
         {
             this.shouldThrow503OnTimeout = shouldThrow503OnTimeout;
@@ -22,7 +29,9 @@ namespace Microsoft.Azure.Cosmos
 
         private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
+            (TimeSpan.FromSeconds(Helpers.GetEnvironmentVariable(
+                        name: HttpFirstRetryTimeoutValue,
+                        defaultValue: firstRetryTimeoutDefault)), TimeSpan.Zero),
             (TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(1)),
             (TimeSpan.FromSeconds(65), TimeSpan.Zero),
         };


Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -15,14 +16,22 @@ internal sealed class HttpTimeoutPolicyControlPlaneRetriableHotPath : HttpTimeou
         public bool shouldThrow503OnTimeout;
         private static readonly string Name = nameof(HttpTimeoutPolicyControlPlaneRetriableHotPath);
 
+        public const string HttpFirstRetryTimeoutValue = "AZURE_COSMOS_SDK_HTTP_FIRST_RETRY_TIMEOUT_VALUE_SECONDS";
```
kirankumarkolli, line 19: Please include in the PR with more details on why and when to use it

ananth7592: Added a more descriptive description

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -15,14 +16,22 @@ internal sealed class HttpTimeoutPolicyControlPlaneRetriableHotPath : HttpTimeou
...
+        private const double firstRetryTimeoutDefault = 0.5;
         private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeout)
         {
             this.shouldThrow503OnTimeout = shouldThrow503OnTimeout;
         }
 
         private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
+            (TimeSpan.FromSeconds(Helpers.GetEnvironmentVariable(
```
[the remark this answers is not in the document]
ananth7592: Kiran, wouldn't this env variable be also useful when the customer wants to fail faster than 0.5 seconds and thereby set the value at LESS THAN 0.5 seconds

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -15,14 +16,22 @@ internal sealed class HttpTimeoutPolicyControlPlaneRetriableHotPath : HttpTimeou
         public bool shouldThrow503OnTimeout;
         private static readonly string Name = nameof(HttpTimeoutPolicyControlPlaneRetriableHotPath);
 
+        public const string HttpFirstRetryTimeoutValue = "AZURE_COSMOS_SDK_HTTP_FIRST_RETRY_TIMEOUT_VALUE_MS";
```
kirankumarkolli, line 19: One alternative is to move all config into `ConfigurationManager` will help us on future discoverability as well.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -15,14 +16,22 @@ internal sealed class HttpTimeoutPolicyControlPlaneRetriableHotPath : HttpTimeou
...
+        private const double firstRetryTimeoutDefault = 500;
         private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeout)
         {
             this.shouldThrow503OnTimeout = shouldThrow503OnTimeout;
         }
 
         private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
+            (TimeSpan.FromMilliseconds(Helpers.GetEnvironmentVariable(
```
kirankumarkolli, line 32: If custom configuration it-self is >5seconds then is second retry with <== first worth?

ananth7592: So if the customer chooses a value as high as 5 seconds, we wait for a timeout of 5 seconds the first time and 5 seconds the second time as well. That should be okay , right? 

Also I am adding a minimum limit of 100ms to ensure the customer is not using a very low first timeout value.

kirankumarkolli: What's expected if they configure >5 seconds?

ananth7592: Then , the first retry is attempted after 5 seconds (unless there is a cancellationToken with a lesser duration cancelling the task) . 

Then if the HTTP call still fails, the second retry kicks in at the 10th (+5 seconds)

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -22,9 +28,14 @@ private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeo
...
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
-            (TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(1)),
-            (TimeSpan.FromSeconds(65), TimeSpan.Zero),
+                (TimeSpan.FromMilliseconds(
+                        Math.Max(100, Helpers.GetEnvironmentVariable(
+                            name: ConfigurationManager.HttpFirstRetryTimeoutValue,
+                            defaultValue: firstRetryTimeoutDefault))),
+                    TimeSpan.Zero
+                ),
+                (TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(1)),
```
xinlian12: I think I have the same question -> if customer has configured a timeout value > 5s, then even for the second or third retry shoud we use Max.max(ConfigurationManager.HttpFirstRetryTimeoutValue, 5)?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -22,9 +28,14 @@ private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeo
 
         private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
-            (TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(1)),
-            (TimeSpan.FromSeconds(65), TimeSpan.Zero),
+                (TimeSpan.FromMilliseconds(
```
xinlian12, line 25: maybe also add some tests to verify the change?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
ananth7592: Agreed. Updated the logic to use the max of the provided value and the defaults for second and third timeouts. Added tests as well.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
ananth7592: Done. Added a variety of scenarios for testing.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
ananth7592: Updated the logic to use the max of the provided value and the defaults for second and third timeouts. Added tests as well.

xinlian12 (approved):
LGTM, thanks

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -22,9 +27,18 @@ private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeo
...
+                        Math.Max(100, ConfigurationManager.GetEnvironmentVariable(
```
kundadebdatta: Do we want to limit it only to `HttpTimeoutPolicyControlPlaneRetriableHotPath` timeout policy ?

kundadebdatta (dismissed):

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
ananth7592: Yes. We want to introduce this env variable based first http timeout override for expensive operations such as "queryPlan" to help customers bail out after a longer duration than the default 500ms

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -22,9 +22,12 @@ private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeo
 
         private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
-            (TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(1)),
-            (TimeSpan.FromSeconds(65), TimeSpan.Zero),
+                (TimeSpan.FromMilliseconds(
+                        Math.Max(100, ConfigurationManager.GetHttpFirstRetryTimeoutValueInMs())),
```
kirankumarkolli, line 26: I suppose this change is because the first timeout is lower, shouldn't then first one be at-least 500ms as before?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -22,9 +22,12 @@ private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeo
...
+                    TimeSpan.Zero
+                ),
+                (TimeSpan.FromMilliseconds(Math.Max(5000, ConfigurationManager.GetHttpFirstRetryTimeoutValueInMs())), TimeSpan.FromSeconds(1)),
```
kirankumarkolli, line 29: Why to override seconds and third

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
ananth7592: This change is to not let the client set a value that is less than 100 ms, the minimum value must be atleast 100ms for these Control plane calls.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
ananth7592: This max(5seconds, EnvVariableValue) helps in cases where customer set a first timeout value that is for example 6 seconds. In this case the second timeout value also should be atleast 6 seconds

PR #5069, opened 2025-03-17
AsyncCache: Adds support for stack trace optimization during exceptions for AsyncCache and AsyncCacheNonblocking

ananth7592:
# Pull Request Template

## Description

This PR addresses the issue of shared exception objects in the AsyncCache class, which was causing significant performance degradation and memory bloat recently for open AI .
 
The fix involves creating shallow copies of known offenders and rethrowing the rest. This approach ensures that each AsyncCache.GetAsync call receives its own exception instance, preventing the dangerous growth of the exception object's StackTrace and reducing GC overhead.

All ICloneable implemenations of exceptions will benefit from this.

https://learn.microsoft.com/en-us/dotnet/api/system.icloneable?view=net-9.0

Key changes:
- Added support to post-process the exception in above mentioned two cache classes that creates a shallow copy of specific exception types to prevent stack trace proliferation and rethrows them, while propagating all other exceptions unchanged.
 - Added support for any exception that is ICloneable supported , TimeoutException and TaskCanceledException in this drop
 - Improved exception handling to avoid shared exception objects across threads.

**Feature is enabled by default. To turn it off, set EnableAsyncCacheExceptionNoSharing CosmosClientOptions to false**

Proof of testing:

**Repro Steps**
To reproduce the issue, I followed these high-level steps:

- Create a Cosmos client with the necessary settings to preempt gateway calls for address refresh to fail with a specific exception.

- Create a TaskCompletionSource that waits until all the threads are created.

- Initiate calls to createItem that will eventually block, with one of them erroring out with the pre-set exception.

- Observe the exception storm as the same exception trace is copied over to all the threads.

**After the fix**

- Tested with TimeoutException (2K Threads)  and TaskCanceledException (10K threads) using TaskCompletionSource with up to 10,000 threads without issues on the devbox.

- Implemented the fix in AsyncNonBlockingCache by creating shallow copies of known offenders and rethrowing the rest.

- Re-ran the setup and observed a significant reduction in the number of stack trace frames.

TaskCancelledExcpetion  with 10K threads ( Even the file size is giving away how bad the stack trace profileration problem was )

[TaskCancExceptionWithFix.log](https://github.com/user-attachments/files/19302320/TaskCancExceptionWithFix.log)

Filesize: 10 KB
  Standard Output: 
     With cache fix and 10K threads 4923736 bytes allocated in  (Gen0:0 1:0 2:0) Frames=2

[TaskCancExceptionWithoutFix.log](https://github.com/user-attachments/files/19302321/TaskCancExceptionWithoutFix.log)

Filesize: 1,168KB
Standard Output: 
     Without cache fix and 10K threads 5612888 bytes allocated in  (Gen0:0 1:0 2:0) Frames=2

TImeoutException  with 2K threads

[TimeoutExceptionWithFix.log](https://github.com/user-attachments/files/19302478/TimeoutExceptionWithFix.log)
[TimeoutExceptionWithoutFix.log](https://github.com/user-attachments/files/19302479/TimeoutExceptionWithoutFix.log)
 

## Type of change

## Closing issues


NaluTripician (commented):
Is there any unit testing we can do to verify change?

ananth7592:
> Is there any unit testing we can do to verify change?

This is more of a performance optimization and will be materialized as such in high scale scenarios. Hence, I performed a local repro with high concurrency and captured the performance improvements in the log attached.

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <param name="e">The exception to process.</param>
+        /// <exception cref="TaskCanceledException">Thrown if the input exception is a TaskCanceledException.</exception>
+        /// <exception cref="TimeoutException">Thrown if the input exception is a TimeoutException.</exception>
+        /// <exception cref="OperationCanceledException">Thrown if the input exception is an OperationCanceledException.</exception>
+        public static void CloneAndRethrowException(Exception e)
+        {
+            if (e is TaskCanceledException)
+            {
+                TaskCanceledException taskCanceledEx = new TaskCanceledException(e.Message, e.InnerException);
+                taskCanceledEx.Data["Message"] = e.Data["Message"];
```
Pilchie, line 30: Does this really need to be done?  Weird that it doesn't propagate from the constructor.

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            if (e is OperationCanceledException)
+            {
+                OperationCanceledException operationCanceledEx = new OperationCanceledException(e.Message, e.InnerException);
+                operationCanceledEx.Data["Message"] = e.Data["Message"];
+                throw operationCanceledEx;
+            }
+
+            // For all other exceptions, throw the original exception.
+            throw e;
```
Pilchie, line 49: Could rewrite the body of this whole method as 

``` csharp
        var newE = e switch
        {
            TaskCanceledException => new TaskCanceledException(e.Message, e),
            TimeoutException => new TimeoutException(e.Message, e),
            OperationCanceledException => new OperationCanceledException(e.Message, e),
            _ => e,
        };

        newE.Data["Message"] = e.Message;
        throw newE;
```

Pilchie (approved):
Looks fine to me.

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <summary>
+        /// Creates a shallow copy of specific exception types (e.g., TaskCanceledException, TimeoutException, OperationCanceledException) 
+        /// to prevent excessive stack trace growth and rethrows them. All other exceptions are propagated as-is.
+        /// </summary>
+        /// <param name="e">The exception to process.</param>
+        /// <exception cref="TaskCanceledException">Thrown if the input exception is a TaskCanceledException.</exception>
+        /// <exception cref="TimeoutException">Thrown if the input exception is a TimeoutException.</exception>
+        /// <exception cref="OperationCanceledException">Thrown if the input exception is an OperationCanceledException.</exception>
+        public static void CloneAndRethrowException(Exception e)
+        {
```
kirankumarkolli, line 26: Will it cover CosmosException and DocumentClientException?

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        /// <summary>
+        /// Creates a shallow copy of specific exception types (e.g., TaskCanceledException, TimeoutException, OperationCanceledException) 
+        /// to prevent excessive stack trace growth and rethrows them. All other exceptions are propagated as-is.
+        /// </summary>
+        /// <param name="e">The exception to process.</param>
+        /// <exception cref="TaskCanceledException">Thrown if the input exception is a TaskCanceledException.</exception>
+        /// <exception cref="TimeoutException">Thrown if the input exception is a TimeoutException.</exception>
+        /// <exception cref="OperationCanceledException">Thrown if the input exception is an OperationCanceledException.</exception>
+        public static void CloneAndRethrowException(Exception e)
```
kirankumarkolli, line 25: One of the proposal is to use the Object.MemberwiseClone and have it as interface on known exceptions, thoughts?

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
```diff
@@ -148,11 +148,14 @@ public async Task<TValue> GetAsync(
...
+            catch (Exception ex) when (object.ReferenceEquals(actualValue, newLazyValue))
             {
                 // If the lambda this thread added to values triggered an exception remove it from the cache.
-                this.TryRemoveValue(key, actualValue);
-                throw;
+                this.TryRemoveValue(key, actualValue);
+
+                // Creates a shallow copy of specific exception types to prevent stack trace proliferation 
+                // and rethrows them, while propagating all other exceptions unchanged.
+                ExceptionHandlingUtility.CloneAndRethrowException(ex);
```
kirankumarkolli, line 158: All new behaviors should be under a feature flag to safe flight.

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
ananth7592: @Pilchie , Ideally I expected the same. 
This Data field is actually a `System.Collections.ListDictionaryInternal` and upstream TransportHandler,  exceptions like TaskCanceledException or OperationCancedExceptions are transformed into 408s and is expecting a valid value in this Dict

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/5e1dd0ca51238caaab72f99a069c8cf534662bf4/Microsoft.Azure.Cosmos/src/Handler/TransportHandler.cs#L60

THe shallow copy was not copying over this Data field.

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
ananth7592: Yep, this makes sense and Object.MemberwiseClone being a protected method can be applied to only custom types which are defined by us. In this case, all the exceptions are default types. And hence I went ahead with static switching.

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
ananth7592: As I am given to understand, the typical cases uses for AsyncCache is to guard against results/calls from Gateway calls such as GetCollection, AddressRefresh .etc. THese calls either fail with Http exceptions or generic Timeout or Task canceled exception (Cancellation TOken based failures). In the upstream, we then convert these exceptions to DocumentClientException or COsmosException.  With respect to the next comment, if these exception types are required to be shallow copied, I think `Object.MemberWiseClone()` would be a good option

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+
+            if (e is TimeoutException)
+            {
+                TimeoutException timeoutEx = new TimeoutException(e.Message, e.InnerException);
+                timeoutEx.Data["Message"] = e.Data["Message"];
+                throw timeoutEx;
+            }
+
+            if (e is OperationCanceledException)
```
kundadebdatta, line 41: Would this be a good utilization and candidate for `switch-case` statements @ananth7592 ?

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
kundadebdatta: But the default value for the flag would be `True` correct ? And this can be deterministically disabled by setting the flag to `False`

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
ananth7592: initially I used this same switch expression as it made sense.

However, after conducting high-concurrency testing with 10,000+ threads, I observed that the switch expression resulted in higher GC allocations compared to the current static approach.

After analyzing the behavior and researching further (including AI-assisted insights), I found that:

Switch expressions create multiple temporary exception objects within the expression itself, increasing short-lived Gen0 allocations.
Under high contention scenarios, the switch expression contributed to GC pressure, leading to more frequent collections.

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
ananth7592: Added CosmosException to the list and made an interface change to do the cloning using `Object.MemberwiseClone`

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
ananth7592: Added DocumentClientException and CosmosException to the loop

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
jaredpar: > Switch expressions create multiple temporary exception objects within the expression itself, increasing short-lived Gen0 allocations.

Can you provide an example of this? In @Pilchie sample the only exceptions the compiler will sythensize are the ones on the RHS of the `=>` and those will only be created once when the condition is true. Creating multiple instances of these exceptions, or creating them before the condition was true, would be a pretty strong violation of the guarantees around `switch` evaluation.

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,56 @@
...
+    using System.Text;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Documents;
+
+    /// <summary>
+    /// Utility for post-processing of exceptions.
+    /// </summary>
+    internal static class ExceptionHandlingUtility
+    {
+        public const string ExceptionHandlingForStackTraceOptimizationEnabled = "AZURE_COSMOS_STACK_TRACE_OPTIMIZATION_ENABLED";
```
kirankumarkolli, line 18: Is it still needed?

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,56 @@
...
+                //TODO. Add support for shallow object clones method in the direct package.
+                DocumentClientException clonedDocumentClientEx = new DocumentClientException(
+                    dce.Message,
+                    dce.InnerException,
+                    dce.StatusCode);
+                clonedDocumentClientEx.Data["Message"] = e.Data["Message"];
+                throw clonedDocumentClientEx;
+            }
+
+            if (e is ICloneable ex)
```
kirankumarkolli, line 50: Move it up first and should be preferred if/when available

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            if (e is TimeoutException)
+            {
+                TimeoutException timeoutEx = new TimeoutException(e.Message, e.InnerException);
+                timeoutEx.Data["Message"] = e.Data["Message"];
+                throw timeoutEx;
+            }
+
+            if (e is DocumentClientException dce)
```
kirankumarkolli, line 39: remove? Once the master PR goes in then DCE is IClonable right?

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
```diff
@@ -148,11 +148,18 @@ public async Task<TValue> GetAsync(
...
-                throw;
+                this.TryRemoveValue(key, actualValue);
+
+                if (isStackTraceOptimizationEnabled)
+                {
+                    // Creates a shallow copy of specific exception types to prevent stack trace proliferation 
+                    // and rethrows them, doesn't process other exceptions.
+                    ExceptionHandlingUtility.CloneAndRethrowException(ex);
+                }
+                throw ex;
```
kirankumarkolli, line 162:
```suggestion
                throw;
```

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -114,9 +116,15 @@ public async Task<TValue> GetAsync(
...
-                    throw;
+                    }
+
+                    if (isStackTraceOptimizationEnabled)
+                    {
+                        // Creates a shallow copy of specific exception types to prevent stack trace proliferation 
+                        // and rethrows them, doesn't process other exceptions.
+                        ExceptionHandlingUtility.CloneAndRethrowException(e);
+                    }
+                    throw e;
```
kirankumarkolli, line 127:
```suggestion
                    throw;
```

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
ananth7592: Done

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
ananth7592: Done

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ExceptionHandlingUtilityTests.cs
```diff
@@ -0,0 +1,92 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos.Query.Core.Exceptions;
+    using Microsoft.Azure.Cosmos.Resource.CosmosExceptions;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [TestClass]
+    public class ExceptionHandlingUtilityTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(TaskCanceledException))]
```
dibahlfi, line 20: nit: may be we can use Assert on the method call that is suppose to throw exception rather than the whole test method.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ExceptionHandlingUtilityTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        [TestMethod]
+        public void DoNotClone_And_Rethrow_OtherException()
+        {
+            // Arrange
+            MalformedContinuationTokenException originalException =
+                new MalformedContinuationTokenException("malformed continuation token", new MalformedContinuationTokenException("Inner"));
+
+            // Act & Assert
+            ExceptionHandlingUtility.CloneAndRethrowException(originalException);
```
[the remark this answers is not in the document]
ananth7592: That's right, the utility function is a no-op for these unsupported exceptions. I modified the test to fail if there is an exception thrown

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ExceptionHandlingUtilityTests.cs
ananth7592: So I am doing a two level assert, which is the asserts inside when the exception is caught and second level assert is on the exception type.
This test construct of `ExpectedException` is at a test method level

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,44 @@
...
+        public static void CloneAndRethrowException(Exception e)
+        {
+            if (e is ICloneable ex)
+            {
+                throw (Exception)ex.Clone();
+            }
+
+            if (e is TaskCanceledException)
+            {
+                TaskCanceledException taskCanceledEx = new TaskCanceledException(e.Message, e.InnerException);
```
kirankumarkolli, line 31:
```suggestion
                TaskCanceledException taskCanceledEx = new TaskCanceledException(e.Message, e);

```

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,44 @@
...
+                taskCanceledEx.Data["Message"] = e.Data["Message"];
+                throw taskCanceledEx;
+            }
+
+            if (e is TimeoutException)
+            {
+                TimeoutException timeoutEx = new TimeoutException(e.Message, e.InnerException);
```
kirankumarkolli, line 38:
```suggestion
                TimeoutException timeoutEx = new TimeoutException(e.Message, e);

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ExceptionHandlingUtilityTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            // Act & Assert
+            try
+            {
+                ExceptionHandlingUtility.CloneAndRethrowException(originalException);
+            }
+            catch (TaskCanceledException ex)
+            {
+                Assert.AreNotSame(originalException, ex); // Ensure a new exception was created
+                Assert.AreEqual(originalException.InnerException.ToString(), ex.InnerException.ToString());
```
kirankumarkolli, line 34: Prefer Object.ReferenceEquals

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -119,8 +119,14 @@ public async Task<TValue> GetAsync(
...
                             removed,
                             e);
-                    }
-
+                    }
+
+                    if (this.enableAsyncCacheExceptionNoSharing)
+                    {
+                        // Creates a shallow copy of specific exception types to prevent stack trace proliferation 
+                        // and rethrows them, doesn't process other exceptions.
```
kirankumarkolli, line 127: UT coverage?

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
```diff
@@ -154,10 +153,17 @@ public async Task<TValue> GetAsync(
...
+            catch (Exception ex) when (object.ReferenceEquals(actualValue, newLazyValue))
             {
                 // If the lambda this thread added to values triggered an exception remove it from the cache.
                 this.TryRemoveValue(key, actualValue);
+
+                if (this.enableAsyncCacheExceptionNoSharing)
+                {
+                    // Creates a shallow copy of specific exception types to prevent stack trace proliferation 
+                    // and rethrows them, doesn't process other exceptions.
+                    ExceptionHandlingUtility.CloneAndRethrowException(ex);
```
kirankumarkolli, line 165: UT coverage?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ExceptionHandlingUtilityTests.cs
ananth7592: Done.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
ananth7592: done

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
ananth7592: Done.

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
```diff
@@ -154,10 +153,17 @@ public async Task<TValue> GetAsync(
...
+                }
```
kirankumarkolli, line 166:
```suggestion
                }
                
```

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -119,8 +119,14 @@ public async Task<TValue> GetAsync(
...
+                        ExceptionHandlingUtility.CloneAndRethrowException(e);
+                    }
```
kirankumarkolli, line 129:
```suggestion
                    }

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AsyncCacheTest.cs
```diff
@@ -306,6 +305,54 @@ await Task.Factory.StartNew(() =>
             TaskCreationOptions.None,
             new SingleTaskScheduler()
             );
+        }
+
+        [TestMethod]
+        public async Task GetAsync_Assert_OnExceptionPostProcessing()
+        {
+            AsyncCache<string, string> cacheDefault = new AsyncCache<string, string>();
```
kirankumarkolli, line 313: How about make the test parameterize and use it to validate both behaviors?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AsyncCacheTest.cs
```diff
@@ -306,6 +305,54 @@ await Task.Factory.StartNew(() =>
...
+        }
+
+        [TestMethod]
+        public async Task GetAsync_Assert_OnExceptionPostProcessing()
+        {
+            AsyncCache<string, string> cacheDefault = new AsyncCache<string, string>();
+
+            // Arrange
+            string key = "testKey";
+            Exception testException = new TimeoutException("Simulated timeout exception");
```
kirankumarkolli, line 317: Also exception can be paramaterized as well, to try different types

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
```diff
@@ -154,10 +153,17 @@ public async Task<TValue> GetAsync(
             {
                 return await generator;
             }
-            catch (Exception) when (object.ReferenceEquals(actualValue, newLazyValue))
+            catch (Exception ex) when (object.ReferenceEquals(actualValue, newLazyValue))
```
kirankumarkolli, line 156: Is it expected to be limited to the conditional catch only?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AsyncCacheTest.cs
ananth7592: done

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
```diff
@@ -154,10 +153,17 @@ public async Task<TValue> GetAsync(
...
             {
                 // If the lambda this thread added to values triggered an exception remove it from the cache.
                 this.TryRemoveValue(key, actualValue);
+
+                if (this.enableAsyncCacheExceptionNoSharing)
```
kirankumarkolli, line 161: Also pending E2E test

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AsyncCacheTest.cs
ananth7592: Done

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
Pilchie: @ananth7592 - can you take another look here? Can you share a trace, or the IL that shows what additional allocations are happening?  @jaredpar is from the C# compiler team, so I'd expect that there is something else going on.

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
ananth7592: Done. Added a test to emulator based ClientTests.cs

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
ananth7592: Yes, Adding another Catch(Exception ex) to handle it when a non conditional exception is thrown

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
ananth7592: @jaredpar and @Pilchie   My original hypothesis regarding high Gen0 allocations was observed due to reasons that are not causally related to using this switch statement. Hence I am unable to reproduce the same behaviour that I observed with this switch statement.

I switched to using this "switch" statement :D  . Thank for you reviewing .

Pilchie: Thanks!

Pilchie (commented):
What happened with the `DotNetSDKApi.json` changes?

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            Exception ex = e switch
+            {
+                ICloneable cloneableEx => (Exception)cloneableEx.Clone(),
+                TaskCanceledException taskCanceledEx => AddMessageData(new TaskCanceledException(taskCanceledEx.Message, taskCanceledEx), e),
+                TimeoutException timeoutEx => AddMessageData(new TimeoutException(timeoutEx.Message, timeoutEx), e),
+                _ => null
+            };
+
+            if (ex is not null)
```
kirankumarkolli, line 32: nit: modeling it as TryClone and the caller doing a throw is preferred

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
```diff
@@ -154,12 +153,30 @@ public async Task<TValue> GetAsync(
...
-            }
+            }
+            catch (Exception ex)
+            {
+                if (this.enableAsyncCacheExceptionNoSharing)
+                {
+                    // Creates a shallow copy of specific exception types to prevent stack trace proliferation
+                    // and rethrows them, doesn't process other exceptions.
+                    ExceptionHandlingUtility.CloneAndRethrowException(ex);
+                }
```
kirankumarkolli, line 176:
```suggestion
                }

```

Microsoft.Azure.Cosmos/src/Routing/AsyncCache.cs
```diff
@@ -154,12 +153,30 @@ public async Task<TValue> GetAsync(
...
             {
                 // If the lambda this thread added to values triggered an exception remove it from the cache.
                 this.TryRemoveValue(key, actualValue);
+
+                if (this.enableAsyncCacheExceptionNoSharing)
+                {
+                    // Creates a shallow copy of specific exception types to prevent stack trace proliferation 
+                    // and rethrows them, doesn't process other exceptions.
+                    ExceptionHandlingUtility.CloneAndRethrowException(ex);
+                }
```
kirankumarkolli, line 166:
```suggestion
                }
                
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -938,6 +938,86 @@ public async Task MultiRegionAccountTest()
...
+                        await container.CreateItemAsync(testItem);
+                    }));
+                }
+
+                // Signal all tasks to start
+                tcs.SetResult(null);
+
+                // Wait for all tasks to complete (they should all fail)
+                await Task.WhenAll(createTasks);
+
```
kirankumarkolli, line 1000:
```suggestion


```

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
ananth7592: Great suggestion.Done

ananth7592:
@Pilchie , The change that resulted in adding support for `ICloneable` for CosmosException and CosmosOperationCanceledException was a contract change and hence had to generate new contract JSON

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -938,6 +938,84 @@ public async Task MultiRegionAccountTest()
...
+            Cosmos.Database db = null;
+            try
+            {
+                CosmosClient cosmosClient = TestCommon.CreateCosmosClient(clientOptions: cosmosClientOptions);
+
+                db = await cosmosClient.CreateDatabaseIfNotExistsAsync("TimeoutFaultTest");
+                Container container = await db.CreateContainerIfNotExistsAsync("TimeoutFaultContainer", "/pk");
+
+                // Shared TaskCompletionSource ensures all tasks start together
+                TaskCompletionSource<object> tcs = new(TaskCreationOptions.RunContinuationsAsynchronously);
```
kirankumarkolli, line 980: Why is it needed? Even with-out it should do right?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ExceptionHandlingUtilityTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            // Arrange
+            MalformedContinuationTokenException originalException =
+                new MalformedContinuationTokenException("malformed continuation token", new MalformedContinuationTokenException("Inner"));
+
+            // Act & Assert
+            bool result = ExceptionHandlingUtility.TryCloneException(originalException, out _);
+
+            Assert.IsFalse(result, "Expected exception to not be cloned.");
+
```
kirankumarkolli, line 87:
```suggestion
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
ananth7592: I use this to TCS as a way to wait and proceed until all the threads are created and trigger the `CreateItem1` call asynchronously

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Routing/AsyncCacheNonBlockingTests.cs
```diff
@@ -530,6 +530,52 @@ public async Task Refresh_WhenThrowsOtherException_ShouldNotRemoveKeyFromTheCach
...
+        {
+            // Arrange
+            Exception testException = new TimeoutException("Simulated timeout exception");
+            CancellationToken cancellationToken = CancellationToken.None;
+
+            AsyncCacheNonBlocking<string, string> asyncCache = new AsyncCacheNonBlocking<string, string>(enableAsyncCacheExceptionNoSharing: enabled);
+
+            Random random = new Random();
+            List<Task> tasks = new List<Task>();
+            for (int i = 0; i < 50; i++)
```
kirankumarkolli, line 548: This test is different from AsyncCacheTest, is that expected?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
ananth7592: Ideally, we could achieve this with a semaphore as well but I chose to use TCS as I was using the same for high concurrency testing

Pilchie (approved):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ExceptionHandlingUtilityTests.cs
ananth7592: Done

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Routing/AsyncCacheNonBlockingTests.cs
ananth7592: Yes, this is expected. This is done to trigger the AsyncNonBlockingCache codepath to hit the exception during GetAsync scenario.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
ananth7592: Modified the approach to block the `tcs` during the sendEventHandler to be more realistic in terms of scenario that we are trying to test. Here is the upgraded scenario of the test

* a TaskCompletionSource is created to track and synchronize tests
* All tasks have only one job i.e. to createItem and are created regularly 
* The sendEvent HTTP presend handler does updates to the Tcs
* If the number of tasks entering the handler are less than the full count, the tcs is blocked
* After all the tasks enter the handler, the tcs is signaled to start execution
* This will simulate waits in AsyncCache and we expect shallow copies of the exception when the feature is enabled

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ClientTests.cs
```diff
@@ -937,6 +937,85 @@ public async Task MultiRegionAccountTest()
...
+                SendingRequestEventArgs = async (sender, e) =>
+                {
+                    if (e.IsHttpRequest())
+                    {
+                        string endWith = "partitionKeyRangeIds=0";
+                        if (e.HttpRequest.Method == HttpMethod.Get &&
+                            e.HttpRequest.RequestUri.OriginalString.EndsWith(endWith))
+                        {
+                            // Wait until all expected threads are in the handler
+                            if (Interlocked.Increment(ref enteredHandlers) == expectedHandlers)
```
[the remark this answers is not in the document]
ananth7592: THe check for URIs ending with `string endWith = "partitionKeyRangeIds=0";` ensures that the address calls will be same as incoming requests

Microsoft.Azure.Cosmos/src/ExceptionHandlingUtility.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>
+        /// Tries to create a shallow copy of specific exception types (e.g., TaskCanceledException, TimeoutException, OperationCanceledException)
+        /// to prevent excessive stack trace growth. Returns true if the exception was cloned, otherwise false.
+        /// </summary>
+        public static bool TryCloneException(Exception e, out Exception clonedException)
+        {
+            clonedException = e switch
+            {
+                ICloneable cloneableEx => (Exception)cloneableEx.Clone(),
+                TaskCanceledException taskCanceledEx => AddMessageData(new TaskCanceledException(taskCanceledEx.Message, taskCanceledEx), e),
```
kirankumarkolli, line 27: Thoughts in changing this to OperationCancelledException? (which will cover TaskCancellationException)

ananth7592: Done. Added support for all OperationCanceledException base types

kirankumarkolli (dismissed):

kirankumarkolli (approved):

kundadebdatta (approved):
LGTM.

PR #5110, opened 2025-04-04
[Internal] Thin Client Integration: Fixes the ThinClientTransportSerializer to avoid creating memory stream.

### LINKED ISSUE #5070: [Thin Client Integration] Optimize the `ThinClientTransportSerializer` to Avoid Creating Memory Stream
Today, the `ThinClientTransportSerializer.SerializeProxyRequestAsync()` uses `MemoryStream` to serialize the transport request payload.

**Acceptance Criteria:**

Optimize the serialization logic by avoiding the buffer writing directly to memory stream.

```
            using Documents.Rntbd.TransportSerialization.SerializedRequest serializedRequest =
                Documents.Rntbd.TransportSerialization.BuildRequestForProxy(request,
                new ResourceOperation(operationType, resourceType),
                activityId,
                bufferProvider.Provider,
                accountName,
                out _,
                out _);

            // TODO: consider using the SerializedRequest directly.
            MemoryStream memoryStream = new MemoryStream(serializedRequest.RequestSize);
            await serializedRequest.CopyToStreamAsync(memoryStream);
            memoryStream.Position = 0;
            return memoryStream;
```

aavasthy:
# Pull Request Template

## Description

Move the TransportSerializer from shared files to v3 repo and Optimize the ThinClientTransportSerializer to Avoid Creating Memory Stream everytime.

## Type of change

## Closing issues

To automatically close an issue: closes #5070


Microsoft.Azure.Cosmos/src/ProxyRequest.cs
```diff
@@ -0,0 +1,156 @@
...
+    using System.Collections.Generic;
+    using Microsoft.Azure.Documents;
+
+    internal class ProxyRequest
+    {
+        private static readonly int[] expectedPositionsForTokensDuringReorder;
+        private static readonly int[] positionsOfTokensToMakeOptionalAfterReorder;
+        private static readonly Action<RntbdToken, RntbdConstants.Request>[] settersForFieldsToMakeOptional;
+
+        static ProxyRequest()
```
FabianMeiswinkel, line 17: NIT: I would usually expect to see a new Test file ProxRequestTests with narrow-scoped unit tests for the logic below.

Microsoft.Azure.Cosmos/src/ThinClientMemoryStreamPool.cs
```diff
@@ -0,0 +1,43 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos
+{
+    using System.Collections.Concurrent;
```
FabianMeiswinkel, line 6: What motivation lead to creating a MemoryStream pool vs. just pooling the underlying byte-arrays? This is an honest question - not an ask to change anything. Just trying to understand the thought process.

Microsoft.Azure.Cosmos/src/TransportSerialization.cs
```diff
@@ -0,0 +1,2764 @@
...
+    using Microsoft.Azure.Cosmos.Rntbd;
+    using Microsoft.Azure.Documents;
+    using Microsoft.Azure.Documents.Collections;
+    using Microsoft.Azure.Documents.Rntbd;
+    using RequestPool = Microsoft.Azure.Documents.RntbdConstants.RntbdEntityPool<Microsoft.Azure.Documents.RntbdConstants.Request, Microsoft.Azure.Documents.RntbdConstants.RequestIdentifiers>;
+
+    internal static class TransportSerialization
+    {
+        // Path format
+        internal static readonly char[] UrlTrim = { '/' };
```
FabianMeiswinkel, line 21: Similar - just a question - why shade/copy all the code instead of keep using the helper functions in direct stack? And just keep the ThinClient specific stuff (ordered headers) in V3 code?

If this means we get significantly better isolation from msdata in V3 SDK this might be worth it - but it is a lot of code duplication - and given that we can't cut ties with msdata (which would IMO justify code duplication for sure) I think we need to understand teh trade-offs.

FabianMeiswinkel (changes requested):
Mostly NITs and questions only - but for the questions I want to be able to form an opinion after review/discussion - so, blocking the PR for now to force these conversations only.

Microsoft.Azure.Cosmos/src/TransportSerialization.cs
aavasthy: Wanted to confirm if we need this change of moving TransportSerialization logic to V3 code? becausew e will continue to have dependency on this class as Direct mode uses the same class for generating request. And direct package does not refer Friends.Proxy so if we eventually move TransportSerializer to that class it wont work.

Also, I tried copying only the methods that are needed for building proxy request but it needs the methods for adding headers to rntbd request and almost the entire class code constitutes of those private headers.

Microsoft.Azure.Cosmos/src/ThinClientMemoryStreamPool.cs
aavasthy: I believe pooling a full MemoryStream object will save creating and initializing a new MemoryStream on every call and pooling the entire MemoryStream will avoid repeated object construction overhead. But with pooling byte arrays, we still have to instantiate a fresh MemoryStream to wrap that array each time. 

Also did a perf comparison using dotnet benchmark tool.

| Method             | Mean      | Error    | StdDev   | P95       | P100      | Op/s     | Allocated |
|-------------------|----------:|---------:|---------:|----------:|----------:|---------:|----------:|
| PooledMemoryStream | 104.24 µs | 4.819 µs | 1.251 µs | 105.32 µs | 105.36 µs |  9,593.1 |         - |
| PooledByteArray    |  97.00 µs | 4.318 µs | 1.121 µs |  98.51 µs |  98.82 µs | 10,309.6 |         - |

FabianMeiswinkel: 7 nanoseconds is nothing. IMO using pooled byte[] is preferrable because that exists as a CLR7.Net FX pooled object - it simply requires less code overhead. But I won't block on this - if you make a different call and can explain why you made the decision I think that is fine.

Microsoft.Azure.Cosmos/src/TransportSerialization.cs
FabianMeiswinkel: My 2 cents - I think this is too mcuh code copying - we should probably wait with this refactoring until we cut the tight coupling with sharedFiles altogether

aavasthy:
Not moving the TransportSerialization logic to v3 as of now. However, following up on the memory stream optimization issue in this PR https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5137

PR #5120, opened 2025-04-10
Query: Adds an environment variable for disabling the hybrid search query plan optimization

neildsh:
## Description
This change adds a configuration whose value will be read from the environment variable `AZURE_COSMOS_HYBRID_SEARCH_QUERYPLAN_OPTIMIZATION_DISABLED`. If this evaluates to true, then the optimization for skipping the order by rewrite for hybrid search is disabled.

The environment variable is read, and the value is propagated as an internally visible query request option. This is plumbed down in the query plan APIs where it is either used to populate the supported query features header in case of a gateway or it is used as a flag passed into ServiceInterop.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
+++ b/Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// Environment variable name to disable sending non streaming order by query feature flag to the gateway.
         /// </summary>
-        internal static readonly string NonStreamingOrderByQueryFeatureDisabled = "AZURE_COSMOS_NON_STREAMING_ORDER_BY_FLAG_DISABLED";
+        internal static readonly string HybridSearchQueryPlanOptimizationDisabled = "AZURE_COSMOS_HYBRID_SEARCH_QUERYPLAN_OPTIMIZATION_DISABLED";
 
         /// <summary>
         /// Environment variable name to enable distributed query gateway mode.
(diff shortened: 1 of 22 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/NonStreamingOrderByQueryTests.cs
```diff
@@ -339,7 +339,7 @@ public static void TraceSupportedFeaturesTestCase(QueryFeaturesTest testCase)
             {
                 if (Enabled)
                 {
-                    Trace.WriteLine($"EnvironmentVariable: {testCase.EnvironmentVariable}, RequestOption: {testCase.RequestOption}, ExpectNonStreamingOrderBy: {testCase.ExpectNonStreamingOrderBy}");
+                    Trace.WriteLine($"EnvironmentVariable: {testCase.EnvironmentVariable}, RequestOption: {testCase.RequestOption}, ExpectNonStreamingOrderBy: {testCase.ExpectHybridSearchOptimizationDisabled}");
```
adityasa, line 342: >ExpectNonStreamingOrderBy [](http://example.com/codeflow?start=132&length=25)

Consider renaming.

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
```diff
@@ -246,15 +246,15 @@ public static bool IsOptimisticDirectExecutionEnabled(
         }
 
         /// <summary>
-        /// Gets the boolean value indicating whether the non streaming order by query feature flag should be sent to the gateway
+        /// Gets the boolean value indicating whether the hybrid search query plan optimization feature flag should be sent to the gateway
         /// based on the environment variable override.
         /// </summary>
-        public static bool IsNonStreamingOrderByQueryFeatureDisabled(
```
adityasa, line 252: >IsNonStreamingOrderByQueryFeatureDisabled [](http://example.com/codeflow?start=27&length=41)

Just to confirm, the purpose of removing old flag is that the feature is graduated and we don't need this anymore?

adityasa (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
neildsh: Yes, this is correct

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/NonStreamingOrderByQueryTests.cs
neildsh: fixed

Maya-Painter (approved):

adityasa (approved):
:shipit:

kirankumarkolli (approved):

PR #5121, opened 2025-04-10
Query: Adds support for the optimized query plan that skips the order by rewrite

neildsh:
## Description

Adds support for the optimized query plan that skips the order by rewrite. The new hybrid search component query plan "lies" to the SDK about there being no ORDER BY on the component queries. However, the hybrid search query pipeline stage knows that the component scores need to be sorted. Since there is no information about this sorting on the query plan anymore, the hybrid search query pipeline stage will always sort descending in this scenario. The component score projections are rewritten to safeguard this assumption.

This change is backward compatible so that the SDK can accept both the old style query plan with rewritten ORDER BY as well as the new one that skips the order by rewrite.

The end user can disable the optimized query plan by setting the environment variable `AZURE_COSMOS_HYBRID_SEARCH_QUERYPLAN_OPTIMIZATION_DISABLED` flag to `true`.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
@@ -404,12 +404,10 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.CrossPartition.HybridSearch
             for (int index = 0; index < hybridSearchQueryInfo.ComponentQueryInfos.Count; ++index)
             {
                 QueryInfo queryInfo = hybridSearchQueryInfo.ComponentQueryInfos[index];
-                Debug.Assert(queryInfo.HasOrderBy, "The component query should have an order by");
-                Debug.Assert(queryInfo.HasNonStreamingOrderBy, "The component query is a non streaming order by");
-                Debug.Assert(queryInfo.OrderBy.Count == 1, "The component query should have exactly one order by expression");
+                SortOrder sortOrder = queryInfo.HasOrderBy ? queryInfo.OrderBy[0] : SortOrder.Descending;
 
                 double componentWeight = useDefaultComponentWeight ? 1.0 : hybridSearchQueryInfo.ComponentWeights[index];
-                result.Add(new ComponentWeight(componentWeight, queryInfo.OrderBy[0]));
+                result.Add(new ComponentWeight(componentWeight, sortOrder));
             }
 
             return result;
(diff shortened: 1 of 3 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/NonStreamingOrderByQueryTests.cs
```diff
@@ -1056,7 +1256,7 @@ private async Task SplitMergeAsync()
         private static class DebugTraceHelpers
         {
 #pragma warning disable CS0162, CS0649 // Unreachable code detected
-            private static readonly bool Enabled;
+            private static readonly bool Enabled = true;
```
sboshra, line 1260: Is this meant to be set to true?

sboshra (dismissed):
:shipit:

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
```diff
@@ -404,12 +404,10 @@ private static IReadOnlyList<ComponentWeight> ExtractComponentWeights(HybridSear
             for (int index = 0; index < hybridSearchQueryInfo.ComponentQueryInfos.Count; ++index)
             {
                 QueryInfo queryInfo = hybridSearchQueryInfo.ComponentQueryInfos[index];
-                Debug.Assert(queryInfo.HasOrderBy, "The component query should have an order by");
```
adityasa, line 407: >queryInfo.HasOrderBy [](http://example.com/codeflow?start=29&length=20)

Is there a value in preserving this assert and making it tighter?

queryInfo.HasOrderBy == !QueryFeatures.Contains(OptimizedQueryPlan)

Although, I am not sure how easy it is to access query features here (and/or they come into play for interop).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/NonStreamingOrderByQueryTests.cs
neildsh: No, it isn't. Good catch!

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Query/Pipeline/NonStreamingOrderByQueryTests.cs
```diff
@@ -321,6 +321,69 @@ public async Task HybridSearchTests()
             }
         }
 
+        [TestMethod]
+        public async Task HybridSearchSkipOrderByRewriteTests()
```
adityasa, line 325: Curious, is the new plan on by default? Or it will be done as a separate change?

neildsh: It is on for the .NET SDK by default. One can use an env variable to disable it.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
neildsh: We could check if the hybrid search query plan optimization was disabled, but that is not sufficient. The query plan that is used a function of both the SDK and the gateway that it is talking to. For example, a newer SDK may be talking to an older gateway that still returns the plan with the rewritten order by. So we have to allow both here.

Maya-Painter (approved):

adityasa (approved):
:shipit:

PR #5128, opened 2025-04-12
Session Consistency: Adds SessionTokenMismatchRetryPolicy optimization through customer supplied region switch hints

dibahlfi:
This PR will allow application developers to configure hints through a SessionRetryOptions instance which will signal to the SDK whether to pin retries on the local region or move quicker to a remote region when READ_SESSION_NOT_AVAILABLE errors are thrown.
[] New feature (non-breaking change which adds functionality)


Microsoft.Azure.Cosmos/src/SessionRetryOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// Sets the maximum number of retries within each region for read and write operations. The minimum value is 1 - the backoff time for the last in-region retry will ensure that the total retry time within the
+        /// region is at least the min. in-region retry time.
+        /// </summary>
+        public int MaxInRegionRetryCount { get; private set; }
+
+        /// <summary>
+        /// hints which guide SDK-internal retry policies on how early to switch retries to a different region. If true, will retry all replicas once and add a minimum delay before switching to the next region.If false, it will
+        /// retry in the local region up to 5s
+        /// </summary>
+        public bool RemoteRegionPreferred { get; set; } = false;
```
FabianMeiswinkel, line 40: This should be captured from Configurationmanager as well, correct? Otherwise it would never actually change to the new session retry behavior. I think you have to expose an envrionment variable to opt-into the new behaior.

FabianMeiswinkel: Never mind - this is public API.

FabianMeiswinkel (approved):
LGTM - Thanks

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -120,7 +121,9 @@ public string ApplicationName
         /// <summary>
         /// Get or set session container for the client
         /// </summary>
-        internal ISessionContainer SessionContainer { get; set; }
+        internal ISessionContainer SessionContainer { get; set; }
+
+        internal SessionRetryOptions SessionRetryOptions { get; private set; }
```
kundadebdatta, line 126: nit: Let's add a `<summary>` section for `SessionRetryOptions`.

dibahlfi: done

kundadebdatta (approved):

FabianMeiswinkel (approved):
LGTM - Thx

PR #5141, opened 2025-04-21
[Internal] Thin Client Integration: Adds HTTP Client Sharing between Gateway and Thin Client

### LINKED ISSUE #4579: [Thin Client Integration] Analyze and Implement HTTP Client Sharing between Gateway and Thin Client
`HttpClient` is capable of handling both `Http1.1` and `Http2.0` protocol. Request will decide which protocol to use (Request can opt for Http2.0 explicitly). 

Therefore, ideally it should be possible to share the same client with both the (Gateway and Thin Client) store models.

aavasthy:
# Pull Request Template

## Description

Adding HttpClient sharing logic between Gateway and Thinclient model. HttpClient is capable of handling both Http1.1 and Http2.0 protocol. The request will decide which protocol to use based on the connection mode.

Therefore, ideally it should be possible to share the same client with both the (Gateway and Thin Client) store models.

## Type of change

## Closing issues

To automatically close an issue: closes #4579


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
```diff
@@ -668,6 +668,54 @@ public async Task NonPartitionKeyLookupCacheTest(bool binaryEncodingEnabledInCli
...
+        [TestMethod]
+        [DataRow(true, DisplayName = "Test scenario when binary encoding is enabled at client level.")]
+        [DataRow(false, DisplayName = "Test scenario when binary encoding is disabled at client level.")]
+        public async Task HttpRequestVersionIsOnePointOneWhenUsingGatewayMode(bool binaryEncodingEnabledInClient)
+        {
+            try
+            {
+                if (binaryEncodingEnabledInClient)
+                {
+                    Environment.SetEnvironmentVariable(ConfigurationManager.BinaryEncodingEnabled, "True");
```
kundadebdatta, line 682: Do we need to validate binary encoding for thin client mode ?

aavasthy: cleaned up the code and removed any binary encoding references from these tests. Added a new tests for validating that http request version is 1.1 during gateway mode and 2.0 during thinclient mode for create item operation.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
```diff
@@ -668,8 +668,94 @@ public async Task NonPartitionKeyLookupCacheTest(bool binaryEncodingEnabledInCli
...
+                Cosmos.Database database = await client.CreateDatabaseIfNotExistsAsync("HttpVersionTestDb");
+                Container container = await database.CreateContainerIfNotExistsAsync("HttpVersionTestContainer", "/pk");
+
+                ToDoActivity testItem = ToDoActivity.CreateRandomToDoActivity();
+
+                try
+                {
+                    await container.CreateItemAsync(testItem, new Cosmos.PartitionKey(testItem.pk));
+                }
+                catch (CosmosException)
```
kundadebdatta, line 740: Remove the `try` and `catch` block and refactor the code to something like the below:

```
await Assert.ThrowsExceptionAsync<CosmosException>(() => container.CreateItemAsync(testItem, new Cosmos.PartitionKey(testItem.pk)));
```

aavasthy: updated

FabianMeiswinkel (approved):
LGTM - I was initially sceptical whether sharing HttpClient between normal Gateway (possibly HTTP 1.1) and ThinClient is a good idea. But trying to poke holes in that idea I don't see any obvious problem. If customer configures HttpClient to not allow HTTP/2 seeing failures for ThinClient is expected i guess.

kundadebdatta (approved):

aavasthy:
> LGTM - I was initially sceptical whether sharing HttpClient between normal Gateway (possibly HTTP 1.1) and ThinClient is a good idea. But trying to poke holes in that idea I don't see any obvious problem. If customer configures HttpClient to not allow HTTP/2 seeing failures for ThinClient is expected i guess. You could mess-up the normal Gateway HttpClient if something goes wrong with ThinClient - but this is something we can properly test.

Tested these changes with both gateway mode and thinclient mode by enabling and disabling thinclient flag. For regression, I am relying on existing pipeline tests we have. Is there anything else we should cover as part of testing this change?

kirankumarkolli (approved):

PR #5145, opened 2025-04-24
Query: Fixes handling of undefined projections in hybrid search

neildsh:
## Description
This changes allows the projections in a hybrid search query to be undefined, bringing the behavior of hybrid search in line with other base pipelines such as the order by pipeline stage and parallel cross partition pipeline stage.

## Type of change


adityasa (approved):
:shipit:

leminh98 (approved):

Issue #5164, opened 2025-05-01
[Per Partition Automatic Failover] - Remove Mandate on Application Region or Application Preferred Regions


kundadebdatta:
**Acceptance Criteria:**

Now that we have the effective preferred region introduced in the .NET SDK, we don't need the validation on mandating `ApplicationRegion` or `ApplicationPreferredRegions`. Therefore, please remove the checks with PPAF.

PR #5165, opened 2025-05-02
[Internal] PPAF: Adds Code to Enable Read Hedging By Default

### LINKED ISSUE #5047: [Per Partition Automatic Failover] - Enable Hedging with Default Threshold of 1 Sec Ehen PPAF is Enabled
**Background:**

We reached to an agreement that when PPAF is enabled, the SDK will opt-in (by default) a cross region availability strategy (hedging) with the following principles:

- An initial threshold limit of one second to get started and tune it afterwards.
- Expose a mechanism to opt-out from hedging, if needed. 
- For the upcoming build session, we should target the read hedging. Post build, default hedging should be applicable for writes as well.

**Acceptance Criteria:**

- Make the above changes in the SDK to enable hedging by default.

kundadebdatta:
# Pull Request Template

## Description

This PR does the following two things:

- Enables Cross Regional Hedging by default with a threshold of either `1` second or half of the request timeout value provided, whichever is minimum, when PPAF is enabled. Note that the default availability strategy can be opted out by setting the `AvailabilityStrategy` to `DisabledAvailabilityStrategy` in the `CosmosClientOptions`. To understand this better, please take a look at the example below:

            CosmosClientOptions cosmosClientOptions = new CosmosClientOptions()
            {
                EnablePartitionLevelFailover = true,
                ConsistencyLevel = Cosmos.ConsistencyLevel.Strong,
                AvailabilityStrategy = new DisabledAvailabilityStrategy(),
            };

- As of today, during the cosmos client initialization, with PPAF enabled we ensure that either the `ApplicationRegion` or `ApplicationPreferredRegions` is present through a validation. Since we now have established the concept of Effective Preferred Region, this PR removes the above mandatory validation on `ApplicationRegion` or `ApplicationPreferredRegions`, as in the if there are no preferred regions passed explicitly, the .NET SDK can populate the effective preferred regions.

## Type of change

## Closing issues

To automatically close an issue: closes #5047, #5164

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -92,6 +92,7 @@ namespace Microsoft.Azure.Cosmos
             this.CustomHandlers = new Collection<RequestHandler>();
             this.CosmosClientTelemetryOptions = new CosmosClientTelemetryOptions();
             this.SessionRetryOptions = new SessionRetryOptions();
+            this.InitializePartitionLevelFailoverWithDefaultHedging();
         }
 
         /// <summary>
@@ -771,17 +772,7 @@ namespace Microsoft.Azure.Cosmos
         internal bool EnablePartitionLevelFailover
         {
             get => this.isPartitionLevelFailoverEnabled;
-            set
-            {
-                if (this.AvailabilityStrategy == null && !ConfigurationManager.IsDefaultHedgingDisabledWithPartitionLevelFailover())
-                {
-                    this.AvailabilityStrategy = AvailabilityStrategy.CrossRegionHedgingStrategy(
-                        threshold: TimeSpan.FromMilliseconds(1000),
-                        thresholdStep: TimeSpan.FromMilliseconds(1));
-                }
-
-                this.isPartitionLevelFailoverEnabled = value || ConfigurationManager.IsPartitionLevelFailoverEnabled(defaultValue: false);
-            }
+            set => this.InitializePartitionLevelFailoverWithDefaultHedging(isPPafEnabled: value);
         }
 
         /// <summary>
@@ -1267,6 +1258,21 @@ namespace Microsoft.Azure.Cosmos
                         features: featureString,
                         regionConfiguration: regionConfiguration,
                         suffix: this.GetUserAgentSuffix());
+        }
+
+        internal void InitializePartitionLevelFailoverWithDefaultHedging(
+            bool isPPafEnabled = false)
+        {
+            this.isPartitionLevelFailoverEnabled = isPPafEnabled || ConfigurationManager.IsPartitionLevelFailoverEnabled(defaultValue: false);
+
+            if (this.isPartitionLevelFailoverEnabled
+                && this.AvailabilityStrategy == null
+                && !ConfigurationManager.IsDefaultHedgingDisabledWithPartitionLevelFailover())
+            {
+                this.AvailabilityStrategy = AvailabilityStrategy.CrossRegionHedgingStrategy(
+                    threshold: TimeSpan.FromMilliseconds(1000),
+                    thresholdStep: TimeSpan.FromMilliseconds(10));
+            }
         }
 
         internal string GetUserAgentSuffix()
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -57,6 +57,12 @@ namespace Microsoft.Azure.Cosmos
         private const ApiType DefaultApiType = ApiType.None;
 
         /// <summary>
+        /// Default thresholds for PPAF request hedging.
+        /// </summary>
+        private const int DefaultHedgingThresholdInMilliseconds = 1000;
+        private const int DefaultHedgingThresholdStepInMilliseconds = 500;
+
+        /// <summary>
         /// Default request timeout
         /// </summary>
         private int gatewayModeMaxConnectionLimit;
@@ -1263,11 +1269,11 @@ namespace Microsoft.Azure.Cosmos
             {
                 // The default threshold is the minimum value of 1 second and a fraction (currently it's half) of
                 // the request timeout value provided by the end customer.
-                double defaultThresholdInMillis = Math.Min(1000, this.RequestTimeout.TotalMilliseconds / 2);
+                double defaultThresholdInMillis = Math.Min(CosmosClientOptions.DefaultHedgingThresholdInMilliseconds, this.RequestTimeout.TotalMilliseconds / 2);
 
                 this.AvailabilityStrategy = AvailabilityStrategy.CrossRegionHedgingStrategy(
                     threshold: TimeSpan.FromMilliseconds(defaultThresholdInMillis),
-                    thresholdStep: TimeSpan.FromMilliseconds(10));
+                    thresholdStep: TimeSpan.FromMilliseconds(CosmosClientOptions.DefaultHedgingThresholdStepInMilliseconds));
             }
         }
 
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -76,6 +76,7 @@ namespace Microsoft.Azure.Cosmos
         private string applicationName;
         private IFaultInjector faultInjector;
         private bool isCustomSerializerProvided;
+        private bool isPartitionLevelFailoverEnabled;
 
         /// <summary>
         /// Creates a new CosmosClientOptions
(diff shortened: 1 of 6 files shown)


Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -44,9 +44,9 @@ internal sealed class GlobalPartitionEndpointManagerCore : GlobalPartitionEndpoi
         private readonly int partitionUnavailabilityDurationInSeconds = ConfigurationManager.GetAllowedPartitionUnavailabilityDurationInSeconds(5);
 
         /// <summary>
-        /// A readonly integer containing the partition failback refresh interval in seconds. The default value is 60 seconds.
+        /// A readonly integer containing the partition failback refresh interval in seconds. The default value is 5 minutes.
         /// </summary>
-        private readonly int backgroundConnectionInitTimeIntervalInSeconds = ConfigurationManager.GetStalePartitionUnavailabilityRefreshIntervalInSeconds(60);
+        private readonly int backgroundConnectionInitTimeIntervalInSeconds = ConfigurationManager.GetStalePartitionUnavailabilityRefreshIntervalInSeconds(300);
```
kundadebdatta, line 49: These value changes are the outcome of my past DR drills. `1` min of failback-refresh interval is too aggressive to failback the partitions. Hence changing it to `5` mins.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -1262,6 +1260,21 @@ internal UserAgentContainer CreateUserAgentContainerWithFeatures(int clientId)
                         suffix: this.GetUserAgentSuffix());
         }
 
+        internal void InitializePartitionLevelFailoverWithDefaultHedging(
+            bool isPPafEnabled = false)
+        {
+            this.isPartitionLevelFailoverEnabled = isPPafEnabled || ConfigurationManager.IsPartitionLevelFailoverEnabled(defaultValue: false);
+
+            if (this.isPartitionLevelFailoverEnabled
+                && this.AvailabilityStrategy == null
```
kirankumarkolli, line 1269: Caller is CTOR => AvailabilityStrategy will always be null right?

Another way we can say that we default initialize and CX can override later, if that's what the intent is.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -1262,6 +1260,21 @@ internal UserAgentContainer CreateUserAgentContainerWithFeatures(int clientId)
...
+            bool isPPafEnabled = false)
+        {
+            this.isPartitionLevelFailoverEnabled = isPPafEnabled || ConfigurationManager.IsPartitionLevelFailoverEnabled(defaultValue: false);
+
+            if (this.isPartitionLevelFailoverEnabled
+                && this.AvailabilityStrategy == null
+                && !ConfigurationManager.IsDefaultHedgingDisabledWithPartitionLevelFailover())
+            {
+                this.AvailabilityStrategy = AvailabilityStrategy.CrossRegionHedgingStrategy(
+                    threshold: TimeSpan.FromMilliseconds(1000),
```
kirankumarkolli, line 1273: Thresholds ideally be based on the user configured RequestTimeouts ad discussed with Abhijeet spec.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kirankumarkolli: What's the side-affect of aggressive retries? 
With 5M the intervals are 5M, 10M, 15M so in-between times are longer. 

Is it about optimizing the client resources?

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -608,7 +608,7 @@ public PartitionKeyRangeFailoverInfo(
                 this.ConsecutiveWriteRequestFailureCount = 0;
                 this.ReadRequestFailureCounterThreshold = ConfigurationManager.GetCircuitBreakerConsecutiveFailureCountForReads(10);
                 this.WriteRequestFailureCounterThreshold = ConfigurationManager.GetCircuitBreakerConsecutiveFailureCountForWrites(5);
-                this.TimeoutCounterResetWindowInMinutes = TimeSpan.FromMinutes(1);
+                this.TimeoutCounterResetWindowInMinutes = TimeSpan.FromMinutes(2);
```
kirankumarkolli, line 611: What's the rationale for changing the window interval to double?

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
```diff
@@ -333,5 +339,21 @@ public static bool IsTcpChannelMultiplexingEnabled()
...
+
+        /// <summary>
+        /// Determines whether the default hedging is disabled/ opted-out in conjunction with partition-level failover.
+        /// This method checks the environment configuration to retrieve the value of the 
+        /// AZURE_COSMOS_SKIP_PPAF_DEFAULT_HEDGING environment variable, defaulting to false if not explicitly set.
+        /// </summary>
+        /// <returns>
+        /// A boolean value indicating whether cross region hedging is disabled with partition-level failover.
+        /// </returns>
+        public static bool IsDefaultHedgingDisabledWithPartitionLevelFailover()
```
kirankumarkolli, line 351: One alternative is to explictly set AvailaiblityStrategy to DisableAvailabilityStratgy. Is that an acceptable outcome?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Agreed - This logic has been moved under `GetConnectionPolicy()`.

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
kundadebdatta: My only concern was that we need to make the Cx aware that, even though they are not providing any availability strategy, there will be a default one which they can disable during cosmos client initialization. This approach works for me. 

I made the changes to get rid of the environment variable.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
kundadebdatta: This was reverted. Will keep it 1 mins for now. The reason to change it to 2 minutes was to provide a longer reset duration for a cold partition (which doesn't have too many requests).

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Per our conversation, updated the threshold with either 1 second or half of the request timeout value provided, whichever is minimum

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -1262,6 +1256,21 @@ internal UserAgentContainer CreateUserAgentContainerWithFeatures(int clientId)
...
+            if (this.EnablePartitionLevelFailover
+                && this.AvailabilityStrategy == null)
+            {
+                // The default threshold is the minimum value of 1 second and a fraction (currently it's half) of
+                // the request timeout value provided by the end customer.
+                double defaultThresholdInMillis = Math.Min(1000, this.RequestTimeout.TotalMilliseconds / 2);
+
+                this.AvailabilityStrategy = AvailabilityStrategy.CrossRegionHedgingStrategy(
+                    threshold: TimeSpan.FromMilliseconds(defaultThresholdInMillis),
+                    thresholdStep: TimeSpan.FromMilliseconds(10));
```
kirankumarkolli, line 1270: Isn't 10ms steps very aggressive? Is this what we closed today?

jeet1995: Being aggressive is the right mindset. If initial step threshold of 1s has elapsed, from a customer latency expectation perspective, it is a breach already.

The tradeoff here will then be how pressured client is after 1s (every 10ms I/O calls will be made) and with tradeoffs we probably need to make threshold step configurable too.

jeet1995 (dismissed):
LGTM - only ask from my side would be to double check how PPCB behaves with hedging on by default. In the Java SDK, we encountered a bug where PPCB doesn't kick in with hedging.

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: Discussed offline and we agreed to keep the threshold step as 500 milli seconds for now.

jeet1995 (dismissed):

jeet1995 (approved):

kirankumarkolli (approved):

aavasthy (approved):

PR #5171, opened 2025-05-05
Query: Fixes default for the query plan optimization for Hybrid Search to be disabled

neildsh:
## Description
#5120 enabled the switch to ask for the optimized query plan by default. However, the pipeline support for the optimized query plan is in #5121. This will lead to exceptions in the hybrid search query pipeline in version 3.49.0 if the SDK talks to a newer gateway that can return the optimized query plan. To prevent this, we change this optimization to be disabled by default.

## Type of change


adityasa (approved):
:shipit:

Maya-Painter (approved):

kundadebdatta (approved):
LGTM.

kirankumarkolli (approved):

PR #5186, opened 2025-05-08
[Internal] Thin Client Integration: Adds proxy request headers startEPK and endEPK for query operation.

### LINKED ISSUE #5185: [Thin Client Integration] Bad request error for query operation due to missing headers.
**Describe the bug**
Bad request error for query operation due to missing headers. Need to add startEPK and endEPK to proxy request for query operation.

aavasthy:
# Pull Request Template

## Description

Adds proxy request headers  startEPK and endEPK for query operation in thinclient mode.
Also, updated and cleaned up ThinClientStoreClient unit tests.

## Type of change

## Closing issues

To automatically close an issue: closes #5185


Microsoft.Azure.Cosmos/src/ThinClientStoreClient.cs
```diff
@@ -90,7 +90,17 @@ private async ValueTask<HttpRequestMessage> PrepareRequestForProxyAsync(
 
             BufferProviderWrapper bufferProviderWrapper = this.bufferProviderWrapperPool.Get();
             try
-            {
+            {
+                PartitionKeyRange partitionKeyRange = request.RequestContext.ResolvedPartitionKeyRange;
+
+                requestMessage.Headers.TryAddWithoutValidation(
```
kundadebdatta, line 96: This would mean that even if `partitionKeyRange` is `null`, we will end up adding the null value in the header correct ? Howe about having the null check before adding it to the headers ?

kirankumarkolli, line 96: Ut coverage?

aavasthy: In serializer we do check for the null for the same header, but I can include it here.

FabianMeiswinkel: Please add a unit test for PK==Null and PK==None

kundadebdatta (approved):
LGTM. The PK `Null` and `None` tests will be added as a part of a follow up PR correct ?

Microsoft.Azure.Cosmos/src/ThinClientStoreClient.cs
```diff
@@ -90,7 +90,20 @@ private async ValueTask<HttpRequestMessage> PrepareRequestForProxyAsync(
 
             BufferProviderWrapper bufferProviderWrapper = this.bufferProviderWrapperPool.Get();
             try
-            {
+            {
+                PartitionKeyRange partitionKeyRange = request.RequestContext.ResolvedPartitionKeyRange;
```
FabianMeiswinkel, line 94: This looks weird - is there any guarantee that ResolvedPartitionKeyRange will only ever be set when only routing information is PhysicalPartitionKeyRangeId?

Wouldn't this property also be set when using ParittionKeyValue or EPK for routing? In which case the startEpk/EndEpk should be the narrower scope?

FabianMeiswinkel: HOnes question - what lead to choosing existence of ResolvedPartitionKeyrange to set these headers?

aavasthy: This  startEpk/EndEpk headers are only being used for non-point operations. That check is added in the thinclient serializer code https://github.com/Azure/azure-cosmos-dotnet-v3/blob/ea26d298279242e019b1317a498fe53b44d0511c/Microsoft.Azure.Cosmos/src/ThinClientTransportSerializer.cs#L91

FabianMeiswinkel: Sounds good

FabianMeiswinkel (approved):
LGTM now - thanks!

kundadebdatta (approved):
LGTM.

Microsoft.Azure.Cosmos/src/ThinClientTransportSerializer.cs
```diff
@@ -74,11 +74,11 @@ public static async Task<Stream> SerializeProxyRequestAsync(
 
             if (operationType.IsPointOperation())
             {
-                string partitionKey = request.Headers.Get(HttpConstants.HttpHeaders.PartitionKey);
-                
-                if (string.IsNullOrEmpty(partitionKey))
-                {
-                    throw new InternalServerErrorException();
```
kirankumarkolli, line 81: nit: Revert

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/ThinClientTransportSerializer.cs
aavasthy: will clean it up as part of follow up with more test changes https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5214

PR #5191, opened 2025-05-09
Change feed: Adds id and pk to ChangeFeedMetadata for delete operations

dibahlfi:
Adds id and partition key to ChangeFeedMetadata. This is populated for delete operations using all versions and deletes change feed mode so that when customers get the delete, they know what was deleted. Note it doesn't have the value of the document that was deleted only the id and partiton key through which a document can be uniquely identified.
Please note that user provided serializer is not being used.

New feature (non-breaking change which adds functionality)


Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -50,9 +51,21 @@ class ChangeFeedMetadata
...
 
         /// <summary>
-        /// Used to distinquish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
+        /// Used to distinguish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
         /// </summary>
         [JsonProperty(PropertyName = ChangeFeedMetadataFields.TimeToLiveExpired, NullValueHandling = NullValueHandling.Ignore)]
         public bool IsTimeToLiveExpired { get; internal set; }
+
+        /// <summary>
+        /// The id of the previous item version. Used for delete operations only.
```
kirankumarkolli, line 60: Its only available for deletes right? Lets please call it explictly. 
And also please add a code doc sample and use-case.

kirankumarkolli:
Please update PR description with usage samples and also call out that user-serializer is not used.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -50,9 +51,21 @@ class ChangeFeedMetadata
...
+        /// </summary>
+        [JsonProperty(PropertyName = ChangeFeedMetadataFields.Id, NullValueHandling = NullValueHandling.Ignore)]
+        public string Id { get; internal set; }
+
+        /// <summary>
+        /// The partition key of the previous item version. Dictionary Key is the partition key property name and Dictionary Value is the partition key property value. Used for delete operations only.
```
kirankumarkolli, line 66: Lets please expand the description with what the property names are

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -16,12 +16,13 @@ namespace Microsoft.Azure.Cosmos
     /// The metadata of a change feed resource with <see cref="ChangeFeedMode"/> is initialized to <see cref="ChangeFeedMode.AllVersionsAndDeletes"/>.
     /// </summary>
     [System.Text.Json.Serialization.JsonConverter(typeof(ChangeFeedMetadataConverter))]
+    [JsonConverter(typeof(ChangeFeedMetadataNewtonSoftConverter))]
```
kirankumarkolli, line 19: Why is newtonsoftconverter needed?

kirankumarkolli (changes requested):
What's the use-case for NewtonSoftConverter needed?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
dibahlfi: I think Justine added one but will double check and if its not there will add it.

jcocchi: Yes we have a sample for all versions and deletes using both CFP and pull model! When this PR is merged I can update those samples

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
jcocchi: Can we add a note about hierarchical keys too- all levels of the hierarchy will be represented

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -50,9 +51,23 @@ class ChangeFeedMetadata
...
 
         /// <summary>
-        /// Used to distinquish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
+        /// Used to distinguish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
         /// </summary>
         [JsonProperty(PropertyName = ChangeFeedMetadataFields.TimeToLiveExpired, NullValueHandling = NullValueHandling.Ignore)]
         public bool IsTimeToLiveExpired { get; internal set; }
+
+        /// <summary>
+        /// Used for delete operations only.
```
kirankumarkolli, line 60:
```suggestion
        /// Applicable for delete operations only, for non-delete it will be null.
```

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataNewtonSoftConverter.cs
```diff
@@ -0,0 +1,202 @@
...
+                    case ChangeFeedMetadataFields.TimeToLiveExpired:
+                        metadata.IsTimeToLiveExpired = reader.Value != null && Convert.ToBoolean(reader.Value);
+                        break;
+
+                    case ChangeFeedMetadataFields.Id:
+                        metadata.Id = reader.Value?.ToString();
+                        break;
+
+                    case ChangeFeedMetadataFields.PartitionKey:
+                        if (reader.TokenType == JsonToken.StartObject)
```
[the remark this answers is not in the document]
dibahlfi: I will add some assert statements which are currently missing to check for null for non-delete operations and fix the Converter.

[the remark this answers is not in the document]
dibahlfi: Fixed.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataNewtonSoftConverter.cs
```diff
@@ -0,0 +1,202 @@
...
+
+                    default:
+                        reader.Skip();
+                        break;
+                }
+
+                reader.Read(); // Move to next property or EndObject
+            }
+
+            metadata.PartitionKey = partitionKey;
```
kirankumarkolli: nit: Can this be set inside the at L:168? 
Will scope all related changes in one place.

dibahlfi:
> Please update PR description with usage samples and also call out that user-serializer is not used.

Justine already has the samples ready and she will update them once the PR goes out.

dibahlfi:
> What's the use-case for NewtonSoftConverter needed?

because of the introduction of List<(string, object)>for which a custom serializator was needed to serialize/deserialize.

dibahlfi:
> What's the use-case for NewtonSoftConverter needed?
because of the introduction of List<(string, object)>for which a custom serializator was needed to serialize/deserialize.

FabianMeiswinkel (dismissed):
LGTM - Thanks

JensSchadron:
Given that this received an approval last week, is there anything still blocking this PR? ( @kirankumarkolli )
Really looking forward to this becoming available in the (preview?) SDK

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
```diff
@@ -76,6 +98,46 @@ public override void Write(Utf8JsonWriter writer, ChangeFeedMetadata value, Json
...
+                        case bool boolValue:
+                            writer.WriteBoolean(key, boolValue);
+                            break;
+
+                        case null:
+                            writer.WriteNull(key);
+                            break;
+
+                        default:
+                            throw new JsonException($"Unexpected value type '{value.GetType()}' for PartitionKey property.");
```
kirankumarkolli: include the key in exception message

kirankumarkolli: It makes JsonException Type as public contract, how about using CosmosException it-self?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
```diff
@@ -56,6 +57,27 @@ public override ChangeFeedMetadata Read(ref Utf8JsonReader reader, Type typeToCo
...
+                    List<(string, object)> partitionKey = new List<(string, object)>();
+                    foreach (JsonProperty pk in property.Value.EnumerateObject())
+                    {
+                        object actualValue = pk.Value.ValueKind switch
+                        {
+                            JsonValueKind.String => pk.Value.GetString(),
+                            JsonValueKind.Number => pk.Value.TryGetInt64(out long longValue) ? longValue : (object)pk.Value.GetDouble(),
+                            JsonValueKind.True or JsonValueKind.False => pk.Value.GetBoolean(),
+                            JsonValueKind.Null => null,
+                            _ => throw new JsonException($"Unexpected JsonValueKind '{pk.Value.ValueKind}' for PartitionKey property."),
```
kirankumarkolli: CosmosException?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
```diff
@@ -56,6 +57,27 @@ public override ChangeFeedMetadata Read(ref Utf8JsonReader reader, Type typeToCo
...
+                        };
+                        partitionKey.Add((pk.Name, actualValue));
+                    }
+                    metadata.PartitionKey = partitionKey;
+                }
```
kirankumarkolli, line 80: How about adding an assertion that both id and paritionKey are present or none?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
kirankumarkolli: Thinking loud here, for non-primitive types respective de-serializers are needed. 
For primitive types may be it not needed., thoughts?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -16,43 +16,49 @@ namespace Microsoft.Azure.Cosmos
...
+        /// Applicable for delete operations only, otherwise null.
+        /// The id of the previous item version. 
+        /// </summary>
+        public string Id { get; internal set; }
+
+        /// <summary>
+        ///  Applicable for delete operations only, otherwise null.
+        /// The partition key of the previous item version. string  is the partition key property name and object is the partition key property value. All levels of hierarchy will be represented in order if a HPK is used.
+        /// </summary>
+        public List<(string, object)> PartitionKey { get; internal set; }
```
kirankumarkolli, line 62: List v Dict: Is List for order of ParittionKeys?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
yash2710: One question I have is what is the purpose of moving all the annotations from individual properties to the class?

yash2710: One drawback I see of this approach is that we loose the capability to do different null value handling behavior for properties

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
```diff
@@ -76,6 +98,46 @@ public override void Write(Utf8JsonWriter writer, ChangeFeedMetadata value, Json
...
+            if (value.Id != null)
+            {
+                writer.WriteString(ChangeFeedMetadataFields.Id, value.Id);
+            }
+
+            if (value.PartitionKey != null)
+            {
+                writer.WriteStartObject(ChangeFeedMetadataFields.PartitionKey);
+
+                foreach ((string key, object objectValue) in value.PartitionKey)
```
yash2710, line 119: I think it will be useful to create a writer util function for reusability

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataNewtonSoftConverter.cs
```diff
@@ -0,0 +1,201 @@
...
+                    }
+
+                    writer.WriteEndObject(); // End PartitionKey object
+                }
+
+                writer.WriteEndObject();
+            }
+            else
+            {
+                throw new JsonSerializationException($"Unexpected value when converting {nameof(ChangeFeedMetadata)}.");
```
yash2710: Is it possible to include the value which was unexpected?

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
yash2710: As discussed, this is no longer needed

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
yash2710: As discussed this is no longer needed

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
dibahlfi: Yes.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
dibahlfi: I ran into it and what I observed was when you have a class-level JsonConverter attribute, the converter takes full control. Any property-level annotations (like [JsonPropertyName], [JsonProperty], etc.) are bypassed.

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
dibahlfi: done

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
dibahlfi: I think its very specific to this class wont add much value unless we are anticipating some future use

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataNewtonSoftConverter.cs
dibahlfi: added

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -16,43 +16,49 @@ namespace Microsoft.Azure.Cosmos
     /// The metadata of a change feed resource with <see cref="ChangeFeedMode"/> is initialized to <see cref="ChangeFeedMode.AllVersionsAndDeletes"/>.
     /// </summary>
     [System.Text.Json.Serialization.JsonConverter(typeof(ChangeFeedMetadataConverter))]
```
kirankumarkolli, line 18: On overall approach: new possible approach is to leverage https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonincludeattribute?view=net-9.0&viewFallbackFrom=netstandard-2.0 (assuming that we can use it with-out rebasing to latest .NET versions.

/cc: @yash2710 , @dibahlfi

Microsoft.Azure.Cosmos/src/Resource/FullFidelity/Converters/ChangeFeedMetadataConverter.cs
dibahlfi: done

kirankumarkolli:
```
Failed PreviewContractChanges [101 ms]
  Error Message:
   Assert.Fail failed. Public API has changed. If this is expected, then run (EnlistmentRoot)\UpdateContracts.ps1 . To see the differences run the update script and use git diff.
  Stack Trace:
     at Microsoft.Azure.Cosmos.Tests.Contracts.ContractEnforcement.ValidateJsonAreSame(String baselineJson, String currentJson) in D:\a\_work\1\s\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.Tests\Contracts\ContractEnforcement.cs:line 542
   at Microsoft.Azure.Cosmos.Tests.Contracts.ContractEnforcement.ValidateContract(String dllName, ContractType contractType, String baselinePattern, String breakingChangesPattern, String officialBaselinePattern) in D:\a\_work\1\s\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.Tests\Contracts\ContractEnforcement.cs:line 423
   at Microsoft.Azure.Cosmos.Tests.Contracts.ContractEnforcementTests.PreviewContractChanges() in D:\a\_work\1\s\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.Tests\Contracts\ContractEnforcementTests.cs:line 22

  Standard Output Messages:
```
Contract changes needs refresh

PR #5201, opened 2025-05-15
[INTERNAL] SessionRetry: Adds support to yield immediately after the original attempt when READ_SESSION_NOT_AVAILABLE error is thrown

dibahlfi:
Exposing MaxInRegionRetryCount variable internally so that if it has value of 0 or less than 0 we will not do any retry/delay and yield immediately for READ_SESSION_NOT_AVAILABLE error from the SessionTokenMismtatchInternalRetry policy.

[SessionRetryEarlyYield.json](https://github.com/user-attachments/files/20229840/SessionRetryEarlyYield.json)

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -754,6 +754,17 @@ namespace Microsoft.Azure.Cosmos
         public AvailabilityStrategy AvailabilityStrategy { get; set; }
 
         /// <summary>
+        /// Provides access to the maximum number of in-region retries for session retry policy.
+        /// If a value of less than or equal to 0 is provided we will not retry and yield after the original attempt, otherwise default is going to be 1
+        /// use value of less than or equal to for this parameter with care as this will have negative consequences for multi-master and some other scenerios.
+        /// </summary>
+        internal int MaxInRegionRetryCountForSessionRetry
+        {
+            get => this.SessionRetryOptions.MaxInRegionRetryCount;
+            set => this.SessionRetryOptions.MaxInRegionRetryCount = value;
+        }
+
+        /// <summary>
         /// Provides SessionTokenMismatchRetryPolicy optimization through customer supplied region switch hints,
         /// which guide SDK-internal retry policies on how early to fallback to the next applicable region.
         /// With a single-write-region account the next applicable region is the write-region, with a 
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -57,12 +57,6 @@ namespace Microsoft.Azure.Cosmos
         private const ApiType DefaultApiType = ApiType.None;
 
         /// <summary>
-        /// Default thresholds for PPAF request hedging.
-        /// </summary>
-        private const int DefaultHedgingThresholdInMilliseconds = 1000;
-        private const int DefaultHedgingThresholdStepInMilliseconds = 500;
-
-        /// <summary>
         /// Default request timeout
         /// </summary>
         private int gatewayModeMaxConnectionLimit;
@@ -772,15 +766,6 @@ namespace Microsoft.Azure.Cosmos
         }
 
         /// <summary>
-        /// Gets or sets a value indicating whether partition-level failover is enabled. When this feature is enabled,
-        /// the SDK by default applies a cross-region hedging strategy with a default threshold of 1 seconds.
-        /// If an availability strategy is provided explicitly, then it will be honored, and the default policy wouldn't be applied. Note that
-        /// the default availability strategy can be opted out by setting <see cref="DisabledAvailabilityStrategy"/> as the availability strategy in
-        /// cosmos client options.
-        /// </summary>
-        internal bool EnablePartitionLevelFailover { get; set; } = ConfigurationManager.IsPartitionLevelFailoverEnabled(defaultValue: false);
-
-        /// <summary>
         /// Enable partition level circuit breaker (aka PPCB). For compute gateway use case, by default per partition automatic failover will be disabled, so does the PPCB.
         /// If compute gateway chooses to enable PPAF, then the .NET SDK will enable PPCB by default, which will improve the read availability and latency. This would mean
         /// when PPAF is enabled, the SDK will automatically enable PPCB as well.
@@ -1028,10 +1013,10 @@ namespace Microsoft.Azure.Cosmos
         {
             this.ValidateDirectTCPSettings();
             this.ValidateLimitToEndpointSettings();
-            this.InitializePartitionLevelFailoverWithDefaultHedging();
 
             ConnectionPolicy connectionPolicy = new ConnectionPolicy()
             {
+                ApplicationName = this.ApplicationName,
                 MaxConnectionLimit = this.GatewayModeMaxConnectionLimit,
                 RequestTimeout = this.RequestTimeout,
                 ConnectionMode = this.ConnectionMode,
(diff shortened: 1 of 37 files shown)


kirankumarkolli (changes requested):
Session immediate yield its only for compute team and is making the types on the Policy type internal set

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -753,6 +753,17 @@ public Func<HttpClient> HttpClientFactory
         /// </remarks>
         public AvailabilityStrategy AvailabilityStrategy { get; set; }
 
+        /// <summary>
```
FabianMeiswinkel, line 756: This change can be reverted - you make the setter available on SessionRetryOptions internally already - and CosmsoClientOPtions.SessionRetryOptions as an internal getter already. So, this is redundant - not needed.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -583,6 +583,18 @@ public CosmosClientBuilder WithSerializerOptions(CosmosSerializationOptions cosm
             return this;
         }
 
+        /// <summary>
```
FabianMeiswinkel, line 586: Dit - change can be reverted

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
```diff
@@ -80,11 +80,10 @@ internal static class ConfigurationManager
         internal static readonly int MinMinInRegionRetryTimeForWritesInMs = 100;
 
         /// <summary>
-        /// intent is If a client specify a value, we will force it to be atleast 1, otherwise default is going to be 1(right now both the values are 1 but we have the provision to change them in future).
+        /// If a client specify a value  less than or equal to 0 we will not retry and yield after the original attempt, otherwise default is going to be 1
```
FabianMeiswinkel, line 83:
```suggestion
        /// If a client specify a value less than or equal to 0 we will not retry and yield after the original attempt, otherwise default is going to be 1
```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
dibahlfi: agree.

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
dibahlfi: done

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
```diff
@@ -112,12 +111,10 @@ public static T GetEnvironmentVariable<T>(string variable, T defaultValue)
 
         public static int GetMaxRetriesInLocalRegionWhenRemoteRegionPreferred()
         {
```
kirankumarkolli, line 113: Is this change needed? 
Compute will set it direclty on the type not through ENV variable.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/SessionRetryOptionsTest.cs
```diff
@@ -39,13 +39,16 @@ public async Task TestInitAsync()
         [DataRow(FaultInjectionOperationType.QueryItem, 1, true, DisplayName = "Validate Query Item operation with remote region preferred.")]
         [DataRow(FaultInjectionOperationType.ReadItem, 2, false, DisplayName = "Validate Read Item operation with local region preferred.")]
         [DataRow(FaultInjectionOperationType.QueryItem, 2, false, DisplayName = "Validate Query Item operation with local region preferred.")]
+        [DataRow(FaultInjectionOperationType.ReadItem, 0, true, DisplayName = "Validate Read Item operation with 0 sessionTokenMismatchRetryAttempts.")]
+        [DataRow(FaultInjectionOperationType.QueryItem, -1, true, DisplayName = "Validate Query Item operation with negative sessionTokenMismatchRetryAttempts.")]
```
kirankumarkolli, line 43: ENV should still limit to at-least ONE as public contract. 
Only for compute directly setting on the Type we are allowing and lets please cover it separately as test.

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
FabianMeiswinkel: +1

FabianMeiswinkel (commented):
I agree with Kiran's comment - Env-variable doesn't need to (or probabyl shouldn't) allow setting ot 0 or -1 - for now let us reduce the scope to just Compute Gw.

Microsoft.Azure.Cosmos/src/SessionRetryOptions.cs
```diff
@@ -26,18 +28,40 @@ public SessionRetryOptions()
...
         /// </summary>
-        public int MaxInRegionRetryCount { get; private set; }
+        public int MaxInRegionRetryCount 
+        {
+            get => this.maxInRegionRetryCount;
+            internal set
+            {
+                if (value <= 0 && !this.RemoteRegionPreferred)
+                {
+                    throw new ArgumentException("MaxInRegionRetryCount can only be set to 0 or less when RemoteRegionPreferred is true.");
```
FabianMeiswinkel, line 43: Sorry, might have not been specific enough. I don't think throwing here and forcing ordering in which properties are set is a good idea. Instead the code consuming SessionRetryOptions should just ignore whatever value is in MaxInRegionRetryCount unless RemoteRegionPreferred is true. For public API it would have been best to not have individual setters but one mehtod to modify all these parameters to allow consistency checks. But since this is internal I think above is sufificent.

FabianMeiswinkel: Reason why I don't like thorwing here is that setting properties on tehse case classes should not depend on the order in which properties are set.

dibahlfi: Thanks. will make the change.

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
kirankumarkolli: Undo changes to this file

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/SessionRetryOptionsUnitTests.cs
```diff
@@ -43,7 +43,6 @@ public void SessionRetryOptionsDefaultValuesTest()
             {
                 EnableRemoteRegionPreferredForSessionRetry = true,
             };
-
```
kirankumarkolli, line 46: Undo

Microsoft.Azure.Cosmos/src/SessionRetryOptions.cs
```diff
@@ -31,7 +31,7 @@ public SessionRetryOptions()
         /// Sets the maximum number of retries within each region for read and write operations. The minimum value is 1 - the backoff time for the last in-region retry will ensure that the total retry time within the
         /// region is at least the min. in-region retry time.
         /// </summary>
-        public int MaxInRegionRetryCount { get; private set; }
+        public int MaxInRegionRetryCount { get; internal set; }
```
kirankumarkolli, line 34: Lets also enable internal for MinInRegionRetryTime

kirankumarkolli (approved):

ananth7592 (approved):

kirankumarkolli (approved):

PR #5207, opened 2025-05-20
[Internal] Thin Client Integration: Fixes runner tests failures for thinclient test scenarios.

aavasthy:
# Pull Request Template

## Description

Fixes runner tests failures for thinclient test scenarios by adding a timeout limit on the tests.

## Type of change

## Closing issues

### DIFF
--- a/templates/build-test.yml
+++ b/templates/build-test.yml
@@ -5,8 +5,8 @@ parameters:
   Arguments: ''
   VmImage: '' # https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops
   OS: 'Windows'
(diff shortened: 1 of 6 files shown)


templates/build-test.yml
```diff
@@ -22,6 +22,7 @@ parameters:
 jobs:
 - job:
   displayName: Microsoft.Azure.Cosmos.Tests
+  timeoutInMinutes: 120
```
kundadebdatta, line 25: Shall we keep the original timeout value and check if your test level timeouts (in the above file) covers it end to end ?

aavasthy: Still running into task cancellation as it times out after 60 minutes. So just experimenting in this branch to get the test jobs to pass.

PR #5221, opened 2025-05-29
Query : Fixes Skip + Order By Bug

### LINKED ISSUE #5016: Regression in LINQ Skip in 3.47.0

**Describe the bug**
When queryable includes a Skip() method, feedIterator.ReadNextAsync throws an ArgumentOutOfRangeException

**To Reproduce**
Use Microsoft.Azure.Cosmos 3.47.0

Create the queryable: var queryable = container.GetItemLinqQueryable<TItem>().OrderBy(x => x.Name).Skip(1);
Get the feediterator: ver feediterator = queryable.ToFeedIterator();
Get the feedresponse: var feedresponse = await feediterator.ReadNextAsync();

Exception is thrown:

An exception of type 'System.AggregateException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'One or more errors occurred.'
 Inner exceptions found, see $exception in variables window for more details.
 Innermost exception 	 System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values. (Parameter 'Value')

**Expected behavior**
No exception. Valid feedresponse is returned and items can be enumerated.

This is the behavior in 3.46.1.

**Actual behavior**
An exception of type 'System.AggregateException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'One or more errors occurred.'
 Inner exceptions found, see $exception in variables window for more details.
 Innermost exception 	 System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values. (Parameter 'Value')

**Environment summary**
SDK Version: 3.47.0
OS Version (e.g. Windows, Linux, MacOSX): macOS Sequoia 15.3

**Additional context**
feediterator queryspec looks like this: SELECT VALUE root FROM root OFFSET 1 LIMIT 2147483647

Maya-Painter:
## Description

Addresses a bug where the top value was being set > int.Max for queries with ORDER BY and SKIP clauses. This triggered an assert and caused query failures.

## Type of change

## Closing issues

closes #5016


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
```diff
@@ -949,6 +949,26 @@ public async Task LinqSelectEverythingWithoutQueryableTest()
             Assert.AreEqual(2, (await this.FetchResults<ToDoActivity>(queryDefinition)).Count);
         }
 
+        [TestMethod]
+        public async Task LinqSkipOrderBy()
```
adityasa, line 953: >LinqSkipOrderBy [](http://example.com/codeflow?start=26&length=15)

Can you also add a baseline test for this? I am curious what the query looks like.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/PipelineFactory.cs
```diff
@@ -172,7 +172,7 @@ public static TryCatch<IQueryPipelineStage> MonadicCreate(
                     top = 0;
                 }
 
-                if (top > int.MaxValue)
+                if (queryInfo.HasTop && (top > int.MaxValue))
```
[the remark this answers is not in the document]
Maya-Painter: Any query that exercises this path - so any query with both TOP and ORDER BY.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
adityasa: Why does that violate the top > int.MaxValue condition?

Maya-Painter: image
This is because QueryInfo.Offset = 1 and queryInfo.Limit.Value = Int.Max. An alternative to changing the validation is making sure that this value never exceeds Int.Max

adityasa: I think that is a better fix.

adityasa:
>                 Debug.Assert(queryInfo.Top.Value <= int.MaxValue, "PipelineFactory Assert!", "Top value must be <= int.MaxValue");

FYI, you can also move the original check here and remove this Debug Assert.

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/PipelineFactory.cs:299 in 4d983f8. [](commit_id = 4d983f82ce291a413102b7613402df129061ceff, deletion_comment = False)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
adityasa: The other alternative is to move the original check to L299 (since that is the only place where top value actually gets used).

adityasa:
>                 Debug.Assert(queryInfo.Top.Value <= int.MaxValue, "PipelineFactory Assert!", "Top value must be <= int.MaxValue");

Do we need to repeat the original check here? What happens if query is missing ORDER BY but has only TOP (did we have coverage for that case in the original change)?

---
Refers to: Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/PipelineFactory.cs:299 in 4d983f8. [](commit_id = 4d983f82ce291a413102b7613402df129061ceff, deletion_comment = False)

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemLinqTests.cs
Maya-Painter: Thats good to know. Before the assert was added this was happening silently.

leminh98 (approved):

PR #5222, opened 2025-05-29
[INTERNAL] Dependency: Fixes Updating immutable to V6

kirankumarkolli:
Testing Immutable V6 package upgrade impact


kirankumarkolli:
Not needed, offline validated

PR #5236, opened 2025-06-04
[msdata/direct] Refactor small piece of Cosmos DB serialization to be AOT-compatible

sbomer:
This enables an AOT-compiled version of the library to be used to read account info and database properties. Other uses of the library are still not AOT-compatible.

This includes breaking API changes that I expect will be problematic, so this is mostly for illustration purposes. It shows the kinds of patterns that could be used to make an AOT-compatible API surface (not suggesting this is a good final API shape). I'm hoping this PR can help inform the discussion for CosmosDB owners. @kirankumarkolli
 
- Introduced new serialization context (CosmosSerializerContext) for handling JSON serialization of various resource types.
- Updated CosmosResource to support deserialization from DocumentServiceResponse using the new context.
- Created new classes (FeedResource_Address, FeedResource_PartitionKeyRange) to encapsulate feed resources for addresses and partition key ranges.
- Replaced Newtonsoft.Json with System.Text.Json for Address and PartitionKeyRange classes, simplifying property definitions.
- Added SerializationHelper for common serialization tasks.
- Updated SnapshotContent to utilize the new serialization methods.
- Refactored Resource and PlainResource classes to implement IResource interface for better abstraction.
- Enhanced error handling during deserialization processes.


Microsoft.Azure.Cosmos/src/direct/DocumentServiceResponse.cs
sbomer, line 1: Fixing the file encoding in https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5237

PR #5238, opened 2025-06-05
Fix: update ReadManyItems test coverage for 429 exceptions

dibahlfi:
Updating the `CosmosReadManyItemsTests` class in the `Microsoft.Azure.Cosmos.SDK.EmulatorTests` namespace to include test coverage for 429 throttling exceptions.


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosReadManyItemsTests.cs
```diff
@@ -362,6 +363,67 @@ public async Task ReadManyTestWithIncorrectIntendedContainerRid()
...
+            List<Task> tasks = new List<Task>();
+
+            for (int i = 0; i < concurrentRequests; i++)
+            {
+                tasks.Add(Task.Run(async () =>
+                {
+                    // Stream API
+                    using (ResponseMessage responseMessage = await container.ReadManyItemsStreamAsync(itemList))
+                    {
+                        if (responseMessage.StatusCode == HttpStatusCode.TooManyRequests)
```
kirankumarkolli, line 402: This is follow-up from an ICM issue. ICM states that with 429 sometimes there are duplicate records. 
Idea is to simulate 429 and also let the retries succeed to simulate the scenario and also validate the response for the expectations

PR #5255, opened 2025-06-20
PPAF: Adds Concurrent Detection of Write Regions During Failover

### LINKED ISSUE #4858: [Per Partition Automatic Failover] Concurrent Detection of Write Regions During Failover
### Acceptance Criteria:

During the partition level failover through the .NET v3 SDK, it is often identified that the detection of the write regions takes longer, sometimes more than a minute. The primary reason for that is the following two:

- The offending partition takes at least a minute from the backend to failover to another region.

- The current (write region) detection logic from the SDK is on a round robin fashion, therefore, the SDK loops through all the regions in the account topology to find out the potential write region.

Though it’s beyond the scope of the SDK to optimize the backend failover time, however the detection logic from the SDK can be made a bit faster. This design proposes an optimization to detect the write regions in parallel by issuing concurrent hedging requests to all the available regions in the account topology.

NaluTripician:
# Pull Request Template

## Description

This pull request enables **write hedging for single-master accounts with PPAF (Per-Partition Automatic Failover)** enabled. When a write request to the primary region is slow or unavailable, the SDK now proactively sends hedged write requests to secondary regions using the read endpoints as failover targets, reducing tail latency and retry amplification during regional outages.

### Production code changes (CrossRegionHedgingAvailabilityStrategy.cs):

* Added `ppafEnabled` field set from `ConnectionPolicy.EnablePartitionLevelFailover`
* **Fixed `ShouldHedge`**: Separated PPAF eligibility from `CanSupportMultipleWriteLocations` check — single-master PPAF accounts now correctly enable write hedging (previously the `ppafEnabled` flag had no effect because `CanSupportMultipleWriteLocations()` always returns false for single-master)
* **Fixed region selection**: Changed `GetApplicableRegions` to use read regions for PPAF write hedging, since single-master PPAF failover routes writes to read regions

### Unit tests (AvailabilityStrategyUnitTests.cs — 9 new tests):

* `ShouldHedge` tests covering PPAF flag interactions with write/read/non-document requests across single-master and multi-master configurations
* `ExecuteAvailabilityStrategyAsync_SetsPpafEnabledFromConnectionPolicy` — validates ppafEnabled is set from ConnectionPolicy
* `ConcurrentExecuteAvailabilityStrategy_PpafEnabledField_NoCorruption` — thread safety under concurrent access
* Added `CreateWriteRequest()` and `CreateMockMultiWriteClient()` test helpers

### Integration tests (CosmosItemIntegrationTests.cs — 3 new tests):

* `CreateItemAsync_WithPPAFEnabledAndSingleMasterAccountWithResponseDelay_ShouldHedgeWriteToMultipleRegions` — validates write hedging with response delay fault injection (DataRow: PPAF enabled/disabled)
* `CreateItemAsync_WithPPAFEnabledAndNoDelay_ShouldNotHedgeWrite` — validates no hedging when primary responds before threshold
* `CreateItemAsync_WithPPAFEnabledAndPrimaryUnavailable_ShouldHedgeAndSucceedInSecondary` — validates hedging succeeds via secondary when primary returns 503

### Updated existing test (GlobalPartitionEndpointManagerTests.cs):

* `TestRequestTimeoutExceptionScenarioAsync` updated to account for write hedging behavior — with hedging enabled, the first `CreateItemAsync` now succeeds via hedging to secondary instead of throwing `RequestTimeout`

### Phase 2 follow-up items (GitHub Issues):

* [#5635](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5635) — Override Map Consistency Under Concurrent Hedging
* [#5637](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5637) — Late-Binding PK-Range Override After Success Confirmation

## Type of change

## Closing issues

closes #4858


FabianMeiswinkel (dismissed):
LGTM except for the PR descriptiona s discussed offline

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -132,6 +134,7 @@ internal override async Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
             RequestMessage request,
             CancellationToken cancellationToken)
         {
+            this.ppafEnabled = client.DocumentClient.ConnectionPolicy.EnablePartitionLevelFailover;
```
ananth7592: Tests please.

ananth7592 (dismissed):
Approved with a comment

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -151,7 +160,7 @@ internal override async Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
                     IReadOnlyCollection<string> hedgeRegions = client.DocumentClient.GlobalEndpointManager
                         .GetApplicableRegions(
                             request.RequestOptions?.ExcludeRegions,
-                            OperationTypeExtensions.IsReadOperation(request.OperationType));
+                            this.ppafEnabled || OperationTypeExtensions.IsReadOperation(request.OperationType));
```
kundadebdatta: So, the `GetApplicableRegions` will now receive the `isReadRequest` as `this.ppafEnabled || OperationTypeExtensions.IsReadOperation(request.OperationType)`.

The catch of doing this is that, for a write request on a single master account with PPAF enabled, it will end up selecting from the available read locations, where as from the current behavior, we would need to get the regions from `GlobalEndpointManager.AccountReadEndpoints`.

We might need to fix this.

NaluTripician: fixed

kushagraThapar (commented):
Blocking routing and concurrency concerns identified during review.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -149,13 +181,27 @@ internal override async Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
...
-                        .GetApplicableRegions(
-                            request.RequestOptions?.ExcludeRegions,
-                            OperationTypeExtensions.IsReadOperation(request.OperationType));
+                    bool isReadRequest = OperationTypeExtensions.IsReadOperation(request.OperationType);
+
+                    // For PPAF write hedging, use all account-level read regions (consistent with
+                    // GlobalPartitionEndpointManagerCore's use of AccountReadEndpoints for PPAF failover).
+                    // GetApplicableRegions filters through EffectivePreferredLocations, which could
+                    // drop valid hedge targets not in the user's PreferredLocations.
+                    IReadOnlyCollection<string> hedgeRegions = this.ppafEnabled && !isReadRequest
```
kushagraThapar: Could we bind each hedge arm to the explicit endpoint selected from this topology snapshot instead of carrying only a region name? `hedgeRegions` can include account-level read regions outside the preferred list, but `ResolveServiceEndpoint` later resolves through `GetApplicableEndpoints(request, isReadRequest: true)`, which is preferred-location based. For primary A with preferred `[B, A]` (or a failover owner C outside that list), the logical B/C arm can therefore resolve back to A, producing duplicate primary writes and potentially caching an A->A override. Each alternate should carry its endpoint through dispatch, with the primary kept separate.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -331,6 +383,20 @@ private async Task<HedgingResponse> CloneAndSendAsync(
                     excludeRegions.RemoveAt(requestNumber);
                     clonedRequest.RequestOptions.ExcludeRegions = excludeRegions;
 
+                    // For PPAF write hedging: suppress partition-level failover cache updates
+                    // on hedged (non-primary) error responses. Without this, hedged request errors
+                    // poison the PPAF cache, causing all subsequent requests for the same
+                    // partition to think the primary region failed over—triggering more hedging
+                    // and amplifying RU consumption. On success, the primary endpoint is used
+                    // to update the cache so future requests route directly to the successful region.
+                    if (this.ppafEnabled
```
kushagraThapar: `ppafEnabled` is mutable state on a strategy instance shared across executions. It is assigned at the start of each request and reread here after awaits, so another concurrent request/client can flip it while this execution is suspended, changing suppression metadata and cache-publication behavior. Please compute one immutable per-execution PPAF flag and pass it through the hedge path; making the field `volatile` would not make the sequence atomic.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -114,6 +137,13 @@ internal bool ShouldHedge(RequestMessage request, CosmosClient client)
                 {
                     return true;
                 }
+
+                // PPAF single-master: hedge writes using read regions as failover targets
+                if (this.ppafEnabled)
```
kushagraThapar: This lets PPAF bypass `EnableMultiWriteRegionHedge=false`, while the SDK-default factory also changes that option's default to `true`. As a result, an opted-out multi-write client - or an operator disabling `AZURE_COSMOS_PPAF_WRITE_HEDGING_ENABLED` - can still fan out writes and observe duplicate-arm 409/412 responses. Please keep the contracts separate: use only the explicit option for multi-write accounts, use gated PPAF only for single-master accounts, and preserve the previous default unless the compatibility change is intentional and separately disclosed.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -373,11 +440,23 @@ private async Task<HedgingResponse> RequestSenderAndResultCheckAsync(
...
             try
             {
                 ResponseMessage response = await sender.Invoke(request, hedgeRequestsCancellationTokenSource.Token);
+
+                // ShouldRetryAsync is only called on error responses (AbstractRetryHandler
+                // short-circuits on success), so the PPAF cache update for successful hedged
+                // writes must happen here, outside the retry policy pipeline.
+                if (response.IsSuccessStatusCode)
+                {
+                    CrossRegionHedgingAvailabilityStrategy.TryUpdatePPAFCacheOnSuccessfulHedge(
```
kushagraThapar: The cache update happens before winner arbitration, so every successful arm can mutate shared PK-range state. If C wins but a slower B arm later succeeds, B can overwrite `Current` and route subsequent writes to the losing or demoted region. Hedge arms should return candidates without shared mutation; after selecting the final response, publish only the winner (ideally generation/CAS guarded) while still observing and disposing late responses.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
NaluTripician: Good catch — agreed on both halves, fixed in 437244d.

`ShouldHedge` now splits the two contracts cleanly instead of letting PPAF fall through:

- **multi-write accounts** → the decision is `EnableMultiWriteRegionHedge` and nothing else. PPAF no longer overrides an explicit `false`.
- **single-master accounts** → PPAF write hedging governs, which is the behavior this PR is actually adding.

I also reverted `AvailabilityStrategy.SDKDefaultCrossRegionHedgingStrategyForPPAF` back to `enableMultiWriteRegionHedge: false`. `DocumentClient` installs that factory without an explicit argument, so flipping the default was silently opting every PPAF-enabled multi-write app into duplicate writes — with the extra 409/412 and the non-deterministic Create-vs-Replace-on-Upsert caveat that the option's own doc comment says the application has to accept up front. The public API default in `contracts/API_*.txt` was already `false` and is unchanged.

Covered by:
- `ShouldHedge_WriteRequest_MultiWriteAccount_PpafDoesNotOverrideOptOut` — this is the old `..._WithPPAFEnabledAndMultiWrite_ReturnsTrue` test, inverted. It was asserting `IsTrue` with `enableMultiWriteRegionHedge: false`, i.e. it was encoding the bug you reported. Calling that out explicitly since an inverted assertion is exactly the kind of edit that deserves a second look.
- `ShouldHedge_WriteRequest_MultiWriteAccount_HonorsMultiWriteOptIn`
- `SDKDefaultStrategyForPPAF_DoesNotEnableMultiWriteRegionHedgeByDefault`

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
NaluTripician: You're right, and it turned out to be reproducible rather than theoretical. Fixed in 437244d.

`LocationCache.GetApplicableEndpoints` filters candidates through `EffectivePreferredLocations` and falls back to `WriteEndpoints[0]` when the filtered set comes back empty. Each hedge arm was setting `ExcludeRegions` to every hedge region except its own target, so any target **outside** the user's `PreferredLocations` produced an empty set and got routed straight back to the primary write endpoint. Two arms then write to the same region, and the PPAF cache override records primary → primary.

Fixed by binding the endpoint at fan-out time rather than re-deriving it per arm:

- `LocationCache` gained `GetApplicableAccountLevelReadRegions(excludeRegions, out effectivePreferredLocationsFallback)` returning `(Region, Endpoint)` pairs from a **single** topology snapshot, so the pair can't tear across a background refresh.
- The hedge context stamps `x-ms-ppaf-hedge-target-endpoint` on each arm.
- `ResolveServiceEndpoint` honors that pin as its first branch and skips the preferred-location filter entirely. The old `GetApplicableEndpoints(request, isReadRequest: true)` + `locationIndex % count` branch is gone.

Ordering is preferred-locations-first (in the user's declared order), then the remaining account read regions — that keeps the latency preference for the first hedge arm while making every read region actually reachable.

`ResolveServiceEndpoint_HonorsPinnedPpafHedgeTargetEndpoint` asserts the old collapse as an explicit baseline first, then asserts the pinned resolution, so the test documents the bug rather than just guarding the fix. `ResolveServiceEndpoint_IgnoresNonUriPpafHedgeTargetProperty` covers a malformed value falling through to normal resolution.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
NaluTripician: Agreed, and agreed that `volatile` wouldn't have been enough — volatility fixes visibility, not interleaving. Execution A can write `true`, execution B write `false`, and A can then legally read `false` after its await with every barrier intact. Fixed in 437244d.

The field is deleted. `ExecuteAvailabilityStrategyAsync` now computes a local `ppafWriteHedgingEnabled` and captures an immutable `PPAFWriteHedgeContext` (primary write endpoint + ordered hedge regions + per-arm target endpoints) **before** any await; nothing reads shared mutable state afterwards. `ShouldHedge` takes the flag as a parameter rather than reading instance state.

Separately, and worth flagging: the old tests reached into this field by reflection, which is a large part of why the race stayed invisible in a green suite. I removed all 10 reflection sites and made every PPAF assertion behavioral. `ConcurrentExecuteAvailabilityStrategy_PpafEnabledField_NoCorruption` is now `..._PpafDecisionIsPerExecution` and asserts that a non-PPAF client never hedges writes while a PPAF client is executing concurrently — which is the property that actually matters, and which the reflection-based test could not have caught.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
NaluTripician: Agreed — fixed in 437244d.

The update was being applied inline for any 2xx, including arms that lose arbitration and arms that complete *after* the winner has already returned to the caller. So a partition could end up pinned to a region the caller never got a response from.

Successful arms now build a `PPAFCacheUpdateCandidate` and stash it on their `HedgingResponse` instead of mutating anything. `PublishPPAFCacheUpdateForWinner` runs at all three `return hedgeResponse.ResponseMessage` sites (first-arm-wins, the `Task.WhenAny` winner, and the all-arms-exhausted path), so exactly one override is applied and it always matches the response the caller actually received.

Covered by `PpafWriteHedge_OnlyTheWinningArmPublishesTheCacheOverride`, which injects a mock `GlobalPartitionEndpointManager` via `DocumentClient.PartitionKeyRangeLocationForTests` and asserts `TrySetPartitionLevelLocationOverrideForSuccessfulHedge` fires exactly once, with the winning arm's endpoint.

NaluTripician:
@kushagraThapar — all four threads addressed in 437244d. Two things I'd flag for your eye rather than leave buried in the diff:

**1. I inverted an existing test.** `ShouldHedge_WriteRequest_WithPPAFEnabledAndMultiWrite_ReturnsTrue` asserted `IsTrue` while passing `enableMultiWriteRegionHedge: false` — it was encoding the exact bug you reported. It's now `ShouldHedge_WriteRequest_MultiWriteAccount_PpafDoesNotOverrideOptOut` and asserts `IsFalse`. Flagging it because "changed a failing assertion" deserves a second look on principle.

**2. I removed all 10 reflection-on-private-state sites.** The old tests reached into `ppafEnabled` via reflection, which is a large part of why the concurrency race stayed invisible in a green suite. Every PPAF assertion is now behavioral. Net test count went up by 7.

The endpoint-collapse issue you spotted was reproducible, not theoretical — `ResolveServiceEndpoint_HonorsPinnedPpafHedgeTargetEndpoint` asserts the old primary-collapse as an explicit baseline before asserting the fix, so the test documents the bug.

Verification: 253/253 unit tests pass across LocationCache / GlobalEndpointManager / GlobalPartitionEndpointManager / ClientRetryPolicy / MetadataHedgingStrategy; `Microsoft.Azure.Cosmos`, `.Tests`, and `.EmulatorTests` all build clean. The emulator tests compile against the new internal signatures but I let CI run them rather than provisioning the emulator locally. Public API contracts are unchanged — the regressed `enableMultiWriteRegionHedge` default lived only on the internal PPAF factory, not on `CrossRegionHedgingStrategy`.

PR #5257, opened 2025-06-24
QueryPlan : Fixes 410 Gone Exception on non-x64 platforms

### LINKED ISSUE #5220: [Hierarchical Partition Key] - Fix the 410/1002 Gone Exception during queries where customer passes a partial PK in "QueryRequestOptions"
Issue Description:

Issue occurs on Cosmos Container with Hierarchical Partition Key (HPK) setup (More than 1 level of Partition Keys)

The issue is primarily occurring primarily in combination of two features

- When the customer passes in the Partial (parent) partition key in the query request options and partial partition key resolves to more than 2 physical partitions

- When SDK determines to obtain the query plan from gateway. - This is done for non x64 platforms on which the SDK is running.

Example: 

Setup: A container setup with two level HPK ["/State", "/City")

Query with partial PK: i.e. State = 'WA' passed in via `QueryRequestOptions`

`                using (FeedIterator<Location> setIterator = container.GetItemLinqQueryable<Location>(
                    requestOptions: new QueryRequestOptions
                    {
                        PartitionKey = firstLevelHpk
                    })
                    .Where(p => p.State == "WA").ToFeedIterator())
                {
                    // Asynchronous query exceution.
                    while (setIterator.HasMoreResults)
                    {
                        foreach (Object item in await setIterator.ReadNextAsync())
                        {
                            Console.WriteLine(item);
                        }
                    }
                } 
`

The above code with the SDK run on a non x64 platform causes a HTTP 410

` Assert.Fail failed. Exception: Microsoft.Azure.Cosmos.CosmosException : Response status code does not indicate success: Gone (410); Substatus: 1002; ActivityId: 2af9a52c-4357-4c78-97fd-d7681f7afcd7; Reason: (Epk Range: [0845FB119899DE50766A2C4CEFC2FA73,0845FB119899DE50766A2C4CEFC2FA73FF) is gone.);`

**BUG:**

**TLDR;**
 The issue is happening because during the HTTP call for obtaining Query plan using RequestInvokeHandler invocation , the code does a getOverlappingRanges computation and this computation will result in more than one partition for a partial (parent) partition key . This causes the code to throw the 410 

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs#L306

Since the query plan itself failed, the SDK fails the request. This doesn;t happen in the case of a x64 platform since the query plan happens via service interop local DLL call.

**Background:**
Query pipeline is going through these series of steps

- Query Context construction: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs#L84

- Determine query plan :https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs#L595

   This will return query ranges (based on the partition key) if provided .An example shown below

- The SDK Capture target ranges from the  https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs#L262

         Based on target ranges and type of the query ( is it single partition bound or cross parttion bound)

* If cross partition bound, create parallel query pipelines https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs#L499

* If single partition, create full query pipeline
The query pipeline materializes by invoking RequestInvokeHandler with query as operationType.

- Inside RequestInvokeHandler, based on the FeedRange , the routing map is invoked to obtain the EPK range of the FeedRange which will be then sent as request.Headers.StartEpk, request.Headers.EndEpk  and request.PartitionKeyRangeId headers

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs#L288

What's happening in this issue when Gateway Query Plan is invoked?

When partition key is passed in QueryRequestOptions, SDK attempts to do a do a "query Plan" call using gateway HTTP call. RequestInvokeHandler receives the request and does a getOverlappingRanges computation and this computation will result in more than one partition for a partial (parent) partition key . This causes the code to throw the 410 

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs#L306

dibahlfi:
Customer faced an issue with their HPK container set up. Problem was when they set the root level partition key in the QueryRequestOptions and run a query they would get the data back when running on windows but for other environments(like docker) they were getting an exception
For Docker they were getting 410(Gone) exception 
Microsoft.Azure.Cosmos.CosmosException : Response status code does not indicate success: Gone (410); Substatus: 1002; ActivityId: 623305e2-9a69-4b65-aa9b-911ee8ac3533; Reason: (Epk Range: [0DCEB8CE51C6BFE84F4BD9409F69B9BB,0DCEB8CE51C6BFE84F4BD9409F69B9BBFF) is gone.);

The root of the problem was that for non-window(x64) environment we were getting the query plan from gateway(unlike windows where we use the ServiceInterop dll) and after we got the plan we funnel it through RequestInvokerHandler pipeline. The pipeline has a check where it sees that if a feedRange is provided and if it returns more than 1 overlapping ranges it throws a gone exception but in this particular case this is a valid scenario(edge case).
e.g
datatable (MinEPK: string, MaxEPK: string, PKrangeId) 
[
"0D4DC2CD8F49C65A8E0C5306B61B4343","0DCEB8CE51C6BFE84F4BD9409F69B9BB2164DEBD78C50C850E0C1E3E3F0579ED",0
 "0DCEB8CE51C6BFE84F4BD9409F69B9BB2164DEBD78C50C850E0C1E3E3F0579ED","1080F600C27CF98DC13F8639E94E7676" 1
]
If our provided partition key results in a hash e.g 0DCEB8CE51C6BFE84F4BD9409F69B9BB that lies between the above 2 records, we will get overlapping ranges back.

Fix:
Fix it to explicitly make a check and see if its a Query Plan , if it is then allow it to progress.

Testing:
I was able to get access to the Defender Team DB where they were facing this issue. I manually tested the following E2E use cases and got the desired results.
On Windows -> with PartitionKey specified
Diagnostics is : {"Summary":{"DirectCalls":{"(449, 5350)":6,"(200, 0)":2},"GatewayCalls":{"(200, 0)":4,"(304, 0)":1}}
Count: 24025; Severity: Medium; 
Count: 24114; Severity: Informational; 
Count: 23; Severity: High; 

On Non_windows - with PartitionKey specified(in the QueryRequestOptions)
Diagnostics is : {"Summary":{"DirectCalls":{"(449, 5350)":2,"(200, 0)":2},"GatewayCalls":{"(200, 0)":5,"(304, 0)":1,"(0, 0)":1}}
Count: 24025; Severity: Medium; 
Count: 24114; Severity: Informational; 
Count: 23; Severity: High;

On Windows - no QueryRequestOption specified
Diagnostics is : {"Summary":{"DirectCalls":{"(200, 0)":2},"GatewayCalls":{"(200, 0)":4,"(304, 0)":1}}
Count: 24025; Severity: Medium; 
Count: 24114; Severity: Informational; 
Count: 23; Severity: High; 

On Non_windows - no QueryRequestOption specified
Diagnostics is : {"Summary":{"DirectCalls":{"(200, 0)":2},"GatewayCalls":{"(200, 0)":5,"(304, 0)":1}}
Count: 24025; Severity: Medium; 
Count: 24114; Severity: Informational; 
Count: 23; Severity: High; 

Problem writing a new reliable E2E test in the pipeline is the set up needed to return overlapping ranges and we don't have control over when partition split happens.

Talked to Ananth on how he tested it and he worked with Elasticity team to create the database in the desired state, we can do the same in the pipeline but need to check if our database accounts/containers get cleaned up by some background job in the future(will check with Nalu on this one).
I also explored the PartitionKeyHashRangeSplitterAndMerger class but that works only with in-memory container and doesn't follow the RequestInvokerHandler pipeline flow.

for the time being I added a unit test which will avoid any regression in the future.

closes #5220
https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5220

### DIFF
--- a/Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
+++ b/Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
@@ -13,7 +13,7 @@
     <Title>Microsoft Azure Cosmos DB client-side encryption library for multi-tenant</Title>
     <PackageId>Microsoft.Azure.Cosmos.Encryption.Custom</PackageId>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageLicenseFile>LICENSE</PackageLicenseFile>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/Azure/azure-cosmos-dotnet-v3</PackageProjectUrl>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
@@ -30,10 +30,6 @@
 
   <ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' == 'True' ">
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.41.0-preview.0" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <None Include="..\..\LICENSE" Pack="true" PackagePath="" Visible="false" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Microsoft.Azure.Cosmos.Encryption/src/Microsoft.Azure.Cosmos.Encryption.csproj
+++ b/Microsoft.Azure.Cosmos.Encryption/src/Microsoft.Azure.Cosmos.Encryption.csproj
@@ -17,7 +17,7 @@
     <Title>Microsoft Azure Cosmos DB client-side encryption library</Title>
     <PackageId>Microsoft.Azure.Cosmos.Encryption</PackageId>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageLicenseFile>LICENSE</PackageLicenseFile>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/Azure/azure-cosmos-dotnet-v3</PackageProjectUrl>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
@@ -34,10 +34,6 @@
 
   <ItemGroup>
     <None Include="..\..\Icon.png" Pack="true" PackagePath=""/>
-  </ItemGroup>
-
-  <ItemGroup>
-    <None Include="..\..\LICENSE" Pack="true" PackagePath="" Visible="false" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' == 'True' ">
--- a/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
+++ b/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/BenchmarkConfig.cs
@@ -31,6 +31,20 @@ namespace CosmosBenchmark
         [JsonIgnore]
         public string Key { get; set; }
 
+        [Option("isthinclientenabled", Required = false, HelpText = "ThinClient enabled")]
+        public string IsThinClientEnabledRaw { get; set; }
+        public bool IsThinClientEnabled => string.Equals(this.IsThinClientEnabledRaw, "true", StringComparison.OrdinalIgnoreCase);
+
+        [Option("isgatewaymodeenabled", Required = false, HelpText = "Gateway mode enabled")]
+        public string IsGatewayModeEnabledRaw { get; set; }
+        public bool IsGatewayModeEnabled => string.Equals(this.IsGatewayModeEnabledRaw, "true", StringComparison.OrdinalIgnoreCase);
+
+        [Option("isdirectmodeenabled", Required = false, HelpText = "Direct mode enabled")]
+        public string IsDirectModeEnabledRaw { get; set; }
+        public bool IsDirectModeEnabled => string.Equals(this.IsDirectModeEnabledRaw, "true", StringComparison.OrdinalIgnoreCase);
+
+
+
         [Option(Required = false, HelpText = "Workload Name, it will override the workloadType value in published results")]
         public string WorkloadName { get; set; }
 
@@ -209,7 +223,7 @@ namespace CosmosBenchmark
             return this.WorkloadName ?? this.WorkloadType ?? BenchmarkConfig.UserAgentSuffix;
         }
 
-        internal Microsoft.Azure.Cosmos.CosmosClient CreateCosmosClient(string accountKey)
+        internal Microsoft.Azure.Cosmos.CosmosClient CreateCosmosClient()
         {
             // Overwrite the default timespan if configured
             if(this.TelemetryScheduleInSec > 0)
@@ -223,6 +237,7 @@ namespace CosmosBenchmark
                 MaxRetryAttemptsOnRateLimitedRequests = 0,
                 MaxRequestsPerTcpConnection = this.MaxRequestsPerTcpConnection,
                 MaxTcpConnectionsPerEndpoint = this.MaxTcpConnectionsPerEndpoint,
+                ConnectionMode = (this.IsThinClientEnabled || this.IsGatewayModeEnabled) ? Microsoft.Azure.Cosmos.ConnectionMode.Gateway: Microsoft.Azure.Cosmos.ConnectionMode.Direct,
                 CosmosClientTelemetryOptions = new Microsoft.Azure.Cosmos.CosmosClientTelemetryOptions()
                 {
                     DisableSendingMetricsToService = !this.EnableTelemetry,
@@ -243,7 +258,7 @@ namespace CosmosBenchmark
 
             return new Microsoft.Azure.Cosmos.CosmosClient(
                         this.EndPoint,
-                        accountKey,
+                        this.Key,
                         clientOptions);
         }
 
--- a/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Program.cs
+++ b/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Program.cs
@@ -35,6 +35,8 @@ namespace CosmosBenchmark
             try
             {
                 BenchmarkConfig config = BenchmarkConfig.From(args);
+                Environment.SetEnvironmentVariable("AZURE_COSMOS_THIN_CLIENT_ENABLED", config.IsThinClientEnabled.ToString());
+
                 await AddAzureInfoToRunSummary();
 
                 MeterProvider meterProvider = BuildMeterProvider(config);
@@ -71,6 +73,8 @@ namespace CosmosBenchmark
             }
             finally
             {
+                Environment.SetEnvironmentVariable("AZURE_COSMOS_THIN_CLIENT_ENABLED", "False");
+
                 Utility.TeeTraceInformation($"{nameof(CosmosBenchmark)} completed successfully.");
                 if (Debugger.IsAttached)
                 {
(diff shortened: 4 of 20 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/RetryHandlerTests.cs
```diff
@@ -18,6 +24,74 @@ namespace Microsoft.Azure.Cosmos.Tests
...
+                .ReturnsAsync(containerProps);
+
+            Mock<Cosmos.Database> databaseMock = new Mock<Cosmos.Database>();
+            databaseMock.Setup(d => d.Id).Returns("testDb");
+            containerMock.Setup(c => c.Database).Returns(databaseMock.Object);
+
+            // Mock PartitionKeyRangeCache
+            List<PartitionKeyRange> overlappingRanges = new List<PartitionKeyRange>
+            {
+                    new PartitionKeyRange { Id = "0", MinInclusive = "A", MaxExclusive = "Z" },
```
kirankumarkolli: Ideal should be non-overlapping

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -305,17 +305,25 @@ public virtual async Task<ResponseMessage> SendAsync(
...
+                                    CosmosException goneException = new CosmosException(
+                                        message: $"Epk Range: {feedRangeEpk.Range} is gone.",
+                                        statusCode: System.Net.HttpStatusCode.Gone,
+                                        subStatusCode: (int)SubStatusCodes.PartitionKeyRangeGone,
+                                        activityId: Guid.NewGuid().ToString(),
+                                        requestCharge: default);
+
+                                    return goneException.ToCosmosResponseMessage(request);
+                                }
+                                // For query operations spanning multiple partitions, we don't set PartitionKeyRangeId
```
kirankumarkolli: Is comment relevant for this change?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/RetryHandlerTests.cs
kirankumarkolli: Prefer use the values from the issue.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/RetryHandlerTests.cs
```diff
@@ -18,6 +24,74 @@ namespace Microsoft.Azure.Cosmos.Tests
     public class RetryHandlerTests
     {
         private static readonly Uri TestUri = new Uri("https://dummy.documents.azure.com:443/dbs");
+        [TestMethod]
+        public async Task ValidatePassingOverlappingRangesInQueryPlanDoesntThrowAnException()
```
kirankumarkolli: Please add E2E test coverage

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
dibahlfi: removed

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/RetryHandlerTests.cs
dibahlfi: fixed

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -305,17 +305,22 @@ public virtual async Task<ResponseMessage> SendAsync(
...
-                                CosmosException goneException = new CosmosException(
-                                    message: $"Epk Range: {feedRangeEpk.Range} is gone.",
-                                    statusCode: System.Net.HttpStatusCode.Gone,
-                                    subStatusCode: (int)SubStatusCodes.PartitionKeyRangeGone,
-                                    activityId: Guid.NewGuid().ToString(),
-                                    requestCharge: default);
-
-                                return goneException.ToCosmosResponseMessage(request);
+                                //If we are running a query plan and our provided partition key results in a hash that resolves to more than one EPKRanges then its a valid use case
+                                bool isQueryOperation = request.ResourceType == ResourceType.Document && request.OperationType == OperationType.QueryPlan;
```
FabianMeiswinkel:
```suggestion
                                bool isQueryPlanOperation = request.ResourceType == ResourceType.Document && request.OperationType == OperationType.QueryPlan;
```

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -305,17 +305,22 @@ public virtual async Task<ResponseMessage> SendAsync(
...
+                                if (!isQueryOperation)
```
FabianMeiswinkel:
```suggestion
                                if (!isQueryPlanOperation)
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/RetryHandlerTests.cs
```diff
@@ -18,6 +24,74 @@ namespace Microsoft.Azure.Cosmos.Tests
...
+
+            // FeedRangeEpk for the test
+            FeedRangeEpk feedRange = new FeedRangeEpk(new Documents.Routing.Range<string>("0DCEB8CE51C6BFE84F4BD9409F69B9BB", "0DCEB8CE51C6BFE84F4BD9409F69B9BBFF", true, false));
+            RequestInvokerHandler invoker = new RequestInvokerHandler(client, null, null, null)
+            {
+                InnerHandler = new TestHandler((request, token) => TestHandler.ReturnSuccess())
+            };
+
+            // Act
+            ResponseMessage response =  await invoker.SendAsync("dbs/testDb/colls/testColl", ResourceType.Document, OperationType.QueryPlan, null, containerMock.Object, feedRange,
```
FabianMeiswinkel: Please add the test that would actually see the GoneException for OperationType == Query. So, the unit test should cover with the same FeedRange overlapping multiple phyiscal partitions that QueryPlan does not throw - while Query throws.

FabianMeiswinkel (dismissed):
LGTM - I am fine with just using unit test instead of E2E test based on the reaosns in the PR  descritpion. But I left a few minor comments - I am good once those are addressed.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
dibahlfi: resolved

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
dibahlfi: changed

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/RetryHandlerTests.cs
dibahlfi: resolved offline in the Cosmos .NET Crew.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/RetryHandlerTests.cs
dibahlfi: done

NaluTripician (approved):
LGTM

kirankumarkolli (approved):

PR #5264, opened 2025-06-30
[FaultInjection] ThinClient Compatibility: Adds compatibility with Thin Client Proxy. 

### LINKED ISSUE #5256: [Thin Client Integration] Add Fault Injection support for Thin Client in .NET SDK
### Acceptance Criteria:

Add Fault Injection support for Thin Client in .NET SDK.

NaluTripician:
# Pull Request Template

## Description

This PR allows FaultInjection rules that target gateway requests to also work with the ThinClient Proxy. 

## Type of change

## Closing issues

To automatically close an issue: closes #5256


Microsoft.Azure.Cosmos/FaultInjection/src/FaultInjectionCondition.cs
```diff
@@ -83,7 +83,7 @@ public FaultInjectionEndpoint GetEndpoint()
         public override string ToString()
         {
             return String.Format(
-                "FaultInjectionCondition{{ OperationType: {0}, ConnectionType: {1}, Region: {2}, Endpoint: {3}",
+                "FaultInjectionCondition{{ OperationType: {0}, ConnectionType: {1}, Region: {2}, Endpoint: {3}, LimitToProxy: {4}",
```
kundadebdatta: Are we still planning to keep the `LimitToProxy` ? We can park this for any future ASKs.

NaluTripician: Removed. This will be added in #5264 if we decide to move forward with the ThinClient proxy gateway fallback. If now we will not add this.

Microsoft.Azure.Cosmos/FaultInjection/tests/Utils/TestCommon.cs
```diff
@@ -24,6 +24,11 @@ internal static string GetConnectionString()
             return ConfigurationManager.GetEnvironmentVariable<string>("COSMOSDB_MULTI_REGION", string.Empty);
         }
 
+        internal static string GetThinClientConnectionString()
+        {
+            return ConfigurationManager.GetEnvironmentVariable<string>("COSMOSDB_THIN_CLIENT", string.Empty);
```
kundadebdatta, line 29: Is this for real account testing ? If so - do you have an proxy setup ?

kirankumarkolli:
> FI is a public library and this change ideal to go into the release notes

Also not all kinds of faulults are not applicable for ThinProxy. At-least worth calling out which ones are applicable vs not. 
Even ideal if we can error out.

Microsoft.Azure.Cosmos/FaultInjection/src/implementation/FaultInjectionConditionInternal.cs
```diff
@@ -6,20 +6,24 @@ namespace Microsoft.Azure.Cosmos.FaultInjection
...
         private string containerResourceId = string.Empty;
         private OperationType? operationType = null;
         private List<Uri> regionEndpoints = new List<Uri>{ };
         private List<Uri> physicalAddresses = new List<Uri> { };
 
-        public FaultInjectionConditionInternal()
+        public FaultInjectionConditionInternal(GlobalEndpointManager globalEndpointManager)
         {
+            this.globalEndpointManager = globalEndpointManager 
+                ?? throw new ArgumentNullException(nameof(globalEndpointManager), "Argument 'globalEndpointManager' cannot be null.");
```
kirankumarkolli: For NUllException the name is already present message is not needed. 
Any static tools warning about it?

Microsoft.Azure.Cosmos/FaultInjection/src/implementation/FaultInjectionConditionInternal.cs
```diff
@@ -201,8 +205,8 @@ public bool IsApplicable(string ruleId, ChannelCallArguments args)
             public bool IsApplicable(string ruleId, DocumentServiceRequest request)
             {
                 bool isApplicable = this.regionEndpoints.Any(uri => 
-                    request.RequestContext.LocationEndpointToRoute.AbsoluteUri
-                    .StartsWith(uri.AbsoluteUri));
+                    request.RequestContext.LocationEndpointToRoute.Host
```
kirankumarkolli, line 207: Isn't this a breaking change? 

Or a new rule should be defined explicitly for Host scope?

Microsoft.Azure.Cosmos/FaultInjection/tests/Utils/TestCommon.cs
NaluTripician: Yes, this will be for the fault injection release gates. We will do not have the fault injection library test run on a PR-to-PR basis, and changes are verified locally. This will use the account that Arooshi has already set up.

Microsoft.Azure.Cosmos/FaultInjection/src/implementation/FaultInjectionConditionInternal.cs
NaluTripician: Will remove check. It looks like there is already one for the global endpoint manager where this is created so it will never be null.

Microsoft.Azure.Cosmos/FaultInjection/src/implementation/FaultInjectionConditionInternal.cs
NaluTripician: No this will not be a breaking change. This is checking for the regional endpoint. 

Currently if an endpoint is `accountName-region.documents.azure.com/` the AbsoluteUri will compare that to the Thin proxy endpoint which will be `accountName-region.documents.azure.com:10250/`. This will cause the rules to fail even if they should be applicable. By comparing just, the host we will only look at the `accountName-region.documents.azure.com` part which is what we care about. This way we can be sure that rules can be applicable regardless of whether they are for the gateway or thin proxy which is the goal.

PR #5273, opened 2025-07-07
Query: Fixes non streaming order by queries to not be tagged as passthrough queries

neildsh:
## Description

Since the addition of non-streaming order by, it is no longer safe to treat single partition order by queries as passthrough, as it can lead to incorrect results for `ORDER BY VectorDistance` and `ORDER BY RANK` queries. This PR fixes the issue by adding a check for non-streaming order by queries in the passthrough query creation logic.

## Type of change


adityasa (dismissed):
:shipit:

adityasa (approved):
:shipit:

PR #5298, opened 2025-07-17
Fix Expression constant folding inside of `MemberInitExpression`

### LINKED ISSUE #1664: Error with Member Indexer when mapping value to a class model in select (if not const value)
I'm having issue with Cosmos DB where it throws an error if I try to map an indexed member to a property on a class.  However it works fine if I switch the query to map to an anonymous type or use a const value.

**Some Code to set the stage**
```csharp
CosmosClient client = new CosmosClient(connectionString);
var container = client.GetContainer(DatabaseId, CollectionId);
var q = container.GetItemLinqQueryable<Entity>(allowSynchronousQueryExecution: true);
var testName = "Test";

//Class used in mapping for select 
public class Model{
	public string Name { get; set; }
}

//Sample CosmosDB document entity structure
public class Entity
{
	public Child Child { get; set; }
}
public class Child
{
	public string Name { get; set; }
	public Dictionary<string, Child> NamedChildren { get; set; }
}
```

# This will work #
```csharp
var items = q.Select(q => new { Name = q.Child.NamedChildren[testName].Name }).ToList();
```
**And produces this working sql**
```sql
SELECT VALUE {"Name": root["Child"]["NamedChildren"]["Test"]["Name"]} FROM root
```
# This will NOT work #
**If I map the select to the model class (see "new Model" below) it will throw an error**
```csharp
var items = q.Select(q => new Model { Name= q.Child.NamedChildren[testName].Name }).ToList()
```
**Produces this incorrect sql**

```sql
SELECT VALUE {"Name": root["Child"]["NamedChildren"][{"testName": "Test"}["testName"]]["Name"]} FROM root
and the following error
```
**and this error from Cosmos DB**

Microsoft.Azure.Cosmos.Query.Core.Exceptions.ExpectedQueryPartitionProviderException: 

```js
{
"errors":[
{
"severity":"Error",
"location":
{
"start":22,
"end":86
},
"code":"SC2990",
"message":"The specified query includes 'member indexer' which is currently not supported."}
]
}
```

# Using a Const Changes the outcome though #
If I change 
```csharp
var testName = "test";
```
to be
```csharp
const string testName = "test";
```

Then the code that is not working works and the 'sql' changes to the expected
```sql
SELECT VALUE {"Name": root["Child"]["NamedChildren"]["Test"]["Name"]} FROM root
```

I could fill the collection of anonymous types and then map to my strong typed collection, but obviously not ideal.

I'm having trouble understanding why it would fail as I would have thought the class mapping happened after the query was finished and returned.  However the actual query/SQL is different.

Is this a bug or is there something I'm missing as to why this should fail or work differently.

In case it gets answered on StackOverflow

https://stackoverflow.com/questions/62563549/cosmos-db-error-with-member-indexer-when-mapping-value-to-a-class-model-in-selec

MichalMarsalek:
# Pull Request Template

## Description

This PR fixes a bug described in #1664 where captured variables in LINQ expressions would not get expanded leading to a `{"testName": "Test"}["testName"]` expression (instead of the expected `"Test"`) in the translated SQL, leading to an error reported by Cosmos.

However the problem appears to be more general as the issue seems to be that the recursion that partially evaluates expressions (so among other things resolves the captured variables and therefore effectively changes `{"testName": "Test"}["testName"]` to `"Test"`) terminates whenever it encounters a node of type `MemberInitExpression`. This explains the difference in behaviour between anonymous types and a class as described in #1664, as anonymous types do not contain the `MemberInitExpression` node.

This is quite surprising since at least in the way I am used to using this library, vast majority of the expressions defining the document projections have the `MemberInitExpression` as the very root of the body of the `Expression<Func<T1, T2>>`, meaning that in this common case no constant folding would *ever* happen. Is that even possible? Or what am I missing here?

## Type of change

## Closing issues

To automatically close an issue: closes #1664


NaluTripician:
Hi @MichalMarsalek, thank you for the contribution! This PR has been idle for ~285 days. Is this change still needed on your side? Flagging for SDK-team triage as part of an open-PR cleanup pass — we'd like to give you a clear yes/no rather than let it sit.

MichalMarsalek:
Hi @NaluTripician. I just rebased my changes and yes, it seems that the problem they are addressing still exist. I think the change is still needed.

NaluTripician:
@sdkReviewAgent

NaluTripician (commented):
## Deep Review — Fix Expression constant folding inside MemberInitExpression

**Verdict:** Net-positive bug fix. No blockers (0 critical, 0 major). The fix correctly recurses into `MemberInit` bindings so closure-captured variables are folded.

The deliberate choice to **not** visit `node.NewExpression` is load-bearing — see the inline comment on `SubtreeEvaluator.cs` for the explanation. This is correct, not a defect.

### Answering your question in the PR description

> "vast majority of the expressions defining the document projections have the `MemberInitExpression` as the very root of the body … no constant folding would ever happen. Is that even possible?"

You aren't missing anything — the original 2018 implementation simply got this wrong, and the existing `LinqTranslationBaselineTests.TestMemberInitializer` baseline only exercises member access on the lambda parameter (`doc.NumericField`), not on captured locals. Anonymous-type projections use `NewExpression` and have always folded correctly, which is why this only surfaced in the relatively rare class-with-member-init case from #1664.

### Observation — design layering (not in scope, worth a follow-up issue)

The `Nominator` nominates expressions bottom-up using the *full* default `ExpressionVisitor`, so it freely adds children of `MemberInitExpression` (including the inner `NewExpression`) to the candidate set. But `SubtreeEvaluator` cannot safely evaluate those candidates — folding a `NewExpression` inside a `MemberInit` would break the parent. The current workaround is "in `SubtreeEvaluator`, manually opt out of visiting paths that would lead us to fold something the parent can't accept." A more principled design would either (a) have the `Nominator` understand parent constraints, or (b) make `EvaluateConstant` return the original expression unchanged when the result would violate a structural constraint of the consumer. Worth filing a tracking issue; the customer-visible fix here is the right next step.

### Findings overview

All inline comments are 🟡 Recommendations or 🟢 Suggestions — mostly test-coverage gaps. Please consider addressing #1, #2, #4, #5 before merge.

Microsoft.Azure.Cosmos/src/Linq/SubtreeEvaluator.cs
```diff
@@ -40,7 +41,9 @@ public override Expression Visit(Expression expression)
 
         protected override Expression VisitMemberInit(MemberInitExpression node)
         {
-            return node;
+            return Expression.MemberInit(
+                node.NewExpression,
```
NaluTripician: 🟡 **Recommendation — Add a code comment explaining why `node.NewExpression` is intentionally NOT visited**

This is correct and load-bearing — calling `base.VisitMemberInit(node)` (or wrapping in `Visit(node.NewExpression)`) would route the parameterless `new TestClass()` through your overridden `Visit`. Since `CanBeEvaluated` returns `true` for any non-Parameter / non-Lambda expression, the `NewExpression` would be in the candidate set and would be folded to a `ConstantExpression`. `VisitAndConvert<NewExpression>` would then throw `InvalidOperationException` because a `ConstantExpression` isn't a `NewExpression`.

Without a comment, a well-intentioned future cleanup will collapse this to `return base.VisitMemberInit(node);` and reintroduce a runtime crash for every expression of the form `new MyClass() { Prop = capturedVar }`.

Suggested comment:

```csharp
// Rebuild the MemberInit manually and intentionally do NOT visit node.NewExpression.
// The Nominator nominates parameterless `new TestClass()` as a candidate, so visiting
// it would fold it into a Constant of the constructed CLR type. Expression.MemberInit
// requires a NewExpression as its first argument, not a Constant, so that path throws
// InvalidOperationException at runtime. We only need to recurse into the bindings
// to fold closure variables in initializers.
```

Microsoft.Azure.Cosmos/src/Linq/SubtreeEvaluator.cs
```diff
@@ -40,7 +41,9 @@ public override Expression Visit(Expression expression)
...
+                node.Bindings.Select(binding => this.VisitMemberBinding(binding)));
```
NaluTripician: 🟢 **Suggestion — Short-circuit the no-change path to avoid allocations**

The current code always allocates a new `MemberInitExpression` and a new bindings sequence (`.Select(...)`) on every visit, even when no binding contains a closure (the overwhelmingly common case in production). The base-class idiom is:

```csharp
ReadOnlyCollection<MemberBinding> newBindings = Visit(node.Bindings, VisitMemberBinding);
if (newBindings == node.Bindings) return node;            // short-circuit
return Expression.MemberInit(node.NewExpression, newBindings);
```

`SubtreeEvaluator` runs on every LINQ query, so this is measurable churn on hot query paths. Not blocking, but worth fixing while you're already in the file.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    using System.Reflection;
+    using System.Text.Json.Serialization;
+    using global::Azure.Core.Serialization;
+    using Microsoft.Azure.Cosmos.Serializer;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Converters;
+
+    [TestClass]
+    public class ConstantEvaluatorTests
```
NaluTripician, line 14: 🟡 **Recommendation — Missing end-to-end SQL baseline coverage**

This unit test verifies the mechanism in `ConstantEvaluator.PartialEval`, but nothing in the PR proves the *SQL output* is correct end-to-end. The existing `Microsoft.Azure.Cosmos.EmulatorTests/Linq/LinqTranslationBaselineTests.cs::TestMemberInitializer` (around line 875) already covers `Select(doc => new DataObject { NumericField = doc.NumericField, ... })` — but **none of its cases use a captured variable**, which is exactly the regression #1664 is about.

Please add a case like:

```csharp
string testName = "Test";
new LinqTestInput("Select w/ DataObject initializer + captured var",
    b => getQuery(b).Select(doc => new DataObject { StringField = doc.StringField + testName }))
```

That will generate a baseline XML showing the captured `"Test"` correctly inlined into the SQL — protecting against regressions at the boundary the customer in #1664 actually hit. Without it, `SubtreeEvaluator` could be re-broken by an unrelated change and only this unit test would catch it.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
```diff
@@ -0,0 +1,46 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Linq
+{
+    using System;
+    using System.Collections.Generic;
```
NaluTripician, line 8: 🟡 **Recommendation — Remove unused `using` directives**

At least 8 of the 12 imports are unused — this looks like template scaffolding from another test file in the same folder:

- `System.Collections.Generic`
- `System.Globalization`
- `System.IO`
- `System.Reflection`
- `System.Text.Json.Serialization`
- `global::Azure.Core.Serialization`
- `Microsoft.Azure.Cosmos.Serializer`
- `Newtonsoft.Json`
- `Newtonsoft.Json.Converters`

The repo doesn't fail CI on `IDE0005`, so this won't block merge, but it adds confusion about what the test actually depends on. Reduce to: `System`, `System.Linq.Expressions`, and `Microsoft.VisualStudio.TestTools.UnitTesting`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    using Microsoft.Azure.Cosmos.Serializer;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Converters;
+
+    [TestClass]
+    public class ConstantEvaluatorTests
+    {
+        [TestMethod]
+        public void ClosuresAreEvaluated()
```
NaluTripician: 🟡 **Recommendation — Split into two tests and use a more descriptive name**

This method conflates two independent scenarios — `MemberInit` (the actual fix target) and anonymous object (a sanity check that already worked before the fix, because anonymous types use `NewExpression`, not `MemberInitExpression`). If the first assertion regresses, the second is never reached, and the reporter shows a single failure masking which scenario broke. Mixing them in one method also blurs the distinction that anonymous-object folding is a baseline assertion, not a regression test for this PR's change.

Suggest:

```csharp
[TestMethod] public void ClosuresInsideMemberInitExpressionAreFolded() { ... }
[TestMethod] public void ClosuresInsideAnonymousObjectAreFolded() { ... }
```

The current name `ClosuresAreEvaluated` also doesn't encode *where* — six months from now a CI failure with that name gives no clue whether the regression is in `MemberInitExpression`, anonymous objects, or some other path.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            int a = 1;
+            // In class member init
+            Expression<Func<int, object>> closureInClassExpression = x => new TestClass { Property = x + a };
```
NaluTripician: 🟡 **Recommendation — Add a test that reproduces the exact #1664 scenario**

The reported bug is specifically about `q.Child.NamedChildren[testName].Name` — a **dictionary indexer keyed by a closure-captured variable**. The test here uses `x + a` (a `BinaryExpression`), which is a different node shape. While the underlying mechanism is the same (recursion into bindings), a regression in `EvaluateMemberAccess` or indexer handling would not be caught by the current test.

Please add a case that mirrors the bug report, e.g.:

```csharp
var dict = new Dictionary<string, string> { ["k"] = "Test" };
string testKey = "k";
Expression<Func<int, object>> expr = x => new TestClass { Name = dict[testKey] };
// Expect the body to fold to `new TestClass { Name = "Test" }`
```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            Expression<Func<int, object>> closureInClassExpressionExpected = x => new TestClass { Property = x + 1 };
+            Expression closureInClassExpressionActual = ConstantEvaluator.PartialEval(closureInClassExpression.Body);
+            Assert.AreEqual(closureInClassExpressionExpected.Body.ToString(), closureInClassExpressionActual.ToString());
```
NaluTripician: 🟢 **Suggestion — `Expression.ToString()` equality is a fragile assertion**

`Expression.ToString()` is the *debug-display* implementation — it is not contractual, and parenthesization / formatting have shifted between .NET versions in the past. A future runtime upgrade can either spuriously fail this test, or silently mask a real regression that happens to render the same way.

Note: functional comparison via `.Compile().Invoke()` does **not** work here either — both expressions yield the same value regardless of whether the closure was inlined. So a structural check is required:

```csharp
var memberInit = (MemberInitExpression)closureInClassExpressionActual;
var binding = (MemberAssignment)memberInit.Bindings[0];
var binary = (BinaryExpression)binding.Expression;
Assert.IsTrue(binary.Right is ConstantExpression { Value: 1 });
```

Or introduce an `ExpressionEqualityComparer`-style helper if the project doesn't already have one.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            // In anonymous object
+            Expression<Func<int, object>> closureInAnonymousObjectExpression = x => new { Property = x + a };
+            Expression<Func<int, object>> closureInAnonymousObjectExpressionExpected = x => new { Property = x + 1 };
+            Expression closureInAnonymousObjectExpressionActual = ConstantEvaluator.PartialEval(closureInAnonymousObjectExpression.Body);
+            Assert.AreEqual(closureInAnonymousObjectExpressionExpected.Body.ToString(), closureInAnonymousObjectExpressionActual.ToString());
```
NaluTripician: 🟡 **Recommendation — Missing coverage of other `MemberBinding` kinds and nested `MemberInit`**

The fix changes behavior from "do nothing" to calling `VisitMemberBinding` on each binding. `VisitMemberBinding` dispatches to three overrides:

- `VisitMemberAssignment` — covered ✓
- `VisitMemberMemberBinding` — for `new Outer { Inner = { Prop = capturedVar } }` — **not covered**
- `VisitMemberListBinding` — for `new Outer { Items = { capturedVar } }` — **not covered**

Also missing: **nested `MemberInit`** like `new Outer { Inner = new Inner { Prop = capturedVar } }`. This is the most likely real-world shape after the bug-report scenario, and the recursive re-entry into `VisitMemberInit` is the trickiest path to get right. Please add at least one nested test.

Microsoft.Azure.Cosmos/src/Linq/SubtreeEvaluator.cs
NaluTripician: Done in 8fc8b9c. Added the load-bearing intent comment explaining why `node.NewExpression` is intentionally not visited (Nominator marks parameterless `new T()` as a candidate, `EvaluateConstant` would fold it to a `ConstantExpression`, and `Expression.MemberInit` requires a `NewExpression` first arg — so the alternative path throws `InvalidOperationException` at runtime).

Microsoft.Azure.Cosmos/src/Linq/SubtreeEvaluator.cs
NaluTripician: Done in 8fc8b9c. Switched to the `protected static ExpressionVisitor.Visit<T>(ReadOnlyCollection<T>, Func<T, T>)` helper and short-circuit return the original node when reference equality holds on the bindings collection. No more per-visit `Select` + `MemberInit` allocations in the (common) no-change case.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
NaluTripician: Acknowledged — deferring this one to a follow-up rather than landing it here. Adding a `LinqTranslationBaselineTests.TestMemberInitializer` entry requires regenerating the corresponding XML baseline against the Cosmos emulator, which I don't have available in this environment, and the deep-review summary already calls out filing a tracking issue. The new `ClosuresUsedAsDictionaryIndexerInsideMemberInitAreFolded` unit test covers the exact #1664 shape at the constant-folding mechanism level (the layer the fix changes). I'll open a follow-up issue to add the baseline coverage; resolving this thread.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
NaluTripician: Done in 8fc8b9c. Trimmed to the minimal set: `System`, `System.Collections.Generic`, `System.Linq`, `System.Linq.Expressions`, `Microsoft.VisualStudio.TestTools.UnitTesting`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
NaluTripician: Done in 8fc8b9c. `ClosuresAreEvaluated` is replaced by six per-scenario methods so a CI failure names the actual broken shape:
- `ClosuresInsideMemberInitExpressionAreFolded`
- `ClosuresInsideAnonymousObjectAreFolded`
- `ClosuresUsedAsDictionaryIndexerInsideMemberInitAreFolded`
- `ClosuresInsideNestedMemberInitAreFolded`
- `ClosuresInsideMemberMemberBindingAreFolded`
- `ClosuresInsideMemberListBindingAreFolded`

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
NaluTripician: Done in 8fc8b9c as `ClosuresUsedAsDictionaryIndexerInsideMemberInitAreFolded`. One subtlety surfaced while writing it: the parameter must be referenced inside the `MemberInit` (here via `q.StringProperty +`), otherwise the Nominator marks the whole `new TestClass { ... }` itself as a candidate and the outer `Visit` short-circuits straight to `EvaluateConstant` — `VisitMemberInit` never runs. The test now mirrors the #1664 shape: a closure-only sub-tree (`map[capturedKey]`) folded inside a parameter-anchored `MemberInit`. Post-fix the right side of the `Add` is the literal `Constant("Test")` instead of a `Dictionary[indexer]` MethodCall.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
NaluTripician: Done in 8fc8b9c. All six tests now assert structurally via small typed helpers (`AssertMemberInit`, `AssertSingleMemberAssignment`, `AssertBinary`, `AssertParameter`, `AssertConstant`) that walk the rewritten tree and check node kinds + values directly — no more `Expression.ToString()` equality.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/ConstantEvaluatorTests.cs
NaluTripician: Done in 8fc8b9c. Added three new tests covering the gaps:
- `ClosuresInsideNestedMemberInitAreFolded` — `new Outer { Inner = new Inner { Prop = captured } }`
- `ClosuresInsideMemberMemberBindingAreFolded` — `new Outer { Inner = { Prop = captured } }` (`MemberMemberBinding`)
- `ClosuresInsideMemberListBindingAreFolded` — `new Outer { Items = { captured } }` (`MemberListBinding`)
Each asserts the closure is folded to its literal while the parameter reference is preserved.

NaluTripician:
## Re-review after  — ready to merge from review-feedback perspective

I pushed [``](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5298/commits/8fc8b9ce25ac0b41f892e5173272b9ff4e729975) directly to `fix-#1664` (maintainer_can_modify) to address the inline review feedback. All 8 review threads are now ✅ resolved.

### What changed since the previous review pass
- **`SubtreeEvaluator.cs`**: added a load-bearing intent comment on `VisitMemberInit` and switched the bindings traversal to the protected static `ExpressionVisitor.Visit<T>` helper with a reference-equality short-circuit (no allocations on the common no-change path).
- **`ConstantEvaluatorTests.cs`**: removed 8 stale `using` directives, split the single combined test into 6 per-shape tests with descriptive names, replaced fragile `Expression.ToString()` equality with structural shape assertions, and added coverage for the gaps:
  - dictionary indexer keyed by a closure variable (the exact #1664 shape)
  - nested `MemberInit`
  - `MemberMemberBinding` (`new Outer { Inner = { Prop = captured } }`)
  - `MemberListBinding` (`new Outer { Items = { captured } }`)

### Verification done locally
- `dotnet build` on `src` + `Microsoft.Azure.Cosmos.Tests`: clean (0 warnings, 0 errors)
- `dotnet test --filter ~ConstantEvaluatorTests`: 6 pass / 0 fail
- `dotnet test --filter ~Linq`: 26 pass / 0 fail
- Full `Microsoft.Azure.Cosmos.Tests` run: 2 637 pass, 11 skipped, 15 failed — all 15 failures are `BinaryEncodingOverTheWireTests` `HttpRequestException: No connection` (pre-existing infra dependency, unrelated to this PR)

### Deferred — needs follow-up issue (not blocking this PR)
- End-to-end SQL baseline coverage in `LinqTranslationBaselineTests.TestMemberInitializer` requires regenerating the XML baseline against the Cosmos emulator. The unit test `ClosuresUsedAsDictionaryIndexerInsideMemberInitAreFolded` covers the #1664 shape at the constant-folding mechanism level (the layer this fix changes); the baseline coverage is a defense-in-depth addition that I'll track separately.

### Remaining merge gates (out of scope for review feedback)
- `/azp run` needs to be issued by an authorized user to start the Azure Pipelines CI
- An approving review

Net: from a review-feedback perspective the PR is in good shape to merge once CI is green.

NaluTripician:
Confirmed the unit tests in this PR pass locally against the PR head (`8fc8b9ce2`):

```
dotnet test Microsoft.Azure.Cosmos.Tests --filter ConstantEvaluatorTests
  Passed ClosuresInsideMemberInitExpressionAreFolded
  Passed ClosuresInsideAnonymousObjectAreFolded
  Passed ClosuresUsedAsDictionaryIndexerInsideMemberInitAreFolded
  Passed ClosuresInsideNestedMemberInitAreFolded
  Passed ClosuresInsideMemberMemberBindingAreFolded
  Passed ClosuresInsideMemberListBindingAreFolded
Total: 6   Passed: 6   Failed: 0

dotnet test Microsoft.Azure.Cosmos.Tests --filter ~Linq
Total: 26  Passed: 26  Failed: 0
```

The two failing CI jobs (`EmulatorTests Release - MultiMaster` and `... - MultiRegion`) are unrelated to this change — they require the `$(COSMOSDB_MULTIMASTER)` and `$(COSMOSDB_MULTI_REGION)` secret variables, which Azure DevOps does not mount on fork PR builds by default, so the env vars arrive empty and `CosmosClient` construction fails before any real test logic runs. The relevant CI job for this PR's changes, `dotnet-v3-ci (Microsoft.Azure.Cosmos.Tests)`, passed.

NaluTripician (dismissed):
Approved, verified live tests pass locally until fork PR issue is fixed

kushagraThapar (changes requested):
PR looks good, request changes on the changelog entry.

NaluTripician:
@kushagraThapar — added a changelog entry under a new `Unreleased` section in `changelog.md` (commit 960135dd8):

> - [5298](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5298) LINQ: Fixes constant folding for closure-captured variables inside MemberInitExpression (resolves #1664). Previously, the recursion that partially evaluates expressions terminated whenever it encountered a `MemberInitExpression` node, so captured variables inside object initializers were not folded, producing invalid translated SQL.

Format matches the existing `Fixed` entries (PR# link → `Category: Verb description`). Ready for re-review.

kushagraThapar (dismissed):
LGTM, thanks @MichalMarsalek !

kushagraThapar (approved):

MichalMarsalek:
@NaluTripician Hi, what is the plan for this? I assumed this was waiting for your PR fixing the CI problem, but I see that's been closed.

NaluTripician (approved):

NaluTripician:
> @NaluTripician Hi, what is the plan for this? I assumed this was waiting for your PR fixing the CI problem, but I see that's been closed.

Hi @MichalMarsalek sorry for the delay I have been out of the office. CI issues are because of issues with the CI for forks. Verified locally and we should get this in soon

MichalMarsalek:
Thank you for resolving this.

PR #5310, opened 2025-07-22
[Internal] PPAF: Adds Dynamic Enablement of PPAF

### LINKED ISSUE #5304: [Per Partition Automatic Failover] - Enable PPAF Dynamically upon change on Account Properties Metadata Response
### Background

As of today, during the cosmos client initialization, we enable PPAF based on the enablement flag, returned by the Gateway in the `DatabaseAccount` response, using the `GET ACCOUNT` metadata call. This setup is good, as the enablement is currently independent of any external factors and depends on the account level settings.

### Problem Statement

However, this approach has a bottleneck. Because the SDK fetches the PPAF enablement flag only during the SDK initialization, in order to get the current status of the flag from the account, the SDK needed to be re-initialized/ bootstrapped. 

This means if the account updates the flag at a later point of time, after the SDK initialization, the customer/ user needs to restart the SDK to take that into effect.

This will become a challenge in the long term.

### Proposed Solution

The SDK needs to do a periodic refresh on the account properties, and update the flag based on the refreshed fetched account metadata. This will remove the dependency on the SDK restart and the customer can have the same SDK running and observe the enablement/ disablement at the run time dynamically.

### Acceptance Criteria

Refactor the SDK code to enable PPAF dynamically, based on the refreshed account properties. We will need to make sure the .NET SDK complies with the PPAF pre-requisites though. i.e. the SDK is able to set and apply the default configs to enable PPAF. They are highlighted below:

- **Read Hedging:** .NET SDK need to ensure that the read hedging is enabled by default, when PPAF is enabled. If a customer has a explicit availability strategy, that needed to be honored and should take the precedence.

- **Per Partition Circuit Breaker:** .NET SDK need to enable circuit breaker, when PPAF is enabled.

NaluTripician:
# Pull Request Template

## Description

This pull request introduces enhancements to the partition-level failover (PPAF) functionality in the Azure Cosmos SDK. The changes include the addition of a default cross-region hedging strategy, dynamic enablement of PPAF based on the database account configuration without a restart of the SDK client, and new tests to validate these behaviors. Below are the most important changes grouped by theme:

### Enhancements to Availability Strategy:
* Added a new method `SDKDefaultCrossRegionHedgingStrategy` in `AvailabilityStrategy.cs` to provide a default hedging strategy for cross-region failover, including support for write requests on multi-region accounts.
* Introduced an internal flag `IsSDKDefaultStrategy` in `CrossRegionHedgingAvailabilityStrategy` to differentiate SDK default strategies from custom ones. Updated the constructor to accept this flag. [[1]](diffhunk://#diff-fdccfbf7c7ab03d54957da31d1545bf40e2d0b854aeb1c1e342e219303c6cc80R47-R52) [[2]](diffhunk://#diff-fdccfbf7c7ab03d54957da31d1545bf40e2d0b854aeb1c1e342e219303c6cc80R61-R66) [[3]](diffhunk://#diff-fdccfbf7c7ab03d54957da31d1545bf40e2d0b854aeb1c1e342e219303c6cc80R81)

### Dynamic PPAF Enablement:
* Updated `GlobalEndpointManager.cs` to dynamically enable or disable PPAF based on the `enablePartitionLevelFailover` flag retrieved from the database account properties. Added logic to reset the availability strategy to null if PPAF is disabled and no custom strategy is set. [[1]](diffhunk://#diff-f63d79326ed627000f22586f341f5d5d2bedfa149f33a1b5c88296b5ba2984e3R768-R774) [[2]](diffhunk://#diff-f63d79326ed627000f22586f341f5d5d2bedfa149f33a1b5c88296b5ba2984e3R812-R844)

### Default Hedging Thresholds:
* Changed the visibility of `DefaultHedgingThresholdInMilliseconds` and `DefaultHedgingThresholdStepInMilliseconds` in `DocumentClient.cs` from private to internal for broader accessibility within the SDK.
* Updated the initialization logic in `InitializePartitionLevelFailoverWithDefaultHedging` to use the new `SDKDefaultCrossRegionHedgingStrategy`.

### Tests for PPAF Functionality:
* Added a new integration test `ReadItemAsync_WithPPAFDynamicOverride` in `CosmosItemIntegrationTests.cs` to validate dynamic PPAF enablement, hedging behavior, and fallback when PPAF is disabled. This includes fault injection and diagnostics validation.

### End to End Validation:

**CreateItemAsync: Strong Consistency Account with Direct Mode:**

<kbd>Screenshot 2025-08-07 162428</kbd>

**CreateItemAsync: Strong Consistency Account with Gateway Mode:**

<kbd>image</kbd>

**CreateItemAsync: Session Consistency Account with Direct Mode:**

<kbd>image</kbd>

**CreateItemAsync: Session Consistency Account with Gateway Mode:**

<kbd>image</kbd>

_[Note: The Orange graph indicates the number of requests processed in the North CentralUS region, where as the Blue graph indicates the number of requests processed in the Central US region.]_

## Type of change

## Closing issues

To automatically close an issue: closes #5304

### DIFF
--- a/Microsoft.Azure.Cosmos/src/UserAgentContainer.cs
+++ b/Microsoft.Azure.Cosmos/src/UserAgentContainer.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Azure.Cosmos
             }
 
             // Matches 'F' followed by one or more digits, then a pipe '|'
-            return this.regex.IsMatch(this.Suffix);
+            return regex.IsMatch(this.Suffix);
         }
 
         internal override string BaseUserAgent => this.cosmosBaseUserAgent ?? string.Empty;


Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -798,6 +806,39 @@ internal async Task<AccountProperties> GetDatabaseAccountAsync(bool forceRefresh
...
                               forceRefresh: forceRefresh);
 #nullable enable
+        }
+
+        private void SetPPAFOnRefresh(bool enablePartitionLevelFailover)
+        {
+            if (enablePartitionLevelFailover)
+            {
+                this.connectionPolicy.EnablePartitionLevelFailover = true;
+                this.connectionPolicy.EnablePartitionLevelCircuitBreaker = true;
```
kundadebdatta: What will happen in the event that the customer has explicitly set the PPCB to be disabled ? We should honor that right ?

kundadebdatta: NM - I see that the contract is when PPAF is enabled, then we agreed to enable PPCB as well by default. We should be good.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -798,6 +806,39 @@ internal async Task<AccountProperties> GetDatabaseAccountAsync(bool forceRefresh
                               cancellationToken: this.cancellationTokenSource.Token,
                               forceRefresh: forceRefresh);
 #nullable enable
+        }
+
+        private void SetPPAFOnRefresh(bool enablePartitionLevelFailover)
+        {
+            if (enablePartitionLevelFailover)
+            {
+                this.connectionPolicy.EnablePartitionLevelFailover = true;
```
kundadebdatta: Discussed Offline. We need to find out a way to reflect this dynamically into the following places:

- `GlobalPartitionEndpointManagerCore`.
- `ClientRetryPolicy`
- `UserAgentFeatureFlag` in `DocumentClient`

Right now, we pass the enablement flag as constructor args. This will be a bottleneck for the dynamic enablement.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -282,6 +282,16 @@ public override bool IsRequestEligibleForPartitionLevelCircuitBreaker(
                 || (!request.IsReadOnlyRequest && this.globalEndpointManager.CanSupportMultipleWriteLocations(request.ResourceType, request.OperationType)));
         }
 
+        public override void SetIsPPAFEnabled(bool isPPAFEnabled)
+        {
+            this.isPartitionLevelFailoverEnabled = isPPAFEnabled;
```
kundadebdatta: There could be a thread contention. How about using `Interlocked.CompareExchange()` ?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Cosmos.Routing
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Azure.Cosmos.Common;
-    using Microsoft.Azure.Cosmos.Core.Trace;
+    using Microsoft.Azure.Cosmos.Core.Trace;
```
kundadebdatta, line 17: Let's remove the unnecessary diffs.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/AvailabilityStrategy.cs
```diff
@@ -46,5 +46,26 @@ public static AvailabilityStrategy CrossRegionHedgingStrategy(
...
+        /// until the request is completed or regions are exausted
+        /// </summary>
+        /// <param name="threshold"> how long before SDK begins hedging</param>
+        /// <param name="thresholdStep">Period of time between first hedge and next hedging attempts</param>
+        /// <param name="enableMultiWriteRegionHedge">Whether hedging for write requests on accounts with multi-region writes are enabled
+        /// Note that this does come with the caveat that there will be more 409 / 412 errors thrown by the SDK.
+        /// This is expected and applications that adopt this feature should be prepared to handle these exceptions.
+        /// Application might not be able to be deterministic on Create vs Replace in the case of Upsert Operations</param>
+        /// <returns>the cross region hedging availability</returns>
+        internal static AvailabilityStrategy SDKDefaultCrossRegionHedgingStrategy(
```
kundadebdatta: Let's name it as `SDKDefaultCrossRegionHedgingStrategyForPPAF()` or something like that ?

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1064,15 +1064,12 @@ private async Task<bool> GetInitializationTaskAsync(IStoreClientFactory storeCli
             this.ConnectionPolicy.UserAgentContainer.AppendFeatures(this.GetUserAgentFeatures());
             this.InitializePartitionLevelFailoverWithDefaultHedging();
 
-            this.PartitionKeyRangeLocation = 
-                this.ConnectionPolicy.EnablePartitionLevelFailover 
-                || this.ConnectionPolicy.EnablePartitionLevelCircuitBreaker
-                    ? new GlobalPartitionEndpointManagerCore(
+            this.PartitionKeyRangeLocation =
+                new GlobalPartitionEndpointManagerCore(
```
kundadebdatta: If you are planning to initialize the partition endpoint manager as `GlobalPartitionEndpointManagerCore` by default, then let's please make sure, when PPAF or PPCB is disabled, the following methods always returns `false` just like the `GlobalPartitionEndpointManagerNoOp`:

```
       public override bool TryAddPartitionLevelLocationOverride(
           DocumentServiceRequest request)
       {
           return false;
       }

       public override bool TryMarkEndpointUnavailableForPartitionKeyRange(
           DocumentServiceRequest request)
       {
           return false;
       }

       public override bool IsRequestEligibleForPartitionLevelCircuitBreaker(DocumentServiceRequest request)
       {
           return false;
       }

       public override bool IsRequestEligibleForPerPartitionAutomaticFailover(DocumentServiceRequest request)
       {
           return false;
       }

       public override bool IncrementRequestFailureCounterAndCheckIfPartitionCanFailover(
           DocumentServiceRequest request)
       {
           return false;
       }
```

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManager.cs
```diff
@@ -53,5 +53,27 @@ public abstract bool IsRequestEligibleForPartitionLevelCircuitBreaker(
...
             Func<Dictionary<PartitionKeyRange, Tuple<string, Uri, TransportAddressHealthState.HealthStatus>>, Task> backgroundConnectionInitTask);
+
+        /// <summary>
+        /// Sets whether per-partition automatic failover is enabled.
+        /// </summary>
+        /// <param name="isEnabled"></param>
+        public abstract void SetIsPPAFEnabled(bool isEnabled);
+
+        /// <summary>
+        /// Sets whether per-partition circuit breaker is enabled.
```
kundadebdatta: Add more details around thread safety, as it's guarded by `Interlocked` for atomic operations.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
NaluTripician: I believe that all of these will check to see if PPAF is enabled and return false if it is not

NaluTripician: Update, had to make one change to ensure this, should be the same as the NoOp now.

Microsoft.Azure.Cosmos/src/UserAgentContainer.cs
```diff
@@ -34,9 +34,15 @@ public void AppendFeatures(
         {
             if (!string.IsNullOrEmpty(features))
             {
+                // Here we have 3 scenarios: 
+                // 1. Suffix is empty, we just set it to the features.
+                // 2. Suffix is not empty, we append the features to the existing suffix.
+                // 3. Suffix already contains features, we the new features in the existing suffix.
                 this.Suffix = string.IsNullOrEmpty(this.Suffix)
```
kundadebdatta, line 44: Toggling between disable and enable features generates the user agent like the below:

`"User Agent": "cosmos-netstandard-sdk/3.52.1|2|X64|Microsoft Windows 10.0.26100|.NET 6.0.36|L|F 00000010|F2|F3"`

We will need to update the features instead of appending.

Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs
```diff
@@ -85,7 +75,17 @@ internal sealed class GlobalPartitionEndpointManagerCore : GlobalPartitionEndpoi
         /// <summary>
         /// A boolean flag indicating if the background connection initialization recursive task is active.
         /// </summary>
-        private bool isBackgroundConnectionInitActive = false;
+        private bool isBackgroundConnectionInitActive = false;
+
+        /// <summary>
+        /// A boolean (represented as an int to allow for thread-safety) flag used to determine if partition level failover is enabled.
+        /// </summary>
+        private int isPartitionLevelFailoverEnabled;
```
kundadebdatta: Let's rename `isPartitionLevelFailoverEnabled` to `isPartitionLevelAutomaticFailoverEnabled`.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -44,6 +44,12 @@ internal class CrossRegionHedgingAvailabilityStrategy : AvailabilityStrategyInte
         /// </summary>
         public bool EnableMultiWriteRegionHedge { get; private set; }
 
+        /// <summary>
+        /// Intenal flag to indicate if this is the default strategy used by the SDK when ebabling 
+        /// PPAF for clients without customer defined availability strategy.
+        /// </summary>
+        public bool IsSDKDefaultStrategy { get; private set; }
```
kundadebdatta: Let's rename this to `IsSDKDefaultStrategyForPPAF`

Microsoft.Azure.Cosmos/src/UserAgentContainer.cs
```diff
@@ -34,9 +34,27 @@ public void AppendFeatures(
...
                     ? features
-                    : $"{features}|{this.Suffix}";
+                    : this.Suffix.Contains("|")
+                        ? $"{features}{this.Suffix.Substring(this.Suffix.IndexOf('|'))}"
+                        : $"{features}|{this.Suffix}";
```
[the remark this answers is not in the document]
kundadebdatta: This is addressed.

ananth7592 (dismissed):

Microsoft.Azure.Cosmos/src/UserAgentContainer.cs
```diff
@@ -34,9 +35,27 @@ public void AppendFeatures(
         {
             if (!string.IsNullOrEmpty(features))
             {
+                // Here we have 3 scenarios:
```
FabianMeiswinkel, line 40: How do you prevent issues when UserAgentSuffix provided by custmer already contains "|"?

FabianMeiswinkel (dismissed):
LGTM except for one small question

Microsoft.Azure.Cosmos/src/UserAgentContainer.cs
NaluTripician: I will add something to address this

NaluTripician: Change here should cover most scenarios unless the customer's application name is in the form of `F{digits}|{rest of string}`

Microsoft.Azure.Cosmos/src/UserAgentContainer.cs
```diff
@@ -53,12 +53,24 @@ public void AppendFeatures(
...
 
+        private bool HasFeatureFlag()
+        {
+            if (string.IsNullOrEmpty(this.Suffix))
+            {
+                return false;
+            }
+
+            // Matches 'F' followed by one or more digits, then a pipe '|'
+            return Regex.IsMatch(this.Suffix, @"F\d+\|");
```
FabianMeiswinkel: Please use a compiled Regex - otherwise this has perf overhead with every execution

FabianMeiswinkel: Since you use the static Regex.IsMatch method it is possible that the regex gets cached - but the default cache has a limit of 16 to be cached regex expressions - in a library like our SDK we should not gamble - and explicitly craete a static read-only variable that has the compiled regex instance - and call its IsMatch method instead - see https://learn.microsoft.com/en-us/dotnet/standard/base-types/compilation-and-reuse-in-regular-expressions

FabianMeiswinkel (changes requested):
LGTM except for the regex compilation comment

Microsoft.Azure.Cosmos/src/UserAgentContainer.cs
```diff
@@ -13,8 +13,10 @@ internal class UserAgentContainer : Documents.UserAgentContainer
         private const int MaxOperatingSystemString = 30;
         private const int MaxClientId = 10;
         private const string PipeDelimiter = "|";
+
         private readonly string cosmosBaseUserAgent;
         private readonly string clientId;
+        private readonly Regex regex = new Regex(@"F\d+\|", RegexOptions.Compiled);
```
FabianMeiswinkel: private static readonly

ananth7592 (approved):

PR #5358, opened 2025-08-18
Switch from ConcurrentDictionary to an approach lazily materializing the Dictionary only while read-only

FabianMeiswinkel:


FabianMeiswinkel:
Closing this - opened https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5361 instead

PR #5361, opened 2025-08-19
Trace: Fixes thread safety issue in Trace class causing high CPU usage and InvalidOperationException

FabianMeiswinkel:
# Pull Request Template

## Description

When cross-region hedging is enabled in the Azure Cosmos DB .NET SDK, users experience high CPU usage or InvalidOperationExceptions ue to thread safety issues when accessing dictionaries in the Trace class. The issue occurs in the AddOrUpdateDatum method where multiple parallel threads attempt to update the same dictionary without proper synchronization.

### Design considerations
The approach chosen is motivated by:
- the need to maintain the order of TraceDatum nodes - the laternative of switching to a ConcurrentDictionary instead off keeping Dictonary and making it thread-safe would result in reordering of trace datum nodes
- non-functional optimization for the normal request flow where materialization o fthe diagnostics only strats after request processing completes (and no more new trace datum nodes/children) are added to a trace. There are a few edge cases like cross regional retries and timeouts where trace datum nodes could still be added after starting to materialize the trace. The current design marks the begin of materialization and would afterwards create new snapshots for every mutation. This allows making sure that Trace.Data or Trace.Children oeprate on a snapshot that is immutable (thread-safe) - but between multiple calls to ITrace.Data/Children different isntances could be returned - and they might have different content.

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Tracing/NoOpTrace.cs
```diff
@@ -40,6 +40,8 @@ private NoOpTrace()
 
         public IReadOnlyDictionary<string, object> Data => NoOpData;
 
+        public bool IsBeingWalked => true;
```
Pilchie: Why true here?

Pilchie (approved):
This approach does seem a bit error prone:
* Needing to remember to call ` SetWalkingStateRecursively`
* Needing to check `IsBeingWalked` and behave accordingly.

I trust that we need this complexity, but I don't love it.

Microsoft.Azure.Cosmos/src/Tracing/NoOpTrace.cs
FabianMeiswinkel: Because otherwise you couldn't materialize ((ITrace)NoOpTrace).Data - or .Children

FabianMeiswinkel: Never mind - that is wrong. I think you are right - should return false instead.

FabianMeiswinkel: After debuggin through it - tehre are scenarios where NoOpTrace gets materialized - and since we validate that IsBeingWalked00true on materialization - it actually needs to return true. The negative side effects are irrelevant because NoOpTrace is always immutable.

Microsoft.Azure.Cosmos/src/Tracing/Trace.cs
```diff
@@ -7,17 +7,15 @@ namespace Microsoft.Azure.Cosmos.Tracing
...
-    using Microsoft.Azure.Cosmos.Tracing.TraceData;
     using Microsoft.Azure.Documents;
 
     internal sealed class Trace : ITrace
     {
         private static readonly IReadOnlyDictionary<string, object> EmptyDictionary = new Dictionary<string, object>();
-        private readonly List<ITrace> children;
-        private readonly Lazy<Dictionary<string, object>> data;
+        private volatile List<ITrace> children;
+        private volatile Dictionary<string, object> data;
```
kirankumarkolli, line 16: In a request flow there are 2 or three places where the Data is captures in all other-places its not used.
Making it non-lazy result in more Dict created

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/Tracing/Trace.cs
FabianMeiswinkel: Discussed offline - the dictionary is lazily created - just not using Lazy<T>

aavasthy (approved):

PR #5252, opened 2025-06-19
TokenCredentialCache: Adds an options to override AAD audience scope.

aavasthy:
# Pull Request Template

## Description

Add AAD audience scope override functionality when this environment variable is set AADScopeOverride = "AZURE_COSMOS_AAD_SCOPE_OVERRIDE".  If no environment variable is et then it uses the existing default scope value.
This will help clients in accessing applications which have a different scope.

## Type of change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)

## Closing issues

To automatically close an issue: closes #IssueNumber

PR #5364, opened 2025-08-19
TokenCredentialCache: Adds a fallback mechanism to AAD scope override.

aavasthy:
# Pull Request Template

## Description

It's a follow up for this PR https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5252, which introduces the ability to override AAD scope value. As part of the current PR, use overridden scope if its provided and there is no fallback for this.
If no overridden scope is provided then use account scope and if this fails with exception [AADSTS500011](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes) then fallback to this scope "https://cosmos.azure.com/.default". Each scope continues to go through existing validations.
## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -190,34 +211,27 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
...
-                                double refreshIntervalInSeconds = (this.cachedAccessToken.Value.ExpiresOn - DateTimeOffset.UtcNow).TotalSeconds * DefaultBackgroundTokenCredentialRefreshIntervalPercentage;
-
-                                // Ensure the background refresh interval is a valid range.
-                                refreshIntervalInSeconds = Math.Max(refreshIntervalInSeconds, TokenCredentialCache.MinimumTimeBetweenBackgroundRefreshInterval.TotalSeconds);
-                                refreshIntervalInSeconds = Math.Min(refreshIntervalInSeconds, TokenCredentialCache.MaxBackgroundRefreshInterval.TotalSeconds);
-                                this.systemBackgroundTokenCredentialRefreshInterval = TimeSpan.FromSeconds(refreshIntervalInSeconds);
-                            }
-
-                            return this.cachedAccessToken.Value;
+                    {
```
FabianMeiswinkel, line 187: Please refactor the code to encapsulate the core code to call
`await this.tokenCredential.GetTokenAsync(`
including the common exception handling (most of it is identical independent of the scope)
from a shared method

FabianMeiswinkel: The RefreshCachedTokenWithRetryHelperAsync is becoming unmaintainable otherwise

kirankumarkolli: One alternative is to model through composition pattern.
That way the TokenCache can be limited to a paramaterized scope with minimal changes.

kirankumarkolli:
```C#
interface ICosmosTokenCredential : IDisposable
{
	Task<string> GetToken();
}

class TokenCredentialCache : ICosmosTokenCredential
{

	TokenCredentialCache GetInstance(string accountendpoint) {
		if (ENV set)
			return new TokenCredentialCache(ENV-value);
			
		return new TokenCredentialCacheMultiScope();
	}
}


class TokenCredentialCacheMultiScope : ICosmosTokenCredential
{
	private string authScope; 
	private ICosmosTokenCredential _instance;

	Task<string> GetToken() {

		try {
			await _instance.GetToken()
		} catch(Excetion ex) where ex.ToString().Contains("AAD....") {
			if (authScope != fallbackScope) {
				_instance = new TokenCredentialCache(fallbackScope);
				authScope = fallbackScope
				return _instance.GetToken()
			}
			
			throw;
		}
	}
}


```

aavasthy: created a separate class for resolving the scope value and fallback.

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+                return false;
+            }
+
+            // If already attempted fallback, do not fallback again
+            if (this.fallbackAttempted)
+            {
+                return false;
+            }
+
+            if (ex.InnerException?.Message.Contains(AadInvalidScopeErrorMessage) == true)
```
kirankumarkolli: ex.ToString() or ex.Message not capturing the AADInvalidaScpoeErrorMessage?

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+        public bool TryFallback(Exception ex)
+        {
+            // If override scope is set, never fallback
+            if (!string.IsNullOrEmpty(this.overrideScope))
+            {
+                return false;
+            }
+
+            // If already attempted fallback, do not fallback again
+            if (this.fallbackAttempted)
```
kirankumarkolli: Nit: This can be (this.currentScope == AadDefaultScope) and fallbackAttempted is not necessary

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -226,25 +220,31 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
...
                             // Don't retry on auth failures
                             if (requestFailedException.Status == (int)HttpStatusCode.Unauthorized ||
                                 requestFailedException.Status == (int)HttpStatusCode.Forbidden)
                             {
                                 this.cachedAccessToken = default;
                                 throw;
+                            }
+
+                            // Fallback logic
+                            if (this.scopeProvider.TryFallback(requestFailedException))
```
kirankumarkolli: Isn't it safe to add this to finally block?

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -226,25 +220,31 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
...
+                            {
+                                continue;
```
kirankumarkolli: Return value has no functional difference right?

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
aavasthy: No only the inner exception has AADSTS500011 error code

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
aavasthy: No i believe its better to keep fallback logic to the catch block.
We need to call continue, right after a fallback so the loop retries immediately with the new scope.
The finally block shouldn't control the loop/retry flow.

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
aavasthy: The return value is functionally important. It shows whether a fallback was actually performed, so the retry loop can immediately retry with the new scope (continue).

ananth7592 (dismissed):

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+        public bool TryFallback(Exception ex)
+        {
+            // If override scope is set, never fallback
+            if (!string.IsNullOrEmpty(this.overrideScope))
+            {
+                return false;
+            }
+
+            // If already using fallback scope, do not fallback again
+            if (this.currentScope == AadDefaultScope)
```
kirankumarkolli: nit: Const use ClassName prefix

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -226,25 +220,31 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
                                 $"RequestFailedException at {DateTime.UtcNow.ToString(CultureInfo.InvariantCulture)}",
                                 requestFailedException.Message);
 
-                            DefaultTrace.TraceError($"TokenCredential.GetToken() failed with RequestFailedException. scope = {string.Join(";", this.tokenRequestContext.Scopes)}, retry = {retry}, Exception = {lastException.Message}");
+                            DefaultTrace.TraceError($"TokenCredential.GetToken() failed with RequestFailedException. scope = {string.Join(";", this.scopeProvider.GetTokenRequestContext().Scopes)}, retry = {retry}, Exception = {lastException.Message}");
```
kirankumarkolli: nit: Its safe to use `tokenRequestContext` instead, it guarantees irrespective of implementation.

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -220,31 +216,37 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
...
                         catch (RequestFailedException requestFailedException)
-                        {
+                        {
                             lastException = requestFailedException;
                             getTokenTrace.AddDatum(
                                 $"RequestFailedException at {DateTime.UtcNow.ToString(CultureInfo.InvariantCulture)}",
                                 requestFailedException.Message);
 
-                            DefaultTrace.TraceError($"TokenCredential.GetToken() failed with RequestFailedException. scope = {string.Join(";", this.tokenRequestContext.Scopes)}, retry = {retry}, Exception = {lastException.Message}");
+                            DefaultTrace.TraceError($"TokenCredential.GetToken() failed with RequestFailedException. scope = {string.Join(";", tokenRequestContext.Scopes ?? Array.Empty<string>())}, retry = {retry}, Exception = {lastException.Message}");
```
kirankumarkolli: Are scoped expected to be empty/null?

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -65,14 +64,7 @@ internal TokenCredentialCache(
...
 
-            string? scopeOverride = ConfigurationManager.AADScopeOverrideValue(defaultValue: null);
-
-            this.tokenRequestContext = new TokenRequestContext(new string[]
-            {
-                !string.IsNullOrEmpty(scopeOverride)
-                    ? scopeOverride
-                    : string.Format(TokenCredentialCache.ScopeFormat, accountEndpoint.Host)
-            });
+            this.scopeProvider = new Microsoft.Azure.Cosmos.Authorization.CosmosScopeProvider(accountEndpoint);
```
kirankumarkolli, line 67: Is scope Dispose needed?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAadTests.cs
```diff
@@ -263,6 +263,136 @@ void GetAadTokenCallBack(
...
+            finally
+            {
+                Environment.SetEnvironmentVariable("AZURE_COSMOS_AAD_SCOPE_OVERRIDE", previous);
+                using CosmosClient cleanup = TestCommon.CreateCosmosClient();
+                await cleanup.GetDatabase(databaseId).DeleteAsync();
+            }
+        }
+
+        [TestMethod]
+        public async Task Aad_AccountScope_Fallbacks_ToCosmosScope()
```
kirankumarkolli, line 338: Test coverage to use the successful scope.

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+            this.accountScope = string.Format(ScopeFormat, accountEndpoint.Host);
+            this.currentScope = this.overrideScope ?? this.accountScope;
+        }
+
+        public TokenRequestContext GetTokenRequestContext()
+        {
+            return new TokenRequestContext(new[] { this.currentScope });
+        }
+
+        public bool TryFallback(Exception exception)
```
kirankumarkolli, line 31: Unit tests please

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAadTests.cs
aavasthy: added a test for success case

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
aavasthy: updated and disposed scopeProvider

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
aavasthy: removed the empty array string

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
aavasthy: added unit tests

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Utils/LocalEmulatorTokenCredential.cs
```diff
@@ -44,7 +59,10 @@ public override ValueTask<AccessToken> GetTokenAsync(TokenRequestContext request
 
         private AccessToken GetAccessToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
         {
-            Assert.AreEqual(this.expectedScope, requestContext.Scopes.First());
+            if (this.expectedScope != null)
```
kirankumarkolli: Lets also ensure that its part of expectedScopes

aavasthy: updated

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -171,11 +164,13 @@ private async Task<AccessToken> GetNewTokenAsync(
...
         private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
             ITrace trace)
-        {
+        {
             try
             {
                 Exception? lastException = null;
-                const int totalRetryCount = 2;
+                const int totalRetryCount = 2;
+                TokenRequestContext tokenRequestContext = default;
```
kirankumarkolli: nit: Can be just before try block

Microsoft.Azure.Cosmos/src/Authorization/IScopeProvider.cs
```diff
@@ -0,0 +1,16 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.Authorization
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+    using global::Azure.Core;
+
+    internal interface IScopeProvider : IDisposable
```
FabianMeiswinkel: Why do we need IDisposable here?

FabianMeiswinkel: There isn't anything being cleaned-up in teh only implementation - since this is internal API we can add this later if needed?

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -220,31 +217,37 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
...
                                 requestFailedException.Status == (int)HttpStatusCode.Forbidden)
                             {
                                 this.cachedAccessToken = default;
                                 throw;
+                            }
+
+                            // Fallback logic
+                            if (this.scopeProvider.TryFallback(requestFailedException))
+                            {
+                                continue;
```
FabianMeiswinkel: continue here effectively is a no-op. Should we do some logging instead?

aavasthy: Added a log trace here

Microsoft.Azure.Cosmos/src/Authorization/IScopeProvider.cs
aavasthy: removed the IDisposable

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -236,6 +235,8 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
                             // Fallback logic
                             if (this.scopeProvider.TryFallback(requestFailedException))
                             {
+                                DefaultTrace.TraceWarning(
+                                    $"TokenCredentialCache: Fallback to default scope triggered due to exception: {requestFailedException.Message}");
```
FabianMeiswinkel: Sorry - I should have been more specific. I think the exception is important to be logged here all the itme - when we actually do the fallback (TryFallback) it is less important (probably info/verbose level sufficient). But when we see errors after TryFallback returns false we should for sure log these exceptions as warning?

FabianMeiswinkel: Also knowing where in the process we are (did TryFallback return true/false) would be helpful?

aavasthy: updated the logs and also logging TryFallback response true/false

Microsoft.Azure.Cosmos/src/Authorization/IScopeProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.Authorization
+{
+    using System;
+    using System.Collections.Generic;
```
FabianMeiswinkel: NIT - unused imports

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -219,32 +215,45 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
...
+                                throw;
+                            }
+
+                            bool didFallback = this.scopeProvider.TryFallback(requestFailedException);
+                            string logMessage = $"TokenCredential.GetToken() failed. scope = {string.Join(";", tokenRequestContext.Scopes)}, retry = {retry}, Exception = {requestFailedException.Message}. Fallback attempted: {didFallback}";
+
+                            if (didFallback)
+                            {
+                                DefaultTrace.TraceInformation(logMessage);
+                                continue;
```
FabianMeiswinkel:
```suggestion
```

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -255,15 +264,25 @@ private async ValueTask<AccessToken> RefreshCachedTokenWithRetryHelperAsync(
...
-                            DefaultTrace.TraceError(
-                                $"TokenCredential.GetTokenAsync() failed. scope = {string.Join(";", this.tokenRequestContext.Scopes)}, retry = {retry}, Exception = {lastException.Message}");
+                        catch (Exception exception)
+                        {
+                            lastException = exception;
+                            getTokenTrace.AddDatum(
+                                $"Exception at {DateTime.UtcNow.ToString(CultureInfo.InvariantCulture)}",
+                                exception.Message);
+
+                            bool didFallback = this.scopeProvider.TryFallback(exception);
```
FabianMeiswinkel: Why try the fallback here in the generic Exception case? This could easily regress non-public clouds because they get forced into cosmos.azure.com on some random transient error - and would never recover?

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                this.currentScope = CosmosScopeProvider.AadDefaultScope;
+                return true;
+            }
+#pragma warning restore CDX1003 // DontUseExceptionToString
+
+            return false;
+        }
+
+        public void Dispose()
```
kirankumarkolli: nit: Remove

Microsoft.Azure.Cosmos/src/Authorization/IScopeProvider.cs
aavasthy: removed

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
aavasthy: removed and tested the change. It retries irrespective of whether continue is there

Microsoft.Azure.Cosmos/src/Authorization/CosmosScopeProvider.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
aavasthy: This exception [AADSTS500011](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes) falls in the generic exception catch block and adding the fallback logic there. The fallback logic internally will check for the specific AAD error code and only then retry.

kirankumarkolli (approved):

PR #5388, opened 2025-09-09
Disable EventSource calls to avoid IL2026 trimming errors

anuchandy:
# [Internal] Telemetry : Disable EventSource calls to avoid IL2026 trimming errors

## Description

Comment out EventSource calls to resolve IL2026 trimming errors

- Comment out WriteEventCore calls in DocumentClientEventSource to fix
  IL2026 trim analysis errors when building trimmed applications
- EventSource.WriteEventCore has RequiresUnreferencedCodeAttribute which
  breaks functionality during trimming as it serializes the whole object graph
- This issue appears in trimmed builds but not AOT builds due to different
  optimization levels
- Safe to disable for preview trimmed versions that don't rely on EventSource telemetry
- 
## Description

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
@@ -41,7 +41,14 @@ namespace Microsoft.Azure.Cosmos
             // must be explicitly be set before writing the event.
             CustomTypeExtensions.SetActivityId(ref activityId);
 
-            this.WriteEventCore(eventId, eventDataCount, dataDesc);
+            // COMMENTED OUT FOR TRIMMED AZMCP: The following call to WriteEventCore is commented out to resolve
+            // IL2026 trim analysis errors when building the trimmed version of azmcp. The error occurs because
+            // EventSource.WriteEventCore has RequiresUnreferencedCodeAttribute which can break functionality 
+            // when trimming application code, as EventSource serializes the whole object graph and the trimmer
+            // cannot safely handle this case. This is specific to the trimmed build - the AOT build doesn't 
+            // report this issue due to more aggressive optimizations that trim away unreachable code paths.
+            // Since the preview trimmed version doesn't rely on EventSource telemetry, this is safe to disable.
+            // this.WriteEventCore(eventId, eventDataCount, dataDesc);
         }
 
         [Event(1,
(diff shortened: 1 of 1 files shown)


joshfree (approved):

Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
```diff
@@ -41,7 +41,14 @@ private unsafe void WriteEventCoreWithActivityId(Guid activityId, int eventId, i
             // must be explicitly be set before writing the event.
             CustomTypeExtensions.SetActivityId(ref activityId);
 
-            this.WriteEventCore(eventId, eventDataCount, dataDesc);
```
agocke, line 44: This works, but I think what you should do is add RequiresUnreferencedCode to this method, then look at the callers. That is, WriteEventCore is annotated with

```
 [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("EventSource will serialize the whole object graph. Trimmer will not safely handle this case because properties may be trimmed. This can be suppressed if the object is a primitive type")]
```

I think the same thing applies here. The responsibility is on the callers to ensure that they're either not calling this in any trim-compatible code path, or they are correctly suppressing the warning because they're only serializing primitive types.

Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
```diff
@@ -266,7 +273,8 @@ private unsafe void Request(
                 dataDesc[32].DataPointer = (IntPtr)(fixedXDate);
                 dataDesc[32].Size = (xDate.Length + 1) * UnicodeEncodingCharSize;
 
-                this.WriteEventCoreWithActivityId(activityId, 1, eventDataCount, dataDesc);
```
agocke, line 269: All of these inputs are primitive types, so this should be OK to suppress with `UnconditionalSuppressMessage` on the parent method.

Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
```diff
@@ -513,7 +521,8 @@ private unsafe void Response(Guid activityId,
                 dataDesc[29].DataPointer = (IntPtr)(fixedVersion);
                 dataDesc[29].Size = (version.Length + 1) * UnicodeEncodingCharSize;
 
-                this.WriteEventCoreWithActivityId(activityId, 2, eventDataCount, dataDesc);
```
agocke, line 516: Same here, it looks like these are all primitive types (Guid, short, double, string).

agocke:
Ah, I see the problem -- this project only builds for netstandard2.0. I will try to add a net8.0 target.

anuchandy:
closing this based on the offline discussion with Andy, he is working on a pr to address this by adding annotations

PR #5402, opened 2025-09-12
[Client encryption] Make Encryption.Custom build always preview

MartinSarkany:
# Pull Request Template

## Description

Removes the preview flag that differentiated between preview and stable builds. There is no stable version being maintained—all work focuses on preview.
Keeping the flag would require adding unnecessary #if blocks when updating the dependency and using its new features, even though the stable version is never built and does not support these features.

#4678


kirankumarkolli (approved):

kirankumarkolli:
Approved after offline syncup with Juraj.

PR #5409, opened 2025-09-17
Query: Fixes to not use passthrough context for collections with HPK

neildsh:
## Description

Avoids using passthrough context for collections with hierarchical partition key.

## Type of change


Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
```diff
@@ -385,9 +385,10 @@ private bool AssertPartitioningPropertiesAndHeaders()
 
             bool partitionKeyRangeIdExists = !string.IsNullOrEmpty(this.Headers.PartitionKeyRangeId);
             if (partitionKeyRangeIdExists)
-            {
+            {
+                OperationType op = this.OperationType;
```
adityasa: >op [](http://example.com/codeflow?start=30&length=2)

Nit, rename to operation?

adityasa: Alternatively, since this is done for brevity/readability, you can separate each condition expression by a newline in original if condition.

adityasa (dismissed):
:shipit:

ananth7592 (dismissed):

Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
neildsh: done

sc978345 (approved):

PR #5412, opened 2025-09-18
Hedging: Adds back diagnostics filed `Response Region` to hedging request diagnostics 

### LINKED ISSUE #5408: Re-add response region to hedging diagnostics
We plan to re-add the response region field in the diagnostics for hedged requests for a general quick way to view this information. 

We want to make it clear that there might be cases the response region may not indicate the correct response region at that point due to some edge case scenarios that can come up depending on what the hedging configuration is. (This is one of the reasons why we decided to remove it originally). The only true way to know where a response came from will be to parse the diagnostic string to see where the storeresponse (or httpresponse if in gatewaymode). We cannot add a 100% accurate standalone parameter in the diagnostics as this will have a negative effect on the response time and CPU overhead which would negate the benefits of hedging. 

The aim of adding this parameter back is to help customers get a quick understanding of where the hedging response came from.

NaluTripician:
# Pull Request Template

## Description

We plan to re-add the response region field in the diagnostics for hedged requests for a general quick way to view this information. This change will affect the diagnostics for all requests with hedging, and PPAF enabled. This will not affect the behavior of those requests, only add one field to the diagnostics. 

We want to make it clear that there might be cases the response region may not indicate the correct response region at that point due to some edge case scenarios that can come up depending on what the hedging configuration is. (This is one of the reasons why we decided to remove it originally). The only true way to know where a response came from will be to parse the diagnostic string to see where the storeresponse (or httpresponse if in gatewaymode). We cannot add a 100% accurate standalone parameter in the diagnostics as this will have a negative effect on the response time and CPU overhead which would negate the benefits of hedging.

The aim of adding this parameter back is to help customers get a quick understanding of where the hedging response came from.

**Diagnostics improvements:**

* Added a new diagnostic datum, `Response Region`, to record the actual region that serviced the request in the diagnostics trace for each hedged response. This helps differentiate between the intended target region and the region that ultimately processed the request. [[1]](diffhunk://#diff-fdccfbf7c7ab03d54957da31d1545bf40e2d0b854aeb1c1e342e219303c6cc80R28) [[2]](diffhunk://#diff-fdccfbf7c7ab03d54957da31d1545bf40e2d0b854aeb1c1e342e219303c6cc80R210-R213) [[3]](diffhunk://#diff-fdccfbf7c7ab03d54957da31d1545bf40e2d0b854aeb1c1e342e219303c6cc80R242-R245)

**Response tracking enhancements:**

* Updated the `CloneAndSendAsync` and `RequestSenderAndResultCheckAsync` methods to pass and record the target region name for each hedged request, ensuring accurate tracking of which region was intended to handle each request. [[1]](diffhunk://#diff-fdccfbf7c7ab03d54957da31d1545bf40e2d0b854aeb1c1e342e219303c6cc80R290) [[2]](diffhunk://#diff-fdccfbf7c7ab03d54957da31d1545bf40e2d0b854aeb1c1e342e219303c6cc80R300)
* Modified the `HedgingResponse` class to include a new `TargetRegionName` field, allowing each response to carry information about its intended region.
* Updated all usages of `HedgingResponse` to properly set and propagate the `TargetRegionName` for both successful and unsuccessful responses.

## Type of change

## Closing issues

To automatically close an issue: closes #5408

### DIFF
--- a/Microsoft.Azure.Cosmos/src/AssemblyInfo.cs
+++ b/Microsoft.Azure.Cosmos/src/AssemblyInfo.cs
@@ -37,6 +37,8 @@ using System.Runtime.CompilerServices;
 [assembly: InternalsVisibleTo("Microsoft.Azure.Cosmos.QueryOracle" + AssemblyKeys.TestPublicKey)]
 [assembly: InternalsVisibleTo("Microsoft.Azure.Cosmos.QueryOracle.JobProcessor" + AssemblyKeys.ProductPublicKey)]
 [assembly: InternalsVisibleTo("Microsoft.Azure.Cosmos.QueryOracle.JobProcessor" + AssemblyKeys.TestPublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.Azure.Cosmos.QueryOracle.QueryTests" + AssemblyKeys.ProductPublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.Azure.Cosmos.QueryOracle.QueryTests" + AssemblyKeys.TestPublicKey)]
 [assembly: InternalsVisibleTo("Microsoft.Azure.Cosmos.QueryOracle.UnitTests" + AssemblyKeys.ProductPublicKey)]
 [assembly: InternalsVisibleTo("Microsoft.Azure.Cosmos.QueryOracle.UnitTests" + AssemblyKeys.TestPublicKey)]
 [assembly: InternalsVisibleTo("CosmosCTL" + AssemblyKeys.ProductPublicKey)]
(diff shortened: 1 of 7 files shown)


Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -206,6 +207,10 @@ internal override async Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
                                     ((CosmosTraceDiagnostics)hedgeResponse.ResponseMessage.Diagnostics).Value.AddOrUpdateDatum(
                                         HedgeContext,
                                         hedgeRegions.Take(requestNumber + 1));
+                                    // Note that the target region can be seperate than the actual region that serviced the request depending on the scenario
+                                    ((CosmosTraceDiagnostics)hedgeResponse.ResponseMessage.Diagnostics).Value.AddOrUpdateDatum(
```
kirankumarkolli, line 211: Add UT coverage to guard against future regressions

FabianMeiswinkel (approved):
LGTM

PR #5422, opened 2025-09-24
Fixes race condition that can cause InvalidOperationException in CosmosOperationCancelledException.ToString()

FabianMeiswinkel:
# Pull Request Template

## Description

Fixes race condition that can cause InvalidOperationException in CosmosOperationCancelledException.ToString() - especially when cross-regional hedging is enabled and E2E timeouts happen (CancellationToken being cancelled).

## Type of change

## Closing issues
ICM#66334029

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs
+++ b/Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs
@@ -20,10 +20,12 @@ namespace Microsoft.Azure.Cosmos
     [Serializable]
     public class CosmosOperationCanceledException : OperationCanceledException, ICloneable
     {
+        private readonly Object thisLock = new object();
         private readonly OperationCanceledException originalException;
-        private readonly Lazy<string> lazyMessage;
-        private readonly Lazy<string> toStringMessage;
         private readonly bool tokenCancellationRequested;
+
+        private string lazyMessage;
+        private string toStringMessage;
 
         /// <summary>
         /// Create an instance of CosmosOperationCanceledException
@@ -38,8 +40,8 @@ namespace Microsoft.Azure.Cosmos
             this.originalException = originalException ?? throw new ArgumentNullException(nameof(originalException));
             this.Diagnostics = diagnostics ?? throw new ArgumentNullException(nameof(diagnostics));
             this.tokenCancellationRequested = originalException.CancellationToken.IsCancellationRequested;
-            this.toStringMessage = this.CreateToStringMessage();
-            this.lazyMessage = this.CreateLazyMessage();
+            this.toStringMessage = null;
+            this.lazyMessage = null;
         }
 
         internal CosmosOperationCanceledException(
@@ -61,8 +63,8 @@ namespace Microsoft.Azure.Cosmos
             }
             this.Diagnostics = new CosmosTraceDiagnostics(trace);
             this.tokenCancellationRequested = originalException.CancellationToken.IsCancellationRequested;
-            this.toStringMessage = this.CreateToStringMessage();
-            this.lazyMessage = this.CreateLazyMessage();
+            this.toStringMessage = null;
+            this.lazyMessage = null;
         }
 
         /// <summary>
@@ -75,8 +77,8 @@ namespace Microsoft.Azure.Cosmos
         {
             this.originalException = (OperationCanceledException)info.GetValue("originalException", typeof(OperationCanceledException));
             this.tokenCancellationRequested = (bool)info.GetValue("tokenCancellationRequested", typeof(bool));
-            this.lazyMessage = new Lazy<string>(() => (string)info.GetValue("lazyMessage", typeof(string)));
-            this.toStringMessage = new Lazy<string>(() => (string)info.GetValue("toStringMessage", typeof(string)));
+            this.lazyMessage = null;
+            this.toStringMessage = null;
             //Diagnostics cannot be serialized
             this.Diagnostics = new CosmosTraceDiagnostics(NoOpTrace.Singleton);
         }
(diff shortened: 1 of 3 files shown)


NaluTripician (approved):

aavasthy (approved):

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs
```diff
@@ -120,18 +122,42 @@ public override Exception GetBaseException()
...
+            lock (this.thisLock)
+            {
+                return this.lazyMessage ??=
+                     $"{this.originalException.Message}{Environment.NewLine}Cancellation Token has expired: {this.tokenCancellationRequested}. Learn more at: https://aka.ms/cosmosdb-tsg-request-timeout{Environment.NewLine}";
+            }
+        }
+
+        internal string EnsureToStringMessage(Boolean skipDiagnostics)
+        {
+            if (this.toStringMessage != null)
```
jeet1995, line 144: q: @FabianMeiswinkel - why is the lock not applied to purely the read paths? is it because once `lazyMessage` or `toStringMessage` is set it cannot be `null` again?

FabianMeiswinkel: Correct - never moves back to null

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs
```diff
@@ -19,11 +19,13 @@ namespace Microsoft.Azure.Cosmos
     /// </summary> 
     [Serializable]
     public class CosmosOperationCanceledException : OperationCanceledException, ICloneable
-    {
+    {
+        private readonly Object thisLock = new object();
```
Pilchie, line 23: Do we need our own lock, or can we just use one of the `Lazy<T>` constructors that is more thread-safe? "https://learn.microsoft.com/dotnet/api/system.lazy-1.-ctor"

Or is the issue when both ToString and Message are being created concurrrently, and need to be protected from each other?

FabianMeiswinkel: Lazy<T> was used before - but is not able to resolve the cyclic dependency between ToString of the CosmosOperationCancelledException and CosmosTraceDiagnostics - now the ValueFactory to materialize the toString value distinguishes between being called from CosmosTraceDiagnostics (skip adding the Diagnostics) vs. in ToString this won't be necessary. Since there are two factory Methods Lazy<T> doesn't work anymore. And thread-safety is sufficiently addressed - given that once value is non-null it is immutabel this is staright forward

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs
```diff
@@ -159,8 +185,8 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
             info.AddValue("originalException", this.originalException);
 #pragma warning restore CDX1000 // DontConvertExceptionToObject
             info.AddValue("tokenCancellationRequested", this.tokenCancellationRequested);
-            info.AddValue("lazyMessage", this.lazyMessage.Value);
-            info.AddValue("toStringMessage", this.toStringMessage.Value);
+            info.AddValue("lazyMessage", this.EnsureLazyMessage());
+            info.AddValue("toStringMessage", this.EnsureToStringMessage(false));
```
Pilchie:
```suggestion
            info.AddValue("toStringMessage", this.EnsureToStringMessage(skipDiagnostics: false));
```

Microsoft.Azure.Cosmos/src/Tracing/TraceWriter.TraceJsonWriter.cs
```diff
@@ -130,9 +130,13 @@ private static void WriteTraceDatum(IJsonWriter writer, object value)
             {
                 writer.WriteStringValue(stringValue);
             }
-            else
-            {
-                writer.WriteStringValue(value.ToString());
+            else if (value is CosmosOperationCanceledException cosmosTimeoutException)
+            {
+                writer.WriteStringValue(cosmosTimeoutException.EnsureToStringMessage(true));
```
Pilchie:
```suggestion
                writer.WriteStringValue(cosmosTimeoutException.EnsureToStringMessage(skipDiagnostics: true));
```

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs
```diff
@@ -120,18 +122,42 @@ public override Exception GetBaseException()
         /// <inheritdoc/>
         public override string ToString()
         {
-            return this.toStringMessage.Value;
+            return this.EnsureToStringMessage(false);
```
Pilchie:
```suggestion
            return this.EnsureToStringMessage(skipDiagnostics: false);
```

Pilchie (dismissed):
Nit on using namad parameters for the boolean so it's easier to see at the call-site what the bool is for.

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs
FabianMeiswinkel: Fixed

Microsoft.Azure.Cosmos/src/Tracing/TraceWriter.TraceJsonWriter.cs
FabianMeiswinkel: Fixed in latest iteration

Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs
FabianMeiswinkel: Fixed in latest iteration

Pilchie (approved):

PR #5373, opened 2025-08-26
[Client encryption]: Fixes failing contract tests on .NET 8

adamnova:
# Pull Request Template

#4678 

## Description

The contracts test is failing due to incorrectly used templates and the contracts being different between .net 6 and .net 8

PR #5423, opened 2025-09-24
[Client encryption]: Adds synchronous initialization to CosmosDataEncryptionKeyProvider

### LINKED ISSUE #5400: [Client encryption]: Initialize CosmosDataEncryptionKeyProvider without making a backend call
Currently, the CosmosDataEncryptionKeyProvider needs to be initialized by calling `InitializeAsync()` method, which tries to create the specified container if it does not exist and initializes its reference. It also validates proper partition key setting.
This is a problem for high-scale services when many instances start in parallel, because the request is a metadata request with limited throughput on the account level. Additionally, this makes the startup vulnerable to temporary DB outages.
Services usually provision resources independently ahead of time, with the proper configuration, so this all is not be necessary.

The CosmosDataEncryptionKeyProvider ultimately needs a reference to the container to work properly, and the services already have it available during initialization. Add `Initialize(Container)` overload method as an alternative initialization method that will not make any backend calls.

We considered deferring the initialization of DEK provider to a later stage, but that only mitigates the potential startup failure during DB outage. It still eats from the quota and causes throttling. The workaround would be also not elegant, implementation-wise. 

**Additional context**
We'll prepare proposal for this change as part of the broader optimizations effort.

MartinSarkany:
# Pull Request Template

## Description

Adds synchronous initialization to CosmosDataEncryptionKeyProvider.
Depends on #5418 due to TestEncryptionKeyStoreProvider being introduced there.

## Type of change

## Closing issues
closes #5400


juraj-blazek (approved):

juraj-blazek (commented):
We need to regenerate public API contracts, too

MartinSarkany:
> We need to regenerate public API contracts, too

I'll update it after #5373 that fixes the contracts is merged. It is broken at the moment.

adamnova:
> > We need to regenerate public API contracts, too
> 
> I'll update it after #5373 that fixes the contracts is merged. It is broken at the moment.

But it should work for .net 6 at least.

juraj-blazek (approved):

adamnova (approved):

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -166,6 +163,25 @@ public async Task InitializeAsync(
             this.container = containerResponse.Container;
         }
 
+        /// <summary>
+        /// Initialize using an existing Cosmos DB container for storing wrapped DEKs.
+        /// </summary>
+        /// <param name="container">Existing Cosmos DB container.</param>
+        public void Initialize(Container container)
```
kirankumarkolli, line 181: There are assumptions about the partition-key which were validated in initialization
Its even high impactful post initialization if CosmosDB is unavailable right?

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -166,6 +163,25 @@ public async Task InitializeAsync(
             this.container = containerResponse.Container;
         }
 
+        /// <summary>
+        /// Initialize using an existing Cosmos DB container for storing wrapped DEKs.
```
kirankumarkolli: Please include the container requirements (ex: partitionKeyDefinition)
And also call out right usage pattern

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -166,6 +163,25 @@ public async Task InitializeAsync(
...
+        /// </summary>
+        /// <param name="container">Existing Cosmos DB container.</param>
+        public void Initialize(Container container)
+        {
+            this.ThrowIfAlreadyInitialized();
+
+            this.container = container ?? throw new ArgumentNullException(nameof(container));
```
kirankumarkolli: Thoughts on handling non-existing containers on Fetch* API's? (i.e. on failure attempt re-create it)

kirankumarkolli: Concurrency control is also an important aspect

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
MartinSarkany: I extended the XML doc with these details.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
MartinSarkany: I added it to XML doc.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
MartinSarkany: I fixed the race condition.

However, I'm not sure about container re-creation as the container should exist after successful initialization and we probably can't re-create the same container if it's initialized using the `Initialize()` method. The caller has an option to react to failure by providing a new Container or using `InitializeAsync()` instead.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -163,7 +172,35 @@ public async Task InitializeAsync(
...
 
-            this.container = containerResponse.Container;
+            this.SetContainer(containerResponse.Container);
+        }
+
+        /// <summary>
+        /// Initializes the provider with an already created Cosmos DB container that meets the required partition key definition (<c>/id</c>).
+        /// </summary>
+        /// <param name="container">Existing Cosmos DB container containing wrapped DEKs or ready to store them.</param>
+        public void Initialize(Container container)
```
kirankumarkolli, line 181: Wondering if we can leverage `CollectionCache` instead of reading every time?

Only side-affect might be freshness aspect.

kirankumarkolli: @juraj-blazek you are right, internals not visible.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
kirankumarkolli, line 166: Push the validation inside `SetContainer`

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -30,9 +37,10 @@ internal Container Container
         {
             get
             {
-                if (this.container != null)
+                Container container = Volatile.Read(ref this.container);
```
kirankumarkolli: Volailte.Read us unnecessary in all places, CLR default guarantees are enough for the scenario.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
MartinSarkany: Done

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
MartinSarkany: Fixed

kirankumarkolli (approved):

kundadebdatta (approved):
LGTM.

PR #5428, opened 2025-10-01
Client Encryption: Adds distributed cache support to DEK caching

### LINKED ISSUE #4678: [Client encryption]: Optimize encryption/decryption flows

adamnova:
## Description

Introduces an optional `IDistributedCache` L2 in `DekCache` so the SDK's in-process DEK-properties cache can be rescued by a peer-populated entry when Cosmos metadata is momentarily unavailable. Adds a proactive background-refresh window, required per-provider cache-key prefix scoping, and format-version-scoped cache keys for rolling-upgrade safety.

Closes https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4678

## Motivation

`CosmosDataEncryptionKeyProvider`'s in-process `DekCache` has a TTL. When that TTL elapses, the next encrypted request must refetch `DataEncryptionKeyProperties` from the Cosmos metadata container. If Cosmos metadata is transient-unavailable at that moment, the request fails even though the wrapped-DEK bytes haven't changed. For fleets that already run an external cache (Redis, or a staged composite), this PR lets peers rescue each other: peer A populates L2, peer B's L1 expires during a Cosmos blip, peer B reads from L2 and the request succeeds.

## Design notes

- **Cache hierarchy:** `DekPropertiesCache` (in-proc, TTL) → `IDistributedCache` (opt-in, consumer-supplied) → Cosmos metadata.
- **L1-expiry path:** refresh goes through the L2-aware fetcher first, falling through to Cosmos only on an L2 miss. A peer-populated L2 rescues the caller when Cosmos is unavailable.
- **L2 lifetime is decoupled from L1.** L2's `AbsoluteExpiration` is stamped at `utcNow() + distributedCacheEntryLifetime` (default `2 × dekPropertiesTimeToLive`) so an L2 entry outlives the writer's (and peer's) L1 expiry and is actually usable at the moment it's needed.
- **Cache key shape:** `{prefix}:v{cacheFormatVersion}:{Uri.EscapeDataString(dekId)}`.
  - `{prefix}` — **required** when `distributedCache != null`. Callers must choose a prefix unique per DEK container scope (for example derived from account/database/container identifiers) so distinct providers sharing one cache cannot collide on identical DEK ids.
  - `v{N}` — embeds the cache-format version so mixed-SDK-version fleets write to disjoint slots; a v1 reader cannot overwrite a v2 peer's entry during a rolling upgrade.
  - `Uri.EscapeDataString(dekId)` — eliminates colon-collision when dekIds contain reserved separator characters.
- **Fail-open contract:** L2 read failures (exceptions, timeouts the implementation surfaces as OCE without caller cancellation) never fail the request — they fall through to Cosmos. Distributed-cache writes are fire-and-forget and swallow exceptions. Caller cancellation is still honoured.
- **Integrity guards:** deserialised payloads are validated for required fields (`Id`, `WrappedDataEncryptionKey`, `EncryptionKeyWrapMetadata`, `EncryptionAlgorithm`); incomplete entries are treated as a miss. Source fetchers returning a DEK whose `Id` disagrees with the requested dekId are rejected before caching.

## Public API additions

Three new `CosmosDataEncryptionKeyProvider` constructor overloads accept `IDistributedCache distributedCache, TimeSpan? refreshBeforeExpiry, string distributedCacheKeyPrefix`. `refreshBeforeExpiry` (previously `proactiveRefreshThreshold`) is the time window before expiry at which a background refresh is triggered; `distributedCacheKeyPrefix` is required when `distributedCache != null`.

## Dependency impact

- **Library:** replaces `Microsoft.Extensions.Caching.Memory` (previously `3.1.7` on `netstandard2.0` / `1.1.2` on `net46`) with `Microsoft.Extensions.Caching.Abstractions` `3.1.7`, unified across TFMs. The library consumes only `IDistributedCache`; the `MemoryCache` reference was dead.
- **Transitive impact:** consumers depending on `Microsoft.Extensions.Caching.Memory` types **through this package** must add a direct reference. Consumers using only `IDistributedCache` are unaffected. The `Abstractions` floor stays at the lowest version the new API surface compiles against, so consumers remain free to unify upward to any LTS.
- `Microsoft.Bcl.TimeProvider` was briefly introduced for testable time and removed in favour of a lighter `Func&lt;DateTime&gt;` seam; consumers inherit no new dependency from it.

## Test coverage

Five behaviour-grounded test files land with the feature:

- `DekCacheResilienceTests.cs` — cold/expired L1 × L2 present/absent × Cosmos up/down matrix, L2 TTL coupling, provider isolation, corrupted/partial payloads, future-version tolerance, background-refresh L1 poisoning.
- `DekCacheResilienceScenariosTests.cs` — concurrency/coalescing under expired L1, cancellation propagation, clock edges, proactive-refresh coalescing, fetcher-Id integrity, slow/failing L2 with caller cancellation.
- `DekCacheInteropTests.cs` — cross-process round-trip fidelity, pinned JSON property names, no `$type` discriminator in payload, forward-compatible extra fields, missing-`v` payload treated as v1, concurrent peer writes, cross-instance `RemoveAsync` invalidation.
- `DekCacheSecurityInvariantsTests.cs` — raw-DEK never leaked to L2, `$type` deserialisation-gadget rejection, cache-key colon-collision isolation, fail-open on L2 errors without leaking error strings into results, write-side fire-and-forget failure containment.
- `DekCacheDistributedCacheContractTests.cs` — `IDistributedCache` contract conformance (null vs empty `byte[]`, exception handling, cancellation semantics, `AbsoluteExpiration` validity, no `RefreshAsync` calls, concurrent-caller safety).

Every test cites a `// REQ:` requirement and `// SOURCE:` grounding (PR description, XML docs, `IDistributedCache` contract, `AsyncCache` semantics). Tests use an injected `Func&lt;DateTime&gt;` and a clock-aware `IDistributedCache` test double; no `Thread.Sleep`, no wall-clock dependencies.

**Totals:** 327 unit tests + 77 emulator tests = **404 passing**, 0 warnings, 0 errors.

## Type of change

- [x] This change requires a documentation update (changelog entry for 1.0.0-preview09 included)

## Closing issues

closes #4678


Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
juraj-blazek, line 128: Note we need to extend also this, in ALE we use both `EncryptionKeyWrapProvider` and `EncryptionKeyStoreProvider` for backwards compatibility

adamnova: I added new constructors

Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
```diff
@@ -50,12 +50,18 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="3.1.7" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Abstractions" Version="3.1.7" />
+    <PackageReference Include="Microsoft.Bcl.TimeProvider" Version="8.0.0" />
```
juraj-blazek: Let's see if we can remove this one

adamnova: Done in 97c4a213 — removed `Microsoft.Bcl.TimeProvider` from both the src and test csproj and replaced `TimeProvider` usage in `DekCache` with a `Func<DateTime>` field (defaults to `() => DateTime.UtcNow`). Tests use a closure-captured `DateTime` instead of `FakeTimeProvider`. All 266 unit tests pass.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -88,11 +111,33 @@ public CosmosDataEncryptionKeyProvider(
...
+        /// <param name="dekPropertiesTimeToLive">Time to live for DEK properties before having to refresh.</param>
+        /// <param name="distributedCache">Optional distributed cache implementation for cross-process/cross-instance caching of DEK properties.</param>
+        /// <param name="refreshBeforeExpiry">Optional time window before expiry to trigger proactive background refresh. For example, TimeSpan.FromMinutes(10) will refresh DEKs 10 minutes before they expire.</param>
+        /// <param name="distributedCacheKeyPrefix">Required prefix for distributed cache keys. Must be unique per DEK container scope (for example, derived from account/database/container identifiers) so that multiple providers sharing the same cache instance do not collide on identical DEK ids.</param>
+        /// <remarks>
+        /// When using a distributed cache, ensure the cache infrastructure is configured with encryption in transit (TLS)
+        /// and encryption at rest. The cache stores wrapped (encrypted) DEK properties including key metadata.
+        /// </remarks>
+        public CosmosDataEncryptionKeyProvider(
```
kundadebdatta, line 110: Three new public ctor overloads (one of which is [Obsolete]) push CosmosDataEncryptionKeyProvider to 6 public constructors with overlapping parameter sets. Two of the new overloads are themselves marked [Obsolete]. An options-bag overload would be cleaner and future-proof, something like this:

```
public CosmosDataEncryptionKeyProvider(
    EncryptionKeyStoreProvider provider,
    DekCacheOptions options) // TTL, IDistributedCache, refreshBeforeExpiry, prefix, lifetime
```

This also avoids the next round of overloads when (not if) more knobs land.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
```diff
@@ -72,10 +161,94 @@ public async Task<InMemoryRawDek> GetOrAddRawDekAsync(
...
+                    {
+                        using (Activity dcActivity = ActivitySource.StartActivity("DekCache.UpdateDistributedCache"))
+                        {
+                            dcActivity?.SetTag("cache.system", "cosmos.encryption.dek");
+                            dcActivity?.SetTag("cache.key", dekId);
+                            dcActivity?.SetTag("cache.operation", "write");
+
+                            try
+                            {
+                                await this.UpdateDistributedCacheAsync(dekId, cachedDekProperties, CancellationToken.None).ConfigureAwait(false);
```
kundadebdatta: The PR description says "Distributed-cache writes are fire-and-forget" but in `FetchFromSourceAndUpdateCachesAsync` the call is awaited:

```
await this.UpdateDistributedCacheAsync(dekId, cachedProperties, cancellationToken);
return cachedProperties;
```
Only SetDekProperties (explicit set) wraps the L2 write in Task.Run. 

Effect: a slow/unreachable Redis on the cold path stretches every L1 miss by the L2 write latency after the customer already paid Cosmos's metadata RTT — exactly the scenario the feature is supposed to defend against. 

Either fire-and-forget here too, or at minimum impose an explicit short timeout on the L2 write so a degraded L2 cannot regress availability.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
```diff
@@ -72,10 +161,94 @@ public async Task<InMemoryRawDek> GetOrAddRawDekAsync(
...
+                                dcActivity?.SetTag("cache.result", "success");
+                            }
+                            catch (Exception ex)
+                            {
+                                dcActivity?.SetTag("cache.result", "failure");
+                                dcActivity?.SetTag("cache.error", ex.GetType().Name);
+                                dcActivity?.SetStatus(ActivityStatusCode.Error, ex.Message);
+
+                                // Log the failure but don't fail the operation
+                                Debug.WriteLine($"Failed to update distributed cache for DEK '{dekId}': {ex.Message}");
```
kundadebdatta: L2 read/write failures and RemoveAsync failures all go through `Debug.WriteLine`, which is stripped in Release builds. 

Production telemetry on L2 health collapses to ActivitySource alone. The rest of the SDK uses `DefaultTrace.TraceWarning` / `TraceError` or `TraceVerbose` — please align with that.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
```diff
@@ -1,25 +1,90 @@
...
+
+            // A distributed cache without a caller-supplied prefix is ambiguous: multiple
+            // providers sharing a single cache would silently collide on identical dekIds. A
+            // prefix is required whenever distributed caching is enabled so the caller must
+            // consciously partition the keyspace (e.g. by container id, tenant id, etc.).
+            if (distributedCache != null)
+            {
+                ArgumentValidation.ThrowIfNullOrWhiteSpace(cacheKeyPrefix, nameof(cacheKeyPrefix));
+            }
+            else if (cacheKeyPrefix != null)
```
kundadebdatta: The `else if` branch validates a prefix that the cache will never use. 

Either reject outright (distributedCache == null && cacheKeyPrefix != null) so the caller gets a clear signal of misconfiguration, or accept silently. Current behavior is the worst of both.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
```diff
@@ -72,10 +161,94 @@ public async Task<InMemoryRawDek> GetOrAddRawDekAsync(
...
+                // Update memory cache
+                this.DekPropertiesCache.Set(dekId, cachedDekProperties);
+                activity?.SetTag("cache.memory.updated", true);
+
+                // Update distributed cache if available (fire-and-forget)
+                if (this.distributedCache != null)
+                {
+                    activity?.SetTag("cache.distributed.enabled", true);
+
+                    this.LastDistributedCacheWriteTask = Task.Run(async () =>
```
kundadebdatta: The fire-and-forget `Task.Run(...)` captures `CancellationToken.None` and this. There's no provider-level CTS that signals "we're shutting down, abort pending L2 writes." 

If a provider is disposed during a burst of SetDekProperties calls, the background tasks keep running, holding references to `IDistributedCache` (which the caller may also be disposing). 

Add a long-lived `CancellationTokenSource` on DekCache and pass its token in?

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
adamnova: Done in 6a6ddeb9 — cold-miss path is now fire-and-forget too, via a shared helper used by both `SetDekProperties` and `FetchFromSourceAndUpdateCachesAsync`.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
adamnova: Done in 6a6ddeb9. Added an internal `EncryptionCustomEventSource` named `Azure-Cosmos-Encryption-Custom` (same shape as `CosmosDbEventSource`) so the four sites emit Warning events that `AzureEventSourceListener` and `dotnet-trace` pick up automatically.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
adamnova: Done in b07babb1 — now throws `ArgumentException` when a prefix is supplied without a distributed cache.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
adamnova: Done in 6a6ddeb9. `DekCache` and the provider now implement `IDisposable` + `IAsyncDisposable`, with a private CTS, tracked in-flight writes, and bounded drain on dispose.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
adamnova: Done in 1ae0bce5 — added `DekCacheOptions` and a `(EncryptionKeyStoreProvider, DekCacheOptions)` ctor for the MDE path. The hybrid wrap+store case goes through `Create(wrap, store, DekCacheOptions)` instead of a 3rd ctor (see the J0 thread for the CS0121 reason). Legacy wrap-only and dual-provider ctors are `[Obsolete]` — 4 public ctors remain.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
```diff
@@ -53,53 +165,515 @@ public async Task<InMemoryRawDek> GetOrAddRawDekAsync(
...
+                dto.ServerPropertiesExpiryUtc);
+        }
+
+        // DTO for serialization to distributed cache.
+        // JsonProperty names are pinned to ensure cross-process interop
+        // regardless of host-level JsonConvert.DefaultSettings.
+        private sealed class CachedDekPropertiesDto
+        {
+            [JsonProperty("v")]
+            public int Version { get; set; } = CurrentCacheFormatVersion;
```
kundadebdatta: **Cross-version safety: `Version` default should be a literal `1`, not `CurrentCacheFormatVersion`.**

The PR comment on `CurrentCacheFormatVersion` is explicit that the version mechanism is meant to isolate mixed-SDK-version fleets so a v1 reader cannot accept a v2 writer's entry. Today (`CurrentCacheFormatVersion = 1`), a payload that lacks the `v` field deserializes with `Version = 1` and the `dto.Version != CurrentCacheFormatVersion` check passes — which matches the documented "missing-`v` payload treated as v1" semantic only by coincidence.

The day this constant is bumped to `2`, a v1-shaped payload missing `v` will deserialize with `Version = 2` (default) and silently sail past the version-equality check, defeating the entire mixed-fleet isolation guarantee at the payload level. The cache *key* segment `v{N}` still provides isolation in steady state, but anyone who bypasses the key (tests, manual cache inspection, the planned v1→v2 migration helper) loses the payload-level guard.

Suggested fix — pin the default to a literal `1` so the "missing `v` ≡ v1" semantic survives version bumps:

```csharp
public int Version { get; set; } = 1;
```

…or use `int?` and reject `null` explicitly. Either way, please add a regression test that asserts the missing-`v` semantic against a hard-coded fixture, so a future version bump that breaks this rule fails the test rather than silently weakening the guard.

kundadebdatta (commented):
Three findings from a deep review focused on the L2 distributed-cache layer added in this PR. All three close real safety/contract gaps that the rest of the design is implicitly asking the reader to trust. The remaining tightenings (lifecycle CTS disposal, sync-Dispose shutdown budget, proactive-refresh telemetry, prefix delimiter validation) are non-blocking and can be follow-ups. Test discipline (`// REQ:` / `// SOURCE:`) is impressive — please preserve that on subsequent PRs.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
```diff
@@ -53,53 +165,515 @@ public async Task<InMemoryRawDek> GetOrAddRawDekAsync(
...
+                    byte[] cachedBytes = await this.distributedCache.GetAsync(
+                        this.GetDistributedCacheKey(dekId),
+                        cancellationToken).ConfigureAwait(false);
+
+                    stopwatch.Stop();
+                    activity?.SetTag("cache.latency_ms", stopwatch.ElapsedMilliseconds);
+
+                    if (cachedBytes != null)
+                    {
+                        CachedDekProperties cachedProps = DeserializeCachedDekProperties(cachedBytes);
```
kundadebdatta, line 410: **L2 hydration should reject payloads whose `Id` doesn't match the requested `dekId` (prefix-collision defense).**

The source path in `FetchFromSourceAndUpdateCachesAsync` already defensively rejects fetcher results with a mismatched id:

```csharp
if (serverProperties == null || !string.Equals(serverProperties.Id, dekId, StringComparison.Ordinal))
{
    throw new InvalidOperationException($"DEK fetcher returned ... '{returnedId}' for requested dekId '{dekId}'. Refusing to cache mismatched properties.");
}
```

…which is exactly the right check on the write path. But the L2 read path performs only a *shape* check inside `DeserializeCachedDekProperties` (non-null Id / WrappedKey / WrapMetadata / Algorithm) — it never compares `props.Id` to the requested `dekId`.

If two providers misconfigure their `DistributedCacheKeyPrefix` so their key spaces overlap, this caller silently returns a `DataEncryptionKeyProperties` whose wrapped bytes belong to a *different* logical DEK whose Id happens to be the requested string. `RawDekCache` then unwraps with the wrong key material and surfaces a confusing failure deep in the unwrap pipeline — not a clear "prefix collision detected" message.

Suggested fix — add the same identity check on the L2 read path, treat mismatch as a miss (fail-open):

```csharp
if (!string.Equals(cachedProps.ServerProperties.Id, dekId, StringComparison.Ordinal))
{
    EncryptionCustomEventSource.DistributedCacheReadFailed(
        dekId,
        new InvalidOperationException($"L2 entry Id '{cachedProps.ServerProperties.Id}' != requested '{dekId}'; treating as miss."));
    return null;
}
```

Worth pairing with a unit test for the colon-collision case — the security-invariants test file already has a `cache-key colon-collision isolation` test that could be extended.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionCustomEventSource.cs
```diff
@@ -0,0 +1,121 @@
...
+        public static void DistributedCacheReadFailed(string dekId, Exception exception)
+        {
+            if (exception == null)
+            {
+                return;
+            }
+
+            if (Singleton.IsEnabled(EventLevel.Warning, EventKeywords.None))
+            {
+                Singleton.DistributedCacheReadFailedCore(dekId ?? string.Empty, exception.GetType().FullName, exception.Message);
```
kundadebdatta: **EventSource doc claims "no full exception strings" but the implementation ships `exception.Message` verbatim.**

The class-level remark states:

> Payloads must not contain key material, wrapped key bytes, or full exception strings.

…but each event method does:

```csharp
Singleton.DistributedCacheReadFailedCore(dekId ?? string.Empty, exception.GetType().FullName, exception.Message);
```

For most `IDistributedCache` implementations `exception.Message` is benign, but Redis/SQL/HTTP-backed implementations can produce messages that embed:

- connection strings, server-side state, partial command text (StackExchange.Redis frequently includes the failing endpoint and sometimes a key prefix),
- parameter values (SqlException),
- URLs with embedded credentials (HttpRequestException).

This contradicts the stated invariant and could leak operational secrets into telemetry that's auto-discovered by `AzureEventSourceListener` — exactly the broad audience the source name was chosen to attract. The current shape is also inconsistent with the rest of the SDK (`CosmosDbEventSource` carefully avoids unstructured `Message` payloads).

Additionally, the `dekId` itself is written verbatim — the "no key material" guard stops at the *bytes*, but customer-meaningful dekIds emitted to broad EventSource subscribers may be a privacy concern worth documenting.

Suggested fix — either (a) tighten the doc claim to match the implementation, or (b) restrict the payload to `exception.GetType().FullName` plus a stable category string (e.g. `"timeout"`, `"connection"`, `"deserialize"`) and rely on the surrounding Activity for unstructured details (which has subscriber-controlled scope). Option (b) is what the rest of the SDK does and is what I'd recommend.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
adamnova: Used a `Create` factory rather than another ctor — `(wrap, store, DekCacheOptions)` was creating CS0121 ambiguity with the existing `(wrap, store, TimeSpan?)` obsolete ctor when callers pass positional null. 8da1f35a

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
adamnova: Pinned to literal `1` in 3cc3ed8a. Two tests in `DekCacheInteropTests` — reflection check on the DTO default, and a hard-coded JSON fixture without the `v` field that has to round-trip as v1.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
adamnova: Done in f20338b9 — Ordinal Id check after deserialize, fail-open as miss, new event id 5 for the mismatch. Observed Id truncated to 128 since it's untrusted peer input. Extended the colon-collision test in `DekCacheSecurityInvariantsTests`.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionCustomEventSource.cs
adamnova: Option (b) in de65daab. payload[2] is now a category (`timeout`/`connection`/`deserialize`/`other`), exception type FullName still in payload[1]. Kept `ex.Message` on Activity tags — matches `OpenTelemetryAttributeKeys.ExceptionMessage` in the main SDK. Sentinel-in-message test covers all 4 channels.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
```diff
@@ -1,25 +1,113 @@
...
+        private static readonly TimeSpan DisposeDrainTimeout = TimeSpan.FromSeconds(5);
+
         private readonly TimeSpan dekPropertiesTimeToLive;
+        private readonly TimeSpan distributedCacheEntryLifetime;
+        private readonly TimeSpan? refreshBeforeExpiry;
+        private readonly IDistributedCache distributedCache;
+        private readonly string cacheKeyPrefix;
+        private readonly Func<DateTime> utcNow;
+
+        private readonly CancellationTokenSource disposalCts = new CancellationTokenSource();
```
kundadebdatta, line 33: `disposalCts` is never disposed — `Dispose()` and `DisposeAsync()` cancel but don't dispose. Add `this.disposalCts.Dispose()` after the bounded drain in both paths. 

`CA2213` will flag this the moment analyzers light up.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -256,5 +340,37 @@ private async Task<DataEncryptionKey> FetchDekAsync(string id, string encryption
...
+
+        /// <summary>
+        /// Cancels in-flight background distributed-cache writes owned by this provider's
+        /// <see cref="DekCache"/> and best-effort drains them. Idempotent. Does NOT dispose
+        /// externally-supplied dependencies (providers, container, or
+        /// <see cref="Microsoft.Extensions.Caching.Distributed.IDistributedCache"/>) — caller
+        /// owns those lifetimes. <see cref="DekCache.RemoveAsync(string,System.Threading.CancellationToken)"/>
+        /// invalidations in flight are not interrupted by disposal.
+        /// </summary>
+        public void Dispose()
```
kundadebdatta, line 348: `Dispose()` / `DisposeAsync()` mutual short-circuit on the provider. If a caller does `provider.Dispose()`; `await provider.DisposeAsync()`;, the await completes before the synchronous drain on the other 
thread returns — DI shutdown ordering can disturb. 

The provider-level isDisposed CAS is structurally redundant (DekCache's own CAS is idempotent). Remove it and forward unconditionally, or document that interleaving isn't safe.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCacheOptions.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// strictly less than <see cref="DekPropertiesTimeToLive"/> when both are specified.
+        /// </summary>
+        public TimeSpan? RefreshBeforeExpiry { get; set; }
+
+        /// <summary>
+        /// Gets or sets the prefix for distributed cache keys. Required when
+        /// <see cref="DistributedCache"/> is non-<see langword="null"/>: must be unique per
+        /// DEK container scope (for example, derived from account/database/container identifiers)
+        /// so that multiple providers sharing the same cache instance do not collide on
+        /// identical DEK ids. Ignored when <see cref="DistributedCache"/> is
```
kundadebdatta: XML doc still says "Ignored when DistributedCache is null" — but the ctor throws. This is a fossil from the round before the prefix-without-cache fi. DI/IConfiguration.Bind consumers who
unconditionally populate the prefix will trust the doc, ship, then crash on first request. Tighten the doc to match the throw.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
```diff
@@ -1,25 +1,113 @@
...
+            Func<DateTime> utcNow = null,
+            TimeSpan? distributedCacheEntryLifetime = null)
         {
             this.dekPropertiesTimeToLive = dekPropertiesTimeToLive.HasValue == true ? dekPropertiesTimeToLive.Value : TimeSpan.FromMinutes(Constants.DekPropertiesDefaultTTLInMinutes);
+
+            // Distributed cache entries live longer than L1 so a peer-populated entry can rescue
+            // a request after L1 expiry. Default = 2x the L1 TTL.
+            this.distributedCacheEntryLifetime = distributedCacheEntryLifetime
+                ?? TimeSpan.FromTicks(this.dekPropertiesTimeToLive.Ticks * 2);
+            if (this.distributedCacheEntryLifetime <= this.dekPropertiesTimeToLive)
```
kundadebdatta: `DistributedCacheEntryLifetime` validation runs even when L2 is disabled. The check here is outside the if `(distributedCache != null)` gate, so a consumer who templates { DekPropertiesTimeToLive = 1h, `DistributedCacheEntryLifetime` = 30m } from config but leaves `DistributedCache = null` (non-prod env, feature-flag flip) gets an `ArgumentOutOfRangeException` complaining about "L2 entries" they don't have. Asymmetric with the same-round fix for cacheKeyPrefix — move into the L2 gate

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
adamnova: 3cb83c13 — dropped the provider-level CAS (truly redundant since `DekCache.Dispose/DisposeAsync` already CAS-idempotent and `this.DekCache?.Dispose()` null-safes), removed the unused `isDisposed` field, and added an XML remark spelling out that concurrent `Dispose`/`DisposeAsync` interleaving is unsupported.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCacheOptions.cs
adamnova: Good catch — fossil from the prefix-without-cache round. Tightened the doc in 3cb83c13 to say `Must be null when DistributedCache is null; the constructor rejects supplying a prefix without a distributed cache as misconfiguration.`

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
adamnova: Done in 3cb83c13 — added `disposalCts.Dispose()` after the bounded drain in both paths. No actual handle leak today (no `CancelAfter`, no `Token.WaitHandle` access) but agreed CA2213 hygiene.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCache.cs
adamnova: Moved into the L2 gate in 3cb83c13 — matches the same-round prefix fix. Added two tests: L2-off with a short lifetime now constructs cleanly, L2-on still throws `ArgumentOutOfRangeException`.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCacheOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <see langword="null"/>.
+        /// </summary>
+        public TimeSpan? DekPropertiesTimeToLive { get; set; }
+
+        /// <summary>
+        /// Gets or sets the optional distributed (L2) cache implementation enabling
+        /// cross-process / cross-instance caching of DEK properties. Leave <see langword="null"/>
+        /// to use only the in-memory cache.
+        /// </summary>
+        public IDistributedCache DistributedCache { get; set; }
```
juraj-blazek: Consider moving all the distributed cache settings to a dedicated `DistributedCacheOptions`. It will make the validation more straightforward (e.g. the prefix will be always required).

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -113,29 +124,100 @@ public CosmosDataEncryptionKeyProvider(
...
+
+        /// <summary>
+        /// Mirrors <see cref="EncryptionKeyStoreProvider.DataEncryptionKeyCacheTimeToLive"/> onto
+        /// the static <see cref="ProtectedDataEncryptionKey.TimeToLive"/>. Process-global mutation;
+        /// shared across providers.
+        /// </summary>
+        private void InitializeProtectedDataEncryptionKeyTimeToLive()
+        {
+            this.PdekCacheTimeToLive = this.EncryptionKeyStoreProvider.DataEncryptionKeyCacheTimeToLive;
+            ProtectedDataEncryptionKey.TimeToLive = this.PdekCacheTimeToLive ?? EffectivelyForeverTimeToLive;
```
juraj-blazek, line 213: We need to get rid of all this static mess in the MDE SDK, can you please create tracking item.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -113,29 +124,100 @@ public CosmosDataEncryptionKeyProvider(
...
+            this.MdeKeyWrapProvider = new MdeKeyWrapProvider(encryptionKeyStoreProvider);
+            this.dataEncryptionKeyContainerCore = new DataEncryptionKeyContainerCore(this);
+            this.DekCache = NewDekCache(dekCacheOptions);
+            this.InitializeProtectedDataEncryptionKeyTimeToLive();
+        }
+
+        /// <summary>
+        /// Materialises a <see cref="DekCache"/> from a (possibly null) <see cref="DekCacheOptions"/>.
+        /// </summary>
+        private static DekCache NewDekCache(DekCacheOptions dekCacheOptions)
```
juraj-blazek, line 200: nit: consider changing the `DekCache` constructor to directly accept the new `DekCacheOptions` and create dedicated validator for the options that we can invoke from the constructor on the passed options.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DekCacheOptions.cs
adamnova: Going to do this in-PR after all. Nesting `DistributedCache` + `KeyPrefix` + `EntryLifetime` into a dedicated `DistributedCacheOptions` so the prefix-required and lifetime-validation invariants live at the type level instead of being construction-time checks. Drive-by: already tightened the `DistributedCacheEntryLifetime` XML doc in 308f216f to match the C4 L2-gated validation; the nested shape will obsolete that doc note.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
adamnova: Yeah this one's been bugging me too — the XML doc on `InitializeProtectedDataEncryptionKeyTimeToLive` already calls out the last-writer-wins. Tracked internally to follow up. The PR already passes the per-provider TTL through to `MdeEncryptionAlgorithm.CreateAsync`, so the static assignment may be vestigial once MDE supports instance-scoped PDEK TTL.

Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
adamnova: Doing it in-PR alongside J1 — collapsing the 6-positional ctor to `DekCache(DekCacheOptions options, Func<DateTime> utcNow = null)` and extracting validation into a separately-testable `DekCacheOptionsValidator`. `utcNow` stays as a ctor-only parameter (test seam, not user config).

juraj-blazek (approved):

juraj-blazek (approved):

kundadebdatta (approved):
LGTM now.

kushagraThapar (approved):
LGTM, thanks @adamnova for the changes!

PR #5437, opened 2025-10-06
MSRC: Fixes dummy account endpoint and keys in tests

sourabh1007:
## Description
Renaming cosmos endpoints and keys to dummy endpoint in unit tests. So that it won't be caught up in MSRC.

### DIFF
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ActiveClientDiagnosticTest.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ActiveClientDiagnosticTest.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Azure.Cosmos.Tests
     [TestClass]
     public class ActiveClientDiagnosticTest
     {
-        private const string ConnectionString = "AccountEndpoint=https://example.documents.azure.com:443/;AccountKey=<placeholder>;";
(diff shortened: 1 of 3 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ActiveClientDiagnosticTest.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Azure.Cosmos.Tests
     [TestClass]
     public class ActiveClientDiagnosticTest
     {
-        private const string ConnectionString = "AccountEndpoint=https://localtestcosmos.documents.azure.com:443/;AccountKey=425Mcv8CXQqzRNCgFNjIhT424GK99CKJvASowTnq15Vt8LeahXTcN5wt3342vQ==;";
+        private const string ConnectionString = "AccountEndpoint=https://dummy.documents.azure.com:443/;AccountKey=dummykey;";
```
Pilchie: I think the normal guidance for credscan is to use `<PLACEHOLDER>` for places where credentials might go.

Pilchie: Also, `dummy` isn't typically a good choice of name.

sourabh1007: I wanted to put something which doesn't sound like a real account name and key.

Pilchie (approved):

PR #5447, opened 2025-10-14
Per Partition Automatic Failover: Adds Hub Region Processing Only While Routing Requests Failed with 404/1002 for single master accounts.

### LINKED ISSUE #5440: [Per Partition Automatic Failover] Use Hub Region Processing Only While Routing Requests Failed with `404/1002` for single master account.
### Background:

This task identifies a possible gap/ issue with the existing PPAF failover logic with session consistency. Let’s imagine the following setup, we have a two-region (R0 and R1) single master session consistency cosmos account with R0 being the primary write region. The account has partition level failover enabled with 2 server partitions, P0 and P1 respectively.

- Let’s consider at time T0, partition P0 in region R0, experienced a full quorum loss on all 4 replicas which triggered the partition to fail over to the next region, i.e. R1. 

- At time T1, the partition P0 fails over to R1 from the backend.

- At time T2, the PPCB and PPAF flows kicks in the SDK, and the reads and writes successfully fail over from the SDK to the region R1.

- Now, at time T3, a write request for the partition P0 lands on to the region R1 and the write is now successfully served from R1, with a session token in return, say session-token#1.

- At time T4, region R0 comes back up, and the failover manager from the backend starts the failback process to R0 for the offending partition P0. Note that the failback process doesn’t have any SLA guarantee. At this time T4, since region R0 is online, it can serve the reads successfully.

- At time T5, region R0 is able to fully resolve any pending conflicts and able to replicate the missing documents from R1, and now it is all caught up to serve the writes from R0. Therefore, partition P0 is now failed back to the primary write region, i.e. R0.

- Things get a bit interesting, between the time T4 and T5. Since at time T4, the Preferred write region R0, is online and able to serve the reads, there can be two possibilities:

  - If a cosmos client bootstraps for initialization (aka cold start client), the reads will always go to the primary write region first, i.e. R0, (The preferred regions may or may not be supplied). If the read request uses the same session token session-token#1 for any document, served from the region R1, then it is possible that the reads will fail (and continue to fail) with a 404/1002 for until the region R0 is caught up with the missing documents from R1.

  - It is also possible that the circuit breaker will fail back the reads, before region R0 is fully caught up with the missing documents written on region R1. In such cases, since the SDK doesn’t do any cross regional retry on the reads, the reads will continue to fail from R0 with 404/1002.

Image

### Is this really an Issue?

Before we dive deep into this, please note there are two types of failovers and failbacks happen at the SDK layer. 

- Write Failover and Failback: Driven by the PPAF code path.
- Read Failover and Failback: Driven by the PPCB (aka Circuit Breaker) code path.

Note that when PPAF is enabled, the SDK enables the circuit breaker by default. This would imply the reads will start failing over and fall back by default with the PPCB flow.

Now, if the write failback takes an extended period, i.e. (T5-T4) is a much higher value than anticipated, then it’s possible that the reads will go to the primary write region and fail with 404/1002. Because the SDK doesn’t do any cross regional retry on 404/1002, the reads will continue to fail until the primary region R0 is able to serve the session reads. This is true for both cold start clients and PPCB based fail back.

### Acceptance Criteria

Goal is to use the newly added BE header `x-ms-cosmos-hub-region-processing-only` to find out the current write region, while routing the 404/1002 during retry attempt.

### SDK Fix

During partition-level failover and failback under session consistency, a timing gap can cause read requests to fail with `404/1002` errors.  
When a partition temporarily fails over to a secondary region and later begins failing back to the primary region, the SDK’s read circuit breaker (PPCB) may start routing reads back to the primary region before it has fully caught up with the writes from the failover region.

To address this, the SDK leverages the newly added backend header  
`x-ms-cosmos-hub-region-processing-only`  
to detect such conditions and route retry requests to the correct write (hub) region, ensuring successful session-consistent reads during the failback window.

---

#### 1. Partition Failover (Backend Initiated)
- Backend detects quorum loss in region **R0** for partition **P0**.
- Partition **P0** fails over to **R1**.
- SDK’s **PPAF** and **PPCB** flows begin directing traffic to **R1**.

#### 2. Write and Session Token Generation
- Writes are served from **R1**, generating session token `session-token#1`.
- Reads under session consistency also go to **R1** and succeed.

#### 3. Region Recovery and Failback Start
- Region **R0** comes back online.
- Backend starts the failback process to **R0**, but synchronization is not instantaneous.
- The SDK’s circuit breaker may interpret R0 as healthy and start sending reads there.

#### 4. Early Failback Window
- Reads using tokens from **R1** fail with **404/1002**, since **R0** has not yet replicated all updates.
- The SDK does **not** perform cross-regional retries for these errors, causing persistent failures.

#### 5. SDK Retry Using Hub-Region Header
- When the SDK encounters a `404/1002`, it now adds the header: x-ms-cosmos-hub-region-processing-only: true
- The retry request is automatically routed to the backend.
- Backend ensures the request is processed in the **active hub (write) region**, currently **R1** during failback.
- This guarantees that reads remain session-consistent even while R0 is catching up.

#### 6. Backend Validation and Response
- Backend checks the region’s write status via the **ReplicationSequencer**:
- If the request is sent to a non-write region, backend returns **403.3 (NotWriteRegion)**.
- For PPAF accounts, if a secondary replica loses its lease renewal, backend returns **410/1022 (LeaseNotFound)**.
- SDK interprets these responses and redirects the request to the correct hub region.

#### 7. Recovery Completion
- Once **R0** fully synchronizes and regains write ownership, SDK and backend resume normal routing.
- Subsequent reads/writes no longer require the hub-region header.

---

#### Hub-Region Processing

A core dependency for this feature is backend support for processing requests in the hub (write) region when the header  
`x-ms-cosmos-hub-region-processing-only`  
is present.

This functionality guarantees that reads are always executed in the active write region during failover and failback windows.

##### Implementation Phases

- **Phase 1 – ReplicationSequencer Check**
- When the header is received, backend validates `WriteStatus_Global` in **ReplicationSequencer**.
- If the target region is not the current write region, backend returns **403.3 (NotWriteRegion)**.
- Applies to both PPAF and non-PPAF accounts.

- **Phase 2 – PPAF-Enabled Partitions**
- In ungraceful failover cases, topology may be outdated.
- Backend uses `EnableWriteRegionSecondaryLeaseEnforcement`, forcing secondaries to return **410/1022 (LeaseNotFound)** if lease renewal fails.
- Introduces `EnableLessThanStrongLeases`, validating leases only for requests with the above header and consistency < Strong.
- Stable non-PPAF regions continue returning **403.3 (NotWriteRegion)** when not the write region.

- **Phase 3 – Non-PPAF Partitions During Offlining**(TODO)
---

### Summary

This enhancement allows the SDK to bridge the timing gap between partition failover and failback.  
By using the backend header `x-ms-cosmos-hub-region-processing-only`, the SDK detects and routes 404/1002 retries directly to the current write region.  
This guarantees session-consistent reads during transient failback windows.  
On the backend, this is supported through topology validation, lease enforcement, and optional RU charging for barrier reads.

aavasthy:
# Pull Request Template

## Description

**This change is limited to single master accounts.** During partition-level failover and failback under session consistency, a timing gap can cause read requests to fail with 404/1002 errors. When a partition temporarily fails over to a secondary region and later begins failing back to the primary region, the SDK’s read circuit breaker (PPCB) may start routing reads back to the primary region before it has fully caught up with the writes from the failover region. As a result, reads using session tokens from the previous write region may fail because the primary region does not yet have the corresponding session state. Since the SDK currently does not perform cross-regional retries for 404/1002 responses, these reads continue to fail until the primary region is fully synchronized. The goal is to leverage the new backend header **x-ms-cosmos-hub-region-processing-only** to detect such conditions and route retry requests to the correct write (hub) region, ensuring successful session-consistent reads during the failback window.

As part of this change, the hub region header is added to the second retry request that failed with 404/1002 Read Session Not Available. In order to avoid running into a loop of retries to find the hub region, the first retry request for 404/1002 does not container header and retries on the available write region from location cache. If that request also fails with 404/1002 then subsequent retry requests all contain the header. If the request has the header and it does not hit the primary hub region then backend return 403/3. SDK will retry the subsequent 403/3 retry requests with the hub region header added.

## Type of change

## Closing issues

To automatically close an issue: closes #5440

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
+++ b/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
@@ -329,19 +329,19 @@ namespace Microsoft.Azure.Cosmos
                     retryOnPreferredLocations: true);
             }
 
-            if (statusCode == HttpStatusCode.NotFound
-                && subStatusCode == SubStatusCodes.ReadSessionNotAvailable)
+            if (statusCode == HttpStatusCode.NotFound && subStatusCode == SubStatusCodes.ReadSessionNotAvailable)
             {
 #if !INTERNAL
                 // Only set the hub region processing header for single master accounts
-                if (!this.canUseMultipleWriteLocations)
+                // Set header only after the first retry attempt fails with 404/1002
+                if (!this.canUseMultipleWriteLocations && this.sessionTokenRetryCount >= 1)
                 {
                     this.addHubRegionProcessingOnlyHeader = true;
                 }
 #endif
                 return this.ShouldRetryOnSessionNotAvailable(this.documentServiceRequest);
             }
-            
+
             // Received 503 due to client connect timeout or Gateway
             if (statusCode == HttpStatusCode.ServiceUnavailable)
             {
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
@@ -4321,12 +4321,12 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests
 
         [TestMethod]
         [Owner("aavasthy")]
-        [Description("Forces a single 404/1002 from the gateway and verifies ClientRetryPolicy adds x-ms-cosmos-hub-region-processing-only on the retry request.")]
+        [Description("Forces two consecutive 404/1002 responses from the gateway and verifies ClientRetryPolicy sets the hub region header flag after the first retry fails.")]
         public async Task ReadItemAsync_ShouldAddHubHeader_OnRetryAfter_404_1002()
         {
-            bool headerObservedOnRetry = false;
             int requestCount = 0;
-            bool shouldReturn404 = true;
+            int return404Count = 0;
+            const int maxReturn404 = 2; // Return 404/1002 twice
 
             // Created HTTP handler to intercept requests
             HttpClientHandlerHelper httpHandler = new HttpClientHandlerHelper
@@ -4340,52 +4340,45 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests
                     {
                         requestCount++;
 
-                        // Check for hub header on retry (2nd+ request)
-                        if (requestCount > 1 &&
-                            request.Headers.TryGetValues(HubRegionHeader, out IEnumerable<string> values) &&
-                            values.Any(v => v.Equals(bool.TrueString, StringComparison.OrdinalIgnoreCase)))
+                        // Header should NOT be present on first retry (2nd request)
+                        if (requestCount == 2 &&
+                            request.Headers.TryGetValues(HubRegionHeader, out IEnumerable<string> firstRetryValues) &&
+                            firstRetryValues.Any())
                         {
-                            headerObservedOnRetry = true;
+                            Assert.Fail("Header should NOT be present on first retry attempt.");
+                        }
+
+                        // Return fake 404/1002 for first two requests
+                        if (return404Count < maxReturn404)
+                        {
+                            return404Count++;
+
+                            var errorResponse = new
+                            {
+                                code = "NotFound",
+                                message = "Message: {\"Errors\":[\"Resource Not Found. Learn more: https://aka.ms/cosmosdb-tsg-not-found\"]}\r\nActivityId: " + Guid.NewGuid() + ", Request URI: " + request.RequestUri,
+                                additionalErrorInfo = ""
+                            };
+
+                            HttpResponseMessage notFoundResponse = new HttpResponseMessage(HttpStatusCode.NotFound)
+                            {
+                                Content = new StringContent(
+                                    JsonConvert.SerializeObject(errorResponse),
+                                    Encoding.UTF8,
+                                    "application/json"
+                                )
+                            };
+
+                            // Add the substatus header for ReadSessionNotAvailable
+                            notFoundResponse.Headers.Add("x-ms-substatus", "1002");
+                            notFoundResponse.Headers.Add("x-ms-activity-id", Guid.NewGuid().ToString());
+                            notFoundResponse.Headers.Add("x-ms-request-charge", "1.0");
+
+                            return Task.FromResult(notFoundResponse);
                         }
                     }
 
                     return Task.FromResult<HttpResponseMessage>(null);
-                },
-
-                ResponseIntercepter = (response, request) =>
-                {
-                    if (shouldReturn404 &&
-                        request.Method == HttpMethod.Get &&
-                        request.RequestUri != null &&
-                        request.RequestUri.AbsolutePath.Contains("/docs/"))
-                    {
-                        shouldReturn404 = false; // Only return 404 once
-
-                        var errorResponse = new
-                        {
-                            code = "NotFound",
-                            message = "Message: {\"Errors\":[\"Resource Not Found. Learn more: https://aka.ms/cosmosdb-tsg-not-found\"]}\r\nActivityId: " + Guid.NewGuid() + ", Request URI: " + request.RequestUri,
-                            additionalErrorInfo = ""
-                        };
-
-                        HttpResponseMessage notFoundResponse = new HttpResponseMessage(HttpStatusCode.NotFound)
-                        {
-                            Content = new StringContent(
-                                JsonConvert.SerializeObject(errorResponse),
-                                Encoding.UTF8,
-                                "application/json"
-                            )
-                        };
-
-                        // Add the substatus header for ReadSessionNotAvailable
-                        notFoundResponse.Headers.Add("x-ms-substatus", "1002");
-                        notFoundResponse.Headers.Add("x-ms-activity-id", Guid.NewGuid().ToString());
-                        notFoundResponse.Headers.Add("x-ms-request-charge", "1.0");
-
-                        return Task.FromResult(notFoundResponse);
-                    }
-
-                    return Task.FromResult(response);
                 }
             };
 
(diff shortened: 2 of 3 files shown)


Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -322,7 +330,8 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
 
             if (statusCode == HttpStatusCode.NotFound
                 && subStatusCode == SubStatusCodes.ReadSessionNotAvailable)
-            {
+            {
+                this.addHubRegionProcessingOnlyHeader = true;
```
ananth7592: The flag for `addHubRegionProcessingOnlyHeader` is set for all the instances of 404/1002(Read session not found) and consequently we would set the header for every 404/1002 retry, is this expected?

aavasthy: yes

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -322,7 +330,8 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
 
             if (statusCode == HttpStatusCode.NotFound
                 && subStatusCode == SubStatusCodes.ReadSessionNotAvailable)
```
xinlian12: also double check: does this change  also targeted for MM as well? For MM, writes can happen in any region, also enable this for MM might cause regression

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -222,6 +224,12 @@ public void OnBeforeSendRequest(DocumentServiceRequest request)
                     // set location-based routing directive based on request retry context
                     request.RequestContext.RouteToLocation(this.retryContext.RetryLocationIndex, this.retryContext.RetryRequestOnPreferredLocations);
                 }
+            }
+            // If previous attempt failed with 404/1002, add the hub-region-processing-only header
+            if (this.addHubRegionProcessingOnlyHeader)
+            {
+                request.Headers[HubRegionHeader] = bool.TrueString;
+                this.addHubRegionProcessingOnlyHeader = false; // reset after applying
```
xinlian12: what would be the errors returned if SDK try to read from non-hub region?

kirankumarkolli: Also falling back to new hub for that partition.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: confirmed this with backend team and this change is not intended to be used in multi-master.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: SDK ends up getting 403/3(WriteForbidden) if it tries to read from a non-hub region. Then SDK has internal logic where every 403/3 response leads to new endpoint discovery. The SDK then retries the request to the hub region.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -22,7 +22,8 @@ internal sealed class ClientRetryPolicy : IDocumentClientRetryPolicy
     {
         private const int RetryIntervalInMS = 1000; // Once we detect failover wait for 1 second before retrying request.
         private const int MaxRetryCount = 120;
-        private const int MaxServiceUnavailableRetryCount = 1;
+        private const int MaxServiceUnavailableRetryCount = 1;
+        private const string HubRegionHeader = "x-ms-cosmos-hub-region-processing-only";
```
Praveen-Msft: Do we have any central place to host these types of headers in the code base?

aavasthy: yes we are actually using the central file HttpConstants.cs. I removed this line as part of code clean up.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -38,7 +38,10 @@ internal sealed class ClientRetryPolicy : IDocumentClientRetryPolicy
         private bool isMultiMasterWriteRequest;
         private Uri locationEndpoint;
         private RetryContext retryContext;
-        private DocumentServiceRequest documentServiceRequest;
+        private DocumentServiceRequest documentServiceRequest;
+#if !INTERNAL
+        private bool addHubRegionProcessingOnlyHeader;
```
jeet1995: Should `volatile` be used for cross-thread visibility?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -222,8 +225,15 @@ public void OnBeforeSendRequest(DocumentServiceRequest request)
                     // set location-based routing directive based on request retry context
                     request.RequestContext.RouteToLocation(this.retryContext.RetryLocationIndex, this.retryContext.RetryRequestOnPreferredLocations);
                 }
+            }
+#if !INTERNAL
+            // If previous attempt failed with 404/1002, add the hub-region-processing-only header
+            if (this.addHubRegionProcessingOnlyHeader)
+            {
+                request.Headers.Add(HttpConstants.HttpHeaders.ShouldProcessOnlyInHubRegion, bool.TrueString);
+                this.addHubRegionProcessingOnlyHeader = false;
```
jeet1995: Is there a need to set it back to `false`? My understanding is with the header; the goal is to have the request land on the hub for the operation - so the header needs to be set to true until hub is discovered.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -322,7 +332,10 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
 
             if (statusCode == HttpStatusCode.NotFound
                 && subStatusCode == SubStatusCodes.ReadSessionNotAvailable)
-            {
+            {
+#if !INTERNAL
+                this.addHubRegionProcessingOnlyHeader = true;
```
jeet1995: I'd gate this (set it for reads when PPAF is enabled).

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: Removed the code for resetting it to false as now for every subsequent retry for 403/3 will also have the header

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: I dont the volatile keyword is required here. The code follows a sequential pattern where ShouldRetryInternalAsync sets the flag, then OnBeforeSendRequest reads it - with an async/await in between which provides memory barrier. Also see other fields with same pattern sessionTokenRetryCount, serviceUnavailableRetryCount etc. Can add volatile keyword for defensive coding.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: based off the design document shared by backend, this header can be added to read requests for both ppaf and non-ppaf accounts.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
jeet1995: @aavasthy - this change should be gated to Single Writer accounts from SDK-perspective. 404:1002 behavior for multi-writer accounts is to cycle through regions from preferred regions and not just hub region. Otherwise, this will be a regression when backend roles out the change for multi-writer accounts.

aavasthy: added a check so that header gets set only for single master

kirankumarkolli:
PR overall LGTM, discussing one aspect offline.

PR #5452, opened 2025-10-21
[Internal] JSON Binary Encoding: Adds support for Base64 String Encoding

sboshra:
## Description
This PR adds support for encoding and decoding Base64 strings in JSON Binary Encoding. Both Standard and URL-safe Base64 formats are supported, including padded and unpadded variants.

**Summary of Changes**

* **JsonBinaryEncoding**

  * Introduced new type markers:

    * `Base64StringLength1`: Standard Base64 string with 1-byte length and 1-byte padding length
    * `Base64StringLength2`: Standard Base64 string with 2-byte length and 1-byte padding length
    * `Base64UrlStringLength1`: URL-safe Base64 string with 1-byte length and 1-byte padding length
    * `Base64UrlStringLength2`: URL-safe Base64 string with 2-byte length and 1-byte padding length
    
  * Added encoding functions:
    * `TryConvertBase64StringToBytes` — decodes a Base64 string (standard or URL-safe) into a byte buffer.
    * `ConvertBytesToBase64String` — encodes a byte buffer into either a standard or URL-safe Base64 string.

* **JsonBinaryWriter**
  * Updated `ForceRewriteRawJsonValue` to respect Base64 string encoding settings.
  * Extended `TryEncodeCompressedString` to support encoding Base64 strings as compressed byte buffers.

* **JsonWriterTest**
  * Added `Base64StringsTest` to validate all supported Base64 variants (standard, URL-safe, padded, and unpadded).
  * Expanded round-trip test coverage to include Base64 encoding scenarios.
  * Improved round-trip validation diagnostics to clearly identify cases where strict buffer comparison is not expected.

## Type of change

## Closing issues

None

### DIFF
--- a/Microsoft.Azure.Cosmos.sln
+++ b/Microsoft.Azure.Cosmos.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.14.36603.0 d17.14
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29123.88
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Cosmos", "Microsoft.Azure.Cosmos\src\Microsoft.Azure.Cosmos.csproj", "{36F6F6A8-CEC8-4261-9948-903495BC3C25}"
 EndProject
@@ -181,18 +181,6 @@ Global
 		{021DDC27-02EF-42C4-9A9E-AA600833C2EE}.Release|Any CPU.Build.0 = Release|Any CPU
 		{021DDC27-02EF-42C4-9A9E-AA600833C2EE}.Release|x64.ActiveCfg = Release|Any CPU
 		{021DDC27-02EF-42C4-9A9E-AA600833C2EE}.Release|x64.Build.0 = Release|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Cover|Any CPU.ActiveCfg = Debug|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Cover|Any CPU.Build.0 = Debug|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Cover|x64.ActiveCfg = Debug|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Cover|x64.Build.0 = Debug|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Debug|x64.ActiveCfg = Debug|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Debug|x64.Build.0 = Debug|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Release|Any CPU.Build.0 = Release|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Release|x64.ActiveCfg = Release|Any CPU
-		{D744906A-1091-403F-B0B6-794DE045169A}.Release|x64.Build.0 = Release|Any CPU
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Cover|Any CPU.ActiveCfg = Debug|Any CPU
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Cover|Any CPU.Build.0 = Debug|Any CPU
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Cover|x64.ActiveCfg = Debug|Any CPU


Microsoft.Azure.Cosmos.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29123.88
+# Visual Studio Version 17
+VisualStudioVersion = 17.14.36603.0 d17.14
```
Maya-Painter: Are changes to this file intentional? #Resolved

sboshra: Not really - I'll revert them.

adityasa (approved):
:shipit:

Maya-Painter (approved):

Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.Strings.cs
```diff
@@ -1465,5 +1595,168 @@ private static void DecodeGuidStringValue(ReadOnlySpan<byte> encodedString, bool
...
+            if (stringValue.Length > (4 * ushort.MaxValue)) return false;
+
+            // Accept the value as a candidate if its length is greater than or equal to the minimum
+            // 6-bit compressed string length, and its length modulo 4 is not equal to 1.
+            if (stringValue.Length >= MinCompressedStringLength6) return (stringValue.Length % 4) != 1;
+
+            // If the length is greater than or equal to 24, accept it as a candidate only if it is a
+            // multiple of 4 and ends with '='.
+            // This primarily ensures that values like the '_rid' property are included.
+            return (stringValue.Length >= 24) && ((stringValue.Length % 4) == 0) && (stringValue[stringValue.Length - 1] == '=');
```
FabianMeiswinkel, line 1759: I don't get this - the length % 4 could be 0,2 or 3 - for 0 and 2 the last char would be "=" (because padding is reuqired). But for legnth % 4 == 3 there is no padding - but it is perfectly valid base64.

"abc" --> "YWJj" #Resolved

Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.ValueLengths.cs
```diff
@@ -25,12 +27,14 @@ private static class ValueLengths
...
             private const int CS5BL1 = -11;      // 5-bit Compressed string w/ 1-byte length followed by 1-byte base char
             private const int CS6BL1 = -12;      // 6-bit Compressed string w/ 1-byte length followed by 1-byte base char
-            private const int Arr1 = -13;        // 1-item array
-            private const int Obj1 = -14;        // 1-property object
-            private const int NC1 = -15;         // Fixed-size numeric items of 1-byte item count
-            private const int NC2 = -16;         // Fixed-size numeric items of 2-byte item count
-            private const int ANC1 = -17;        // Array of fixed-size numeric items of 1-byte item count
-            private const int ANC2 = -18;        // Array of fixed-size numeric items of 2-byte item count
+            private const int B64L1 = -13;       // Base64 string 1-byte encoded length + 1-byte padding length
+            private const int B64L2 = -14;       // Base64 string 2-byte encoded length + 1-byte padding length
```
Pilchie, line 31: It's ok that these are renumbered? #Resolved

Microsoft.Azure.Cosmos/src/Json/JsonWriter.JsonBinaryWriter.cs
```diff
@@ -59,8 +63,14 @@ private sealed class JsonBinaryWriter :
...
-                Token,
-                StrUsr,
+                Arr,
+                Arr1,
+                ArrArrNum,
+                ArrNum,
+                NumUI64,
+                Obj,
+                Obj1,
+                StrBase64,
```
Pilchie, line 73: Renumbering is ok here too? #Resolved

Microsoft.Azure.Cosmos/src/Json/JsonWriter.JsonBinaryWriter.cs
```diff
@@ -296,16 +306,19 @@ private enum RawValueType : byte
             /// <param name="initialCapacity">The initial capacity to avoid intermediary allocations.</param>
             /// <param name="enableNumberArrays">Determines whether to enable writing of uniform number arrays.</param>
             /// <param name="enableUint64Values">Determines whether to enable writing of full-precision unsigned 64-bit integer values</param>
+            /// <param name="enableBase64Strings">Determines whether to enable writing of Base64-encoded strings.</param>
             /// <param name="enableEncodedStrings">Determines whether to enable reference string encoding.</param>
             public JsonBinaryWriter(
                 int initialCapacity,
                 bool enableNumberArrays,
                 bool enableUint64Values,
+                bool enableBase64Strings,
```
Pilchie, line 315: By putting this not last, with optional parameters after it, you may take calls that were passing a value to `enableEncodedStrings` and now that value is going to `enableBase64Strings` and `enableEncodedStrings` gets the default value. It's unfortunate when there are default parameter values :-/ #Resolved

Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.Strings.cs
sboshra: We're here doing the opposite - we have a Base64 encoded string and we're checking whether it's candidate for storing it as its decoded bytes ("YWJj" --> "abc").

Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.ValueLengths.cs
sboshra: Yes. The exact value itself isn’t important; what matters at the top level is whether it’s greater than or equal to zero or less than zero. For negative values, the constant name is referenced instead of the value itself.

Microsoft.Azure.Cosmos/src/Json/JsonWriter.JsonBinaryWriter.cs
sboshra: Yes, these are used exclusively in switch statements.

Microsoft.Azure.Cosmos/src/Json/JsonWriter.JsonBinaryWriter.cs
sboshra: There are only a couple of references to the constructor, and both have been updated.

That’s a valid point - I’ll remove the default values for the other two arguments to ensure this isn’t overlooked. I may do that in a subsequent change.

PR #5454, opened 2025-10-21
[ClientEncryption] Remove JsonProcessor from EncryptionOptions

MartinSarkany:
# Pull Request Template

#4678

## Description

This removes `JsonProcessor` from `EncryptionOptions`.
`EncryptionOptions` are used in `RequestOptions` that already contain `JsonProcessor` in `RequestOptions.Properties` dictionary. `EncryptionOptions.JsonProcessor` is redundant and cannot be used in `RequestOptions` subclasses that do not contain `EncryptionOptions`.

## Type of change


Microsoft.Azure.Cosmos.Encryption.Custom/src/JsonProcessor.cs
```diff
@@ -0,0 +1,25 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.Encryption.Custom
+{
+    /// <summary>
+    /// API for JSON processing
+    /// </summary>
+    public enum JsonProcessor
```
juraj-blazek: Do we need to keep this public?

MartinSarkany: Depends on how we want users to set the processor in `RequestOptions`. `RequestOptions.Properties` can store it as enum or string. Do you suggest we only keep string? Or we add helper methods that set it without actually exposing the enum?

adamnova: At this point I would say we might as well keep it as a string. The API will likely only remain for "finite" amount of time and since we will not be adding the convenience extension as per Kiran's request I do not see much value in this enum being public.

MartinSarkany: I made it internal. Please review the changes.

Microsoft.Azure.Cosmos.Encryption.Custom/tests/EmulatorTests/MdeCustomEncryptionTests.cs
```diff
@@ -364,7 +364,10 @@ public async Task StreamProcessor_EndToEnd_RoundTrip()
                     DataEncryptionKeyId = dekProperties.Id,
                     EncryptionAlgorithm = CosmosEncryptionAlgorithm.MdeAeadAes256CbcHmac256Randomized,
                     PathsToEncrypt = new List<string> { "/Sensitive" },
-                    JsonProcessor = JsonProcessor.Stream
+                },
+                Properties = new Dictionary<string, object>
+                {
+                    { "encryption-json-processor", JsonProcessor.Stream }
```
kirankumarkolli: JsonProcessor is an internal type and not accessible for external usage. 
Please update test exactly the way external CX use.

MartinSarkany: I updated tests to use string instead of enum.

PR #5466, opened 2025-10-29
Proposal: Change feed metadata to use attribute level json converter

yash2710:
Adding on to #5191, the proposal is to use `JsonProperty` and `JsonConverter` at attribute level instead of class level. 

This PR includes all changes from #5191 and the additional changes.

Updates tests to reflect that the previous image might not be available for delete operations


Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs
```diff
@@ -50,9 +50,24 @@ class ChangeFeedMetadata
...
+        /// Used to distinguish explicit deletes (e.g. via DeleteItem) from deletes caused by TTL expiration (a collection may define time-to-live policy for documents).
         /// </summary>
         [JsonProperty(PropertyName = ChangeFeedMetadataFields.TimeToLiveExpired, NullValueHandling = NullValueHandling.Ignore)]
         public bool IsTimeToLiveExpired { get; internal set; }
+
+        /// <summary>
+        /// Applicable for delete operations only, otherwise null.
+        /// The id of the previous item version. 
+        /// </summary>
+        [JsonProperty(PropertyName = ChangeFeedMetadataFields.Id, NullValueHandling = NullValueHandling.Ignore)]
```
kirankumarkolli, line 62: Its only addresses for Newtonsoft right?

kirankumarkolli: Closed offline to use a different approach

PR #4995, opened 2025-02-04
Open Telemetry: Fixes attribute name as per OTel Guidance

sourabh1007:
# Pull Request Template

## Description

For Open Telemetry, we got new semantic conventions from Otel Community to follow. 

Hence, as part of this PR, making changes accordingly.
1. Changing prefix from `db.client.cosmosdb` to `azure.cosmosdb.client` (for attributes and metrics)
2. `db.system` => `db.system.name`
3. `db.cosmosdb.client_id` =>  `azure.cosmosdb.client.id`
4. `db.cosmosdb.connection_mode` => `azure.cosmosdb.connection.mode`
5. `db.cosmosdb.request_content_length` => `azure.cosmosdb.request.body.size`
6. `db.cosmosdb.response_content_length` =>` azure.cosmosdb.response.body.size`
7. `db.cosmosdb.regions_contacted`=> `azure.cosmosdb.contacted_regions`
8. `db.cosmosdb.row_count` => `azure.cosmosdb.row.count`
9. `db.cosmosdb.consistency_level` => `azure.cosmosdb.consistency.level`

## Type of change

- [] New feature (non-breaking change which adds functionality)

PR #4765, opened 2024-10-08
Open Telemetry: Fixes attribute name following otel convention

sourabh1007:
### What is the change?
`db.cosmosdb.operation_type `--> removed (moved to classic sdk support)
`db.cosmosdb.machine_id` --> removed (moved to classic sdk support)
`db.cosmosdb.client_id` --> ~~`db.cosmosdb.client.id`~~ it will remain same ~~(aligns with OTel messaging conventions)~~
`db.cosmosdb.request_content_length` --> ~~`db.cosmosdb.request.content_length`~~  it will remain same ~~(aligns with OTel HTTP conventions)~~
`db.cosmosdb.response_content_length` --> ~~`db.cosmosdb.request.content_length`~~  it will remain same ~~(aligns with OTel HTTP conventions)~~
`db.cosmosdb.status_code` --> `db.response.status_code`  (moved `db.cosmosdb.status_code`  to classic sdk support)
`db.cosmosdb.consistency_level` -> added
`db.cosmosdb.request_content_length_bytes` -> added for classic sdk support

### Why?

Lot of discussions are going on with Azure Core and Open Telemetry Community teams. As part of that discussion, we are making these changes as our goal is to be Open Telemetry and Azure compliant,

## Type of change

Please delete options that are not relevant.

- [] Bug fix (non-breaking change which fixes an issue)

## Closing issues

To automatically close an issue: closes #4553

PR #5472, opened 2025-11-02
Fixes attribute names as per OTel Semantic Conventions

ghost:
* `db.system.name` value `cosmosdb` => `azure.cosmosdb`
* `azure.cosmosdb.client.id` => `azure.client.id`
* `azure.cosmosdb.request.request_charge` => `azure.cosmosdb.operation.request_charge`
* `azure.cosmosdb.contacted_regions` => `azure.cosmosdb.operation.contacted_regions`
* `azure.cosmosdb.row.count` => `db.response.returned_rows`

Tested manually with both the Application Insights SDK and OpenTelemetry SDK.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionLoadBalancerCore.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionLoadBalancerCore.cs
@@ -89,14 +89,14 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.FeedManagement
                             catch (Exception e)
                             {
                                 Extensions.TraceException(e);
-                                DefaultTrace.TraceError("Partition load balancer lease add/update iteration failed");
+                                DefaultTrace.TraceError("Failed to acquire lease: Id={0}, Owner={1}, Timestamp={2}", lease.Id, lease.Owner, lease.Timestamp);
                             }
                         }
                     }
                     catch (Exception e)
                     {
                         Extensions.TraceException(e);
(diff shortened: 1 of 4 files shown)


joegoldman2:
@sourabh1007 can you please have a look? Thanks!

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryAttributeKeys.cs
```diff
@@ -84,9 +84,9 @@ public OpenTelemetryAttributeKeys(OperationMetricsOptions operationMetricsOption
         // Cosmos DB specific attributes
 
         /// <summary>
-        /// Represents the client ID for Cosmos DB.
+        /// Represents the unique identifier of the client instance.
         /// </summary>
-        public const string ClientId = "azure.cosmosdb.client.id";
+        public const string ClientId = "azure.client.id";
```
FabianMeiswinkel, line 89: I don't think we can just replace the attribute names - this would be a breakign change. Instead populating attributes with both names seems the appropriate way. Maybe we can add an opt-in to allow only populating teh new names if a customer explicitly opt-in into it for perf reasons.

joegoldman2:
@FabianMeiswinkel thank you for the review.

Is the support for OpenTelemetry already stable in this lib? As far as I know, it's not yet stable and something like `AppContext.SetSwitch("Azure.Experimental.EnableActivitySource", true);` is still necessary today.

Also, such changes have already been made in the past (https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4995 or https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4765 for example).

PR #5478, opened 2025-11-07
[Client encryption] Adds streaming JSON processing support to feed iterators

MartinSarkany:
# Pull Request Template

### What this PR does

Adds **stream-mode JSON processing for encryption feed iterators** (query, LINQ, change-feed) to `Microsoft.Azure.Cosmos.Encryption.Custom`. Consumers opt in per-call via `RequestOptions` or per-container via `Container.UseStreamingJsonProcessingByDefault()`. **Default remains Newtonsoft** — existing callers see no behavioral change.

### New public API

- `EncryptionContainerExtensions.UseStreamingJsonProcessingByDefault(Container)` — opt-in to stream mode per container
- `DecryptableItem.DisposeAsync()` — `DecryptableItem` is now `IAsyncDisposable` so `StreamDecryptableItem` can release its pooled buffer

No public API removals. No behavioral changes to existing surface.

### Dependency changes

| Package | Project | Change |
|---|---|---|
| `Microsoft.IO.RecyclableMemoryStream 3.0.1` | `Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests` | **Removed.** Used for benchmarks; replaced with `PooledMemoryStream` |
| `System.Text.RegularExpressions 4.3.1` | `Microsoft.Azure.Cosmos.Encryption.Custom` | **Removed** (unused) |

No additions.

#4678

## Type of change


Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
```diff
@@ -38,8 +38,8 @@
   <ItemGroup>
      <PackageReference Include="Microsoft.Data.Encryption.Cryptography" Version="2.0.0-pre015" />
      <PackageReference Include="Azure.Identity" Version="1.11.4" />
-     <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
      <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.6.3" />
+    <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.1" />
```
MartinSarkany: This will be replaced with pooled streams introduced in #5479 after it is merged.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainer.cs
```diff
@@ -820,7 +831,10 @@ public override FeedIterator<T> GetChangeFeedIterator<T>(
                     changeFeedStartFrom,
                     changeFeedMode,
                     changeFeedRequestOptions),
-                this.ResponseFactory);
+                this.ResponseFactory,
+                this.Encryptor,
+                this.CosmosSerializer,
+                changeFeedRequestOptions?.GetJsonProcessor(this.DefaultJsonProcessor) ?? this.DefaultJsonProcessor);
```
juraj-blazek: nit: the last `?? this.DefaultJsonProcessor` should not be needed, the extension method already returns the fallback processor if no override is found in the request options

MartinSarkany: fixed

juraj-blazek:
Review findings:

1. **High** - `Microsoft.Azure.Cosmos.Encryption.Custom\src\Transformation\StreamProcessor.Decryptor.cs:452` / `:467`: streaming decrypt matches encrypted property names and `_ei` by name at any depth. Since encryption paths are root-only, nested properties like `{ "nested": { "Sensitive": "plain" } }` can be treated as encrypted when `_ei` lists `/Sensitive`, causing decrypt failures/data corruption; nested `_ei` user data can also be stripped. Add `reader.CurrentDepth == 1` guards for encrypted-path matching and `_ei` skipping, plus nested-property tests.

2. **High** - `Microsoft.Azure.Cosmos.Encryption.Custom\src\StreamDecryptableItem.cs:54-78`: if decryption succeeds but `cosmosSerializer.FromStream<T>` throws, `decryptedStream` is never disposed because disposal only happens after successful deserialization. In stream mode this can leave a `ReadOnlyBufferWriterStream`/rented buffer containing plaintext undisposed. Track `decryptedStream` outside the inner try and dispose it in a `finally`/catch before wrapping the exception.

3. **Medium** - `Microsoft.Azure.Cosmos.Encryption.Custom\src\EncryptionProcessor.cs:414-421` + `Microsoft.Azure.Cosmos.Encryption.Custom\src\DecryptableFeedResponse.cs:58-60`: stream-mode feed responses return lazy `StreamDecryptableItem`s backed by pooled streams, but the response does not own/cascade disposal and public examples do not dispose items. If callers skip items or stop early, rented buffers are not returned/cleared. Either avoid exposing pooled streams through the lazy public item path, or provide/enforce a response/item disposal pattern and update examples accordingly.

MartinSarkany:
> Review findings:
> 
> 1. **High** - `Microsoft.Azure.Cosmos.Encryption.Custom\src\Transformation\StreamProcessor.Decryptor.cs:452` / `:467`: streaming decrypt matches encrypted property names and `_ei` by name at any depth. Since encryption paths are root-only, nested properties like `{ "nested": { "Sensitive": "plain" } }` can be treated as encrypted when `_ei` lists `/Sensitive`, causing decrypt failures/data corruption; nested `_ei` user data can also be stripped. Add `reader.CurrentDepth == 1` guards for encrypted-path matching and `_ei` skipping, plus nested-property tests.
> 2. **High** - `Microsoft.Azure.Cosmos.Encryption.Custom\src\StreamDecryptableItem.cs:54-78`: if decryption succeeds but `cosmosSerializer.FromStream<T>` throws, `decryptedStream` is never disposed because disposal only happens after successful deserialization. In stream mode this can leave a `ReadOnlyBufferWriterStream`/rented buffer containing plaintext undisposed. Track `decryptedStream` outside the inner try and dispose it in a `finally`/catch before wrapping the exception.
> 3. **Medium** - `Microsoft.Azure.Cosmos.Encryption.Custom\src\EncryptionProcessor.cs:414-421` + `Microsoft.Azure.Cosmos.Encryption.Custom\src\DecryptableFeedResponse.cs:58-60`: stream-mode feed responses return lazy `StreamDecryptableItem`s backed by pooled streams, but the response does not own/cascade disposal and public examples do not dispose items. If callers skip items or stop early, rented buffers are not returned/cleared. Either avoid exposing pooled streams through the lazy public item path, or provide/enforce a response/item disposal pattern and update examples accordingly.

All should be fixed now.

kundadebdatta (commented):
## Deep Review — `[Client encryption] Adds streaming JSON processing support to feed iterators` @ `e09e0bd`

**Overall assessment: in good shape.** The major concerns raised earlier by @juraj-blazek (depth‑1 guard for `_ei`/encrypted-path matching, `decryptedStream` disposal on `FromStream` failure, pooled-stream cascade through `DecryptableFeedResponse`) all hold up at head. No correctness blockers found in the streaming decrypt path itself.

Cross-PR interaction check: PRs #5780 / #5829 / #5870 / #5844 / #5920 all touch the core `Microsoft.Azure.Cosmos` client; this PR is entirely contained in `Microsoft.Azure.Cosmos.Encryption.Custom`. No file or surface overlap.

Findings below in severity order. None are merge-blockers; 1–3 are worth addressing before merge.

---

### 🟡 1. `DecryptableFeedResponse.DisposeAsync` aborts cascade on first item exception (buffer leak)

**File:** `Microsoft.Azure.Cosmos.Encryption.Custom/src/DecryptableFeedResponse.cs:87-96`

```csharp
if (this.Resource != null)
{
    foreach (T item in this.Resource)
    {
        if (item is IAsyncDisposable asyncDisposable)
        {
            await asyncDisposable.DisposeAsync().ConfigureAwait(false);
        }
    }
}
```

If any one item''s `DisposeAsync` throws (e.g., a `StreamDecryptableItem` whose `SemaphoreSlim.WaitAsync` observes cancellation), the `foreach` is aborted and the remaining items never have their pooled buffers returned. The class''s own XML doc explicitly says callers "rely on this cascade to release those buffers" — that contract is silently violated.

**Suggested fix:** Capture per-item exceptions, dispose all items, then aggregate/rethrow:

```csharp
List<Exception> errors = null;
foreach (T item in this.Resource)
{
    if (item is IAsyncDisposable asyncDisposable)
    {
        try { await asyncDisposable.DisposeAsync().ConfigureAwait(false); }
        catch (Exception ex) { (errors ??= new()).Add(ex); }
    }
}
if (errors != null) throw new AggregateException(errors);
```

---

### 🟡 2. `StreamDecryptableItem` never disposes its `SemaphoreSlim` (handle leak)

**File:** `Microsoft.Azure.Cosmos.Encryption.Custom/src/StreamDecryptableItem.cs:30, 252-282`

```csharp
private readonly SemaphoreSlim asyncLock = new (1, 1);
...
public override async ValueTask DisposeAsync()
{
    await this.asyncLock.WaitAsync().ConfigureAwait(false);
    try { ... }
    finally { this.asyncLock.Release(); }
    GC.SuppressFinalize(this);
}
```

`SemaphoreSlim` allocates a kernel `ManualResetEvent` lazily on first contended wait — that handle is only released by `Dispose()`. A query returning 1000 docs creates 1000 `StreamDecryptableItem`s; under any contention these accumulate OS wait handles until GC.

**Suggested fix:** after the `finally` block, call `this.asyncLock.Dispose();`.

---

### 🟡 3. Over-broad `NotSupportedException` catch silently triggers fallback for unrelated errors

**File:** `Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs:456-475`

```csharp
try
{
    return await MdeEncryptionProcessor.DecryptJsonArrayStreamInPlaceAsync(
        content, encryptor, CosmosDiagnosticsContext.Create(null), cancellationToken);
}
catch (NotSupportedException)
{
    content.Position = 0;
    return await DecryptJsonArrayNewtonsoftAsync(content, encryptor, cancellationToken);
}
```

The catch is intended for the seekability check at `StreamProcessor.Decryptor.cs:45-48`. But `NotSupportedException` is **also** thrown from `DecryptStreamAsync` for `EncryptionFormatVersion != Mde` (`StreamProcessor.Decryptor.cs:333`) and from `MdeEncryptor.GetAdapter` (`MdeEncryptionProcessor.cs:178`) for an unsupported processor enum. Both would now be silently retried on Newtonsoft.

**Suggested fix:** Pre-validate seekability up-front and let the stream path''s exceptions propagate, or introduce a dedicated `StreamNotSupportedException`.

---

### 🟢 4. Fallback path is unreachable for its only legitimate trigger

**File:** `Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs:471`

```csharp
catch (NotSupportedException)
{
    content.Position = 0;   // <- throws NotSupportedException on non-seekable streams
    return await DecryptJsonArrayNewtonsoftAsync(content, encryptor, cancellationToken);
}
```

The `NotSupportedException` this catch was designed to handle is raised exactly when `!content.CanSeek`. Setting `Position = 0` on a non-seekable stream throws another `NotSupportedException`, masking the original and breaking the fallback. In practice Cosmos returns seekable `MemoryStream` content, so this never trips today, but the safety net doesn''t catch the case it advertises. Combined with finding 3 if you take the pre-validate approach this disappears.

---

### 🟢 5. `_ei` skip is fragile if the value isn''t a JSON object

**File:** `Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs:469-477` (set) and `:403-407` (reset)

```csharp
else if (reader.ValueTextEquals(this.encryptionPropertiesNameBytes))
{
    if (!reader.TrySkip())
    {
        isIgnoredBlock = true;       // <- assumes the value is an object
    }
    break;
}
...
if (isIgnoredBlock && reader.CurrentDepth == 1 && tokenType == JsonTokenType.EndObject)
{
    isIgnoredBlock = false;          // <- only resets on EndObject
    continue;
}
```

The current serializer always emits `_ei` as a JSON object, but the reader makes no such guarantee. If a forged/corrupt response had `_ei` as an array or a primitive, `isIgnoredBlock` would never reset (no depth‑1 `EndObject`) and the *remainder of the document* would be silently dropped — including the closing `}`, which would then corrupt downstream JSON state.

**Suggested fix:** Validate that the next non-skipped token is `StartObject` before entering ignore mode, or also reset on `EndArray`/primitives at depth 1. Add a unit test covering `_ei: []` and `_ei: "foo"`.

---

### 🟢 6. `StreamDecryptableItem` caches the deserialized item as `object` — second `GetItemAsync<U>` cast throws opaquely

**File:** `Microsoft.Azure.Cosmos.Encryption.Custom/src/StreamDecryptableItem.cs:34-90`

```csharp
private object cachedItem;
...
if (this.isDecrypted)
{
    return ((T)this.cachedItem, this.cachedDecryptionContext);
}
```

After a successful `GetItemAsync<Foo>()`, the stream is disposed and only the boxed `Foo` remains. A later `GetItemAsync<Bar>()` does `(Bar)this.cachedItem` — throws `InvalidCastException` with no context. The previous `DecryptableItemCore` (Newtonsoft) didn''t have this restriction since it cached the `JToken`. Subtle behavioural divergence between the two processors.

**Suggested fix:** Store the first-materialized type and throw a descriptive exception on mismatch, or store enough state to re-deserialize.

---

### 🟢 7. Missing `ConfigureAwait(false)` on several awaits in `netstandard2.0` paths

**Files:**
- `Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionFeedIterator.cs:42, 50`
- `Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainer.cs:1054, 1059`
- `Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs:399, 401, 463, 467, 473`

Several awaits lack `ConfigureAwait(false)` while the rest of the file/PR uses it. The library targets `netstandard2.0` consumers (ASP.NET classic / WPF SynchronizationContext) where capturing the context can deadlock or cause unnecessary thread switches.

---

### 💬 8. Observation: `UseStreamingJsonProcessingByDefault` is a write-only, one-shot toggle with no synchronization story

**File:** `Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainer.cs:19, 1028-1031` + `EncryptionContainerExtensions.cs:31-47`

- Mutable state, no opposite method, no idempotency contract documented.
- Reads in `GetItemQueryIterator` etc. aren''t synchronized with the setter (enum writes are atomic in .NET so this is observably benign today, but the design is awkward).
- The extension on `Container` returns the same `Container` reference after mutation — easy to misuse (caller may assume a wrapping pattern).

Not a blocker. Consider an options/builder shape or constructor-injected default if revisiting the public API.

---

### 💬 9. Observation: `TryExtractEncryptionProperties` re-parses each captured object

**File:** `Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs:280-294, 173`

For every document the streaming path first buffers the entire object, then `JsonSerializer.Deserialize<EncryptionPropertiesWrapper>(...)` *just to check for `_ei`*, then re-parses with `Utf8JsonReader` to do the actual transform. Two full passes per document on the supposedly "stream-fast" path. A single targeted pre-scan via `Utf8JsonReader` for the `_ei` property at depth 1 would avoid the second deserialize. Performance, not correctness — but a worthwhile follow-up.

---

## Recommendation

**No blocking findings.** The two highest-impact items raised in the prior review thread are clearly addressed. Recommend tightening before merge:

- **Finding 1** (cascade aborts on first failure) — small change, real leak.
- **Finding 2** (SemaphoreSlim leak) — one-line fix.
- **Finding 3** (overly-broad catch) — tighten to avoid silent fallback for unrelated errors.

Findings 4–9 are nice-to-haves and can land here or as a small follow-up. Otherwise approve-with-nits.

adamnova:
Picking this up while @MartinSarkany is on leave. New commit `6e0d6d5` addresses one finding from an adversarial pass and strengthens the disposal-contract docs/changelog around the fixes Martin already made.

**Adversarial-review verdict on the existing fixes** (HEAD `e09e0bd`):

| Prior finding | File | Verdict |
|---|---|---|
| Redundant `?? this.DefaultJsonProcessor` | `EncryptionContainer.cs:837` | ✅ Fixed. `GetJsonProcessor` is null-safe so the dropped `?.` is fine. |
| Depth guard for encrypted-path and `_ei` matching | `StreamProcessor.Decryptor.cs:452` | ✅ Fixed. `reader.CurrentDepth == 1` gates both encrypted-path lookup and `_ei` skip. `EncryptionPropertiesStreamReader` also scans `_ei` only at depth 1. |
| `decryptedStream` leak when `FromStream<T>` throws | `StreamDecryptableItem.cs:56-79` | ✅ Fixed. Aliasing check + inner `finally` dispose; `StreamDecryptableItemTests.GetItemAsync_WhenSerializerThrowsAfterDecryption_DisposesDecryptedStream` covers it. |
| Lazy `StreamDecryptableItem`s leak if caller skips | `DecryptableFeedResponse.cs` | ✅ Cascade works. The compile-time return type is still `FeedResponse<T>` — callers must runtime-cast to `IAsyncDisposable`. Locked in by `DecryptableFeedResponseTests.DisposeAsync_ResponseIsCastableToIAsyncDisposable` and the updated example on `DecryptableItem`. Strengthened the XML docs on `UseStreamingJsonProcessingByDefault` / `DecryptableItem.DisposeAsync` to call this out at the opt-in surface (this commit). |

**One NEW issue this commit fixes** (HIGH — orphan-leak narrower than #4 above):

`EncryptionProcessor.ConvertResponseToDecryptableItemsStreamAsync` accumulates `StreamDecryptableItem`s in a local `List<DecryptableItem>` while consuming `JsonArrayStreamSplitter.SplitIntoSubstreamsAsync`. If the splitter threw **after** yielding one or more documents (mid-feed transport error, cancellation, malformed payload past the first doc), every already-appended `StreamDecryptableItem` owned a `PooledMemoryStream` rented from `ArrayPool<byte>.Shared`. Because the partial list never reaches the caller and is never wrapped in a `DecryptableFeedResponse`, the disposal cascade addressed by the previous review round never sees those items. The rentals stay out of the pool (eventual GC reclamation only) and may retain plaintext residue.

The fix wraps the `await foreach` in `try`/`catch`, drains any partial items via `IAsyncDisposable` (swallowing per-item failures so we don't mask the original cause), then rethrows the original exception unchanged. `EncryptionProcessorTests.ConvertResponseToDecryptableItemsAsync_Stream_WhenSplitterThrowsMidFeed_PreservesOriginalException` feeds a `ThrowAfterPrefixStream` containing one complete document followed by an `IOException` and asserts the original `IOException` identity is preserved through the catch/rethrow.

Also added the missing changelog entry for #5478 under `1.0.0-preview09`.

**Local validation**: `net8.0` unit-test suite `689` passed / `0` failed (previously `688` — adds my new regression test). Encryption.Custom build clean (0 warnings).

**CI**: The two failures on the previous build (`EmulatorTests Release - MultiRegion` / `MultiMaster`) are core-SDK MultiRegion/MultiMaster tests; this PR touches only files under `Microsoft.Azure.Cosmos.Encryption.Custom` and the `Encryption.Custom EmulatorTests Release` job passed. Treating those failures as flaky/infra unrelated to this PR. Re-triggering with `/azp run` below — if MultiRegion/MultiMaster still fail on the rerun and the failure modes are not encryption-related, they should be ignored for merge gating per the standard flaky-test handling.

/azp run

adamnova:
Second-pass review verdict (commit `38e6e38`). Independently verified every finding before fixing, applied four code-changing fixes (one HIGH regression, one HIGH cascade-leak, two MEDIUMs), one doc-strengthening, and added three regression tests.

| # | Sev | Finding | Verdict | Action |
|---|---|---|---|---|
| **H-1** | HIGH | `StreamDecryptableItem` loses DEK id in `EncryptionException` (regression vs `DecryptableItemCore`) | ✅ CONFIRM | **Fixed.** Catch block now extracts the DEK id from `DecryptionContext` (if decrypt succeeded then serializer threw) or from on-stream `_ei` via `EncryptionPropertiesStreamReader.ReadAsync`. Falls back to `string.Empty` only when both sources are unavailable (preserves the original no-throw-in-catch behavior). Test: `GetItemAsync_WhenDecryptionFails_PopulatesDataEncryptionKeyIdOnException`. |
| **H-2** | HIGH | `DecryptableFeedResponse.DisposeAsync` aborts cascade on first throw | ✅ CONFIRM | **Fixed.** Per-item `try`/`catch`; single failure rethrown via `ExceptionDispatchInfo.Capture` (preserves identity; existing `DisposeAsync_PropagatesExceptionFromItemDispose` still green); multiple failures surface as `AggregateException`. `isDisposed` tightened to `int` + `Interlocked.Exchange` for the L-1 concurrent-disposer race. Tests: `DisposeAsync_WhenItemThrows_StillDisposesRemainingItems`, `DisposeAsync_WhenMultipleItemsThrow_AggregatesAndStillDrains`. |
| **H-3** | HIGH | Missing changelog entry | ✅ CONFIRM | **Done in `6e0d6d5`.** Extended in `38e6e38` with three #### Fixes entries for the H-1/H-2/M-2 fixes. |
| **H-4** | HIGH | No unit test exercises actual MDE stream-decryption path through `StreamDecryptableItem` | ✅ CONFIRM | **Will file as a follow-up issue.** Out of scope for this commit — addressing it cleanly requires either a fake MdeStreamProcessor or a coverage hook around the algorithm dispatch in `SystemTextJsonStreamAdapter`. The new H-1 test does exercise the MDE encrypt → stream-decrypt-attempt → MDE-failure path, but that's a partial credit, not full coverage. |
| **M-1** | MED | `EncryptionContainer.DefaultJsonProcessor` mutation racy and one-way | ✅ CONFIRM | **Doc-only fix.** Added XML doc note on `EncryptionContainer.UseStreamingJsonProcessingByDefault()` calling out the configure-once-before-use contract and the undefined behavior of in-flight mutation. Skipping `Interlocked.Exchange` here intentionally — for a configure-once setter, adding atomics misleads callers into thinking mid-operation mutation is supported; the right contract is "set during setup, before any encryption op". Matches the `PooledStreamConfiguration` pattern in the same package. |
| **M-2** | MED | `StreamProcessor` per-document buffer grows unbounded | ✅ CONFIRM | **Fixed.** Both `StreamProcessor.Decryptor` (line 368-378) and `StreamProcessor.Encryptor` (line 68-77) now enforce the same `MaxBufferSize = 64 MiB` cap that `JsonArrayStreamSplitter` / `JsonFeedStreamHelper` already enforce, throwing the same `InvalidOperationException` style. Note: confirmed via `git log -L` that the unbounded path was introduced by this PR (commit `0779f5130` "Replace RecyclableMemoryStream with ArrayPool-backed streams"), so this isn't a pure pre-existing concern — appropriate to fix here. |
| **M-3** | MED | Misleading `HandleLeftOver` error message | ✅ CONFIRM | **Fixed.** Reworded to "JSON document or token does not fit within the maximum buffer size of {N} bytes" — trip condition is no-progress on the chunk, which is most often the document not fitting, not strictly a token. M-2 fixes use the same wording for consistency. |
| **M-4** | MED | `UseStreamingJsonProcessingByDefault` return type is `Container` | ⚠️ ACKNOWLEDGE | **Skipping.** Reviewer correctly notes this can't be cleanly improved today (`EncryptionContainer` is `internal sealed`); raising the surface would require a wider public-API change. Worth a follow-up if the customization surface grows. |
| **L-1** | LOW | `isDisposed` not atomic | ✅ CONFIRM | **Fixed alongside H-2** — see above. |
| **L-2** | LOW | PR description has empty checkbox `- []` | ✅ CONFIRM | **Can't fix from my account** — editing the base-repo PR body needs maintainer access I don't have (my token has `repo` scope but not write on `Azure/azure-cosmos-dotnet-v3`). Cosmetic only; please flip `- []` → `- [x]` when someone with the perm does the merge pass. |
| **L-3** | LOW | Doc example forces customers to manually cast to `IAsyncDisposable` | ⚠️ ACKNOWLEDGE | **Skipping.** The reviewer's own suggestion (`throw if not IAsyncDisposable`) would break the inheritance contract (`DecryptableItem` base could legitimately not need disposal). Strengthened the XML doc in `6e0d6d5` to make the contract loud at the opt-in surface; further hardening is an SDK-level concern. |
| **L-4** | LOW | `DecryptContentStreamAsync` drops `CancellationToken` | ⚠️ ACKNOWLEDGE | **Skipping** — same limitation as `DecryptableItemCore`, would require a new public overload on `DecryptableItem` (worth a follow-up issue). |
| **L-5** | LOW | Two `EncryptionFeedIterator` ctor overloads with same arity | ⚠️ ACKNOWLEDGE | **Skipping** — small ergonomic concern, removing the unused overload is a non-trivial refactor for the value. Filed in my head for follow-up. |
| **L-6** | LOW | Duplicate-`Documents`-property test pinning unusual semantics | ⚠️ ACKNOWLEDGE | **Skipping** — a code comment would be nice but pinning the behavior is intentional defense-in-depth for backend payloads. |

**Local validation**: `692` passed / `0` failed on `net8.0` (up from `689` after the three new regression tests). Build clean (0 warnings, 0 errors).

**Status**: PR HEAD is now `38e6e38`. Still waiting for an authorized user (e.g. @kundadebdatta) to comment `/azp run` to re-trigger the `dotnet-v3-ci` pipeline on the new SHA — my comment alone is silently ignored by the bot, same restriction Martin's PR always had. Will file the H-4 coverage-gap and L-4 cancellation-token concerns as follow-up issues separately so they don't block this merge.

/azp run

adamnova:
H-4 and L-4 follow-up fixes pushed in commit `b62e01e`. Both deferred items now resolved in-PR.

| # | Status | Summary |
|---|---|---|
| **L-4** `CancellationToken` plumbed through stream-mode decrypt | ✅ Fixed | New public virtual `DecryptableItem.GetItemAsync<T>(CancellationToken)` overload. Default impl honors pre-call cancellation and delegates to `GetItemAsync<T>()` so out-of-package subclasses are binary-compat. `StreamDecryptableItem` overrides it to plumb the token through `SemaphoreSlim.WaitAsync` and `EncryptionProcessor.DecryptAsync`. `DecryptableItemCore` overrides it to forward to the existing `EncryptionProcessor.DecryptAsync(JObject, ...)` overload (which already took a CT param that the parameterless caller was hard-coding to `default`). `OperationCanceledException` is rethrown unwrapped on both paths. The parameterless `GetItemAsync<T>()` now delegates to `GetItemAsync<T>(default)` so the impl lives in one place. Tests: `GetItemAsync_WithCancellationToken_PreCancelled_Throws` + `_NotCancelled_DecryptsNormally` + `GetItemAsync_DefaultOverload_DelegatesToCancellationTokenOverload`. Public API surface contracts (net6 + net8) updated; diff is purely additive. |
| **H-4** MDE stream-decrypt path coverage | ✅ Fixed | `GetItemAsync_StreamMode_WithMdeAlgorithm_TakesGenuineMdeStreamPath` encrypts with `MdeAeadAes256CbcHmac256Randomized` via `TestEncryptorFactory.CreateMde` (the only mock that supplies the `DataEncryptionKey` methods the MDE stream encryptor needs — `EncryptByteCount`/`DecryptByteCount`/`EncryptData`/`DecryptData`), then asserts via an `ActivityListener` on the `Microsoft.Azure.Cosmos.Encryption.Custom` source that the `ScopeDecryptModeSelectionPrefix+JsonProcessor.Stream` scope WAS created AND that the `ScopeDecryptModeSelectionPrefix+JsonProcessor.Newtonsoft` scope was NOT — locking in "we actually take the MDE/Stream branch", not the legacy fallback the existing tests silently exercised. |

**Validation**: `net8.0` unit-test suite `696` passed / `0` failed. Build clean (0 warnings, 0 errors).

**Status**: PR HEAD is now `b62e01e`. Still pending an authorized user to comment `/azp run` to re-trigger CI on the new SHA.

/azp run

adamnova:
Reverting **L-4** (the `GetItemAsync<T>(CancellationToken)` overload) and force-pushing — your pushback on the default impl was correct, and looking at where the token could actually fire, the value is near zero:

**Where CT could observe cancellation inside `GetItemAsync<T>`:**
- `SemaphoreSlim.WaitAsync(CT)` — only under contention on the same item (rare)
- `encryptor.GetEncryptionKeyAsync(CT)` — only if the customer's DEK provider does remote I/O and the DEK isn't cached
- `encryptor.DecryptAsync(CT)` — only if the customer's encryptor is slow

**Where CT cannot fire** (the bulk of the work):
- `PooledMemoryStream.ReadAsync` — in-memory, ignores CT
- `StreamProcessor.TransformDecryptBuffer` per-byte inner loop — purely synchronous, no `ThrowIfCancellationRequested`
- `Utf8JsonWriter` / `Utf8JsonReader` — sync
- `cosmosSerializer.FromStream<T>` — sync
- Newtonsoft path's `JObject` traversal — sync

For the typical case (page already fetched, DEK cached, small-to-medium documents) `GetItemAsync<T>` is sub-millisecond sync CPU work and CT will never fire even with proper plumbing. Customers who want fail-fast can already do `ct.ThrowIfCancellationRequested(); await item.GetItemAsync<T>();` from their side. The original L-4 rationale ("important for the stream path where decryption can be slow for large documents") doesn't survive scrutiny — the per-byte loop is exactly where slowness happens, and it's CT-deaf.

**Cost vs benefit**: API-surface expansion + breaking change for external `DecryptableItem` subclassers (when made abstract) for a CT that fires only in the HSM-slow-DEK-fetch corner case. Not worth it.

Reverted L-4 in commit `5cecb2b33` (force-pushed over `b62e01e63`). Kept the H-4 MDE-path coverage test — that one is standalone and lands real coverage on a path that was silently falling through to the Newtonsoft fallback. The H-4 commit message is now test-only.

**Validation**: `693` passed / `0` failed on `net8.0` after the revert (was 696; -3 for the dropped L-4 regression tests, kept the +1 H-4 test). Build clean.

**Final PR HEAD**: `5cecb2b33`. Still pending an authorized commenter for `/azp run`.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -14,18 +14,285 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom.Transformation
...
+
+            Span<byte> destination = objectBuffer.GetSpan(segment.Length);
+            segment.CopyTo(destination);
+            objectBuffer.Advance(segment.Length);
+        }
+
+        private static async Task OverwriteStreamAsync(Stream stream, Stream source, CancellationToken cancellationToken)
+        {
+            stream.Position = 0;
+            stream.SetLength(0);
```
kundadebdatta: **🟡 Recommendation** · Reliability: Data Loss On Failure

**`OverwriteStreamAsync` truncates the input stream before writing — cancellation or IO failure leaves the response body irrecoverably corrupted**

`SetLength(0)` permanently destroys the original ciphertext **before any byte of the decrypted payload is written**. If `CopyToAsync` throws (cancellation request between buffer writes, IO failure, OOM), the caller in `EncryptionProcessor.DecryptJsonArrayStreamAsync` calls `MdeEncryptionProcessor.DecryptJsonArrayStreamInPlaceAsync(content, ...)` which returns `input` — i.e., the now-truncated `content`. That same reference is wrapped in `DecryptedResponseMessage` and returned up the stack. Downstream consumers see either an empty body or a partial decrypted prefix, with no signal that anything went wrong other than the propagated exception (which a careless `catch` could swallow). The original encrypted payload cannot be recovered for retry.

The typical input is a `MemoryStream` (no IO failure), but `CancellationToken` cancellation during `CopyToAsync` is realistic for long pages, and the resulting state is silently inconsistent. The whole point of buffering into `tempOutputStream` is to avoid mid-stream corruption — but the final copy step throws that guarantee away.

**Suggested fix:** Materialize `tempOutputStream` to a `byte[]` first and then write under a try/catch that restores the buffer on failure; or simply return `tempOutputStream` and let the caller replace its reference (`EncryptionProcessor` already discards the input).

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/StreamDecryptableItem.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        public override async Task<(T, DecryptionContext)> GetItemAsync<T>()
+        {
+            await this.asyncLock.WaitAsync().ConfigureAwait(false);
+            try
+            {
+                ObjectDisposedException.ThrowIf(this.isDisposed, this);
+
+                // Return cached value if already decrypted
+                if (this.isDecrypted)
```
kundadebdatta, line 49: **🟡 Recommendation** · Reliability: Missing Cancellation

**`StreamDecryptableItem.GetItemAsync` and `DisposeAsync` ignore caller cancellation — decryption cannot be aborted**

The base `DecryptableItem.GetItemAsync<T>()` abstract signature does not take a `CancellationToken`, so this implementation passes `cancellationToken: default` to `EncryptionProcessor.DecryptAsync`. That path includes a remote DEK fetch (`Encryptor.FetchDataEncryptionKeyWithoutRawKeyAsync`) which can do KMS / Key Vault round-trips. A stuck DEK fetch — or a misbehaving encryptor — cannot be cancelled by the application; the only escape is process teardown. Likewise the `SemaphoreSlim.WaitAsync()` on the async lock here (and in `DisposeAsync`) cannot be cancelled.

Streaming mode is positioned for ""hot-path workloads""; those workloads typically use overall request-deadline cancellation. The streaming path silently loses that capability vs. the Newtonsoft path (which is sync and benefits from the existing CT cascade in `DecryptableItemCore`).

**Suggested fix:** Add a CT-accepting virtual overload on `DecryptableItem` (e.g. `GetItemAsync<T>(CancellationToken)`) — the type is already being extended in this PR with `IAsyncDisposable`, so a new optional virtual is acceptable — then thread the token through `asyncLock.WaitAsync(ct)` and `EncryptionProcessor.DecryptAsync(..., ct)`.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/StreamDecryptableItem.cs
```diff
@@ -0,0 +1,199 @@
...
+                    // Falls back to string.Empty rather than null because EncryptionException's ctor rejects null DataEncryptionKeyId.
+                    string dataEncryptionKeyId = decryptionContextForDiagnostics?.DecryptionInfoList?.FirstOrDefault()?.DataEncryptionKeyId
+                        ?? await this.TryReadDataEncryptionKeyIdAsync().ConfigureAwait(false)
+                        ?? string.Empty;
+
+                    string encryptedContent = await this.TryReadStreamAsStringAsync().ConfigureAwait(false);
+
+                    await this.DisposeContentStreamAsync().ConfigureAwait(false);
+
+                    throw new EncryptionException(dataEncryptionKeyId: dataEncryptionKeyId, encryptedContent: encryptedContent ?? string.Empty, exception);
```
kundadebdatta, line 104: **🟡 Recommendation** · Reliability: Inconsistent Error State

**Failure path leaves the instance ""not disposed, not decrypted, but unusable"" — second call NREs and masks the original error**

After this `throw` fires, `contentStream` has been nulled by `DisposeContentStreamAsync()` just above, but `isDisposed` stays `false` and `isDecrypted` stays `false`. A subsequent `GetItemAsync<T>()` will:

1. Pass the `ObjectDisposedException.ThrowIf` guard,
2. Skip the ""already decrypted"" cache,
3. Call `DecryptContentStreamAsync()` — which forwards `contentStream == null` to `EncryptionProcessor.DecryptAsync` → `MdeEncryptionProcessor.DecryptAsync` → `IMdeJsonProcessorAdapter.DecryptAsync(null, ...)` → NRE inside `ReadMdeEncryptionPropertiesStreamingAsync`,
4. This same catch wraps that NRE in a fresh `EncryptionException`, but `TryReadDataEncryptionKeyIdAsync()` returns `null` (contentStream is null), `TryReadStreamAsStringAsync()` returns `null` too, so the user sees a second `EncryptionException` with empty `DataEncryptionKeyId` and empty `EncryptedContent` — masking the real first failure.

This is observable as a confusing UX (retry of a failed decrypt produces a different, less-useful exception than the first one). It also nullifies the recent improvements that made the diagnostic DEK-ID extraction more robust.

**Suggested fix:** Set `this.isDisposed = true` (or introduce a dedicated `this.isFaulted = true`) just before this `throw`, so subsequent calls throw `ObjectDisposedException` directly. Alternatively, cache the first `EncryptionException` and rethrow it on subsequent calls.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/MdeEncryptionProcessor.cs
```diff
@@ -131,43 +143,15 @@ public async Task EncryptAsync(
...
-        public async Task<DecryptionContext> DecryptStreamAsync(
-            Stream input,
-            Stream output,
-            Encryptor encryptor,
-            EncryptionProperties properties,
-            CosmosDiagnosticsContext diagnosticsContext,
-            CancellationToken cancellationToken)
-        {
-            return await this.StreamProcessor.DecryptStreamAsync(input, output, encryptor, properties, diagnosticsContext, cancellationToken);
+            return input;
```
kundadebdatta, line 154: **🟡 Recommendation** · Maintainability: Discarded Result

**`DecryptJsonArrayStreamInPlaceAsync` silently discards the aggregated `DecryptionContext`**

The inner `StreamProcessor.DecryptJsonArrayStreamInPlaceAsync` returns a `DecryptionContext` that aggregates `(DEK-id => set-of-decrypted-paths)` across the entire page (see `CreateAggregatedContext`). This wrapper drops it and returns only `input`.

The Newtonsoft path (`DecryptJsonArrayNewtonsoftAsync`) is similarly stream-only, so on the surface this looks symmetrical. But this is also the only place where per-page aggregated decryption context is computed — once discarded here it is irretrievable. Any future enhancement that wants to surface ""which paths/keys touched this page"" through diagnostics has nothing to consume, and today the work in `UpdateAggregatedPaths` / `CreateAggregatedContext` is effectively dead.

**Suggested fix:** Either thread the context out to `diagnosticsContext` as a scoped attribute, or remove the aggregation machinery in `StreamProcessor.Decryptor` when called from the in-place array path so we don't pay for results we throw away.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/JsonArrayStreamSplitter.cs
```diff
@@ -0,0 +1,126 @@
...
+                        buffer.AsSpan(0, dataLength),
+                        isFinalBlock,
+                        ref readerState,
+                        ref traversalState,
+                        writeSegment);
+
+                    leftOver = dataLength - result.BytesConsumed;
+                    buffer = JsonFeedStreamHelper.HandleLeftOver(buffer, dataLength, leftOver, result.BytesConsumed, MaxBufferSize);
+
+                    if (isFinalBlock && leftOver > 0)
```
kundadebdatta, line 86: **🟡 Recommendation** · Reliability: Misleading Error On Malformed Input

**Truncated/malformed JSON degrades to an inflated ""buffer too large"" error after wasteful repeated buffer growth**

When the stream is exhausted but there are leftover unparseable bytes (truncated JSON, missing closing brackets, etc.), this loop:

1. Reads 0 bytes (`bytesRead == 0`, `isFinalBlock = true`),
2. `ProcessChunk` consumes 0 bytes (incomplete token),
3. `leftOver == dataLength`, so `HandleLeftOver` rents a buffer 2x larger,
4. Resets `isFinalBlock = false` here, repeats — next iteration reads 0 again, doubles again,
5. ... until `newSize > MaxBufferSize` and `HandleLeftOver` throws `InvalidOperationException(""JSON document or token does not fit within the maximum buffer size of 67108864 bytes"")`.

A truncated 200-byte response wastes ~26 buffer-doublings (16 KB → 64 MB) and surfaces with an error message that points the operator at ""max buffer size"" when the actual cause is ""malformed/truncated JSON"". This is both wasteful and misleading. The same pattern exists in `StreamProcessor.Decryptor.cs`.

**Suggested fix:** After `HandleLeftOver`, detect the ""no progress on a final block"" condition and throw a clean `JsonException` directly:

```csharp
if (isFinalBlock && result.BytesConsumed == 0 && leftOver > 0)
{
    throw new JsonException($""Truncated or malformed Cosmos feed envelope: {leftOver} unread bytes after end of stream."");
}
```

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionFeedIterator{T}.cs
```diff
@@ -31,18 +50,49 @@ public override async Task<FeedResponse<T>> ReadNextAsync(CancellationToken canc
             if (typeof(T) == typeof(DecryptableItem))
             {
                 IReadOnlyCollection<T> resource;
-                (responseMessage, resource) = await this.feedIterator.ReadNextWithoutDecryptionAsync<T>(cancellationToken);
+                (responseMessage, resource) = await this.ReadNextWithoutDecryptionAsync(cancellationToken).ConfigureAwait(false);
```
kundadebdatta, line 53: **🟡 Recommendation** · Reliability: Buffer Pool Leak

**`ReadNextAsync` (DecryptableItem branch) leaks pooled buffers if anything between item-list construction and `CreateResponse` throws**

`ConvertResponseToDecryptableItemsStreamAsync` is careful to drain partial items on failure, but once it returns successfully here, the items are held in the local `resource` collection with **no exception-safety net**. If the subsequent `DecryptableFeedResponse<T>.CreateResponse` call throws (or an OOM occurs allocating the response wrapper, or any user-supplied subclass overrides things), every `StreamDecryptableItem` in that list still holds a rented `ArrayPool<byte>` buffer that will only be freed when GC eventually reclaims them — and only after possibly retaining plaintext residue.

The PR explicitly documents the ""buffers must be returned to the pool to clear plaintext residue"" contract in multiple places (e.g. `DecryptableItem` XML doc, `UseStreamingJsonProcessingByDefault` doc). This is a narrow but symmetric leak window that contradicts that contract.

**Suggested fix:** Wrap the construction+return in a try/catch that disposes the rented items on failure:

```csharp
try
{
    return DecryptableFeedResponse<T>.CreateResponse(responseMessage, resource);
}
catch
{
    foreach (T item in resource)
    {
        if (item is IAsyncDisposable d) await d.DisposeAsync().ConfigureAwait(false);
    }
    throw;
}
```

This mirrors the disposal pattern that already exists inside `ConvertResponseToDecryptableItemsStreamAsync`.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainerExtensions.cs
```diff
@@ -28,6 +28,43 @@ public static Container WithEncryptor(
...
+        /// caller skipped, did not enumerate, or did not call <c>GetItemAsync</c> on return their pooled buffers to
+        /// the pool. Forgetting to dispose leaks <c>ArrayPool</c> rentals and risks leaving plaintext residue in
+        /// pooled memory until GC. See the example on <see cref="DecryptableItem"/> for the recommended pattern.
+        /// </para>
+        /// </remarks>
+        public static Container UseStreamingJsonProcessingByDefault(this Container container)
+        {
+            if (container is not EncryptionContainer encryptionContainer)
+            {
+                throw new NotSupportedException($"{nameof(UseStreamingJsonProcessingByDefault)} is only supported with {nameof(EncryptionContainer)}.");
```
kundadebdatta: **🟢 Suggestion** · Style: Wrong Exception Type

**`UseStreamingJsonProcessingByDefault` throws `NotSupportedException` where `ArgumentException` is the conventional choice**

`NotSupportedException` conventionally means ""this operation is fundamentally unsupported by the implementation""; that's not what's happening here. The user passed the *wrong kind of container* — that's an invalid argument. `ArgumentException` (or `InvalidOperationException` if interpreted as ""this `Container` hasn't been wrapped by `WithEncryptor`"") is more accurate and matches how exception filters typically partition handling logic in user code.

Minor, but exception-type contracts matter for callers writing structured `catch` blocks. Every other null/argument guard in the new code (`ArgumentValidation.ThrowIfNull(content)`, `ArgumentNullException.ThrowIfNull(...)`) is consistent — this is the odd one out.

**Suggested fix:**

```suggestion
                throw new ArgumentException($""{nameof(UseStreamingJsonProcessingByDefault)} is only supported with {nameof(EncryptionContainer)}."", nameof(container));
```

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainerExtensions.cs
```diff
@@ -57,7 +94,10 @@ public static FeedIterator<T> ToEncryptionFeedIterator<T>(
 
             return new EncryptionFeedIterator<T>(
                 (EncryptionFeedIterator)encryptionContainer.ToEncryptionStreamIterator(query),
```
kundadebdatta, line 124: **🟢 Suggestion** · API Design: Inconsistent Override Surface

**LINQ `ToEncryptionFeedIterator<T>` (and `ToEncryptionStreamIterator<T>`) ignore per-call request options — only the container default is honored**

Every other entry point routes the JSON processor through `requestOptions.GetJsonProcessor(this.DefaultJsonProcessor)`, so the per-call `Properties[""encryption-json-processor""]` opt-in works uniformly. This LINQ overload (and its sibling `ToEncryptionStreamIterator<T>` a few lines below) bypasses that path entirely — there is no `RequestOptions` parameter on the public extension, so LINQ users have no way to override the processor per call. The PR description claims ""Consumers opt in per-call via `RequestOptions` or per-container"", but for LINQ users that's only half-true.

This is a surface-area inconsistency. LINQ users who set the property bag to enable the streaming path (matching what they'd do for `ReadItemAsync` / `CreateItem...`) will quietly fall back to the container default and won't get the streaming code path they thought they enabled.

**Suggested fix:** Either add an optional `RequestOptions` parameter on both LINQ extensions and route through `requestOptions.GetJsonProcessor(encryptionContainer.DefaultJsonProcessor)`, or document the limitation prominently on the LINQ extension XML doc.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/DecryptableFeedResponse.cs
```diff
@@ -7,9 +7,13 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom
     using System;
     using System.Collections.Generic;
     using System.Net;
+    using System.Threading;
+    using System.Threading.Tasks;
 
-    internal class DecryptableFeedResponse<T> : FeedResponse<T>
+    internal class DecryptableFeedResponse<T> : FeedResponse<T>, IAsyncDisposable
```
kundadebdatta, line 13: **💬 Observation** · API Design: Foot-gun Disposal Contract

**`DecryptableFeedResponse<T>` advertises `IAsyncDisposable` only at runtime — `await foreach` / `await using` callers will silently leak**

The class implements `IAsyncDisposable`, but the public `FeedIterator<T>` return type is `FeedResponse<T>`, which does not. The PR documents this in `UseStreamingJsonProcessingByDefault`'s XML comment (""Callers MUST cast each page to `IAsyncDisposable`...""), and the test code does exactly that (`if (readDocsLazily is IAsyncDisposable disposable) await disposable.DisposeAsync();`). But:

- C# `await using` won't compile against `FeedResponse<T>` (base type implements only `IDisposable`), so the natural disposal idiom does nothing.
- `await foreach` cleanup relies on `IAsyncDisposable` known at compile time; it won't trigger either.
- Any caller who reads the existing Cosmos `FeedIterator<T>` documentation will not learn about this contract.

This is an opt-in feature where the failure mode is silent buffer leak + potential plaintext residue in pooled memory. The ""you must cast and dispose"" contract is easy to forget — especially because the base `FeedResponse<T>` looks complete.

Marked as observation (not recommendation) because fixing it cleanly likely requires a return-type change — e.g., introducing a derived `DisposableFeedResponse<T>` that compile-time implements `IAsyncDisposable` and is returned from the streaming-decryption iterators. Worth a deliberate design decision before this stabilizes into a non-preview surface.

---

kushagraThapar:
**Changelog gap — `Microsoft.IO.RecyclableMemoryStream` removal is unannounced**

Commit `0779f5130` ("Replace RecyclableMemoryStream with ArrayPool-backed streams") removes both the `Microsoft.IO.RecyclableMemoryStream 3.0.1` `PackageReference` from `Microsoft.Azure.Cosmos.Encryption.Custom.csproj` and the corresponding reference from `Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests.csproj`. The `1.0.0-preview09` entry in `changelog.md` mentions the `Microsoft.Extensions.Caching.Memory → Abstractions` swap (line 38) but does not mention the `RecyclableMemoryStream` removal.

Evidence:
```
grep -n "RecyclableMemoryStream\|System.Text.RegularExpressions" Microsoft.Azure.Cosmos.Encryption.Custom/changelog.md
  → 0 matches
```

**Why it matters:** Transitive dependency removal. Consumers that picked up `Microsoft.IO.RecyclableMemoryStream` types through this package's dependency closure (perf-test harnesses, downstream packages that didn't declare their own ref) will silently lose the transitive flow on upgrade — build break shows up at restore time on the consumer's side, with no signal in our changelog.

**Suggested addition** under `#### Updates`, mirroring the existing `Microsoft.Extensions.Caching.Memory` entry:

> Removes the package's `Microsoft.IO.RecyclableMemoryStream` reference (previously `3.0.1`). The package's stream-pooling needs are now served by the internal `PooledMemoryStream` (`ArrayPool<byte>`-backed, with `clearOnReturn` for plaintext-residue zeroing). Consumers transitively depending on `Microsoft.IO.RecyclableMemoryStream` types **through this package** must add a direct reference.

kushagraThapar:
**Per-call streaming opt-in is not type-discoverable on the public surface**

The new `EncryptionContainerExtensions.UseStreamingJsonProcessingByDefault(Container)` extension gives the per-*container* default a typed public path, but the per-*call* opt-in (documented in `EncryptionContainer.cs:1035` remarks as `requestOptions.Properties["encryption-json-processor"] = JsonProcessor.Stream`) has no typed public path. Every typed symbol it composes is `internal`:

| Surface | Visibility |
|---|---|
| `UseStreamingJsonProcessingByDefault(Container)` (container default) | ✅ public |
| `JsonProcessor` enum (`src/JsonProcessor.cs:10`) | ❌ `internal` |
| `JsonProcessorRequestOptionsExtensions` (`src/JsonProcessorRequestOptionsExtensions.cs:14`) | ❌ `internal` |
| `JsonProcessorPropertyBagKey` constant (`"encryption-json-processor"`) | ❌ `internal` |
| `EncryptionItemRequestOptions.JsonProcessor` typed property | ❌ does not exist |

Evidence:
```
grep -rnE 'public.*JsonProcessor|JsonProcessor.*public' Microsoft.Azure.Cosmos.Encryption.Custom/src
  → 0 hits where JsonProcessor is the public symbol
grep -nE 'JsonProcessor[^P]' tests/.../Contracts/DotNetSDKEncryptionCustomAPI.net8.json
  → 0 occurrences on the shipped public surface
```

The shipped contract has zero references to `JsonProcessor` as a typed symbol. A customer who wants per-call streaming opt-in is forced to write `requestOptions.Properties["encryption-json-processor"] = (object)1;` — magic-string key, magic-int value, mistype-detected only at runtime. Equally, an MCP tool, Roslyn-driven code-completion generator, or Copilot inline-suggestion source cannot expose "decode in streaming mode" as a typed flag.

This shape is hard to walk back once it ships in a GA preview, so worth pinning down now.

**Suggested fix** (either — option (a) is more idiomatic and matches the existing `EncryptionItemRequestOptions.EncryptionOptions` shape):

- (a) Promote `JsonProcessor` enum to `public` and add `JsonProcessor? JsonProcessorOverride { get; set; }` on `EncryptionItemRequestOptions` (and `EncryptionChangeFeedRequestOptions`).
- (b) Expose `JsonProcessorPropertyBagKey` as `public const string` plus a public static helper class with `RequestOptions.UseStreamingJsonProcessing()` / `.UseNewtonsoftJsonProcessing()` extension methods.

Happy to file as a separate API-design issue if you'd rather not block this PR on it.

adamnova:
Third-pass review verdict. Commit `1e8ba74` fixes 4 findings; 2 are refuted (one via experiment); 5 are real but deferred as design calls. Each was verified against the code before acting.

### ✅ Fixed in `1e8ba74`

| # | Sev | Finding | Fix |
|---|---|---|---|
| **Inconsistent error state** (`StreamDecryptableItem.cs:101`) | 🟡 | Confirmed real bug. After a failed decrypt the catch nulled `contentStream` but left `isDisposed`/`isDecrypted` false; a retry NRE'd in the MDE adapter and produced a second `EncryptionException` with empty diagnostics, masking the first. | Set `isDisposed = true` before throwing → retries surface a clean `ObjectDisposedException`. Test added. |
| **Buffer pool leak window** (`EncryptionFeedIterator{T}.cs:53`) | 🟡 | Confirmed. Items sit in the local `resource` list with no safety net until `CreateResponse` takes ownership; if it throws, pooled buffers leak until GC. | try/catch drain on `CreateResponse` failure, mirroring the existing converter pattern. |
| **Wrong exception type** (`EncryptionContainerExtensions.cs:59`) | 🟢 | Confirmed. | `NotSupportedException` → `ArgumentException(nameof(container))`. Test updated. |
| **Changelog dependency-removal gap** | 🟡 | Partially correct — but the named package (`Microsoft.IO.RecyclableMemoryStream`) was never in `main`'s shipped csproj (only transient on this branch). The **actual** undeclared removal is `System.Text.RegularExpressions 4.3.1`. | Added an `#### Updates` entry for `System.Text.RegularExpressions`. |

### 🚫 Refuted

**`OverwriteStreamAsync` … wait, no — the "misleading buffer-size error on truncated input"** (`JsonArrayStreamSplitter.cs:86` / `StreamProcessor.Decryptor.cs`) — **refuted by experiment.** The scenario doesn't reproduce: the strict final-block `Utf8JsonReader` in `ProcessChunk` throws `JsonReaderException` (a `JsonException`) on truncated input *before* any buffer growth. I verified with a three-way experiment:

| Experiment | Guard | Safety | Truncated-input test |
|---|---|---|---|
| 1 | removed | intact | ✅ passes — guard is redundant in normal operation |
| 2 | removed | broken (forced lenient parsing) | ❌ fails — reproduces the exact 64 MiB-growth → `"maximum buffer size"` error the review feared; **the test catches it** |
| 3 | present | broken | ❌ fails (3/4) — guard only rescues the growth case, doesn't mask the regression |

So a product-side guard would be dead code. Instead I kept a **regression test** (`SplitIntoSubstreamsAsync_TruncatedInput_…`) that pins the contract (truncated input → `JsonException`, never a buffer-size error) and is proven non-vacuous by experiment 2. No product change.

**Missing cancellation** (`StreamDecryptableItem.cs:49`) — this is the same as the `GetItemAsync<T>(CancellationToken)` overload we **deliberately added and then reverted** on 2026-06-04: the slow part of the decrypt (the per-byte `StreamProcessor` inner loop) is synchronous and CT-deaf, and the DEK fetch is cached after first use, so the token would fire only in a rare HSM-slow corner case. Customers can already `ct.ThrowIfCancellationRequested(); await item.GetItemAsync<T>();`. Holding the revert.

### ⏸️ Real, but deferred as design calls (not in this commit)

These are legitimate but need a deliberate decision rather than a reflexive patch:

- **`OverwriteStreamAsync` truncate-before-write** (🟡 labeled high) — `SetLength(0)` then async copy; if `CopyToAsync` throws mid-copy the in-memory body is left partial. In practice the input is always a `MemoryStream` and the exception propagates (the corrupted stream isn't returned), so impact is limited to a caller that swallows the exception and re-reads. Fix would be a two-phase commit (materialize → swap) — worth doing but is a behavioral change to the in-place contract.
- **Discarded `DecryptionContext`** in `DecryptJsonArrayStreamInPlaceAsync` (🟡) — symmetric with the Newtonsoft path; the aggregation is currently dead. Thread it to diagnostics *or* remove the machinery — a design choice.
- **No typed per-call opt-in / LINQ ignores `RequestOptions`** (🟡, the two `kushagraThapar` items + `ToEncryptionFeedIterator`) — promoting `JsonProcessor` to public + a typed property on `EncryptionItemRequestOptions` + a `RequestOptions` param on the LINQ extensions is a public-API surface decision that should be made deliberately before this leaves preview.
- **Runtime-only `IAsyncDisposable`** on the feed response (💬 observation) — a clean fix needs a compile-time-disposable return type; flagged by the reviewer themselves as "worth a deliberate design decision."

@kushagraThapar / @kundadebdatta — happy to take any of the deferred items in this PR or as follow-ups; they're tracked. Just let me know which you want to gate the merge on.

**Validation**: net8.0 unit suite 698 passed / 0 failed. Build clean. PR HEAD now `1e8ba74`.

adamnova:
Follow-up on the deferred design-call items from the previous round. After discussion, all four are now resolved (commits `c1717f5`, `08ce0ebe5`, `861fa02`).

| Item | Resolution |
|---|---|
| **`OverwriteStreamAsync` truncate-then-write data loss** | **Fixed.** The in-place overwrite did `SetLength(0)` (destroying the original ciphertext) then an async `CopyToAsync`; a cancellation between the two could leave the response body empty/partial. The decrypted payload is already fully materialized in the pooled buffer, so the copy-back is now a **synchronous** memcpy over the in-memory seekable destination — no interruptible gap, corruption window eliminated. Cancellation is still observed during the decrypt loop that precedes it. |
| **Discarded page-level `DecryptionContext`** | **Removed the dead aggregation.** Verified it was completely unused: the per-page aggregate (DEK-id → decrypted paths) was computed and then dropped one level up (`MdeEncryptionProcessor.DecryptJsonArrayStreamInPlaceAsync` ignored the return), matching the Newtonsoft feed path. It is *success* metadata, not an error signal — decryption **failures** propagate as exceptions (and `EncryptionException` with the DEK id on the lazy path), unaffected. Wiring it into telemetry was rejected as a **privacy regression** (it would leak the encryption schema + key ids into traces); the per-item `DecryptionContext` remains available on the point-read/lazy path. A comment notes any future per-page summary should be **returned to the caller** (who already holds the plaintext), not emitted to telemetry. Saves a per-page allocation. |
| **Disposal foot-gun (`IAsyncDisposable` runtime-only)** | **Finalizer safety net.** Compile-time discoverability is genuinely blocked — `EncryptionFeedIterator<T>.ReadNextAsync` *overrides* `FeedIterator<T>.ReadNextAsync`, whose return type is fixed to the non-disposable `FeedResponse<T>` by the base SDK, so the page cannot advertise `IAsyncDisposable` at compile time (and the reviewer's "return a `DisposableFeedResponse<T>`" idea can't compile against that contract). Instead we made forgetting-to-dispose **benign**: `PooledMemoryStream` now has a finalizer that returns **and zeroes** the rented buffer if disposal was missed. A missed dispose degrades to GC-timed cleanup instead of permanent pool-slot loss + plaintext residue. Disposing remains the prompt path and suppresses the finalizer (via `GC.SuppressFinalize`). Regression test forces finalization and asserts the buffer is zeroed. |
| **Wrong exception type** | **Fixed** (previous round): `UseStreamingJsonProcessingByDefault` now throws `ArgumentException(nameof(container))`. Docs corrected to match this round. |
| **Typed per-call opt-in + LINQ `RequestOptions`** | **Docs improved; typed public surface deferred to a focused follow-up.** The per-call channel must be the `RequestOptions.Properties` bag because streaming applies to feed ops whose options types (`QueryRequestOptions`/`ChangeFeedRequestOptions`/`ReadManyRequestOptions`) are owned by the **core SDK**, not this package — so a typed property can't simply be added. A proper public shape (making `JsonProcessor` public + an extension + LINQ `RequestOptions`) is a one-way GA API decision that deserves its own review, so I've documented the supported mechanism instead: `Properties["encryption-json-processor"] = "Stream"` / `"Newtonsoft"`, with a code example, plus an explicit note that the LINQ entry points honor only the container default. Happy to file the follow-up API-design issue (cc @kushagraThapar) — say the word and I'll draft it.

Refuted (for the record): the **"missing cancellation"** item is the same `GetItemAsync<T>(CancellationToken)` overload we added and then deliberately reverted last round (the slow inner decrypt loop is synchronous/CT-deaf; near-zero practical value). The **"truncated JSON → misleading buffer-size error"** item didn't reproduce — the strict final-block `Utf8JsonReader` throws a clean `JsonException` before any buffer growth (proven by a guard-removed/safety-broken experiment matrix); kept a regression test that pins the contract.

**Validation**: net8.0 unit suite **699 passed / 0 failed**, build clean on both TFMs. PR HEAD `861fa02`. Still pending an authorized `/azp run` to exercise the full CI pipeline.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
adamnova: Good catch — fixed in `c1717f5`. The in-place overwrite now copies the decrypted bytes back **synchronously** from the already-materialized pooled buffer instead of an async `CopyToAsync`, so there's no cancellation/partial-write gap after `SetLength(0)`. Cancellation is still observed during the decrypt loop that runs beforehand.

Microsoft.Azure.Cosmos.Encryption.Custom/src/StreamDecryptableItem.cs
adamnova: We actually added this exact `GetItemAsync<T>(CancellationToken)` overload earlier in the PR and then deliberately reverted it. The slow part of the decrypt — the per-byte `StreamProcessor` loop — is synchronous and ignores the token, and the DEK is cached after first use, so the token would only ever fire in a rare HSM-slow corner case. Callers wanting fail-fast can `ct.ThrowIfCancellationRequested()` before the call. Happy to revisit if there's a concrete request-deadline scenario you're hitting.

Microsoft.Azure.Cosmos.Encryption.Custom/src/StreamDecryptableItem.cs
adamnova: Fixed in `1e8ba74` — the catch now sets `isDisposed = true` before throwing, so a retry after a failed decrypt surfaces a clean `ObjectDisposedException` instead of NRE-ing on the nulled stream and masking the original error. Added a regression test (`GetItemAsync_WhenDecryptionFailsThenCalledAgain_SurfacesObjectDisposedException`).

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/MdeEncryptionProcessor.cs
adamnova: Resolved in `c1717f5` by **removing the dead aggregation** rather than wiring it out. It was confirmed completely unused — the feed path only ever returns the decrypted stream (same as the Newtonsoft path) — and surfacing it via telemetry would leak the encryption schema + key ids. Decryption failures still propagate as exceptions, and the per-item `DecryptionContext` is still surfaced on the point-read / lazy path.

Microsoft.Azure.Cosmos.Encryption.Custom/src/JsonArrayStreamSplitter.cs
adamnova: Looked into this closely and it doesn't reproduce: the strict final-block `Utf8JsonReader` throws a clean `JsonException` (`JsonReaderException`) on truncated input *before* any buffer growth. I verified with a guard-on/off × safety-intact/broken experiment matrix — the contract holds with no added guard. Rather than ship dead code, I kept a regression test (`SplitIntoSubstreamsAsync_TruncatedInput_...`) that pins the "clean JsonException, never a buffer-size error" contract (and is proven non-vacuous: breaking the parser's final-block handling does fail it).

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionFeedIterator{T}.cs
adamnova: Fixed in `1e8ba74` — `ReadNextAsync` now drains the partially-built items in a `catch` if `CreateResponse` throws, mirroring the orphan-cleanup already inside `ConvertResponseToDecryptableItemsStreamAsync`.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainerExtensions.cs
adamnova: Fixed in `1e8ba74` — now throws `ArgumentException` with `nameof(container)`. Thanks.

Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainerExtensions.cs
adamnova: You're right. The full fix (a typed per-call surface + `RequestOptions` on the LINQ entry points) is a public-API decision we're deferring to a focused follow-up, because the per-call channel has to go through the property bag — the feed options types (`QueryRequestOptions` etc.) are owned by the core SDK, so a typed property can't simply be added here. For now I documented the limitation on both LINQ extensions in `861fa02`: they honor only the container default. I'll cc you on the follow-up issue.

Microsoft.Azure.Cosmos.Encryption.Custom/src/DecryptableFeedResponse.cs
adamnova: Agreed it's a real foot-gun. The compile-time fix is unfortunately blocked: `ReadNextAsync` overrides `FeedIterator<T>.ReadNextAsync`, whose return type is fixed to the non-disposable `FeedResponse<T>`, so the page can't advertise `IAsyncDisposable` at compile time (a derived `DisposableFeedResponse<T>` can't satisfy that override). Instead I made forgetting-to-dispose **benign** in `08ce0ebe5`: `PooledMemoryStream` now has a finalizer that returns and zeroes the buffer if disposal is missed, so it degrades to GC-timed cleanup instead of pool-slot loss + plaintext residue. Disposing remains the prompt path and suppresses the finalizer.

adamnova:
@kushagraThapar — both addressed:

**Changelog dependency removal:** good flag, though the specific package was off — `Microsoft.IO.RecyclableMemoryStream` was never in `main`'s shipped csproj (it lived only transiently on this branch). The *actual* undeclared removal is `System.Text.RegularExpressions 4.3.1`, which this PR drops. Added an `#### Updates` entry for it in `1e8ba74`.

**Typed per-call opt-in:** agreed the magic-string/-int shape isn't great. The full typed surface (public `JsonProcessor` + a typed override on the options + `RequestOptions` on the LINQ entries) is a one-way GA API decision, and the per-call channel structurally has to go through the property bag since the feed options types are owned by the core SDK — so I'd rather not commit the shape inside this already-large PR. For now I documented the supported mechanism (`Properties["encryption-json-processor"] = "Stream"` / `"Newtonsoft"`, with an example) and the LINQ limitation in `861fa02`. I'll file the API-design follow-up and cc you — happy to take your steer on (a) public enum + typed property vs (b) helper extensions there.

kundadebdatta:
## 🔴 PR Deep Review · 2 findings (1 🔴 Blocking, 1 🟢)

Reviewed the streaming JSON processing feature in depth -- including focused passes over the pooled-buffer lifecycle, the decryptor, and the streaming JSON splitter. Overall a strong, careful feature: I verified that the prior review round's fixes hold (`c1717f5`, `1e8ba74`, `08ce0ebe5`), and that the `PooledMemoryStream` finalizer/dispose lifecycle is correct (no double-return; `Stream.Dispose` suppresses the finalizer; plaintext is cleared on return). The decryptor and `StreamDecryptableItem` paths are clean.

One correctness bug surfaced in the streaming buffer management:

1. 🔴 **Premature buffer growth on short reads** (`JsonFeedStreamHelper.HandleLeftOver`) -- growth triggers on `leftOver == dataLength` (nothing consumed) instead of `leftOver == buffer.Length` (buffer full); a large token spanning multiple short reads doubles the buffer each read until it throws the max-buffer-size error on valid input.
2. 🟢 **Leading BOM not stripped** (low confidence; Cosmos responses are normally BOM-less).

Details inline.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/JsonFeedStreamHelper.cs
```diff
@@ -0,0 +1,211 @@
...
+            }
+
+            readerState = reader.CurrentState;
+
+            return new ProcessResult(checked((int)reader.BytesConsumed), objectCompleted: false);
+        }
+
+        internal static byte[] HandleLeftOver(byte[] buffer, int dataLength, int leftOver, int bytesConsumed, int maxBufferSize)
+        {
+            if (leftOver == dataLength)
```
kundadebdatta: **🔴 Blocking** · Correctness: Premature Buffer Growth On Short Reads

**`HandleLeftOver` grows the buffer when nothing was consumed, not when the buffer is full**

The growth condition `leftOver == dataLength` is equivalent to `bytesConsumed == 0` (nothing consumed this round) -- **not** "the buffer is full." The streaming read loop parses after a **single** `ReadAsync(buffer, leftOver, buffer.Length - leftOver)`, and streams are permitted to return **short reads**. Once parsing is part-way through one large JSON token (e.g., a large base64 encrypted property value) that spans multiple short reads, `Utf8JsonReader` consumes 0 bytes each iteration, so `leftOver == dataLength` holds **even though the buffer still has free space** -- so the buffer doubles on every short read, exhausts `maxBufferSize`, and throws `InvalidOperationException` ("JSON document or token does not fit within the maximum buffer size...") on a perfectly valid feed that would have parsed fine had reads filled the buffer first.

This is the standard `Utf8JsonReader` streaming pitfall: grow only when the buffer is genuinely full. Change the condition to `if (leftOver == buffer.Length)`.

It also explains why the earlier "misleading error on malformed input" concern (`JsonArrayStreamSplitter.cs:86`) is not fully resolved under chunked reads: a truncated large string hits this max-buffer error *before* the strict final-block `JsonException` that was verified for single-read input.

**Suggested fix:** change `if (leftOver == dataLength)` to `if (leftOver == buffer.Length)`, and add a regression test using a stream that returns short reads (e.g., 64-128 byte chunks) over a document whose value is larger than one chunk. (If the feed stream is always an already-buffered `MemoryStream` in practice, this is latent today, but the one-line fix removes a real failure mode for streamed responses.)

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/JsonFeedStreamHelper.cs
```diff
@@ -0,0 +1,211 @@
...
+        private const string DocumentsMustBeArrayMessage = "The Documents property must be a JSON array.";
+
+        internal static ProcessResult ProcessChunk(
+            ReadOnlySpan<byte> dataSegment,
+            bool finalBlock,
+            ref JsonReaderState readerState,
+            ref JsonArrayTraversalState traversalState,
+            JsonSegmentWriter writeSegment)
+        {
+            Utf8JsonReader reader = new (dataSegment, finalBlock, readerState);
```
kundadebdatta, line 23: **🟢 Suggestion** · Robustness: Leading BOM Not Stripped

**A UTF-8 BOM at the start of the payload makes `Utf8JsonReader` throw**

`Utf8JsonReader` does not skip a leading byte-order mark; an input beginning with `EF BB BF` throws a `JsonException` at this reader construction. Cosmos JSON responses are normally BOM-less, so this is low-probability -- but if any feed / change-feed payload can carry a BOM, strip it at stream start before constructing the reader. Flagging for confirmation rather than as a definite defect.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/JsonFeedStreamHelper.cs
adamnova: Confirmed and fixed in `43775b1d4`. You're exactly right — `leftOver == dataLength` is really `bytesConsumed == 0` ("nothing consumed this round"), which under short reads fires while the buffer still has free space, so it doubled on every read until the 64 MiB cap threw on otherwise-valid input. Changed the guard to grow only when the buffer is genuinely full (`leftOver == buffer.Length`).

I reproduced it before fixing: a valid 64 KiB-value document fed through a `ChunkedReadStream` (128 bytes/read) threw `InvalidOperationException: ...maximum buffer size of 67108864 bytes` on the old code, and parses cleanly after the change. Added that as a regression test (`SplitIntoSubstreamsAsync_ShortReadsAcrossLargeValue_DoesNotPrematurelyGrowBuffer`), and as you predicted the existing truncated-input test now also covers the chunked case — short-read truncation surfaces the clean `JsonException` instead of the buffer-size error. Thanks for catching this.

Microsoft.Azure.Cosmos.Encryption.Custom/src/JsonFeedStreamHelper.cs
adamnova: Thanks — I dug into this and I'm going to skip the strip here. The only stream that reaches `SplitIntoSubstreamsAsync` is the Cosmos feed / change-feed response body, which is UTF-8 without a BOM by contract (neither the gateway nor the SDK serializer prepends `EF BB BF`), so `Utf8JsonReader` never sees a BOM on this path. Adding a strip would be defensive code on an input that can't occur here. Happy to revisit if a concrete BOM-bearing payload ever turns up. Resolving.

kundadebdatta (approved):
LGTM Now.

kushagraThapar (approved):
LGTM, thanks @MartinSarkany

PR #5479, opened 2025-11-07
[Client Encryption] Add ArrayPool-backed pooled streams for reduced allocations in streaming encrypt/decrypt operations

adamnova:
# [Internal] Client Encryption: Refactors stream-processor to reduce decrypt-path allocations

## What this PR changes

Surgical optimisations on the `JsonProcessor.Stream` path in `Microsoft.Azure.Cosmos.Encryption.Custom`. No public-API changes, no changes to the `Newtonsoft` path.

1. **Decrypt writer routed through `IBufferWriter<byte>`.** `Utf8JsonWriter(Stream)` on .NET 8 eagerly constructs an internal `ArrayBufferWriter<byte>` that is GC-heap backed (`Array.Resize` doubling from 256 B), producing ~2× the final JSON size in short-lived GC garbage per op. The decrypt core now uses `Utf8JsonWriter(IBufferWriter<byte>)` over a pooled `RentArrayBufferWriter`, eliminating that internal buffer entirely. The new-output adapter path returns a `ReadOnlyBufferWriterStream` that owns the rented buffer (cleared on dispose for defense-in-depth). The caller-provided-output path shares the same core and memcpy-copies out to the user stream at the end.
2. **`_ei` metadata subtree extraction streams via `Utf8JsonReader`** instead of `JsonSerializer.DeserializeAsync<EncryptionPropertiesWrapper>(stream)`. The old call forces `ReadBufferState` to grow `16K → 32K → 64K → 128K` (final rental lands on the LOH) because `Skip()` over every unknown root property needs the complete value in-buffer. The replacement (`EncryptionPropertiesStreamReader`) uses `Utf8JsonReader` with `TrySkip` and `isFinalBlock: false`, staying in a 4 KB pooled buffer and only growing when a single value exceeds the current buffer. Safe-rewind snapshotting handles `_ei` truncation across chunk boundaries; an explicit guard prevents pathological growth on partial-read (trickle) transports.
3. **Property-name matching uses `Utf8JsonReader.ValueTextEquals`** with pre-encoded UTF-8 path bytes instead of allocating `"/" + reader.GetString()` per property.
4. **`PooledMemoryStream` rents its backing buffer lazily** on first write.
5. **`ArrayPoolManager.rentedBuffers` is pre-sized** to cover the typical decrypt rent count.
6. **`PooledStreamConfiguration` XML docs** clarify that `SetConfiguration` is configure-once-before-first-use.

## How `PooledMemoryStream` and friends are actually used

All pooling types introduced by this PR are `internal sealed` and live under `Microsoft.Azure.Cosmos.Encryption.Custom.Common`. No public API surface changes.

### The types

| Type | Role | Lifetime |
|------|------|----------|
| `PooledMemoryStream` | `Stream` backed by an `ArrayPool<byte>.Shared` rental that grows by doubling. Clears the buffer before return by default. | Constructed by product code, handed back to user as the `Stream` return of `EncryptAsync` / `DecryptStreamAsync`; disposal returns the rental. |
| `ReadOnlyBufferWriterStream` *(new)* | Read-only, seekable `Stream` wrapper around a `RentArrayBufferWriter`. Owns the writer and disposes it (clearing the rented buffer) on `Dispose`. | Constructed only inside the decrypt adapter; handed back to user as the returned `Stream`; disposal returns the rental. |
| `RentArrayBufferWriter` | `IBufferWriter<byte>` backed by `ArrayPool<byte>.Shared`. Internal-only. | Constructed per operation for `Utf8JsonWriter` to write through; disposed at end of op (or transferred into `ReadOnlyBufferWriterStream`). |
| `ArrayPoolManager` | Bookkeeping for multiple `ArrayPool<byte>` rentals that all return at end of op. | One instance per encrypt/decrypt op; disposed at end. |
| `PooledJsonSerializer` | Static helpers: shared `JsonSerializerOptions`, serialize-to-pooled-stream, deserialize-from-stream. | Stateless. |
| `PooledStreamConfiguration` | Static configure-once knobs: `StreamInitialCapacity` (4 KB), `BufferWriterInitialCapacity` (256 B), `StreamProcessorBufferSize` (16 KB). | Process-lifetime singleton. |

### The three real call-flows

All external traffic enters through the public `EncryptionProcessor` static class, which is driven by `EncryptionContainer` on the user's behalf. Every returned pooled stream is eventually disposed by the Cosmos SDK (`ResponseMessage.Dispose()` chains to `Content.Dispose()`, which walks into our `Dispose` that returns the rental to the pool).

#### 1. Encrypt to a new stream — `EncryptAsync(Stream input, …) → Stream`

```
user Cosmos call
  └─ EncryptionContainer.CreateItemStreamAsync / ReadItemStreamAsync (ETM-enabled path)
      └─ EncryptionProcessor.EncryptAsync (public static, dispatches on algorithm)
          └─ MdeEncryptionProcessor.EncryptAsync
              └─ IMdeJsonProcessorAdapter.EncryptAsync
                  ├─ SystemTextJsonStreamAdapter (JsonProcessor.Stream)
                  │    [1] new PooledMemoryStream()          ◄── rents from pool on first write
                  │    [2] streamProcessor.EncryptStreamAsync(input, ms, …)   ── writes via Utf8JsonWriter(Stream)
                  │    [3] return ms as Stream               ◄── ownership transfers to caller
                  └─ NewtonsoftAdapter (JsonProcessor.Newtonsoft) — unchanged by this PR
```

Caller (`EncryptionContainer.cs:78–166`) wraps the returned stream in `using (Stream streamPayload = …)` so the pool rental is returned after `CreateItemStreamAsync` sends the payload.

**Pool-managed allocations in this flow:** 1 × `PooledMemoryStream` output + per-operation ArrayPool rents for intermediate buffers. Returned.

#### 2. Encrypt into a caller-supplied stream — `EncryptAsync(Stream input, Stream output, …)`

```
EncryptionProcessor.EncryptAsync(input, output, …) (NET8+ overload)
  └─ MdeEncryptionProcessor.EncryptAsync(input, output, …)
      └─ SystemTextJsonStreamAdapter.EncryptAsync(input, output, …)
          └─ StreamProcessor.EncryptStreamAsync(input, output, …)
              ── writes directly to caller's output Stream via Utf8JsonWriter(Stream)
```

No `PooledMemoryStream` is constructed in this path — the caller already supplied a stream. Used by performance-critical callers that want to bring their own pooled output (the benchmark harness supplies `RecyclableMemoryStream`; production can use any `Stream`).

**Pool-managed allocations in this flow:** 0 × `PooledMemoryStream`; only intermediate `RentArrayBufferWriter` used by the stream processor for per-property encryption buffers.

#### 3. Decrypt to a new stream — `DecryptAsync(Stream input, …) → (Stream, DecryptionContext)`

This is the path this PR reshapes the most.

```
EncryptionContainer.*StreamAsync post-processing (decryptResponse branch)
  └─ EncryptionProcessor.DecryptAsync (public static, peeks for legacy algo)
      └─ MdeEncryptionProcessor.DecryptAsync (dispatches on JsonProcessor)
          └─ IMdeJsonProcessorAdapter.DecryptAsync
              └─ SystemTextJsonStreamAdapter.DecryptAsync (JsonProcessor.Stream)
                  [1] EncryptionPropertiesStreamReader.ReadAsync(input)      ◄── NEW: streaming _ei extractor
                  [2] new RentArrayBufferWriter(StreamInitialCapacity)       ◄── NEW: writes through IBufferWriter
                  [3] streamProcessor.DecryptStreamAsync(input, bufferWriter, …) ── Utf8JsonWriter(IBufferWriter<byte>)
                  [4] return (new ReadOnlyBufferWriterStream(bufferWriter), context)
                       │
                       └── the ReadOnlyBufferWriterStream takes ownership;
                           user's ResponseMessage.Dispose() disposes the stream,
                           which disposes the RentArrayBufferWriter,
                           which returns the rented buffer to ArrayPool<byte>.Shared.
```

`EncryptionContainer.cs` assigns the returned stream directly to `responseMessage.Content`. When the application disposes the `ResponseMessage`, the pool rental is returned.

**Pool-managed allocations in this flow:** 1 × `RentArrayBufferWriter` (output) + per-op rents for decrypt buffers and `_ei` scan. All returned on `Dispose`.

#### 3b. Decrypt into a caller-supplied stream — `DecryptAsync(Stream input, Stream output, …)`

Same as (3) but `SystemTextJsonStreamAdapter.DecryptAsync(input, output, …)` calls `StreamProcessor.DecryptStreamAsync(Stream, Stream, …)`, which internally uses the same `IBufferWriter<byte>` core and then `CopyToAsync`'s the result to the caller's stream. No `ReadOnlyBufferWriterStream` is constructed. This is the lowest-allocation path: the caller owns and disposes their output stream and all the rest stays inside our pool.

### Ownership & disposal contract (summary)

- Everything the library hands back to the caller via a `Stream` return value is owned by the caller and MUST be disposed to return the rental. In the Cosmos pipeline this happens transparently via `ResponseMessage.Dispose()`.
- Methods taking `Stream output` do NOT take ownership of the caller's stream — the caller disposes it.
- Every internal `ArrayPool` rental is scoped to `using RentArrayBufferWriter`/`using ArrayPoolManager` blocks so exceptions never leak pool memory.
- `clearOnReturn: true` is the default on every rental path; decrypted plaintext and ciphertext buffers are zeroed before they go back to the shared pool.

### Should these types be public?

**Short answer: not in this PR, and likely not worth doing at all.** Detailed reasoning:

| Type | Case for making public | Case against |
|------|------------------------|--------------|
| `PooledMemoryStream` | Callers could supply it to `EncryptAsync(input, output, …)` instead of `MemoryStream` or `RecyclableMemoryStream`. | `Microsoft.IO.RecyclableMemoryStream` already solves the same problem with a mature API, multi-pool support, and telemetry; we'd be reinventing a slightly worse version. The class is deliberately minimal (lazy rent, clear-on-return) — it is *tightly coupled* to our security contract, not a general-purpose pooled stream. |
| `ReadOnlyBufferWriterStream` | Callers could reuse it to wrap their own `IBufferWriter<byte>`. | `System.IO.Pipelines` + `PipeReader.AsStream()` already cover this case in the BCL. Our wrapper is narrowly scoped to "take ownership of one `RentArrayBufferWriter` and dispose it" — that ownership model isn't general. |
| `RentArrayBufferWriter` | Callers could drive `Utf8JsonWriter` through it to eliminate STJ's internal buffer, same trick we apply. | Duplicate of an existing gist-quality `ArrayPoolBufferWriter` that various libraries ship; a public version in this library is not where downstream consumers would look for it. |
| `PooledStreamConfiguration` | Hosts that run many tenants with very different doc sizes could tune `StreamInitialCapacity`. | The defaults (4 KB / 16 KB / 256 B) are well-chosen for encrypted Cosmos documents. The comment from @kirankumarkolli asked for an **opt-in** story for the new pooled behavior as a whole, which is a different question than exposing the knobs. |

**Recommendation (separate from this PR):** keep the pooling types `internal` and, if we want an opt-in, instead surface a single new `RequestOptions`-like property such as `EncryptionAllocationMode = { Default, PooledStream }` gated behind a feature flag. That addresses @kirankumarkolli's opt-in concern without committing to public pool-helper types we'd have to maintain forever.

## Benchmark methodology

Both runs executed on the same idle machine (Windows 11, .NET SDK 10.0.202, .NET 8.0.26 X64 RyuJIT AVX2, BenchmarkDotNet v0.13.3). `MediumRun` job: 15 iterations × 2 launches × 10 warmup, `OperationsPerInvoke = 16`, `InProcessEmitToolchain`, `MemoryDiagnoser`. The benchmark harness (`EncryptionBenchmark.cs`) is **byte-identical** between the two runs — only product code differs.

- **`master`**: commit `79d18b73` with this PR's harness back-ported (harness-only changes: concrete `BenchmarkKeyStoreProvider`, drop of the spurious `ENCRYPTION_CUSTOM_PREVIEW` gate, `OperationsPerInvoke = 16`).
- **`this PR`**: commit `9fea360f` on `feature/stream-processor-optimizations` (subsequent commits add tests, docs, and comment cleanup only — measured numbers unchanged).

Run with:
```powershell
cd Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests
dotnet run -c Release --framework net8.0 -- --filter '*EncryptionBenchmark*'
```

## Newtonsoft paths — sanity check (not touched by this PR)

All Newtonsoft scenarios agree within run-to-run noise (≤ 0.52 %), confirming the comparison is apples-to-apples.

| Scenario | master Alloc | this PR Alloc | Δ |
|---|---:|---:|---:|
| Encrypt 1 KB                    |    36,552 B |    36,648 B | +0.26% |
| Decrypt 1 KB                    |    54,768 B |    54,960 B | +0.35% |
| DecryptToProvidedStream 1 KB    |    36,688 B |    36,880 B | +0.52% |
| Encrypt 10 KB                   |   171,353 B |   171,449 B | +0.06% |
| Decrypt 10 KB                   |   198,722 B |   198,913 B | +0.10% |
| DecryptToProvidedStream 10 KB   |   125,401 B |   125,593 B | +0.15% |
| Encrypt 100 KB                  | 1,693,159 B | 1,693,133 B | −0.00% |
| Decrypt 100 KB                  | 1,584,352 B | 1,584,568 B | +0.01% |
| DecryptToProvidedStream 100 KB  | 1,013,492 B | 1,013,660 B | +0.02% |

## Stream paths — where this PR moves the numbers

| Scenario | master Alloc | this PR Alloc | Alloc Δ | master Mean | this PR Mean | Time Δ |
|---|---:|---:|---:|---:|---:|---:|
| 1 KB  Encrypt                  |    16,552 B |    13,816 B | **−16.5%** |    27.54 μs |    19.50 μs | **−29.2%** |
| 1 KB  EncryptToProvidedStream  |    10,392 B |     9,704 B |  **−6.6%** |    24.24 μs |    18.96 μs | **−21.8%** |
| 1 KB  Decrypt                  |    27,328 B |    24,704 B |  **−9.6%** |    51.53 μs |    42.25 μs | **−18.0%** |
| 1 KB  DecryptToProvidedStream  |    11,072 B |     5,720 B | **−48.3%** |    26.12 μs |    25.75 μs |   −1.4% |
| 10 KB Encrypt                  |    81,953 B |    45,873 B | **−44.0%** |    69.84 μs |    64.00 μs |  **−8.4%** |
| 10 KB EncryptToProvidedStream  |    36,048 B |    29,473 B | **−18.2%** |    62.92 μs |    65.16 μs |   +3.6% |
| 10 KB Decrypt                  |    70,673 B |    63,489 B | **−10.2%** |   129.35 μs |   128.67 μs |   −0.5% |
| 10 KB DecryptToProvidedStream  |    17,984 B |     5,720 B | **−68.2%** |    59.03 μs |    49.01 μs | **−17.0%** |
| 100 KB Encrypt                 |   677,078 B |   425,690 B | **−37.1%** |   976.77 μs |   647.31 μs | **−33.7%** |
| 100 KB EncryptToProvidedStream |   229,131 B |   163,545 B | **−28.6%** |   637.89 μs |   484.90 μs | **−24.0%** |
| 100 KB Decrypt                 |   539,936 B |   446,452 B | **−17.3%** | 1,065.20 μs |   946.17 μs | **−11.2%** |
| 100 KB DecryptToProvidedStream |   118,682 B |     5,906 B | **−95.0%** |   468.04 μs |   386.53 μs | **−17.4%** |

### Takeaway

All 12 Stream-processor scenarios are **neutral or better than master on both allocations and wall time**. Highlights:

- **100 KB DecryptToProvidedStream: −95 % allocations** (118 KB → 6 KB) and **−17 % wall time**. Gen2 collections per 1 000 ops drop from 2.93 to 0.
- **100 KB Encrypt: −37 % allocations and −34 % wall time** (677 KB → 426 KB; 977 μs → 647 μs).
- **100 KB Decrypt: −17 % allocations and −11 % wall time** (540 KB → 446 KB; 1065 μs → 946 μs).

## Testing

- 375 unit tests pass (302 pre-existing + 73 new covering the new/modified types to 100 % of the lines this PR touches).
- Newtonsoft path is untouched and allocation numbers confirm parity (≤ 0.52 % noise).
- Cosmos emulator integration tests were not run.

## Risks

- The pooled output streams are only safe as long as callers dispose them. In the Cosmos pipeline this is automatic (`ResponseMessage.Dispose()` chains through). Any hypothetical direct caller of `EncryptionProcessor.EncryptAsync` / `DecryptAsync` that forgets to dispose the returned stream leaks one `ArrayPool` rental — identical risk surface to the pre-existing `PooledMemoryStream`, so no new leak class.
- `EncryptionPropertiesStreamReader` fails fast on non-seekable input (`ArgumentException`). This matches pre-PR behaviour, which threw `NotSupportedException` from `input.Position = 0` on the same input.

## Checklist

- [x] Branch name matches `users/<user>/...`
- [x] PR title regex `(\[Internal\]|\[v4\] )?.{3}.+: (Adds|Fixes|Refactors|Removes) .{3}.+`
- [x] `Co-authored-by: Copilot` trailer on every commit
- [x] No `Directory.Build.props` / versioning / packaging changes
- [x] Unit tests added; full suite passes on net8.0
- [x] Benchmark README updated with apples-to-apples numbers


Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/MdeEncryptionProcessor.cs
```diff
@@ -138,14 +138,25 @@ public async Task EncryptAsync(
             CosmosDiagnosticsContext diagnosticsContext,
             CancellationToken cancellationToken)
         {
-            MemoryStream ms = new ();
-            DecryptionContext context = await this.StreamProcessor.DecryptStreamAsync(input, ms, encryptor, properties, diagnosticsContext, cancellationToken);
-            if (context == null)
+            PooledMemoryStream ms = new ();
```
kirankumarkolli, line 141: How about make it opt-in based on ENV variable. 
When opted will use new behavior? Later we can revisit the defaults

kirankumarkolli: .NET 8 re-targetted version is not yet released. 
Post release would be good to have optionality to opt-in new mechanism with current state being default?

Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests/Readme.md
```diff
@@ -1,61 +1,92 @@
...
+
+### Results by Document Size
+
+#### Small Document: 917 bytes (0.90 KB)
+
+| Processor | Operation | Total Allocated | Overhead | Alloc per Input KB |
+|-----------|-----------|----------------|----------|-------------------|
+| **Stream** | **Encrypt** | 7,556 bytes (7.38 KB) | 824.0% | 8,438 bytes |
+| **Stream** | **Decrypt** | 8,175 bytes (7.98 KB) | 612.4% | 6,271 bytes |
+| Newtonsoft (baseline) | Encrypt | ~41,784 bytes (40.8 KB) | ~4,457% | ~46,649 bytes |
```
kirankumarkolli: STJ as baseline?

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/MdeEncryptionProcessor.cs
adamnova: I don't think it would be a good idea to introduce complexity with opt-in to such behaviors. The whole release is significant change under the hood of Encryption.Custom and cherry picking something like PooledMemoryStream to be opt-in does not make sense to me, when there are 20 other changes. The changes we made are a major release and it is unavoidable that significant parts of the code will change.

Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests/Readme.md
```diff
@@ -1,61 +1,262 @@
...
+This project benchmarks the memory allocation and throughput of the encryption/decryption pipeline
+using [BenchmarkDotNet](https://benchmarkdotnet.org/). The benchmarks compare the **Newtonsoft** (legacy)
+and **Stream** (pooled, ArrayPool-backed) JSON processors across different document sizes.
 
+## Benchmark Methods
+
+| Method | Description |
+|--------|-------------|
+| `Encrypt` | Encrypts a document, returning a new stream |
+| `EncryptToProvidedStream` | Encrypts into a caller-provided `RecyclableMemoryStream` |
```
kirankumarkolli: How about use RecyclableMemoryStream not RecyclableMemoryStream (which is a separate package)

adamnova: What do you mean? I renamed it to `Stream`

Microsoft.Azure.Cosmos.Encryption.Custom/tests/Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests/Readme.md
adamnova: Done

PR #5482, opened 2025-11-10
HttpTimeoutPolicy Improvements Phase 2: Refactors Code to Separate out Retry Policy Timeouts for point-reads and non-point-reads on PPAF

### LINKED ISSUE #5484: [Fundamentals] HttpTimeoutPolicy‎: Update the `HttpTimeoutPolicyForPartitionFailover` to Use More Relaxed Timeouts for Query Workloads
### Acceptance Criteria:
 
- Refactor `HttpTimeoutPolicyForPartitionFailover` to Use More Relaxed Timeouts for Query Workloads

Praveen-Msft:
# Pull Request Template

## Description
This change fixes the aggressive retry timeouts for requests when PPAF enabled.  This change contains distinct retry policy timeouts for respective operations and allows us to manage the timeouts separately.

## Type of change

## Closing issues

To automatically close an issue: closes #5484

### DIFF
--- a/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
@@ -57,10 +57,9 @@ namespace Microsoft.Azure.Cosmos
                 {
                     if (isPartitionLevelFailoverEnabled)
                     {
-                        return documentServiceRequest.OperationType == OperationType.Query ||
-                               documentServiceRequest.OperationType == OperationType.SqlQuery
-                            ? HttpTimeoutPolicyForPartitionFailover.InstanceShouldThrow503OnTimeoutForQuery
-                            : HttpTimeoutPolicyForPartitionFailover.InstanceShouldThrow503OnTimeoutForReads;
+                        return documentServiceRequest.OperationType == OperationType.Read 
+                            ? HttpTimeoutPolicyForPartitionFailover.InstanceShouldThrow503OnTimeoutForPointReads
+                            : HttpTimeoutPolicyForPartitionFailover.InstanceShouldThrow503OnTimeoutForNonPointReads;
                     }
                     else
                     {
--- a/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
@@ -9,27 +9,27 @@ namespace Microsoft.Azure.Cosmos
 
     internal sealed class HttpTimeoutPolicyForPartitionFailover : HttpTimeoutPolicy
     {
-        public static readonly HttpTimeoutPolicy InstanceShouldThrow503OnTimeoutForQuery = new HttpTimeoutPolicyForPartitionFailover(isQuery: true);
-        public static readonly HttpTimeoutPolicy InstanceShouldThrow503OnTimeoutForReads = new HttpTimeoutPolicyForPartitionFailover(isQuery: false);
-        private readonly bool isQuery;
+        public static readonly HttpTimeoutPolicy InstanceShouldThrow503OnTimeoutForNonPointReads = new HttpTimeoutPolicyForPartitionFailover(isPointRead: false);
+        public static readonly HttpTimeoutPolicy InstanceShouldThrow503OnTimeoutForPointReads = new HttpTimeoutPolicyForPartitionFailover(isPointRead: true);
+        private readonly bool isPointRead;
         private static readonly string Name = nameof(HttpTimeoutPolicyDefault);
 
-        private HttpTimeoutPolicyForPartitionFailover(bool isQuery)
+        private HttpTimeoutPolicyForPartitionFailover(bool isPointRead)
         {
-            this.isQuery = isQuery;
+            this.isPointRead = isPointRead;
         }
 
         // Timeouts and delays are based on the following rationale:
-        // For reads: 3 agressive attempts with timeouts of 1s, 6s, and 6s respectively.
-        // For queries: 3 attempts with timeouts of 6s, 6s, and 10s respectively.
-        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForReads = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
+        // For point reads: 3 attempts with timeouts of 1s, 6s, and 6s respectively.
+        // For non-point reads: 3 attempts with timeouts of 6s, 6s, and 10s respectively.
+        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForPointReads = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
             (TimeSpan.FromSeconds(1), TimeSpan.Zero),
             (TimeSpan.FromSeconds(6), TimeSpan.Zero),
             (TimeSpan.FromSeconds(6), TimeSpan.Zero),
         };
 
-        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForQueries = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
+        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForNonPointReads = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
             (TimeSpan.FromSeconds(6), TimeSpan.Zero),
             (TimeSpan.FromSeconds(6), TimeSpan.Zero),
@@ -38,11 +38,11 @@ namespace Microsoft.Azure.Cosmos
 
         public override string TimeoutPolicyName => HttpTimeoutPolicyForPartitionFailover.Name;
 
-        public override int TotalRetryCount => this.isQuery ? this.TimeoutsAndDelaysForQueries.Count : this.TimeoutsAndDelaysForReads.Count;
+        public override int TotalRetryCount => this.isPointRead ? this.TimeoutsAndDelaysForPointReads.Count : this.TimeoutsAndDelaysForNonPointReads.Count;
 
         public override IEnumerator<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> GetTimeoutEnumerator()
         {
-            return this.isQuery ? this.TimeoutsAndDelaysForQueries.GetEnumerator() : this.TimeoutsAndDelaysForReads.GetEnumerator();
+            return this.isPointRead ? this.TimeoutsAndDelaysForPointReads.GetEnumerator() : this.TimeoutsAndDelaysForNonPointReads.GetEnumerator();
         }
 
         public override bool ShouldRetryBasedOnResponse(HttpMethod requestHttpMethod, HttpResponseMessage responseMessage)


Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
```diff
@@ -9,30 +9,42 @@ namespace Microsoft.Azure.Cosmos
...
+            this.isQuery = isQuery;
         }
 
-        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
+        // Timeouts and delays are based on the following rationale:
+        // For reads: 3 agressive attempts with timeouts of .5s, .5s, and 1s respectively.
+        // For queries: 3 attempts with timeouts of 5s, 5s, and 10s respectively.
+        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForReads = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
             (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
```
kundadebdatta: This is also too agressive. Let's go with `.5s`, `1s` and `5s`.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
```diff
@@ -9,30 +9,42 @@ namespace Microsoft.Azure.Cosmos
...
+
         public override string TimeoutPolicyName => HttpTimeoutPolicyForPartitionFailover.Name;
 
-        public override int TotalRetryCount => this.TimeoutsAndDelays.Count;
+        public override int TotalRetryCount => this.isQuery ? this.TimeoutsAndDelaysForQueries.Count : this.TimeoutsAndDelaysForReads.Count;
 
         public override IEnumerator<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> GetTimeoutEnumerator()
         {
-            return this.TimeoutsAndDelays.GetEnumerator();
+            return this.isQuery ? this.TimeoutsAndDelaysForQueries.GetEnumerator() : this.TimeoutsAndDelaysForReads.GetEnumerator();
```
kundadebdatta: On line `62` we should always return `ShouldThrow503OnTimeout` as `True`

```
public override bool ShouldThrow503OnTimeout => true;
```

Without this, the cross regional retry won't happen.

kundadebdatta: Fixed it with latest commit.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -55,9 +55,16 @@ public static HttpTimeoutPolicy GetTimeoutPolicy(
                 // Data Plane Reads.
                 else if (documentServiceRequest.IsReadOnlyRequest)
                 {
-                    return isPartitionLevelFailoverEnabled
-                        ? HttpTimeoutPolicyForPartitionFailover.InstanceShouldThrow503OnTimeout
-                        : HttpTimeoutPolicyDefault.InstanceShouldThrow503OnTimeout;
+                    if (isPartitionLevelFailoverEnabled)
+                    {
+                        return documentServiceRequest.OperationType == OperationType.Query
```
ananth7592: Curious, `documentServiceRequest.IsReadOnlyRequest` covers both `OperationType.Query` and `OperationType.SqlQuery` both. Do we need to select the query timeout for `sqlquery` as well?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemIntegrationTests.cs
```diff
@@ -2259,7 +2259,134 @@ public async Task ClinetOverrides0msRequestTimeoutValueForPPAF()
...
+
+            List<FaultInjectionRule> rules = new List<FaultInjectionRule> { serviceResponseDelay1};
+            FaultInjector faultInjector = new FaultInjector(rules);
+
+            List<string> preferredRegions = new List<string> { region1, region2, region3 };
+            CosmosClientOptions cosmosClientOptions = new CosmosClientOptions()
+            {
+                ConsistencyLevel = ConsistencyLevel.Session,
+                FaultInjector = faultInjector,
+                RequestTimeout = TimeSpan.FromSeconds(5),
```
ananth7592: Is `RequestTimeout` needed here? Since fault injection is already applied to not return the response for 70 seconds, the timeout poliicy of [5,5,10] would be triggered for all the retries?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
```diff
@@ -9,37 +9,47 @@ namespace Microsoft.Azure.Cosmos
 
     internal sealed class HttpTimeoutPolicyForPartitionFailover : HttpTimeoutPolicy
     {
-        public static readonly HttpTimeoutPolicy Instance = new HttpTimeoutPolicyForPartitionFailover(false);
-        public static readonly HttpTimeoutPolicy InstanceShouldThrow503OnTimeout = new HttpTimeoutPolicyForPartitionFailover(true);
-        public bool shouldThrow503OnTimeout;
+        public static readonly HttpTimeoutPolicy InstanceShouldThrow503OnTimeoutForQuery = new HttpTimeoutPolicyForPartitionFailover(true);
+        public static readonly HttpTimeoutPolicy InstanceShouldThrow503OnTimeoutForReads = new HttpTimeoutPolicyForPartitionFailover(false);
```
FabianMeiswinkel: I don't get why pint read and query would be different - can you explian (and add context in PR description and code comment here?)

FabianMeiswinkel: Ah - never mind - the boolean flag is just an indicator whether it is querxy - not whether to throw 503 or not. Can you please use a named paraemeter isQuery: true/false here ?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
```diff
@@ -9,37 +9,47 @@ namespace Microsoft.Azure.Cosmos
...
         }
 
-        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
+        // Timeouts and delays are based on the following rationale:
+        // For reads: 3 agressive attempts with timeouts of .5s, 1s, and 5s respectively.
+        // For queries: 3 attempts with timeouts of 5s, 5s, and 10s respectively.
+        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForReads = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
             (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
```
FabianMeiswinkel: 500ms also seesm too agressive. And seocnd retry should already be 5 seocnds, Like 1s, 5s, 5s

FabianMeiswinkel: Actually replace 5s with default NetworkRequestTimeout usedin .Net - probably either 6 s or 10s?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
```diff
@@ -9,37 +9,47 @@ namespace Microsoft.Azure.Cosmos
...
             (TimeSpan.FromSeconds(1), TimeSpan.Zero),
+            (TimeSpan.FromSeconds(5), TimeSpan.Zero),
+        };
+
+        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForQueries = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
+        {
+            (TimeSpan.FromSeconds(5), TimeSpan.Zero),
```
FabianMeiswinkel: 5s is backend time-budget - instead of 5s should add some buffer - I would just choose default Network request timeout for direct mode (i believe either 6s or 10s)

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
Praveen-Msft: I believe this is the default request timeout, which is 6s. would like to confirm if I am looking at the right field. 
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs#L21

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
kirankumarkolli: +1 for 6s

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
```diff
@@ -9,37 +9,47 @@ namespace Microsoft.Azure.Cosmos
...
-        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
+        // Timeouts and delays are based on the following rationale:
+        // For reads: 3 agressive attempts with timeouts of .5s, 1s, and 5s respectively.
+        // For queries: 3 attempts with timeouts of 5s, 5s, and 10s respectively.
+        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForReads = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
             (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
             (TimeSpan.FromSeconds(1), TimeSpan.Zero),
+            (TimeSpan.FromSeconds(5), TimeSpan.Zero),
```
kirankumarkolli: Last timeout values: 65s?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
kundadebdatta: Thanks for the suggestion @FabianMeiswinkel. I have refactored the code to reflect this.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
kundadebdatta: Updated the timeout to `1s`, `5s` and `6s`, with the last one with 6s default timeout.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
kundadebdatta: Updated the timeout to `6s`, as default timeout.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
kundadebdatta: This policy is particularly for PPAF and I think 65 secs will be too high to meet the failover threshold. I think we should keep the last attempt as 6 seconds, which should give enough time to successfully process most of the queries.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
Praveen-Msft: I think it should be 1s, 6s, 6s. since 5s is backend timeout

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
```diff
@@ -9,37 +9,47 @@ namespace Microsoft.Azure.Cosmos
...
 
-        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
+        // Timeouts and delays are based on the following rationale:
+        // For reads: 3 agressive attempts with timeouts of 1s, 6s, and 6s respectively.
+        // For queries: 3 attempts with timeouts of 6s, 6s, and 10s respectively.
+        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelaysForReads = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
             (TimeSpan.FromSeconds(1), TimeSpan.Zero),
```
FabianMeiswinkel, line 27: Sorry, probably better to make the 1 second configurable via env config - Reason behind it is that we technically allow overrides in config to use 16 MB documents - for 16 MB i could see certain edge cases where point reads legitimately take > 1 second E2E?

FabianMeiswinkel (dismissed):
LGTM except for the one comment about PPAF min. retry interval of 1 second

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyForPartitionFailover.cs
kundadebdatta: Discussed offline. This can be taken as a follow up PR.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemIntegrationTests.cs
kundadebdatta: No - it should not be needed. I have removed it from the test.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
kundadebdatta: Great catch. I see there are usage across msdata and it's worth adding `OperationType.SqlQuery` as well.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -55,9 +55,17 @@ public static HttpTimeoutPolicy GetTimeoutPolicy(
                 // Data Plane Reads.
                 else if (documentServiceRequest.IsReadOnlyRequest)
                 {
-                    return isPartitionLevelFailoverEnabled
-                        ? HttpTimeoutPolicyForPartitionFailover.InstanceShouldThrow503OnTimeout
-                        : HttpTimeoutPolicyDefault.InstanceShouldThrow503OnTimeout;
+                    if (isPartitionLevelFailoverEnabled)
+                    {
+                        return documentServiceRequest.OperationType == OperationType.Query ||
```
FabianMeiswinkel: What about ReadFeed (ChangeFeed) - shouldn't this logic be isReadOnlyRequest && !Read? and then only special case Read?

Praveen-Msft:
> LGTM except for the one comment about PPAF min. retry interval of 1 second

@FabianMeiswinkel, min retry interval already updated to 1sec. Do you have anything else on that?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
kundadebdatta: This is now addressed.

kundadebdatta (approved):
LGTM Now.

kirankumarkolli (approved):

PR #5498, opened 2025-11-14
Query: Fixes hybrid search query plan optimization to be enabled by default

neildsh:
Enable the hybrid search query plan optimization by default

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -135,7 +135,8 @@ namespace Microsoft.Azure.Cosmos
         //Auth
         internal readonly AuthorizationTokenProvider cosmosAuthorization;
 
-        private bool isThinClientEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
+        private readonly bool isThinClientFeatureFlagEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
+        internal bool isThinClientEnabled;
 
         // Gateway has backoff/retry logic to hide transient errors.
         private RetryPolicy retryPolicy;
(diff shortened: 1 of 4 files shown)


Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -227,7 +227,7 @@ public ConsistencyLevel? ConsistencyLevel
 
         internal FeedRange FeedRange { get; set; }
 
-        internal bool IsHybridSearchQueryPlanOptimizationDisabled { get; set; } = ConfigurationManager.IsHybridSearchQueryPlanOptimizationDisabled(defaultValue: true);
+        internal bool IsHybridSearchQueryPlanOptimizationDisabled { get; set; } = ConfigurationManager.IsHybridSearchQueryPlanOptimizationDisabled(defaultValue: false);
```
adityasa, line 230: >defaultValue: false [](http://example.com/codeflow?start=147&length=19)

Is local emulator updated to generate the new plan?

My concern is that if emulator still generates the old plan, we may actually not be activating logic required for new plan (in checkin gates).

adityasa: Also, maybe we should run some tests with old plan some with new one to ensure continuity?

neildsh: That is a good point. However, the plan is generated locally for windows, and the current version of the direct package seems to have the necessary changes.

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
```diff
@@ -1043,27 +1055,31 @@ private static StringBuilder AppendQueryResult(HybridSearchQueryResult queryResu
...
-                string header = builder.ToString();
-                System.Diagnostics.Trace.WriteLine(header);
+                    builder.Remove(builder.Length - 1, 1); // remove extra tab
+
+                    string header = builder.ToString();
+                    System.Diagnostics.Trace.WriteLine(header);
+                }
             }
 
             private static void TraceFullDebugTSVHeader(int componentCount)
```
khdang, line 1085: should we update this function to have the [Conditional("DEBUG")] and put it under the Enabled condition like the other trace functions?

khdang (approved):

Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
neildsh: All the public methods conform to that. The private methods can only be called from the public methods so they dont have the redundant check.

PR #5502, opened 2025-11-17
VectorIndex Policy[No Release]: Adds Support for QuantizerType in IndexingPolicy

aavasthy:
# Pull Request Template

## Description

- Updated direct package version as this PR needed some code references from shared files.

- Add a new field "quantizerType": "product/spherical" to vector index definition in indexing policy. 

Example:

        "vectorIndexes": [
          {
            "path": "/vector1",
            "type": "diskANN",
            "quantizerType": "product"
            "quantizationByteSize": 4,
            "indexingSearchListSize": 50
          }
        ],
## Type of change

Testing against a live account that has Vector Search for NoSQL API feature enabled:
- Ran e2e test and created a container with 

 ```
       Collection<Cosmos.VectorIndexPath> vectorIndexes = new Collection<Cosmos.VectorIndexPath>
{
    new Cosmos.VectorIndexPath
    {
        Path = "/embedding",
            Type = Cosmos.VectorIndexType.DiskANN,
        QuantizerType = Cosmos.QuantizerType.Product
    }
};
        properties.IndexingPolicy.VectorIndexes = vectorIndexes;
```
        
Checked in portal and it has the quantizer type value getting reflected as **product**: 
image
And checked with quantizer type as **Spherical**
image

## Closing issues

### DIFF
--- a/azure-pipelines-rolling.yml
+++ b/azure-pipelines-rolling.yml
@@ -18,7 +18,7 @@ schedules:
   always: true # whether to always run the pipeline or only if there have been source code changes since the last run. The default is false
 
 variables:
-  ReleaseArguments: ' --filter "TestCategory!=Flaky & TestCategory!=Quarantine & TestCategory!=Functional & TestCategory!=Ignore" --verbosity normal '
+  ReleaseArguments: ' --filter "TestCategory!=Flaky & TestCategory!=Quarantine & TestCategory!=Functional & TestCategory!=Ignore" --verbosity normal --logger "console;verbosity=detailed" --blame-hang --blame-crash'
   VmImage: windows-latest # https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops
 
 jobs:

### CODE REVIEW

File: Microsoft.Azure.Cosmos/src/Resource/Settings/QuantizerType.cs [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+    public
+#else
+    internal
+#endif
+    enum QuantizerType
+    {
+        /// <summary>
+        /// Represents a product quantizer type.
+        /// </summary>
+        [EnumMember(Value = "product")]
```
kirankumarkolli, line 24: Are these paths or only possible values?

aavasthy: quantizerType will be added in the vector index definition and spherical and product are the values
https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB/pullrequest/1777785?path=/Product/Microsoft.Azure.Documents/SharedFiles/QuantizerType.cs&version=GBmaster&line=14&lineEnd=15&lineStartColumn=1&lineEndColumn=1&type=2&lineStyle=plain&_a=files&iteration=12&base=0

Directory.Build.props
```diff
@@ -3,7 +3,7 @@
 		<ClientOfficialVersion>3.55.0</ClientOfficialVersion>
 		<ClientPreviewVersion>3.56.0</ClientPreviewVersion>
 		<ClientPreviewSuffixVersion>preview.0</ClientPreviewSuffixVersion>
-		<DirectVersion>3.41.2</DirectVersion>
+		<DirectVersion>3.41.3</DirectVersion>
```
NaluTripician, line 6: Are there any other changes with this direct package? If so, can we put that in the PR description?

aavasthy: No per se any changes in direct code but we need some shared files updates for this PR.

Microsoft.Azure.Cosmos/src/Resource/Settings/IndexingPolicy.cs
```diff
@@ -591,6 +591,114 @@ public int GetHashCode(IndexingPolicy indexingPolicy)
...
+            public static readonly VectorIndexPathEqualityComparer Singleton = new VectorIndexPathEqualityComparer();
+
+            public bool Equals(VectorIndexPath vectorIndexPath1, VectorIndexPath vectorIndexPath2)
+            {
+                if (Object.ReferenceEquals(vectorIndexPath1, vectorIndexPath2))
+                {
+                    return true;
+                }
+
+                if (vectorIndexPath1 == null || vectorIndexPath2 == null)
```
allenkim0129: Nit: Instead of using `if` here, it would be better to use `else if`, since this condition assumes both index paths are not equal. Ideally, put related conditions in the same block

PR #5527, opened 2025-12-10
[Internal] Logging: Refactors CFP traces to contain more information

simorenoh:
If there are errors here that don't directly translate within the exception, there is useful context that is potentially missing for debugging here. Making these changes to add some more information that can be useful in debugging CFP scenarios for users.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,8 +1,8 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 	<PropertyGroup>
-		<ClientOfficialVersion>3.56.0</ClientOfficialVersion>
+		<ClientOfficialVersion>3.56.1</ClientOfficialVersion>
 		<ClientPreviewVersion>3.57.0</ClientPreviewVersion>
-		<ClientPreviewSuffixVersion>preview.0</ClientPreviewSuffixVersion>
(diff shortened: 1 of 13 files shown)


Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/PartitionLoadBalancerCore.cs
```diff
@@ -89,14 +89,14 @@ private async Task RunAsync()
                             catch (Exception e)
                             {
                                 Extensions.TraceException(e);
-                                DefaultTrace.TraceError("Partition load balancer lease add/update iteration failed");
+                                DefaultTrace.TraceError($"Failed to acquire lease: Id={lease.Id}, Owner={lease.Owner ?? "null"}, Timestamp={lease.Timestamp}");
```
kirankumarkolli: nit: Recommended pattern is index based arguments {0} 
Reason: Perf but since is rare may be acceptable

simorenoh: changed it

NaluTripician (approved):

PR #5535, opened 2025-12-18
[Thin Client Integration]: Adds thinclient header for refresh account data requests.

### LINKED ISSUE #5536: [ThinClient Integration]: Missing thinclient header for refresh account data requests leads to stale endpoint information
**Describe the bug**
In ThinClient mode when endpoint refresh happens or request encounters a HttpRequestException , the ClientRetryPolicy triggers a location refresh by calling GlobalEndpointManager.RefreshLocationAsync(). This refresh fetches updated account information from the service. However, the refresh request was missing the thinclient endpoint discovery header (x-ms-cosmos-thinclient-endpoint-discovery).
Without this header, the service returned standard endpoint information instead of thin client-specific endpoint information. Due to this after a failover or endpoint refresh, the SDK was not receiving the correct thin client endpoints.

aavasthy:
# Pull Request Template

## Description
In ThinClient mode when endpoint refresh happens or request encounters a HttpRequestException , the ClientRetryPolicy triggers a location refresh by calling GlobalEndpointManager.RefreshLocationAsync(). This refresh fetches updated account information from the service. However, the refresh request was missing the thinclient endpoint discovery header (x-ms-cosmos-thinclient-endpoint-discovery).
Without this header, the service returned standard endpoint information instead of thin client-specific endpoint information. Due to this after a failover or endpoint refresh, the SDK was not receiving the correct thin client endpoints.
As part of this PR added thinclient header with refresh request for get database account info.
Tested this change for region failover scenario and tested it for endpoint manager refresh scenario.

## Type of change

## Closing issues

To automatically close an issue: closes #5536

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -136,7 +136,7 @@ namespace Microsoft.Azure.Cosmos
         internal readonly AuthorizationTokenProvider cosmosAuthorization;
 
         private readonly bool isThinClientFeatureFlagEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
-        private bool isThinClientEnabled;
+        internal bool isThinClientEnabled;
 
         // Gateway has backoff/retry logic to hide transient errors.
         private RetryPolicy retryPolicy;
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
@@ -545,7 +545,7 @@ namespace Microsoft.Azure.Cosmos.Tests
                     EnablePartitionLevelFailover = ppaf
                 };
 
-                CosmosClient cosmosClient = cosmosClientBuilder.Build(new MockDocumentClient(policy));
+                CosmosClient cosmosClient = cosmosClientBuilder.Build(new MockDocumentClient(policy, thinClient));
 
                 CosmosClientOptions cosmosClientOptions = cosmosClient.ClientOptions;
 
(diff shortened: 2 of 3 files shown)


Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6538,6 +6539,16 @@ await this.cosmosAuthorization.AddAuthorizationHeaderAsync(
                         "GET",
                         AuthorizationTokenType.PrimaryMasterKey);
 
+                    // Added the thinclient endpoint discovery header for account data refresh requests.
+                    // This header signals to the service that the client supports thin client mode
+                    // and needs thinclient-specific endpoint information in the response.
+                    if (this.isThinClientFeatureFlagEnabled)
+                    {
+                        request.Headers.Add(
```
kundadebdatta: Instead, can we add the header in the `headersCollection` itself ? That way, the following `foreach` loop will take care of adding the header anyway right.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -135,7 +135,8 @@ internal partial class DocumentClient : IDisposable, IAuthorizationTokenProvider
         //Auth
         internal readonly AuthorizationTokenProvider cosmosAuthorization;
 
-        private bool isThinClientEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
+        private readonly bool isThinClientFeatureFlagEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
```
kundadebdatta, line 138: Why do we need two boolean flags here ? Can we not leverage just the `isThinClientEnabled` flag ? That alone should suffice isn't it ?

aavasthy: private readonly bool isThinClientFeatureFlagEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
 private bool isThinClientEnabled;
 
 
 Keeping these two separate because isThinClientFeatureFlagEnabled is just when the feature flag is enabled whereas isThinClientEnabled is a combination of flag enabled + thinclient endpoint available+ gateway mode

Microsoft.Azure.Cosmos/src/DocumentClient.cs
aavasthy: updated

PR #5543, opened 2025-12-19
Query: Fixes GetItemQueryIterator to honor the supplied (optional) FeedRange

### LINKED ISSUE #5451: Fix support for container.GetItemQueryIterator using Serialized FeedRange [StartEPK, EndEPK] input
**Issue Description**

The .NET SDK currently has a bug in how it handles Query requests with custom serialized FeedRange values that specify StartEPK and EndEPK.

When such a FeedRange overlaps only a portion of a physical partition (rather than the full partition), the SDK incorrectly maps the request to the entire physical partition.

Specifically:

For partial FeedRange inputs, the SDK is not populating the headers x-ms-start-epk and x-ms-end-epk.

Instead, it only sets x-ms-documentdb-partitionkeyrangeid (the physical partition ID), which results in broader-than-intended scoping.

This leads to incorrect query behavior where partial ranges are expanded unintentionally to the full partition.

Acceptance Criteria

The SDK must correctly honor the contract of FeedRange:

- When a StartEPK and EndEPK are provided, the SDK must populate x-ms-start-epk and x-ms-end-epk headers accordingly.

- Queries should be scoped only to the specified logical range, not the entire physical partition.

- Unit and integration tests must validate that partial FeedRange queries respect the defined boundaries across different partition overlap scenarios.

- Regression tests confirm that full partition FeedRange queries continue to behave as expected.

Code sample of the issue

          var queryDefinition = new QueryDefinition("Select * from c");
          string feedRangeSerialization = JsonConvert.SerializeObject(new 
                  { Range = new { min = "0", max = "1DF48064A289030124119FB7736C62FB"} });
          FeedRange feedRange = FeedRange.FromJsonString(feedRangeSerialization);

          // Query multiple items from container
          using FeedIterator<JObject> feed = container.GetItemQueryIterator<JObject>(
              feedRange,
              queryDefinition);

            // Iterate query result pages
            while (feed.HasMoreResults)
            {
                Console.WriteLine($"ReadNextAsync");
                FeedResponse<JObject> response = await feed.ReadNextAsync();

                Console.WriteLine(response.Diagnostics.ToString());
                // Iterate query results
                foreach (JObject item in response)
                {
                    Console.WriteLine(item);
                }
            }`

adityasa:
# Query: Fixes GetItemQueryIterator to honor the supplied (optional) FeedRange

## Description

This change supports query execution using serialized FeedRange supplied to GetItemQueryIterator. It limits the query execution to specified feed ranges by using start epk and end epk values.
This is done by adjusting PartitionKeyRange of each partition to be >= min and <= max of the supplied range after target partitions have been identified. While this is not logically accurate, since PartitionKeyRange represents the ranges of a physical partition, and technically there may not be a physical partition with exact ranges after they have been adjusted using user supplied range and (the adjusted range is a subset of a physical partition range), this is rather only a semantic issue. Also, once the target range (adjusted or not) is obtained, they are anyways converted to FeedRangeEpk which represents a slice of a partition.
This logic of "clipping" a partition is similar to what we do in case of Hierarchical partition and we repurpose the existing class for locality of knowledge.

Note: QueryRequestOptions has a FeedRange property. However this is NOT the property that gets used when the GetItemQueryIterator API with FeedRange is called. In fact, QueryRequestOptions.FeedRange is an internal property.

I also explored the option of populating the start-epk and end-epk headers in QueryRequestOptions::PopulateRequestOptions as an alternative to "clipping" the PartitionKeyRange from target ranges.
This has following challenges:
1. Generally RequestInvokerHandler is in charge of setting epk and partition key related fields. It does so for all operations (including query). Breaking this pattern in a subset of cases is difficult from maintainability perspective.
2. QueryRequestOptions is user input and should be treated as immutable. Modifying it to store FeedRange breaks this principal

Caveats:
Please note due to the difference in types being utilized internally and externally (deserialized type), we are _only_ able to support feedranges with min being inclusive and max being exclusive - [min, max)
Moreover there is no validation against "unsupported values" and they are automatically interpreted as [min, max).

## Type of change

## Closing issues

Closes #5451

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -135,7 +135,8 @@ namespace Microsoft.Azure.Cosmos
         //Auth
         internal readonly AuthorizationTokenProvider cosmosAuthorization;
 
-        private bool isThinClientEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
+        private readonly bool isThinClientFeatureFlagEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
+        internal bool isThinClientEnabled;
 
         // Gateway has backoff/retry logic to hide transient errors.
         private RetryPolicy retryPolicy;
@@ -1058,7 +1059,7 @@ namespace Microsoft.Azure.Cosmos
                 this.ConnectionPolicy.EnablePartitionLevelFailover = this.accountServiceConfiguration.AccountProperties.EnablePartitionLevelFailover.Value;
             }
 
-            this.isThinClientEnabled = (this.ConnectionPolicy.ConnectionMode == ConnectionMode.Gateway) &&
+            this.isThinClientEnabled = this.isThinClientFeatureFlagEnabled && (this.ConnectionPolicy.ConnectionMode == ConnectionMode.Gateway) &&
                 (this.accountServiceConfiguration.AccountProperties?.ThinClientWritableLocationsInternal?.Count ?? 0) > 0;
             this.ConnectionPolicy.EnablePartitionLevelCircuitBreaker |= this.ConnectionPolicy.EnablePartitionLevelFailover;
             this.ConnectionPolicy.UserAgentContainer.AppendFeatures(this.GetUserAgentFeatures());
@@ -6538,6 +6539,14 @@ namespace Microsoft.Azure.Cosmos
                         "GET",
                         AuthorizationTokenType.PrimaryMasterKey);
 
+                    // Added the thinclient endpoint discovery header for account data refresh requests.
+                    // This header signals to the service that the client supports thin client mode
+                    // and needs thinclient-specific endpoint information in the response.
+                    if (this.isThinClientFeatureFlagEnabled)
+                    {
+                        headersCollection[ThinClientConstants.EnableThinClientEndpointDiscoveryHeaderName] = true.ToString();
+                    }
+
                     foreach (string key in headersCollection.AllKeys())
                     {
                         request.Headers.Add(key, headersCollection[key]);
@@ -6822,7 +6831,7 @@ namespace Microsoft.Azure.Cosmos
                     connectionPolicy: this.ConnectionPolicy,
                     httpClient: this.httpClient,
                     cancellationToken: this.cancellationTokenSource.Token,
-                    isThinClientEnabled: this.isThinClientEnabled);
+                    isThinClientEnabled: this.isThinClientFeatureFlagEnabled);
 
             this.accountServiceConfiguration = new CosmosAccountServiceConfiguration(accountReader.InitializeReaderAsync);
 
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CrossPartition/HybridSearch/HybridSearchCrossPartitionQueryPipelineStage.cs
@@ -653,6 +653,7 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.CrossPartition.HybridSearch
             }
 
             string rewrittenQuery = FormatComponentQueryTextWorkaround(queryInfo.RewrittenQuery, statistics, componentCount);
+            HybridSearchDebugTraceHelpers.TraceComponentQueryText(rewrittenQuery);
 
             QueryInfo result = new QueryInfo()
             {
@@ -955,6 +956,17 @@ namespace Microsoft.Azure.Cosmos.Query.Core.Pipeline.CrossPartition.HybridSearch
 #pragma warning disable CS0162 // Unreachable code detected
 
             [Conditional("DEBUG")]
+            public static void TraceComponentQueryText(string queryText)
+            {
+                if (Enabled)
+                {
+                    System.Diagnostics.Trace.WriteLine("Component Query Text:");
+                    System.Diagnostics.Trace.WriteLine(queryText);
+                    System.Diagnostics.Trace.WriteLine("\n");
+                }
+            }
+
+            [Conditional("DEBUG")]
             public static void TraceQuerySpec(SqlQuerySpec querySpec)
             {
                 if (Enabled)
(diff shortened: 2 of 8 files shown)


Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
```diff
@@ -108,5 +109,75 @@ public static FeedRangeInternal LimitFeedRangeToSinglePartition(PartitionKey? pa
...
+
+        /// <summary>
+        /// Limits the partition key ranges to fit within the provided EPK ranges.
+        /// Computes the overall min and max from the provided ranges, then trims each partition key range to fit within those boundaries.
+        /// </summary>
+        /// <param name="partitionKeyRanges">The list of partition key ranges to trim</param>
+        /// <param name="providedRanges">The EPK ranges to use as boundaries</param>
+        /// <returns>A list of trimmed partition key ranges that fit within the provided ranges</returns>
+        public static List<Documents.PartitionKeyRange> LimitPartitionKeyRangesToProvidedRanges(
+            List<Documents.PartitionKeyRange> partitionKeyRanges,
```
adityasa, line 122: Documents.PartitionKeyRange and Documents.Routing.Range are not directly comparable (despite the code below). This is because one of the structs only supports MinInclusive and MaxExclusive whereas the other is open ended (both min and max values can be either inclusive or exclusive).

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
```diff
@@ -108,5 +109,75 @@ public static FeedRangeInternal LimitFeedRangeToSinglePartition(PartitionKey? pa
...
+
+                // Trim max: use the lesser of range.Max and overallMax
+                if (maxComparer.Compare(new Range<string>(range.MaxExclusive, range.MaxExclusive, true, true),
+                                        new Range<string>(overallMax, overallMax, true, true)) > 0)
+                {
+                    trimmedMax = overallMax;
+                }
+
+                trimmedRanges.Add(
+                    new Documents.PartitionKeyRange
```
adityasa, line 179: This is technically not correct PartitionKeyRange, since no logical partition key value might actually correspond to the values below.
This is actually effective partition key values.
This seems like a theoretical concern only since at the callsite, we always seem to turn PartitionKeyRange into FeedRangeEpk. We could return a different type from these functions.

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
```diff
@@ -108,5 +109,75 @@ public static FeedRangeInternal LimitFeedRangeToSinglePartition(PartitionKey? pa
...
+                    {
+                        Id = range.Id,
+                        MinInclusive = trimmedMin,
```
adityasa, line 182: Due to the return type limitation, we implicitly lose nuance around min value's inclusive|exclusive-ness and max value's inclusive|exclusive-ness from provided ranges.

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
```diff
@@ -231,14 +231,16 @@ public override async Task<List<PartitionKeyRange>> GetTargetPartitionKeyRangeBy
...
-                return await this.GetTargetPartitionKeyRangesAsync(
+                List<PartitionKeyRange> ranges = await this.GetTargetPartitionKeyRangesAsync(
                     resourceLink,
                     collectionResourceId,
-                    ranges,
+                    providedRanges,
                     forceRefresh,
                     childTrace);
+
+                return QueryRangeUtils.LimitPartitionKeyRangesToProvidedRanges(ranges, providedRanges);
```
adityasa, line 243: >return QueryRangeUtils.LimitPartitionKeyRangesToProvidedRanges(ranges, providedRanges); [](http://example.com/codeflow?start=16&length=87)

Another alternative is to do this at the callsite. However there may be too many callers for this method.

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
```diff
@@ -108,5 +109,75 @@ public static FeedRangeInternal LimitFeedRangeToSinglePartition(PartitionKey? pa
...
+                // Trim min: use the greater of range.Min and overallMin
+                if (minComparer.Compare(new Range<string>(range.MinInclusive, range.MinInclusive, true, true),
+                                        new Range<string>(overallMin, overallMin, true, true)) < 0)
+                {
+                    trimmedMin = overallMin;
+                }
+
+                // Trim max: use the lesser of range.Max and overallMax
+                if (maxComparer.Compare(new Range<string>(range.MaxExclusive, range.MaxExclusive, true, true),
+                                        new Range<string>(overallMax, overallMax, true, true)) > 0)
```
adityasa: Needs to be fixed up to exclude max (per PR documentation).

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
```diff
@@ -108,5 +109,75 @@ public static FeedRangeInternal LimitFeedRangeToSinglePartition(PartitionKey? pa
...
+            // Trim each range to fit within the overall boundaries
+            List<Documents.PartitionKeyRange> trimmedRanges = new List<Documents.PartitionKeyRange>(partitionKeyRanges.Count);
+            foreach (Documents.PartitionKeyRange range in partitionKeyRanges)
+            {
+                string trimmedMin = range.MinInclusive;
+                string trimmedMax = range.MaxExclusive;
+
+                // Trim min: use the greater of range.Min and overallMin
+                if (minComparer.Compare(new Range<string>(range.MinInclusive, range.MinInclusive, true, true),
+                                        new Range<string>(overallMin, overallMin, true, true)) < 0)
```
adityasa: Needs to be updated to include min (per PR documentation).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/QueryFeedRangeTest.cs
```diff
@@ -0,0 +1,512 @@
...
+    using System.Collections.ObjectModel;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos;
+    using Microsoft.Azure.Cosmos.CosmosElements;
+    using Microsoft.Azure.Documents;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [TestClass]
+    public class QueryFeedRangeTest : QueryTestsBase
```
adityasa, line 14: >QueryFeedRangeTest [](http://example.com/codeflow?start=17&length=18)

Match file name with class name.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/QueryFeedRangeTest.cs
```diff
@@ -0,0 +1,512 @@
...
+            }
+
+            return epkList;
+        }
+
+        private static FeedRange CreateFeedRangeFromEpkRange(string minEpk, string maxEpk)
+        {
+            string feedRangeSerialization = Newtonsoft.Json.JsonConvert.SerializeObject(new
+            {
+                Range = new { min = minEpk, max = maxEpk }
```
adityasa, line 393: >Range = new { min = minEpk, max = maxEpk } [](http://example.com/codeflow?start=16&length=42)

Exercise inclusive exclusive combinations. At least to document unsupported scenarios.

adityasa: Added Debug.Assert in the product code.

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
adityasa: The inclusion/exclusion doesn't matter for comparison purposes since both ranges are being constructed on the fly.

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
adityasa: The inclusion/exclusion doesn't matter for comparison purposes since both ranges are being constructed on the fly.

Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
adityasa: Sticking with this approach.

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
adityasa: Added comment in PR description. Will be listed in known issues.

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
adityasa: Semantic issue only. The return type on CosmosQueryClient interface needs to be changed from PartitionKeyRange to just Range<string>.

Microsoft.Azure.Cosmos/src/Query/Core/QueryRangeUtils.cs
adityasa: Added an assert to this effect. Also will be listed as a known issue.

PR #5544, opened 2025-12-23
Query : Fixes LINQ API to support builtin functions - ARRAY_CONTAINS_ALL and ARRAY_CONTAINS_ANY

adityasa:
# Query : Fixes LINQ API to support builtin functions - ARRAY_CONTAINS_ALL and ARRAY_CONTAINS_ANY

## Description

This change implements support for ARRAY_CONTAINS_ALL and ARRAY_CONTAINS_ANY system functions to LINQ API.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Batch/BatchAsyncBatcher.cs
+++ b/Microsoft.Azure.Cosmos/src/Batch/BatchAsyncBatcher.cs
@@ -36,6 +36,7 @@ namespace Microsoft.Azure.Cosmos
         private readonly int maxBatchOperationCount;
         private readonly InterlockIncrementCheck interlockIncrementCheck = new InterlockIncrementCheck();
         private readonly CosmosClientContext clientContext;
+        private readonly ItemRequestOptions options;
         private long currentSize = 0;
         private bool dispatched = false;
         private bool isClientEncrypted = false;
@@ -49,7 +50,8 @@ namespace Microsoft.Azure.Cosmos
             CosmosSerializerCore serializerCore,
             BatchAsyncBatcherExecuteDelegate executor,
             BatchAsyncBatcherRetryDelegate retrier,
-            CosmosClientContext clientContext)
+            CosmosClientContext clientContext,
+            ItemRequestOptions options = null)
         {
             if (maxBatchOperationCount < 1)
             {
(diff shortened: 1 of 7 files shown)


Microsoft.Azure.Cosmos/src/Linq/BuiltinFunctions/OtherBuiltinSystemFunctions.cs
```diff
@@ -173,22 +173,84 @@ protected override SqlScalarExpression VisitExplicit(MethodCallExpression method
...
+                        new Type[]{typeof(object), typeof(object[])}
+                    })
+            {
+            }
+
+            protected override SqlScalarExpression VisitImplicit(MethodCallExpression methodCallExpression, TranslationContext context)
+            {
+                if (methodCallExpression.Arguments.Count != 2)
+                {
+                    return null;
```
neildsh, line 192: >return null; [](http://example.com/codeflow?start=20&length=12)

for my own edification, why return null instead of throwing an exception? Is it because we are relying on the binder for the syntax check?

neildsh (approved):

Microsoft.Azure.Cosmos/src/Linq/BuiltinFunctions/OtherBuiltinSystemFunctions.cs
adityasa: Returning null ends up being a signal for the higher level component to throw the exception. This is a (rather unfortunate and confusing) pattern with these builtin function translators (string function file, array function file and this file). I do plan on addressing this so the code flow and intent is clearer.

Microsoft.Azure.Cosmos/src/Linq/BuiltinFunctions/OtherBuiltinSystemFunctions.cs
```diff
@@ -17,9 +17,9 @@ namespace Microsoft.Azure.Cosmos.Linq
 
     internal static class OtherBuiltinSystemFunctions
     {
-        private class RRFVisit : SqlBuiltinFunctionVisitor
+        private class RRFVisitor : SqlBuiltinFunctionVisitor
```
sboshra: >RRF [](http://example.com/codeflow?start=22&length=3)

RrfVisitor

kirankumarkolli (approved):

PR #5548, opened 2026-01-05
[Internal] Testing: Fixes CFP container cleanup

yash2710:
# Pull Request Template

## Description

Add cleanup steps for monitored and leased containers in ChangeFeedProcessor tests

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/BaseChangeFeedClientHelper.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/BaseChangeFeedClientHelper.cs
@@ -4,6 +4,7 @@
 
 namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
 {
+    using System.Net;
     using System.Threading.Tasks;
     using Microsoft.Azure.Cosmos.SDK.EmulatorTests;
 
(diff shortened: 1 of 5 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/BaseChangeFeedClientHelper.cs
```diff
@@ -24,5 +25,23 @@ public async Task ChangeFeedTestInit(string leaseContainerPk = "/id")
...
             this.LeaseContainer = response;
         }
+
+        public new async Task TestCleanup()
+        {
+            if (this.LeaseContainer != null)
+            {
+                try
+                {
+                    await this.LeaseContainer.DeleteContainerStreamAsync(
```
NaluTripician, line 35: Is this container in the default preexisting database for the integration tests?

yash2710: No, it is created in the base class with `ChangeFeedTestInit` on test run

NaluTripician: Ok, if it is not needed to stay in between tests lets delete the database instead of the container. Other than that, LGTM

yash2710: Well, seems the database is already being deleted in `base.TestCleanup()`
In that case we don't need this change at all?

PR #5552, opened 2026-01-13
[Internal] DTS: Adds initial skeleton for DTS core data models

Meghana-Palaparthi:
## Description

This pull request introduces the initial SDK‑side implementation for Distributed Transactions in the Azure Cosmos DB SDK. It lays the groundwork by adding the core abstractions, foundational classes, and operation/result models needed to support distributed write transactions across partitions and collections.

The primary goal of this change is to establish the structural framework and API surface for distributed transactions. As such, several components are intentionally scaffolding-only, with full behavior to be implemented in subsequent phases.

Since this API is not yet intended for public consumption, it is currently gated behind a preprocessor directive - enabled for internal end‑to‑end testing while remaining internal by default.

## Type of change

[] Bug fix (non-breaking change which fixes an issue)
[✓] New feature (non-breaking change which adds functionality)
[] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[] This change requires a documentation update


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    /// <summary>
+    /// Represents a distributed transaction that will be performed across partitions and/or collections. 
+    /// </summary>
+    public abstract class DistributedTransaction
+    {
+        /// <summary>
+        /// Commits the distributed transaction.
+        /// </summary>
+        public abstract void CommitTransaction();
```
Praveen-Msft: Is there a plan to support rollback txn?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
```diff
@@ -0,0 +1,88 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    using System;
+    using System.Collections.Generic;
+    using Microsoft.Azure.Documents;
+
+    internal class DistributedWriteTransactionCore : DistributedWriteTransaction
```
Praveen-Msft, line 14: Do we support concurrent distributed txns? aka, client initiating multiple distributed txns? with the current info, does it make sense to limit `DistributedWriteTransactionCore` to be singleton?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
Meghana-Palaparthi: No, there's no plan to support rollback explicitly as a user‑initiated operation. Rollbacks are and will be handled internally by the coordinator

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
Meghana-Palaparthi: At a high level the design does not prohibit concurrent distributed transactions per client

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Represents an operation on a document whichwill be executed as a part of a distributed transaction.
+    /// </summary>
+    internal class DistributedTransactionOperation
+    {
+        public DistributedTransactionOperation(
+            Documents.OperationType operationType,
+            PartitionKey partitionKey,
+            string database,
```
jeet1995, line 20: 1. Is using `string` type intentional here instead of `CosmosContainer` or `CosmosDatabase`? `string` does not encourage encapsulation or reusability.
2. It will make it easier for reviewers to work of a real-world use case and correlate with the APIs you introduce here. How would an application developer use your APIs.
3. Wouldn't a `DistributedTransactionRequestOptions` be required too?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
```diff
@@ -0,0 +1,32 @@
+﻿// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    internal abstract class DistributedWriteTransaction : DistributedTransaction
```
jeet1995: Is non-distributed `Batch` scoped out here?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// List of operations in a distributed transaction
+        /// </summary>
+        protected List<DistributedTransactionOperation> operations;
+
+        internal DistributedWriteTransactionCore()
+        {
+            this.operations = new List<DistributedTransactionOperation>();
+        }
+
+        public override DistributedTransaction Create<T>(string database, string collection, PartitionKey partitionKey)
```
jeet1995: How are we modelling a `Conditional`?

jeet1995: Is there a requirement for a per-`DistributedTransaction` request option or existing `*ItemRequestOption`? `SessionToken` for example is a per-(logical partition) concept when the end user has to define the session, custom (de)serializer etc. to name a few.

Microsoft.Azure.Cosmos/src/CosmosClient.cs
```diff
@@ -1162,6 +1162,20 @@ public virtual FeedIterator GetDatabaseQueryStreamIterator(
...
+        /// <summary>
+        /// Creates a new instance of a distributed write transaction.
+        /// </summary>
+        /// <returns>An instance of Distributed transaction.</returns>
+#if INTERNAL
+        public 
+#else
+        internal
+#endif
+        virtual DistributedTransaction CreateDistributedTransaction()
```
jeet1995: For the DTC endpoint - will it be the same as the Gateway Service Endpoint (I'd like to think so).

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
```diff
@@ -0,0 +1,88 @@
...
+                    partitionKey,
+                    database,
+                    collection));
+            return this;
+        }
+
+        public override DistributedTransaction Patch(string database, string collection, PartitionKey partitionKey)
+        {
+            this.operations.Add(
+                new DistributedTransactionOperation(
```
jeet1995: `Patch` semantics have more nuances which have to be modelled. See - https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update and https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/ItemManagement/Program.cs#L520

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
Meghana-Palaparthi: yeah, using `string` is intentional here. We will resolve collection RID from the string. Right now, from the design, It seems like database is not required. I'll remove this param

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
Meghana-Palaparthi: This PR is scoped to distributed transactions only.
Non‑distributed `TransactionalBatch` behavior is unchanged and out of scope.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
Meghana-Palaparthi: We have not finalized the contract for `ConditionalCheck` yet. That will be finalized and implemented as a part of phase 2.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
Meghana-Palaparthi: Yeah, each operation needs it's params revised. I added there just as placeholders for now

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
Meghana-Palaparthi: Yeah, we need request options too. And I'll update the design with a real-world example of how users might build a DTS

kirankumarkolli: string vs OO is interesting aspect.

IMO at RAW API layer OOB type is overhead forces applications to unnecessarily create them. 
Future builders API are interesting, order matters in execution also.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
kirankumarkolli: > Is non-distributed Batch scoped out here?
Can you please expand more? 
Distributed here doesn't just mean across containers, its referring to cross-partition (can be any)

Microsoft.Azure.Cosmos/src/CosmosClient.cs
kirankumarkolli: Its hosted on gateway (compute), we will follow the same pattern as today (i.e. use regional endpoint names)

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
```diff
@@ -0,0 +1,43 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    using System.Collections.Generic;
+
+    internal abstract class DistributedWriteTransaction : DistributedTransaction
+    {
+        public abstract DistributedTransaction Create<T>(
```
kirankumarkolli: Is it for builder pattern? 

Typically we isolate builder from regular API. 
Also for writes the retrun type also should be WriteTransaction.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
jeet1995: Can a `TransactionalBatch` be part of a distributed transaction along with other point-write operation types or is this intentionally kept out of scope?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
Meghana-Palaparthi: Yeah, it makes more sense to have return type as WriteTransaction to restrict the user to only add write-specific operations to the transaction. I've updated it return types

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
Meghana-Palaparthi: I don't think that's in scope

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
```diff
@@ -0,0 +1,48 @@
...
+    using System.Collections.Generic;
+
+#if INTERNAL
+    public 
+#else
+    internal
+#endif
+    abstract class DistributedWriteTransaction : DistributedTransaction
+    {
+        public abstract DistributedWriteTransaction Create<T>(
```
kirankumarkolli: nit: Just to be consistent prefer `Item' suffix

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
```diff
@@ -0,0 +1,26 @@
...
+        public 
+#else
+    internal
+#endif
+    abstract class DistributedTransaction
+    {
+        /// <summary>
+        /// Commits the distributed transaction.
+        /// </summary>
+        public abstract void CommitTransaction();
```
kirankumarkolli: Will it be a blocking execution? 

Ideally Async method like 

```suggestion
        public abstract Task<Response>CommitTransactionAsync();
```

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs
```diff
@@ -0,0 +1,48 @@
...
+            string database,
+            string collection,
+            PartitionKey partitionKey,
+            T resource);
+
+        public abstract DistributedWriteTransaction Replace<T>(
```
kirankumarkolli: nit: Follow-up streaming API's later (non-blocker)

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
Meghana-Palaparthi: Yeah, this is async. I realized this and changed it in [PR #5570](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5570/changes#diff-ecb644e5955fc1e31cb3bf2a26abe457e417d0982e9ed573307d98619cb37e51R24). I'll update it here as well

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        public DistributedTransactionOperation(
+            Documents.OperationType operationType,
+            int operationIndex,
+            string database,
+            string container,
+            PartitionKey partitionKey,
+            string id = null)
+        {
+            this.OperationType = operationType;
```
kirankumarkolli, line 25: Assert that mandated arguments are not-null (optional ones in the caller which seems already happening)

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -0,0 +1,73 @@
...
+            this.StatusCode = statusCode;
+        }
+
+        internal DistributedTransactionOperationResult(DistributedTransactionOperationResult other)
+        {
+            this.StatusCode = other.StatusCode;
+            this.SubStatusCode = other.SubStatusCode;
+            this.ETag = other.ETag;
+            this.ResourceStream = other.ResourceStream;
+            this.SessionToken = other.SessionToken;
```
kirankumarkolli, line 34: ITrace?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <summary>
+        /// Gets a value indicating whether the HTTP status code returned by the operation indicates success.
+        /// </summary>
+        public virtual bool isSuccessStatusCode => ((int)this.StatusCode >= 200) && ((int)this.StatusCode <= 299);
+
+        /// <summary>
+        /// Gets the entity tag (ETag) associated with the operation result.
+        /// The ETag is used for concurrency control and represents the version of the resource.
+        /// </summary>
+        public virtual string ETag { get; internal set; }
```
kirankumarkolli, line 60: How compute can set it?

Meghana-Palaparthi: SDK deserializes the response and sets ETag internally during response parsing. Compute doesn't directly set it

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        public virtual bool IsSuccessStatusCode => ((int)this.StatusCode >= 200) && ((int)this.StatusCode <= 299);
+
+        /// <summary>
+        /// Gets the entity tag (ETag) associated with the operation result.
+        /// The ETag is used for concurrency control and represents the version of the resource.
+        /// </summary>
+        public virtual string ETag { get; internal set; }
+
+        internal virtual string SessionToken { get; set; }
```
kirankumarkolli: public?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -0,0 +1,109 @@
...
+        public virtual HttpStatusCode StatusCode { get; internal set; }
+
+        internal virtual SubStatusCodes SubStatusCode { get; }
+
+        internal IReadOnlyList<DistributedTransactionOperation> Operations { get; set; }
+
+        /// <summary>
+        /// Gets the error message associated with the distributed transaction response, if any.
+        /// </summary>
+        public virtual string ErrorMessage { get; internal set; }
```
kirankumarkolli, line 82: Diagnostics contract follow-up

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
```diff
@@ -0,0 +1,151 @@
...
+        protected List<DistributedTransactionOperation> operations;
+
+        internal DistributedWriteTransactionCore()
+        {
+            this.operations = new List<DistributedTransactionOperation>();
+        }
+
+        public override DistributedWriteTransaction CreateItem<T>(string database, string collection, PartitionKey partitionKey, T resource)
+        {
+            ValidateContainerReference(database, collection);
```
kirankumarkolli: Convention, for static methos please prefix with classname ex: DistributedWriteTransactionCore.ValidateContainerReference(..);

PR #5565, opened 2026-01-15
[Preview] SemanticReranking: Adds HttpTimeout Policy to Reranking API

### LINKED ISSUE #5564: Semantic Reranking: Request Timeouts
The Semantic Reranking API needs to add default request timeout. This will initially be 5 seconds however this time may be changed in the future. The SDK should also accept a request timeout as a request option that can override the default value.

NaluTripician:
# Pull Request Template

## Description

This pull request introduces support for configurable request timeouts and improved fault injection for inference service operations (such as semantic reranking) in Cosmos DB. The changes allow users to set a custom timeout for inference requests, ensure proper handling of timeouts and retries, and enhance testing capabilities through chaos/fault injection mechanisms.

### Inference service timeout configuration and handling

* Added a new `InferenceRequestTimeout` property to `CosmosClientOptions`, allowing users to configure the timeout for inference service operations separately from standard requests. Default is 5 seconds.
* Added a fluent API method `WithInferenceRequestTimeout` to `CosmosClientBuilder` for setting the inference request timeout.
* Implemented `HttpTimeoutInferencePolicy`, a timeout policy specific to inference requests, supporting single-attempt requests with no retries and future extensibility.
* Updated `InferenceService` to use the configured timeout and policy, ensuring the HTTP client timeout is set appropriately and handling timeout exceptions by throwing Cosmos-specific exceptions with detailed context.
* As of now, there will be no retries for inference call, however this PR adds the infrastructure to add retry capabilities in the future.

### Shared HTTP retry pattern abstraction

* **Extracted `HttpTimeoutPolicyHelper.ExecuteWithTimeoutAsync<T>()`** — a new shared static helper that encapsulates the common timeout/retry loop pattern previously duplicated between `CosmosHttpClientCore` and `InferenceService`. The helper manages:
  - Timeout enumerator initialization and advancement
  - Linked `CancellationTokenSource` creation with per-attempt `CancelAfter`
  - User cancellation propagation (`OperationCanceledException` when user token is cancelled)
  - Out-of-retries detection
  - Inter-retry delays
* **Refactored `CosmosHttpClientCore.SendHttpHelperAsync`** to use the new `HttpTimeoutPolicyHelper`, with callbacks for:
  - Request execution (including fault injection and statistics recording)
  - Response-based retry decisions (`ShouldRetryBasedOnResponse`)
  - Exception handling (`IsSafeToRetry`, `WebException` classification, 503 timeout conversion)
* **Refactored `InferenceService.SemanticRerankAsync`** to use the same helper, with simpler callbacks (no response-based retry, timeout → `RequestTimeoutException`, HTTP failure → `ServiceUnavailableException`)
* Changed `InjectFaultsAsync` to accept `CancellationToken` instead of `CancellationTokenSource` (only `.Token` was used)
* Removed unused `IsOutOfRetries` private method (now handled by the shared helper)

### Fault injection and chaos testing enhancements

* Integrated chaos interceptor support into `InferenceService`, enabling fault and response delay injection for robust testing of inference operations. New helper methods `InjectFaultsAsync` and `InjectResponseDelayAsync` were added to facilitate this.
* Updated integration tests to use the new fault injection capabilities and refactored token credential initialization for consistency.

## Type of change

## Closing issues

To automatically close an issue: closes #5564


Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -387,6 +387,24 @@ public CosmosClientBuilder WithRequestTimeout(TimeSpan requestTimeout)
             return this;
         }
 
+        /// <summary>
+        /// Sets the request timeout for inference service operations (e.g., semantic reranking).
+        /// </summary>
+        /// <param name="inferenceRequestTimeout">A time to use as timeout for inference operations.</param>
+        /// <value>Default value is 5 seconds.</value>
```
Praveen-Msft, line 396: 5 seconds - is it because the backend takes 5sec? if so, we should update it to 6 seconds similar to other timeout policies.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyInference.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    internal sealed class HttpTimeoutPolicyInference : HttpTimeoutPolicy
+    {
+        private static readonly string Name = nameof(HttpTimeoutPolicyInference);
+        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> timeoutsAndDelays;
+
+        /// <summary>
+        /// Creates an instance with default timeout of 5 seconds and no retries.
+        /// </summary>
+        public static readonly HttpTimeoutPolicyInference InstanceDefault = new HttpTimeoutPolicyInference(TimeSpan.FromSeconds(5));
```
Praveen-Msft: I put in the comment in the CosmosClientBuilder.cs, I believe the 6seconds update should be here.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
NaluTripician: 5 seconds is based on the values on request duration that was given to us by semantic reranking team. (I believe this is the P999 time)

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutInferencePolicy.cs
```diff
@@ -0,0 +1,61 @@
...
+        public override int TotalRetryCount => this.timeoutsAndDelays.Count;
+
+        public override IEnumerator<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> GetTimeoutEnumerator()
+        {
+            return this.timeoutsAndDelays.GetEnumerator();
+        }
+
+        public override bool ShouldRetryBasedOnResponse(HttpMethod requestHttpMethod, HttpResponseMessage responseMessage)
+        {
+            // For now, no response-based retries
```
Praveen-Msft, line 54: does it mean it retries all failures?

NaluTripician: No, it will not retry any (which is the design we have discussed with the team)

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyInference.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Net.Http;
+
+    /// <summary>
+    /// Timeout policy for inference service requests without retry on timeout.
+    /// This policy allows for future expansion including handling errors and cross-regional retries.
+    /// </summary>
+    internal sealed class HttpTimeoutPolicyInference : HttpTimeoutPolicy
```
kirankumarkolli:
```suggestion
    internal sealed class HttpTimeoutInferencePolicy : HttpTimeoutPolicy
```

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
```diff
@@ -58,6 +61,12 @@ public InferenceService(CosmosClient client)
                 "AZURE_COSMOS_SEMANTIC_RERANKER_INFERENCE_SERVICE_MAX_CONNECTION_LIMIT",
                 inferenceServiceDefaultMaxConnectionLimit) ?? inferenceServiceDefaultMaxConnectionLimit;
 
+            // Get the inference timeout from client options, or use default
+            this.inferenceRequestTimeout = client.ClientOptions?.InferenceRequestTimeout ?? TimeSpan.FromSeconds(5);
```
kirankumarkolli: How about assert that client inferencetimout will not be null? WIll help have all defaults in one place only

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
```diff
@@ -113,46 +124,111 @@ public async Task<SemanticRerankResult> SemanticRerankAsync(
...
+                    // Build the request payload.
+                    Dictionary<string, object> body = this.AddSemanticRerankPayload(rerankContext, documents, options);
+
+                    message.Content = new StringContent(
+                        Newtonsoft.Json.JsonConvert.SerializeObject(body),
+                        Encoding.UTF8,
+                        RuntimeConstants.MediaTypes.Json);
+
+                    // Create linked cancellation token source to honor both user cancellation and timeout policy
+                    using CancellationTokenSource cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```
kirankumarkolli:
```suggestion
                    using CancellationTokenSource infereanceCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
```diff
@@ -113,46 +124,111 @@ public async Task<SemanticRerankResult> SemanticRerankAsync(
...
+                    catch (OperationCanceledException operationCanceledException)
+                    {
+                        // Throw if the user passed in cancellation was requested
+                        if (cancellationToken.IsCancellationRequested)
+                        {
+                            throw;
+                        }
+
+                        // Convert OperationCanceledException to timeout exception when the HTTP client throws it
+                        bool isOutOfRetries = !timeoutEnumerator.MoveNext();
```
kirankumarkolli: This is a common pattern on exisitng GW interactions also can it be abstreacted out and then shared across?

NaluTripician: Added abstraction and unit tests to ensure everything is functioning correctly.

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
```diff
@@ -93,6 +104,8 @@ public InferenceService(CosmosClient client)
         /// </summary>
         internal InferenceService(HttpMessageHandler messageHandler, Uri inferenceEndpoint, AuthorizationTokenProvider cosmosAuthorization)
         {
+            this.inferenceRequestTimeout = TimeSpan.FromSeconds(5);
```
kirankumarkolli: Define a static const with default value and reference it from CosmosClientOptions default initalization for consistency.

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
```diff
@@ -161,7 +224,11 @@ await this.cosmosAuthorization.AddInferenceAuthorizationHeaderAsync(
         /// <param name="httpClient">The HttpClient to configure.</param>
         private void CreateClientHelper(HttpClient httpClient)
         {
-            httpClient.Timeout = TimeSpan.FromSeconds(120);
+            // Set the timeout to be higher than the configured request timeout to allow the
+            // CancellationTokenSource to handle the timeout and retry logic properly
+            httpClient.Timeout = this.inferenceRequestTimeout > TimeSpan.FromSeconds(5)
```
kirankumarkolli: What's the goal here? 

If its to ensure that min is 35 seconds then 

```
            httpClient.Timeout = this.inferenceRequestTimeout >= TimeSpan.FromSeconds(35)
                ? this.inferenceRequestTimeout
                : TimeSpan.FromSeconds(35);
```

kirankumarkolli: Also please define the 35 secopnds as static const at the top.

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
```diff
@@ -113,46 +126,96 @@ public async Task<SemanticRerankResult> SemanticRerankAsync(
...
+                },
+                shouldRetryOnResult: null,
+                onException: (Exception exception, bool isOutOfRetries, TimeSpan requestTimeout) =>
+                {
+                    switch (exception)
+                    {
+                        case OperationCanceledException operationCanceledException:
+                            if (isOutOfRetries)
+                            {
+                                string errorMessage = $"Inference Service Request Timeout. Start Time UTC:{startDateTimeUtc}; Total Duration:{(DateTime.UtcNow - startDateTimeUtc).TotalMilliseconds} Ms; Request Timeout {requestTimeout.TotalMilliseconds} Ms; Http Client Timeout:{this.httpClient.Timeout.TotalMilliseconds} Ms; Activity id: {System.Diagnostics.Trace.CorrelationManager.ActivityId};";
```
kirankumarkolli: Why exception handling for Inference needs to be different? 

Any future reliability improvements might miss probabvly for one of the branches

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -383,121 +383,104 @@ private async Task<HttpResponseMessage> SendHttpHelperAsync(
...
-                                            headers: new Headers()
-                                            {
-                                                ActivityId = System.Diagnostics.Trace.CorrelationManager.ActivityId.ToString(),
-                                                SubStatusCode = SubStatusCodes.TransportGenerated503
-                                            },
-                                            trace: trace,
-                                            innerException: e);
-                                    }
-
-                                    throw;
```
kirankumarkolli, line 452: Throw is retaiing the original call stack not re-throwing right? Is ExceptionDispatchInfo changes needed?

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
kirankumarkolli, line 89: How about leverage `CosmosHttpClientCore` factory methods to create and maintain the `HttpClient`?

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
```diff
@@ -1677,93 +1677,117 @@ public abstract ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilder(
...
+        /// </example>
         public abstract Task<SemanticRerankResult> SemanticRerankAsync(
             string rerankContext,
             IEnumerable<string> documents,
-            IDictionary<string, object> options = null,
+            IDictionary<string, object> options = null,
             CancellationToken cancellationToken = default);
-#endif
-
+#else
```
kirankumarkolli: [WAITING] Why is it needed?

kirankumarkolli (changes requested):
Please check comments

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
NaluTripician: Done. Added `InferenceService.DefaultInferenceRequestTimeout` as an `internal static readonly` constant and referenced it from `CosmosClientOptions.InferenceRequestTimeout`, `HttpTimeoutInferencePolicy.InstanceDefault`, and the test constructor.

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
NaluTripician: Done. Defined `MinimumHttpClientTimeout` (35s) as a `private static readonly` constant and simplified the logic to match your suggestion:
```csharp
httpClient.Timeout = this.inferenceRequestTimeout >= InferenceService.MinimumHttpClientTimeout
    ? this.inferenceRequestTimeout
    : InferenceService.MinimumHttpClientTimeout;
```

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
NaluTripician: Good callout. The difference is intentional: `CosmosHttpClientCore` re-throws raw exceptions because `TransportHandler` catches and wraps them upstream. `InferenceService` has no such upstream handler, so it must wrap exceptions in `CosmosException` types (`RequestTimeoutException` / `ServiceUnavailableException`) directly for the caller.

Added a code comment above the `onException` callback explaining this rationale so future maintainers don't accidentally diverge or miss alignment opportunities.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
NaluTripician: The `onException` callback runs outside the original `catch` block (it's a delegate invoked from `HttpTimeoutPolicyHelper`), so `throw;` can't be used there — `ExceptionDispatchInfo.Capture(e).Throw()` is the correct way to preserve the original stack trace in this context. The existing code already uses `ExceptionDispatchInfo` for all re-throw paths in both `CosmosHttpClientCore` and `InferenceService`, so no changes are needed here.

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
NaluTripician: We already use `CosmosHttpClientCore.CreateHttpClientHandler()` to create the underlying `HttpMessageHandler`. The `HttpClient` itself is constructed directly because `InferenceService` needs different configuration (separate timeout, different default headers like the inference user agent, and a distinct auth scope). Wrapping this in a `CosmosHttpClientCore` factory would couple inference to gateway-specific behavior that doesn't apply here.

Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
NaluTripician: The `#else` block provides an `internal` stub that throws `NotSupportedException` for non-preview builds. This is needed because `ContainerCore` overrides this method, so the base class must declare it in all build configurations. Without it, the non-preview build would fail to compile. This follows the same pattern used for other preview-only APIs in the SDK (e.g., `VectorEmbeddingPolicy`).

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -383,121 +383,104 @@ private async Task<HttpResponseMessage> SendHttpHelperAsync(
...
                             datum.RecordHttpException(requestMessage, e, resourceType, requestStartTime);
-                            trace = datum.Trace;
+                            currentTrace = datum.Trace;
                         }
-                        bool isOutOfRetries = CosmosHttpClientCore.IsOutOfRetries(timeoutEnumerator);
 
-                        switch (e)
-                        {
-                            case OperationCanceledException operationCanceledException:
-                                // Throw if the user passed in cancellation was requested
```
kirankumarkolli, line 424: This section is removed, is it intentional? Might be breaking in behavior.

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
kirankumarkolli: Prox pattern as discussed offline

NaluTripician:
## Reverted `CosmosHttpClientCore` Refactoring

In commit e1b5884, I reverted the `CosmosHttpClientCore.SendHttpHelperAsync()` refactoring and removed all whitespace-only changes. Here's what was done and why:

### What was reverted
- **`CosmosHttpClientCore.cs`** — fully reverted to master. The `SendHttpHelperAsync()` method keeps its original inline retry loop, `InjectFaultsAsync` keeps its `CancellationTokenSource` parameter, and the `IsOutOfRetries` method is restored.
- **`Container.cs`** — reverted whitespace-only changes (no semantic content was changed).
- **`CosmosClientOptions.cs` / `CosmosClientBuilder.cs`** — removed ~200 lines of whitespace noise (trailing space / CRLF changes), keeping only the `InferenceRequestTimeout` property and `WithInferenceRequestTimeout()` method additions.

### Why
1. **`InferenceService` is completely independent from `CosmosHttpClientCore`** — it creates its own `HttpClient` with its own handler, timeout, headers, and auth scope. It never calls `CosmosHttpClientCore.SendHttpAsync()`. The shared `HttpTimeoutPolicyHelper` abstraction is useful, but forcing `CosmosHttpClientCore` to adopt it in the same PR introduces unnecessary risk to the critical gateway HTTP path.

2. **Unresolved reviewer concern** — @kirankumarkolli flagged a potential behavioral break in the refactored exception handling (`cancellationToken.IsCancellationRequested` check was moved from the switch case to the helper's catch filter). While analysis shows functional equivalence, this change to a hot production code path deserves its own focused review.

3. **Cleaner diff** — the PR now only touches files that are directly related to the inference timeout feature, making review straightforward.

### What's kept
- `HttpTimeoutPolicyHelper` (new shared helper) ✅
- `HttpTimeoutInferencePolicy` (new timeout policy) ✅
- `InferenceService` using the helper ✅
- All new tests ✅
- `CosmosClientOptions.InferenceRequestTimeout` ✅
- `CosmosClientBuilder.WithInferenceRequestTimeout()` ✅

### Test results
- All 14 `HttpTimeoutPolicyHelperTests` pass ✅
- All 4 `InferenceServiceTests` pass ✅
- All 15/16 `CosmosHttpClientCoreTests` pass ✅ (1 pre-existing failure unrelated to this PR — `HttpTimeoutThrow503TestAsync` has a .NET runtime version mismatch between `OperationCanceledException` and `TaskCanceledException`, present on master as well)

NaluTripician:
## Future Work: Migrate `CosmosHttpClientCore` to `HttpTimeoutPolicyHelper`

With `HttpTimeoutPolicyHelper` now available and independently tested, the next step is to migrate `CosmosHttpClientCore.SendHttpHelperAsync()` to use it. This should be done in a **separate, focused PR** to ensure proper review of behavioral equivalence.

### Scope of the future PR
1. **Refactor `SendHttpHelperAsync()`** to use `HttpTimeoutPolicyHelper.ExecuteWithTimeoutAsync()` with callbacks for:
   - `executeAsync`: fault injection, HTTP execution, response stats recording
   - `shouldRetryOnResult`: `timeoutPolicy.ShouldRetryBasedOnResponse()`
   - `onException`: `OperationCanceledException` → 408/503 conversion, `WebException` / `HttpRequestException` retry logic
2. **Change `InjectFaultsAsync`** parameter from `CancellationTokenSource` to `CancellationToken` (only `.Token` is used)
3. **Remove `IsOutOfRetries`** private method (handled by the helper)

### Key review points for that PR
- **User cancellation equivalence**: Verify that `catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)` in the helper is equivalent to the inline `if (cancellationToken.IsCancellationRequested) { throw; }` check in the current switch case
- **Stack trace preservation**: Confirm `throw;` inside the helper's catch block preserves the original stack trace when `onException` returns the same exception reference
- **Closure-captured state**: Review that `currentMethod` and `currentTrace` closures behave correctly across retry iterations
- **Integration test coverage**: Run gateway-mode integration tests to validate no behavioral regression in the hot HTTP path

### Benefits of the migration
- Eliminates duplicated retry/timeout loop logic between `CosmosHttpClientCore` and `InferenceService`
- Centralizes cancellation token management and retry delay logic
- Makes it easier to add new HTTP-based services (future inference endpoints, etc.) with consistent timeout/retry behavior

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
kirankumarkolli, line 396: Is it worth adding that its only one time no-retries?

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// Return true to retry, false to return the result. Only called when retries are available.
+        /// Pass null to never retry on a successful result.
+        /// </param>
+        /// <param name="onException">
+        /// Function called when an exception occurs (after user cancellation is already handled).
+        /// Receives the exception, whether retries are exhausted, and the per-attempt request timeout.
+        /// Return null to retry. Return the original exception to re-throw it with preserved stack trace.
+        /// Return a different exception to throw that instead.
+        /// </param>
+        internal static async Task<TResult> ExecuteWithTimeoutAsync<TResult>(
```
NaluTripician: Can we just use the method in HttpClientCore for this? We have a very similar one there and this class might not be needed.

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
```diff
@@ -163,7 +229,11 @@ await this.cosmosAuthorization.AddAuthorizationHeaderAsync(
         /// <param name="httpClient">The HttpClient to configure.</param>
         private void CreateClientHelper(HttpClient httpClient)
         {
-            httpClient.Timeout = TimeSpan.FromSeconds(120);
+            // Set the timeout to be at least MinimumHttpClientTimeout so the per-request
```
kirankumarkolli, line 232: Resultant might be changes from 120 -> 35, is it intentional?

Microsoft.Azure.Cosmos/src/Inference/InferenceService.cs
```diff
@@ -115,46 +136,91 @@ public async Task<SemanticRerankResult> SemanticRerankAsync(
...
-
-            // Add all headers to the HTTP request.
-            foreach (string key in additionalHeaders.AllKeys())
+            DateTime startDateTimeUtc = DateTime.UtcNow;
+
+            try
             {
-                message.Headers.Add(key, additionalHeaders[key]);
-            }
+                IEnumerator<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> timeoutEnumerator =
```
kirankumarkolli, line 143: IMO its not really needed.

NaluTripician:
Superseded by #5804.

We decided to scope this down to only adding the **non-retriable per-request timeout** for the Semantic Reranking inference endpoint. Since no retries are being added at this time, the `HttpTimeoutInferencePolicy`, `HttpTimeoutPolicyHelper`, and `CosmosHttpClientCore` refactor from this PR are being deferred — this PR can be reopened/revived if/when we decide to add retries for the inference endpoint.

The follow-up PR (#5804) also addresses @kirankumarkolli's review comment about `httpClient.Timeout` changing from 120s to 35s: in #5804 `httpClient.Timeout` is left unchanged at 120s, and the per-request budget is enforced purely via a linked `CancellationTokenSource` + `CancelAfter` around `SendAsync`.

Closing in favor of #5804.

PR #5566, opened 2026-01-15
[Internal] gitignore: Adds vscode files to gitignore

NaluTripician:
# Pull Request Template

## Description

Adds vscode files to gitignore

## Type of change

## Closing issues


.gitignore
```diff
@@ -328,3 +328,7 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
+.vscode/extensions.json
+.vscode/launch.json
+.vscode/settings.json
+.vscode/tasks.json
```
Pilchie: I think just putting `.vscode/` is typical. Are there things in that folder that we *would* want to add to git?

NaluTripician: This is the question that I also had. When I was doing some experiments in VSCode these where the only files that got added when setting up the SDK. If there are not any files we want to include, I am happy to change it to `.vscode/`

kirankumarkolli: Hmm unsure mt-self. 

@copilot can you please help us guide?

NaluTripician: Based on copilot's response just having `.vscode/` should be sufficient.

PR #5568, opened 2026-01-16
[Query] Fixes EPK performance by providing partial partition key in the request

adityasa:
# [Query] Fixes EPK performance by providing partial partition key in the request

## Description

The change provides partition key in the request header even when EPK routing is used. EPK routing is generally used when partial partition key is used for querying hierarchically partitioned container. As a result of this change, we could have query request having all 3 headers:
- Partition Key Range Id
- StartEPK, EndEPK
- Partition Key (logical partition key, partial or full)

The change only applies for query operations.

This allows backend to push the partition key (even when it's partial) to the index. This provides significant gains for containers where EPK indexing is not enabled (which applies to most containers).

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
+++ b/Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
@@ -397,7 +397,7 @@ namespace Microsoft.Azure.Cosmos
                 }
             }
 
-            if (pkExists && partitionKeyRangeIdExists)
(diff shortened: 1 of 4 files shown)


adityasa:
Run query tests with this change against a live account to validate stability.

NaluTripician:
Hi @adityasa — this PR has been idle for ~92 days. What's the current ETA / status? Flagging as part of an open-PR cleanup pass; if it's no longer being pursued we can close it. No auto-close — just looking for a status update.

PR #5572, opened 2026-01-23
Add support for constructor invocation binding

### LINKED ISSUE #3728: Support LINQ querying for records
Currently when invoking `Select()` for records, an exception is thrown

"Constructor invocation is not supported., Windows/10.0.19045 cosmos-netstandard-sdk/3.29.4"

Support constructors to select the members specified in constructors and let the deserializer handle the rest.

There are no alternatives except to not use records or not use LINQ.

**Additional context**
Example:

```C#
// record
record UserRecord(string Id)
{
    public string? Name { get; set; }
}

// query:
query.Select(x => new UserRecord(x.id, x.Name));
```

could result in

```sql
SELECT c.id as Id, c.Name
FROM c
```

which yields

```json
[
   { "Id": "12345", "Name": "Joe" }
]
```

JoasE:
# Pull Request Template

## Description

Adds support for projecting to types with constructor arguments.
This is specifically useful when working with records that are defined with primary constructors
```csharp
record UserRecord(string Id, string Name);

query.Select(x => new UserRecord(x.id, x.name));
```

## Type of change

## Closing issues
closes #3728


Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -916,12 +916,22 @@ private static SqlScalarExpression VisitNew(NewExpression inputExpression, Trans
...
             {
-                if (inputExpression.Members == null)
+                IReadOnlyList<MemberInfo> members = inputExpression.Members;
+                if (members == null)
                 {
-                    throw new DocumentQueryException(ClientResources.ConstructorInvocationNotSupported);
+                    MemberInfo[] typeMembers = inputExpression.Type.GetMembers(BindingFlags.Instance | BindingFlags.Public);
+                    members = inputExpression.Constructor.GetParameters().Select(param =>
+                    {
+                        MemberInfo member = typeMembers.FirstOrDefault(x => x.Name.Equals(param.Name, StringComparison.OrdinalIgnoreCase));
```
bartelink: at least have a comment to justify the IgnoreCase

JoasE: I've added a comment, but also did some other improvements:
Reduce complexity
Use GetFields and GetProperties instead of GetMembers

bartelink: Saw that, looks better (NOTE I'm just a lurker; someone with authority on to merge/approve will need to do the real reviewing work here)

JoasE: Thought you were a mantainer :p

Pilchie:
@adityasa  - this is in Linq. Can someone from the query team review?

Microsoft.Azure.Cosmos/src/Linq/ExpressionToSQL.cs
```diff
@@ -916,12 +916,37 @@ private static SqlScalarExpression VisitNew(NewExpression inputExpression, Trans
 
             if (inputExpression.Arguments.Count > 0)
             {
-                if (inputExpression.Members == null)
+                BindingFlags flags = BindingFlags.Instance | BindingFlags.Public;
+                PropertyInfo[] properties = inputExpression.Type.GetProperties(flags);
+                FieldInfo[] fields = inputExpression.Type.GetFields(flags);
```
adityasa, line 921: We generally try to avoid using raw reflection in this manner.

Can you add some more details to the PR (especially around what is the missing functionality that does NOT have workarounds) so it's a bit more clear?

AFAIK, current scenario can be resolved with a parameter-less constructor and property setters on the object. Correct?

adityasa, line 921: Assuming there is a good reason to continue down this path (TBD), at the very least we should move this and other (raw) reflection usage into a separate class so that it can be better tracked and minimized.

JoasE: @adityasa Thanks for the review! This is specifically for working with records defined using a primary constructor:
```csharp
record UserRecord(string Id, string Name);
```

There is definitely a possible workaround, which is using required init properties:
```csharp
record UserRecord
{
    public required string Id { get; init; }
    public required string Name { get; init; }
}
```

But there appears to be community interest in allowing the primary constructor syntax in #3728. Most likely because it is less verbose in both the query and the object definition.

The reflection is currently needed to match the parameters in the expression to members on the type. We need this to align with modern serialization standards where json properties will be matched to members, and members are matched to constructor parameters. Meaning the parsed json structure must match the property definitions in order to be able to use them as constructor parameters, and not the other way around. 

To avoid reflection: We could do it without matching the parameter to a member, by assuming a naming convention where the parameter is always the same case or only first case is lower instead of upper and using the parameter name string in the query. This does divert from most used serialization library's behavior, but only in (edge) cases where naming conventions don't match. On top of that, this method wouldn't allow use of attributes like `JsonPropertyName`, which are applied to the member instead of the argument. Example:
```csharp
public record Person([property: JsonPropertyName("first-name")] string FirstName,
    [property: JsonPropertyName("last-name")] string LastName);
```
This would require a different approach and some more refactoring to make possible in the query parser. 

Another option would be caching the reflection in a concurrent dictionary if possible.

What do you think about these options?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/BaselineTest/TestBaseline/LinqSQLTranslationBaselineTest.ValidateSQLTranslation.xml
```diff
@@ -235,8 +235,10 @@ WHERE ((root["x"] >= 0) AND (root["x"] < 3))]]></SqlQuery>
       <Expression><![CDATA[query.Select(x => new TimeSpan(Convert(x.x, Int64)))]]></Expression>
     </Input>
     <Output>
-      <SqlQuery><![CDATA[]]></SqlQuery>
-      <ErrorMessage><![CDATA[Constructor invocation is not supported.]]></ErrorMessage>
+      <SqlQuery><![CDATA[
+SELECT VALUE {"Ticks": root["x"]} 
+FROM root]]></SqlQuery>
+      <ErrorMessage><![CDATA[Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'System.TimeSpan' because the type requires a JSON primitive value (e.g. string, number, boolean, null) to deserialize correctly.To fix this error either change the JSON to a JSON primitive value (e.g. string, number, boolean, null) or change the deserialized type so that it is a normal .NET type (e.g. not a primitive type like integer, not a collection type like an array or List<T>) that can be deserialized from a JSON object. JsonObjectAttribute can also be added to the type to force it to deserialize from a JSON object.Path '[0].Ticks', line 1, position 10.]]></ErrorMessage>
```
adityasa, line 241: This seems like an issue. Will likely affect other types than just TimeSpan that don't support deserialization in this manner.

JoasE: Could you clarify the exact issue?
Both before and after my changes this scenario would result in an exception telling the user what they are trying to do isn't possible.
As far as I can see, the things that changed compared to previous behaviour are:
- Exception type and message
- Exception thrown at deserialization time instead of query compile time

Which makes sense to me as whether or not the constructor can be used for deserialization is determined by the serializer. 

If we want to have more specific behaviour, an option might be to add methods to the CosmosLinqSerializer to retreive info from there, but I personally feel it might be a bit overkill for this specific edge case? Tho I think it might be justified for matching the paramter names to properties in the comment above, which might help in  cases like this aswell. 

Have you had time to evaluate the direction options for the parameter matching?

adityasa: I was referring to the error message and timing and the fact that new error message is rather confusing and less actionable. TimeSpan also seems like one of the many times that could hit this issue (degraded error experience).

Regarding error message timing - when you say this is at deserialization time, does this make it data dependent?

JoasE: @adityasa 
> Regarding error message timing - when you say this is at deserialization time, does this make it data dependent?

well if the document has a property called "Ticks" that is a string representing a timespan this could possibly deserialize without an exception, but there would be a mismatch between the model used and the document being retreived.

adityasa: If the document successfully parses in some cases, that is not the intended design. Deserializing TimeSpan should consistently fail (at compile time, regardless of the document contents)

JoasE: @adityasa
> Deserializing TimeSpan should consistently fail (at compile time, regardless of the document contents)

 I think you mean 'Deserializing TimeSpan **via the Ticks constructor**' right?

If so then, I don't think this is going to be feasible, unless we can somehow easily check whether a certain constructor is viable for deserialization and add a method for that to the CosmosLinqSerializer. But that seems cumbersome either way.

I would like to emphasize that this only happens when there is a mismatch between the model used in the query and the document being retreived and deserialized. All sorts of errors can happen in the sdk at that point, all outside of query compilation.

For example, if I have 2 documents in the same container:
```json
{
  "id": "1",
  "someProp": "string"
}
```
```json
{
  "id": "2",
  "someProp": { 
     "nestedProp: "someProp is not a string but an object"
   }
}
```

And I query them using this model:
```csharp
var query = await cosmosClient.GetContainer("test", "test").GetItemQueryIterator<MyModel>("SELECT * FROM c").ReadNextAsync();
class MyModel
{
    public string Id { get; set; }
    public string SomeProp { get; set; }
}
```

I also get an exception at deserialization time, not query compile time.
I think the scenario with TimeSpan and Ticks is comparable.

PR #5582, opened 2026-01-30
Query: Adds ability to choose global vs local/focused statistics for FullTextScore 

neildsh:
# Enabling users to choose global vs local/focused statistics for FullTextScore 

## Why? 

Cosmos DB’s implementation of FullTextScore computes BM25 statistics (term frequency, inverse document frequency, and document length) across all documents in the container, including all physical and logical partitions. 

While this provides a valid and comprehensive representation of statistics for the entire dataset, it introduces challenges for several common use cases. 

In multi-tenant scenarios, it is often necessary to isolate queries to data belonging to a specific tenant, typically defined by the partition key or a component of a hierarchical partition key. This enables scoring to reflect statistics that are accurate for that tenant’s dataset, rather than for the entire container. For customers such as Veeam and Sitecore, which operate large multi-tenant containers, this is not just an optimization but a requirement. Their tenants often operate in very different domains, which can significantly change the distribution and importance of keywords and phrases. Using global statistics in these cases leads to distorted relevance rankings. 

In other scenarios involving hundreds or thousands of physical partitions, computing statistics across the entire container can become both time-consuming and expensive. Customers may prefer to use statistics derived from only a subset of partitions to improve performance and reduce RU consumption. Indeed, there is precedence for this as  Azure AI Search defaults to this “local” method.  

## What? 

We propose extending the flexibility of BM25 scoring in Cosmos DB so that developers can choose between a *global FullTextScore* (existing behavior) or *Scoped FullTextScore* (statistics computed restricted to the partition key(s) used in the query). The key aspects: 

For _*global*_ BM25, FullTextScore retains its existing behavior and computes BM25 statistics, such as IDF and average document length, across all documents in the container regardless of any partition key filters in the query. In _*scoped*_ BM25, when a query includes a partition key filter or explicitly requests scoped scoring, the engine computes these statistics only over the subset of documents within the specified partition key values. Query results are still returned only from the filtered partitions, and the resulting scores and ranking reflect relevance within that partition-specific slice of data.

## How?

The user issues query like: 

```
SELECT TOP 10 * FROM c   
WHERE c.tenantId = @tenantId   
ORDER BY RANK FullTextScore(c.text, "keywords") 
```
 

And sets a new [QueryRequestOption](../Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs) called `FullTextScoreScope` which can be set to one of two values: `local` or `global`. The request option is inspected, and the query uses scoped/full stats accordingly. 

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
@@ -222,7 +222,7 @@ namespace Microsoft.Azure.Cosmos
         /// </para>
         /// <para>
         /// When set to <see cref="FullTextScoreScope.Local"/>, statistics are computed only over the subset of documents
-        /// within the partition key values specified in the query. This is useful for multi-tenant scenarios where scoring
+        /// within the partition key values specified in the query request. This is useful for multi-tenant scenarios where scoring
         /// should reflect statistics that are accurate for a specific tenant's dataset.
         /// </para>
         /// </remarks>
--- a/Microsoft.Azure.Cosmos/src/Resource/Settings/FullTextScoreScope.cs
+++ b/Microsoft.Azure.Cosmos/src/Resource/Settings/FullTextScoreScope.cs
@@ -14,13 +14,13 @@ namespace Microsoft.Azure.Cosmos
         /// across all documents in the container, including all physical and logical partitions.
         /// This is the default behavior.
         /// </summary>
-        Global = 0,
+        Global,
 
         /// <summary>
         /// Compute BM25 statistics only over the subset of documents within the partition key values
         /// specified in the query. This is useful for multi-tenant scenarios where scoring should
         /// reflect statistics that are accurate for a specific tenant's dataset.
         /// </summary>
-        Local = 1
+        Local
     }
 }


docs/query/local_statistics_for_hybrid_search.md
```diff
@@ -0,0 +1,117 @@
+# Enabling users to choose global vs local/focused statistics for FullTextScore
```
adityasa, line 1: Do you intent to include this in PR? If so, consider putting this in a subfolder: query\HubridSearch\

adityasa:
Is it possible to add some emulator based e2e tests?
One validation that is of interest is partition filtering based on :
1. query filter
2. QueryRequestOptions

In both cases, the query should honor FullTextScore Scope (local v/s global).

docs/query/local_statistics_for_hybrid_search.md
neildsh: done

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -209,6 +209,25 @@ public ConsistencyLevel? ConsistencyLevel
...
+        /// The scope for computing BM25 statistics. Defaults to <see cref="FullTextScoreScope.Global"/>.
+        /// </value>
+        /// <remarks>
+        /// <para>
+        /// When set to <see cref="FullTextScoreScope.Global"/>, BM25 statistics (term frequency, inverse document frequency,
+        /// and document length) are computed across all documents in the container, including all physical and logical partitions.
+        /// </para>
+        /// <para>
+        /// When set to <see cref="FullTextScoreScope.Local"/>, statistics are computed only over the subset of documents
+        /// within the partition key values specified in the query. This is useful for multi-tenant scenarios where scoring
```
adityasa: >specified in the query [](http://example.com/codeflow?start=44&length=22)

We also honor partition key values from query request options. Consider a slightly more generic phrasing ("specified in the query request").

Microsoft.Azure.Cosmos/src/Resource/Settings/FullTextScoreScope.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// Specifies the scope for computing BM25 statistics used by FullTextScore in hybrid search queries.
+    /// </summary>
+    public enum FullTextScoreScope
+    {
+        /// <summary>
+        /// Compute BM25 statistics (term frequency, inverse document frequency, and document length) 
+        /// across all documents in the container, including all physical and logical partitions.
+        /// This is the default behavior.
+        /// </summary>
+        Global = 0,
```
sboshra: What's the reason for having explicit enum values?

neildsh: fixed

PR #5600, opened 2026-02-02
HPK: Adds id to partition key when "/id" is the last path in partition key definition

ramarag:
# Pull Request Template

## Description

The current change ensures that SDK users don't need to repeat themselves when "/id" is the last part of HPK/

With a traditional single-level partition key—for example, "/pk"—a given partition key value is constrained to a single physical partition.
Because of this:

Throughput (RU/s) for that partition key cannot exceed the limit of a single partition.
Storage available to that partition key is also bounded by that single partition’s storage capacity.
As a result, a “hot” partition key can become a scalability bottleneck.

By Using Hierarchical Partition key (HPK) with "/id" as the last level path. 1 partition key essentially maps to a single document.
"/pk" from above becomes a prefix and uses the prefix-based functionality of query in HPK. The partition key "/pk" now becomes boundless and can span multiple partitions.

Cons:
The following features can work only on one document, largely reducing their efficacy
1. Atomic Batch, 
2. Stored Procedure
3.  DeleteAllDocuments by Partition Key

## Example
Without this change reads will be like:

```
pKeyBuilder = new PartitionKeyBuilder()
        .Add(doc.GetPropertyValue<string>("ZipCode"))
        .Add(doc.GetPropertyValue<string>("City"));
       
pKey = pKeyBuilder.Add(doc.GetPropertyValue<string>("id1"))
        .Build();
var doc1 = await this.container.ReadItemAsync<Document>("id1",  pKey)

pKey = pKeyBuilder.Add(doc.GetPropertyValue<string>("id1"))
        .Build();
var doc2 = await this.container.ReadItemAsync<Document>("id2",  pKey)
```

But with the change this can be transformed into:
```
pKey = new PartitionKeyBuilder()
        .Add(doc.GetPropertyValue<string>("ZipCode"))
        .Add(doc.GetPropertyValue<string>("City"))
        .Build();

var doc1 = await this.container.ReadItemAsync<Document>("id1",  pKey)
var doc2 = await this.container.ReadItemAsync<Document>("id2",  pKey)
```

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -1065,6 +1067,70 @@ private static bool TryParseTokenListForElement(CosmosObject pathTraversal, IRea
...
+                if (streamPayload.CanSeek)
+                {
+                    streamPayload.Position = 0;
+                }
+            }
+        }
+
+        public override async Task<string> GetItemIdFromStreamIfRequiredAsync(
+            string itemId,
+            Stream streamPayload,
```
FabianMeiswinkel: Stream becomes unusable because it is drained and we can't rely on Steream to be seekable. Instead eother have this be a transofrmation that will also return a usable/consumable stream or make sure any caller is required to take care of making the stream seekable first.

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -1065,6 +1067,70 @@ private static bool TryParseTokenListForElement(CosmosObject pathTraversal, IRea
             return true;
         }
 
+        private static string GetIdFromStreamPayload(Stream streamPayload)
+        {
+            if (streamPayload == null || !streamPayload.CanSeek)
```
FabianMeiswinkel: Nothing requires Stream to be seekable - we can't just fail if customer is using non-seekable stream?

Microsoft.Azure.Cosmos/src/Batch/BatchAsyncContainerExecutor.cs
```diff
@@ -81,6 +81,7 @@ public virtual async Task<TransactionalBatchOperationResult> AddAsync(
                 throw new ArgumentNullException(nameof(operation));
             }
 
+            await operation.UpdatePartitionKeyIfRequiredAsync(this.cosmosContainer, trace, cancellationToken);
```
FabianMeiswinkel: Not necessary - atomic batch is dead with id as part of PK

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.cs
```diff
@@ -707,5 +709,59 @@ public override FeedIterator<T> GetChangeFeedIteratorWithQuery<T>(
                 changeFeedIteratorCore,
                 responseCreator: this.ClientContext.ResponseFactory.CreateChangeFeedUserTypeResponse<T>);
         }
+
+        public override async Task<PartitionKey?> EnsureIdGetAppendedtoPartitionKeyIfneededAsync(
```
FabianMeiswinkel: Unit tests missing

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerProperties.cs
```diff
@@ -712,6 +713,44 @@ internal ContainerProperties(string id, PartitionKeyDefinition partitionKeyDefin
...
+                    return this.isLastPartitionKeyPathId.Value;
+                }
+
+                if (this.PartitionKey?.Kind != PartitionKind.MultiHash)
+                {
+                    this.isLastPartitionKeyPathId = false;
+                    return this.isLastPartitionKeyPathId.Value;
+                }
+
+                IReadOnlyList<string> partitionKeyPaths = this.PartitionKey?.Paths;
```
FabianMeiswinkel, line 729: Wasn't the plan to make pk-def mutable? This code would only be safe as long as pk-def is immutable?

FabianMeiswinkel (changes requested):
Few open questions

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
ramarag: returning a seekable stream

Microsoft.Azure.Cosmos/src/Batch/BatchAsyncContainerExecutor.cs
ramarag: would like to keep it so that, functionality remains consistent.
It will also throw an error from SDK 
https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5600/changes#diff-9b561374ce2a28251e2afbb39b291196f2f40bb52d4dc121d205b924a81a6becR728

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.cs
ramarag: added

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerProperties.cs
ramarag: this will be a subsequent change, this one is only about adding syntactic sugar

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,176 @@
...
+            if (streamPayload.CanSeek)
+            {
+                originalPosition = streamPayload.Position;
+            }
+
+            if (streamPayload is not MemoryStream memoryStream)
+            {
+                memoryStream = new MemoryStream();
+
+                streamPayload.CopyTo(memoryStream);
```
FabianMeiswinkel: async overload should be used.

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            long originalPosition = 0;
+            if (streamPayload.CanSeek)
+            {
+                originalPosition = streamPayload.Position;
+            }
+
+            if (streamPayload is not MemoryStream memoryStream)
+            {
+                memoryStream = new MemoryStream();
```
FabianMeiswinkel: Buffer pooling? See https://github.com/Azure/azure-cosmos-dotnet-v3/blob/d1393f188fe3e93fd17fe9c196edd7bff7e6b1d3/Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/JsonHelper.cs#L22 for an example - ComputeGateway also heavily relies on Buffer pooling.

FabianMeiswinkel: See https://msdata.visualstudio.com/CosmosDB/_git/pgcosmos?path=/gateway/Cosmos.CoreAndTrace/RecyclableMemoryManager/ComputeRecyclableMemoryStream.cs&_a=contents&version=GBmain

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,176 @@
...
+
+                streamPayload.CopyTo(memoryStream);
+                memoryStream.Position = 0;
+            }
+
+            string idValue = null;
+            try
+            {
+                IJsonNavigator jsonNavigator = JsonNavigator.Create(memoryStream.ToArray());
```
FabianMeiswinkel: ToArray creates another copy - on a MemoryStream you own you should be able to use getBuffer

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,176 @@
...
+                CosmosObject cosmosObject = CosmosObject.Create(jsonNavigator, jsonNavigatorNode);
+
+                if (cosmosObject.TryGetValue("id", out CosmosElement idElement) && idElement is CosmosString cosmosString)
+                {
+                    idValue = cosmosString.Value;
+                }
+            }
+            catch
+            {
+            }
```
FabianMeiswinkel, line 201: Some logging should happen here at least

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: done

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: done

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: done

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: done

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+            byte[] rentedBuffer = null;
+            if (streamPayload is MemoryStream memoryStream)
+            {
+                if (memoryStream.TryGetBuffer(out ArraySegment<byte> existingBuffer))
+                {
+                    payload = existingBuffer.AsMemory();
+                }
+            }
+
+            if (payload.IsEmpty)
```
FabianMeiswinkel, line 166: If MemoryStream has empty content and TryGetBuffer succeeds you would still enter this if-condition and try to read again. Probably functionally fine - but unnecessary?

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            if (containerProperties.IsLastPartitionKeyPathId)
+            {
+                return await ContainerPropertiesExtensions.GetIdFromStreamPayloadAsync(streamPayload);
+            }
+
+            return (itemId, streamPayload);
+        }
+
+        private static async Task<(string, Stream)> GetIdFromStreamPayloadAsync(Stream streamPayload)
```
FabianMeiswinkel: NIT - I think the API docs should clearly spell out that the streamPayload gets consumed and it is unsafe to use this stream passed into GetIdFromStreamPayloadAsync afterwards - This is especially important because it most cases it would still work (when streamPayload is seeekable you reset the Position on streamPayload) - but code that relies on it would break when streams are non-seekable.

FabianMeiswinkel (dismissed):
LGTM except for two NITs - thanks for driving these changes!

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            ContainerProperties containerProperties;
+            try
+            {
+                containerProperties = await container.GetCachedContainerPropertiesAsync(
+                    forceRefresh: false,
+                    trace: NoOpTrace.Singleton,
+                    cancellationToken: cancellationToken);
+            }
+            catch (Exception)
```
kushagraThapar: shall we swallow all kinds of exceptions here? That could contains OOM or something fatal. How about catching only CosmosException instead and letting any other exception bubble up to the customer?

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -131,8 +131,11 @@ internal async Task<List<ResponseMessage>[]> ReadManyTaskHelperAsync(IDictionary
             IDictionary<PartitionKeyRange, List<(string, PartitionKey)>> partitionKeyRangeItemMap = new
                 Dictionary<PartitionKeyRange, List<(string, PartitionKey)>>();
 
-            foreach ((string id, PartitionKey pk) item in items)
+            foreach ((string id, PartitionKey pk) in items)
             {
+                PartitionKey? partitionKey = await this.container.EnsureIdGetAppendedToPartitionKeyIfNeededAsync(pk, id, cancellationToken);
+                (string id, PartitionKey pk) item = (id, partitionKey.Value);
```
kushagraThapar: what if the partitionKey is null? should we check it before getting the .Value from it?

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerProperties.cs
```diff
@@ -712,6 +713,44 @@ internal ContainerProperties(string id, PartitionKeyDefinition partitionKeyDefin
...
+        /// <summary>
+        /// Gets a value indicating whether the last partition key path is "id".
+        /// This property is used to determine if the item's "id" field is part of the hierarchical partition key.
+        /// </summary>
+        /// <value>
+        /// <c>true</c> if the last partition key path is "id"; otherwise, <c>false</c>.
+        /// Returns <c>false</c> if the partition key is not defined or has no paths.
+        /// </value>
+        [JsonIgnore]
+        public bool IsLastPartitionKeyPathId
```
kushagraThapar: I am curious, why do we need to make this a public API?

kushagraThapar (commented):
Thanks @ramarag  for the changes, my only concern is the catch (Exceptions) code path - which will swallow everything and not just cosmos exceptions.

xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
```diff
@@ -329,6 +329,10 @@ internal override Task<ResponseMessage> ProcessResourceOperationStreamAsync(
             CancellationToken cancellationToken)
         {
             this.ThrowIfDisposed();
+
+            (itemId, streamPayload) = await cosmosContainerCore.GetItemIdFromStreamIfRequiredAsync(itemId, streamPayload, cancellationToken);
+            partitionKey = await cosmosContainerCore.EnsureIdGetAppendedToPartitionKeyIfNeededAsync(partitionKey, itemId, cancellationToken);
```
xinlian12: 🔴 **Correctness — `DeleteAllItemsByPartitionKeyStreamAsync` will throw `ArgumentException` for HPK+/id containers**

This interception point is too broad — it applies to **all** operations going through this overload, not just item CRUD. `DeleteAllItemsByPartitionKeyStreamAsync` (`ContainerCore.Items.cs:1036`) calls this method with `itemId: null` and `streamPayload: null`.

The call chain:
1. `GetItemIdFromStreamIfRequiredAsync(null, null, ...)` → returns `(null, null)` (stream is null guard)
2. `EnsureIdGetAppendedToPartitionKeyIfNeededAsync(partitionKey, null, ...)` → for a 2-path HPK `["/pk", "/id"]`, when user provides a 1-component PK: `existingComponents.Count (1) == partitionKeyPaths.Count - 1 (1)` → match → `string.IsNullOrEmpty(null)` → **throws `ArgumentException`**

This completely breaks `DeleteAllItemsByPartitionKeyStreamAsync` for any container with HPK ending in `/id`. The PR description lists this as a "con" with reduced efficacy (implying it should still function), but the code throws instead.

**Suggested fix:** Guard these calls with `resourceType == ResourceType.Document`:

```csharp
if (cosmosContainerCore != null && resourceType == ResourceType.Document)
{
    (itemId, streamPayload) = await cosmosContainerCore.GetItemIdFromStreamIfRequiredAsync(itemId, streamPayload, cancellationToken);
    partitionKey = await cosmosContainerCore.EnsureIdGetAppendedToPartitionKeyIfNeededAsync(partitionKey, itemId, cancellationToken);
}

```

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+                return partitionKey;
+            }
+
+            if (string.IsNullOrEmpty(itemId))
+            {
+                throw new ArgumentException("itemId needs to be specified if LastPartitionKeyPath is id");
+            }
+
+            IReadOnlyList<Documents.Routing.IPartitionKeyComponent> existingComponents = Array.Empty<Documents.Routing.IPartitionKeyComponent>();
+            if (partitionKey.HasValue)
```
xinlian12: 🔴 **Correctness — `PartitionKey.None` causes `NullReferenceException`**

`PartitionKey.None` is a struct with `IsNone = true`. When passed as `PartitionKey?`, `.HasValue` is `true` (it's a non-null struct value), so this code enters the `HasValue` branch. However, `PartitionKey.None.InternalKey.Components` is `null` — confirmed by `PartitionKey.cs:209` which maps null `Components` to `PartitionKey.None`.

Accessing `.Count` on `null` here throws `NullReferenceException`.

**Concrete scenario:** Any item operation on an HPK+/id container that passes `PartitionKey.None` as the partition key (e.g., operations on non-partitioned items transitioning to HPK).

**Suggested fix:** Add a guard before accessing Components:

```csharp
if (partitionKey.HasValue)
{
    if (partitionKey.Value.IsNone)
    {
        return partitionKey;
    }
    existingComponents = partitionKey.Value.InternalKey.Components;
    // ... rest of logic
}

```

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+                    cancellationToken: cancellationToken);
+            }
+            catch (Exception)
+            {
+                // Container may not exist yet (e.g. operations on non-existent containers).
+                // Swallow the exception and let the actual operation surface the proper error.
+                return (itemId, streamPayload);
+            }
+
+            if (containerProperties.IsLastPartitionKeyPathId)
```
xinlian12: 🔴 **Correctness — Missing null check on `containerProperties`**

Unlike `EnsureIdGetAppendedToPartitionKeyIfNeededAsync` (line 48) which properly checks `containerProperties == null`, this method accesses `containerProperties.IsLastPartitionKeyPathId` without a null guard. If `GetCachedContainerPropertiesAsync` returns `null`, this will throw `NullReferenceException`.

**Suggested fix:** Add the same pattern used in the sibling method:

```csharp
if (containerProperties == null || !containerProperties.IsLastPartitionKeyPathId)
{
    return (itemId, streamPayload);
}

```

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            long originalPosition = 0;
+            if (streamPayload.CanSeek)
+            {
+                originalPosition = streamPayload.Position;
+            }
+
+            ReadOnlyMemory<byte> payload = default;
+            byte[] rentedBuffer = null;
+            if (streamPayload is MemoryStream memoryStream)
```
xinlian12, line 158: 🟡 **Reliability — Missing `typeof(MemoryStream)` type-safety check**

The existing SDK pattern in `BatchExecUtils.cs:34-38` explicitly adds a `memStream.GetType() == typeof(MemoryStream)` check with a detailed comment:

> *"Some derived implementations of MemoryStream (such as versions of RecyclableMemoryStream prior to 1.2.2 that we may be using) return an incorrect response from TryGetBuffer."*

This code skips that guard. If a derived `MemoryStream` (e.g., `CloneableStream`, `RecyclableMemoryStream`) is passed, `TryGetBuffer` may return incorrect data, leading to corrupted JSON parsing and silent wrong-partition routing.

**Suggested fix:**

```csharp
if (streamPayload is MemoryStream memoryStream
    && memoryStream.GetType() == typeof(MemoryStream))
{
    if (memoryStream.TryGetBuffer(out ArraySegment<byte> existingBuffer))
    {
        payload = existingBuffer.AsMemory();
    }
}

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Batch/BatchAsyncContainerExecutorTests.cs
```diff
@@ -241,7 +241,7 @@ public async Task DoesNotRecalculatePartitionKeyRangeOnNoSplits()
             TransactionalBatchOperationResult result = await executor.AddAsync(itemBatchOperation, NoOpTrace.Singleton);
 
             Mock.Get(mockContainer.Object)
-                .Verify(x => x.GetCachedContainerPropertiesAsync(It.IsAny<bool>(), It.IsAny<ITrace>(), It.IsAny<CancellationToken>()), Times.Once);
+                .Verify(x => x.GetRoutingMapAsync(It.IsAny<CancellationToken>()), Times.Once);
```
xinlian12, line 244: 🟡 **Testing — Verification weakened; HPK feature not actually exercised**

This verification was changed from asserting `GetCachedContainerPropertiesAsync` was called to asserting `GetRoutingMapAsync`. The problem is: the mock setup (line 54) returns `ContainerProperties` with `PartitionKind` defaulting to `Hash` (not `MultiHash`), so `IsLastPartitionKeyPathId` returns **false**, and `EnsureIdGetAppendedToPartitionKeyIfNeededAsync` becomes a **silent no-op**. The HPK auto-append feature is not exercised at all by these tests.

`GetRoutingMapAsync` was already called in the original code (line 245 of `BatchAsyncContainerExecutor.cs`) — this verification proves nothing new about the PR's changes.

Per review rules: "DO verify new tests are additive, not substitutive."

**Suggested fix:** Either keep the old verification alongside the new one, or update the mock to use `Kind = PartitionKind.MultiHash` and verify the partition key was actually modified by the append logic.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Batch/BatchAsyncContainerExecutorTests.cs
```diff
@@ -25,9 +25,9 @@ public async Task InitializeAsync()
         {
             this.cosmosClient = TestCommon.CreateCosmosClient(useGateway: true);
             DatabaseResponse db = await this.cosmosClient.CreateDatabaseAsync(Guid.NewGuid().ToString());
-            PartitionKeyDefinition partitionKeyDefinition = new PartitionKeyDefinition();
-            partitionKeyDefinition.Paths.Add("/Status");
-            this.cosmosContainer = (ContainerInlineCore)await db.Database.CreateContainerAsync(new ContainerProperties() { Id = Guid.NewGuid().ToString(), PartitionKey = partitionKeyDefinition }, 10000);
+            ContainerProperties containerProperties = new ContainerProperties("mycoll", new List<string> { "/Status", "/id" });
```
xinlian12, line 28: 🟡 **Testing — Emulator tests substitutive, not additive**

Per review rules: *"DO verify new tests are additive, not substitutive. New feature tests should NOT replace existing test coverage for related features."*

This file previously tested **single-PK batch** operations (with `"/Status"`). Now it only tests HPK with `"/Status", "/id"`. The original single-PK batch/bulk executor scenario is no longer covered by this test class.

Similarly, `CosmosItemBulkTests.cs` changed from single-PK `"/pk"` to HPK `"/pk", "/id"`.

The new HPK coverage is valuable, but the original single-PK scenarios should be preserved (either keep the existing tests alongside new ones, or add dedicated single-PK test methods).

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
```diff
@@ -927,6 +927,7 @@ private async Task<ResponseMessage> ProcessItemStreamAsync(
 
             ContainerInternal.ValidatePartitionKey(partitionKey, requestOptions);
             string resourceUri = this.GetResourceUri(requestOptions, operationType, itemId);
+            (itemId, streamPayload) = await ContainerPropertiesExtensions.GetItemIdFromStreamIfRequiredAsync(this, itemId, streamPayload, cancellationToken);
```
xinlian12: 🟡 **Design — Redundant call; `GetItemIdFromStreamIfRequiredAsync` called twice per operation**

This call extracts the id from the stream, but the same method is called **again** at `ClientContextCore.cs:333` inside `ProcessResourceOperationStreamAsync`. The second call short-circuits (itemId is already non-null), but it still allocates an async state machine and calls `GetCachedContainerPropertiesAsync` (cache lookup) on every item operation.

Since `ClientContextCore.ProcessResourceOperationStreamAsync` is the central hub that all item operations pass through, it's the right place for the HPK logic. This call here is redundant.

**Suggested fix:** Remove this call. Let `ClientContextCore.cs:333` handle it as the single point of responsibility for HPK augmentation.

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+            {
+                IJsonNavigator jsonNavigator = JsonNavigator.Create(payload);
+                IJsonNavigatorNode jsonNavigatorNode = jsonNavigator.GetRootNode();
+                CosmosObject cosmosObject = CosmosObject.Create(jsonNavigator, jsonNavigatorNode);
+
+                if (cosmosObject.TryGetValue("id", out CosmosElement idElement) && idElement is CosmosString cosmosString)
+                {
+                    idValue = cosmosString.Value;
+                }
+            }
```
xinlian12, line 190: 🟡 **Error Handling — Exception swallowing produces misleading downstream error**

When JSON parsing fails here (e.g., malformed JSON, corrupt stream), `idValue` remains `null`. The caller `GetItemIdFromStreamIfRequiredAsync` returns `(null, stream)`. Then `EnsureIdGetAppendedToPartitionKeyIfNeededAsync` receives `itemId: null` and throws:

```
ArgumentException: "itemId needs to be specified if LastPartitionKeyPath is id"

```

The user sees a partition-key error instead of the real problem (invalid JSON payload). The trace log is written here, but most users won't have trace logging enabled.

Consider at minimum re-throwing `OperationCanceledException` (to respect user cancellation tokens) and `OutOfMemoryException`. This aligns with @kushagraThapar's concern about the `GetCachedContainerPropertiesAsync` catch blocks swallowing fatal exceptions.

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos.Core.Trace;
+    using Microsoft.Azure.Cosmos.CosmosElements;
+    using Microsoft.Azure.Cosmos.Json;
+    using Microsoft.Azure.Cosmos.Tracing;
+
+    internal static class ContainerPropertiesExtensions
```
xinlian12, line 19: 💬 **Observation — No design document and novel cross-SDK behavior**

Per review rules: *"DO look for design documents in the `docs/` directory and the `openspec/` directory for major features."*

No design document was found for this feature. This is a significant behavioral change that **silently mutates user-provided partition keys** — a pattern with no precedent in Java, Python, or Rust Cosmos SDKs. The behavioral contract for edge cases (DeleteAllByPK, PartitionKey.None, partial PKs, already-complete PKs) would benefit from a written spec, particularly since the implicit PK modification could surprise users debugging partition routing issues.

This is not blocking, but documenting the expected behavior for edge cases would increase confidence and help future maintainers.

xinlian12:
✅ **Review complete** (48:31)

Posted **9** inline comment(s).

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerProperties.cs
```diff
@@ -712,6 +713,44 @@ internal ContainerProperties(string id, PartitionKeyDefinition partitionKeyDefin
...
+                IReadOnlyList<string> partitionKeyPaths = this.PartitionKey?.Paths;
+                if (partitionKeyPaths == null || partitionKeyPaths.Count <= 0)
+                {
+                    this.isLastPartitionKeyPathId = false;
+                    return this.isLastPartitionKeyPathId.Value;
+                }
+
+                string lastPartitionKeyPath = partitionKeyPaths[partitionKeyPaths.Count - 1];
+
+                this.isLastPartitionKeyPathId = string.Equals(lastPartitionKeyPath, "/id", StringComparison.OrdinalIgnoreCase);
```
kirankumarkolli: Single /id as PK also will match this case right? 
If we change above `PK.Count <= 1` then it will be more robust.

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+                // Swallow the exception and let the actual operation surface the proper error.
+                return partitionKey;
+            }
+
+            if (containerProperties == null || !containerProperties.IsLastPartitionKeyPathId)
+            {
+                return partitionKey;
+            }
+
+            if (string.IsNullOrEmpty(itemId))
```
kirankumarkolli, line 100: Promote up as early as possible for check (best practice)

Microsoft.Azure.Cosmos/src/Batch/BatchAsyncContainerExecutor.cs
```diff
@@ -238,6 +238,10 @@ private async Task<string> ResolvePartitionKeyRangeIdAsync(
                 trace: trace,
                 cancellationToken: cancellationToken);
             PartitionKeyDefinition partitionKeyDefinition = cachedContainerPropertiesAsync?.PartitionKey;
+
+            operation.PartitionKey = await this.cosmosContainer.EnsureIdGetAppendedToPartitionKeyIfNeededAsync(
```
xinlian12: 🟡 **Observability — Cache priming with `NoOpTrace` makes "Read Collection" diagnostics invisible for bulk operations**

Both `EnsureIdGetAppendedToPartitionKeyIfNeededAsync` and `GetItemIdFromStreamIfRequiredAsync` call `GetCachedContainerPropertiesAsync` with `NoOpTrace.Singleton` (lines 36-39 and 113-116 of `ContainerPropertiesExtensions.cs`). In the bulk path, these calls execute in `ClientContextCore.ProcessResourceOperationStreamAsync` (line 333-334) **before** the batch executor's own call at line 236-239 of this file, which passes a real `ITrace`.

**What happens:**
1. `ClientContextCore` calls `EnsureIdGetAppendedToPartitionKeyIfNeededAsync` → internal `GetCachedContainerPropertiesAsync(NoOpTrace.Singleton)` → cache miss → network call → **trace entries discarded** (NoOp)
2. `BatchAsyncContainerExecutor.ResolvePartitionKeyRangeIdAsync` calls `GetCachedContainerPropertiesAsync(trace)` → **cache hit** → no network → no trace entries generated

**Evidence:** The baseline XML diff confirms this — "Waiting for Initialization" and "Read Collection" entries are removed from `EndToEndTraceWriterBaselineTests.BulkOperationsAsync.xml` (two separate test baselines).

**Impact:** Customers using `CosmosDiagnostics` to debug collection metadata fetch latency in bulk operations lose visibility into the "Read Collection" transport trace for the first operation. This affects **all containers** (not just HPK-with-/id), since the extension methods always call `GetCachedContainerPropertiesAsync` before short-circuiting.

**Suggested fix:** Accept an `ITrace` parameter in both extension methods and pass it through, or restructure the calls so the real-trace call in `BatchAsyncContainerExecutor` executes first (before the extension methods).

xinlian12:
✅ **Review complete** (04:51)

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerProperties.cs
ramarag: actaully it will work, if Cx want to keep their code elegant and not specify id multiple times

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: we only want to throw if it has the last level as id, and that check happens in L48

NaluTripician:
Hi @ramarag — I cherry-picked four follow-up commits from my mirror PR #5785 onto this branch so the fixes live alongside your original work. Pushed as `a1f4129c1..f7bfc2799` on top of your existing tip.

## Commits added

1. **`a1f4129` — Restore Gateway mode substatus guard in `CosmosNotFoundTests`**
   The new `EnsureIdGetAppendedToPartitionKeyIfNeededAsync` pre-flight container read changes the substatus surfaced for not-found errors in Gateway mode. Restored the original Gateway-vs-Direct guard in the test so it asserts the correct substatus per connection mode.

2. **`74ceb0b` — Tests: Adds request URI diagnostics to `InitTaskThreadSafe`**
   Added `RequestUri` capture/logging to the diagnostics asserted by the test. Pure test-side diagnostic improvement to make the next failure of this test much easier to triage.

3. **`eadc96c` — PartitionKey: Fixes latent issues in `ContainerPropertiesExtensions` and hardens `InitTaskThreadSafe`**
   - In `ContainerPropertiesExtensions.EnsureIdGetAppendedToPartitionKeyIfNeededAsync`, the in-memory cache lookup now uses `ClientCollectionCache.TryGetCachedContainerProperties` (sync, in-memory only) instead of `GetCachedContainerPropertiesAsync`. The async path was triggering an extra HTTP/init outside the handler pipeline trace and breaking `GatewayClientSideRequestStatsTests` request-count expectations.
   - Hardened `InitTaskThreadSafe` so it no longer relies on resetting `VmMetadataApiHandler`'s static `isInitialized` latch between iterations (was racing with parallel tests). Cumulative `metadataCallCount=1` across iterations is the correct invariant.

4. **`f7bfc27` — PartitionKey: Refactors exception filter for HPK id-append pre-flight lookup**
   Cleaned up the exception filter on the pre-flight container properties lookup in `EnsureIdGetAppendedToPartitionKeyIfNeededAsync` — narrower catch + no behavior change, just clearer intent.

## Why
PR #5785 was a mirror of this PR opened from the Azure org so CI gates (which can't run on fork PRs) could exercise the change end-to-end. While running CI I hit two real failures — `GatewayClientSideRequestStatsTests` (caused by the async cache call adding a hidden HTTP request) and the `InitTaskThreadSafe` race — plus a couple of test-quality issues. Those are the four commits above.

All previous review feedback on this PR still applies; these are purely additive on top of your work. Happy to break any of them out into separate commits or amend if you'd prefer a different shape.

NaluTripician:
Branch might still not have correct CI because of testing branch rules. See my mirror PR for testing validation.

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: you are correct, i will not fix it because functionality is fine

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: Disagree -> if id can not be parsed, the correct error is throw "itemId needs to be specified if LastPartitionKeyPath is id"
so that users can explicitly be specified via partition key path.

modified the exception message to reflect the intent

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
ramarag: EnsureIdGetAppendedToPartitionKeyIfNeededAsync should at least return pk, but i will change it to default to pk, @kushagraThapar  please let me know if you have alternate suggestions.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Batch/BatchAsyncContainerExecutorTests.cs
ramarag: this is by design, supporting 2 paths should be sufficient

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Batch/BatchAsyncContainerExecutorTests.cs
ramarag: The intention of the test is to check for Routing Map refresh is not triggered, modifying the logic to reflect that.

Microsoft.Azure.Cosmos/src/Resource/Container/ContainerCore.Items.cs
ramarag: this is by design, so that we read the id from stream before it gets encrypted

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: Disagree-> It is a synctatic sugar for what users can already achieve by specifying partition key explicitly

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
ramarag: Disagree -> The actual RecyclableMemoryStream bug (pre-1.2.2) returned either:

 - true with an empty ArraySegment (the common case — it couldn't expose its chunked internal buffers as a single contiguous segment)
 - false (correct behavior, fixed in
  1.2.2+)

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
ramarag: Agreed -> This breaks the feature by design. We will follow up with another work item to get it working

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
kushagraThapar: That should be enough, thanks!

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -0,0 +1,240 @@
...
+            {
+                throw new ArgumentException("itemId needs to be specified if LastPartitionKeyPath is id or add it to the partition key paths");
+            }
+
+            PartitionKeyBuilder builder = new PartitionKeyBuilder();
+
+            if (!partitionKey.HasValue)
+            {
+                for (int i = 0; i < containerProperties.PartitionKey.Paths.Count - 1;; i++)
```
kushagraThapar:
```suggestion
                for (int i = 0; i < containerProperties.PartitionKey.Paths.Count - 1; i++)
```

kushagraThapar (dismissed):
LGTM, thanks @ramarag

NaluTripician:
Pushed commit `37c0da4` to address the `ClientTests.InitTaskThreadSafe` regression that this PR was introducing.

**Root cause:** `EnsureIdGetsAppendedToPartitionKeyIfNeededAsync` is called pre-pipeline. Its call to `GetCachedContainerPropertiesAsync` routes through `EnsureValidClientAsync` and can re-enter the init path before `isSuccessfullyInitialized` has flipped, producing extra HTTP traffic outside the `AsyncLazy` coalescing - so `InitTaskThreadSafe` saw more than one URL hit to localhost during init.

**Fix (gate on init state):**
1. Try the in-memory `ClientCollectionCache` first (zero HTTP via new `AsyncCache.TryGetCachedValue` / `CollectionCache.TryGetCachedContainerProperties` / `ContainerCore.TryGetCachedContainerPropertiesFromMemory`).
2. If miss and `DocumentClient.IsClientInitialized == false`: skip the auto-append for this op; the request pipeline will warm the cache.
3. If miss and the client *is* initialized: fall through to `GetCachedContainerPropertiesAsync` - safe because `EnsureValidClientAsync` short-circuits on the flag.

Also drops the per-iteration `metadataCallCount` reset that the PR had added to `InitTaskThreadSafe`: `VmMetadataApiHandler.isInitialized` is a process-wide latch and resetting it between iterations races with parallel tests.

**Local emulator results on this branch (head `37c0da4`):**
- `ClientTests.InitTaskThreadSafe` - 5/5 PASS (repeated)
- `CosmosMultiHashTest` - 12/12 PASS
- `CosmosItemBulkTests` - 17/17 PASS
- `BatchAsyncContainerExecutorTests.ValidateRequestOptions` - PASS

Total: 30/30 on the regression suite.

NaluTripician:
Hi @ramarag — I verified this PR locally and pushed a small fix that resolves the failing unit tests and the missing changelog entry. Branch: [`nalutripician/verify-pr-5600`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/nalutripician/verify-pr-5600), commit [`461ae9ff`](https://github.com/Azure/azure-cosmos-dotnet-v3/commit/461ae9ffcc15f31c5fe5e29222a1d3af24f55a2d).

### What was broken
9 of the 10 new `CosmosItemUnitTests.EnsureIdGet[s]Appended_*` tests were failing locally:

| Symptom | Cause |
|---|---|
| `Assert.IsNotNull failed` / `["tenant1"]` instead of `["tenant1","item123"]` / no `ArgumentException` thrown | The mock helper `CreateMockContainerWithProperties` only stubs `GetCachedContainerPropertiesAsync`, but commit 37c0da4 added a fast-path that early-returns when `!container.IsClientInitialized` *before* reaching that mock. The new virtual default is `false`, so every test silently skips the auto-append logic. |

### Fix (cherry-pickable)
Two surgical edits, +7 lines total:

1. **`Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosItemUnitTests.cs`** — stub `IsClientInitialized` to `true` in the helper so the slow path is actually exercised:
   ```csharp
   mock.Setup(x => x.IsClientInitialized).Returns(true);
   ```
2. **`changelog.md`** — add the missing entry under `### Unreleased` → `#### Features Added` (the changelog-soft-check CI status was failing).

### Verification
- `dotnet build Microsoft.Azure.Cosmos.sln -c Release` → 0 errors.
- All 15 `CosmosItemUnitTests.EnsureIdGet*` tests pass.
- Full `Microsoft.Azure.Cosmos.Tests` suite: only failure left is `InvalidKey_ExceptionFullStacktrace`, which is unrelated and pre-existing (needs an emulator-reachable endpoint to throw the network exception it asserts on).
- Emulator integration tests against a local emulator:
  - `CosmosMultiHashTest` (PR-touched, +195 lines) — **12/12 pass**, including the new HPK `/id` integration tests
  - `Batch.BatchAsyncContainerExecutorTests` — pass
  - `CosmosItemBulkTests` — pass
  - `ClientTests` — 20/20 pass

(I also ran broader emulator suites and saw failures in circuit-breaker, PPAF, multi-region, and `ContainterReCreateStatelessTest`, but baselining on `main` reproduces the same failures — they're environmental, not caused by this PR.)

Feel free to cherry-pick `461ae9ff` onto `users/ramarag/hpkwithid`, or I can open a PR back to your branch if that's easier.

FabianMeiswinkel (dismissed):
LGTM - except for the ignored tests - and we discussed it offline and came-up with a plan to re-enable. So, not blocking merge.

Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerPropertiesExtensions.cs
```diff
@@ -14,11 +14,28 @@ namespace Microsoft.Azure.Cosmos
...
+    using Microsoft.Azure.Documents;
 
     internal static class ContainerPropertiesExtensions
     {
         private const int DefaultStreamBufferSize = 4096;
-       
+
+        internal static bool ShouldValidatePartitionKeyHasId(ResourceType resourceType, OperationType operationType)
+        {
+            if (resourceType == ResourceType.PartitionKey &&
```
FabianMeiswinkel: I think a safer approach would just be to return false if ResourceType != Document - I mean whether PK/Delete or StoredProc/Execute etc. none of the qualify as pint operations - and the "add id magically" approach will only ever work for point operations (strictly ResourceType==Document)

ramarag: agreed, made the changes

FabianMeiswinkel (approved):
LGTM - Thanks!

kushagraThapar (approved):
LGTM, thanks @ramarag

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosMultiHashTest.cs
```diff
@@ -197,6 +199,144 @@ public async Task MultiHashReadItemTest()
...
+            new List<(string, Cosmos.PartitionKey)> { ("document3", default) });
+
+            Assert.AreEqual(1, feedResponse.Count());
+
+            await idPkContainer.DeleteItemAsync<Document>("document3", default);
+
+            CosmosException clientException = await Assert.ThrowsExceptionAsync<CosmosException>(() =>
+                idPkContainer.ReadItemAsync<Document>("document3", default)
+            );
+        }
```
ramarag, line 307: example usage, when the first level of key is optional, which translates to PartitionKey.None

Issue #5268, opened 2025-07-03
`ChangeFeedProcessor` skips the first change unless start time is specified


BalassaMarton:
**Describe the bug**

Using the latest `3.52.0` version of `Microsoft.Azure.Cosmos`, unless we specify the start time, the first change is consistently lost.

**To Reproduce**

Run below application:

`TestProject.csproj`
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.52.0" />
    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
  </ItemGroup>

</Project>
```

`Program.cs`
```cs
using System.Collections.Concurrent;
using System.Diagnostics;
using Microsoft.Azure.Cosmos;
using Newtonsoft.Json;

const int documentCount = 100;
const int timeoutSeconds = 30;

var cosmosConnectionUri = Environment.GetEnvironmentVariable("COSMOS_CONNECTION_URI");
var cosmosConnectionKey = Environment.GetEnvironmentVariable("COSMOS_CONNECTION_KEY");

var iteration = 0;

using var cts = new CancellationTokenSource();

Console.CancelKeyPress += (_, eventArgs) =>
{
    eventArgs.Cancel = true;
    Console.WriteLine("Shutting down...");
    cts.Cancel();
};

while (!cts.IsCancellationRequested)
{
    await RunTestAsync(cts.Token);
}

async Task RunTestAsync(CancellationToken cancellationToken)
{
    var uniqueId = Guid.NewGuid().ToString("N");

    Console.WriteLine($"Running iteration #{++iteration} with ID {uniqueId}");

    var databaseId = $"DB-{uniqueId}";
    var containerName = $"Test-{uniqueId}";
    var leaseContainerName = $"Leases-{uniqueId}";
    var processorName = $"Processor-{uniqueId}";
    var instanceName = $"Instance-{uniqueId}";

    using var client = new CosmosClient(cosmosConnectionUri, cosmosConnectionKey);

    var database = (await client.CreateDatabaseAsync(databaseId)).Database;

    try
    {
[...]

PR #5617, opened 2026-02-18
ChangeFeedProcessor: Fixes first-change skip during initial startup by anchoring StartTime

NaluTripician:
## Summary
Fixes issue #5268 where `ChangeFeedProcessor` can skip the first change when started without explicit start options.

## Bug Description
When `ChangeFeedProcessor` is started with default options (no `WithStartTime`, no continuation, no start-from-beginning), it uses the default start-from-now path.

This path anchors at the first service read (`IfNoneMatch: *`) rather than at `StartAsync()` call time. Because startup includes asynchronous lease acquisition and load balancing, there is a window where documents can be written after `StartAsync()` returns but before the first read reaches the backend. Those documents can be skipped.

## Root Cause
- `PartitionLoadBalancerCore.Start()` schedules background work and returns immediately.
- The first change feed read happens later after lease operations.
- Default `Now` behavior is based on first read timing, not `StartAsync()` timing.
- Documents created in that gap may be missed.

## Fix
In `ChangeFeedProcessorCore.StartAsync()`:
- On first initialization only, if no explicit start option is configured:
  - `StartFromBeginning == false`
  - `StartTime == null`
  - `StartContinuation` is null/empty
- set `changeFeedProcessorOptions.StartTime = DateTime.UtcNow` before async initialization.

This anchors start behavior at processor startup time and closes the timing gap.

## Why This Is Safe
- Does not override explicit user start settings.
- Applies only on first startup when no lease continuation exists.
- Existing continuation/checkpoint behavior remains unchanged.

### Unit tests
`Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedProcessorCoreTests.cs`
- `StartAsync_SetsStartTime_WhenNoStartOptionsProvided`
- `StartAsync_DoesNotOverrideExplicitStartTime`
- `StartAsync_DoesNotSetStartTime_WhenStartFromBeginning`

### Emulator regression test
`Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs`
- `TestWithRunningProcessor_ImmediateWriteAfterStart`
- Starts processor and immediately writes documents without setup delay.

### A/B test on the new emulator regression
- With fix enabled: **10/10 passed**
- With fix temporarily removed: **7/10 failed**, showing the original issue reproduces with high probability
- Fix restored and revalidated: stable pass behavior

### Additional checks
- Targeted unit tests pass.
- No diagnostics errors in changed files.

## Files Changed
- `Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs`
- `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedProcessorCoreTests.cs`
- `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs`


Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -59,6 +59,13 @@ public override async Task StartAsync()
         {
             if (!this.initialized)
             {
+                if (!this.changeFeedProcessorOptions.StartFromBeginning
```
Praveen-Msft, line 62: Is there any other way to ensure there is no lease document exists rather than inferring (assuming when these 3 conditions are met)?

NaluTripician: Good question — the three conditions here aren't actually inferring whether lease documents exist. They're checking whether the **user explicitly configured a start behavior** (StartFromBeginning, StartTime, or StartContinuation). The fix targets the specific case where the user provided no start options at all, which defaults to "start from now" — the bug is that "now" was anchored at first-read time rather than `StartAsync()` call time.

Directly checking for existing leases would be tricky here because `documentServiceLeaseStoreManager` isn't initialized until `InitializeAsync()` on the next line, so we'd need to restructure the initialization flow and add an extra async query.

More importantly, it wouldn't change behavior — `ChangeFeedProcessorOptions.StartTime` is already documented as being ignored when a lease has a continuation token (see the xml doc comments on `StartTime` and `StartFromBeginning`). The precedence in `ChangeFeedPartitionKeyResultSetIteratorCore.Create()` is: existing continuation token > StartTime > StartFromBeginning > default "Now". So setting StartTime here is safe regardless of whether leases exist.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs
```diff
@@ -78,6 +78,44 @@ public async Task TestWithRunningProcessor()
...
+            await processor.StartAsync();
+            foreach (int id in Enumerable.Range(0, 10))
+            {
+                await this.Container.CreateItemAsync<dynamic>(new { id = id.ToString(), pk = partitionKey });
+            }
+
+            bool isStartOk = allDocsProcessed.WaitOne(10 * BaseChangeFeedClientHelper.ChangeFeedSetupTime);
+            await processor.StopAsync();
+            Assert.IsTrue(isStartOk, "Timed out waiting for docs to process");
+            Assert.AreEqual("0.1.2.3.4.5.6.7.8.9.", accumulator);
```
kirankumarkolli, line 116: Its possible that more than 10 can be processed?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -59,6 +59,15 @@ public override async Task StartAsync()
...
+                    && this.changeFeedProcessorOptions.StartTime == null
```
kirankumarkolli, line 63: What is JAVA SDK behavior?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedProcessorCoreTests.cs
```diff
@@ -131,6 +131,142 @@ public async Task StartAsync()
...
+                    leaseStoreManager.Object,
+                    null,
+                    "instanceName",
+                    new ChangeFeedLeaseOptions(),
+                    options,
+                    ChangeFeedProcessorCoreTests.GetMockedContainer("monitored"));
+
+                await processor.StartAsync();
+
+                Assert.AreEqual(explicitStartTime, options.StartTime);
```
kirankumarkolli, line 213: Isn't there an exsiting equivalent test?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AvailabilityStrategyUnitTests.cs
```diff
@@ -298,7 +298,7 @@ public async Task AppCancellationDuringHedging_DoesNotSpawnNewHedgeRequests()
                 {
                     // First request: cancel the app token after a brief delay
                     // This simulates an e2e timeout scenario
-                    _ = Task.Delay(15).ContinueWith(_ => appCts.Cancel());
+                    _ = Task.Delay(1).ContinueWith(_ => appCts.Cancel());
```
kirankumarkolli: Is it related to this change?

kirankumarkolli:
Please update Title with initialization state also

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs
NaluTripician: Good catch — the `== 10` check is fragile. If the callback fires with a batch that crosses 10 (e.g., count goes from 8 to 12), the `ManualResetEvent` would never be set and the test would time out. Changed to `>= 10` to be defensive. Pushed in a4dec3fb0.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
NaluTripician: The Java SDK has the same default "start from now" semantics when no explicit start options are configured (``startFromBeginning`` defaults to ``false``). However, Java also anchors at first-read time rather than at ``start()`` call time, so the same timing gap exists there. This fix makes the .NET SDK more robust for immediate-write-after-start scenarios by anchoring at ``StartAsync()`` call time instead.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedProcessorCoreTests.cs
NaluTripician: The existing ``StartAsync`` test (line 92-131) verifies lease initialization calls (``IsInitializedAsync`` and ``GetOwnedLeasesAsync`` are each called once). It does not assert anything about ``StartTime``. This new test specifically validates the new behavior — that ``StartTime`` is set to ``DateTime.UtcNow`` when no explicit start options are provided.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AvailabilityStrategyUnitTests.cs
NaluTripician: No, this is unrelated — it was a test reliability fix for the ``AppCancellationDuringHedging_DoesNotSpawnNewHedgeRequests`` test. The 15ms delay was sometimes too long relative to the 10ms hedging threshold, causing flaky behavior. Reducing to 1ms makes the cancellation trigger more deterministically. Happy to revert this from the PR and submit separately if you prefer.

NaluTripician:
Updated the PR title to include initialization state context as requested.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
kirankumarkolli: In-case of application stricter semantics there is already a way for application to explicitly specify

kirankumarkolli (approved):

Issue #5095, opened 2025-03-27
Contacted Regions returns empty string when contacting hub region after 410:21005 for multimaster accounts


NaluTripician:
>We are continuously addressing and improving the SDK, if possible, make sure the problem persist in the [latest SDK version](https://www.nuget.org/packages/Microsoft.Azure.Cosmos).

**Describe the bug**
For multimaster read requests in scenarios where exclude regions is enabled excluding the hub region, after the SDK exhausts retries on remaining region (returning 410:21005), the SDK will correctly route request to hub region. Here once response is returned and GetContactedRegions is called on the diagnostics, the region name will appear as null, with the global endpoint returned correctly as the URI. This issue does not appear in single master calls. 

**To Reproduce**
Steps to reproduce the behavior. If you can include code snippets or links to repositories containing a repro of the issue that can helps us in detecting the scenario it would speed up the resolution.

**Expected behavior**

**Actual behavior**
Provide a description of the actual behavior observed. 

**Environment summary**
SDK Version:
OS Version (e.g. Windows, Linux, MacOSX)

**Additional context**
Add any other context about the problem here (for example, complete stack traces or logs).

PR #5618, opened 2026-02-18
Diagnostics: Fixes null contacted region name for multimaster hub fallback (410/21005)

NaluTripician:
## Summary
Fixes issue #5095 where `GetContactedRegions()` can return `(null, <global-endpoint-uri>)` for multimaster reads when requests fall back to the account default endpoint after `410/21005` retries with `ExcludeRegions`.

### What this change does
- Updates `LocationCache.GetLocation(Uri endpoint)` to resolve the default/global endpoint to the first available write location name for multimaster as well as single-master.
- Updates `LocationCache.TryGetLocationForGatewayDiagnostics(Uri endpoint, out string regionName)` so that for default-endpoint hostnames in multimaster accounts, diagnostics resolve to hub/write region instead of returning `null`.
- Adds regression coverage in `LocationCacheTests` for multimaster default-endpoint diagnostics mapping.

## Root cause
For multimaster accounts:
- `GetApplicableEndpoints(...)` can fall back to `defaultEndpoint` when preferred/applicable regional endpoints are exhausted or excluded.
- Existing diagnostics resolution in `TryGetLocationForGatewayDiagnostics` treated endpoints with host matching `defaultEndpoint` as non-regional and returned `regionName = null`.
- `GetLocation(defaultEndpoint)` also returned `null` in multimaster scenarios due to a guard condition.

This caused diagnostics to record contacted regions with an empty/null region name despite correct endpoint routing.

### Repro test added
- `ValidateTryGetLocationForGatewayDiagnosticsOnDefaultEndpointForMultiMaster`
  - File: `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/LocationCacheTests.cs`
  - Validates that in multimaster mode:
    - `GetLocation(DefaultEndpoint)` returns the hub/write region name.
    - `TryGetLocationForGatewayDiagnostics(DefaultEndpoint, out regionName)` returns `true` and non-null hub region.
    - Same behavior for default endpoint with path (`new Uri(DefaultEndpoint, "random/path")`).

### Red/green proof
- **Without fix** (temporarily reverted locally):
  - Command: `dotnet test .\\Microsoft.Azure.Cosmos\\tests\\Microsoft.Azure.Cosmos.Tests\\Microsoft.Azure.Cosmos.Tests.csproj --filter "FullyQualifiedName~ValidateTryGetLocationForGatewayDiagnosticsOnDefaultEndpointForMultiMaster" --nologo`
  - Result: **Failed** with `Expected:<location1>. Actual:<(null)>`.
- **With fix restored**:
  - Same command
  - Result: **Passed**.

## Files changed
- `Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs`
- `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/LocationCacheTests.cs`

## Notes
- Single-master behavior for default endpoint diagnostics remains unchanged (`false` + null region name) in existing test coverage.
- Change is focused to diagnostics-region resolution behavior and corresponding regression test coverage.


Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -203,12 +200,18 @@ public bool TryGetLocationForGatewayDiagnostics(Uri endpoint, out string regionN
                     UriFormat.SafeUnescaped, 
                     StringComparison.OrdinalIgnoreCase) == 0)
             {
+                if (this.CanUseMultipleWriteLocations())
+                {
+                    regionName = this.GetLocation(this.defaultEndpoint);
+                    return regionName != null;
+                }
```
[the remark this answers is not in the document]
NaluTripician: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5618#discussion_r2849889632)

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -203,12 +200,18 @@ public bool TryGetLocationForGatewayDiagnostics(Uri endpoint, out string regionN
                     UriFormat.SafeUnescaped, 
                     StringComparison.OrdinalIgnoreCase) == 0)
             {
+                if (this.CanUseMultipleWriteLocations())
```
NaluTripician: **Medium concern**: `CanUseMultipleWriteLocations()` returns `this.useMultipleWriteLocations && this.enableMultipleWriteLocations` — it requires **both** client opt-in and account setting. If the account is multi-master but the client has `UseMultipleWriteLocations = false` (a valid configuration), this branch won't fire and diagnostics will still return `null` for the region name.
Consider using `this.enableMultipleWriteLocations` (server-side setting only) here instead. PR #5640 already proposes this fix — recommend folding it into this PR before merge rather than shipping a known diagnostic gap.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -168,15 +168,11 @@ public ReadOnlyCollection<Uri> ThinClientWriteEndpoints
...
-        /// <remarks>
-        /// Today we return null for defaultEndPoint if multiple write locations can be used.
-        /// This needs to be modifed to figure out proper location in such case.
-        /// </remarks>
         public string GetLocation(Uri endpoint)
         {
             string location = this.locationInfo.AvailableWriteEndpointByLocation.FirstOrDefault(uri => uri.Value == endpoint).Key ?? this.locationInfo.AvailableReadEndpointByLocation.FirstOrDefault(uri => uri.Value == endpoint).Key;
 
-            if (location == null && endpoint == this.defaultEndpoint && !this.CanUseMultipleWriteLocations())
+            if (location == null && endpoint == this.defaultEndpoint)
```
NaluTripician: **Low concern**: With the guard removal, the fallback path in `GetLocation` now reaches `AvailableWriteEndpointByLocation.First().Key` for multimaster scenarios too. Dictionary enumeration order is not contractually guaranteed. Consider using `AvailableWriteLocations[0]` from the ordered `ReadOnlyCollection` to pick the hub/first write location deterministically.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -203,12 +200,18 @@ public bool TryGetLocationForGatewayDiagnostics(Uri endpoint, out string regionN
...
+                    return regionName != null;
+                }
+
                 regionName = null;
                 return false;
             }
 
             regionName = this.GetLocation(endpoint);
-            return true;
+            return regionName != null;
```
NaluTripician, line 220: :+1: Good defensive fix — previously `TryGetLocationForGatewayDiagnostics` could return `true` with a null `regionName` if `GetLocation` returned null for a non-default endpoint.

NaluTripician (commented):
## Review Summary

The core fix is correct — removing the guard that prevented multimaster accounts from resolving the default endpoint to a write region name, and adding a null-safety check on the return path. A couple of concerns below.

**Whitespace noise**: ~200 lines of trailing-whitespace-only changes obscure the ~15 lines of real logic. Consider a separate formatting commit or `.editorconfig` enforcement to keep semantic diffs clean.

**Test coverage gap**: The new test covers `useMultipleWriteLocations: true` but not the asymmetric case (account is multi-master, client didn't opt in). PR #5640 addresses this — recommend folding it in before merge.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -168,19 +168,15 @@ public ReadOnlyCollection<Uri> ThinClientWriteEndpoints
...
-        /// This needs to be modifed to figure out proper location in such case.
-        /// </remarks>
         public string GetLocation(Uri endpoint)
         {
             string location = this.locationInfo.AvailableWriteEndpointByLocation.FirstOrDefault(uri => uri.Value == endpoint).Key ?? this.locationInfo.AvailableReadEndpointByLocation.FirstOrDefault(uri => uri.Value == endpoint).Key;
 
-            if (location == null && endpoint == this.defaultEndpoint && !this.CanUseMultipleWriteLocations())
+            if (location == null && endpoint == this.defaultEndpoint)
             {
-                if (this.locationInfo.AvailableWriteEndpointByLocation.Any())
```
kirankumarkolli, line 184: Is this change MUST? Isn't Any() better than count?

NaluTripician: Good question! Yes, this change is intentional — two reasons:

1. **Deterministic ordering**: The old code used `AvailableWriteEndpointByLocation.First().Key` which relies on `ReadOnlyDictionary` enumeration order — not contractually guaranteed in .NET. Since we removed the `!CanUseMultipleWriteLocations()` guard and this path now applies to multimaster, we need deterministic hub-region selection. `AvailableWriteLocations[0]` from the ordered `ReadOnlyCollection` guarantees that. Using `.Count > 0` on the same collection we index into is the natural guard.

2. **`.Count > 0` vs `.Any()` for `ReadOnlyCollection<string>`**: `.Count` is an O(1) property (delegates directly to the underlying list's count field) with zero overhead. `.Any()` goes through LINQ extension method dispatch and in older runtimes allocates an enumerator (in .NET 6+ it's optimized to just call `.Count > 0` internally anyway). The file already follows this convention consistently — `.Any()` for `ConcurrentDictionary` (where `.Count` is expensive due to lock acquisition), `.Count` for `ReadOnlyCollection` (where it's O(1)).

kirankumarkolli: Discussed offline

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -203,12 +200,21 @@ public bool TryGetLocationForGatewayDiagnostics(Uri endpoint, out string regionN
                     UriFormat.SafeUnescaped, 
                     StringComparison.OrdinalIgnoreCase) == 0)
             {
+                // Use account-level enableMultipleWriteLocations (not CanUseMultipleWriteLocations which also
+                // requires client opt-in) because diagnostics should resolve the hub region regardless of whether
+                // the client uses multi-write. The default endpoint routes to the hub/write region server-side.
+                if (this.enableMultipleWriteLocations)
```
kirankumarkolli: nit: Can be collapsed into above if

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -235,9 +241,9 @@ public void OnDatabaseAccountRead(AccountProperties databaseAccount)
         {
             this.UpdateLocationCache(
                 databaseAccount.WritableRegions,
-                databaseAccount.ReadableRegions,
```
kirankumarkolli, line 241: Unrelated changes )tabs to white space probably

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -168,19 +168,15 @@ public ReadOnlyCollection<Uri> ThinClientWriteEndpoints
...
         {
             string location = this.locationInfo.AvailableWriteEndpointByLocation.FirstOrDefault(uri => uri.Value == endpoint).Key ?? this.locationInfo.AvailableReadEndpointByLocation.FirstOrDefault(uri => uri.Value == endpoint).Key;
 
-            if (location == null && endpoint == this.defaultEndpoint && !this.CanUseMultipleWriteLocations())
+            if (location == null && endpoint == this.defaultEndpoint)
             {
-                if (this.locationInfo.AvailableWriteEndpointByLocation.Any())
+                if (this.locationInfo.AvailableWriteLocations.Any())
                 {
-                    return this.locationInfo.AvailableWriteEndpointByLocation.First().Key;
```
kirankumarkolli, line 186: Its possible in theory but not really needed for now. Unrelated

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -168,19 +168,15 @@ public ReadOnlyCollection<Uri> ThinClientWriteEndpoints
...
         /// </summary>
-        /// <remarks>
-        /// Today we return null for defaultEndPoint if multiple write locations can be used.
-        /// This needs to be modifed to figure out proper location in such case.
-        /// </remarks>
         public string GetLocation(Uri endpoint)
         {
             string location = this.locationInfo.AvailableWriteEndpointByLocation.FirstOrDefault(uri => uri.Value == endpoint).Key ?? this.locationInfo.AvailableReadEndpointByLocation.FirstOrDefault(uri => uri.Value == endpoint).Key;
 
-            if (location == null && endpoint == this.defaultEndpoint && !this.CanUseMultipleWriteLocations())
```
kirankumarkolli, line 182: MM first write region might not Hub most of the time.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -168,28 +168,24 @@ public ReadOnlyCollection<Uri> ThinClientWriteEndpoints
         /// For the defaultEndPoint, we will return the first available write location.
         /// Returns null, in other cases.
```
kushagraThapar: Isn't this documentation stale now?

NaluTripician: Good catch -- you're right, the summary was stale. Updated in 591c926 to cover all three cases: (1) regional write/read endpoint match, (2) default-endpoint fallback now applies to both single- and multi-master (since this PR removed the `!CanUseMultipleWriteLocations()` guard), and (3) clarified per your earlier note that the returned location is just the first entry in the write locations list -- not necessarily the hub region for multi-master accounts.

kushagraThapar (approved):
LGTM, thanks @NaluTripician

aavasthy (approved):

PR #5643, opened 2026-02-26
[Internal] Tests: Fixes test reliability across flaky and nightly tests

NaluTripician:
## Summary

This PR improves the reliability of tests across the nightly rolling pipeline and the flaky-tagged test suite. Changes are targeted at the root causes of intermittent failures identified from rolling test data analysis.

#### AvailabilityStrategyNoTriggerTest (10.8% failure rate — 23/212 runs)
**File:** `CosmosAvailabilityStrategyTests.cs`
- **Root cause:** Injected response delay (200ms) barely exceeded the hedging threshold (150ms), leaving only 50ms margin. On loaded CI agents, timing jitter caused inconsistent behavior.
- **Fix:** Increased delay to 500ms and threshold to 300ms, providing a 200ms safety margin.

#### DistributedTransaction E2E Tests (8.6% failure rate — 16/187 runs, 4 tests)
**File:** `DistributedTransactionE2ETests.cs`
**Tests:** ValidateConflictResponseReturnsErrorStatus, ValidateHappyPathRequestAndResponse, ValidateMixedOperationsRequestStructure, ValidateResponseDeserializesCorrectly
- **Root cause:** All 4 tests share a `TestInitialize` that creates a new container via `BaseCosmosClientHelper.TestInit()` (which calls `DeleteAllDatabasesAsync`). Concurrent test execution caused resource contention on the emulator.
- **Fix:** Added `[DoNotParallelize]` attribute and retry logic with exponential backoff around container creation.

#### AvailabilityStrategyAllFaultsTests (5.2% failure rate — 11/212 runs)
**File:** `CosmosAvailabilityStrategyTests.cs`
- **Root cause:** 90 DataRow combinations with a 100ms hedging threshold that was too tight for CI environments.
- **Fix:** Increased hedging threshold from 100ms to 200ms.

#### AppCancellationDuringHedging_DoesNotSpawnNewHedgeRequests (5.4% failure rate — 9/168 runs)
**File:** `AvailabilityStrategyUnitTests.cs`
- **Root cause:** Threshold of 10ms and app cancellation at 15ms gave only 5ms margin. `Task.Delay` precision at these scales is unreliable on loaded CI agents.
- **Fix:** Increased threshold from 10ms to 100ms, thresholdStep from 10ms to 100ms, and cancellation delay from 15ms to 150ms (10x margin increase).

#### CircuitBreaker Integration Tests (2.8% and 1% failure rates)
**File:** `CosmosItemIntegrationTests.cs`
- **Root cause:** 3000ms replication delay insufficient for multi-region data propagation on slow networks.
- **Fix:** Increased replication delay from 3s to 5s for both `ReadItemAsync_WithCircuitBreakerEnabled...ThirdRegion` and `ReadItemAsync_WithCircuitBreakerDisabled...` tests.

#### ValidatesCongestionControlAsync (0.3% failure rate — 1/359 runs)
**File:** `BatchAsyncStreamerTests.cs`
- **Root cause:** Fixed `Task.Delay(2000)` followed by immediate assertion on semaphore count. On slow CI, background semaphore release may not have completed.
- **Fix:** Replaced with polling loop (200ms intervals, 10s timeout) that waits for the semaphore count to reach the expected value.

#### Controller_ShouldReleasesLease_IfObserverExits (0.3% failure rate — 1/359 runs)
**File:** `PartitionControllerTests.cs`
- **Root cause:** 100ms delay insufficient for async observer completion and lease release.
- **Fix:** Increased delay from 100ms to 500ms.

#### CosmosHttpClientCoreTests (RetryTransientIssuesTestAsync, RetryTransientIssuesForQueryPlanTestAsync)
- **Root cause:** Delays only 0.1s beyond timeout policy thresholds, causing race conditions.
- **Fix:** Increased delay margins to ~1s beyond policy thresholds. Added `[Timeout(120000)]` (2 min) to both tests.

#### GlobalEndpointManagerTest (EndpointFailureMockTest)
- **Root cause:** 3s delay with 2s `UnavailableLocationsExpirationTimeInSeconds` left only 1s margin.
- **Fix:** Increased delay to 5s (3s margin). Added `[Timeout(30000)]`.

#### CosmosAuthorizationTests (TestTokenCredentialBackgroundRefreshAsync)
- **Root cause:** Unbounded polling loop (`while (NumTimesInvoked == 1) await Task.Delay(500)`) could spin indefinitely.
- **Fix:** Added bounded polling with `Stopwatch` and 20s timeout with descriptive assertion message.

#### ClientTelemetryTests (all 12 test methods)
- **Root cause:** No `[Timeout]` attributes on any tests in this flaky class, allowing individual tests to consume the entire 60-minute job budget.
- **Fix:** Added `[Timeout(300000)]` (5 min) to all 12 test methods.

#### EndToEndTraceWriterBaselineTests (QueryAsync, TypedPointOperationsAsync)
- **Root cause:** No `[Timeout]` attributes despite being marked `[TestCategory("Flaky")]`.
- **Fix:** Added `[Timeout(300000)]` (5 min) to both tests.

## Test Impact

| Category | Tests Improved | Expected Improvement |
|----------|---------------|---------------------|
| Nightly rolling (>5% failure) | 4 test groups | Eliminate timing races |
| Nightly rolling (1-3% failure) | 4 tests | Reduce replication/sync failures |
| Flaky-tagged tests | 8 tests across 6 files | Bounded timeouts prevent job overruns |

## Pipeline Duration Impact

The `[Timeout]` additions to flaky-tagged tests prevent any single test from consuming the entire 60-minute pipeline job timeout. Previously, a hanging flaky test with `retryCountOnTaskFailure: 4` could consume the full budget across retries.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/SqlObjects/SqlFunctionCallScalarExpression.cs
+++ b/Microsoft.Azure.Cosmos/src/SqlObjects/SqlFunctionCallScalarExpression.cs
@@ -103,6 +103,8 @@ namespace Microsoft.Azure.Cosmos.SqlObjects
             { Names.IsObject, Identifiers.IsObject },
             { Names.IsPrimitive, Identifiers.IsPrimitive },
             { Names.IsString, Identifiers.IsString },
+            { Names.LastSubstringAfter, Identifiers.LastSubstringAfter },
+            { Names.LastSubstringBefore, Identifiers.LastSubstringBefore },
             { Names.Left, Identifiers.Left },
             { Names.Length, Identifiers.Length },
             { Names.Like, Identifiers.Like },
@@ -144,6 +146,8 @@ namespace Microsoft.Azure.Cosmos.SqlObjects
             { Names.StringToNumber, Identifiers.StringToNumber },
             { Names.StringToObject, Identifiers.StringToObject },
             { Names.Substring, Identifiers.Substring },
+            { Names.SubstringAfter, Identifiers.SubstringAfter },
+            { Names.SubstringBefore, Identifiers.SubstringBefore },
             { Names.Sum, Identifiers.Sum },
             { Names.Tan, Identifiers.Tan },
             { Names.TicksToDateTime, Identifiers.TicksToDateTime },
@@ -351,6 +355,8 @@ namespace Microsoft.Azure.Cosmos.SqlObjects
             public const string IsPrimitive = "IS_PRIMITIVE";
             public const string IsString = "IS_STRING";
             public const string LastIndexOf = "LastIndexOf";
+            public const string LastSubstringAfter = "LastSubstringAfter";
+            public const string LastSubstringBefore = "LastSubstringBefore";
             public const string Left = "LEFT";
             public const string Length = "LENGTH";
             public const string Like = "LIKE";
@@ -399,6 +405,8 @@ namespace Microsoft.Azure.Cosmos.SqlObjects
             public const string StringToNumber = "StringToNumber";
             public const string StringToObject = "StringToObject";
             public const string Substring = "SUBSTRING";
+            public const string SubstringAfter = "SubstringAfter";
+            public const string SubstringBefore = "SubstringBefore";
             public const string Sum = "SUM";
             public const string Tan = "TAN";
             public const string TicksToDateTime = "TicksToDateTime";
@@ -529,6 +537,8 @@ namespace Microsoft.Azure.Cosmos.SqlObjects
             public static readonly SqlIdentifier IsPrimitive = SqlIdentifier.Create(Names.IsPrimitive);
             public static readonly SqlIdentifier IsString = SqlIdentifier.Create(Names.IsString);
             public static readonly SqlIdentifier LastIndexOf = SqlIdentifier.Create(Names.LastIndexOf);
+            public static readonly SqlIdentifier LastSubstringAfter = SqlIdentifier.Create(Names.LastSubstringAfter);
+            public static readonly SqlIdentifier LastSubstringBefore = SqlIdentifier.Create(Names.LastSubstringBefore);
             public static readonly SqlIdentifier Left = SqlIdentifier.Create(Names.Left);
             public static readonly SqlIdentifier Length = SqlIdentifier.Create(Names.Length);
             public static readonly SqlIdentifier Like = SqlIdentifier.Create(Names.Like);
@@ -577,6 +587,8 @@ namespace Microsoft.Azure.Cosmos.SqlObjects
             public static readonly SqlIdentifier StringToNumber = SqlIdentifier.Create(Names.StringToNumber);
             public static readonly SqlIdentifier StringToObject = SqlIdentifier.Create(Names.StringToObject);
             public static readonly SqlIdentifier Substring = SqlIdentifier.Create(Names.Substring);
+            public static readonly SqlIdentifier SubstringAfter = SqlIdentifier.Create(Names.SubstringAfter);
+            public static readonly SqlIdentifier SubstringBefore = SqlIdentifier.Create(Names.SubstringBefore);
             public static readonly SqlIdentifier Sum = SqlIdentifier.Create(Names.Sum);
             public static readonly SqlIdentifier Tan = SqlIdentifier.Create(Names.Tan);
             public static readonly SqlIdentifier TicksToDateTime = SqlIdentifier.Create(Names.TicksToDateTime);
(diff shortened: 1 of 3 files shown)


Pilchie:
It would be so much better if we could convert these delays into deterministic processing via `ManualResetEvent`'s.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTransaction/DistributedTransactionE2ETests.cs
```diff
@@ -30,11 +31,23 @@ public async Task TestInitialize()
         {
             await this.TestInit();
 
-            ContainerResponse response = await this.database.CreateContainerAsync(
-                new ContainerProperties(id: Guid.NewGuid().ToString(), partitionKeyPath: PartitionKeyPath),
-                cancellationToken: this.cancellationToken);
+            const int maxRetries = 3;
```
kirankumarkolli: Undo extra retries

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -68,15 +69,15 @@ public async Task RetryTransientIssuesTestAsync()
             {
                 {HttpTimeoutPolicyControlPlaneRead.Instance,  new List<TimeSpan>()
                 {
-                    TimeSpan.FromSeconds(5.1),
-                    TimeSpan.FromSeconds(10.1),
-                    TimeSpan.FromSeconds(20.1)
```
kirankumarkolli, line 73: Is this analysis right?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Batch/BatchAsyncStreamerTests.cs
```diff
@@ -150,9 +150,14 @@ public async Task ValidatesCongestionControlAsync()
             // 300 batch request should atleast sum up to 1000 ms barrier with wait time of 20ms in executor
             await Task.WhenAll(contexts);
 
-            await Task.Delay(2000);
+            // Poll for semaphore count to increase, with a reasonable timeout
+            System.Diagnostics.Stopwatch sw = System.Diagnostics.Stopwatch.StartNew();
```
kirankumarkolli, line 154: What's the evidence from logs its the delay?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTransaction/DistributedTransactionE2ETests.cs
NaluTripician: Done — removed the retry loop. \[DoNotParallelize]\ should be sufficient to prevent the concurrent container creation contention.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
NaluTripician: Yes — verified from the source. The \HttpTimeoutPolicyControlPlaneRead\ defines requestTimeouts of **5s, 10s, 20s** and \HttpTimeoutPolicyControlPlaneRetriableHotPath\ defines **0.5s, 5s, 65s**.

The original test delays (5.1s, 10.1s, 20.1s / 0.6s, 5.1s, 65.1s) had only **100ms margin** beyond these thresholds. On loaded CI agents, \Task.Delay\ precision at that scale is unreliable — the cancellation token from the timeout policy can fire before the delay completes, causing the \Assert.Fail\ to not trigger as expected.

The new values (6s, 11s, 21s / 1s, 6s, 66s) provide **~1s margin**, which is well within \Task.Delay\ reliability bounds.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Batch/BatchAsyncStreamerTests.cs
NaluTripician: The failure pattern is structural rather than log-based — the original code was:

\\\csharp
await Task.WhenAll(contexts);
await Task.Delay(2000);
Assert.IsTrue(newLimiter.CurrentCount >= 2, ...);
\\\

The semaphore count increases asynchronously in background threads as batch execution completes and releases slots. A fixed 2s delay is inherently a race condition: on slow CI agents (or when the thread pool is saturated), background threads may not have released the semaphore within that window.

The polling pattern replaces the guess with an explicit wait-for-condition:

\\\csharp
while (newLimiter.CurrentCount < 2 && sw.Elapsed < TimeSpan.FromSeconds(10))
    await Task.Delay(200);
\\\

This is a well-established reliability pattern for asserting async conditions — it succeeds as soon as the condition is met rather than guessing how long to sleep, and fails deterministically with a bounded timeout if the condition never occurs.

PR #5651, opened 2026-03-03
Serializer: Fixes unsafe stream cast in FromStream<T>

### LINKED ISSUE #5620: Cast may not always works in certain edge cases

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/SystemTextJson/CosmosSystemTextJsonSerializer.cs

**Describe the bug**

The following code in the function:
 public override T FromStream<T>(Stream stream)
...
                if (typeof(Stream).IsAssignableFrom(typeof(T)))
                {
                    return (T)(object)stream;
                }

when theoretically calling the function like this:

        using var stream = new MemoryStream([1, 2, 3]);
        serializer.FromStream<FileStream>(stream));

the cast:

`(T)(object)stream;
`

may results in a **InvalidCastException**

**To Reproduce**
Steps to reproduce the behavior. If you can include code snippets or links to repositories containing a repro of the issue that can helps us in detecting the scenario it would speed up the resolution.

as given in the description

**Expected behavior**

May the code can be enhanced by throwing an exception that describes itself better

        if (typeof(Stream).IsAssignableFrom(typeof(T)))
        {
            if (stream is T typedStream)
            {
                return typedStream;
            }

            throw new ArgumentException($"Cannot convert stream type '{stream.GetType().FullName}' to requested '{typeof(T).FullName}'.");
        }

**Actual behavior**
Provide a description of the actual behavior observed. 

a InvalidCastException is thrown

**Environment summary**
SDK Version:

not OS specific

**Additional context**

NaluTripician:
## Description

Fixes #5620

Replaces the unsafe `(T)(object)stream` cast pattern with safe `is` pattern matching in all `FromStream<T>` serializer implementations across the SDK.

### Problem

The `FromStream<T>` method in multiple serializer implementations uses the following pattern:

```csharp
if (typeof(Stream).IsAssignableFrom(typeof(T)))
{
    return (T)(object)stream;
}
```

`typeof(Stream).IsAssignableFrom(typeof(T))` returns `true` when `T` is `Stream` **or any subclass** (e.g., `MemoryStream`, `FileStream`). If `T` is a specific `Stream` subclass but the runtime `stream` parameter is a different `Stream` type, the cast `(T)(object)stream` throws a raw `InvalidCastException` with no context about what went wrong.

**Example that throws:**
```csharp
// T = MemoryStream, but stream is actually a FileStream at runtime
serializer.FromStream<MemoryStream>(someFileStream); // InvalidCastException!
```

### Fix

Replaced the unsafe cast with safe `is` pattern matching:

```csharp
if (typeof(Stream).IsAssignableFrom(typeof(T)))
{
    if (stream is T typedStream)
    {
        return typedStream;
    }

    throw new InvalidCastException(
        $"Stream of type '{stream.GetType().FullName}' is not compatible "
        + $"with the requested type '{typeof(T).FullName}'.");
}
```

This provides:
- ✅ Safe runtime type checking (no unexpected `InvalidCastException`)
- ✅ A descriptive error message identifying both the actual and expected types
- ✅ No behavioral change for the common case (`T = Stream`)

### Files Changed

**Core SDK Serializers (2):**
- `Microsoft.Azure.Cosmos/src/Serializer/CosmosJsonDotNetSerializer.cs`
- `Microsoft.Azure.Cosmos/src/Serializer/CosmosSystemTextJsonSerializer.cs`

**Sample Code (3):**
- `Microsoft.Azure.Cosmos.Samples/Usage/SystemTextJson/CosmosSystemTextJsonSerializer.cs`
- `Microsoft.Azure.Cosmos.Samples/Usage/ReEncryption/ReEncryptionSupport/ReEncryptionJsonSerializer.cs`
- `Microsoft.Azure.Cosmos.Samples/Usage/ItemManagement/Program.cs`

**Encryption Modules (2):**
- `Microsoft.Azure.Cosmos.Encryption/src/CosmosJsonDotNetSerializer.cs`
- `Microsoft.Azure.Cosmos.Encryption.Custom/src/Common/CosmosJsonDotNetSerializer.cs`

**Unit Tests (2):**
- `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosJsonSerializerUnitTests.cs` — 2 new tests
- `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Json/CosmosSystemTextJsonSerializerTest.cs` — 2 new tests

### Not Changed

- **`PatchOperationCore{T}.cs`** — Has a similar-looking pattern but casts FROM `T` TO `Stream` (upcast), which always succeeds. No fix needed.
- **Test utility serializers** — Internal test code, not customer-facing.

### Testing

- 4 new unit tests added (2 per serializer):
  - `ValidateFromStreamWithBaseStreamType` / `TestFromStreamWithBaseStreamType` — Confirms `FromStream<Stream>(memoryStream)` succeeds (regression test)
  - `ValidateFromStreamWithIncompatibleStreamTypeThrowsDescriptiveError` / `TestFromStreamWithIncompatibleStreamTypeThrowsDescriptiveError` — Confirms `FromStream<FileStream>(memoryStream)` throws `InvalidCastException` with a descriptive message containing both type names
- All existing tests continue to pass

### DIFF
--- a/Microsoft.Azure.Cosmos/contracts/API_3.55.1.txt
+++ b/Microsoft.Azure.Cosmos/contracts/API_3.55.1.txt
@@ -1,0 +1,1766 @@
+namespace Microsoft.Azure.Cosmos
+{
+    public class AccountConsistency
+    {
+        public AccountConsistency();
+        public ConsistencyLevel DefaultConsistencyLevel { get; }
+        public int MaxStalenessIntervalInSeconds { get; }
+        public int MaxStalenessPrefix { get; }
+    }
+    public class AccountProperties
+    {
+        public AccountConsistency Consistency { get; }
+        public string ETag { get; }
+        public string Id { get; }
+        public IEnumerable<AccountRegion> ReadableRegions { get; }
+        public IEnumerable<AccountRegion> WritableRegions { get; }
+    }
+    public class AccountRegion
+    {
+        public AccountRegion();
+        public string Endpoint { get; }
+        public string Name { get; }
+    }
+    public abstract class AvailabilityStrategy
+    {
+        public static AvailabilityStrategy CrossRegionHedgingStrategy(TimeSpan threshold, Nullable<TimeSpan> thresholdStep, bool enableMultiWriteRegionHedge=false);
+        public static AvailabilityStrategy DisabledStrategy();
+    }
+    public sealed class BoundingBoxProperties
+    {
+        public BoundingBoxProperties();
+        public double Xmax { get; set; }
+        public double Xmin { get; set; }
+        public double Ymax { get; set; }
+        public double Ymin { get; set; }
+    }
+    public abstract class ChangeFeedEstimator
+    {
+        protected ChangeFeedEstimator();
+        public abstract FeedIterator<ChangeFeedProcessorState> GetCurrentStateIterator(ChangeFeedEstimatorRequestOptions changeFeedEstimatorRequestOptions=null);
+    }
+    public sealed class ChangeFeedEstimatorRequestOptions
+    {
+        public ChangeFeedEstimatorRequestOptions();
+        public Nullable<int> MaxItemCount { get; set; }
+    }
+    public abstract class ChangeFeedMode
+    {
+        public static ChangeFeedMode Incremental { get; }
+        public static ChangeFeedMode LatestVersion { get; }
+    }
+    public abstract class ChangeFeedProcessor
+    {
+        protected ChangeFeedProcessor();
+        public abstract Task StartAsync();
+        public abstract Task StopAsync();
+    }
+    public class ChangeFeedProcessorBuilder
+    {
+        public ChangeFeedProcessor Build();
+        public ChangeFeedProcessorBuilder WithErrorNotification(Container.ChangeFeedMonitorErrorDelegate errorDelegate);
+        public virtual ChangeFeedProcessorBuilder WithInMemoryLeaseContainer();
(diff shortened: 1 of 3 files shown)


Microsoft.Azure.Cosmos.Encryption.Custom/src/Common/CosmosJsonDotNetSerializer.cs
```diff
@@ -44,7 +44,13 @@ public T FromStream<T>(Stream stream)
 
             if (typeof(Stream).IsAssignableFrom(typeof(T)))
             {
-                return (T)(object)stream;
+                if (stream is T typedStream)
+                {
+                    return typedStream;
+                }
+
+                throw new InvalidCastException(
```
kirankumarkolli: Isn't it supposed to fallback to the copy?

Microsoft.Azure.Cosmos.Encryption.Custom/src/Common/CosmosJsonDotNetSerializer.cs
```diff
@@ -44,7 +44,13 @@ public T FromStream<T>(Stream stream)
 
             if (typeof(Stream).IsAssignableFrom(typeof(T)))
             {
-                return (T)(object)stream;
+                if (stream is T typedStream)
```
kirankumarkolli: Is above if still needed?

kirankumarkolli (approved):

PR #5685, opened 2026-03-09
Read Consistency Strategy: Adds Read Consistency Strategy option for read requests

aavasthy:
# Pull Request Template

## Description
Currently, Cosmos DB accounts are configured with a single default consistency level, and the existing per-request override only allows weakening it — there's no way to strengthen reads or choose a fundamentally different read strategy without changing the account-level setting. This becomes a limitation for workloads where certain reads need the latest committed data via quorum reads.

ReadConsistencyStrategy introduces a new dimension of control by allowing applications to specify their desired read behavior (Eventual, Session, LatestCommitted, GlobalStrong) per-request or per-client, completely independent of the
account's default consistency. Unlike the existing ConsistencyLevel override, these strategies map directly to how the Direct layer reads from replicas — single replica reads for Eventual, session-token-aware reads for Session, quorum reads with GLSN barrier for LatestCommitted, and quorum reads with GCLSN barrier for GlobalStrong.

This PR adds the ReadConsistencyStrategy property is exposed on all read-path request options (ItemRequestOptions, QueryRequestOptions, ChangeFeedRequestOptions, ReadManyRequestOptions), CosmosClientOptions. The SDK sets the x-ms-cosmos-read-consistency-strategy header and propagates the strategy to DocumentServiceRequestContext, where the Direct package's ConsistencyReader and QuorumReader handle the actual read mode selection.

## Type of change

## Closing issues


kirankumarkolli:
>ReadConsistencyStrategy introduces a new dimension of control by allowing applications to specify their desired read behavior (Eventual, Session, LatestCommitted, GlobalStrong) per-request or per-client, completely independent of the
account's default consistency. Unlike the existing ConsistencyLevel override, these strategies map directly to how the Direct layer reads from replicas — single replica reads for Eventual, session-token-aware reads for Session, quorum reads with GLSN barrier for LatestCommitted, and quorum reads with GCLSN barrier for GlobalStrong.
Most developers might needs guidance is when to use what, so some guidance around it in PR description also in code-doc would be great. 

Few of the aspects are 
- Consistency upgrades (currently its only downgrade)
- ConsistentPrefix: Call out that Eventual == ConsistencyPrefix or may be we can add ConsistencyPrefix also explictly just to avoid any confusion.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -504,6 +508,58 @@ private async Task ValidateAndSetConsistencyLevelAsync(RequestMessage requestMes
...
+                    throw new ArgumentException(string.Format(
+                        CultureInfo.CurrentUICulture,
+                        "ReadConsistencyStrategy.GlobalStrong is only valid for accounts configured with Strong consistency. Current account consistency: {0}",
+                        this.AccountConsistencyLevel.Value));
+                }
+
+                // The Direct layer's ServerStoreModel rejects requests with both
+                // x-ms-consistency-level and x-ms-cosmos-read-consistency-strategy headers.
+                // Remove ConsistencyLevel if it was set by ValidateAndSetConsistencyLevelAsync.
+                requestMessage.Headers.Remove(HttpConstants.HttpHeaders.ConsistencyLevel);
```
kirankumarkolli: Its unclear to developers on priorities between them? Why not just fail if both are set?

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -504,6 +508,58 @@ private async Task ValidateAndSetConsistencyLevelAsync(RequestMessage requestMes
...
+            }
+
+            if (readConsistencyStrategy.HasValue)
+            {
+                if (!this.AccountConsistencyLevel.HasValue)
+                {
+                    this.AccountConsistencyLevel = await this.client.GetAccountConsistencyLevelAsync();
+                }
+
+                // GlobalStrong is only valid for Strong consistency accounts
```
kirankumarkolli: Nit: How about let it fail on service?

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
kirankumarkolli: And it also takes dependency on setting them in-order.

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.Azure.Cosmos
+{
+    /// <summary>
+    /// Represents the read consistency strategies supported by the Azure Cosmos DB service.
+    /// The requested strategy can be chosen independent of the account's default consistency level.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// When set, <see cref="ReadConsistencyStrategy"/> takes precedence over <see cref="ConsistencyLevel"/>
```
kirankumarkolli, line 13: Since service has no override semantics why is it needed in SDK?

Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
```diff
@@ -139,6 +139,12 @@ public virtual Stream Content
 
         internal PartitionKeyRangeIdentity PartitionKeyRangeId { get; set; }
 
+        /// <summary>
+        /// Set by RequestInvokerHandler when a ReadConsistencyStrategy is configured.
+        /// Uses Documents namespace type for Direct layer compatibility.
+        /// </summary>
+        internal Documents.ReadConsistencyStrategy? ReadConsistencyStrategy { get; set; }
```
kirankumarkolli: Where is it used?

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: This we are failing in service, if they both are set. Should we do this at SDK level?

aavasthy: updated logic and now service will throw exception if both consistencylevel and readconsistencystrategy are set.

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: ConsistencyLevel also has a copy on local v3 repo and also the direct package ReadConsistencyStrategy file has accessibility issues as it is under preview flag https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB?path=/Product/Microsoft.Azure.Documents/SharedFiles/ReadConsistencyStrategy.cs&version=GBmaster&_a=contents&line=13&lineStyle=plain&lineEnd=14&lineStartColumn=1&lineEndColumn=1

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: Updated the logic and now all validation for readconsistencystrategy is being done on the service.

Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
aavasthy: It's set by RequestInvokerHandler, then copied to DocumentServiceRequestContext.ReadConsistencyStrategy in OnBeforeRequestHandler so the Direct layer's ConsistencyReader/QuorumReader can pick it up.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -38,12 +39,21 @@ internal class RequestInvokerHandler : RequestHandler
...
             Cosmos.ConsistencyLevel? requestedClientConsistencyLevel,
+            Cosmos.ReadConsistencyStrategy? requestedClientReadConsistencyStrategy,
             Cosmos.PriorityLevel? requestedClientPriorityLevel,
             int? requestedClientThroughputBucket)
         {
             this.client = client;
 
-            this.RequestedClientConsistencyLevel = requestedClientConsistencyLevel;       
+            if (requestedClientConsistencyLevel.HasValue && requestedClientReadConsistencyStrategy.HasValue)
+            {
```
kirankumarkolli: Let it be driven from service fully

Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
kirankumarkolli: As discussed its not needed.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: yes removed this from here and now service will throw exception if both consistencylevel and readconsistencystrategy are set

Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
aavasthy: yes removed this and readconsistency strategy will now only be set by header

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -77,6 +79,7 @@ public override async Task<ResponseMessage> SendAsync(
                 request.Headers.Add(HttpConstants.HttpHeaders.SupportedSerializationFormats, RequestInvokerHandler.BinarySerializationFormat);
             }
 
+            await this.ValidateAndSetReadConsistencyStrategyAsync(request);
```
kirankumarkolli, line 82: Should the overrides be limited to Dataplane only (i.e exclude metadata)?

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /// <summary>
+        /// Eventual consistency - reads from a single random secondary replica.
+        /// This strategy also covers ConsistentPrefix consistency level behavior,
+        /// as both use the same underlying read mechanism (single replica read).
+        /// </summary>
+        Eventual = 1,
+
+        /// <summary>
+        /// Session consistency - monotonic reads, writes and read-your-writes guarantees within a session.
```
kirankumarkolli, line 35: Lets also include that sessionToken from writes needed and a code reference.

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        Session = 2,
+
+        /// <summary>
+        /// Quorum read with GLSN barrier - returns the latest committed version in the current region.
```
kirankumarkolli:
```suggestion
        /// Quorum read with GLSN barrier - returns the latest committed version in the current read region.
```

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
```
kirankumarkolli, line 44: Call it our as Bounded equivalent from ConssitencyLevel?

Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
```diff
@@ -345,8 +345,8 @@ internal RequestMessage Clone(ITrace newTrace, CloneableStream cloneContent)
                     : new PartitionKeyRangeIdentity(this.PartitionKeyRangeId.CollectionRid, this.PartitionKeyRangeId.PartitionKeyRangeId);
             }
 
-            clone.UseGatewayMode = this.UseGatewayMode;
-            clone.ContainerId = this.ContainerId;
+            clone.UseGatewayMode = this.UseGatewayMode;
```
kirankumarkolli: undo?

Microsoft.Azure.Cosmos/src/RequestOptions/ChangeFeedRequestOptions.cs
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.Cosmos
     using System;
     using System.Diagnostics;
     using Microsoft.Azure.Cosmos.Serializer;
+    using Microsoft.Azure.Documents;
```
kirankumarkolli: Is it needed?

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -504,6 +507,33 @@ private async Task ValidateAndSetConsistencyLevelAsync(RequestMessage requestMes
             }
         }
 
+        /// <summary>
+        /// Validate and set the ReadConsistencyStrategy header.
+        /// </summary>
+        private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMessage)
```
xinlian12, line 513: **🔴 Blocking** · Cross-SDK: Missing ConsistencyLevel Header Mapping

**ConsistencyLevel header not set when ReadConsistencyStrategy is used — breaks service telemetry**

The Java SDK's `RequestHelper.getReadConsistencyStrategyToUse()` maps `ReadConsistencyStrategy` back to a `ConsistencyLevel` header for service telemetry:
- `LatestCommitted` → `BoundedStaleness`
- `GlobalStrong` → `Strong`
- `Session` → `Session`
- `Eventual` → `Eventual`

This method only sets the `x-ms-cosmos-read-consistency-strategy` header and does NOT set `ConsistencyLevel`. Service-side telemetry and Kusto logging depend on the `ConsistencyLevel` header. Without it, operations using `ReadConsistencyStrategy` will appear as 'no consistency level' in service dashboards, breaking observability.

---

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <para>
+    /// <see cref="GlobalStrong"/> is only valid for accounts configured with Strong consistency.
+    /// </para>
+    /// </remarks>
+#if PREVIEW
+    public
+#else
+    internal
+#endif
+    enum ReadConsistencyStrategy
```
xinlian12, line 25: **🟡 Recommendation** · Cross-SDK: Missing DEFAULT Enum Value

**Java SDK has a DEFAULT value that .NET is missing**

The Java SDK's `ReadConsistencyStrategy` enum includes a `DEFAULT` value meaning 'use the normal consistency level behavior':
```java
DEFAULT("Default"),  // Use default read behavior
```n
Without `DEFAULT`, there's no way for a request-level option to explicitly say 'don't override — use whatever the client or account specifies.' Setting the property to `null` works but is less explicit and asymmetric with Java. Consider adding `Default = 0` to match.

---

File: Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs [hunk 14]
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /// <summary>
+        /// Eventual consistency - reads from a single random secondary replica.
+        /// This strategy also covers ConsistentPrefix consistency level behavior,
+        /// as both use the same underlying read mechanism (single replica read).
+        /// </summary>
+        Eventual = 1,
```
xinlian12, line 32: **🟡 Recommendation** · Correctness: Enum Zero Value Gap

**`default(ReadConsistencyStrategy)` produces invalid value 0**

The enum starts at `Eventual = 1` with no value for `0`. In C#, `default(ReadConsistencyStrategy)` produces `0`, which doesn't correspond to any defined strategy. If any code path accidentally uses `default` instead of `null` (e.g., uninitialized struct field, deserialization edge case), it would silently produce an invalid `"0"` header value the service won't recognize.

Consider adding `None = 0` or `Default = 0` explicitly (which also addresses the missing DEFAULT value from the Java SDK).

---

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
xinlian12, line 513: **🟡 Recommendation** · Design: Method Name Implies Validation That Doesn't Exist

**`ValidateAndSetReadConsistencyStrategyAsync` does zero validation and no async work**

Despite the name containing 'Validate', this method performs no validation. Compare with `ValidateAndSetConsistencyLevelAsync` which fetches the account consistency level and validates the override is a valid weakening. The Java SDK validates that `GlobalStrong` is only used with Strong accounts in `RequestHelper.validateReadConsistencyStrategy()`.

Additionally, the method returns `Task.CompletedTask` — it does no async work but is declared as `Task`-returning.

Suggestions:
1. Add `GlobalStrong` validation against the account consistency level (matching Java SDK)
2. Either make the method synchronous (`void`) or add real validation that justifies the async signature

---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
```diff
@@ -123,7 +124,7 @@ public void VerifyCosmosConfigurationPropertiesGetUpdated()
                 .WithThrottlingRetryOptions(maxRetryWaitTime, maxRetryAttemptsOnThrottledRequests)
                 .WithBulkExecution(true)
                 .WithSerializerOptions(cosmosSerializerOptions)
-                .WithConsistencyLevel(consistencyLevel)
+                .WithReadConsistencyStrategy(Cosmos.ReadConsistencyStrategy.LatestCommitted)
```
xinlian12: **🟡 Recommendation** · Testing: Existing ConsistencyLevel Test Coverage Removed

**Test replaces ConsistencyLevel verification with ReadConsistencyStrategy**

This line changed from `.WithConsistencyLevel(consistencyLevel)` to `.WithReadConsistencyStrategy(...)`, and the assertions below were updated to check `ReadConsistencyStrategy` instead of `ConsistencyLevel`. This means the `WithConsistencyLevel()` builder path is no longer verified in this test.

If a future change breaks `ConsistencyLevel` builder integration, there's no test to catch it. The new feature's test should be **additive** — keep the original `ConsistencyLevel` test intact and add a separate test method for `ReadConsistencyStrategy`.

---

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: Dont see ConsistencyLevel filtering by resource type. We can let the service handle this, for metadata operations anyway the service will ignore this header.

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Handler/RequestMessage.cs
aavasthy: removed

Microsoft.Azure.Cosmos/src/RequestOptions/ChangeFeedRequestOptions.cs
aavasthy: removed

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: we did discuss this before when we were creating the ReadConsistencyStrategy class in direct package and decided to remove Default enum and instead just treat null as Default.
https://msdata.visualstudio.com/DefaultCollection/CosmosDB/_git/CosmosDB/pullRequest/1948372?discussionId=31193987#1771546448

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: we did discuss this before when we were creating the ReadConsistencyStrategy class in direct package and decided to remove Default enum and instead just treat null as Default.
https://msdata.visualstudio.com/DefaultCollection/CosmosDB/_git/CosmosDB/pullRequest/1948372?discussionId=31193987#1771546448

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: Both methods get called to set ReadConsistencyStrategy and COnsistencyLevel respectively
            await this.ValidateAndSetReadConsistencyStrategyAsync(request);
            await this.ValidateAndSetConsistencyLevelAsync(request);

aavasthy: if both headers are set then service will actually fail as ideally client should only set one of these two values.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosClientOptionsUnitTests.cs
aavasthy: updated the test so that it continues to test consistencylevel and added new test for readconsistencystrategy

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: At SDK level it just validates if the value is null. This validation(Add GlobalStrong validation against the account consistency level (matching Java SDK)) is added at service level which is already in direct package PR https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB/pullrequest/1948372

PR #5686, opened 2026-03-09
[Internal] DTS: Refactors casing in DTS response and adds tests

Meghana-Palaparthi:
## Description

This pull request introduces several important changes to the distributed transactions implementation in Cosmos DB. The main focus is on improving the serialization and deserialization of distributed transaction operation results, updating property names to follow consistent casing, and modernizing how resource bodies are handled in responses. Additionally, it removes the end-to-end tests for distributed transactions and adds other test files.

Key changes include:

### Serialization and Deserialization Improvements

* Replaced the base64-encoded string property `resourcebody` with a new `resourceBody` property of type `JsonElement` in `DistributedTransactionOperationResult`, allowing direct handling of JSON objects instead of base64 strings. The setter now serializes the JSON element to bytes and populates the `ResourceStream`.
* Updated the deserialization logic in `FromJson` to use `json.GetRawText()` for more accurate deserialization.

### Property Naming and API Consistency

* Changed JSON property names to use camelCase (e.g., `statuscode` → `statusCode`, `substatuscode` → `subStatusCode`) for consistency with common JSON conventions.
* Added `subStatusCodeValue` as a public property for serialization, mapping to the internal `SubStatusCode` enum.

### Header and Operation Type Handling

* Updated header value serialization for `OperationType` and `ResourceType` to use explicit conversion methods (`ToOperationTypeString()`, `ToResourceTypeString()`) instead of `.ToString()`, ensuring correct string representations in headers.

### Test Code Cleanup

* Removed the entire `DistributedTransactionE2ETests.cs` file, which contained end-to-end tests for distributed transactions.
* Added below test files -
```
Microsoft.Azure.Cosmos.EmulatorTests\DistributedTransaction
----> DistributedTransactionTests.cs

Microsoft.Azure.Cosmos.Tests\DistributedTransaction
----> DistributedTransactionResponseTests.cs
----> DistributedTransactionSerializerTests.cs
----> DistributedWriteTransactionTests.cs
```

### Minor Cleanups

* Removed `SessionToken` from response header and deserialization logic

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
@@ -138,19 +138,7 @@ namespace Microsoft.Azure.Cosmos
         /// <returns>The deserialized operation result.</returns>
         internal static DistributedTransactionOperationResult FromJson(JsonElement json)
         {
-            DistributedTransactionOperationResult result = null;
-            try
-            {
-                result = JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json, DistributedTransactionOperationResult.CaseInsensitiveOptions);
-            }
-            catch (JsonException)
-            {
-            }
-
-            if (result == null)
-            {
-                return new DistributedTransactionOperationResult(HttpStatusCode.InternalServerError);
-            }
+            DistributedTransactionOperationResult result = JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json, DistributedTransactionOperationResult.CaseInsensitiveOptions);
 
             if (json.TryGetProperty("resourceBody", out JsonElement resourceBody)
                 && resourceBody.ValueKind != JsonValueKind.Undefined


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -127,14 +117,35 @@ internal string ResourceBodyBase64
         [JsonIgnore]
         internal ITrace Trace { get; set; }
 
+        /// <summary>
+        /// Gets or sets the resource body as a JSON element.
+        /// Setting this property populates <see cref="ResourceStream"/> with the serialized bytes.
+        /// </summary>
+        [JsonInclude]
+        [JsonPropertyName("resourceBody")]
+        public virtual JsonElement? ResourceBody
```
kirankumarkolli: JsonElement as contract will tie to a specific serializer. 
Ideally this contract should be Stream

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -314,7 +314,6 @@ private static async Task<DistributedTransactionResponse> PopulateFromJsonConten
 
                             DistributedTransactionOperationResult operationResult = DistributedTransactionOperationResult.FromJson(operationElement);
                             operationResult.Trace = trace;
-                            operationResult.SessionToken ??= responseMessage.Headers.Session;
```
kirankumarkolli, line 317: Is it not expected?

Meghana-Palaparthi: Session token details aren't available in response headers. So we can't populate it this way. Also, we already have it in [DistributedTransactionOperationResult](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/236b4401c280207d3887331030b163a34b40d8fa/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs#L85), where it is getting deserialized and added to operation level responses

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
Meghana-Palaparthi: Agreed. I'll change `ResourceBody` from public to internal so `JsonElement` is not exposed in the public contract. The public API surface only exposes `Stream ResourceStream`. The intermediary `ResourceBody` property is needed because json cannot deserialize directly into a stream. It deserializes the JSON token into a JsonElement, and the setter converts it to a memoryStream to populate ResourceStream.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -134,7 +127,22 @@ internal string ResourceBodyBase64
         /// <returns>The deserialized operation result.</returns>
         internal static DistributedTransactionOperationResult FromJson(JsonElement json)
         {
-            return JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json);
+            DistributedTransactionOperationResult result = JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json.GetRawText());
```
Praveen-Msft: protect from deserialization failures?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedWriteTransactionTests.cs
```diff
@@ -0,0 +1,399 @@
...
+
+            return contextMock;
+        }
+
+        private static ResponseMessage BuildSuccessResponse(int operationCount)
+        {
+            List<string> results = new List<string>();
+            for (int i = 0; i < operationCount; i++)
+            {
+                results.Add($@"{{""index"":{i},""statuscode"":201}}");
```
NaluTripician: **🟡 Recommendation** · Test Quality: Incorrect Mock JSON

**Mock response uses old lowercase `statuscode` instead of camelCase `statusCode`**

This `BuildSuccessResponse` helper uses `statuscode` (lowercase):

`csharp
results.Add($@"{{""index"":{i},""statuscode"":201}}");
``n
But the production code now declares `[JsonPropertyName("statusCode")]` (camelCase). System.Text.Json uses **case-sensitive** property matching by default, so the deserialized `StatusCode` property will silently remain at its default value (`0`) instead of `201`.

Current tests pass only because they don't assert on individual operation `StatusCode` values, but this helper is a trap for any future test that does. The same issue applies to `BuildErrorResponse` below.

**Suggestion:** Change `statuscode` to `statusCode` in both helpers to match the updated wire contract.

---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedTransactionSerializerTests.cs
```diff
@@ -0,0 +1,330 @@
...
+
+            return contextMock;
+        }
+
+        private ResponseMessage BuildSuccessResponse(int operationCount)
+        {
+            List<string> results = new List<string>();
+            for (int i = 0; i < operationCount; i++)
+            {
+                results.Add($@"{{""index"":{i},""statuscode"":201}}");
```
NaluTripician: **🟡 Recommendation** · Test Quality: Incorrect Mock JSON

**Same casing issue as DistributedWriteTransactionTests.cs**

This `BuildSuccessResponse` also uses `statuscode` (lowercase). Commit `b84dc1e` fixed this in `DistributedTransactionTests.cs` (emulator tests) but missed this file and `DistributedWriteTransactionTests.cs`.

**Suggestion:** Change `statuscode` to `statusCode` here as well.

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
NaluTripician: **🟡 Recommendation** · Reliability: Error Handling [Already posted by @Praveen-Msft — Unresolved]

**`JsonSerializer.Deserialize` call is unprotected against malformed JSON**

`csharp
DistributedTransactionOperationResult result = JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json.GetRawText());
``n
If the JSON element contains unexpected types (e.g., `statusCode` as a string instead of an integer, or an unrecognized enum value for `SubStatusCode`), this throws an unhandled `JsonException`. Depending on whether the caller (`PopulateFromJsonContent`) has a try-catch, this could fail the entire response parsing for all operations, not just the malformed one.

Reinforcing @Praveen-Msft's earlier comment — consider wrapping this in a try-catch and returning a result with `InternalServerError` status on failure, so one bad operation result doesn't crash the entire batch.

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -60,7 +59,7 @@ protected DistributedTransactionOperationResult()
         /// Gets the HTTP status code returned by the operation.
         /// </summary>
         [JsonInclude]
-        [JsonPropertyName("statuscode")]
+        [JsonPropertyName("statusCode")]
```
NaluTripician, line 68: **🟡 Recommendation** · Compatibility: Wire Protocol Change

**JSON property casing change is a breaking wire protocol change**

The property name changed from `statuscode` (lowercase) to `statusCode` (camelCase). Since System.Text.Json uses case-sensitive matching by default, if the server still sends lowercase property names, deserialization will silently return default values: `StatusCode = 0`, `SubStatusCode = 0`, and no `ResourceStream`.

Please confirm this change is coordinated with a corresponding server-side update. If backward compatibility with older server responses is needed, consider using `JsonSerializerOptions { PropertyNameCaseInsensitive = true }` in the `FromJson` method.

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
NaluTripician: **🟡 Recommendation** · Performance: Unnecessary Allocation

**`GetRawText()` allocates a string unnecessarily**

The old code used the `JsonElement` overload directly:
`csharp
return JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json);
``nThe new code allocates a string via `GetRawText()` just to re-parse it:
`csharp
JsonSerializer.Deserialize<...>(json.GetRawText());
``n
The `JsonSerializer.Deserialize<T>(JsonElement)` overload was already available (the old code used it). Reverting to the direct overload avoids an unnecessary string allocation per operation result, which matters for transactions with many operations.

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -91,33 +90,27 @@ protected DistributedTransactionOperationResult()
...
-        [JsonPropertyName("substatuscode")]
+        [JsonIgnore]
         internal virtual SubStatusCodes SubStatusCode { get; set; }
 
+        /// <summary>
+        /// Gets the sub-status code value as an unsigned integer.
+        /// </summary>
+        [JsonInclude]
+        [JsonPropertyName("subStatusCode")]
+        public virtual uint SubStatusCodeValue
```
NaluTripician, line 114: **🟡 Recommendation** · Test Coverage: Missing Tests

**No test verifies `SubStatusCodeValue` round-trip deserialization**

This new property maps `SubStatusCodes` enum to `uint` via `[JsonPropertyName("subStatusCode")]`. No test supplies a non-zero `subStatusCode` in mock JSON and verifies it deserializes correctly. The cast `(SubStatusCodes)value` could silently produce unexpected enum values for codes not defined in the enum.

**Suggestion:** Add a test with e.g. `subStatusCode: 449` in the mock JSON and assert the deserialized value.

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -91,33 +90,27 @@ protected DistributedTransactionOperationResult()
...
-            }
-        }
-
         /// <summary>
         /// Request charge in request units for the operation.
         /// </summary>
+        [JsonInclude]
         [JsonPropertyName("requestCharge")]
-        internal virtual double RequestCharge { get; set; }
+        public virtual double RequestCharge { get; internal set; }
```
NaluTripician, line 104: **🟡 Recommendation** · Test Coverage: Missing Tests

**No test verifies `RequestCharge` deserialization**

`RequestCharge` was changed from `internal` to `public` with `[JsonInclude]` added, but no test in the new suite verifies it deserializes correctly from response JSON.

**Suggestion:** Add `requestCharge: 5.43` to a mock response and assert the deserialized value.

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -42,10 +42,9 @@ internal DistributedTransactionOperationResult(DistributedTransactionOperationRe
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DistributedTransactionOperationResult"/> class.
-        /// This protected constructor is intended for use by derived classes.
         /// </summary>
         [JsonConstructor]
-        protected DistributedTransactionOperationResult()
+        public DistributedTransactionOperationResult()
```
NaluTripician, line 53: **🟡 Recommendation** · API Surface: Constructor Visibility

**`[JsonConstructor]` changed from `protected` to `public`**

While the class is normally `internal` (via `#if INTERNAL`), when compiled as `public`, this lets consumers construct result objects with default (zero) values, bypassing SDK construction logic. Result types are typically not meant to be user-constructed.

If `[JsonConstructor]` requires accessibility for deserialization, consider whether `internal` would suffice (it does for System.Text.Json when the type itself is internal).

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -134,7 +127,22 @@ internal string ResourceBodyBase64
         /// <returns>The deserialized operation result.</returns>
         internal static DistributedTransactionOperationResult FromJson(JsonElement json)
         {
-            return JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json);
+            DistributedTransactionOperationResult result = JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json.GetRawText());
+
+            if (json.TryGetProperty("resourceBody", out JsonElement resourceBody)
```
NaluTripician, line 143: **🟢 Suggestion** · Maintainability: Mixed Deserialization Pattern

**`FromJson` mixes automatic and manual deserialization**

This method uses `JsonSerializer.Deserialize` for most properties, then manually processes `resourceBody` via `TryGetProperty`. This dual pattern is fragile — any future property needing special handling requires updates in both the class attributes AND the manual code in `FromJson`, with no compile-time safety net.

Consider a custom `JsonConverter<DistributedTransactionOperationResult>` to centralize all deserialization logic in one place.

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -134,7 +127,22 @@ internal string ResourceBodyBase64
...
+                && resourceBody.ValueKind != JsonValueKind.Undefined
+                && resourceBody.ValueKind != JsonValueKind.Null)
+            {
+                if (resourceBody.ValueKind != JsonValueKind.Object)
```
NaluTripician, line 148: **🟢 Suggestion** · Robustness: Type Restriction

**`resourceBody` is restricted to JSON objects only**

`csharp
if (resourceBody.ValueKind != JsonValueKind.Object)
    throw new JsonException(...);
``n
This throws if the server returns a non-object `resourceBody` (e.g., an array). If the wire protocol could evolve to support non-object payloads, this restriction would need to change. Worth documenting this assumption, even if correct today.

---

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
Meghana-Palaparthi: `resourcebody` is the only property that requires custom handling. Changing deserialization of other properties to reflect the deserialization of `resourcebody` could introduce a lot of unnecessary manual handling

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
Meghana-Palaparthi: The project resolves System.Text.Json 6.0.10. In STJ 6.x, [JsonConstructor] is only honored on public constructors. internal would silently break deserialization.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -127,14 +126,47 @@ internal string ResourceBodyBase64
...
+
         /// <summary>
         /// Creates a <see cref="DistributedTransactionOperationResult"/> from a JSON element.
         /// </summary>
         /// <param name="json">The JSON element containing the operation result.</param>
         /// <returns>The deserialized operation result.</returns>
         internal static DistributedTransactionOperationResult FromJson(JsonElement json)
         {
-            return JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json);
+            DistributedTransactionOperationResult result = null;
```
kirankumarkolli: Error handling: Lets spec the error handling and then implement

PR #5689, opened 2026-03-10
[Internal] DTS: Adds retries in DTS when isRetriable is true and on timeout

Meghana-Palaparthi:
### Two-loop architecture
 
 DTX commits flow through two retry loops with distinct responsibilities:
 
 ```
 DistributedTransactionCommitter.ExecuteCommitWithRetryAsync   ← OUTER LOOP
   └─ clientContext.ProcessResourceOperationStreamAsync
        └─ AbstractRetryHandler (inner while-loop)
             └─ ClientRetryPolicy.ShouldRetryInternalAsync       ← INNER LOOP
 ```
 
 The coordinator returns **two response shapes** for the same HTTP status codes, which is why a single loop cannot own retries:
 
 | Shape | Status / sub-status | Body | Retriability signal | Owner |
 |---|---|---|---|---|
 | **Envelope failure** | 408, 449/5352, 500/5411-5413 | `Content == null` | implicit (status code) | Inner (CRP) |
 | **Semantic failure** | 408, 449/5352, 452 | JSON with per-op results | `"isRetriable": true` in body | Outer (committer) |
 | **Throttle** | 429/3200 | varies | `Retry-After` header | `ResourceThrottleRetryPolicy` |
 
 ### The amplification problem
 
 Without a routing rule, a 449/5352 carrying a body would be retried by **both** loops:
 inner CRP (10 attempts) × outer committer (10 attempts) = up to **100 wire requests per user call**.
 
 ### Options considered
 
 #### Option A — Status-code partition (outer-loop exclusion list)
 
 CRP retries its owned codes unconditionally. The committer adds an `IsOuterLoopRetriable()` helper that mirrors CRP's owned set and excludes them.
 
 - ❌ **Dual-maintenance hazard.** The owned-code set lives in two files. Any drift re-introduces the 10×10 amplification.
 - ❌ **Discards body data.** `AbstractRetryHandler` abandons the `ResponseMessage` without reading it on inner-loop retries. Per-op session tokens, ETags, 
and `ResourceStream` payloads are silently dropped on every CRP retry.
 - ❌ **Brittle to contract evolution.** If the coordinator starts attaching bodies to 449 in a future revision, CRP retries blindly and the `isRetriable` 
signal never reaches the committer.
 
 #### Option B — Body-presence deferral ✅ **(chosen)**
 
 Before CRP retries `{408, 449/5352, 429/3200}`, it checks `hasResponseBody = (responseMessage?.Content != null)`. If a body is present, CRP returns 
`NoRetry()` and defers to the outer loop, which parses `isRetriable` from that body. If no body, CRP retries immediately.
 
 - ✅ **Single source of truth** — one predicate, one file.
 - ✅ **Preserves per-op data** — body-bearing responses always reach `FromResponseMessageAsync → MergeSessionTokens`.
 - ✅ **Forward compatible** — new body-bearing codes route through `isRetriable` automatically with no SDK change.
 
 ### Retry budgets and delays (after this PR)
 
 | Code | Loop | Budget | Delay |
 |---|---|---|---|
 | 408, 449/5352 (no body) | Inner (CRP) | 10 | 1 s flat (honors `Retry-After`) |
 | 500/5411-5413 | Inner (CRP) | 10 | 100 ms → 5 s exponential + jitter |
 | 408, 449/5352, 452 (body + `isRetriable: true`) | Outer (committer) | 10 | 1 s → 32 s exponential + jitter |
 | 429/3200 | `ResourceThrottleRetryPolicy` | 9 (default) | `Retry-After` header |
 
 ### Other notable changes in this PR
 
 - **408 endpoint-marking guard.** Added `if (!isDtxRequest)` before `TryMarkEndpointUnavailableForPkRange`. DTX always goes through gateway; marking a 
partition endpoint unavailable on a coordinator 408 would skew routing for unrelated requests.
 - **Idempotency on retry.** `DistributedTransactionServerRequest` pre-serializes the body to `byte[]` and vends a fresh `MemoryStream` per attempt via 
`CreateBodyStream()`. The idempotency token header is re-applied by the `requestEnricher` closure on each attempt. Both inner and outer retries are safe.
 - **Body parser hardening.** `DistributedTransactionResponse` disposes the original `ResponseMessage` before constructing the synthetic 500 fallback, and 
emits a `DefaultTrace.TraceWarning` on `JsonException`.

---

## PR Summary

This pull request introduces significant improvements to the retry logic for distributed transactions (DTX) in Cosmos DB, focusing on more robust, efficient, and nuanced handling of retriable error codes and infrastructure failures. The changes include new retry limits and backoff strategies specifically for DTX scenarios, improved differentiation of retry logic based on request type, and the introduction of jitter to avoid synchronized client retries. Additionally, the code is refactored to better support testing and configuration of retry behavior.

**Distributed Transaction (DTX) Retry Logic Enhancements:**

* Introduced new retry counters and configurable limits for DTX retries (`MaxDtxRetryCount`, `MaxDtxInfraFailureRetryCount`) and added request-type detection (`isDtxRequest`) to apply DTX-specific logic. [[1]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4L26-R32) [[2]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4R43-R48) [[3]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4R229-R231)
* Implemented DTX-specific handling in `ShouldRetryInternalAsync`, including distinct logic for various DTX error codes (e.g., 408, 449/5352, 429/3200, 500/5411-5413), with separate retry budgets and backoff strategies for infrastructure failures.
* Added bounded exponential backoff with ±25% jitter for DTX infrastructure failures to prevent synchronized retries across clients. [[1]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4L26-R32) [[2]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4R724-R748)
* Ensured DTX retries are not mistakenly counted as endpoint failures for non-DTX traffic, preventing negative impact on routing.

**General Retry and Committer Improvements:**

* Updated `ShouldRetryAsync` and related methods to pass additional context (e.g., `RetryAfter`, response body presence) to internal retry logic, allowing more nuanced retry decisions. [[1]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4L121-R131) [[2]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4L135-R146) [[3]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4R185-R191) [[4]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4L249-R269)
* Refactored `DistributedTransactionCommitter` to allow configurable base delay and injected delay provider for improved testability and flexibility, and added thread-local jitter for retry delays.

**Session Token Retry Policy Adjustments:**

* Updated session token retry logic to clarify and simplify when to set the hub region processing header, and to stop retries after the correct number of attempts. [[1]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4R361-R368) [[2]](diffhunk://#diff-2b056512ca285b1d95e025e31f60345059fa92d958becc38f90a6fb54ce1bbb4L452-L474)

These changes collectively make the retry policy for distributed transactions more resilient, efficient, and easier to maintain and test, especially in the face of transient infrastructure failures and complex error scenarios.
## Type of change

- [✓] New feature (non-breaking change which adds functionality)

## Closing issues


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -15,15 +16,29 @@ namespace Microsoft.Azure.Cosmos
 
     internal class DistributedTransactionCommitter
     {
+        private const int MaxRetryAttempts = 3;
```
Praveen-Msft: Is there a timeout for each retry? is it same for all retries?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -41,27 +55,75 @@ await DistributedTransactionCommitterUtils.ResolveCollectionRidsAsync(
...
                 throw;
             }
         }
 
+        private async Task<DistributedTransactionResponse> ExecuteCommitWithRetryAsync(
+            DistributedTransactionServerRequest serverRequest,
+            CancellationToken cancellationToken)
+        {
+            int attempt = 0;
+            while (true)
```
xinlian12: 🟡 **Recommendation: Unbounded retry loop needs a safety ceiling**

This `while (true)` loop retries indefinitely — the only exits are success, a non-retriable error, or `CancellationToken` cancellation. Every other retry policy in this SDK enforces a hard bound:

| Policy | Max |
|--------|-----|
| `ClientRetryPolicy` | 120 |
| `ResourceThrottleRetryPolicy` | 9 + 30s cumulative |
| `BulkExecutionRetryPolicy` | 10 |
| `WebExceptionRetryPolicy` | 30s window |

The public API `CommitTransactionAsync(CancellationToken cancellationToken = default)` means a caller can invoke this with `CancellationToken.None`. If the server keeps returning `isRetriable: true` (e.g. during a prolonged service issue), the task never completes — a silent hang that's extremely hard to diagnose.

I understand this is intentional per commit `d547817fa` ("Change to unbounded retries") and that idempotent transactions can safely retry. However, even a generous safety ceiling (e.g. 30–120 attempts, consistent with `ClientRetryPolicy`) would prevent pathological hangs without limiting real-world usage. A final trace error before giving up would also aid diagnostics.

At minimum, consider documenting in the public API that callers **must** provide a meaningful cancellation token with a timeout.

<sub>📎 Validated against SDK retry conventions (`BackoffRetryUtility`, `ClientRetryPolicy`, `ResourceThrottleRetryPolicy`) — all use bounded retries.</sub>

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -41,27 +55,75 @@ await DistributedTransactionCommitterUtils.ResolveCollectionRidsAsync(
...
+                    DefaultTrace.TraceWarning(
+                        $"Distributed transaction commit timed out (attempt {attempt + 1}). " +
+                        $"Retrying with idempotency token {serverRequest.IdempotencyToken}.");
+                    await Task.Delay(this.GetRetryDelay(attempt), cancellationToken);
+                    attempt++;
+                    continue;
+                }
+
+                if (!response.IsSuccessStatusCode
+                    && (response.IsRetriable || response.StatusCode == HttpStatusCode.RequestTimeout))
```
xinlian12: 🟡 **Recommendation: Potential double-retry amplification on HTTP 408**

The outer DTS loop retries on 408 both via the `CosmosException` catch (line 84) and the response status check (line 95). However, `ExecuteCommitAsync` calls `ProcessResourceOperationStreamAsync`, which flows through `RetryHandler` → `ClientRetryPolicy`. `ClientRetryPolicy.ShouldRetryOnEndpointFailureAsync` also handles 408 by marking endpoints unavailable and retrying with failover.

This creates two nested retry loops for 408:
- **Inner (pipeline):** Up to 120 retries for endpoint failover
- **Outer (DTS):** Unbounded retries

Each outer retry invocation spins up a fresh inner retry cycle, so a persistent 408 condition could generate `120 × N` total HTTP requests.

The `isRetriable` JSON flag is clearly DTS-specific and justified at this layer. For 408 specifically, could you confirm whether the gateway-mode pipeline (`UseGatewayMode = true`) already handles 408 retries at the transport level? If so, removing the explicit 408 check from the outer loop (relying only on `isRetriable`) would avoid amplification. If gateway mode bypasses transport-level 408 retries, then this outer check is necessary — a code comment explaining why would help future maintainers.

<sub>📎 Traced through `RequestHandler` → `RetryHandler` → `ClientRetryPolicy` pipeline.</sub>

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -41,27 +55,75 @@ await DistributedTransactionCommitterUtils.ResolveCollectionRidsAsync(
...
+
+                return response;
+            }
+        }
+
+        private TimeSpan GetRetryDelay(int attempt)
+        {
+            const int maxExponent = 5; // Cap backoff at 32x base delay
+            int exponent = Math.Min(attempt, maxExponent);
+            return TimeSpan.FromTicks((long)(this.retryBaseDelay.Ticks * Math.Pow(2, exponent)));
```
xinlian12: 🟡 **Recommendation: Add jitter to exponential backoff**

The backoff is fully deterministic (`1s, 2s, 4s, 8s, 16s, 32s, 32s...`). If multiple clients hit retriable DTS failures simultaneously (common during service disruptions), they will all retry at identical intervals, creating synchronized retry storms that amplify server pressure exactly when it's already struggling.

The SDK already has precedent for jitter via `RetryOptions.RandomSaltForRetryWithMilliseconds`. A simple approach:

```csharp
private TimeSpan GetRetryDelay(int attempt)
{
    const int maxExponent = 5;
    int exponent = Math.Min(attempt, maxExponent);
    double baseDelayMs = this.retryBaseDelay.TotalMilliseconds * Math.Pow(2, exponent);
    // Full jitter: uniform random between 0 and base delay
    double jitter = baseDelayMs * Random.Shared.NextDouble();
    return TimeSpan.FromMilliseconds(jitter + baseDelayMs * 0.5);
}
```

This decorrelates concurrent clients' retry timing with minimal effort.

<sub>📎 Cross-referenced with `RetryOptions.RandomSaltForRetryWithMilliseconds` in the SDK.</sub>

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -122,6 +129,11 @@ public virtual DistributedTransactionOperationResult this[int index]
         /// </summary>
         public virtual string ServerDiagnostics { get; }
```
xinlian12: 🟢 **Suggestion: `ServerDiagnostics` is now permanently null — consider cleanup**

This property is still declared as public API surface, but it's no longer populated. The constructor parameter `string serverDiagnostics` was replaced with `bool isRetriable` (commit `3d69cebe8`), and no other code path sets this property.

A consumer inspecting a `DistributedTransactionResponse` would see `ServerDiagnostics` and reasonably expect it to contain diagnostic information, but it will always be `null`.

Consider either removing the property or adding `[Obsolete]` / a comment indicating it's reserved for future use.

<sub>📎 Grepped for all references to `ServerDiagnostics` — zero read sites found.</sub>

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
Meghana-Palaparthi: I traced through the pipeline and the retry amplification doesn't occur.

ClientRetryPolicy does not retry 408. In [ShouldRetryInternalAsync](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/c2080630451a7fb10f42ca8de7a83087122ecaf6/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs#L258-L266), the 408 block only calls `TryMarkEndpointUnavailableForPkRange` and then falls through with no return. It ends up causing `AbstractRetryHandler` to return the ResponseMessage(408) straight to the caller — no inner retry.

So the outer DTS retry is the only retry for 408. Each outer attempt = exactly one inner pipeline call.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -302,16 +348,34 @@ private static async Task<DistributedTransactionResponse> PopulateFromJsonConten
...
+                // Unparseable body — fall back to default response construction.
+                return null;
+            }
+
+            using (responseJson)
+            {
+                JsonElement root = responseJson.RootElement;
+
+                if (root.TryGetProperty("isRetriable", out JsonElement isRetriableElement) &&
+                    isRetriableElement.ValueKind == JsonValueKind.True)
```
xinlian12, line 325: 🟡 **Recommendation | Test Coverage: `IsRetriable` parsing has no isolated unit tests**

The new `isRetriable` JSON field is parsed here with a strict boolean check (`ValueKind == JsonValueKind.True`), but `DistributedTransactionResponseTests.cs` has **zero tests** for this property. The only coverage is indirect — via the committer retry integration tests which use `CreateRetriableErrorResponseMessage()`.

**Why this matters:** If a future refactor changes the property name in the `TryGetProperty` call, or switches from `ValueKind` to `.GetBoolean()`, the regression would only surface through multi-layered retry tests — not at the unit where the parsing occurs.

**Suggested tests to add in `DistributedTransactionResponseTests`:**

1. `"isRetriable": true` → `response.IsRetriable == true`
2. `"isRetriable": false` → `response.IsRetriable == false`
3. Absent `isRetriable` field → `response.IsRetriable == false`
4. `"isRetriable": "true"` (string, not boolean) → `response.IsRetriable == false` (documents the contract)

These 4 tests pin the parsing behavior at the response level, independent of the retry loop.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedTransactionCommitterTests.cs
```diff
@@ -263,6 +266,688 @@ public async Task CommitTransactionAsync_MergesSessionTokens_OnFailureResponse()
                 "Session token should still be merged even when the DTC response indicates a failure.");
         }
 
+        // ─── Retry / Spec-Compliance Tests ─────────────────────────────────────
+
+        [TestMethod]
+        public async Task CommitTransaction_SucceedsOnFirstAttempt()
```
xinlian12, line 282: 🟡 **Recommendation | Test Coverage: Exponential backoff formula never exercised with real values**

Every retry test constructs the committer with `TimeSpan.Zero` as the base delay, which makes `baseDelayMs = 0 * 2^n = 0` — collapsing the entire exponential backoff + jitter formula to zero. The `maxExponent = 5` cap, the `Math.Pow(2, exponent)` growth, and the `(0.5 * base) + (base * random)` jitter spread are all untested.

**Why this matters:** The formula in `GetRetryDelay` is the core mechanism preventing retry storms during service disruptions. A bug here (e.g., wrong exponent, missing cap, overflow at high attempt counts) would be invisible with zero-delay tests.

**Suggested test:** Add a test with a non-zero `retryBaseDelay` (e.g., `TimeSpan.FromSeconds(1)`) and a `delayProvider` that captures delays across ~6 retries. Assert that:
- Delays increase monotonically (on average) with attempt number
- Delays are capped after `maxExponent` (attempt ≥ 5 should produce similar magnitude delays)
- Each delay falls within the expected range `[0.5 × base × 2^n, 1.5 × base × 2^n]`

This would catch formula regressions without making the test flaky (assert ranges, not exact values).

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -111,17 +111,24 @@ public virtual DistributedTransactionOperationResult this[int index]
         /// <summary>
         /// Gets the number of operation results in the distributed transaction response.
         /// </summary>
-        public virtual int Count => this.results?.Count ?? 0;
+        public virtual int Count
```
xinlian12: 🟢 **Suggestion | Test Coverage: New `ThrowIfDisposed` guards on `Count` and `GetEnumerator` lack post-dispose tests**

The existing `Indexer_AfterDispose_ThrowsObjectDisposedException` test verifies the indexer throws after dispose. However, the newly added `ThrowIfDisposed()` calls on `Count` (line 118) and `GetEnumerator()` (line 147) have no corresponding post-dispose test.

Two small tests (`Count_AfterDispose_ThrowsObjectDisposedException` and `GetEnumerator_AfterDispose_ThrowsObjectDisposedException`) in `DistributedTransactionResponseTests` would pin this defensive behavior and prevent accidental removal during future refactors.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -17,15 +17,31 @@ namespace Microsoft.Azure.Cosmos
 
     internal class DistributedTransactionCommitter
     {
+        private static readonly TimeSpan DefaultRetryBaseDelay = TimeSpan.FromSeconds(1);
+
         private readonly IReadOnlyList<DistributedTransactionOperation> operations;
         private readonly CosmosClientContext clientContext;
+        private readonly TimeSpan retryBaseDelay;
+        private readonly Random jitter = new Random();
```
xinlian12: 💬 **Observation | Thread Safety: `new Random()` seed collision on .NET Framework**

On `netstandard2.0` targeting .NET Framework, `new Random()` seeds from `Environment.TickCount` (~15ms resolution). If multiple `DistributedTransactionCommitter` instances are created within the same tick (e.g., concurrent transactions during a thundering herd), they produce **identical jitter sequences** — defeating the purpose of jitter exactly when it matters most.

On .NET Core / .NET 6+, `new Random()` uses a thread-safe interleaved seed, so this is not an issue there.

Since the SDK targets `netstandard2.0` and `Random.Shared` isn't available, no action is needed now. Just noting this for awareness in case the DTS feature is used in high-concurrency .NET Framework scenarios.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedTransactionCommitterTests.cs
```diff
@@ -345,5 +1030,188 @@ private Mock<CosmosClientContext> CreateMockContext(
...
+
+            json.Append("]}");
+
+            return new ResponseMessage(HttpStatusCode.OK)
+            {
+                Content = new MemoryStream(Encoding.UTF8.GetBytes(json.ToString()))
+            };
+        }
+
+        private static ResponseMessage CreateRetriableErrorResponseMessage()
```
xinlian12, line 966: 💬 **Observation | Testing: Missing `[Description]` attributes on new test methods**

All 5 pre-existing tests in this file use `[Description("...")]` attributes (e.g., `"Verifies that when the DTC response carries a session token, the token is merged into the SessionContainer"`). None of the 22 new tests include descriptions.

This is a minor pattern deviation from the file's own convention. The test method names are descriptive enough to be self-documenting, but adding `[Description]` attributes would maintain consistency with the existing tests in this file.

xinlian12:
✅ **Review complete** (39:02)

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -43,24 +59,136 @@ await DistributedTransactionCommitterUtils.ResolveCollectionRidsAsync(
...
                 // await this.AbortTransactionAsync(cancellationToken);
                 throw;
             }
         }
 
+        private async Task<DistributedTransactionResponse> ExecuteCommitWithRetryAsync(
+            DistributedTransactionServerRequest serverRequest,
+            CancellationToken cancellationToken)
+        {
+            int attempt = 0;
```
kirankumarkolli, line 79: On Retries PR: There are two kinds of retries 
- In-region reties: Timeout retry, retry-after (on retries) 
- Cross-region retries: when we get 403.3 from the service falling back to other regions etc...
--403.3 here should not impact current mainline flows as well

These are driven by ClientRetryPolicy in the request pipeline. 
Thoughts on leveraging similar model for DTX?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionServerRequest.cs
```diff
@@ -38,11 +38,21 @@ public static async Task<DistributedTransactionServerRequest> CreateAsync(
...
         {
-            MemoryStream bodyStream = this.bodyStream;
-            this.bodyStream = null;
-            return bodyStream;
+            if (this.serializedBody == null)
+            {
+                throw new InvalidOperationException("Request body has not been initialized. Use CreateAsync to construct a request.");
+            }
+
+            return new MemoryStream(this.serializedBody, writable: false);
```
xinlian12, line 50: 🟢 **Suggestion | Test Coverage: `CreateBodyStream()` reusability — the foundational retry enabler — lacks a direct unit test**

The refactor from `TransferBodyStream()` (one-shot ownership transfer that nullified the internal reference) to `CreateBodyStream()` (factory vending fresh read-only `MemoryStream` instances from a pre-serialized `byte[]`) is the **single most critical change** enabling retry safety. If this method regressed — e.g., someone reverted to the old move-semantic pattern or introduced a shared mutable stream — every retry attempt would fail with a disposed/empty body.

However, the committer tests mock `ProcessResourceOperationStreamAsync`, so the body stream is never actually read during retries. A regression in `CreateBodyStream()` would be invisible to those tests.

**Suggested test in a `DistributedTransactionServerRequest` test file:**

```csharp
[TestMethod]
public async Task CreateBodyStream_CalledMultipleTimes_ReturnsIndependentStreams()
{
    var request = await DistributedTransactionServerRequest.CreateAsync(
        CreateTestOperations(), serializer, CancellationToken.None);

    using MemoryStream stream1 = request.CreateBodyStream();
    using MemoryStream stream2 = request.CreateBodyStream();

    Assert.AreNotSame(stream1, stream2, "Each call must return a new instance.");
    Assert.AreEqual(0, stream1.Position);
    Assert.AreEqual(0, stream2.Position);
    Assert.IsTrue(stream1.Length > 0);
    CollectionAssert.AreEqual(stream1.ToArray(), stream2.ToArray(),
        "Both streams must contain identical serialized bytes.");

    stream1.Dispose();
    // stream2 must still be usable after stream1 is disposed
    Assert.IsTrue(stream2.CanRead, "Disposing one stream must not affect siblings.");
}

```

This pins the retry-safety contract at the source, independent of the mocked retry loop.

xinlian12:
✅ **Review complete** (40:04)

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -43,24 +59,136 @@ await DistributedTransactionCommitterUtils.ResolveCollectionRidsAsync(
...
-            catch (Exception ex)
+            catch (Exception ex) when (ex is not OperationCanceledException)
             {
                 DefaultTrace.TraceError($"Distributed transaction failed: {ex.Message}");
                 // await this.AbortTransactionAsync(cancellationToken);
                 throw;
             }
         }
 
+        private async Task<DistributedTransactionResponse> ExecuteCommitWithRetryAsync(
```
xinlian12, line 75: 🟡 **Recommendation | Diagnostics: Retry attempts produce disconnected traces — intermediate diagnostics are lost**

Each call to `ExecuteCommitAsync` creates a new root `ITrace` (line 189), which is disposed when the method returns. During retries, this means:

1. Attempt 1 → root trace created, request executed, trace disposed on return
2. Retry decision → `response.Dispose()` called, attempt's diagnostic data is gone
3. Attempt 2 → new, unrelated root trace created
4. ...repeat...

The returned `DistributedTransactionResponse` carries the trace from only the **final** attempt. All intermediate attempts' diagnostics — which contain the most useful debugging information (why did they fail? what latency? what server error details?) — are disposed and lost.

**Why this matters:** When debugging a DTx commit that took 30 seconds due to 5 retries, `CosmosDiagnostics` will only show the last (successful) 2-second call. The four failed attempts that explain *why* it took 30 seconds are invisible. This makes production diagnosis of retry-heavy scenarios significantly harder.

**Suggested fix:** Create a parent trace in `ExecuteCommitWithRetryAsync` and pass it into `ExecuteCommitAsync`, which would use `parentTrace.StartChild(...)` instead of `Trace.GetRootTrace(...)`:

```csharp
private async Task<DistributedTransactionResponse> ExecuteCommitWithRetryAsync(
    DistributedTransactionServerRequest serverRequest,
{
    using (ITrace retryTrace = Trace.GetRootTrace(
        "Distributed Transaction Commit", TraceComponent.Batch, TraceLevel.Info))
    {
        int attempt = 0;
        while (true)
        {
            // Each attempt becomes a child trace under the parent
            response = await this.ExecuteCommitAsync(serverRequest, retryTrace, attempt, cancellationToken);
            // ...
        }
    }
}

```

This preserves the full retry history in a single trace tree, consistent with how `AbstractRetryHandler` preserves diagnostics across retries in the standard pipeline.

xinlian12:
✅ **Review complete** (37:13)

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -17,15 +17,33 @@ namespace Microsoft.Azure.Cosmos
 
     internal class DistributedTransactionCommitter
     {
+        private static readonly TimeSpan DefaultRetryBaseDelay = TimeSpan.FromSeconds(1);
+
+        internal const int MaxIsRetriableRetryCount = 100;
```
kushagraThapar: I am curious, if this retries 100 times, and inside ClientRetryPolicy we retry 100 times, wouldn't it be like 100k retries for a single request?
To avoid this confusion, maybe we can combine these two retry mechanisms into a single layer?

NaluTripician: Good catch, I think worst this is 10k retries on a single request (before throttling) . If we want to combine maybe a passing response header to the client retry policy could work? If we wanted to keep the 2 retry layers and not have a new header we can probably just limit it to 3-5 retries here to be safe.

Meghana-Palaparthi: Thank you for pointing this out. The coordinator typically does not provide a response body for many failure status codes, which means isRetriable isn't available for the outer loop. Therefore, the scenario of 10k retries is unlikely. I'll double-check the coordinator code and add a safeguard, just in case the SDK receives a response body for failure status codes.

NaluTripician (approved):
LGTM

kushagraThapar (approved):

PR #5644, opened 2026-02-26
[Internal] Diagnostics: Adds spec for CosmosDiagnostics compaction summary mode

NaluTripician:
## Summary

Adds a spec sheet for the diagnostics compaction feature that introduces a `DiagnosticsVerbosity` option (`Detailed` vs `Summary`) to reduce unbounded diagnostics output size during high-retry scenarios (429 throttling, transient failures, cross-region failovers).

**Spec location:** `specs/diagnostics-compaction.md`

## Key Design Decisions

- **Summary mode** groups requests by region, keeps first/last in full detail, deduplicates middle requests with aggregate stats (count, RU, min/max/P50/avg latency)
- **Default is `Detailed`** — no behavioral change for existing users
- **Configuration:** `CosmosClientOptions.DiagnosticsVerbosity` + per-request `RequestOptions.DiagnosticsVerbosity` override + env vars
- **Max summary size:** 8KB default (4KB min), with truncated fallback
- **Compaction is serialization-only** — in-memory `ITrace` tree is unchanged

## Reference

Modeled after the Rust SDK's approach: Azure/azure-sdk-for-rust#3592

## Work Items

| # | Work Item | Description |
|---|-----------|-------------|
| 1 | Enum & Options Plumbing | `DiagnosticsVerbosity` enum, properties on `CosmosClientOptions`/`RequestOptions`, env var support |
| 2 | Summary Computation Engine | Trace tree walk, region grouping, first/last/aggregated groups |
| 3 | Serialization Integration | Branch `ToString()` on verbosity, size enforcement, truncated fallback |
| 4 | Contract Updates | Update `ContractEnforcementTests` baselines |
| 5 | Unit Tests | 21 tests covering summary engine, verbosity flow, edge cases |
| 6 | Integration Tests | Emulator-based tests for real operations with summary mode |
| 7 | Baseline Tests | Golden-file JSONs for summary mode serialization stability |

**Please review the spec and leave feedback before implementation begins.**

#5325

Issue #5325, opened 2025-07-29
[Fundamentals] Diagnostics - Reduce Cosmos Diagnostics Size for Storage Optimization


kundadebdatta:
### Acceptance Criteria:

- Find out ways to reduce Cosmos Diagnostics Size for Storage Optimization

PR #5695, opened 2026-03-17
Diagnostics: Adds DiagnosticsVerbosity Summary mode for compacted diagnostics output

NaluTripician:
## Summary

**⚠️ This is a v1 implementation and is subject to change based on feedback from the spec review (#5644).**

Implements the core diagnostics compaction feature described in the [diagnostics compaction spec](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5644). When opted into, `Summary` mode reduces unbounded `CosmosDiagnostics.ToString()` output (100+ KB in high-retry scenarios) down to ~2-4 KB by grouping requests by region and aggregating retry statistics.

### Problem

`CosmosDiagnostics.ToString()` grows **unboundedly** with retries. Each retry adds a full `StoreResponseStatistics` entry to the trace tree. In pathological scenarios (sustained 429 throttling, transient failures, cross-region failovers), a single operation's diagnostics can grow to hundreds of KB, causing:
- **Log truncation** — monitoring systems silently drop oversized entries
- **Memory pressure** — large diagnostic strings increase GC overhead
- **Readability** — operators cannot extract signal from noise

### New Public API

| API | Description |
|-----|-------------|
| `DiagnosticsVerbosity` enum | `Detailed` (default, current behavior) and `Summary` (compacted) |
| `CosmosClientOptions.DiagnosticsVerbosity` | Client-level preferred verbosity |
| `CosmosClientOptions.MaxDiagnosticsSummarySizeBytes` | Max bytes for summary output (default: 8KB, min: 4KB) |
| `CosmosDiagnostics.ToString(DiagnosticsVerbosity)` | New overload for explicit verbosity control |
| `CosmosClientBuilder.WithDiagnosticsVerbosity()` | Fluent builder method |
| `CosmosClientBuilder.WithMaxDiagnosticsSummarySizeBytes()` | Fluent builder method |
| `AZURE_COSMOS_DIAGNOSTICS_VERBOSITY` | Environment variable override (accepts `Summary` or `Detailed`) |
| `AZURE_COSMOS_DIAGNOSTICS_MAX_SUMMARY_SIZE` | Environment variable override (integer, minimum 4096) |

### How Summary Mode Works

1. **Walks** the `ITrace` tree to collect all `StoreResponseStatistics` and `HttpResponseStatistics`
2. **Groups** entries by region (preserving chronological order)
3. **Per region**: keeps **First** and **Last** request in full detail
4. **Aggregates** middle entries by `(StatusCode, SubStatusCode)` with: count, total RU, min/max/P50/avg latency
5. **Enforces size limit**: falls back to truncated JSON if output exceeds `MaxDiagnosticsSummarySizeBytes`

### Key Design Decisions

- **Default is `Detailed`** — zero behavioral change for existing users
- **Parameterless `ToString()` always returns `Detailed`** — backward compatibility guaranteed
- **In-memory `ITrace` tree unchanged** — compaction only happens at serialization time
- **GA (non-preview)** — additive, backward-compatible feature
- **`Lazy<string>` caching** — summary is computed once and reused
- **`MaxDiagnosticsSummarySizeBytes` wired through pipeline** — propagated from `CosmosClientOptions` to `CosmosTraceDiagnostics` at call sites with `ClientContext` access

### Example Summary Output

```json
{
  "Summary": {
    "DiagnosticsVerbosity": "Summary",
    "TotalDurationMs": 1234.5,
    "TotalRequestCharge": 245.5,
    "TotalRequestCount": 60,
    "RegionsSummary": [
      {
        "Region": "West US 2",
        "RequestCount": 50,
        "First": { "StatusCode": 429, "SubStatusCode": 3200, "DurationMs": 5, ... },
        "Last": { "StatusCode": 200, "SubStatusCode": 0, "DurationMs": 12, ... },
        "AggregatedGroups": [
          { "StatusCode": 429, "SubStatusCode": 3200, "Count": 48, "P50DurationMs": 12, ... }
        ]
      }
    ]
  }
}
```

### Files Changed

| File | Change |
|------|--------|
| `DiagnosticsVerbosity.cs` | **New** — public enum |
| `DiagnosticsSummaryWriter.cs` | **New** — core compaction engine |
| `DiagnosticsSummaryWriterTests.cs` | **New** — unit tests |
| `DiagnosticsSummaryBaselineTests.cs` | **New** — schema baseline tests |
| `DiagnosticsVerbosityEmulatorTests.cs` | **New** — emulator integration tests |
| `CosmosClientOptions.cs` | Added `DiagnosticsVerbosity`, `MaxDiagnosticsSummarySizeBytes`, env var support |
| `CosmosClientBuilder.cs` | Added `WithDiagnosticsVerbosity`, `WithMaxDiagnosticsSummarySizeBytes` |
| `ConfigurationManager.cs` | Added env var constants |
| `CosmosDiagnostics.cs` | Added abstract `ToString(DiagnosticsVerbosity)` overload |
| `CosmosTraceDiagnostics.cs` | Implemented overload with `Lazy<string>` caching |
| `EncryptionCosmosDiagnostics.cs` | Implemented overload (`SDKPROJECTREF`-gated) |
| `ContainerCore.cs` | Wired `MaxDiagnosticsSummarySizeBytes` from options |
| `ReadManyQueryHelper.cs` | Wired `MaxDiagnosticsSummarySizeBytes` from options |
| `CosmosLinqQuery.cs` | Wired `MaxDiagnosticsSummarySizeBytes` from options |
| `ChangeFeedEstimatorIterator.cs` | Wired `MaxDiagnosticsSummarySizeBytes` from options |
| `DotNetSDKAPI.net6.json` | Updated contract baseline |
| `CosmosClientOptionsUnitTests.cs` | Added tests for new properties, env vars, builder |

### Test Coverage (56 tests)

**Unit Tests (DiagnosticsSummaryWriterTests — 27 tests)**
- Enum/options defaults and validation
- Single region: 1 request, 2 requests, many retries (429)
- Multi-region failover with separate region summaries
- Mixed status codes producing multiple aggregated groups
- P50 computation: odd count, even count, single item
- Size enforcement: under limit, over limit (truncation)
- Empty trace, null region, null trace, invalid enum fallback
- Backward compatibility: parameterless `ToString()` unchanged
- Caching: `Lazy<string>` returns same instance

**Schema Baseline Tests (DiagnosticsSummaryBaselineTests — 9 tests)**
- Exact field set validation for each JSON level (summary, region, request entry, aggregated group, truncated)
- Field type consistency (numeric, string, array, object)
- DiagnosticsVerbosity field always "Summary"
- Truncation message content

**Options + Builder Tests (CosmosClientOptionsUnitTests — 12 new tests)**
- Default values, validation, large values
- Env var parsing (Summary, case-insensitive, invalid ignored, below-min ignored)
- Explicit property overrides env var
- Builder methods with validation

**Emulator Integration Tests (DiagnosticsVerbosityEmulatorTests — 8 tests)**
- CRUD operations (Create, Read, Replace, Delete) with Summary mode
- Query with Summary mode
- Summary output smaller than Detailed
- Parameterless ToString unchanged by Summary option
- Summary caching returns same instance

### Related

- **Spec PR:** #5644
- **Tracking issue:** #5325


Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -34,6 +40,15 @@ public CosmosTraceDiagnostics(ITrace trace)
 
             this.Value = rootTrace;
             this.accumulatedMetrics = new Lazy<ServerSideCumulativeMetrics>(() => PopulateServerSideCumulativeMetrics(this.Value));
+            this.cachedSummaryJson = new Lazy<string>(() =>
+            {
+                if (this.Value is Tracing.Trace rootConcreteTrace2)
+                {
+                    rootConcreteTrace2.SetWalkingStateRecursively();
```
NaluTripician: **🔴 Blocking** · Correctness: Redundant Operation

**Redundant `SetWalkingStateRecursively()` call — remove from Lazy lambda**

This `SetWalkingStateRecursively()` call is redundant. The `Lazy<string>` lambda calls it here, then immediately invokes `DiagnosticsSummaryWriter.WriteSummary()`, which calls `SetWalkingStateRecursively()` **again** at its top (line 38-41 of DiagnosticsSummaryWriter.cs).

Since `WriteSummary()` is a `public static` method that must be safe to call standalone (and tests call it directly), the call inside `WriteSummary()` is the correct one — it owns the responsibility. This one in the Lazy lambda should be removed.

**Why it matters:** This performs a full recursive tree traversal on every summary materialization — exactly the deep trace trees this feature targets (hundreds of retries). If `SetWalkingStateRecursively()` is ever made non-idempotent, this double-call becomes a real bug.

---

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -17,8 +17,14 @@ namespace Microsoft.Azure.Cosmos.Diagnostics
...
     {
         private readonly Lazy<ServerSideCumulativeMetrics> accumulatedMetrics;
+        private readonly Lazy<string> cachedSummaryJson;
 
         public CosmosTraceDiagnostics(ITrace trace)
+            : this(trace, CosmosClientOptions.DefaultMaxDiagnosticsSummarySizeBytes)
+        {
+        }
+
+        internal CosmosTraceDiagnostics(ITrace trace, int maxDiagnosticsSummarySizeBytes)
```
NaluTripician, line 27: **🟡 Recommendation** · Completeness: Missing Propagation Path

**`MaxDiagnosticsSummarySizeBytes` not propagated through `ResponseMessage.Diagnostics` getter**

This new internal constructor correctly accepts `maxDiagnosticsSummarySizeBytes`, and the PR wires it at 4 explicit call sites (ChangeFeedEstimatorIterator, CosmosLinqQuery, ReadManyQueryHelper, ContainerCore). However, `ResponseMessage.Diagnostics` — the **primary** code path for normal CRUD operations — lazily creates `new CosmosTraceDiagnostics(this.Trace)` using the parameterless constructor, which always defaults to 8192.

This means users who customize `MaxDiagnosticsSummarySizeBytes` (e.g., to 16384) will only see their custom value take effect for change feed estimator, LINQ aggregation, ReadMany, and throughput operations — not for standard ReadItem/CreateItem/etc.

**Suggestion:** Consider storing `maxDiagnosticsSummarySizeBytes` on the `ResponseMessage` and threading it through, or document this as a known v1 limitation. The PR description notes ''at call sites with ClientContext access'' which is accurate.

---

Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs
```diff
@@ -0,0 +1,465 @@
...
+            }
+
+            // For odd count, take the middle element.
+            // For even count, take the lower of the two middle elements
+            // (matching the Rust SDK's floor-based approach).
+            int midIndex = (count - 1) / 2;
+            return sortedValues[midIndex];
+        }
+
+        private static string BuildTruncatedJson(ITrace trace, int totalRequestCount)
```
NaluTripician: **🟡 Recommendation** · Completeness: Missing Field in Truncated Output

**Truncated summary omits `TotalRequestCharge`**

The full summary JSON includes `TotalRequestCharge`, but `BuildTruncatedJson` does not. At the point this method is called (line 50 in `WriteSummary`), the `entries` list has already been collected and the total charge was already computed in `BuildSummaryJson`.

**Why it matters:** `TotalRequestCharge` is the most operationally useful field — it tells operators the total RU cost. When the summary is truncated (meaning the request was complex enough to exceed the size limit), this is exactly when operators need the total RU cost the most.

**Suggestion:** Compute `totalRequestCharge` in `WriteSummary` before calling `BuildSummaryJson`, then pass it to `BuildTruncatedJson`. The change is minimal:`n`csharp
private static string BuildTruncatedJson(ITrace trace, int totalRequestCount, double totalRequestCharge)
`$footer

Microsoft.Azure.Cosmos.Encryption/src/EncryptionCosmosDiagnostics.cs
```diff
@@ -88,6 +88,43 @@ public override string ToString()
         }
 
 #if SDKPROJECTREF
+        public override string ToString(DiagnosticsVerbosity verbosity)
```
NaluTripician, line 98: **🟡 Recommendation** · Performance: Missing Cache

**`ToString(DiagnosticsVerbosity)` does not cache, unlike existing `ToString()`**

The existing parameterless `ToString()` in this class caches its result in `this.cachedDiagnostics`. However, this new `ToString(DiagnosticsVerbosity)` override rebuilds the JSON wrapper (StringBuilder, JsonTextWriter, encryption diagnostics serialization) on **every call**. While the inner `coreDiagnostics.ToString(verbosity)` is cached via `Lazy<string>`, the encryption wrapper is not.

**Why it matters:** `ToString()` may be called multiple times per response (logging, telemetry, error handlers). The inconsistency with the existing caching pattern is also a maintenance concern — developers may assume all `ToString` variants are cached.

**Suggestion:** Add a `Lazy<string>` or null-check cache for the summary path, matching the existing `cachedDiagnostics` pattern.

---

Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs
```diff
@@ -0,0 +1,465 @@
...
+            if (trace is Tracing.Trace concreteTrace)
+            {
+                concreteTrace.SetWalkingStateRecursively();
+            }
+
+            List<RequestEntry> entries = CollectRequestEntries(trace);
+
+            string summaryJson = BuildSummaryJson(trace, entries);
+
+            if (Encoding.UTF8.GetByteCount(summaryJson) <= maxSizeBytes)
```
NaluTripician, line 61: **🟡 Recommendation** · Performance: Wasteful Allocation

**Full summary JSON always computed before size check**

The `WriteSummary` method always computes the complete summary JSON via `BuildSummaryJson`, then checks `Encoding.UTF8.GetByteCount(summaryJson) <= maxSizeBytes`. If it exceeds the limit, it discards the full summary and builds a truncated version instead.

For pathological scenarios (the exact use case this feature targets — hundreds of retries across multiple regions), the full summary could be several KB. The method allocates this string, measures it, discards it, and builds the smaller truncated version.

**Suggestion:** Consider estimating output size before full serialization (e.g., `entries.Count * ~200` bytes as a rough estimate per entry), or use a streaming approach where `IJsonWriter` can report bytes written and bail early. Acceptable for v1 if documented as a future optimization.

---

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -48,6 +63,16 @@ public override string ToString()
             return this.ToJsonString();
         }
 
+        public override string ToString(DiagnosticsVerbosity verbosity)
+        {
+            if (verbosity == DiagnosticsVerbosity.Summary)
```
NaluTripician: **🟡 Recommendation** · Maintainability: Future-Proofing

**`if (Summary) else fallthrough` pattern won't catch future enum values**

This uses `if (verbosity == DiagnosticsVerbosity.Summary)` and falls through to `ToString()` (Detailed) for any other value. If a third verbosity level is added later (e.g., `Minimal`), it would silently get `Detailed` output with no compile-time or runtime warning.

**Suggestion:** Consider a `switch` with an explicit `default` case:`n`csharp
return verbosity switch
{
    DiagnosticsVerbosity.Summary => this.cachedSummaryJson.Value,
    DiagnosticsVerbosity.Detailed => this.ToString(),
    _ => this.ToString(), // Fallback to detailed for unknown values
};
`$footer

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
```diff
@@ -125,6 +125,18 @@ internal static class ConfigurationManager
         /// </summary>
         internal static readonly string UseLengthAwareRangeComparator = "AZURE_COSMOS_USE_LENGTH_AWARE_RANGE_COMPARATOR";
 
+        /// <summary>
+        /// Environment variable name for overriding the default diagnostics verbosity.
+        /// Accepts "Summary" or "Detailed" (case-insensitive).
+        /// </summary>
+        internal static readonly string DiagnosticsVerbosityVariable = "AZURE_COSMOS_DIAGNOSTICS_VERBOSITY";
```
NaluTripician, line 132: **🟢 Suggestion** · Maintainability: Consistency

**Env var constants defined but tests use string literals**

These constants (`DiagnosticsVerbosityVariable`, `DiagnosticsMaxSummarySizeVariable`) are well-defined here, but the unit tests in `CosmosClientOptionsUnitTests.cs` reference the env var names as string literals (e.g., `"AZURE_COSMOS_DIAGNOSTICS_VERBOSITY"`). If the env var name changes, the constant would be updated but tests would silently test the wrong variable.

**Suggestion:** Reference `ConfigurationManager.DiagnosticsVerbosityVariable` in tests instead of duplicating the string literal.

---

Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs
```diff
@@ -0,0 +1,465 @@
...
+        }
+
+        private static List<RequestEntry> CollectRequestEntries(ITrace trace)
+        {
+            List<RequestEntry> entries = new List<RequestEntry>();
+            CollectRequestEntriesRecursive(trace, entries);
+            return entries;
+        }
+
+        private static void CollectRequestEntriesRecursive(ITrace currentTrace, List<RequestEntry> entries)
```
NaluTripician: **🟢 Suggestion** · Reliability: Defensive Programming

**Recursive trace tree walk has no cycle guard**

`CollectRequestEntriesRecursive` walks `currentTrace.Children` recursively. If the `ITrace` tree ever had a cycle (child pointing back to a parent), this would stack overflow. The existing `TraceJsonWriter` likely has the same limitation, so this is pre-existing behavior.

**Suggestion:** Low priority — cycles are unlikely in practice. But since this is new code, adding a `HashSet<ITrace> visited` parameter would be a defensive improvement.

---

Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs
```diff
@@ -0,0 +1,465 @@
...
+                    && int.TryParse(first, NumberStyles.Integer, CultureInfo.InvariantCulture, out int sub))
+                {
+                    return sub;
+                }
+            }
+
+            return 0;
+        }
+
+        private static string BuildSummaryJson(ITrace trace, List<RequestEntry> entries)
```
NaluTripician: **🟢 Suggestion** · Observability: Missing Correlation ID

**Summary output doesn't include `ActivityId`**

The summary includes region, status codes, latency, and RU charges — but no `ActivityId`. This is essential for correlating client-side diagnostics with server-side logs during support investigations. The detailed output includes it via the full trace tree.

**Suggestion:** Add the `ActivityId` at the top level of the summary JSON (before `RegionsSummary`). It's a single string field and critical for production troubleshooting in summary mode.

---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Diagnostics/DiagnosticsSummaryWriterTests.cs
```diff
@@ -0,0 +1,578 @@
...
+            // Call ToString(Summary) multiple times — should return same cached instance
+            string summary1 = diagnostics.ToString(DiagnosticsVerbosity.Summary);
+            string summary2 = diagnostics.ToString(DiagnosticsVerbosity.Summary);
+
+            Assert.AreSame(summary1, summary2, "Summary should be cached via Lazy<string>");
+        }
+
+        #region Helpers
+
+        private static void AddStoreResponseStatistic(
```
NaluTripician, line 473: **🟢 Suggestion** · Test Coverage: Missing Path

**No test for `HttpResponseStatistics` (Gateway mode) collection path**

All tests use `AddStoreResponseStatistic` which exercises the Direct mode (`StoreResponseStatistics`) path. The `CollectRequestEntriesRecursive` method also handles `HttpResponseStatistics` (Gateway mode) with different header parsing logic, including `GetHttpSubStatusCode` and request charge extraction from HTTP headers.

**Suggestion:** Add a helper like `AddHttpResponseStatistic` and a test that exercises Gateway mode entry collection, sub-status code extraction, and request charge parsing from HTTP headers.

---

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -34,6 +40,15 @@ public CosmosTraceDiagnostics(ITrace trace)
 
             this.Value = rootTrace;
             this.accumulatedMetrics = new Lazy<ServerSideCumulativeMetrics>(() => PopulateServerSideCumulativeMetrics(this.Value));
+            this.cachedSummaryJson = new Lazy<string>(() =>
```
NaluTripician, line 43: **💬 Observation** · Design: Caching Strategy

**`Lazy<string>` caching means summary is immutable after first access**

The `Lazy<string>` computes the summary once and caches it forever. If the trace tree were modified after the first `ToString(Summary)` call, the cached summary would be stale. Since `CosmosDiagnostics` is created at response completion time (trace is finalized), this is safe.

Just noting for future reference — if diagnostics ever become mutable/appendable, this caching strategy would need revisiting.

---

openspec/changes/diagnostics-compaction/design.md
```diff
@@ -0,0 +1,105 @@
...
+| `Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs` | Summary computation and JSON serialization logic |
+
+## Files to Modify
+
+| File | Change |
+|------|--------|
+| `CosmosClientOptions.cs` | Add `DiagnosticsVerbosity` and `MaxDiagnosticsSummarySizeBytes` properties with validation |
+| `CosmosDiagnostics.cs` | Add `ToString(DiagnosticsVerbosity)` abstract overload |
+| `CosmosTraceDiagnostics.cs` | Implement `ToString(DiagnosticsVerbosity)` overload; delegate to `DiagnosticsSummaryWriter` when verbosity is `Summary` |
+| `TraceWriter.TraceJsonWriter.cs` | Add summary serialization path that delegates to `DiagnosticsSummaryWriter` when verbosity is `Summary` |
```
NaluTripician: **💬 Observation** · Documentation: Spec Drift

**Design spec lists files not modified by this PR**

This spec lists modifications to `TraceWriter.TraceJsonWriter.cs`, `SummaryDiagnostics.cs`, and `ClientSideRequestStatisticsTraceDatum.cs` — none of which are actually modified in the implementation. The actual approach (standalone `DiagnosticsSummaryWriter`) is arguably cleaner than what the spec proposed.

**Suggestion:** Update the spec to reflect the actual implementation to avoid confusion for future readers.

---

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
NaluTripician: Lets keep this as an internal value and only store it if its explicitly stated. We can also explore if there is a better alternative than storing it per response.

Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs
NaluTripician: lets document this.

Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs
NaluTripician: add this.

Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs
NaluTripician: add this.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Diagnostics/DiagnosticsSummaryWriterTests.cs
NaluTripician: add this

openspec/changes/diagnostics-compaction/design.md
NaluTripician: update spec

Microsoft.Azure.Cosmos.Encryption/src/EncryptionCosmosDiagnostics.cs
```diff
@@ -17,6 +17,9 @@ internal sealed class EncryptionCosmosDiagnostics : CosmosDiagnostics
         private readonly JObject encryptContent;
         private readonly JObject decryptContent;
         private readonly TimeSpan processingDuration;
+#if SDKPROJECTREF
+        private string cachedSummaryDiagnostics;
```
NaluTripician: 🔴 **High — Thread Safety: Race condition in caching**

This manual check-and-set pattern is not thread-safe:
```csharp
if (this.cachedSummaryDiagnostics != null) return ...;
// ...build...
this.cachedSummaryDiagnostics = stringWriter.ToString();
```

Multiple concurrent callers can all see `null`, compute independently, and overwrite each other. `CosmosTraceDiagnostics` correctly uses `Lazy<string>` for the same purpose.

**Suggestion:** Use `Lazy<string>` initialized in the constructor (matching `CosmosTraceDiagnostics`), or use `Interlocked.CompareExchange` for lock-free caching.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -17,8 +17,14 @@ namespace Microsoft.Azure.Cosmos.Diagnostics
     internal sealed class CosmosTraceDiagnostics : CosmosDiagnostics
     {
         private readonly Lazy<ServerSideCumulativeMetrics> accumulatedMetrics;
+        private readonly Lazy<string> cachedSummaryJson;
 
         public CosmosTraceDiagnostics(ITrace trace)
+            : this(trace, CosmosClientOptions.DefaultMaxDiagnosticsSummarySizeBytes)
+        {
+        }
```
NaluTripician, line 25: 🟡 **Medium — Completeness: Core CRUD paths use default max size**

The parameterless constructor defaults to `DefaultMaxDiagnosticsSummarySizeBytes` (8192). Several core code paths use this constructor:
- `ResponseMessage.Diagnostics` getter (lazy creation)
- `TransactionalBatchResponse` constructor
- `CosmosException` constructor

This means **standard CRUD operations** (ReadItem, CreateItem, etc.) won't respect a user-configured `MaxDiagnosticsSummarySizeBytes` — only ChangeFeed, LINQ, ReadMany, and throughput operations do.

**Suggestion:** At minimum, document which code paths respect the custom value. Ideally, thread the configured value through `ResponseMessage` during pipeline construction.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -384,7 +405,57 @@ public ConnectionMode ConnectionMode
...
+        /// <remarks>
+        /// This property is only relevant when <see cref="DiagnosticsVerbosity"/> is set to
+        /// <see cref="Microsoft.Azure.Cosmos.DiagnosticsVerbosity.Summary"/>.
+        /// </remarks>
+        public int MaxDiagnosticsSummarySizeBytes
+        {
+            get => this.maxDiagnosticsSummarySizeBytes;
+            set
+            {
+                if (value < 4096)
```
NaluTripician: 🟡 **Medium — Validation: No upper bound on MaxDiagnosticsSummarySizeBytes**

The setter validates `>= 4096` but has no upper bound. Setting `int.MaxValue` (~2 GB) via the property or environment variable would be silently accepted, potentially causing huge memory allocations.

**Suggestion:** Add a reasonable upper bound (e.g., 10 MB) with `ArgumentOutOfRangeException`:
```csharp
if (value < 4096 || value > 10_485_760)
{
    throw new ArgumentOutOfRangeException(...);
}
```
Apply the same check in the env var parsing path.

Microsoft.Azure.Cosmos/src/Diagnostics/DiagnosticsSummaryWriter.cs
```diff
@@ -0,0 +1,522 @@
...
+                if (first != null
+                    && int.TryParse(first, NumberStyles.Integer, CultureInfo.InvariantCulture, out int sub))
+                {
+                    return sub;
+                }
+            }
+
+            return 0;
+        }
+
```
NaluTripician, line 170: 🟡 **Medium — API Evolution: No version field in Summary JSON**

The Summary JSON is shipping as GA with no format version indicator. Once customers build log parsers against this schema, structural changes become breaking.

**Suggestion:** Add a version field early in the JSON output:
```csharp
writer.WriteFieldName("SummaryFormatVersion");
writer.WriteNumberValue(1);
```
This is cheap insurance that enables non-breaking format evolution. If a future version restructures the JSON, consumers can branch on the version field.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -89,8 +95,23 @@ public CosmosClientOptions()
...
+            this.SessionRetryOptions = new SessionRetryOptions();
+
+            string envVerbosity = Environment.GetEnvironmentVariable(ConfigurationManager.DiagnosticsVerbosityVariable);
+            if (!string.IsNullOrEmpty(envVerbosity)
+                && Enum.TryParse(envVerbosity, ignoreCase: true, out DiagnosticsVerbosity parsedVerbosity))
+            {
+                this.DiagnosticsVerbosity = parsedVerbosity;
+            }
+
+            string envMaxSize = Environment.GetEnvironmentVariable(ConfigurationManager.DiagnosticsMaxSummarySizeVariable);
```
NaluTripician, line 121: 🟢 **Low — Observability: Silent env var validation failures**

When `AZURE_COSMOS_DIAGNOSTICS_MAX_SUMMARY_SIZE` is set to an invalid value (e.g., `"abc"` or `"100"`), it is silently ignored with no feedback. Same for `AZURE_COSMOS_DIAGNOSTICS_VERBOSITY` with invalid enum names. Operators debugging misconfiguration would have no signal.

**Suggestion:** Consider emitting a `DefaultTrace.TraceWarning` when an env var is present but fails validation:
```csharp
else if (!string.IsNullOrEmpty(envMaxSize))
{
    DefaultTrace.TraceWarning("Invalid AZURE_COSMOS_DIAGNOSTICS_MAX_SUMMARY_SIZE: {0}", envMaxSize);
}
```

NaluTripician (commented):
## PR Review: Diagnostics Summary Mode

**Overall:** Well-designed feature with solid architecture. The standalone `DiagnosticsSummaryWriter` cleanly separates compaction from the trace tree, backward compatibility is properly preserved, and test coverage is substantial (56 tests). Several issues from a prior review round have been addressed.

### Remaining Findings Summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🔴 High | Thread safety in `EncryptionCosmosDiagnostics` caching (see inline) |
| 2 | 🟡 Medium | Missing `MaxDiagnosticsSummarySizeBytes` wiring in core CRUD paths (see inline) |
| 3 | 🟡 Medium | No upper bound on `MaxDiagnosticsSummarySizeBytes` (see inline) |
| 4 | 🟡 Medium | No version field in Summary JSON format (see inline) |
| 5 | 🟡 Medium | Test coverage gaps (see below) |
| 6 | 🟢 Low | Environment variable validation errors are silent (see inline) |

### Test Coverage Gaps (Finding #5)

The following scenarios lack test coverage:

1. **Concurrent access** — All tests are single-threaded. The `Lazy<string>` caching is never stress-tested with parallel callers. A `Parallel.For` test verifying `ReferenceEquals` across threads would validate the thread-safety guarantee.

2. **Size boundary conditions** — The truncation test uses 512 bytes with 200 entries (massive overshoot). No test validates the exact boundary: output at exactly `maxSizeBytes` (should pass) vs. `maxSizeBytes + 1` (should truncate).

3. **`ActivityId` extraction paths** — `FindActivityId` was added per prior feedback, but no unit test creates a trace with `PointOperationStatisticsTraceDatum` to verify ActivityId appears in output or is correctly omitted when absent.

4. **Multi-byte UTF-8 region names** — Size enforcement uses `Encoding.UTF8.GetByteCount` (correct), but no test validates that non-ASCII region names are counted by byte size, not char count.

### Items Already Addressed from Prior Review ✅
Redundant `SetWalkingStateRecursively` in Lazy lambda, `TotalRequestCharge` in truncated output, cycle guard (`HashSet<ITrace>`), `ActivityId` in summary, switch expression for future enum values, `HttpResponseStatistics` test coverage, and spec alignment.

NaluTripician:
Closing PR until rust work is done

PR #5705, opened 2026-03-19
[Internal] DTS: Adds logic to merge session data from DTC response into SessionContainer

Meghana-Palaparthi:
# Pull Request Template

## Description

After a CommitDistributedTransaction, per-operation session tokens in the JSON response body were never pushed into the  SDK's SessionContainer. Subsequent reads with Session consistency on the affected collections could therefore fail with ReadSessionNotAvailable (404/1002).

Changes:
- `DistributedTransactionCommitter`: add MergeSessionTokens() private helper, called after FromResponseMessageAsync returns.  Iterates (result[i], operation[i]) pairs, builds a minimal header collection carrying the per-operation session token, and calls `sessionContainer.SetSessionToken(collectionRid, collectionFullname, headers)`  mirroring the pattern used by `GatewayStoreModel.CaptureSessionTokenAndHandleSplitAsync()` for TransactionalBatch.
- New unit tests in `DistributedTransactionCommitterTests`
- Merge session data from DTC response into `SessionContainer`

## Type of change

## Closing issues

### DIFF
--- a/.github/agents/issue-fix-agent.agent.md
+++ b/.github/agents/issue-fix-agent.agent.md
@@ -163,6 +163,11 @@ github_access:
         - "github-mcp-server-issue_read"
         - "github-mcp-server-pull_request_read"
         - "All github-mcp-server-* tools for Azure org"
+      important: |
+        Do NOT reference github-mcp-server-* tools anywhere in your workflow
+        for Azure org repos. Always use gh CLI equivalents instead.
+        This includes: list_issues, issue_read, pull_request_read,
+        get_review_comments, search_code, search_issues, etc.
         
   recommended_approach:
     primary: "Use gh CLI for all GitHub operations"
@@ -685,6 +690,38 @@ troubleshooting:
     fix: "First MCP call opens browser - complete Microsoft login"
 ```
 
+### 0.8 Whitespace & Line Endings
+
+```yaml
+line_endings:
+  git_config:
+    # Repository uses core.autocrlf=input (LF in repo, no conversion on checkout)
+    # Verify with: git config core.autocrlf
+    expected: "input"
+    
+  common_pitfall: |
+    When editing a file, some editors normalize all line endings to LF or CRLF.
+    If the original file has mixed line endings (some CRLF, some LF), your edit
+    may produce a diff with spurious whitespace-only changes that reviewers will flag.
+    
+  prevention:
+    before_editing:
+      - "Check original line endings: git show master:{file} | python -c \"import sys; d=sys.stdin.buffer.read(); print(f'CRLF:{d.count(chr(13).encode()+chr(10).encode())} LF:{d.count(chr(10).encode())}')\""
+      
+    after_editing:
+      - "Verify no whitespace-only changes: git diff master --ignore-all-space -- {file}"
+      - "If whitespace diff appears, restore from master and re-apply only intentional changes"
+      
+    restore_pattern: |
+      # If you accidentally introduced whitespace changes:
+      git checkout master -- {file}          # Restore original
+      # Then re-apply only your intentional additions using edit tool
+      
+  diff_verification:
+    always_run: "git diff master --ignore-cr-at-eol -- {file}"
+    if_spurious_changes: "git checkout master -- {file} and re-apply edits"
+```
+
 ---
 
 ## 1. Executive Summary
@@ -796,6 +833,7 @@ mandatory_checkpoints:
     - "Search for existing PRs: `gh pr list --search 'ISSUE_NUMBER'`"
     - "Search for existing branches: `git branch -a | grep ISSUE_NUMBER`"
     - "If PR exists, use existing branch instead of creating new one"
+    - "If existing PR is stale (no activity >30 days), comment on it and create new branch"
     - "Document search results as evidence"
     
   before_creating_pr:
@@ -803,12 +841,16 @@ mandatory_checkpoints:
     - "Local tests MUST pass with captured output"
     - "Root cause MUST be documented with code path"
     - "Fix MUST be verified with before/after evidence"
+    - "If before/after evidence is not possible (environment-specific issue), document why and provide unit test evidence instead"
     - "PR MUST be created as DRAFT: `gh pr create --draft`"
+    - "PR title MUST follow format: `[Internal] Category: (Adds|Fixes|Refactors|Removes) Description` (enforced by PR Lint bot)"
+    - "Verify no spurious whitespace changes: `git diff master --ignore-all-space`"
     
   before_marking_pr_ready:
-    - "ALL CI gates MUST show COMPLETED status"
+    - "ALL CI gates MUST show green (passed) status — not just completed"
     - "PR description MUST follow full template"
     - "Issue comment MUST be posted with investigation summary"
+    - "All reviewer comments MUST be addressed (see Section 7.4.1)"
     - "Convert from draft: `gh pr ready <PR_NUMBER>`"
     
   before_claiming_issue_resolved:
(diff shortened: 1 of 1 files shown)


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -93,10 +93,11 @@ protected DistributedTransactionOperationResult()
 
         /// <summary>
         /// Used for JSON deserialization of the base64-encoded resource body.
+        /// System.Text.Json requires [JsonInclude] members to be public in reflection mode.
         /// </summary>
         [JsonInclude]
         [JsonPropertyName("resourcebody")]
-        internal string ResourceBodyBase64
+        public string ResourceBodyBase64
```
kirankumarkolli: Is below base64 logic still needed?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -100,6 +107,47 @@ private static void EnrichRequestMessage(RequestMessage requestMessage, Distribu
...
+            if (response == null || response.Count == 0)
+            {
+                return;
+            }
+
+            RequestNameValueCollection headers = new RequestNameValueCollection();
+
+            for (int i = 0; i < response.Count && i < serverRequest.Operations.Count; i++)
+            {
+                DistributedTransactionOperationResult result = response[i];
```
kirankumarkolli, line 135: Exclude sessionCapture for 404/1002

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -100,6 +107,47 @@ private static void EnrichRequestMessage(RequestMessage requestMessage, Distribu
...
+
+                if (string.IsNullOrEmpty(result.SessionToken) || string.IsNullOrEmpty(operation.CollectionResourceId))
+                {
+                    continue;
+                }
+
+                headers.Clear();
+                headers[HttpConstants.HttpHeaders.SessionToken] = result.SessionToken;
+
+                sessionContainer.SetSessionToken(
```
kirankumarkolli, line 155: Optimization: Each operation inside does a locking, wondering if a new API for batch updates for a collection will help. 

/cc: @FabianMeiswinkel

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -75,13 +77,27 @@ private async Task<DistributedTransactionResponse> ExecuteCommitAsync(
...
+                    try
+                    {
+                        DistributedTransactionCommitter.MergeSessionTokens(
+                            response,
+                            serverRequest,
+                            this.clientContext.DocumentClient.sessionContainer);
+                    }
+                    catch (Exception ex)
+                    {
+                        DefaultTrace.TraceWarning($"DTC session token merge failed (non-fatal): {ex.Message}");
```
kirankumarkolli: What are the possible errors and is ignoring fine? 
Can you please validate similar flows

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -100,8 +116,76 @@ private static void EnrichRequestMessage(RequestMessage requestMessage, Distribu
...
+            // but we must explicitly push them into the SessionContainer.
+
+            if (response == null || response.Count == 0 || serverRequest == null || sessionContainer == null)
+            {
+                return;
+            }
+
+            if (response.Count != serverRequest.Operations.Count)
+            {
+                DefaultTrace.TraceWarning(
```
kirankumarkolli: TraceError?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -100,8 +116,76 @@ private static void EnrichRequestMessage(RequestMessage requestMessage, Distribu
...
+                int operationIndex = result.Index;
+                if (operationIndex < 0 || operationIndex >= serverRequest.Operations.Count)
+                {
+                    DefaultTrace.TraceWarning(
+                        $"DTC session token merge: result at position {i} has out-of-range Index={operationIndex} " +
+                        $"(operations count: {serverRequest.Operations.Count}); skipping.");
+                    continue;
+                }
+
+                DistributedTransactionOperation operation = serverRequest.Operations[operationIndex];
```
kirankumarkolli: In-case of sub-set return will it result in IndexOutOfBounds?

kirankumarkolli: Alternative is to iterate through every response if possible and consume?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
Meghana-Palaparthi: DTS commit has already succeeded on the server, so a failure here should not propagate back to the caller as a transaction error. That's why we are ignoring it here. If we want to propagate it, then we need to make it clear that the DTS commit has succeeded and there's an error during session token merge. 

Possible exceptions:
- `IndexOutOfRangeException` / `FormatException` from SessionContainer.SetSessionToken if the server returns a malformed session token.
- `FormatException` from `ResourceId.Parse` if the collection RID is malformed.

GatewayStoreModel.CaptureSessionTokenAndHandleSplitAsync does NOT wrap SetSessionToken in a try/catch, so an exception there would propagate even though the server write already succeeded.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
Meghana-Palaparthi: `IndexOutOfRangeException` is not possible in either the full-count or subset case. The loop already iterates on every response and consumes it.  We also have explicit bounds check to guard against any out-of-range Index value before the access.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
Meghana-Palaparthi: This is fixed in PR https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5686/

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
kirankumarkolli: That's a good call out and for now we can continue with current model and let the call stack covey the failure. Lets also please track this work item.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
kirankumarkolli: Can you please create a follow-up GitHub issue to track it.

Meghana-Palaparthi: Created a GitHub issue: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5729

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -100,8 +116,76 @@ private static void EnrichRequestMessage(RequestMessage requestMessage, Distribu
...
+            }
+
+            RequestNameValueCollection headers = new RequestNameValueCollection();
+
+            for (int i = 0; i < response.Count; i++)
+            {
+                DistributedTransactionOperationResult result = response[i];
+
+                int operationIndex = result.Index;
+                if (operationIndex < 0 || operationIndex >= serverRequest.Operations.Count)
```
kirankumarkolli: Its totally unexpected. 
Validations right place is upstrteam when response is composed, thougts?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -100,8 +116,76 @@ private static void EnrichRequestMessage(RequestMessage requestMessage, Distribu
             requestMessage.UseGatewayMode = true;
         }
 
+        internal static void MergeSessionTokens(
```
kirankumarkolli, line 112: Optimization: Follow-up limit session merge for only SESSION, MM

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
kirankumarkolli: Very little value with tracing as CX already sees the final failure. try-catch is not really needed.

kirankumarkolli (approved):

PR #5706, opened 2026-03-19
Performance: Cache URL-encoded AAD authorization signature

reedz:
## Description

The AAD token credential path calls HttpUtility.UrlEncode on the full authorization string (~1.5KB JWT) on every Cosmos DB request, even though the underlying AAD token is cached and only refreshes every ~30 minutes.

At high throughput (thousands of requests/sec), HttpUtility.IsSafe character-by-character scanning accounts for significant amount of CPU. Benchmarks show 8.9μs and 6KB allocation per call, while the cached path is 1.7ns with zero allocations.

This change caches the URL-encoded result alongside the token using Volatile reads/writes for lock-free thread safety. The static GenerateAadAuthorizationSignature method is preserved for backward compatibility.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Authorization/AuthorizationTokenProviderTokenCredential.cs
+++ b/Microsoft.Azure.Cosmos/src/Authorization/AuthorizationTokenProviderTokenCredential.cs
@@ -6,6 +6,7 @@ namespace Microsoft.Azure.Cosmos
 {
     using System;
     using System.Globalization;
+    using System.Threading;
     using System.Threading.Tasks;
     using global::Azure.Core;
     using Microsoft.Azure.Cosmos.Core.Trace;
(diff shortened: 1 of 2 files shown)


Microsoft.Azure.Cosmos/src/Authorization/AuthorizationTokenProviderTokenCredential.cs
```diff
@@ -108,6 +117,20 @@ public static string GenerateAadAuthorizationSignature(string aadToken)
                 aadToken));
         }
 
+        internal string GetOrCreateAadAuthorizationSignature(string aadToken)
+        {
+            string currentCachedToken = Volatile.Read(ref this.cachedAadToken);
```
kirankumarkolli, line 122: One alternative might be to push it inside the TokenCredentialCache it-self (Renaming it might help). 
It will help all consumers including inference and any future integrations.

PR #5718, opened 2026-03-24
Add equality logic to CosmosClientOptions and other option bags

### LINKED ISSUE #4965: Implement value semantics for CosmosClientOptions
When the options are provided by the user it would be useful to be able to compare them to each other to be able to return an existing CosmosClient instance if the options used to create it are equivalent.

Override `Equals` and `GetHashCode`. Using reference equality for options of reference type would be acceptable.

Currently EF accomplishes this by wrapping all the options: https://github.com/dotnet/efcore/blob/38dc78fcc3bfff413cb9ad8a7080dc13a1ca8ee4/src/EFCore.Cosmos/Infrastructure/Internal/CosmosDbOptionExtension.cs#L642

roji:
Closes #4965

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.Cosmos
     /// ]]>
     /// </code>
     /// </example>
-    public class CosmosClientOptions
+    public class CosmosClientOptions : IEquatable<CosmosClientOptions>
     {
         /// <summary>
         /// Default connection mode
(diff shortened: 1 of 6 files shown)


Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -1072,6 +1072,158 @@ public IFaultInjector FaultInjector
...
+                && this.EnableContentResponseOnWrite == other.EnableContentResponseOnWrite
+                && this.idleTcpConnectionTimeout == other.idleTcpConnectionTimeout
+                && this.openTcpConnectionTimeout == other.openTcpConnectionTimeout
+                && this.maxRequestsPerTcpConnection == other.maxRequestsPerTcpConnection
+                && this.maxTcpConnectionsPerEndpoint == other.maxTcpConnectionsPerEndpoint
+                && this.portReuseMode == other.portReuseMode
+                && this.webProxy == other.webProxy
+                && this.httpClientFactory == other.httpClientFactory
+                && object.Equals(this.serializerOptions, other.serializerOptions)
+                && this.serializerInternal == other.serializerInternal
```
AndriySvyryd, line 1112: Is this something the user can set?

Issue #5684, opened 2026-03-05
Thin Client Integration] Separate thinclient logic from gateway logic in GatewayStoreModel.


aavasthy:
Need to separate the logic between thinclient and gateway in GatewayStoreModel. This will reduce thinclient flag checks everywhere in the code and will have different code flow between thinclient mode and gateway mode.

PR #5614, opened 2026-02-13
[ThinClient Integration]: Adds support for QueryPlan in thinclient mode.

aavasthy:
# Pull Request Template

## Description

Adds support for QueryPlan in thinclient mode. The change also now converts partition key ranges from PartitionKeyInternal format to EPK hex strings required by the RNTBD protocol. This enables complex queries (ORDER BY, GROUP BY, aggregates) to work in ThinClient mode by translating backend QueryPlan responses.

## Type of change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)

## Closing issues

To automatically close an issue: closes #IssueNumber

PR #5720, opened 2026-03-25
[Internal] Thin Client Integration: Adds changefeed support in thinclient mode

aavasthy:
- [x] Understand the PR review comments
- [x] Fix fragile negation logic in `IsLatestVersionChangeFeedRequest` (use positive matching for `IncrementalFeed` instead of negation of `FullFidelityFeed`)
- [x] Update XML doc comment to accurately describe the method behavior (fail-safe routing for future A-IM values)
- [x] Add test for unknown A-IM header values (should NOT be routed to ThinClient)
- [x] Change `OperationType.Invalid` to `OperationType.ReadFeed` (no A-IM header) in unsupported operations test for a realistic scenario
- [x] Source project builds successfully, changes validated

---

💬 Send tasks to Copilot coding agent from [Slack](https://gh.io/cca-slack-docs) and [Teams](https://gh.io/cca-teams-docs) to turn conversations into code. Copilot posts an update in your thread when it's finished.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
+++ b/Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
@@ -590,15 +590,16 @@ namespace Microsoft.Azure.Cosmos
         }
 
         /// <summary>
-        /// Determines if the request is a LatestVersion (Incremental) change feed request.
-        /// A change feed request is identified by the presence of the A-IM header.
-        /// LatestVersion uses "Incremental feed" while AllVersionsAndDeletes uses "Full-Fidelity Feed".
+        /// Determines if the request is a LatestVersion (Incremental) change feed request that can
+        /// be routed to the thin client. Returns true only when the A-IM header is exactly
+        /// <c>HttpConstants.A_IMHeaderValues.IncrementalFeed</c>. Any other value — including
+        /// Full-Fidelity Feed (AllVersionsAndDeletes) or an unknown future mode — falls back to
+        /// Compute Gateway so that new modes are not accidentally routed to the thin client.
         /// </summary>
         internal static bool IsLatestVersionChangeFeedRequest(DocumentServiceRequest request)
         {
             string aImHeaderValue = request.Headers[HttpConstants.HttpHeaders.A_IM];
-            return !string.IsNullOrEmpty(aImHeaderValue)
-                && !string.Equals(aImHeaderValue, HttpConstants.A_IMHeaderValues.FullFidelityFeed, StringComparison.OrdinalIgnoreCase);
+            return string.Equals(aImHeaderValue, HttpConstants.A_IMHeaderValues.IncrementalFeed, StringComparison.OrdinalIgnoreCase);
         }
         private async Task<AccountProperties> GetDatabaseAccountPropertiesAsync()
         {
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
@@ -1332,8 +1332,10 @@ namespace Microsoft.Azure.Cosmos
                 It.IsAny<DocumentServiceRequest>()))
                 .ReturnsAsync(successResponse);
 
+            // Plain ReadFeed without an A-IM header is a realistic unsupported operation
+            // (document enumeration, not a change feed).
             DocumentServiceRequest request = DocumentServiceRequest.Create(
-                operationType: OperationType.Invalid,
+                operationType: OperationType.ReadFeed,
                 resourceType: ResourceType.Document,
                 resourceId: "NH1uAJ6ANm0=",
                 body: null,
@@ -1589,6 +1591,17 @@ namespace Microsoft.Azure.Cosmos
             nonDocumentRequest.Headers[HttpConstants.HttpHeaders.A_IM] = HttpConstants.A_IMHeaderValues.IncrementalFeed;
             Assert.IsFalse(GatewayStoreModel.IsOperationSupportedByThinClient(nonDocumentRequest),
                 "ReadFeed on non-Document resource should NOT be supported by ThinClient");
+
+            // Unknown A-IM header value should NOT be supported (fail-safe for future feed modes)
+            DocumentServiceRequest unknownAImRequest = DocumentServiceRequest.Create(
+               operationType: OperationType.ReadFeed,
+               resourceType: ResourceType.Document,
+               resourceId: "NH1uAJ6ANm0=",
+               body: null,
+               authorizationTokenType: AuthorizationTokenType.PrimaryMasterKey);
+            unknownAImRequest.Headers[HttpConstants.HttpHeaders.A_IM] = "Unknown-Feed";
+            Assert.IsFalse(GatewayStoreModel.IsOperationSupportedByThinClient(unknownAImRequest),
+                "Unknown A-IM values should NOT be supported by ThinClient");
         }
 
         [TestMethod]


Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -552,31 +552,53 @@ internal static bool IsStoredProcedureCrudOperation(
...
+
+            return false;
+        }
+
+        /// <summary>
+        /// Determines if the request is a LatestVersion (Incremental) change feed request.
+        /// A change feed request is identified by the presence of the A-IM header.
+        /// LatestVersion uses "Incremental feed" while AllVersionsAndDeletes uses "Full-Fidelity Feed".
+        /// </summary>
+        internal static bool IsLatestVersionChangeFeedRequest(DocumentServiceRequest request)
```
NaluTripician, line 599: **🟡 Recommendation** · Correctness: Fragile Negation Logic

**Prefer positive matching for `IncrementalFeed` over negation of `FullFidelityFeed`**

This method checks `!string.Equals(aImHeaderValue, FullFidelityFeed)` instead of positively matching `string.Equals(aImHeaderValue, IncrementalFeed)`. While the doc comment says *'LatestVersion uses Incremental feed'*, the code accepts **any** A-IM value that isn't FullFidelityFeed.

**Why it matters**: If a third change feed mode is added with a new A-IM value (e.g., `Streaming-Feed`), it would be automatically routed to thin client without anyone explicitly validating thin client support. The failure mode could be errors or incorrect results. Using the positive check makes new modes default to gateway (the safe path) until explicitly added.

Currently there are only two A-IM values (`IncrementalFeed` and `FullFidelityFeed` per `ChangeFeedModeIncremental` and `ChangeFeedModeFullFidelity`), so this is correct today — but fragile for tomorrow.

Suggested change:
```csharp
return string.Equals(aImHeaderValue, HttpConstants.A_IMHeaderValues.IncrementalFeed, StringComparison.OrdinalIgnoreCase);
```

---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
```diff
@@ -1384,15 +1384,216 @@ public async Task ThinClient_ProcessMessageAsync_WithUnsupportedOperations_Shoul
...
+
+            // Assert
+            Assert.IsNotNull(response);
+            Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+            Assert.IsFalse(thinClientInvoked, "AllVersionsAndDeletes change feed should NOT be routed to ThinClient");
+        }
+
+        [TestMethod]
+        [Owner("aavasthy")]
+        public void IsOperationSupportedByThinClient_WithChangeFeedModes_ReturnsCorrectResults()
```
NaluTripician, line 1550: **🟡 Recommendation** · Testing: Missing Test for Unknown A-IM Values

**Add a test case for an unexpected A-IM header value**

This test covers IncrementalFeed (supported), FullFidelityFeed (not supported), no A-IM header (not supported), and non-Document ReadFeed (not supported). But there's no test for an **unknown** A-IM value like `Unknown-Feed`.

**Why it matters**: With the current negation logic in `IsLatestVersionChangeFeedRequest`, an unknown A-IM value would be routed to thin client (likely unintended). Adding a test for this case would:
- Document the current behavior explicitly
- Catch regressions if the logic is changed to positive matching (Finding #1)
- Surface the question of whether unknown values *should* go to thin client

Suggested test case:
```csharp
// Unknown A-IM header value should NOT be supported
DocumentServiceRequest unknownRequest = DocumentServiceRequest.Create(
   operationType: OperationType.ReadFeed,
   resourceType: ResourceType.Document,
   resourceId: "NH1uAJ6ANm0=",
   body: null,
   authorizationTokenType: AuthorizationTokenType.PrimaryMasterKey);
unknownRequest.Headers[HttpConstants.HttpHeaders.A_IM] = "Unknown-Feed";
Assert.IsFalse(GatewayStoreModel.IsOperationSupportedByThinClient(unknownRequest),
    "Unknown A-IM values should NOT be supported by ThinClient");
```

---

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -552,31 +552,53 @@ internal static bool IsStoredProcedureCrudOperation(
...
+            {
+                return true;
+            }
+
+            // LatestVersion (Incremental) ChangeFeed on documents.
+            // AllVersionsAndDeletes (FullFidelity) is excluded because it requires
+            // split-handling logic in Compute Gateway (UseGatewayMode is set by ChangeFeedModeFullFidelity).
+            if (request.ResourceType == ResourceType.Document
+                && request.OperationType == OperationType.ReadFeed
+                && GatewayStoreModel.IsLatestVersionChangeFeedRequest(request))
```
NaluTripician, line 577: **🟡 Recommendation** · Reliability: Defense-in-Depth with `UseGatewayMode`**

**Consider checking `request.UseGatewayMode` as an additional guard**

`ProcessMessageAsync` doesn't check `request.UseGatewayMode`. `ChangeFeedModeFullFidelity.Accept()` sets both `UseGatewayMode = true` AND the `A-IM: Full-Fidelity Feed` header, but only the A-IM header is checked here for thin client routing.

**Why it matters**: `UseGatewayMode = true` semantically means 'this request needs Compute Gateway's special handling (split logic).' If a future change sets `UseGatewayMode` without properly setting the A-IM header, the request could be misrouted to thin client. Adding a check like `&& !(request.UseGatewayMode == true)` would provide defense-in-depth.

This is a belt-and-suspenders improvement, not a current bug.

---

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -552,31 +552,53 @@ internal static bool IsStoredProcedureCrudOperation(
...
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Determines if the request is a LatestVersion (Incremental) change feed request.
+        /// A change feed request is identified by the presence of the A-IM header.
```
NaluTripician: **🟡 Recommendation** · Maintainability: Misleading XML Doc Comment

**Doc comment doesn't describe the actual logic**

The comment says *'A change feed request is identified by the presence of the A-IM header'* but the method also checks that the value is NOT `FullFidelityFeed`. A developer reading only the summary would expect a simple null-or-empty check.

**Why it matters**: When the routing logic is this critical (determining thin client vs gateway), the documentation should precisely describe what the method actually checks. This is especially important because the negation pattern is non-obvious.

Suggested update:
```xml
/// <summary>
/// Determines if the request is a LatestVersion (Incremental) change feed request.
/// Returns true when the A-IM header is present and its value is NOT
/// Full-Fidelity Feed (which requires Compute Gateway for split-handling).
/// </summary>
```

---

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -552,31 +552,53 @@ internal static bool IsStoredProcedureCrudOperation(
...
+                || request.OperationType == OperationType.Upsert
+                || request.OperationType == OperationType.Replace
+                || request.OperationType == OperationType.Delete
+                || request.OperationType == OperationType.Query
+                || request.OperationType == OperationType.QueryPlan))
+            {
+                return true;
+            }
+
+            // LatestVersion (Incremental) ChangeFeed on documents.
```
NaluTripician, line 572: **🟡 Recommendation** · Maintainability: Growing Thin Client Logic in GatewayStoreModel

**Consider referencing issue #5684 and designing for future extraction**

Issue [#5684](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5684) ('Separate thinclient logic from gateway logic in GatewayStoreModel') is still open. In PR #5614, reviewer @neildsh explicitly asked: *'Please consider removing the thin client logic from this class and creating a new type like ThinClientStoreModel.'* This PR adds more thin client-specific logic (`IsLatestVersionChangeFeedRequest`) to `GatewayStoreModel`.

**Why it matters**: Each new thin client feature adds more conditional branching, making it harder to reason about gateway-only vs thin-client code paths. Not blocking for this PR, but the debt is accumulating. Consider:
- Referencing #5684 in the PR description
- Ensuring the new method is easy to extract later (it already is, being static)

---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
```diff
@@ -1354,6 +1354,52 @@ public async Task TestThinClientQueryPlanMultiPartitionFanout()
             }
         }
 
+        [TestMethod]
+        [TestCategory("ThinClient")]
+        public async Task TestThinClientChangeFeedLatestVersionAsync()
```
NaluTripician, line 1359: **🟢 Suggestion** · Testing: Add Emulator Test for AllVersionsAndDeletes

**Consider adding a companion emulator test for FullFidelity change feed**

This test verifies LatestVersion change feed routes through thin client by checking `|F4` in diagnostics. A companion test verifying AllVersionsAndDeletes does **not** contain `|F4` would increase confidence that the header-based routing works end-to-end in the real stack.

Unit tests already cover the FullFidelity → gateway path, so this is a nice-to-have for integration confidence.

---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
```diff
@@ -1333,7 +1333,7 @@ public async Task ThinClient_ProcessMessageAsync_WithUnsupportedOperations_Shoul
                 .ReturnsAsync(successResponse);
 
             DocumentServiceRequest request = DocumentServiceRequest.Create(
-                operationType: OperationType.ReadFeed,
+                operationType: OperationType.Invalid,
```
NaluTripician: **🟢 Suggestion** · Testing: Use Realistic Operation Type in Unsupported Operations Test

**Consider using a realistic unsupported operation instead of `OperationType.Invalid`**

This was changed from `OperationType.ReadFeed` to `OperationType.Invalid` because ReadFeed is now conditionally supported. While functionally correct, `Invalid` is a sentinel value that would never appear in production.

Consider using `OperationType.ReadFeed` **without** an A-IM header (a plain document enumeration) — this better documents what a real unsupported ReadFeed looks like and exercises the new change feed routing logic more naturally.

---

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
NaluTripician, line 599: **🟢 Suggestion** · Naming: Method Name vs. Implementation Mismatch

**Method name implies positive check but implementation is negative**

`IsLatestVersionChangeFeedRequest` suggests a positive check for LatestVersion, but the implementation is a negative check (not FullFidelity). If Finding #1 is addressed (positive matching for `IncrementalFeed`), this mismatch resolves naturally. If the negation is kept intentionally, consider renaming to `IsNonFullFidelityChangeFeedRequest()` to match the actual semantics.

---

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -552,31 +552,53 @@ internal static bool IsStoredProcedureCrudOperation(
...
+        }
+
+        /// <summary>
+        /// Determines if the request is a LatestVersion (Incremental) change feed request.
+        /// A change feed request is identified by the presence of the A-IM header.
+        /// LatestVersion uses "Incremental feed" while AllVersionsAndDeletes uses "Full-Fidelity Feed".
+        /// </summary>
+        internal static bool IsLatestVersionChangeFeedRequest(DocumentServiceRequest request)
+        {
+            string aImHeaderValue = request.Headers[HttpConstants.HttpHeaders.A_IM];
```
NaluTripician, line 601: **🟢 Suggestion** · Design: Consider Structured Routing Over Header Inspection

**String header comparison for routing is fragile**

The thin client routing decision for change feed relies on parsing string header values. A more structured approach could be a boolean property on `DocumentServiceRequest` (e.g., `CanRouteThroughThinClient`) set by each `ChangeFeedMode.Accept()`. This would be explicit, type-safe, and avoid string comparison fragility.

This is a larger refactor and could be part of #5684 — noting it here as a design direction rather than a request for this PR.

---

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
Praveen-Msft: This is a good suggestion, Possibly IsFullFidelityChangeFeedRequest will be better given the functionality

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
aavasthy: @copilot we land on gatewaystoremodel class only if gatewaymode = true

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
aavasthy: Will address this as part of the other open issue  #5684

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
aavasthy: Will consider this as design change improvement when working on  https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5684

Praveen-Msft (approved):

kirankumarkolli (approved):

PR #5597, opened 2026-02-02
CosmosClient: Fixes ObjectDisposedException message when client is disposed during request

kirankumarkolli:
## Description

🤖 **This PR was authored by GitHub Copilot** as part of an automated issue triage workflow.

Fixes #5596

This PR fixes the confusing `InvalidOperationException: StoreProxy cannot be null` error that occurs when `CosmosClient.Dispose()` is called while requests are still in-flight.

---

## Investigation Steps Summary

1. **Reproduced the error path**: Traced `TransportHandler.cs:120` where error originates
2. **Identified all StoreModel assignments**: Found 3 locations in DocumentClient.cs
   - Line 1114: `= this.GatewayStoreModel` (Gateway mode init)
   - Line 6821: `= new ServerStoreModel(...)` (Direct mode init)  
   - Line 1362: `= null` (**Dispose() - ONLY place setting to null**)
3. **Verified GatewayStoreModel cannot be null at line 1114**: Assigned at line 1098, just 16 lines before
4. **Concluded root cause**: Race condition - Dispose() called while requests in-flight
5. **Analyzed PR #5197 (version 3.51)**: Confirmed it did NOT introduce the bug, only changed timing

---

## Root Cause

**Location**: `DocumentClient.cs:1362` (Dispose) and `DocumentClient.cs:6586` (GetStoreProxy)

**Analysis**:
When `CosmosClient.Dispose()` is called while requests are in-flight:
1. `Dispose()` sets `StoreModel = null` (DocumentClient.cs:1362)
2. In-flight request calls `GetStoreProxy()` which returns `null`
3. Caller throws unhelpful `InvalidOperationException("StoreProxy cannot be null")`

**Evidence**: `StoreModel` is set to `null` only in `Dispose()` - no other code path sets it to `null`.

---

### Before Fix (Confusing Error)

```mermaid
sequenceDiagram
    participant App
    participant CosmosClient
    participant TransportHandler
    participant DocumentClient
    
    App->>CosmosClient: ReadItemAsync()
    activate CosmosClient
    
    Note over App: Another thread calls Dispose()
    App->>CosmosClient: Dispose()
    CosmosClient->>DocumentClient: Dispose()
    DocumentClient->>DocumentClient: StoreModel = null
    
    CosmosClient->>TransportHandler: ProcessMessageAsync()
    TransportHandler->>DocumentClient: GetStoreProxy()
    DocumentClient-->>TransportHandler: null (disposed!)
    TransportHandler->>TransportHandler: Retry once
[...]

PR #5155, opened 2025-04-30
QuorumReader code not yielding on 429

dibahlfi:
When receiving 429s on a single region account with strong consistency, quorum reader code does not yield on QuorumRead and Barrier. 

[Single429Failure-BarrierCalls.json](https://github.com/user-attachments/files/20106664/Single429Failure-BarrierCalls.json)

[429WriteBarrierCallsWithoutFix.json](https://github.com/user-attachments/files/20093289/429WriteBarrierCallsWithoutFix.json)

[429WriteBarrierCallsWithFix.json](https://github.com/user-attachments/files/20093288/429WriteBarrierCallsWithFix.json)

// original issue diagnostics - QuorumRead
[429-OriginalIssueWithOutFix.json](https://github.com/user-attachments/files/19980803/429-OriginalIssueWithOutFix.json)

// verifying existing behavior by turning off exceptionless for 429
[429-OriginalIssueWithExceptionLessTurnedoOfFor429.json](https://github.com/user-attachments/files/19980788/429-OriginalIssueWithExceptionLessTurnedoOfFor429.json)

//fix Diagnostics with a large number of 429 failures- - QuorumRead
[429-WithFixAll429s.json](https://github.com/user-attachments/files/19980805/429-WithFixAll429s.json)

//fix Diagnostics with fewer 429 failures resulting in eventual success- - QuorumRead 
[429-Withfewer429sResultsInEventualSuccess.json](https://github.com/user-attachments/files/19980804/429-Withfewer429sResultsInEventualSuccess.json)

// original issue diagnostics - Barrier calls
[429 BarrierCallsOriginalIssueWithDiagnostics.json](https://github.com/user-attachments/files/20005545/429.BarrierCallsOriginalIssueWithDiagnostics.json)

//  Barrier calls fix with Diagnostics
[429-BarrierCallsFixWithDianogstics.json](https://github.com/user-attachments/files/20005546/429-BarrierCallsFixWithDianogstics.json)

Unit tests for QuorumReader is pending which I will submit along with the Direct code changes.
- [] Bug fix (non-breaking change which fixes an issue)

To automatically close an issue: closes #5035

PR #3870, opened 2023-05-23
CosmosClient: Fixes crashing CosmosClient on dispose stage

rinatmini:
# Pull Request Template

## Description

Fixes crashing CosmosClient.Dispose if there are any in-flight requests.
We already handle failing HTTPClient during disposing of the CosmosClient if there are any in-flight requests and this PR applies the same for disposing CosmosClient.IStoreClientFactory.
IStoreClientFactory internally closes Microsoft.Azure.Documents.Rntbd.LoadBalancingChannel which in the end also disposes of the own internal [ReaderWriterLockSlim ](https://learn.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim.dispose?view=net-6.0#system-threading-readerwriterlockslim-dispose) and throws SynchronizationLockException if there any of threads are waiting to enter the lock.

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)

## Closing issues

To automatically close an issue: closes #3700

PR #5722, opened 2026-03-26
[Internal] Direct package: Fixes thread pool starvation from blocking calls in RNTBD Dispatcher

### LINKED ISSUE #4393: [FEATURE REQ] Avoid blocking Asynchronous calls as it leads to thread starvation.
For Azure Cosmos DB SDK v3.36 on Linux, customer reported issue where the thread pool is rapidly exhausted, leading to the service becoming unresponsive. Investigation indicates that hundreds of threads are being blocked due to waits on tasks through Task.Wait method calls. This blocking behavior is causing threadpool starvation, severely impacting the performance.
Need to update SDK so it avoids blocking asynchronous calls.

Reference document:
https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/performance-tips-dotnet-sdk-v3?tabs=trace-net-core#logging-and-tracing

NaluTripician:
## Problem

Fixes #4393

On Linux, thread pool threads are blocked by synchronous `.Wait()` calls in the RNTBD `Dispatcher` class. When many connections go idle simultaneously, the `OnIdleTimer` callback runs on thread pool threads via `ContinueWith` and each one calls `WaitTask()` → `t.Wait()`, blocking the thread. This causes **thread pool starvation** and makes the service unresponsive.

### Two Blocking Paths

**Path 1: Idle Timer Callbacks (PRIMARY)**
```
TimerPool fires for N connections
  → ContinueWith(OnIdleTimer) × N thread pool threads
    → OnIdleTimer() → WaitTask(receiveTask) → t.Wait()
      → N threads BLOCKED simultaneously → STARVATION
```

**Path 2: Mass Channel Disposal (SECONDARY)**
```
ChannelDictionary.Dispose()
  → foreach channel: channel.Close()
    → Channel.Dispose() → initTask.Wait()        ← BLOCKS
      → Dispatcher.Dispose()
        → WaitTask(idleTimerTask)                 ← BLOCKS
        → WaitTask(receiveTask)                   ← BLOCKS
```

### `Dispatcher.cs`
- **`WaitTaskAsync`**: New async counterpart to `WaitTask` using `await` instead of `.Wait()`
- **`OnIdleTimerAsync`**: Converted from sync `OnIdleTimer` — **the critical fix** that eliminates thread pool starvation from idle timer callbacks
- **`ScheduleIdleTimer`**: Updated to use `.ContinueWith(OnIdleTimerAsync).Unwrap()` for proper async continuation tracking
- **`IAsyncDisposable` + `DisposeAsync`**: Non-blocking disposal path

### `IChannel.cs`
- Added `CloseAsync()` method to the interface

### `Channel.cs`
- **`IAsyncDisposable` + `DisposeAsync`**: Uses `await initTask` and `dispatcher.DisposeAsync()`
- **`CloseAsync`**: Delegates to `DisposeAsync`

### `LoadBalancingChannel.cs`
- **`IAsyncDisposable` + `DisposeAsync`**: Concurrent partition disposal via `Task.WhenAll`
- **`CloseAsync`**: Delegates to `DisposeAsync`

### `LoadBalancingPartition.cs`
- **`DisposeAsync`**: Concurrent channel state disposal via `Task.WhenAll`

### `LbChannelState.cs`
- **`DisposeAsync`**: Calls `channel.CloseAsync()` instead of `channel.Close()`

### `ChannelDictionary.cs`
- **`IAsyncDisposable` + `DisposeAsync`**: Concurrent channel closure via `Task.WhenAll`

## Design Decisions

- **Backward compatible**: All existing sync methods (`Dispose()`, `Close()`, `WaitTask()`) kept unchanged
- **Lock safety**: `lock` blocks remain synchronous; `await` is always outside lock scope
- **`.Unwrap()` is essential**: Ensures `idleTimerTask` properly represents the full async operation lifecycle for `StopIdleTimer()` cancellation tracking
- **Concurrent disposal**: `DisposeAsync` at each level uses `Task.WhenAll` for parallel channel/partition cleanup

## Testing

- ✅ Build: 0 errors, 0 warnings
- ✅ Existing RNTBD transport tests pass
- 8 pre-existing test failures (resource embedding issues on msdata/direct branch, unrelated to this change)

### DIFF
--- a/Microsoft.Azure.Cosmos/src/direct/Channel.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/Channel.cs
@@ -298,6 +298,7 @@ namespace Microsoft.Azure.Documents.Rntbd
             await this.DisposeAsync().ConfigureAwait(false);
         }
 
+        // Keep in sync with DisposeAsync().
         void IDisposable.Dispose()
         {
             this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
@@ -344,11 +345,17 @@ namespace Microsoft.Azure.Documents.Rntbd
             this.stateLock.Dispose();
         }
 
+        // Keep in sync with Dispose().
         public async ValueTask DisposeAsync()
         {
-            this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
-            this.ThrowIfDisposed();
+            if (this.disposed)
+            {
+                return;
+            }
+
             this.disposed = true;
+            GC.SuppressFinalize(this);
+            this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
             DefaultTrace.TraceInformation("[RNTBD Channel {0}] Async disposing RNTBD Channel {1}", this.ConnectionCorrelationId, this);
 
             Task initTask = null;
@@ -386,8 +393,14 @@ namespace Microsoft.Azure.Documents.Rntbd
                 }
             }
             Debug.Assert(this.dispatcher != null);
-            await this.dispatcher.DisposeAsync().ConfigureAwait(false);
-            this.stateLock.Dispose();
+            try
+            {
+                await this.dispatcher.DisposeAsync().ConfigureAwait(false);
+            }
+            finally
+            {
+                this.stateLock.Dispose();
+            }
         }
 
         #region Test hook.
--- a/Microsoft.Azure.Cosmos/src/direct/Channel.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/Channel.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Azure.Documents.Rntbd
 #endif
 
     // The RNTBD RPC channel. Supports multiple parallel requests and timeouts.
-    internal sealed class Channel : IChannel, IDisposable
+    internal sealed class Channel : IChannel, IDisposable, IAsyncDisposable
     {
         private readonly Dispatcher dispatcher;
         private readonly TimerPool timerPool;
@@ -293,6 +293,11 @@ namespace Microsoft.Azure.Documents.Rntbd
             ((IDisposable) this).Dispose();
         }
 
+        public async Task CloseAsync()
+        {
+            await this.DisposeAsync().ConfigureAwait(false);
+        }
+
         void IDisposable.Dispose()
         {
             this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
@@ -336,6 +341,52 @@ namespace Microsoft.Azure.Documents.Rntbd
             }
             Debug.Assert(this.dispatcher != null);
             this.dispatcher.Dispose();
+            this.stateLock.Dispose();
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
+            this.ThrowIfDisposed();
+            this.disposed = true;
+            DefaultTrace.TraceInformation("[RNTBD Channel {0}] Async disposing RNTBD Channel {1}", this.ConnectionCorrelationId, this);
+
+            Task initTask = null;
+            this.stateLock.EnterWriteLock();
+            try
+            {
+                if (this.state != State.Closed)
+                {
+                    initTask = this.initializationTask;
+                }
+                this.state = State.Closed;
+            }
+            finally
+            {
+                this.stateLock.ExitWriteLock();
+            }
+            if (initTask != null)
+            {
+                try
+                {
+                    await initTask.ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    DefaultTrace.TraceWarning(
+                        "[RNTBD Channel {0}] {1} initialization failed. Consuming the task " +
+                        "exception in {2}. Server URI: {3}. Exception: {4}",
+                        this.ConnectionCorrelationId,
+                        nameof(Channel),
+                        nameof(DisposeAsync),
+                        this.serverUri,
+                        e.Message);
+                    // Intentionally swallowing the exception. The caller can't
+                    // do anything useful with it.
+                }
+            }
+            Debug.Assert(this.dispatcher != null);
+            await this.dispatcher.DisposeAsync().ConfigureAwait(false);
             this.stateLock.Dispose();
         }
 
--- a/Microsoft.Azure.Cosmos/src/direct/Channel.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/Channel.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Documents.Rntbd
         private readonly int requestTimeoutSeconds;
         private readonly Uri serverUri;
         private readonly bool localRegionRequest;
-        private bool disposed = false;
+        private int disposed;
 
         private readonly ReaderWriterLockSlim stateLock =
             new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
@@ -79,7 +79,7 @@ namespace Microsoft.Azure.Documents.Rntbd
 
         public void InjectFaultInjectionConnectionError(TransportException transportException)
         {
-            if (!this.disposed)
+            if (this.disposed == 0)
             {
                 this.dispatcher.InjectFaultInjectionConnectionError(transportException);
             }
@@ -293,17 +293,18 @@ namespace Microsoft.Azure.Documents.Rntbd
             ((IDisposable) this).Dispose();
         }
 
-        public async Task CloseAsync()
-        {
-            await this.DisposeAsync().ConfigureAwait(false);
-        }
+        public Task CloseAsync() => this.DisposeAsync().AsTask();
 
         // Keep in sync with DisposeAsync().
         void IDisposable.Dispose()
         {
+            if (Interlocked.CompareExchange(ref this.disposed, 1, 0) != 0)
+            {
+                return;
+            }
+
+            GC.SuppressFinalize(this);
             this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
-            this.ThrowIfDisposed();
-            this.disposed = true;
             DefaultTrace.TraceInformation("[RNTBD Channel {0}] Disposing RNTBD Channel {1}", this.ConnectionCorrelationId, this);
 
             Task initTask = null;
@@ -341,19 +342,24 @@ namespace Microsoft.Azure.Documents.Rntbd
                 }
             }
             Debug.Assert(this.dispatcher != null);
-            this.dispatcher.Dispose();
-            this.stateLock.Dispose();
+            try
+            {
+                this.dispatcher.Dispose();
+            }
+            finally
+            {
+                this.stateLock.Dispose();
+            }
         }
 
         // Keep in sync with Dispose().
         public async ValueTask DisposeAsync()
         {
-            if (this.disposed)
+            if (Interlocked.CompareExchange(ref this.disposed, 1, 0) != 0)
             {
                 return;
             }
 
-            this.disposed = true;
             GC.SuppressFinalize(this);
             this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
             DefaultTrace.TraceInformation("[RNTBD Channel {0}] Async disposing RNTBD Channel {1}", this.ConnectionCorrelationId, this);
@@ -428,7 +434,7 @@ namespace Microsoft.Azure.Documents.Rntbd
 
         private void ThrowIfDisposed()
         {
-            if (this.disposed)
+            if (this.disposed != 0)
             {
                 throw new ObjectDisposedException(nameof(Channel));
             }
--- a/Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
@@ -101,13 +101,14 @@ namespace Microsoft.Azure.Documents.Rntbd
                 closeTasks.Add(channel.CloseAsync());
             }
 
+            Task whenAllTask = Task.WhenAll(closeTasks);
             try
             {
-                await Task.WhenAll(closeTasks).ConfigureAwait(false);
+                await whenAllTask.ConfigureAwait(false);
             }
-            catch (AggregateException ae)
+            catch (Exception)
             {
-                foreach (Exception inner in ae.Flatten().InnerExceptions)
+                foreach (Exception inner in whenAllTask.Exception.Flatten().InnerExceptions)
                 {
                     DefaultTrace.TraceWarning(
                         "[RNTBD ChannelDictionary] Async dispose encountered error during channel closure: {0}",
--- a/Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Documents.Rntbd
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos.Core.Trace;
     using Microsoft.Azure.Documents.FaultInjection;
 
     // ChannelDictionary maps server keys to load-balanced channels. There is
@@ -82,16 +83,30 @@ namespace Microsoft.Azure.Documents.Rntbd
 
         public async ValueTask DisposeAsync()
         {
-            this.ThrowIfDisposed();
-            this.disposed = true;
+            if (this.disposed)
+            {
+                return;
+            }
 
-            List<Task> closeTasks = new List<Task>();
+            this.disposed = true;
+            GC.SuppressFinalize(this);
+
+            List<Task> closeTasks = new List<Task>(this.channels.Count);
             foreach (IChannel channel in this.channels.Values)
             {
                 closeTasks.Add(channel.CloseAsync());
             }
 
-            await Task.WhenAll(closeTasks).ConfigureAwait(false);
+            try
+            {
+                await Task.WhenAll(closeTasks).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                DefaultTrace.TraceWarning(
+                    "[RNTBD ChannelDictionary] Async dispose encountered errors during channel closure: {0}",
+                    e.Message);
+            }
         }
 
         private void ThrowIfDisposed()
--- a/Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
@@ -5,12 +5,14 @@ namespace Microsoft.Azure.Documents.Rntbd
 {
     using System;
     using System.Collections.Concurrent;
+    using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Threading.Tasks;
     using Microsoft.Azure.Documents.FaultInjection;
 
     // ChannelDictionary maps server keys to load-balanced channels. There is
     // one load-balanced channel per back-end server.
-    internal sealed class ChannelDictionary : IChannelDictionary, IDisposable
+    internal sealed class ChannelDictionary : IChannelDictionary, IDisposable, IAsyncDisposable
     {
         private readonly ChannelProperties channelProperties;
         private bool disposed = false;
@@ -76,6 +78,20 @@ namespace Microsoft.Azure.Documents.Rntbd
             {
                 channel.Close();
             }
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            this.ThrowIfDisposed();
+            this.disposed = true;
+
+            List<Task> closeTasks = new List<Task>();
+            foreach (IChannel channel in this.channels.Values)
+            {
+                closeTasks.Add(channel.CloseAsync());
+            }
+
+            await Task.WhenAll(closeTasks).ConfigureAwait(false);
         }
 
         private void ThrowIfDisposed()
--- a/Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.Documents.Rntbd
     using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Azure.Cosmos.Core.Trace;
     using Microsoft.Azure.Documents.FaultInjection;
@@ -16,7 +17,7 @@ namespace Microsoft.Azure.Documents.Rntbd
     internal sealed class ChannelDictionary : IChannelDictionary, IDisposable, IAsyncDisposable
     {
         private readonly ChannelProperties channelProperties;
-        private bool disposed = false;
+        private int disposed;
 
         private ConcurrentDictionary<ServerKey, IChannel> channels =
             new ConcurrentDictionary<ServerKey, IChannel>();
@@ -73,8 +74,12 @@ namespace Microsoft.Azure.Documents.Rntbd
 
         public void Dispose()
         {
-            this.ThrowIfDisposed();
-            this.disposed = true;
+            if (Interlocked.CompareExchange(ref this.disposed, 1, 0) != 0)
+            {
+                return;
+            }
+
+            GC.SuppressFinalize(this);
             foreach (IChannel channel in this.channels.Values)
             {
                 channel.Close();
@@ -83,12 +88,11 @@ namespace Microsoft.Azure.Documents.Rntbd
 
         public async ValueTask DisposeAsync()
         {
-            if (this.disposed)
+            if (Interlocked.CompareExchange(ref this.disposed, 1, 0) != 0)
             {
                 return;
             }
 
-            this.disposed = true;
             GC.SuppressFinalize(this);
 
             List<Task> closeTasks = new List<Task>(this.channels.Count);
@@ -101,17 +105,20 @@ namespace Microsoft.Azure.Documents.Rntbd
             {
                 await Task.WhenAll(closeTasks).ConfigureAwait(false);
             }
-            catch (Exception e)
+            catch (AggregateException ae)
             {
-                DefaultTrace.TraceWarning(
-                    "[RNTBD ChannelDictionary] Async dispose encountered errors during channel closure: {0}",
-                    e.Message);
+                foreach (Exception inner in ae.Flatten().InnerExceptions)
+                {
+                    DefaultTrace.TraceWarning(
+                        "[RNTBD ChannelDictionary] Async dispose encountered error during channel closure: {0}",
+                        inner.Message);
+                }
             }
         }
 
         private void ThrowIfDisposed()
         {
-            if (this.disposed)
+            if (this.disposed != 0)
             {
                 throw new ObjectDisposedException(nameof(ChannelDictionary));
             }
--- a/Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.Documents.Rntbd
 
     // Dispatcher encapsulates the state and logic needed to dispatch multiple requests through
     // a single connection.
-    internal sealed class Dispatcher : IDisposable
+    internal sealed class Dispatcher : IDisposable, IAsyncDisposable
     {
         // Connection is thread-safe for sending.
         // Receiving is done only from the receive loop.
@@ -483,6 +483,38 @@ namespace Microsoft.Azure.Documents.Rntbd
             DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] RNTBD Dispatcher {1} is disposed", this.ConnectionCorrelationId, this);
         }
 
+        public async ValueTask DisposeAsync()
+        {
+            this.ThrowIfDisposed();
+            this.disposed = true;
+
+            DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] Async disposing RNTBD Dispatcher {1}", this.ConnectionCorrelationId, this);
+
+            Task idleTimerTaskCopy = null;
+            Debug.Assert(!Monitor.IsEntered(this.connectionLock));
+            lock (this.connectionLock)
+            {
+                this.StartConnectionShutdown();
+                idleTimerTaskCopy = this.StopIdleTimer();
+            }
+
+            await this.WaitTaskAsync(idleTimerTaskCopy, "idle timer").ConfigureAwait(false);
+
+            Task receiveTaskCopy = null;
+            Debug.Assert(!Monitor.IsEntered(this.connectionLock));
+            lock (this.connectionLock)
+            {
+                Debug.Assert(this.idleTimer == null);
+                Debug.Assert(this.idleTimerTask == null);
+
+                receiveTaskCopy = this.CloseConnection();
+            }
+
+            await this.WaitTaskAsync(receiveTaskCopy, "receive loop").ConfigureAwait(false);
+
+            DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] RNTBD Dispatcher {1} is disposed", this.ConnectionCorrelationId, this);
+        }
+
         private void StartIdleTimer()
         {
             DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] RNTBD idle connection monitor: Timer is starting...", this.ConnectionCorrelationId);
@@ -522,7 +554,7 @@ namespace Microsoft.Azure.Documents.Rntbd
             }
         }
 
-        private void OnIdleTimer(Task precedentTask)
+        private async Task OnIdleTimerAsync(Task precedentTask)
         {
             Task receiveTaskCopy = null;
 
@@ -572,7 +604,7 @@ namespace Microsoft.Azure.Documents.Rntbd
                 receiveTaskCopy = this.CloseConnection();
             }
 
-            this.WaitTask(receiveTaskCopy, "receive loop");
+            await this.WaitTaskAsync(receiveTaskCopy, "receive loop").ConfigureAwait(false);
         }
 
         // this.connectionLock must be held.
@@ -580,7 +612,7 @@ namespace Microsoft.Azure.Documents.Rntbd
         {
             Debug.Assert(Monitor.IsEntered(this.connectionLock));
             this.idleTimer = this.idleTimerPool.GetPooledTimer((int)timeToIdle.TotalSeconds);
-            this.idleTimerTask = this.idleTimer.StartTimerAsync().ContinueWith(this.OnIdleTimer, TaskContinuationOptions.OnlyOnRanToCompletion);
+            this.idleTimerTask = this.idleTimer.StartTimerAsync().ContinueWith(this.OnIdleTimerAsync, TaskContinuationOptions.OnlyOnRanToCompletion).Unwrap();
             this.idleTimerTask.ContinueWith(
                 failedTask =>
                 {
@@ -670,6 +702,28 @@ namespace Microsoft.Azure.Documents.Rntbd
                 Debug.Assert(!Monitor.IsEntered(this.callLock));
                 Debug.Assert(!Monitor.IsEntered(this.connectionLock));
                 t.Wait();
+            }
+            catch (Exception e)
+            {
+                DefaultTrace.TraceWarning(
+                    "[RNTBD Dispatcher {0}][{1}] Parallel task failed: {2}. Exception: {3}",
+                    this.ConnectionCorrelationId, this, description, e.Message);
+                // Intentionally swallowing the exception. The caller can't
+                // do anything useful with it.
+            }
+        }
+
+        private async Task WaitTaskAsync(Task t, string description)
+        {
+            if (t == null)
+            {
+                return;
+            }
+            try
+            {
+                Debug.Assert(!Monitor.IsEntered(this.callLock));
+                Debug.Assert(!Monitor.IsEntered(this.connectionLock));
+                await t.ConfigureAwait(false);
             }
             catch (Exception e)
             {
--- a/Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Azure.Documents.Rntbd
         private readonly bool enableChannelMultiplexing;
         private readonly Action<string, Exception> clientCertificateFailureHandler;
 
-        private bool disposed = false;
+        private int disposed;
 
         private ServerProperties serverProperties = null;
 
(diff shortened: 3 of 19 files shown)


Microsoft.Azure.Cosmos/src/direct/Channel.cs
```diff
@@ -339,6 +344,52 @@ void IDisposable.Dispose()
             this.stateLock.Dispose();
         }
 
+        public async ValueTask DisposeAsync()
+        {
+            this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
+            this.ThrowIfDisposed();
+            this.disposed = true;
```
NaluTripician: **🟡 Recommendation** · Design: Contract Violation

**`DisposeAsync()` violates `IAsyncDisposable` idempotency contract and has side-effect ordering issue**

Per the [.NET `IAsyncDisposable` guidelines](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync), `DisposeAsync()` **must be idempotent** — calling it multiple times should be a no-op. The current `ThrowIfDisposed() + disposed = true` pattern throws `ObjectDisposedException` on the second call, which can crash a disposal chain if both `Close()` and `DisposeAsync()` are called on the same object (realistic during the transition to async callers).

Additionally, `chaosInterceptor?.OnChannelDispose()` fires **before** the disposal guard on line 349. If the object is already disposed, this side-effect still fires before `ThrowIfDisposed()` throws, potentially corrupting test/fault-injection state.

This same pattern applies to `Dispatcher.DisposeAsync()`, `ChannelDictionary.DisposeAsync()`, and `LoadBalancingChannel.DisposeAsync()`.

`LbChannelState.DisposeAsync()` already demonstrates the correct pattern with `Interlocked.Increment` — adopt it everywhere:

```csharp
if (this.disposed) return;
this.disposed = true;
// Move chaosInterceptor call AFTER this guard
this.chaosInterceptor?.OnChannelDispose(this.ConnectionCorrelationId);
```

---

Microsoft.Azure.Cosmos/src/direct/Channel.cs
```diff
@@ -339,6 +344,52 @@ void IDisposable.Dispose()
...
+                        nameof(Channel),
+                        nameof(DisposeAsync),
+                        this.serverUri,
+                        e.Message);
+                    // Intentionally swallowing the exception. The caller can't
+                    // do anything useful with it.
+                }
+            }
+            Debug.Assert(this.dispatcher != null);
+            await this.dispatcher.DisposeAsync().ConfigureAwait(false);
```
NaluTripician: **🟡 Recommendation** · Robustness: Resource Leak

**`stateLock.Dispose()` not protected by `try/finally` — `ReaderWriterLockSlim` leaks if dispatcher disposal throws**

If `dispatcher.DisposeAsync()` throws (e.g., `StartConnectionShutdown()` propagates an `AggregateException` from cancellation callbacks), the `ReaderWriterLockSlim` is never disposed, leaking kernel wait handles. While `WaitTaskAsync` internally swallows exceptions, `StartConnectionShutdown()` can still throw.

The sync `Dispose()` has the same vulnerability (pre-existing), but since `DisposeAsync()` is new code, this is an opportunity to get it right.

```csharp
try
{
    await this.dispatcher.DisposeAsync().ConfigureAwait(false);
}
finally
{
    this.stateLock.Dispose();
}
```

---

Microsoft.Azure.Cosmos/src/direct/LbChannelState.cs
```diff
@@ -96,5 +97,14 @@ public void Dispose()
                 this.channel.Close();
             }
         }
+
+        public async Task DisposeAsync()
```
NaluTripician: **🟡 Recommendation** · Design: Interface Inconsistency

**`LbChannelState` and `LoadBalancingPartition` add `DisposeAsync()` but don't implement `IAsyncDisposable`**

Both classes add a `public async Task DisposeAsync()` method but do **not** implement `IAsyncDisposable`. They return `Task` instead of `ValueTask`. Meanwhile, `Channel`, `Dispatcher`, `ChannelDictionary`, and `LoadBalancingChannel` all implement `IAsyncDisposable` with `ValueTask`.

This creates two issues:
1. Callers can't use `await using` for these types
2. The method name shadows the well-known `IAsyncDisposable` pattern with the wrong return type, creating friction if the interface is added later

Either implement `IAsyncDisposable` properly (returning `ValueTask`) or rename to avoid confusion (e.g., `CloseInternalAsync()`).

*Context: The rest of the hierarchy (`Channel`, `Dispatcher`, `ChannelDictionary`, `LoadBalancingChannel`) all implement `IAsyncDisposable` correctly.*

---

Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
```diff
@@ -78,6 +80,20 @@ public void Dispose()
...
+            this.ThrowIfDisposed();
+            this.disposed = true;
+
+            List<Task> closeTasks = new List<Task>();
+            foreach (IChannel channel in this.channels.Values)
+            {
+                closeTasks.Add(channel.CloseAsync());
+            }
+
+            await Task.WhenAll(closeTasks).ConfigureAwait(false);
```
NaluTripician: **🟡 Recommendation** · Robustness: Exception Handling

**`Task.WhenAll` in concurrent disposal lacks exception handling — one failing channel crashes the disposal chain**

If any individual `CloseAsync()` throws, the `AggregateException` propagates unhandled — preventing cleanup telemetry and potentially crashing the caller. This is particularly important because the PR targets scenarios with many connections (N simultaneous idle connections), so the probability of at least one failing is non-trivial.

The same issue exists in `LoadBalancingChannel.DisposeAsync()` (line 213). `WaitTaskAsync` in `Dispatcher` swallows exceptions correctly — the same defensive pattern should apply at these higher levels:

```csharp
try
{
    await Task.WhenAll(closeTasks).ConfigureAwait(false);
}
catch (Exception ex)
{
    DefaultTrace.TraceWarning(
        "[RNTBD ChannelDictionary] Async dispose encountered errors: {0}", ex.Message);
}
```

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -483,6 +483,38 @@ public void Dispose()
             DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] RNTBD Dispatcher {1} is disposed", this.ConnectionCorrelationId, this);
         }
 
+        public async ValueTask DisposeAsync()
+        {
+            this.ThrowIfDisposed();
+            this.disposed = true;
```
NaluTripician: **🟡 Recommendation** · Maintainability: Code Duplication

**Dual `Dispose()`/`DisposeAsync()` logic duplication creates maintenance risk**

~30 lines of identical business logic (`StartConnectionShutdown`, `StopIdleTimer`, `WaitTask`, `CloseConnection`) are duplicated between sync `Dispose()` (line 454) and async `DisposeAsync()` (line 486). Over time, fixes in one path may not be applied to the other — the classic dual-dispose maintenance trap. `Channel.cs` has the same issue (~45 lines duplicated).

Also, the `ThrowIfDisposed(); this.disposed = true;` pattern on lines 488-489 is a TOCTOU race — two threads can both pass the check before either sets the flag. `LbChannelState` already uses `Interlocked.Increment` for its disposal guard — this pattern should be adopted.

At minimum, add cross-reference comments: `// Keep in sync with DisposeAsync()` / `// Keep in sync with Dispose()`.

*Context: PR #5597 established that disposal flag ordering is critical in this codebase — `isDisposing` must be set before resource teardown. The current PR correctly sets `disposed = true` early, but the non-atomic check widens the race surface.*

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -483,6 +483,38 @@ public void Dispose()
             DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] RNTBD Dispatcher {1} is disposed", this.ConnectionCorrelationId, this);
         }
 
+        public async ValueTask DisposeAsync()
```
NaluTripician, line 492: **🟡 Recommendation** · Coverage: Missing Tests

**No tests added for the async disposal paths or `OnIdleTimerAsync` fix**

The core behavioral change (switching `OnIdleTimer` to `OnIdleTimerAsync`) changes the thread execution model of a critical infrastructure callback. The `DisposeAsync` chain across 6 classes also has no test coverage.

PR #5300 (FeedIterator deadlock fix) established the team pattern of adding tests with mock `SynchronizationContext` and timeout-based deadlock detection for starvation fixes. PR #5155 reviewer asked to "check all analogous paths."

At minimum, tests should verify:
1. `OnIdleTimerAsync` completes without blocking a thread pool thread (the core fix)
2. `DisposeAsync` on the channel hierarchy completes cleanly
3. Concurrent `DisposeAsync` + `OnIdleTimerAsync` doesn't double-close (relies on `connection.Disposed` guard in `CloseConnection()`, which is currently untested)

---

Microsoft.Azure.Cosmos/src/direct/LoadBalancingPartition.cs
```diff
@@ -356,6 +356,37 @@ public void Dispose()
...
+                this.capacityLock.ExitWriteLock();
+            }
+
+            await Task.WhenAll(disposeTasks).ConfigureAwait(false);
+
+            try
+            {
+                this.capacityLock.Dispose();
+            }
+            catch(SynchronizationLockException)
```
NaluTripician: **🟢 Suggestion** · Observability: Silent Exception Swallowing

**`SynchronizationLockException` swallowed without logging — makes production debugging difficult**

This catch block silently `return`s with no trace or log. The sync `Dispose()` has the same pattern (pre-existing), but consistent with how `WaitTaskAsync` in `Dispatcher` logs swallowed exceptions, this should include tracing:

```csharp
catch (SynchronizationLockException e)
{
    DefaultTrace.TraceWarning(
        "[RNTBD LoadBalancingPartition] SynchronizationLockException during async dispose: {0}", e.Message);
    return;
}
```

---

Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
```diff
@@ -78,6 +80,20 @@ public void Dispose()
             }
         }
 
+        public async ValueTask DisposeAsync()
+        {
+            this.ThrowIfDisposed();
+            this.disposed = true;
+
+            List<Task> closeTasks = new List<Task>();
```
NaluTripician: **🟢 Suggestion** · Efficiency: Memory Allocation

**`List<Task>` created without capacity hint — will resize multiple times with many channels**

Since `channels` is a `ConcurrentDictionary` with `Count` available, pre-size the list to avoid repeated resizing in the N-connections scenario this PR targets:

```csharp
List<Task> closeTasks = new List<Task>(this.channels.Count);
```

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -483,6 +483,38 @@ public void Dispose()
             DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] RNTBD Dispatcher {1} is disposed", this.ConnectionCorrelationId, this);
         }
 
+        public async ValueTask DisposeAsync()
+        {
+            this.ThrowIfDisposed();
+            this.disposed = true;
+
+            DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] Async disposing RNTBD Dispatcher {1}", this.ConnectionCorrelationId, this);
```
NaluTripician, line 501: **🟢 Suggestion** · Best Practice: Defensive Disposal

**Missing `GC.SuppressFinalize(this)` in `DisposeAsync()` implementations**

The [.NET `IAsyncDisposable` pattern](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync) recommends calling `GC.SuppressFinalize(this)` in `DisposeAsync()` when the class also implements `IDisposable`. This prevents the finalizer from running after async disposal. Defensive practice even without explicit finalizers today — applies to `Channel`, `Dispatcher`, `ChannelDictionary`, and `LoadBalancingChannel`.

```csharp
public async ValueTask DisposeAsync()
{
    GC.SuppressFinalize(this);
    ...
}
```

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -572,15 +604,15 @@ private void OnIdleTimer(Task precedentTask)
...
+            await this.WaitTaskAsync(receiveTaskCopy, "receive loop").ConfigureAwait(false);
         }
 
         // this.connectionLock must be held.
         private void ScheduleIdleTimer(TimeSpan timeToIdle)
         {
             Debug.Assert(Monitor.IsEntered(this.connectionLock));
             this.idleTimer = this.idleTimerPool.GetPooledTimer((int)timeToIdle.TotalSeconds);
-            this.idleTimerTask = this.idleTimer.StartTimerAsync().ContinueWith(this.OnIdleTimer, TaskContinuationOptions.OnlyOnRanToCompletion);
+            this.idleTimerTask = this.idleTimer.StartTimerAsync().ContinueWith(this.OnIdleTimerAsync, TaskContinuationOptions.OnlyOnRanToCompletion).Unwrap();
```
NaluTripician, line 631: **💬 Observation** · Design: Complexity

**`ContinueWith + Unwrap` pattern works correctly but is harder to reason about than pure async**

The `.Unwrap()` is essential and the PR description correctly explains why. Without it, `idleTimerTask` would complete when `OnIdleTimerAsync` *starts* (returns its `Task`), not when it *finishes*, causing `StopIdleTimer`/`DisposeAsync` to proceed before the callback finishes closing the connection.

For a future refactor, a pure async wrapper would eliminate the `ContinueWith`/`Unwrap`/`TaskContinuationOptions` complexity entirely:

```csharp
private async Task RunIdleTimerCycleAsync(TimeSpan timeToIdle)
{
    await this.idleTimerPool.GetPooledTimer(...).StartTimerAsync();
    await this.OnIdleTimerAsync();
}
```

---

Microsoft.Azure.Cosmos/src/direct/LoadBalancingChannel.cs
```diff
@@ -188,6 +194,25 @@ void IDisposable.Dispose()
...
+                disposeTasks.Add(this.singlePartition.DisposeAsync());
+            }
+            if (this.partitions != null)
+            {
+                for (int i = 0; i < this.partitions.Length; i++)
+                {
+                    disposeTasks.Add(this.partitions[i].DisposeAsync());
+                }
+            }
+            await Task.WhenAll(disposeTasks).ConfigureAwait(false);
```
NaluTripician: **💬 Observation** · Design: Phased Rollout

**Async disposal chain exists but has no upstream caller — only `OnIdleTimerAsync` is actively wired in**

The `DisposeAsync` methods form a complete chain: `ChannelDictionary` → `LoadBalancingChannel` → `LoadBalancingPartition` → `LbChannelState` → `Channel` → `Dispatcher`. However, `IChannelDictionary` was **not updated** to include `DisposeAsync()`, so no upstream code can call async disposal through the interface.

This means Path 2 (mass disposal starvation from `ChannelDictionary.Dispose()`) is not fixed by this PR — only Path 1 (idle timer via `OnIdleTimerAsync`) is active. This is a valid phased approach, but worth documenting the intent for the follow-up PR that wires in `DisposeAsync` at the top level.

---

Microsoft.Azure.Cosmos/src/direct/LoadBalancingPartition.cs
```diff
@@ -356,6 +356,49 @@ public void Dispose()
...
+            {
+                disposeTasks = new List<Task>(this.openChannels.Count);
+                foreach (LbChannelState channelState in this.openChannels)
+                {
+                    disposeTasks.Add(channelState.DisposeAsync().AsTask());
+                }
+            }
+            finally
+            {
+                this.capacityLock.ExitWriteLock();
```
NaluTripician, line 387: **🟡 Recommendation** · Concurrency: Behavioral Divergence

**Early write lock release in `DisposeAsync()` creates a window where new channels won't be disposed**

The sync `Dispose()` holds the `capacityLock` write lock during the entire channel disposal loop (lines 334-344), preventing `OpenChannel()` from adding new `LbChannelState` entries. The async `DisposeAsync()` releases the write lock here before awaiting `Task.WhenAll(disposeTasks)`.

During this window, another thread could acquire the write lock and add a new channel to `openChannels` that won't be disposed — creating a resource leak. While this is a shutdown path and the concurrent add is unlikely, the behavioral divergence from the sync path is worth noting.

**Suggestion:** Either set a disposal flag inside the lock before releasing (so `OpenChannel` can reject new channels), or document why the early release is safe in this context.

---

Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
```diff
@@ -78,6 +81,34 @@ public void Dispose()
...
+            }
+
+            try
+            {
+                await Task.WhenAll(closeTasks).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                DefaultTrace.TraceWarning(
+                    "[RNTBD ChannelDictionary] Async dispose encountered errors during channel closure: {0}",
```
NaluTripician: **🟡 Recommendation** · Observability: Partial Exception Logging

**`Task.WhenAll` catch only logs the first exception — other failures are invisible**

`Task.WhenAll` throws an `AggregateException` where `e.Message` only contains the first inner exception's message. If multiple channels fail during concurrent disposal (the exact N-connections scenario this PR targets), only the first failure is visible in traces.

The same pattern applies to `LoadBalancingChannel.DisposeAsync()` (line 229) and `LoadBalancingPartition.DisposeAsync()` (line 383).

**Suggestion:** Log `e.ToString()` for full stack traces, or iterate `((AggregateException)e).InnerExceptions` to log each failure individually.

---

Microsoft.Azure.Cosmos/src/direct/Channel.cs
```diff
@@ -293,6 +293,12 @@ public void Close()
             ((IDisposable) this).Dispose();
         }
 
+        public async Task CloseAsync()
+        {
+            await this.DisposeAsync().ConfigureAwait(false);
+        }
+
+        // Keep in sync with DisposeAsync().
```
NaluTripician, line 298: **🟡 Recommendation** · Maintainability: Sync/Async Parity Gap

**'Keep in sync' comments promise behavioral parity that doesn't exist**

The async `DisposeAsync()` has several improvements the sync `Dispose()` lacks:
- `DisposeAsync()` wraps `dispatcher.DisposeAsync()` in try/finally for `stateLock.Dispose()` (lines 396-403), but sync `Dispose()` does not (lines 344-345) — if `dispatcher.Dispose()` throws, the `ReaderWriterLockSlim` leaks
- Async paths use idempotent `if (disposed) return` while sync paths use `ThrowIfDisposed()`  
- Async paths call `GC.SuppressFinalize(this)` while sync paths don't

The 'Keep in sync' cross-reference comments create an expectation that the two methods are structurally equivalent, but they already diverge. Future maintainers may assume parity exists when it doesn't.

Same pattern applies to `Dispatcher.cs` (lines 454 vs 487).

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -483,6 +484,44 @@ public void Dispose()
             DefaultTrace.TraceInformation("[RNTBD Dispatcher {0}] RNTBD Dispatcher {1} is disposed", this.ConnectionCorrelationId, this);
         }
 
+        // Keep in sync with Dispose().
+        public async ValueTask DisposeAsync()
+        {
+            if (this.disposed)
```
NaluTripician: **🟡 Recommendation** · Concurrency: Cross-Path Disposal Race

**Non-volatile, non-atomic `disposed` flag — concurrent `Dispose()` + `DisposeAsync()` can double-execute shutdown**

The `disposed` field is `private bool` (not volatile) and is checked/set non-atomically. If `Dispose()` and `DisposeAsync()` are called concurrently from different threads (realistic during the transition to async callers):
1. Thread A calls `Dispose()`, passes `ThrowIfDisposed()` (disposed still false)
2. Thread B calls `DisposeAsync()`, passes `if (this.disposed)` check (still false)
3. Both set `disposed = true` and execute full shutdown logic

The internal guards (`cancellation.IsCancellationRequested`, `connection.Disposed`) prevent crashes, but double-execution wastes resources and produces confusing interleaved trace output.

`LbChannelState` already uses `Interlocked.Increment` for an atomic disposal guard — this pattern should be adopted by the other classes.

Same issue exists in `Channel.cs`, `ChannelDictionary.cs`, and `LoadBalancingChannel.cs`.

---

Microsoft.Azure.Cosmos/src/direct/Channel.cs
```diff
@@ -293,6 +293,12 @@ public void Close()
             ((IDisposable) this).Dispose();
         }
 
+        public async Task CloseAsync()
```
NaluTripician: **🟢 Suggestion** · Performance: Unnecessary Async State Machine

**`CloseAsync()` creates an unnecessary async state machine — use `.AsTask()` instead**

```csharp
public async Task CloseAsync()
{
    await this.DisposeAsync().ConfigureAwait(false);
}
```

This allocates an async state machine to convert `ValueTask` to `Task`. In the N-connections-going-idle scenario, this creates N unnecessary allocations. Simplify to:

```csharp
public Task CloseAsync() => this.DisposeAsync().AsTask();
```

Same applies to `LoadBalancingChannel.CloseAsync()` (line 174).

---

Microsoft.Azure.Cosmos/src/direct/LoadBalancingChannel.cs
```diff
@@ -188,6 +195,42 @@ void IDisposable.Dispose()
...
+        // Keep in sync with Dispose().
+        // TODO: Wire upstream callers (IChannelDictionary) to call DisposeAsync
+        // to fully address Path 2 (mass disposal starvation).
+        public async ValueTask DisposeAsync()
+        {
+            if (this.disposed)
+            {
+                return;
+            }
+
```
NaluTripician, line 209: **🟢 Suggestion** · Performance: List Capacity

**`List<Task>` created without capacity hint — will resize with many partitions**

```csharp
List<Task> disposeTasks = new List<Task>();
```

The count is knowable: `(singlePartition != null ? 1 : 0) + (partitions?.Length ?? 0)`. `ChannelDictionary.DisposeAsync()` was correctly fixed with `new List<Task>(this.channels.Count)` but this one was missed.

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -522,7 +561,7 @@ private void StartIdleTimer()
             }
         }
 
-        private void OnIdleTimer(Task precedentTask)
+        private async Task OnIdleTimerAsync(Task precedentTask)
```
NaluTripician, line 567: **🟢 Suggestion** · Clarity: Unused Parameter

**`precedentTask` parameter is never used in `OnIdleTimerAsync`**

```csharp
private async Task OnIdleTimerAsync(Task precedentTask)
```

The `precedentTask` parameter is required by the `ContinueWith` delegate signature (`Func<Task, Task>`), but it's never referenced in the method body. Consider renaming to `_` or `_precedentTask` to signal it's intentionally unused — this helps future maintainers understand it's a delegate-signature requirement, not a missed usage.

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -681,6 +720,28 @@ private void WaitTask(Task t, string description)
...
+                Debug.Assert(!Monitor.IsEntered(this.callLock));
+                Debug.Assert(!Monitor.IsEntered(this.connectionLock));
+                await t.ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                DefaultTrace.TraceWarning(
+                    "[RNTBD Dispatcher {0}][{1}] Parallel task failed: {2}. Exception: {3}",
+                    this.ConnectionCorrelationId, this, description, e.Message);
+                // Intentionally swallowing the exception. The caller can't
```
NaluTripician, line 749: **🟢 Suggestion** · Observability: Stack Trace Logging

**`WaitTaskAsync` logs `e.Message` only — use `e.ToString()` for full stack traces**

```csharp
DefaultTrace.TraceWarning(
    "[RNTBD Dispatcher {0}][{1}] Parallel task failed: {2}. Exception: {3}",
    this.ConnectionCorrelationId, this, description, e.Message);
```

For infrastructure-level components, `e.ToString()` (or at least `e.GetType().Name + ": " + e.Message`) provides crucial diagnostic information. When these fire in production, the exception type often matters as much as the message. Same applies to the `initTask` catch in `Channel.DisposeAsync()` (line 390).

---

Microsoft.Azure.Cosmos/src/direct/LoadBalancingChannel.cs
```diff
@@ -188,6 +195,42 @@ void IDisposable.Dispose()
             }
         }
```
NaluTripician, line 199: **💬 Observation** · Design: Phased Rollout

**Path 2 TODO should reference a tracking issue**

```csharp
// TODO: Wire upstream callers (IChannelDictionary) to call DisposeAsync
// to fully address Path 2 (mass disposal starvation).
```

The phased approach (fixing Path 1 now, Path 2 later) is reasonable. However, TODOs without tracking issues tend to get lost. Consider creating a GitHub issue and referencing it here: `// TODO(#XXXX): Wire upstream callers...`  

Note: `IChannelDictionary` was not updated to include `DisposeAsync()`, confirming the async disposal chain has no upstream entry point yet.

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -572,15 +611,15 @@ private void OnIdleTimer(Task precedentTask)
...
+            await this.WaitTaskAsync(receiveTaskCopy, "receive loop").ConfigureAwait(false);
         }
 
         // this.connectionLock must be held.
         private void ScheduleIdleTimer(TimeSpan timeToIdle)
         {
             Debug.Assert(Monitor.IsEntered(this.connectionLock));
             this.idleTimer = this.idleTimerPool.GetPooledTimer((int)timeToIdle.TotalSeconds);
-            this.idleTimerTask = this.idleTimer.StartTimerAsync().ContinueWith(this.OnIdleTimer, TaskContinuationOptions.OnlyOnRanToCompletion);
+            this.idleTimerTask = this.idleTimer.StartTimerAsync().ContinueWith(this.OnIdleTimerAsync, TaskContinuationOptions.OnlyOnRanToCompletion).Unwrap();
```
NaluTripician, line 631: **💬 Observation** · Architecture: Positive Assessment

**Core `ContinueWith(OnIdleTimerAsync).Unwrap()` fix is architecturally sound**

```csharp
this.idleTimerTask = this.idleTimer.StartTimerAsync()
    .ContinueWith(this.OnIdleTimerAsync, TaskContinuationOptions.OnlyOnRanToCompletion)
    .Unwrap();
```

The `.Unwrap()` is essential and correctly applied — without it, `idleTimerTask` would complete when `OnIdleTimerAsync` *starts* (returns its `Task`), not when it *finishes*, causing `StopIdleTimer()`/`DisposeAsync()` to proceed before the callback finishes closing the connection.

The concurrency model between `OnIdleTimerAsync` and `DisposeAsync` is properly handled: `DisposeAsync` waits for `idleTimerTask` (via `WaitTaskAsync`), so the idle timer callback always completes before disposal proceeds. The lock serialization on `connectionLock` ensures `CloseConnection()` is never called concurrently.

---

NaluTripician:
## PR Review Summary

**Overall Assessment:** The core fix (`ContinueWith(OnIdleTimerAsync).Unwrap()`) is architecturally sound and correctly eliminates thread pool starvation from idle timer callbacks. The author addressed feedback from 2 prior review iterations (16 AI-generated comments), fixing atomic disposal guards, try/finally patterns, GC.SuppressFinalize, and more.

**New findings (after deduplication):** 6 new issues found, 1 blocking.

| # | Severity | Finding |
|---|----------|---------|
| 1 | Blocking | `catch (AggregateException)` after `await Task.WhenAll()` is dead code in 3 files |
| 2 | Recommendation | `LoadBalancingPartition.Dispose()` doesn't share atomic disposal guard with `DisposeAsync()` |
| 3 | Recommendation | Missing `GC.SuppressFinalize` in `LoadBalancingPartition` and `LbChannelState` |
| 4 | Suggestion | Inconsistent disposal guard idiom (`Increment` vs `CompareExchange`) |
| 5 | Suggestion | `IChannel.CloseAsync()` returns `Task` forcing `.AsTask()` allocation |
| 6 | Observation | Dormant DisposeAsync chain compounds untested bugs |

**What the PR gets right:**
- Core `ContinueWith(OnIdleTimerAsync).Unwrap()` fix is correct and essential
- Atomic `Interlocked.CompareExchange` disposal guards eliminate TOCTOU race
- All `await` calls correctly placed outside lock scopes
- `try/finally` around dispatcher/stateLock disposal prevents resource leaks
- Backward compatible: all sync paths preserved

---
<sub>AI-generated review summary</sub>

Microsoft.Azure.Cosmos/src/direct/ChannelDictionary.cs
```diff
@@ -70,17 +74,51 @@ public bool TryGetChannel(Uri requestUri, out IChannel channel)
...
+            foreach (IChannel channel in this.channels.Values)
+            {
+                closeTasks.Add(channel.CloseAsync());
+            }
+
+            try
+            {
+                await Task.WhenAll(closeTasks).ConfigureAwait(false);
+            }
+            catch (AggregateException ae)
```
NaluTripician: **🔴 Blocking** · Correctness: Wrong Exception Type

**`catch (AggregateException)` after `await Task.WhenAll()` is dead code — exception handling will never execute**

In .NET, `await` on a faulted task unwraps the `AggregateException` and throws only the **first inner exception**. So `catch (AggregateException ae)` will never execute for common exceptions (`IOException`, `ObjectDisposedException`, etc.). The code looks like it defensively handles all disposal errors, but actually lets the first exception propagate unhandled.

The same pattern exists in `LoadBalancingChannel.DisposeAsync()` and `LoadBalancingPartition.DisposeAsync()`.

**Fix:** Save the `Task.WhenAll` result to a variable, catch `Exception` broadly, and access `task.Exception.Flatten().InnerExceptions` for the full error list:

```csharp
Task whenAllTask = Task.WhenAll(closeTasks);
try
{
    await whenAllTask.ConfigureAwait(false);
}
catch (Exception)
{
    foreach (Exception inner in whenAllTask.Exception.Flatten().InnerExceptions)
    {
        DefaultTrace.TraceWarning(
            "[RNTBD ChannelDictionary] Async dispose encountered error during channel closure: {0}",
            inner.Message);
    }
}
```

**Why it matters:** When the disposal chain is wired up (Path 2, per TODO #4393), exceptions during concurrent channel closure will propagate instead of being caught/logged, potentially crashing the disposal chain. `Task.WhenAll` ensures all tasks complete before throwing, so the individual close operations still run — but the caller gets an unexpected unhandled exception.

---

Microsoft.Azure.Cosmos/src/direct/LoadBalancingPartition.cs
```diff
@@ -356,6 +357,57 @@ public void Dispose()
             }
         }
 
+        public async ValueTask DisposeAsync()
+        {
+            if (Interlocked.CompareExchange(ref this.disposed, 1, 0) != 0)
```
NaluTripician, line 368: **🟡 Recommendation** · Concurrency: Disposal Guard Mismatch

**`LoadBalancingPartition.Dispose()` doesn't share the atomic disposal guard with `DisposeAsync()`**

All other classes in the PR (Channel, Dispatcher, ChannelDictionary, LoadBalancingChannel) modified **both** `Dispose()` and `DisposeAsync()` to share the same `disposed` int field with `Interlocked.CompareExchange`. But `LoadBalancingPartition.Dispose()` was **not updated** — it has no disposal guard at all. It just enters the write lock and disposes channels.

This means:
- `Dispose()` + `DisposeAsync()` on the same instance both execute fully (no cross-guard)
- `Dispose()` can be called multiple times (no guard)
- `capacityLock.Dispose()` could be called twice, throwing `ObjectDisposedException`

**Fix:** Add the same `Interlocked.CompareExchange` guard to sync `Dispose()`, matching the pattern in all other classes. Also add `GC.SuppressFinalize(this)` for consistency.

*Context: Past PR #4005 attempted to fix `Dispose()` race conditions in `LoadBalancingPartition` but was closed. PR #3870 addressed crashes disposing with in-flight requests. The disposal guard mismatch in this PR creates a similar race surface.*

---

Microsoft.Azure.Cosmos/src/direct/LoadBalancingPartition.cs
```diff
@@ -356,6 +357,57 @@ public void Dispose()
             }
         }
 
+        public async ValueTask DisposeAsync()
```
NaluTripician, line 366: **🟡 Recommendation** · Consistency: Missing Defensive Pattern

**Missing `GC.SuppressFinalize(this)` in both `Dispose()` and `DisposeAsync()`**

Channel, Dispatcher, ChannelDictionary, and LoadBalancingChannel all call `GC.SuppressFinalize(this)` in both disposal methods. `LoadBalancingPartition` calls it in **neither**. `LbChannelState` also omits it.

While neither class has an explicit finalizer today, `GC.SuppressFinalize` is defensive best practice per the .NET `IAsyncDisposable` pattern. The inconsistency within the PR's own changes creates confusion — a reader expects the pattern to be uniform across all 7 files.

---

Microsoft.Azure.Cosmos/src/direct/LbChannelState.cs
```diff
@@ -96,5 +97,14 @@ public void Dispose()
                 this.channel.Close();
             }
         }
+
+        public async ValueTask DisposeAsync()
+        {
+            int disposeInvocationCounter = Interlocked.Increment(ref this.stateDisposeCounter);
```
NaluTripician, line 104: **🟢 Suggestion** · Consistency: Disposal Guard Idiom

**Inconsistent disposal guard: `Interlocked.Increment` here vs `Interlocked.CompareExchange` everywhere else**

This class uses the pre-existing `Interlocked.Increment` pattern for its disposal guard, while all other classes in the PR use `Interlocked.CompareExchange(ref this.disposed, 1, 0)`. Both achieve atomic single-execution, but the different patterns make the disposal hierarchy harder to reason about at a glance. Not worth changing in this PR since it's pre-existing, but worth noting for future consistency.

---

Microsoft.Azure.Cosmos/src/direct/IChannel.cs
```diff
@@ -27,5 +27,7 @@ public Task OpenChannelAsync(
         bool Healthy { get; }
 
         void Close();
+
+        Task CloseAsync();
```
NaluTripician, line 31: **🟢 Suggestion** · Design: Return Type

**`CloseAsync()` returns `Task` — forces `.AsTask()` allocation at every bridge site**

All implementations bridge from `ValueTask DisposeAsync()` to `Task CloseAsync()` via `.AsTask()`. In the N-connections-going-idle scenario this PR targets, that's N unnecessary `Task` allocations. If the interface returned `ValueTask`, the bridge would be free.

This is a minor point — the interface was designed for maximum compatibility, and changing it now would affect other `IChannel` implementations. But worth considering for a future revision.

---

Microsoft.Azure.Cosmos/src/direct/LoadBalancingChannel.cs
```diff
@@ -188,9 +197,48 @@ void IDisposable.Dispose()
             }
         }
 
+        // Keep in sync with Dispose().
+        // TODO(#4393): Wire upstream callers (IChannelDictionary) to call DisposeAsync
```
NaluTripician, line 201: **💬 Observation** · Risk: Dormant Code With Bugs

**The dormant `DisposeAsync` chain has untested code with known bugs**

The entire `DisposeAsync` hierarchy (`ChannelDictionary` → `LoadBalancingChannel` → `LoadBalancingPartition` → `LbChannelState` → `Channel` → `Dispatcher`) exists but has no upstream caller. `IChannelDictionary` doesn't include `DisposeAsync()`, and `TransportClient.Dispose()` still calls sync `Dispose()`.

The dormant code already has a correctness bug (the `catch (AggregateException)` dead code in Finding #1) that would be caught by tests. When the follow-up PR wires this chain up (per this TODO), those bugs will surface in production rather than in testing.

The phased approach is reasonable, but testing the dormant chain *now* — even without wiring it to upstream callers — would catch issues like Finding #1 before they ship.

---

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -572,15 +614,15 @@ private void OnIdleTimer(Task precedentTask)
...
```
kirankumarkolli, line 631: Thoughts on tests/validation?

NaluTripician: Will add testing and validation to this PR and then merge the tests once direct package is updated

NaluTripician:
## Reproduction Results — Before/After Proof

A standalone repro (`repro/Program.cs`) simulates the exact blocking pattern from `OnIdleTimer`: 200 `ContinueWith` callbacks that need to wait for receive tasks. Run with `dotnet run -- both` to see both modes.

### Sync Mode (base `msdata/direct` branch behavior)
```
=== SYNC MODE (simulates base msdata/direct branch) ===
Connections: 200
Thread pool min threads: 12

Callbacks started:     0/200
Callbacks completed:   0/200
Thread pool threads:   0 → 30 (peak: 0)
Thread spike:          +0
Probe latency:         10193ms

╔══════════════════════════════════════════════════════════════╗
║  ❌ THREAD POOL STARVATION DETECTED                         ║
║  QueueUserWorkItem could not execute within 10 seconds.     ║
║  This confirms the bug from issue #4393.                    ║
╚══════════════════════════════════════════════════════════════╝
Total time:            12237ms
```

### Async Mode (fix branch behavior)
```
=== ASYNC MODE (simulates fix branch) ===
Connections: 200
Thread pool min threads: 12

Callbacks started:     200/200
Callbacks completed:   0/200
Thread pool threads:   30 → 34 (peak: 31)
Thread spike:          +1
Probe latency:         0ms

✅ Thread pool remained responsive (probe latency: 0ms)
Total time:            2060ms
```

### Comparison

| Metric | Sync (`t.Wait()`) | Async (`await t`) |
|--------|-------------------|-------------------|
| **Thread pool probe** | ❌ **STARVED** (10,193ms) | ✅ Responsive (**0ms**) |
| **Callbacks started** | 0/200 | 200/200 |
| **Thread spike** | Pool exhausted | +1 |
| **Total time** | 12,237ms | 2,060ms |

The sync path couldn't even **start** 1 of the 200 callbacks — the thread pool was completely saturated by `Task.Run` work items that immediately blocked on `t.Wait()`, preventing the `QueueUserWorkItem` probe from executing for over 10 seconds. The async path started all 200 callbacks and the probe executed in 0ms.

repro/Program.cs
```diff
@@ -0,0 +1,178 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+
+class Program
+{
+    const int ConnectionCount = 200;
+
+    static async Task Main(string[] args)
```
kirankumarkolli, line 25: Is this a conceptual possibility repro? 
Ideal is to repro with SDK code.

NaluTripician:
PR scope too large, fix inconsistent and not easily testable. Will create new PR with better testing framework

PR #5727, opened 2026-03-31
[Internal] DTS: Adds stream variants to DistributedWriteTransaction

Meghana-Palaparthi:
## Description

Adds `CreateItemStream`, `ReplaceItemStream`, `UpsertItemStream`, and `PatchItemStream` to `DistributedWriteTransaction` and `DistributedWriteTransactionCore`, bringing the API surface to parity with Transactional Batch

## Type of change

## Closing issues

### DIFF
--- a/.github/agents/release-copilot-agent.agent.md
+++ b/.github/agents/release-copilot-agent.agent.md
@@ -50,6 +50,7 @@ I need to add missed PRs to a release
 8. Create a release PR with API diff in the description
 9. Ensure that a full test suite passes, including contract enforcement tests.
 10. Guide through post-merge pipeline queuing and NuGet publish
+11. Update the `azure-docs-sdk-dotnet` metadata file with the new version (fork/sync, worktree clone, PR)
 
 ---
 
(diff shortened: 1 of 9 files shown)


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
```diff
@@ -44,6 +45,33 @@ public override DistributedWriteTransaction CreateItem<T>(
...
+            string collection,
+            PartitionKey partitionKey,
+            Stream streamPayload,
+            DistributedTransactionRequestOptions requestOptions = null)
+        {
+            DistributedWriteTransactionCore.ValidateContainerReference(database, collection);
+            if (streamPayload == null)
+            {
+                throw new ArgumentNullException(nameof(streamPayload));
+            }
```
Praveen-Msft, line 59: ValidateStreamPayload(..) similar to other validations?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
```diff
@@ -44,6 +45,33 @@ public override DistributedWriteTransaction CreateItem<T>(
...
+
+            this.operations.Add(
+                new DistributedTransactionOperation(
+                    operationType: OperationType.Create,
+                    operationIndex: this.operations.Count,
```
Praveen-Msft, line 64: by default the .Count is not threadsafe. Don't we need to employ some mechanism here to ensure the correct count?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
```diff
@@ -44,6 +45,33 @@ public override DistributedWriteTransaction CreateItem<T>(
...
+            PartitionKey partitionKey,
+            Stream streamPayload,
+            DistributedTransactionRequestOptions requestOptions = null)
+        {
+            DistributedWriteTransactionCore.ValidateContainerReference(database, collection);
+            if (streamPayload == null)
+            {
+                throw new ArgumentNullException(nameof(streamPayload));
+            }
+
```
Praveen-Msft, line 60: Also, validating the streamPayload's content, is it deferred to the client?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
Meghana-Palaparthi: This is a single-threaded builder API by design, consistent with TransactionalBatch.  `CosmosClient.CreateDistributedWriteTransaction()` returns a fresh instance per call so there's no shared state across callers. Adding synchronization here would add complexity without a real benefit given the intended usage model

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
Meghana-Palaparthi: Yes, the caller is responsible for providing a valid JSON stream. Any further stream content validation is deferred to the service. This is the same behavior as `TransactionalBatch` stream methods. SDK only guards against a null stream reference on the client side

PR #5732, opened 2026-04-03
fix: CosmosException when reading from change feed

### LINKED ISSUE #5707: CosmosException when reading from change feed
Getting CosmosException with following message: {"code":"InternalServerError","message":"Specified argument was out of the range of valid values. (Parameter 'Partition has more than two children which is not supported')"} while reading the change feed with AllVersionsAndDeletes enabled (using standard ChnageFeedProcessor).

Not sure if that's SDK or internal CosmosDb issue.

**Environment summary**
SDK Version: 3.58.0-preview.1
OS Version (Linux)

SillyZir:
## Summary
Added documentation for the `InternalServerError` (500) exception related to partition children limits when using `AllVersionsAndDeletes` change feed mode, and clarified the "owner resource not found" error description in `Exceptions.md`.

Fixes #5707

## Problem
Users encountering a `CosmosException` with message `"Partition has more than two children which is not supported"` when reading the change feed with `AllVersionsAndDeletes` enabled had no guidance in the SDK documentation on what causes this error or how to work around it (#5707).

## Solution
Updated `Exceptions.md` to add a new section documenting the `InternalServerError` (500) status code specific to the partition children limitation in `AllVersionsAndDeletes` change feed mode, including a workaround to switch to `LatestVersion` mode when full version tracking isn't required. Also simplified the "owner resource not found" description from the generic "(container or database)" phrasing to be more direct about the referenced resource.

## Impact
Users hitting this partition-related exception now have a documented explanation and actionable workaround directly in the exceptions reference, reducing confusion about whether this is an SDK bug or a known service limitation.

### DIFF
--- a/Exceptions.md
+++ b/Exceptions.md
@@ -44,4 +44,8 @@ When you receive a 404 (Not Found) status code from Cosmos DB, it can indicate t
 
 When using the `AllVersionsAndDeletes` change feed mode, you may encounter an `InternalServerError` with the message **"Partition has more than two children which is not supported"**. This error occurs when a partition split results in more than two child partitions, which is not supported by the `AllVersionsAndDeletes` change feed mode.
 
-**Workaround:** Switch to `LatestVersion` change feed mode if your scenario does not require all versions and deletes. If you require `AllVersionsAndDeletes`, retry the change feed request with a continuation token from before the partition split occurred, or re-initialize the change feed processor to pick up the new partition ranges.
+### Workaround
+
+Currently, there is no known workaround for this scenario when using AllVersionsAndDeletes mode.
+
+This behavior is being tracked, and a fix is expected from the CosmosDB team.
--- a/Exceptions.md
+++ b/Exceptions.md
@@ -42,7 +42,7 @@ When you receive a 404 (Not Found) status code from Cosmos DB, it can indicate t
 
 ### InternalServerError (500) - Partition has more than two children <a id="partition-split-change-feed"></a>
 
-When using the `AllVersionsAndDeletes` change feed mode, you may encounter an `InternalServerError` with the message **"Partition has more than two children which is not supported"**. This error occurs when a partition split results in more than two child partitions, which is not supported by the `AllVersionsAndDeletes` change feed mode.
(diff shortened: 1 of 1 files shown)


Exceptions.md
```diff
@@ -38,55 +38,10 @@ To see a list of common error code and issues please see [.NET SDK troubleshooti
...
-    else
-    {
-        // The item doesn't exist in an existing container
-        Console.WriteLine("Item not found");
-    }
-}
-```
-
-This distinction is particularly useful when implementing retry logic or error handling strategies, as you may want to handle these scenarios differently (e.g., creating the container if it doesn't exist vs. handling a missing item).
+**Workaround:** Switch to `LatestVersion` change feed mode if your scenario does not require all versions and deletes. If you require `AllVersionsAndDeletes`, retry the change feed request with a continuation token from before the partition split occurred, or re-initialize the change feed processor to pick up the new partition ranges.
```
kirankumarkolli: Currently there is no workaround for AVAD in-case such cases. There is a backlog we are tracking. Meanwhile CosmosDB is working on fix to not allow enabling AVAD on such accounts.

SillyZir: Thanks for the clarification — that makes sense.

I’ll remove the suggested workaround and update the documentation to reflect that there is currently no workaround for AVAD in this scenario, and that a fix is being tracked.

Exceptions.md
```diff
@@ -38,55 +38,14 @@ To see a list of common error code and issues please see [.NET SDK troubleshooti
...
-}
-```
-
-This distinction is particularly useful when implementing retry logic or error handling strategies, as you may want to handle these scenarios differently (e.g., creating the container if it doesn't exist vs. handling a missing item).
+2. **Owner resource not found**: The container or database referenced in the request doesn't exist
+
+### InternalServerError (500) - Partition has more than two children <a id="partition-split-change-feed"></a>
+
+When using the `AllVersionsAndDeletes` change feed mode, you may encounter an `InternalServerError` with the message **"Partition has more than two children which is not supported"**. This error occurs when a partition split results in more than two child partitions, which is not supported by the `AllVersionsAndDeletes` change feed mode.
+
```
Praveen-Msft, line 46: Another not supported scenario for `AllVersionAndDeletes`: when the partitions have undergone merges in the past. Can you please add it as well?

SillyZir: Good catch — added the merge scenario in the same paragraph. Should show up in the latest commit.

kushagraThapar:
@SillyZir this PR actually makes the docs worse than what they are right now, closing this.

PR #5741, opened 2026-04-07
Hedging Detection API: Adds IsHedged() and GetHedgedRegions() to CosmosDiagnostics

NaluTripician:
## Proposal

Users of the Azure Cosmos DB .NET SDK who enable cross-region hedging via `AvailabilityStrategy.CrossRegionHedgingStrategy()` have no performant, first-class API to determine whether a given response was produced by a hedged request. Today the only way to detect hedging is to call `Diagnostics.ToString()` and parse the resulting JSON for `\"Hedge Context\"` — a string-parsing approach that is fragile, allocates heavily, and is inappropriate for hot paths.

## What Changes

- **New virtual method `IsHedged()` on `CosmosDiagnostics`** — returns `bool` indicating whether the availability strategy activated hedging. Default returns `false` (back-compat).
- **New virtual method `GetHedgedRegions()` on `CosmosDiagnostics`** — returns `IReadOnlyList<string>` containing all regions that received hedge requests, in dispatch order. Default returns an empty list. This mirrors the `\"Hedge Context\"` datum already stored in diagnostics.
- **Internal boolean + list fields on `CosmosTraceDiagnostics`** — set directly by `CrossRegionHedgingAvailabilityStrategy`, true O(1) with zero allocations.

## Design Decisions

1. **Methods on `CosmosDiagnostics`** (not `ResponseMessage`) — follows established pattern (`GetClientElapsedTime`, `GetContactedRegions`, etc.)
2. **Direct fields** (not trace Data dictionary lookup) — avoids `SetWalkingStateRecursively()` overhead
3. **Virtual methods with defaults** (not abstract) — non-breaking change
4. **All hedged regions** (not just responding region) — gives users the full observability picture matching the diagnostics trace data

## Impact

- `CosmosDiagnostics.cs` — 2 new virtual methods
- `CosmosTraceDiagnostics.cs` — internal fields + overrides
- `CrossRegionHedgingAvailabilityStrategy.cs` — set flags on response

## OpenSpec Artifacts

Full design, specs, and task breakdown in `openspec/changes/hedging-detection-api/`


NaluTripician:
closing, new APIs being proposed in offline discussion

PR #5742, opened 2026-04-07
Availability Strategy: Fixes GetContactedRegions() to include all hedged regions

NaluTripician:
## Description

Fixes a bug where `CosmosDiagnostics.GetContactedRegions()` does not include regions targeted by cancelled hedged requests. When cross-region hedging dispatches requests to multiple regions and a winning response arrives before other regions respond, the losing requests are cancelled before `RecordResponse()` is called, so their regions are never recorded.

### Root Cause

**Primary issue:** Contacted regions are only recorded when `RecordResponse()` / `RecordHttpResponse()` / `RecordHttpException()` is called on response receipt. Cancelled hedged requests never receive a response, so their target region is never added to `TraceSummary`.

**Secondary issue:** `ClientSideRequestStatisticsTraceDatum.RecordResponse()` calls `TraceSummary.AddRegionContacted()` directly but does not populate its own `RegionsContacted` HashSet. This makes the `UpdateRegionContacted()` safety-net call in `TransportHandler.ProcessMessageAsync()` a no-op (it returns early when `RegionsContacted.Count == 0`).

### Fix

1. **Record regions at dispatch time** in `CloneAndSendAsync()`: adds `trace.Summary.AddRegionContacted(targetRegion, targetEndpoint)` when the hedged request is dispatched, ensuring the region is tracked even if the request is cancelled.

2. **Populate `CSRSD.RegionsContacted`** in `RecordResponse`, `RecordHttpResponse`, and `RecordHttpException` so the `TransportHandler` safety-net works correctly.

### Testing

Added 3 new unit tests:
- `CancelledHedgeRequest_RegionStillAppearsInContactedRegions` — verifies cancelled primary region appears
- `ThreeRegionHedging_AllRegionsAppearInContactedRegions` — verifies all 3 regions when last wins
- `PrimaryWinsBeforeHedge_OnlyOneRegionInContactedRegions` — verifies only 1 region when no hedge triggers

All 16 `AvailabilityStrategyUnitTests` and all 3 `ContactedRegionsTests` pass.

### Impact

| Affected API | Impact |
|---|---|
| `CosmosDiagnostics.GetContactedRegions()` | Now includes cancelled hedge regions |
| `ClientTelemetryHelper.GetContactedRegions()` | Telemetry correctly reports all contacted regions |
| `CosmosDbMeterUtil.GetRegions()` | OTel metrics include hedge regions |


Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
```diff
@@ -266,6 +266,7 @@ public void RecordResponse(
             {
                 if (locationEndpoint != null)
                 {
+                    this.RegionsContacted.Add((regionName, locationEndpoint));
```
kirankumarkolli: Please expand on why explictly its needed? 
Below TraceSummary.AddRegionContacted do capture it inside the TraceSummary and is the current pattern right?

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -148,10 +149,16 @@ internal override async Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
...
                 {
+                    bool isReadRequest = OperationTypeExtensions.IsReadOperation(request.OperationType);
+
                     IReadOnlyCollection<string> hedgeRegions = client.DocumentClient.GlobalEndpointManager
                         .GetApplicableRegions(
                             request.RequestOptions?.ExcludeRegions,
-                            OperationTypeExtensions.IsReadOperation(request.OperationType));
+                            isReadRequest);
+
+                    ReadOnlyDictionary<string, Uri> endpointsByRegion = isReadRequest
```
kirankumarkolli: Remote possible race between the two calls the data can be inconsistent.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -314,6 +323,15 @@ private async Task<HedgingResponse> CloneAndSendAsync(
             {
                 clonedRequest.RequestOptions ??= new RequestOptions();
 
+                string targetRegion = hedgeRegions.ElementAt(requestNumber);
```
kirankumarkolli, line 325: It's possible that diagnostics will show regions which might not have corresponding NW interactions. Just for Hedging this seems like a big side-affect. 

Is the requirement to programmatically know if hedging happened or not? (even just that hint also conficts with the response diagnostics).

Unsure if we have a better option.

NaluTripician (commented):
**Re: CSRSD.RegionsContacted additions (ClientSideRequestStatisticsTraceDatum.cs)**

You're right — \TraceSummary.AddRegionContacted\ is the correct and sufficient pattern. The \TraceSummary\ is already thread-safe (uses \lock\ on its internal \HashSet\) and is the canonical source of contacted regions.

The \CSRSD.RegionsContacted\ additions were intended to make the \TransportHandler.UpdateRegionContacted()\ safety-net work, but since \TraceSummary.AddRegionContacted\ is already called in the same code path, the safety-net's \UnionWith\ adds nothing new. Additionally, \CSRSD.RegionsContacted\ is a plain \HashSet\ without comprehensive lock protection across all callers — so populating it from multiple methods is a thread-safety concern.

**Reverted** all three \	his.RegionsContacted.Add(...)\ lines in the latest push.

NaluTripician (commented):
**Re: Race between GetApplicableRegions and endpoint dictionary pre-fetch (CrossRegionHedgingAvailabilityStrategy.cs:159)**

Good catch. The pre-fetched \ndpointsByRegion\ dictionary was a separate call from \GetApplicableRegions()\, creating a window where the endpoint data could change between the two lookups.

**Fixed** by removing the pre-fetched dictionary entirely. The endpoint is now resolved lazily inside \CloneAndSendAsync\ via \GlobalEndpointManager.GetAvailableReadEndpointsByLocation()\ (or write equivalent) at the exact point where it's needed. This eliminates the temporal gap and ensures region-to-endpoint consistency.

NaluTripician (commented):
**Re: Regions without NW interactions / diagnostics side-effect (CrossRegionHedgingAvailabilityStrategy.cs:326)**

Valid concern. Recording at dispatch-time (before \RequestSenderAndResultCheckAsync\) meant a region could appear in diagnostics even if the request hadn't actually been handed off to the network layer yet.

**Fixed** by moving the region recording into a \inally\ block *after* \RequestSenderAndResultCheckAsync\. This ensures:

1. The region is only recorded after the request has been dispatched to the sender (so there IS a corresponding NW interaction — either completed, failed, or cancelled in-flight)
2. The \inally\ block still runs on cancellation (\RequestSenderAndResultCheckAsync\ catches \OperationCanceledException\ and wraps it), so cancelled hedge requests still get their regions tracked
3. \TraceSummary.AddRegionContacted\ uses a \HashSet\ internally, so duplicate recordings from both this path and the normal \RecordResponse\ path are deduplicated

The endpoint is also now resolved at recording time (not pre-fetched), which addresses the race concern from the other comment.

NaluTripician:
Closing based on offline discussion

PR #5777, opened 2026-04-10
﻿DocumentClient: Fixes missing Dispose of GlobalPartitionEndpointManagerCore causing timer leak

pcwiese:
## Description

DocumentClient.Dispose() disposes GlobalEndpointManager, HttpClient, StoreModel, AddressResolver, and other owned resources, but never disposes PartitionKeyRangeLocation (GlobalPartitionEndpointManagerCore). This leaves the circuit breaker failback loop running indefinitely, leaking Task.Delay timers.

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -1405,6 +1405,12 @@ namespace Microsoft.Azure.Cosmos
                 this.cosmosAuthorization.Dispose();
             }
 
+            if (this.PartitionKeyRangeLocation is IDisposable partitionKeyRangeLocationDisposable)
+            {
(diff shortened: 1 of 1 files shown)


Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -1405,6 +1405,12 @@ public void Dispose()
                 this.cosmosAuthorization.Dispose();
             }
 
+            if (this.PartitionKeyRangeLocation is IDisposable partitionKeyRangeLocationDisposable)
```
kirankumarkolli, line 1408: Thank you for the proposed fix.

kirankumarkolli:
Fixed part of #5778 with tests

PR #5783, opened 2026-04-14
[Preview] Container: Fixes SemanticRerankAsync TypeLoadException in derived classes

NaluTripician:
## Problem
`System.TypeLoadException`: Method `SemanticRerankAsync` in type `Microsoft.Azure.Cosmos.Encryption.Custom.EncryptionContainer` does not have an implementation.

## Root Cause
`SemanticRerankAsync` was declared as `abstract` in the `Container` base class (under `#if PREVIEW`). Derived classes like `EncryptionContainer` in the Encryption.Custom project only override this method when both `PREVIEW && SDKPROJECTREF` preprocessor symbols are defined. When `PREVIEW` is not defined in the consuming project but the NuGet preview package includes the abstract method, the derived class is missing the required implementation, causing a `TypeLoadException` at runtime.

## Fix
Changed `SemanticRerankAsync` from `abstract` to `virtual` with a default implementation that throws `NotImplementedException`. This follows the existing pattern used by other methods in `Container` (e.g., `DeleteAllItemsByPartitionKeyStreamAsync`) and prevents the TypeLoadException in derived classes that don't override the method.


kirankumarkolli (dismissed):

kirankumarkolli:
@juraj-blazek FYI

kirankumarkolli (approved):

PR #5785, opened 2026-04-15
[DO NOT MERGE] PartitionKey: Adds id to partition key when '/id' is the last path in HPK definition

NaluTripician:
# Pull Request Template

## Description

> **This PR mirrors [#5600](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5600) by @ramarag** to allow CI gates to run (fork PRs cannot access pipeline secrets/agents). All code is authored by @ramarag — previous reviews on #5600 still apply.

The current change ensures that SDK users don't need to repeat themselves when "/id" is the last part of HPK.

With a traditional single-level partition key—for example, "/pk"—a given partition key value is constrained to a single physical partition. Because of this:

- Throughput (RU/s) for that partition key cannot exceed the limit of a single partition.
- Storage available to that partition key is also bounded by that single partition's storage capacity.
- As a result, a "hot" partition key can become a scalability bottleneck.

By using Hierarchical Partition Key (HPK) with "/id" as the last level path, 1 partition key essentially maps to a single document. "/pk" from above becomes a prefix and uses the prefix-based functionality of query in HPK. The partition key "/pk" now becomes boundless and can span multiple partitions.

### Cons:
The following features can work only on one document, largely reducing their efficacy:
1. Atomic Batch
2. Stored Procedure
3. DeleteAllDocuments by Partition Key

## Type of change

- New feature (non-breaking change which adds functionality)

## Original PR

- Original fork PR: #5600 by @ramarag
- Code author: @ramarag
- All previous review feedback on #5600 has been addressed


NaluTripician:
Bringing changes to original branch

PR #5786, opened 2026-04-15
Region Availability: Adds missing regions from Direct 3.42.4

NaluTripician:
## Description

Adds 3 regions introduced in Direct SDK 3.42.3/3.42.4 that were missing from `Cosmos.Regions` and the API contract baseline:

- `SoutheastAsia3` → `Southeast Asia 3`
- `SaudiArabiaEast` → `Saudi Arabia East`
- `WestCentralUSFRE` → `West Central US FRE`

These were added to `LocationNames` in the Direct package bump (#5779) but the corresponding `Regions.cs` entries and contract baseline were not updated, causing `MappedRegionsTest` to fail.

## Type of change

## Does this introduce a breaking change?

No. Additive-only: 3 new public `const string` fields.

### DIFF
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetSDKAPI.net6.json
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetSDKAPI.net6.json
@@ -8193,6 +8193,11 @@
           "Attributes": [],
           "MethodInfo": "System.String QatarCentral;IsInitOnly:False;IsStatic:True;"
         },
+        "System.String SaudiArabiaEast": {
+          "Type": "Field",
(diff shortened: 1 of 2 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetSDKAPI.net6.json
```diff
@@ -8193,6 +8193,11 @@
           "Attributes": [],
           "MethodInfo": "System.String QatarCentral;IsInitOnly:False;IsStatic:True;"
         },
+        "System.String SaudiArabiaEast": {
```
kirankumarkolli, line 8196: Preview contracts also might need this change

kirankumarkolli (approved):

PR #5788, opened 2026-04-15
[Socket Handler]: Adds HTTP/2 PING keep-alive to detect broken connections in pool.

aavasthy:
# Pull Request Template

## Description
Configures HTTP/2 PING keep-alive on the SDK-managed SocketsHttpHandler to detect and evict broken connections from the pool.

Without this, a broken HTTP/2 connection can linger in the pool indefinitely, causing persistent 500 errors that only resolve after application restart.
This was observed across multiple Go SDK customers using HTTP/2 and applies to the .NET SDK in thin client mode which explicitly uses HTTP/2.

Settings added:
KeepAlivePingDelay     │ 30s    │ Send PING after 30s of inactivity
KeepAlivePingTimeout   │ 20s    │ Mark connection dead if no PONG within 20s
KeepAlivePingPolicy    │ Always │ Ping idle connections too — critical for detecting broken pooled connections  

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
@@ -187,13 +187,22 @@ namespace Microsoft.Azure.Cosmos
             // Without this, a broken HTTP/2 connection (e.g. after a network blip or load balancer
             // reset) can remain in the pool indefinitely, causing persistent request failures
             // that only resolve after application restart.
+            // KeepAlivePingDelay/Timeout/Policy are available on SocketsHttpHandler in .NET 5.0+.
             try
             {
+                int pingDelayInSeconds = ConfigurationManager.GetEnvironmentVariable<int>(
+                    ConfigurationManager.Http2KeepAlivePingDelayInSeconds,
+                    defaultValue: 1);
+
+                int pingTimeoutInSeconds = ConfigurationManager.GetEnvironmentVariable<int>(
+                    ConfigurationManager.Http2KeepAlivePingTimeoutInSeconds,
+                    defaultValue: 2);
+
                 PropertyInfo keepAlivePingDelayInfo = socketHandlerType.GetProperty("KeepAlivePingDelay");
-                keepAlivePingDelayInfo?.SetValue(socketHttpHandler, TimeSpan.FromSeconds(30));
+                keepAlivePingDelayInfo?.SetValue(socketHttpHandler, TimeSpan.FromSeconds(pingDelayInSeconds));
 
                 PropertyInfo keepAlivePingTimeoutInfo = socketHandlerType.GetProperty("KeepAlivePingTimeout");
-                keepAlivePingTimeoutInfo?.SetValue(socketHttpHandler, TimeSpan.FromSeconds(20));
+                keepAlivePingTimeoutInfo?.SetValue(socketHttpHandler, TimeSpan.FromSeconds(pingTimeoutInSeconds));
 
                 // HttpKeepAlivePingPolicy.Always = 1: send pings even for idle connections,
                 // which is critical for detecting broken connections lingering in the pool.
--- a/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
@@ -183,6 +183,33 @@ namespace Microsoft.Azure.Cosmos
                 DefaultTrace.TraceWarning("Failed to set EnableMultipleHttp2Connections on SocketsHttpHandler: {0}", ex.Message);
             }
 
+            // Enable HTTP/2 PING keep-alive to detect broken connections.
+            // Without this, a broken HTTP/2 connection (e.g. after a network blip or load balancer
+            // reset) can remain in the pool indefinitely, causing persistent request failures
+            // that only resolve after application restart.
+            try
+            {
+                PropertyInfo keepAlivePingDelayInfo = socketHandlerType.GetProperty("KeepAlivePingDelay");
+                keepAlivePingDelayInfo?.SetValue(socketHttpHandler, TimeSpan.FromSeconds(30));
+
+                PropertyInfo keepAlivePingTimeoutInfo = socketHandlerType.GetProperty("KeepAlivePingTimeout");
+                keepAlivePingTimeoutInfo?.SetValue(socketHttpHandler, TimeSpan.FromSeconds(20));
+
+                // HttpKeepAlivePingPolicy.Always = 1: send pings even for idle connections,
+                // which is critical for detecting broken connections lingering in the pool.
+                PropertyInfo keepAlivePingPolicyInfo = socketHandlerType.GetProperty("KeepAlivePingPolicy");
+                if (keepAlivePingPolicyInfo != null)
+                {
+                    Type pingPolicyType = keepAlivePingPolicyInfo.PropertyType;
+                    object alwaysValue = Enum.ToObject(pingPolicyType, 1);
+                    keepAlivePingPolicyInfo.SetValue(socketHttpHandler, alwaysValue);
+                }
+            }
+            catch (Exception ex)
+            {
+                DefaultTrace.TraceWarning("Failed to configure HTTP/2 keep-alive ping on SocketsHttpHandler: {0}", ex.Message);
+            }
+
             if (serverCertificateCustomValidationCallback != null)
             {
                 //Get SslOptions Property
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
@@ -467,8 +467,8 @@ namespace Microsoft.Azure.Cosmos.Tests
             Assert.IsTrue(socketsHandler.EnableMultipleHttp2Connections, "EnableMultipleHttp2Connections should be true for HTTP/2 thin client support");
 
             // HTTP/2 PING keep-alive: detects broken connections lingering in the pool
-            Assert.AreEqual(TimeSpan.FromSeconds(30), socketsHandler.KeepAlivePingDelay, "KeepAlivePingDelay should be 30 seconds for HTTP/2 connection health monitoring");
-            Assert.AreEqual(TimeSpan.FromSeconds(20), socketsHandler.KeepAlivePingTimeout, "KeepAlivePingTimeout should be 20 seconds");
+            Assert.AreEqual(TimeSpan.FromSeconds(1), socketsHandler.KeepAlivePingDelay, "KeepAlivePingDelay should be 1 second for HTTP/2 connection health monitoring");
+            Assert.AreEqual(TimeSpan.FromSeconds(2), socketsHandler.KeepAlivePingTimeout, "KeepAlivePingTimeout should be 2 seconds");
             Assert.AreEqual(HttpKeepAlivePingPolicy.Always, socketsHandler.KeepAlivePingPolicy, "KeepAlivePingPolicy should be Always to detect broken idle connections");
 
             //Create cert for test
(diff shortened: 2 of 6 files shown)


Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -183,6 +183,33 @@ public static HttpMessageHandler CreateSocketsHttpHandlerHelper(
...
+            // Without this, a broken HTTP/2 connection (e.g. after a network blip or load balancer
+            // reset) can remain in the pool indefinitely, causing persistent request failures
+            // that only resolve after application restart.
+            try
+            {
+                PropertyInfo keepAlivePingDelayInfo = socketHandlerType.GetProperty("KeepAlivePingDelay");
+                keepAlivePingDelayInfo?.SetValue(socketHttpHandler, TimeSpan.FromSeconds(30));
+
+                PropertyInfo keepAlivePingTimeoutInfo = socketHandlerType.GetProperty("KeepAlivePingTimeout");
+                keepAlivePingTimeoutInfo?.SetValue(socketHttpHandler, TimeSpan.FromSeconds(20));
```
kirankumarkolli: How are the values determined?

kirankumarkolli: HOw about customization through ENV variables?

xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
FabianMeiswinkel: Yeah - definitely do not hard-code - and to provide value they will likely have to be more aggressive

xinlian12: **Cross-SDK data to support this thread:**

The Go SDK (which motivated this PR  "observed across multiple Go SDK customers") already has this fix since October 2023 ([PR #21468](https://github.com/Azure/azure-sdk-for-go/pull/21468)) with **significantly more aggressive** values:

| Setting | .NET (this PR) | Go SDK (production since 2023) |
|---------|---------------|-------------------------------|
| Ping delay | 30s | 10s |
| Ping timeout | 20s | 5s |
| **Total detection time** | **~50s** | **~15s** |

That's 3.3 slower broken-connection detection in .NET. During that 50s window, every request routed to the broken connection fails.

The Go values were chosen after real customer impact analysis ([golang/go#59690](https://github.com/golang/go/issues/59690)) and have been stable in production for 2.5 years. If the goal is to match the customer-facing behavior that resolved the Go SDK issue, aligning closer to those values (or making them ENV-configurable as @kirankumarkolli suggested) would be prudent.

---
 AI-generated review  may be incorrect. Agree?  resolve the conversation. Disagree?  reply with your reasoning.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -183,6 +183,33 @@ public static HttpMessageHandler CreateSocketsHttpHandlerHelper(
...
+
+                // HttpKeepAlivePingPolicy.Always = 1: send pings even for idle connections,
+                // which is critical for detecting broken connections lingering in the pool.
+                PropertyInfo keepAlivePingPolicyInfo = socketHandlerType.GetProperty("KeepAlivePingPolicy");
+                if (keepAlivePingPolicyInfo != null)
+                {
+                    Type pingPolicyType = keepAlivePingPolicyInfo.PropertyType;
+                    object alwaysValue = Enum.ToObject(pingPolicyType, 1);
```
xinlian12, line 204: 🟢 **Suggestion — Resilience: Prefer `Enum.Parse` by name over `Enum.ToObject` by ordinal**

```csharp
object alwaysValue = Enum.ToObject(pingPolicyType, 1);

```

The hardcoded `1` is correct today (`HttpKeepAlivePingPolicy.Always = 1`), but it's an ordinal assumption with no compile-time validation. If Microsoft ever inserted a new enum member before `Always`, this would silently apply the wrong policy.

Consider using name-based lookup instead:

```csharp
object alwaysValue = Enum.Parse(pingPolicyType, "Always");

```

This is self-documenting, resilient to ordinal changes, and equally correct. The performance difference is irrelevant — this runs once at client startup. `Enum.Parse` will throw `ArgumentException` if the name changes (surfacing the problem immediately), whereas `Enum.ToObject` would silently apply the wrong policy.

---

xinlian12:
✅ **Review complete** (23:26)

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
aavasthy: Added environment variables for setting this value and referred the rust sdk config for default values
https://github.com/Azure/azure-sdk-for-rust/blob/955b4c4803f6a96f082f3ecd394fd4ae58bc8cd6/sdk/cosmos/azure_data_cosmos_driver/src/options/connection_pool.rs#L254-L255

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -183,6 +183,42 @@ public static HttpMessageHandler CreateSocketsHttpHandlerHelper(
...
+            // Enable HTTP/2 PING keep-alive to detect broken connections.
+            // Without this, a broken HTTP/2 connection (e.g. after a network blip or load balancer
+            // reset) can remain in the pool indefinitely, causing persistent request failures
+            // that only resolve after application restart.
+            // KeepAlivePingDelay/Timeout/Policy are available on SocketsHttpHandler in .NET 5.0+.
+            try
+            {
+                int pingDelayInSeconds = ConfigurationManager.GetEnvironmentVariable<int>(
+                    ConfigurationManager.Http2KeepAlivePingDelayInSeconds,
+                    defaultValue: 1);
```
kirankumarkolli, line 195: Defaults seems very aggressive. 
Lets align with eventual RUST expected values.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -495,6 +500,46 @@ public void CreateHttpClientHandlerCreatesCorrectValueType()
             Assert.IsFalse(clientHandler.ServerCertificateCustomValidationCallback.Invoke(new HttpRequestMessage(), x509Certificate2, x509Chain, sslPolicyErrors));
         }
 
+        [TestMethod]
```
kirankumarkolli, line 503: Possible concurrency issue with other tests?

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
kirankumarkolli: Seems like there are the RUST defaults.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -183,6 +183,42 @@ public static HttpMessageHandler CreateSocketsHttpHandlerHelper(
...
             }
 
+            // Enable HTTP/2 PING keep-alive to detect broken connections.
+            // Without this, a broken HTTP/2 connection (e.g. after a network blip or load balancer
+            // reset) can remain in the pool indefinitely, causing persistent request failures
+            // that only resolve after application restart.
+            // KeepAlivePingDelay/Timeout/Policy are available on SocketsHttpHandler in .NET 5.0+.
+            try
+            {
+                int pingDelayInSeconds = ConfigurationManager.GetEnvironmentVariable<int>(
```
xinlian12, line 193: 🟡 **Recommendation — Resilience: Add `Math.Max` clamping on env var overrides to prevent silent feature disablement**

```csharp
int pingDelayInSeconds = ConfigurationManager.GetEnvironmentVariable<int>(
    ConfigurationManager.Http2KeepAlivePingDelayInSeconds,
    defaultValue: 1);

int pingTimeoutInSeconds = ConfigurationManager.GetEnvironmentVariable<int>(
    ConfigurationManager.Http2KeepAlivePingTimeoutInSeconds,
    defaultValue: 2);

```

Both `KeepAlivePingDelay` and `KeepAlivePingTimeout` on `SocketsHttpHandler` require values strictly greater than `TimeSpan.Zero` (the setter throws `ArgumentOutOfRangeException` for `<= TimeSpan.Zero`). If a user sets `AZURE_COSMOS_HTTP2_KEEPALIVE_PING_DELAY_IN_SECONDS=0` or a negative value:

1. `GetEnvironmentVariable<int>` parses successfully (valid int)
2. `TimeSpan.FromSeconds(0)` → `TimeSpan.Zero`
3. `SetValue` throws `ArgumentOutOfRangeException`
4. Caught by `catch (Exception ex)` → **all three settings skipped** (delay, timeout, AND policy)
5. Handler has .NET defaults: `KeepAlivePingDelay = Timeout.InfiniteTimeSpan` → **pings disabled entirely**

The feature this PR adds is silently defeated with only a trace warning. This is especially subtle because the delay failure also prevents the `KeepAlivePingPolicy = Always` from being set — so even if only the timeout env var is invalid, the policy reverts to `WithActiveRequests` (which doesn't ping idle connections, defeating the purpose).

The codebase already has an established pattern for this in `ConfigurationManager.cs:159-166`:

```csharp
// Existing pattern
return Math.Max(
    ConfigurationManager.GetEnvironmentVariable(..., defaultValue: ...),
    MinMaxRetriesInLocalRegionWhenRemoteRegionPreferred);

```

**Suggested fix:**

```csharp
int pingDelayInSeconds = Math.Max(1, ConfigurationManager.GetEnvironmentVariable<int>(
    ConfigurationManager.Http2KeepAlivePingDelayInSeconds,
    defaultValue: 1));

int pingTimeoutInSeconds = Math.Max(1, ConfigurationManager.GetEnvironmentVariable<int>(
    ConfigurationManager.Http2KeepAlivePingTimeoutInSeconds,
    defaultValue: 2));

```

This guarantees values are always ≥ 1 second, which is the minimum `SocketsHttpHandler` accepts, and follows the existing clamping convention.

xinlian12:
✅ **Review complete** (27:01)

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
aavasthy: The set defaults are what have been set in Rust as well https://github.com/Azure/azure-sdk-for-rust/blob/955b4c4803f6a96f082f3ecd394fd4ae58bc8cd6/sdk/cosmos/azure_data_cosmos_driver/src/options/connection_pool.rs#L254-L255

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -183,6 +183,42 @@ public static HttpMessageHandler CreateSocketsHttpHandlerHelper(
                 DefaultTrace.TraceWarning("Failed to set EnableMultipleHttp2Connections on SocketsHttpHandler: {0}", ex.Message);
             }
 
+            // Enable HTTP/2 PING keep-alive to detect broken connections.
+            // Without this, a broken HTTP/2 connection (e.g. after a network blip or load balancer
+            // reset) can remain in the pool indefinitely, causing persistent request failures
+            // that only resolve after application restart.
+            // KeepAlivePingDelay/Timeout/Policy are available on SocketsHttpHandler in .NET 5.0+.
+            try
```
xinlian12, line 191: 🟡 **Recommendation — Resilience: Separate try-catch blocks for independent property sets**

The single `try/catch` block (lines 191–220) groups env var parsing, `KeepAlivePingDelay`, `KeepAlivePingTimeout`, **and** `KeepAlivePingPolicy` together. If any one operation fails — including a `FormatException` from a non-numeric env var string, or a `TargetInvocationException` from the delay setter — the entire block is skipped, including the critical `KeepAlivePingPolicy = Always`.

The sibling `EnableMultipleHttp2Connections` just above (lines 176–184) uses its own **isolated** try-catch, which is the established pattern in this method. Each logical property set is independent and should fail independently.

**Concrete failure scenario:** A user sets `AZURE_COSMOS_HTTP2_KEEPALIVE_PING_DELAY_IN_SECONDS=fast` (non-numeric). `Convert.ChangeType` throws before any property is set → the catch fires → delay, timeout, AND policy are all skipped → the handler has .NET defaults (`KeepAlivePingDelay = Timeout.InfiniteTimeSpan`, `KeepAlivePingPolicy = WithActiveRequests`) → PING keep-alive is completely disabled. A bad delay value shouldn't prevent the timeout and policy from being configured with their defaults.

**Suggested structure:**

```csharp
// Block 1: Parse env vars (with defaults)
int pingDelayInSeconds = /* ... defaultValue: 1 */;
int pingTimeoutInSeconds = /* ... defaultValue: 2 */;

// Block 2: Set delay (independent)
try { keepAlivePingDelayInfo?.SetValue(...); }
catch (Exception ex) { DefaultTrace.TraceWarning(...); }

// Block 3: Set timeout (independent)
try { keepAlivePingTimeoutInfo?.SetValue(...); }
catch (Exception ex) { DefaultTrace.TraceWarning(...); }

// Block 4: Set policy (independent)
try { /* Enum.ToObject + SetValue */ }
catch (Exception ex) { DefaultTrace.TraceWarning(...); }

```

This way, a failure in one property still allows the others to be applied — all partial states are safe since the .NET defaults are reasonable fallbacks.

xinlian12:
✅ **Review complete** (30:35)

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
kirankumarkolli, line 195: Define static const and use them here for visibility

kirankumarkolli (approved):

PR #5792, opened 2026-04-19
[Internal] PPAF: Fixes Hub region header never sent due to premature NoRetry on single-master 404/1002.

aavasthy:
# Pull Request Template

## Description

# PR Description

## Summary

Fixes a bug in `ClientRetryPolicy.ShouldRetryOnSessionNotAvailable` where the hub region processing header (`x-ms-cosmos-hub-region-processing-only`) was set but **never actually sent** to the backend on single-master accounts. The retry threshold `sessionTokenRetryCount > 1` caused `NoRetry` immediately after the header flag was set, preventing the retry that would carry the header.

## Fix

Changed `sessionTokenRetryCount > 1` → `sessionTokenRetryCount > 2`.

### Corrected Flow

| Attempt | `count` before method | Guard sets header? | `count++` | Check `count > 2` | Result | Header on request? |
|---------|----------------------|-------------------|-----------|-------------------|--------|--------------------|
| 1st request | — | — | — | — | Sent to preferred read region | ❌ No |
| 1st 404/1002 | 0 | No (`0 >= 1` = false) | → 1 | `1 > 2` = false | ✅ Retry to write region (index=0) | ❌ No |
| 2nd 404/1002 | 1 | **Yes** (`1 >= 1` = true) | → 2 | `2 > 2` = false | ✅ Retry to write region (index=0) | ✅ **Yes** |
| 3rd 404/1002 | 2 | Yes (already true) | → 3 | `3 > 2` = true | ⛔ NoRetry — hub confirmed not available | ✅ Yes |

### End-to-End Hub Region Discovery (Happy Path with 403/3)

| Step | Request | Response | Hub Header? | What happens |
|------|---------|----------|-------------|-------------|
| 1 | Read to preferred region | 404/1002 | ❌ | Session not available → retry to write region |
| 2 | Read to write region | 404/1002 | ❌ | Write region also returns session not available → header flag set → retry |
| 3 | Read to write region | 403/3 | ✅ | Non-hub region rejects hub-only request → PPAF marks endpoint unavailable → retry to next region |
| 4 | Read to next region | 200 OK | ✅ | Hub region found, request succeeds, hub cached for partition |

## Type of change

## Closing issues


Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -456,10 +457,13 @@ private ShouldRetryResult ShouldRetryOnSessionNotAvailable(DocumentServiceReques
                 }
                 else
                 {
-                    if (this.sessionTokenRetryCount > 1)
+                    if (this.sessionTokenRetryCount > MaxSessionTokenRetryCount)
```
kundadebdatta, line 461: From line `336` - `341`, can we defer the logic of setting the boolean flag into this method `ShouldRetryOnSessionNotAvailable` itself ?

```
                // Only set the hub region processing header for single master accounts
                // Set header only after the first retry attempt fails with 404/1002
                if (!this.canUseMultipleWriteLocations && this.sessionTokenRetryCount >= 1)
                {
                    this.addHubRegionProcessingOnlyHeader = true;
                }
```

aavasthy: updated

kundadebdatta (dismissed):
LGTM.

Praveen-Msft (dismissed):

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -456,10 +449,22 @@ private ShouldRetryResult ShouldRetryOnSessionNotAvailable(DocumentServiceReques
                 }
                 else
                 {
-                    if (this.sessionTokenRetryCount > 1)
+#if !INTERNAL
+                    // Only set the hub region processing header for single master accounts.
+                    // Set header after the second consecutive 404/1002 (count >= 2 means both
+                    // the initial request and the first retry to the write region have failed).
+                    if (this.sessionTokenRetryCount >= MaxSessionTokenRetryCount)
```
kushagraThapar, line 456: I am curious, does having this `if condition` inside `INTERNAL` cause any issues?
Since half of the conditioning is internal and half is outside on line 461. Can we please double check the actual conditions of this code path? would it make sense to have both of the conditions inside INTERNAL as well as outside INTERNAL?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -456,10 +449,22 @@ private ShouldRetryResult ShouldRetryOnSessionNotAvailable(DocumentServiceReques
...
+                    {
+                        this.addHubRegionProcessingOnlyHeader = true;
```
kushagraThapar, line 458: would this instance be visible to the INTERNAL builds?

kushagraThapar (commented):
Since this is specifically fixed for single master, I wonder if we can add some tests for multi-master account? making sure the retries are same in multi-master and are correct. 
@jeet1995 can you please review this PR as well, thanks!

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
jeet1995: Conceptually, `x-ms-hub-region-processing-only` pinning is isolated to the client SDK, so any x-region retry behavior change based on this header should also be isolated to the client SDK (such as the count gate change).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -528,7 +501,100 @@ public async Task ClientRetryPolicy_HubRegionHeader_AddedOn404_1002_BasedOnAccou
...
             }
-        }
+        }
+
+        /// <summary>
+        /// End-to-end test for the hub region discovery flow on a single-master account (Direct mode):
+        /// 1st request → 404/1002 (no hub header) → retry to write region
+        /// 2nd request → 404/1002 (no hub header) → hub header flag set, retry
+        /// 3rd request → assert hub header present → 403/3 from non-hub → retry
+        /// 4th request → assert hub header present → 200 success
```
jeet1995, line 511: Can we add a test to verify header persistence when 3rd request onwards we hit a 503/429 (especially the other retriable error codes such as 503s/408s/(410s?))? What is the x-region retry behavior when these are hit say for the 3rd request onwards (do we revert to 503-based preferred region cycle through with the header set)?

jeet1995: Also, do we have the capability to carry over the header to hedged requests too when primary request has header pinned?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: AddHubRegionProcessingOnlyHeader is declared inside #if !INTERNAL, so is completely stripped from INTERNAL builds. INTERNAL builds won't see the field, the flag-setting, or the header-writing code.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: Updated this and now added with #if !INTERNAL / #else:
  
  - !INTERNAL build: Sets the hub header flag at count >= 2, then allows one more retry (count > 2 → NoRetry). This gives the request a chance to be sent with the hub header.
  - #else (INTERNAL) build: Preserves the original count > 1 → NoRetry behavior unchanged since there's no hub header concept.
 
 The addHubRegionProcessingOnlyHeader field itself is declared inside #if !INTERNAL, so it doesn't exist in INTERNAL builds at all.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
aavasthy: - Added ClientRetryPolicy_HubRegionHeader_PersistsThroughRetriableErrors — this test:
  1. Triggers two 404/1002 to set the hub header flag
  2. Verifies the hub header is present on the 3rd request
  3. Simulates a 503 ServiceUnavailable on the 3rd request
  4. Verifies the 503 triggers a retry via ShouldRetryOnServiceUnavailable (separate code path)
  5. Verifies the hub header still persists on the 4th request
 
- Hedging creates a separate DocumentServiceRequest with its own headers, so the hub header on the primary request wouldn't automatically propagate to a hedged request.

aavasthy:
> Since this is specifically fixed for single master, I wonder if we can add some tests for multi-master account? making sure the retries are same in multi-master and are correct. @jeet1995 can you please review this PR as well, thanks!

We have multi-master tests already added ensuring that this header is not there 
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/f8f93e55c65a435fec2de7b7a566ecdca8d80dee/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs?plain=1#L415C23-L415C23

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kushagraThapar: Thanks @aavasthy

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kushagraThapar: Thanks @aavasthy

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
jeet1995: > Hedging creates a separate DocumentServiceRequest with its own headers, so the hub header on the primary request wouldn't automatically propagate to a hedged request. 

- I'd track this then - there could be a correctness risk where primary request is going through its retries but hedged region returned a success without the header - thoughts @kundadebdatta , @FabianMeiswinkel

kundadebdatta: Yes - completely agreed. This needed to be in sync with the original request and te hedged ones, and better to do it in this PR itself.

How is it handled in the Java toda @jeet1995 ? Do we just set the header in the hedged request, whenever it's present in the original one and keep the decision making to the retry policy ?

aavasthy: We will cover hedging part in this https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5824

NaluTripician:
@sdkReviewAgent-2

jeet1995 (approved):

kundadebdatta (approved):
LGTM Now.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -456,12 +449,29 @@ private ShouldRetryResult ShouldRetryOnSessionNotAvailable(DocumentServiceReques
...
+                    }
+
+                    if (this.sessionTokenRetryCount > MaxSessionTokenRetryCount)
```
xinlian12, line 461: 🟢 **Maintainability: Shared `else` across preprocessor boundaries**

The `else` on line 475 binds to different `if` statements depending on which preprocessor path is active:

- In `!INTERNAL` builds: binds to `if (sessionTokenRetryCount > MaxSessionTokenRetryCount)` (line 461)
- In `INTERNAL` builds: binds to `if (sessionTokenRetryCount > 1)` (line 468)

Both are syntactically correct, but the pattern is unusual — a future editor could easily break one configuration while fixing the other, and IDE code folding / "go to matching brace" will behave differently depending on which constants are defined.

**Consider making each branch self-contained:**

```csharp

#if !INTERNAL
                    if (this.sessionTokenRetryCount >= MaxSessionTokenRetryCount)
                    {
                        this.addHubRegionProcessingOnlyHeader = true;
                    }

                    if (this.sessionTokenRetryCount > MaxSessionTokenRetryCount)
                    {
                        return ShouldRetryResult.NoRetry();
                    }
                    else
                    {
                        this.retryContext = new RetryContext { ... };
                        return ShouldRetryResult.RetryAfter(TimeSpan.Zero);
                    }

#else
                    if (this.sessionTokenRetryCount > 1)
                    {
                        return ShouldRetryResult.NoRetry();
                    }
                    else
                    {
                        this.retryContext = new RetryContext { ... };
                        return ShouldRetryResult.RetryAfter(TimeSpan.Zero);
                    }

#endif

```

This duplicates ~5 lines but eliminates the cross-boundary `else` binding. Not blocking — the current code compiles correctly in both configurations.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -466,43 +466,16 @@ public async Task ClientRetryPolicy_HubRegionHeader_AddedOn404_1002_BasedOnAccou
...
+                // For single master, after the second 404/1002, the hub region header flag has been set
+                // and the retry policy should allow one more retry so the request can be sent with the header.
+                Assert.IsTrue(shouldRetry.ShouldRetry, "Single master should retry once more after second 404/1002 so the hub region header is sent.");
+
+                // Verify the header is now present on the retry
+                retryPolicy.OnBeforeSendRequest(request);
+                headerValues = request.Headers.GetValues(HubRegionHeader);
+                Assert.IsNotNull(headerValues, "Hub region header should be present on retry after second 404/1002.");
+                Assert.AreEqual(1, headerValues.Length, "Header should have exactly one value.");
+                Assert.AreEqual(bool.TrueString, headerValues[0], "Header value should be 'True'.");
```
xinlian12, line 478: 🟡 **Testing: Terminal NoRetry condition lost at unit test level**

The old version of this test explicitly asserted `Assert.IsFalse(shouldRetry.ShouldRetry)` after the 2nd 404/1002 — that was the unit-level guard for the "stop retrying" threshold. The new version correctly asserts `Assert.IsTrue(shouldRetry.ShouldRetry)` (verifying the fix), verifies the hub header is present on the retry, and then **stops**. It never fires a 3rd 404/1002 to assert `ShouldRetry == false`.

The `EndToEnd_DirectMode` test simulates 403/3 on the 3rd request (different code path), and the emulator test catches the terminal `CosmosException` implicitly — but neither exercises the exact `sessionTokenRetryCount > MaxSessionTokenRetryCount → NoRetry()` branch with a direct unit assertion.

Per `.pragentrules`: *"DO verify new tests are additive, not substitutive. If an existing test is modified, verify the original scenario is still covered."*

**Suggestion:** Extend the single-master branch of this test (or `EndToEnd_DirectMode`) by adding one more step after the hub header is verified — fire a 3rd 404/1002 and assert `shouldRetry.ShouldRetry == false`. This directly guards the `> MaxSessionTokenRetryCount` threshold:

```csharp
// Fire 3rd 404/1002 → should be terminal NoRetry
shouldRetry = await retryPolicy.ShouldRetryAsync(sessionNotAvailableException2, CancellationToken.None);
Assert.IsFalse(shouldRetry.ShouldRetry, "Single master should stop retrying after hub header retry also fails with 404/1002.");

```

xinlian12:
✅ **Review complete** (22:21)

PR #5793, opened 2026-04-20
[Internal] DTS: Adds required document id to all DTS operations

Meghana-Palaparthi:
## Description

- The coordinator needs `documentId` to reliably commit or abort stuck transactions—the transaction log doesn’t store document content.
- Always sending it enables deterministic replay without request bodies and avoids special cases (especially for create).
- Making it mandatory keeps the SDK–backend contract future‑proof and correct.

All distributed transaction operations now require an explicit 'id' parameter, consistent with Delete, Replace, and Patch. Previously, Create and Upsert omitted id from the metadata sent to the DTC coordinator (the id was only in the resource body).

- Add string id parameter to CreateItem, CreateItemStream, UpsertItem, UpsertItemStream in both the abstract class and the core implementation
- Validate id at call site via existing ValidateItemId (fails fast with ArgumentNullException before serialization)
- Simplify DistributedTransactionOperation<T>: one constructor (with id), MaterializeResourceAsync just serializes and calls base
- Update all tests

> Note: To ensure consistency with Create/Upsert operations in other SDK flows, we will implement a long-term solution by extracting the id from the resource body instead of having users provide it explicitly.
 
## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
@@ -94,13 +94,15 @@ namespace Microsoft.Azure.Cosmos
 
         public T Resource { get; internal set; }
 
-        internal override async Task MaterializeResourceAsync(CosmosSerializerCore serializerCore, CancellationToken cancellationToken)
+        internal override Task MaterializeResourceAsync(CosmosSerializerCore serializerCore, CancellationToken cancellationToken)
         {
             if (this.body.IsEmpty && this.Resource != null)
             {
                 this.ResourceStream = serializerCore.ToStream(this.Resource);
-                await base.MaterializeResourceAsync(serializerCore, cancellationToken);
+                return base.MaterializeResourceAsync(serializerCore, cancellationToken);
             }
+
+            return Task.CompletedTask;
         }
     }
 }
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------
+// ------------------------------------------------------------
 // Copyright (c) Microsoft Corporation.  All rights reserved.
 // ------------------------------------------------------------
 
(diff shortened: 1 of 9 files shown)


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
```diff
@@ -78,19 +78,6 @@ internal virtual async Task MaterializeResourceAsync(CosmosSerializerCore serial
...
-            string database,
-            string container,
-            PartitionKey partitionKey,
-            T resource,
-            DistributedTransactionRequestOptions requestOptions = null)
-            : base(operationType, operationIndex, database, container, partitionKey, id: null, requestOptions)
-        {
-            this.Resource = resource;
-        }
-
```
Praveen-Msft, line 93: why this is being removed?

Also, the PR description mentions removal of extractIdFromBody, BOM handling, Utf8JsonReader etc but could not find it in the PR.

Meghana-Palaparthi: This constructor previously set `id=null` when the id wasn't provided, such as with `CreateItem`. Since `id` is now a required parameter, this constructor is no longer necessary, and the base constructor is sufficient.

I have now updated the PR description, as there was some leftover content from the initial commit.

Praveen-Msft (approved):

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionSerializer.cs
```diff
@@ -62,6 +63,9 @@ private static void WriteOperation(Utf8JsonWriter jsonWriter, DistributedTransac
             // collectionName
             jsonWriter.WriteString("collectionName", operation.Container);
 
+            // id
```
kirankumarkolli, line 66: nit: do order matter for ORCH perf? Please check with crew once.

kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionSerializer.cs
Meghana-Palaparthi: Confirmed that there's no dependency on order in the orchestrator. Any changes here won't break the flow

PR #5799, opened 2026-04-22
Change Feed Processor: Fixes duplicate lease docs by using id as partition-key value

NaluTripician:
# Change Feed Processor: Fixes duplicate lease docs by using id as partition-key value

Closes IcM 768856224.

---

### What the customer saw

The customer's Change Feed Processor (CFP) silently stopped processing changes. Restarting hosts did not help. Their lease container contained documents like:

| id | LeaseToken | partitionKey | Owner | _ts |
|---|---|---|---|---|
| `MyProcessorhost1_abc_xyz..0` | `0` | `d4e2…` (Guid) | `null` | frozen |
| `MyProcessorhost1_abc_xyz..0` | `0` | `7a9f…` (different Guid) | `null` | frozen |

Every lease had a duplicate entry: **same `id`, same `LeaseToken`, different Guid `partitionKey`**, `Owner: null`, timestamps frozen at the moment the processor died.

### Root cause

The CFP lease container was partitioned by `/partitionKey`. On a partition split, `PartitionSynchronizerCore.HandlePartitionGoneAsync` delegates child-lease creation to `DocumentServiceLeaseManagerCosmos.CreateLeaseIfNotExistAsync`, which generated the partition-key value with `Guid.NewGuid().ToString()`:

```csharp
// DocumentServiceLeaseManagerCosmos.cs (before)
this.requestOptionsFactory.AddPartitionKeyIfNeeded(
    (string pk) => documentServiceLease.LeasePartitionKey = pk,
    Guid.NewGuid().ToString());
```

`TryCreateItemAsync` relies on Cosmos's per-partition-key `id` uniqueness check to turn concurrent creates into a `409 Conflict`. **That check only catches duplicates with the *same* partition key.** Every retry of split handling — host restart mid-split, transient error retry, or two hosts racing on the same parent lease — rolled a *new* random Guid, bypassing the uniqueness check and silently persisting a second document with identical `id`/`LeaseToken` but a different `partitionKey`.

Once duplicates exist, every balance tick (~13 s) lands here in `EqualPartitionsBalancingStrategy.CategorizeLeases`:

```csharp
allPartitions.Add(lease.CurrentLeaseToken, lease); // ArgumentException on duplicate
```

The `ArgumentException` propagates out of `CalculateLeasesToTake`, the balancer catches it, logs, and returns zero leases — **every tick, forever**. No host ever claims a lease; `Owner` stays `null`; the feed permanently stalls. There is no automatic recovery: the duplicate documents must be deleted manually.

---

## The fix

Set the partition-key value on each new lease document to the lease's own deterministic id (already computed as `this.GetDocumentId(leaseToken)`), in both overloads of `DocumentServiceLeaseManagerCosmos.CreateLeaseIfNotExistAsync`:

```csharp
// DocumentServiceLeaseManagerCosmos.cs (after)
// Use the lease document id as the partition-key value so that retries / concurrent
// creates for the same lease resolve to the same (id, partitionKey) tuple. This lets the
// Cosmos per-partition-key id-uniqueness check turn duplicates into a 409 Conflict
// instead of silently persisting cross-partition-key duplicates.
this.requestOptionsFactory.AddPartitionKeyIfNeeded(
    (string pk) => documentServiceLease.LeasePartitionKey = pk,
    leaseDocId);
```

With this change, **concurrent or retried creates of the same lease resolve to the same `(id, partitionKey)` tuple**, so Cosmos's per-partition-key id-uniqueness check fires normally and returns `409 Conflict`. `TryCreateItemAsync` turns that into a benign "already exists" outcome — the duplicate is prevented at the source. This is the pattern already used by `DocumentServiceLeaseStoreCosmos` for its marker and lock documents (which use `markerDocId` / `lockId` as both id and pk).

### Backward compatibility

Fully compatible with existing lease containers. The read path (`TryGetLeaseAsync`, `ReleaseAsync`, `UpdateLeaseAsync`, `CheckpointAsync`, `DeleteAsync`) pulls the partition-key value off the deserialized lease document via `requestOptionsFactory.GetPartitionKey(lease.Id, lease.PartitionKey)`, which for `/partitionKey`-partitioned containers returns `new PartitionKey(partitionKey)` — i.e. whatever was stored. Pre-existing lease documents that have Guid-based `partitionKey` values continue to load, refresh, and be released normally. New lease documents created after this change are written with `partitionKey == id`; once all parent leases have been split / replaced, the container naturally converges to the new scheme with no manual migration.

### Affected containers

Only `/partitionKey`-partitioned lease containers were exposed to the bug. Lease containers partitioned by `/id` or on single-partition (fixed) collections never exercised the affected code path (`AddPartitionKeyIfNeeded` is a no-op on those request-options factories — see `PartitionedByIdCollectionRequestOptionsFactory` and `SinglePartitionRequestOptionsFactory`) and require no action.

---

### Unit tests

Extended the existing `ValidateRequestOptionsFactory` helper used by `DocumentServiceLeaseManagerCosmosTests.CreatesEPKBasedLease` and `CreatesPartitionKeyBasedLease` (both `[DataTestMethod]` with rows for each of the three `RequestOptionsFactory` implementations). For the `PartitionedByPartitionKeyCollectionRequestOptionsFactory` case the helper now asserts `lease.PartitionKey == lease.Id`, proving the new deterministic wiring for both the PK-range overload and the EPK overload.

```
Passed!  - Failed: 0, Passed: 15, Skipped: 0, Total: 15
```

Full ChangeFeed unit-test slice: **261 / 261 pass**.

### Manual verification

- `dotnet build Microsoft.Azure.Cosmos.sln -c Debug` — 0 warnings, 0 errors.
- ChangeFeed unit-test slice passes locally.

---

## Performance impact

None. The change replaces one `Guid.NewGuid().ToString()` call with a variable reference (`leaseDocId`, already computed one line earlier in the same method). No extra IO.

---

## Customer remediation (for customers already in this state)

Delete the duplicate lease documents (same `id`, different `partitionKey`) from the lease container, then restart the processors. A one-time query is sufficient:

```sql
SELECT c.id, c.partitionKey, c.LeaseToken FROM c
```

Group by `id`; for each group with more than one document, keep one and delete the rest (using the matching `partitionKey` as the partition-key value on the delete).

Once this fix rolls out, new child leases created during a split/merge are immune to the duplicate-creation race — concurrent or retried creates for the same lease collide on `(id, partitionKey)` and Cosmos enforces uniqueness.

---

## Type of change

- [x] Bug fix (non-breaking)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Checklist

- [x] My code follows the code style of this project
- [x] I have added tests to cover my changes (extended `ValidateRequestOptionsFactory` to assert deterministic pk == id on `/partitionKey`-partitioned containers)
- [x] All new and existing tests pass locally (261 / 261 ChangeFeed unit tests)
- [x] I have verified the fix resolves the customer incident scenario (IcM 768856224)

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
@@ -128,7 +128,11 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                 Mode = this.GetChangeFeedMode()
             };
 
-            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, Guid.NewGuid().ToString());
+            // Use the lease document id as the partition-key value so that retries / concurrent
+            // creates for the same lease resolve to the same (id, partitionKey) tuple. This lets the
+            // Cosmos per-partition-key id-uniqueness check turn duplicates into a 409 Conflict
+            // instead of silently persisting cross-partition-key duplicates.
+            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, leaseDocId);
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
@@ -153,7 +157,11 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                 Mode = this.GetChangeFeedMode()
             };
 
-            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, Guid.NewGuid().ToString());
+            // Use the lease document id as the partition-key value so that retries / concurrent
+            // creates for the same lease resolve to the same (id, partitionKey) tuple. This lets the
+            // Cosmos per-partition-key id-uniqueness check turn duplicates into a 409 Conflict
+            // instead of silently persisting cross-partition-key duplicates.
+            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, leaseDocId);
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/DocumentServiceLeaseManagerCosmosTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/DocumentServiceLeaseManagerCosmosTests.cs
@@ -628,6 +628,13 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Tests
             else if (requestOptionsFactory is PartitionedByPartitionKeyCollectionRequestOptionsFactory)
             {
                 Assert.IsNotNull(lease.PartitionKey);
+
+                // The lease's partition-key value must equal its id so that concurrent/retry creates
+                // for the same lease resolve to the same (id, partitionKey) tuple and Cosmos's per-
+                // partition-key id-uniqueness check turns duplicates into a 409 Conflict. Using a
+                // random Guid here would silently persist cross-partition-key duplicates and stall
+                // the change feed load balancer. See DocumentServiceLeaseManagerCosmos.CreateLeaseIfNotExistAsync.
+                Assert.AreEqual(lease.Id, lease.PartitionKey);
             }
             else
             {


xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/EqualPartitionsBalancingStrategy.cs
```diff
@@ -138,7 +138,28 @@ private void CategorizeLeases(
...
                 Debug.Assert(lease.CurrentLeaseToken != null, "TakeLeasesAsync: lease.CurrentLeaseToken cannot be null.");
 
+                // Defensive: duplicate lease documents for the same lease token can exist when the lease
+                // container is partitioned by /partitionKey and split handling re-ran with a different
+                // Guid-based partition key (see PartitionSynchronizerCore.HandlePartitionGoneAsync).
+                // Use TryAdd so the load balancer keeps the first lease seen and continues functioning
+                // instead of throwing and blocking lease acquisition for the whole container. The
+                // duplicate documents must be deleted manually (or the lease container recreated with
+                // /id partitioning) to fully resolve the condition.
+                if (allPartitions.ContainsKey(lease.CurrentLeaseToken))
```
xinlian12: 🟡 **Recommendation — Design: Non-deterministic "first wins" can cause transient dual-processing**

The `continue` keeps whichever duplicate `GetAllLeasesAsync()` enumerates first and skips the rest. But `GetAllLeasesAsync()` runs a cross-partition `SELECT * FROM c WHERE STARTSWITH(...)` without `ORDER BY` — result ordering across logical partitions is non-deterministic.

**Concrete failure scenario:**

| Tick | Query returns first | Kept | Skipped | Effect |
|------|---|---|---|---|

| N | Lease-A (owner=host1) | ✅ host1 counted | Lease-B (owner=null) skipped | Correct |

| N+1 | Lease-B (owner=null) | Added to `expiredLeases` | Lease-A (owner=host1) skipped, **host1 undercounted** | ⚠️ Another host acquires Lease-B → two hosts process same feed range |

This only manifests when one duplicate is actively owned and the other isn't (the cross-host TOCTOU race), and the IcM scenario had all duplicates `Owner: null`, so it wouldn't apply there. But for the residual cross-host race this PR acknowledges, a 3-line tiebreaker would make the fix strictly more robust:

```csharp
if (allPartitions.ContainsKey(lease.CurrentLeaseToken))
{
    DocumentServiceLease existing = allPartitions[lease.CurrentLeaseToken];

    // Prefer the actively-owned lease to avoid offering a stale
    // duplicate for acquisition while another host is already processing.
    if (string.IsNullOrWhiteSpace(existing.Owner) && !string.IsNullOrWhiteSpace(lease.Owner))
    {
        allPartitions[lease.CurrentLeaseToken] = lease;
    }

    DefaultTrace.TraceWarning(...);
    continue;
}

```

If you prefer to defer this, the current code is still strictly better than the pre-fix crash — this is a refinement, not a correctness bug.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DuplicateLeaseRegressionTests.cs
```diff
@@ -0,0 +1,236 @@
...
+                2,
+                duplicateCount,
+                "Expected exactly two lease documents sharing the same LeaseToken after seeding; the cross-PK duplicate state was not reproduced.");
+
+            // Phase 3: Start a second processor. On un-patched code
+            // EqualPartitionsBalancingStrategy.CategorizeLeases throws on the first
+            // balance tick and the processor never picks up any change. On patched code
+            // the balancer emits a warning, keeps the first occurrence of each lease
+            // token, and continues.
+            ManualResetEventSlim allDocsProcessed = new ManualResetEventSlim(false);
```
xinlian12: 🟢 **Suggestion — Resource Cleanup: `ManualResetEventSlim` is `IDisposable` but not disposed**

`ManualResetEventSlim` implements `IDisposable` — when `Wait` contends (as it does here with a 60s timeout), it allocates a kernel `ManualResetEvent` handle that leaks until GC. Wrap in `using`:

```csharp
using ManualResetEventSlim allDocsProcessed = new ManualResetEventSlim(false);

```

Minor issue in test code only — no production impact.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/EqualPartitionsBalancingStrategyTests.cs
```diff
@@ -194,6 +194,32 @@ public void CalculateLeasesToTake_NeedTwoAndOtherOwnersEqualThanTargetCount_Retu
...
+                this.CreateLease(ownerNone, "1"),
+                this.CreateLease(ownerNone, "1"),
+                this.CreateLease(ownerNone, "2"),
+            };
+
+            // Must not throw ArgumentException from the internal Dictionary.Add.
+            List<DocumentServiceLease> leasesToTake = strategy.SelectLeasesToTake(allLeases).ToList();
+
+            // At minimum the balancer should consider at least one lease per distinct token.
+            Assert.IsTrue(leasesToTake.Count >= 1);
```
xinlian12: 🟢 **Suggestion — Testing: Assertion is too weak to catch regressions**

With 2 distinct tokens (both `ownerNone`), 1 host (`"self"`), and default strategy (`minPartitionCount=0`, `maxPartitionCount=0`):
- `partitionCount = 2`, `workerCount = 1`, `target = ceil(2/1) = 2`
- `partitionsNeededForMe = 2 - 0 = 2`
- Both leases are expired → `expiredLeases.Take(2)` returns 2

The assertion `>= 1` would still pass if the balancer erroneously dropped one of the two distinct tokens — consider:

```csharp
Assert.AreEqual(2, leasesToTake.Count, "Should consider all distinct lease tokens");

```

xinlian12:
✅ **Review complete** (38:33)

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/FeedManagement/EqualPartitionsBalancingStrategy.cs
jeet1995: Hey @NaluTripician - what would be the risk of any newly created leases having id and user chosen partition key path to have identical values as a rule by getting rid of using `Guid` (do flows break when you have some leases with different id and pk values and some leases with identical id and pk values)? The comment does bring up a good point that choosing the first lease amongst duplicates is a brittle/non-deterministic contract. 

Also, I feel in this scenario asking the customer to migrate should be the new container approach (asking customers to delete amongst duplicates, they wouldn't necessarily know which duplicate to delete).

NaluTripician: Hey @jeet1995, good questions. Let me take them in turn.

**On dropping the Guid in favor of a deterministic partition-key value:**
That is the long-term fix I called out under *Out of scope* in the description, and I agree it is the right end state. The cleanest value is the lease `id` itself (or `LeaseToken`) so `id`/PK collide and the service-side `409 Conflict` actually fires on retries. A few things kept me from rolling it into this PR:

1. **Mixed-state coexistence.** Existing customers already have leases written with `partitionKey = Guid.NewGuid()`. If a new SDK starts writing children with `partitionKey = id`, lookups still work (the read paths in `DocumentServiceLeaseManagerCosmos` always pass the PK through; nothing reads cross-partition by `id` alone), but a deterministic-PK insert will not *collide* with a leftover Guid-PK lease for the same token, so you still need the dedup pre-check from Fix 2 during the migration window. Mixed state is functionally fine, but the deterministic PK only buys you full idempotency once *all* leases on that container have been rewritten.
2. **Migration semantics.** There is no in-place rewrite of an existing lease''s PK in Cosmos. It would be a delete + re-insert per lease, which has to be coordinated against active hosts (you cannot yank a lease out from under a running observer). That is a bigger design conversation and I would rather land the unblock first and follow up with a properly designed migration story.

Happy to open a follow-up work item for the deterministic-PK change with these constraints captured if you agree.

**On the customer remediation guidance:**
Fair point. Telling customers to pick a duplicate to delete is unfriendly when both rows have `Owner: null` and frozen `_ts` (which is exactly the IcM state). I will update the warning message and the PR description to lead with "recreate the lease container with `/id` partitioning (or just delete the existing one and let CFP rebuild from your configured start position)" and demote the manual-dedup-query approach to a fallback for customers who cannot afford to lose checkpoints. Want me to push that as part of this PR?

xinlian12: I think:
  - we can still make the pkValue=id change, it is just within SDK, we still can not replying on it to make sure there is only 1 documents created per feed range. But at least it prevents multiple lease documents will be created in future
  - I also kind feel we should not add the de-dup logic within the SDK -  because we can not guarantee there is only 1 lease document exists for each feed range even with the fix. It will cause duplicate processing issues which is harder to debug I feel.

kirankumarkolli:
Duplicate #5807

PR #5802, opened 2026-04-22
Fix OS condition

jaredpar:
The `$(OS)` property on Linux machines is `Unix` not `Linux`. This was causing builds to fail on Linux. Flipped to making it a positive check for Windows since it looks like none of this should fire for non-Windows.

# Pull Request Template

## Description

## Type of change

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Microsoft.Azure.Cosmos.Tests.csproj
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Microsoft.Azure.Cosmos.Tests.csproj
@@ -69,7 +69,7 @@
     <ProjectReference Include="..\..\src\Microsoft.Azure.Cosmos.csproj" />
   </ItemGroup>  
 
-  <ItemGroup Condition="$(OS) != 'Linux' AND '$(ProjectRef)' != 'True' ">
+  <ItemGroup Condition="$(OS) == 'Windows_NT' AND '$(ProjectRef)' != 'True' ">
     <None Include="$(NugetPackageRoot)\Microsoft.HybridRow\$(HybridRowVersion)\microsoft.hybridrow.nuspec">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Microsoft.Azure.Cosmos.Tests.csproj
```diff
@@ -69,7 +69,7 @@
     <ProjectReference Include="..\..\src\Microsoft.Azure.Cosmos.csproj" />
   </ItemGroup>  
 
-  <ItemGroup Condition="$(OS) != 'Linux' AND '$(ProjectRef)' != 'True' ">
+  <ItemGroup Condition="$(OS) == 'Windows_NT' AND '$(ProjectRef)' != 'True' ">
```
Pilchie, line 72: This means that this will no longer run on macOS, right?  I think that's correct, but want to call it out explicitly.

Pilchie (approved):

Pilchie:
@kirankumarkolli - can you take a look?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Microsoft.Azure.Cosmos.Tests.csproj
jaredpar: Correct. That seemed to be the _intent_ here. Also MSBuild doesn't have a great way to differentiate between Linux and MacOS. The `$(OS)` value for both is `"Unix"`. Can differentiate but you have to use a different type of condition

kirankumarkolli: Since the original intent was not honored and so-far not know of any broken/impact. How about remove condition it-self?

Pilchie: Well, @jaredpar was trying to build on Linux and it didn't work, right? So presumably something about that doesn't work on Linux?

kirankumarkolli: > /usr/lib/dotnet/sdk/10.0.106/Microsoft.Common.CurrentVersion.targets(5381,5): error MSB3030: Could not copy the file "/home/kirankk/.nuget/packages/Microsoft.HybridRow/1.1.0-preview3/microsoft.hybridrow.nuspec" because it was not found.

Yup. lets take this PR then.

kirankumarkolli: Seems this was primarily driven through the yml files where its set. 
One option might be `<ItemGroup Condition="$(TargetFramework.Contains('windows'))">` (for build time?)

kirankumarkolli: WSL varified that it works

jaredpar: Yeah I was trying to review the other PR and I didn't quite understand how the build laid out files such that the targets could actually work. Opened up the repo in codespaces to play around and the build failed. The fix was essentially just changing the condition after which build worked fine.

PR #5803, opened 2026-04-22
Diagnostics: Fixes Authorization header leaked to DocumentDBClient ETW EventSource

NaluTripician:
# Diagnostics: Fixes Authorization header leaked to DocumentDBClient ETW EventSource

## Summary

The SDK's `DocumentDBClient` ETW `EventSource` (`GUID f832a342-0a53-5bab-b57b-d5bc65319768`) emits **Event ID 1 (`Request`)** at `Verbose` level / keyword `HttpRequestAndResponse` (`0x1`). Field #5 of that event's 33-field payload is the **raw value of the outbound HTTP `Authorization` header**. Any consumer that legitimately subscribes to this provider — for example a Geneva / GCS `<EtwProvider name="DocumentDBClient" …>`, PerfView, or `dotnet-trace` — captures the credential **in plaintext**:

- master-key HMAC signature (`type=master&ver=1.0&sig=…`),
- resource token, or
- AAD access token (`type=aad&…Bearer <JWT>`).

This is **secure-by-default-violation**. The Traffic Manager team recently enabled the `DocumentDBClient` provider in their Geneva GCS config and observed the leak in their CentralBond / Kusto pipeline. Bug has existed since the initial commit of `DocumentClientEventSource.cs`; customer is on `Microsoft.Azure.Cosmos 3.51.0` but every released 3.x version through current `master` (3.58.0 / 3.59.0-preview) is affected.

## Root cause

In `Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs`, the `[NonEvent]` public wrapper:

```csharp
[NonEvent]
public void Request(Guid activityId, Guid localId, string uri, string resourceType,
                    HttpRequestHeaders requestHeaders)
{
    if (this.IsEnabled(EventLevel.Verbose, Keywords.HttpRequestAndResponse))
    {
        string[] keysToExtract = { …, HttpConstants.HttpHeaders.Authorization, … };
        string[] headerValues = Helpers.ExtractValuesFromHTTPHeaders(requestHeaders, keysToExtract);
        this.Request(…, headerValues[1] /* Authorization */, …);   // <-- raw secret into ETW payload
    }
}
```

calls into the `[Event(1)]`-decorated method which pins the Authorization string directly into `dataDesc[5]` and emits it to ETW. `Helpers.ExtractValuesFromHTTPHeaders` performs no redaction — it returns the header verbatim.

The `Response` event (Event ID 2) is **not** affected; only the outbound request path leaks.

## Fix

Redact the Authorization slot **inside the `[NonEvent]` wrapper**, before it reaches the ETW `[Event]` method. The 33-field ETW manifest, event IDs, keywords, and field types are all preserved so existing Geneva / Kusto schemas parsing this provider continue to work.

```csharp
string[] headerValues = Helpers.ExtractValuesFromHTTPHeaders(requestHeaders, keysToExtract);

// SECURITY: The Authorization header contains a live credential
// (master-key HMAC signature, resource token, or AAD Bearer access token).
// It must never be written to the ETW event payload where any listener
// subscribing to the "DocumentDBClient" EventSource at Verbose level
// (e.g., Geneva MonitoringAgent, PerfView, dotnet-trace) would capture it.
// Replace the value with a fixed placeholder while preserving the 33-field
// ETW manifest so existing consumers remain compatible.
const int AuthorizationIndex = 1; // matches position in keysToExtract above
if (!string.IsNullOrEmpty(headerValues[AuthorizationIndex]))
{
    headerValues[AuthorizationIndex] = "REDACTED";
}
```

## Test

New file `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DocumentClientEventSourceTests.cs` adds two regression tests that run an in-proc `EventListener` subscribed to `DocumentDBClient` at Verbose / keyword `0x1`:

1. **`Request_AuthorizationHeader_IsRedactedInEtwPayload`** — emits a request with a recognisable secret authorization value; asserts `payload[5] == "REDACTED"` and **no** payload field contains the secret substring.
2. **`Request_NoAuthorizationHeader_EmitsEmptyString`** — verifies the redaction logic preserves the prior empty-string behaviour when the header is absent (internal plumbing requests).

```
dotnet test --filter FullyQualifiedName~DocumentClientEventSourceTests
  Passed!  - Failed: 0, Passed: 2, Skipped: 0, Total: 2
```

Full solution build: 0 warnings, 0 errors.

## Customer / field mitigation (ship alongside this PR in docs / known-issues)

For any team that already enabled the `DocumentDBClient` ETW provider in Geneva GCS (or similar), until they pick up an SDK version containing this fix:

1. **Remove** the `DocumentDBClient` `<EtwProvider>` block from their GCS XML. There is no level / keyword combination that captures request tracing without Event 1, and Event 1 is the only event that carries the Authorization field.
2. **Purge** existing CentralBond / Kusto rows for the retention window the provider was active.
3. **Rotate Cosmos DB master keys** if master-key auth was in use during the logging window. Resource tokens expire naturally; AAD tokens expire in ~1 h and cannot be rotated — treat as compromised until expiry.
4. File an ICM with `cosmosdbsec` following the standard secret-exposure IR runbook.
5. Optionally, enable the safer `Azure-Cosmos-Operation-Request-Diagnostics` EventSource (`CosmosDbEventSource`) instead, which emits operation-level diagnostics without HTTP headers.

## Suggested servicing

Because this is a **credential-disclosure** fix, this commit should be cherry-picked onto at least:

- `master` → next GA `3.59.0` and preview `3.60.0-preview`
- `3.51.x` (customer's pinned version)
- `3.58.x` (current LTS)

Engage MSRC / `cosmosdbsec` for CVE assignment prior to publishing the NuGet package; advisory wording should explicitly call out: *"affects callers who enable the `DocumentDBClient` ETW EventSource at Verbose level (Geneva MonitoringAgent, PerfView, dotnet-trace with that provider name)."*

## Checklist

- [x] PR title matches the enforced regex (`Diagnostics: Fixes …`).
- [x] Branch follows `users/<user>/<feature>` convention.
- [x] Changelog entry added under `3.59.0-preview.0` → `Fixed`.
- [x] Regression test added (2 cases, both passing).
- [x] No ETW manifest change — 33-field `Request` event, same field types, same event IDs / keywords — backward compatible for downstream Kusto / Geneva schemas.
- [ ] MSRC / cosmosdbsec coordination before release (human action required).
- [ ] Cherry-pick to `3.51.x` / `3.58.x` hotfix branches (human action required, per `ReleaseCopilotAgent` hotfix runbook).

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
@@ -278,7 +278,7 @@ namespace Microsoft.Azure.Cosmos
                 string[] keysToExtract =
                 {
                     HttpConstants.HttpHeaders.Accept,
-                    HttpConstants.HttpHeaders.Authorization,
+                    HttpConstants.HttpHeaders.Authorization, // SECURITY: value is redacted to "REDACTED" before Event 1 is emitted to ETW. Do not remove this entry or reorder the array without updating the redaction below - the [Event(1)] signature still declares the `authorization` parameter, so the slot must continue to be populated.
                     HttpConstants.HttpHeaders.ConsistencyLevel,
                     HttpConstants.HttpHeaders.ContentType,
                     HttpConstants.HttpHeaders.ContentEncoding,
--- a/Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
@@ -309,6 +309,20 @@ namespace Microsoft.Azure.Cosmos
                 };
 
                 string[] headerValues = Helpers.ExtractValuesFromHTTPHeaders(requestHeaders, keysToExtract);
+
+                // SECURITY: The Authorization header contains a live credential
+                // (master-key HMAC signature, resource token, or AAD Bearer access token).
+                // It must never be written to the ETW event payload where any listener
+                // subscribing to the "DocumentDBClient" EventSource at Verbose level
+                // (e.g., Geneva MonitoringAgent, PerfView, dotnet-trace) would capture it.
+                // Replace the value with a fixed placeholder while preserving the 33-field
+                // ETW manifest so existing consumers remain compatible.
+                const int AuthorizationIndex = 1; // matches position in keysToExtract above
+                if (!string.IsNullOrEmpty(headerValues[AuthorizationIndex]))
+                {
+                    headerValues[AuthorizationIndex] = "REDACTED";
+                }
+
                 this.Request(activityId, localId, uri, resourceType, headerValues[0], headerValues[1], headerValues[2], headerValues[3], headerValues[4],
                     headerValues[5], headerValues[6], headerValues[7], headerValues[8], headerValues[9], headerValues[10], headerValues[11], headerValues[12],
                     headerValues[13], headerValues[14], headerValues[15], headerValues[16], headerValues[17], headerValues[18], headerValues[19], headerValues[20],
--- a/Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
@@ -34,6 +34,51 @@ namespace Microsoft.Azure.Cosmos
             public const EventKeywords HttpRequestAndResponse = (EventKeywords)1;
         }
 
+        // Ordered list of headers extracted and forwarded positionally to Event(1).
+        // IMPORTANT: The order of entries here matches the field order declared by [Event(1)]
+        // and the `{index}` placeholders in its Message template. Do not reorder, insert, or
+        // remove entries without updating Event(1) and the redaction logic in Request(...).
+        // The Authorization slot is populated from the raw header and then overwritten with
+        // "REDACTED" in the [NonEvent] wrapper before being emitted to ETW.
+        private static readonly string[] RequestHeaderKeysToExtract =
+        {
+            HttpConstants.HttpHeaders.Accept,
+            HttpConstants.HttpHeaders.Authorization, // SECURITY: redacted to "REDACTED" in Request(...) before ETW emission. Do not remove.
+            HttpConstants.HttpHeaders.ConsistencyLevel,
+            HttpConstants.HttpHeaders.ContentType,
+            HttpConstants.HttpHeaders.ContentEncoding,
+            HttpConstants.HttpHeaders.ContentLength,
+            HttpConstants.HttpHeaders.ContentLocation,
+            HttpConstants.HttpHeaders.Continuation,
+            HttpConstants.HttpHeaders.EmitVerboseTracesInQuery,
+            HttpConstants.HttpHeaders.EnableScanInQuery,
+            HttpConstants.HttpHeaders.ETag,
+            HttpConstants.HttpHeaders.HttpDate,
+            HttpConstants.HttpHeaders.IfMatch,
+            HttpConstants.HttpHeaders.IfNoneMatch,
+            HttpConstants.HttpHeaders.IndexingDirective,
+            HttpConstants.HttpHeaders.KeepAlive,
+            HttpConstants.HttpHeaders.OfferType,
+            HttpConstants.HttpHeaders.PageSize,
+            HttpConstants.HttpHeaders.PreTriggerExclude,
+            HttpConstants.HttpHeaders.PreTriggerInclude,
+            HttpConstants.HttpHeaders.PostTriggerExclude,
+            HttpConstants.HttpHeaders.PostTriggerInclude,
+            HttpConstants.HttpHeaders.ProfileRequest,
+            HttpConstants.HttpHeaders.ResourceTokenExpiry,
+            HttpConstants.HttpHeaders.SessionToken,
+            HttpConstants.HttpHeaders.SetCookie,
+            HttpConstants.HttpHeaders.Slug,
+            HttpConstants.HttpHeaders.UserAgent,
+            HttpConstants.HttpHeaders.XDate
+        };
+
+        // Index of the Authorization header in RequestHeaderKeysToExtract, computed once at class
+        // initialization so the per-request redaction path in Request(...) does not pay an O(n)
+        // Array.IndexOf lookup on every HTTP call.
+        private static readonly int AuthorizationHeaderIndex
+            = Array.IndexOf(RequestHeaderKeysToExtract, HttpConstants.HttpHeaders.Authorization);
+
         [NonEvent]
         private unsafe void WriteEventCoreWithActivityId(Guid activityId, int eventId, int eventDataCount, EventSource.EventData* dataDesc)
         {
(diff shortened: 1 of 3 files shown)


kundadebdatta (dismissed):
LGTM

Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
```diff
@@ -309,6 +309,20 @@ public void Request(Guid activityId, Guid localId, string uri, string resourceTy
...
                 string[] headerValues = Helpers.ExtractValuesFromHTTPHeaders(requestHeaders, keysToExtract);
+
+                // SECURITY: The Authorization header contains a live credential
+                // (master-key HMAC signature, resource token, or AAD Bearer access token).
+                // It must never be written to the ETW event payload where any listener
+                // subscribing to the "DocumentDBClient" EventSource at Verbose level
+                // (e.g., Geneva MonitoringAgent, PerfView, dotnet-trace) would capture it.
+                // Replace the value with a fixed placeholder while preserving the 33-field
+                // ETW manifest so existing consumers remain compatible.
+                const int AuthorizationIndex = 1; // matches position in keysToExtract above
```
kirankumarkolli: Isn't it safe to remove it from the collection list above instead?

kirankumarkolli: Actual Event still has it part of the contract. 
Options are 
- Breaking new Event
- Redaction: Lets at-least add a comment in above list near Auth entry

kundadebdatta: Also, ordering wise, is the `AuthorizationToken` gurenteed to be at the 1th place in the headers ?

kirankumarkolli: Yes it taht how its extracted above.

Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
```diff
@@ -309,6 +309,26 @@ public void Request(Guid activityId, Guid localId, string uri, string resourceTy
...
+                // (master-key HMAC signature, resource token, or AAD Bearer access token).
+                // It must never be written to the ETW event payload where any listener
+                // subscribing to the "DocumentDBClient" EventSource at Verbose level
+                // (e.g., Geneva MonitoringAgent, PerfView, dotnet-trace) would capture it.
+                // Replace the value with a fixed placeholder while preserving the 33-field
+                // ETW manifest so existing consumers remain compatible. The IsNullOrEmpty
+                // guard preserves the existing "" semantics for requests that never had an
+                // Authorization header attached (e.g. internal plumbing); we only overwrite
+                // when a real value is present.
+                int authorizationIndex = Array.IndexOf(keysToExtract, HttpConstants.HttpHeaders.Authorization);
```
kirankumarkolli: This is overkill and unnecessary for every call lookup overhead. At-max once static lookup best. 

Or IMO just Debug assert is fine for index assumption.

Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
NaluTripician: Good call-out, and thanks for reasoning through both options so I could land on the minimal one.

You're right that the `[Event(1)]` signature is the real contract — the `string authorization` parameter (and its `{5}` slot in the `Message` template, plus the `dataDesc[5]` write) is what actually ships the ETW manifest. Removing `HttpConstants.HttpHeaders.Authorization` from the key list alone doesn't remove it from the wire; it only shifts where the placeholder comes from. The two options that fully drop it are:

1. **Breaking new Event** — bump Event ID to 3 (or rev `[Event(1, Version = …)]`), drop the `authorization` parameter, regenerate the manifest. That breaks every downstream Kusto/Geneva consumer already parsing `DocumentDBClient` rows — including the customer that reported this. Not worth it for a security patch we also want to cherry-pick to the 3.51.x / 3.58.x servicing lines.
2. **Redaction at the boundary** — what this PR does. Manifest, event IDs, keywords, and field types stay identical, so existing schemas parse the same rows, they just see `"REDACTED"` in field 5.

I kept `Authorization` in the key list so the positional forwarding `headerValues[0..28]` stays 1:1 with the `[Event(1)]` parameter order and a future refactor can't silently shift indices.

Applied your minimum ask in [`ee0be9948`](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5803/commits/ee0be9948) (comment next to the Authorization entry) and then folded it into [`9a1ecd8ba`](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5803/commits/9a1ecd8ba) when I hoisted the array to a `private static readonly` field per your other comment.

NaluTripician: Great catch — that exact concern came up in a local review pass and was addressed in [`8fee68628`](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5803/commits/8fee68628) (and further tightened in [`9a1ecd8ba`](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5803/commits/9a1ecd8ba)). The index is no longer assumed to be `1`:

```csharp
private static readonly int AuthorizationHeaderIndex
    = Array.IndexOf(RequestHeaderKeysToExtract, HttpConstants.HttpHeaders.Authorization);
```

Computed once at class init (no per-request overhead), with a `Debug.Assert(AuthorizationHeaderIndex >= 0, …)` inside `Request(...)` to catch any future refactor that removes the entry from the array. So if the list is alphabetised, extended, or reordered, redaction still targets the right slot.

(Your earlier `LGTM` got auto-dismissed by GitHub when the round-1 review commits went in — happy to get a fresh re-review once you've looked at the last two commits.)

Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
NaluTripician: Agreed, done in [`9a1ecd8ba`](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5803/commits/9a1ecd8ba). Both the key list and the Authorization index are now `private static readonly` fields at the top of the class — the index is computed once at class initialization, so the per-request path in `Request(...)` is just a `Debug.Assert` + an `IsNullOrEmpty` guard + one array write. No `Array.IndexOf` on the hot path.

I kept `Debug.Assert(AuthorizationHeaderIndex >= 0, …)` rather than falling back to a hardcoded index, because the cost is zero in Release builds and it gives us a loud failure in debug/tests if anyone ever deletes the Authorization entry from the array. Let me know if you'd prefer I drop that and rely purely on the static init throwing at class load instead.

ananth7592 (dismissed):

Microsoft.Azure.Cosmos/src/DocumentClientEventSource.cs
```diff
@@ -275,40 +320,28 @@ public void Request(Guid activityId, Guid localId, string uri, string resourceTy
...
-                    HttpConstants.HttpHeaders.UserAgent,
-                    HttpConstants.HttpHeaders.XDate
-                };
+                    headerValues[AuthorizationHeaderIndex] = "REDACTED";
+                }
 
-                string[] headerValues = Helpers.ExtractValuesFromHTTPHeaders(requestHeaders, keysToExtract);
                 this.Request(activityId, localId, uri, resourceType, headerValues[0], headerValues[1], headerValues[2], headerValues[3], headerValues[4],
                     headerValues[5], headerValues[6], headerValues[7], headerValues[8], headerValues[9], headerValues[10], headerValues[11], headerValues[12],
                     headerValues[13], headerValues[14], headerValues[15], headerValues[16], headerValues[17], headerValues[18], headerValues[19], headerValues[20],
```
Pilchie, line 362: I know this isn't new, but this pattern is weird. We can't just pass the whole array in instead of each value as a separate parameter?

NaluTripician: I can look at this in a separate issue

NaluTripician: Dug into this on a separate worktree off `master` — short answer: **no, not without breaking the ETW wire contract**, and the awkwardness is a direct consequence of classic-manifest `EventSource` constraints rather than a stylistic choice.

### Why the 33 positional parameters are forced

`DocumentClientEventSource` is a **classic manifest** EventSource — explicit `[EventSource(Name="DocumentDBClient", Guid="f832a342-…")]`, explicit `[Event(1, Message="{0}..{32}")]`, hand-rolled `WriteEventCore` over `stackalloc EventData[33]`. On that path:

1. **The ETW manifest is generated from the method signature via reflection at class init.** Each parameter becomes a distinct `<data name="…" inType="…"/>` entry. There is **no manifest type for `string[]`** on the classic path — EventSource will either refuse to register the event or emit it in a shape existing consumers can't parse.
2. **`Message = "{0} … {32}"` binds positional slots 1:1 to manifest field indices.** Collapsing to one array parameter collapses all 29 header fields into one blob, breaking the template and the named fields downstream parsers key on.
3. **`dataDesc[0..32]` is the literal payload layout on the wire.** Each slot is a discrete `EventData` descriptor (pointer + length) that PerfView / Geneva/MDS / Kusto ingestion read positionally.
4. **`params string[]` on the target doesn't help** — EventSource reflects the declared parameter list, and C# has no spread-into-positional-arguments for non-`params` methods, so the caller-side `headerValues[0]..[28]` unpacking is also unavoidable given the target signature.

### Why "just switch to self-describing EventSource" isn't free

`EventSourceSettings.EtwSelfDescribingEventFormat` *does* support array payloads and would let us write `WriteEvent(1, headerValues)`. But:

- It's a **different wire format** — same blast radius as the "Breaking new Event" option [@kirankumarkolli](https://github.com/kirankumarkolli) raised earlier in this review ([r3126060841](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5803#discussion_r3126060841)), which we explicitly rejected to keep this cherry-pickable into 3.51.x / 3.58.x servicing.
- The positional `Message="{0}..{32}"` template goes away, so PerfView consumers lose the formatted message.
- Per-header field names disappear from the manifest — everything becomes one opaque `string[]` blob consumers index into by convention.

### Recommendation for the follow-up

I'll still file a follow-up issue as promised, but based on this the resolution is **"won't fix — by design"** with this analysis attached. The only way to make the signature prettier is to break the event contract (new `[Event(3)]`, coordinated deprecation with the Geneva/Kusto telemetry consumers), which is a major-version move and the same reason we chose redaction over breaking here.

kirankumarkolli (approved):

PR #5806, opened 2026-04-22
Metadata Retry: Fixes cross-region failover preempted by caller cancellation

### LINKED ISSUE #5805: Metadata retry: cross-region failover preempted by caller cancellation during control-plane HTTP timeout escalation
## Summary

When the .NET Cosmos SDK warms a cold collection-metadata cache (`ClientCollectionCache.GetByRidAsync` / `GetByNameAsync`), the `ReadCollection` call is routed through `TaskHelper.InlineIfPossible(..., retryPolicy, cancellationToken)`, which delegates to `BackoffRetryUtility<T>.ExecuteAsync` in Cosmos.Direct. That utility evaluates `cancellationToken.ThrowIfCancellationRequested()` **before** consulting `IDocumentClientRetryPolicy.ShouldRetryAsync`.

The control-plane HTTP timeout policy escalates at 0.5 s → 5 s → 30 s (~36 s total) against an unhealthy region. When the caller's request-level timeout (typically ~36.5 s) trips during that escalation, `ThrowIfCancellationRequested` fires on the next iteration and throws `OperationCanceledException` without ever consulting `ClientRetryPolicy` — which would otherwise have returned "retry to next preferred region". The customer sees a `CosmosOperationCanceledException` instead of a successful cross-region failover.

This is the bug described in the diagnostic bundle attached to PR #5787.

## Reproduction

The repro is included in PR #5787 (`ClientRetryPolicyTests` + `ControlPlaneHedgingReproTests`). A concise end-to-end repro:

1. Preferred regions = `[East US 2, Central US]` where East US 2 is unhealthy (gateway returns 503 after slow timeouts).
2. Cold collection cache.
3. Issue `QueryDocument` with a ~36.5 s client-side timeout.
4. Observe: `CosmosOperationCanceledException`; no HTTP call to Central US.

Expected: the cross-region retry against Central US executes and the query succeeds.

## Root cause

`BackoffRetryUtility<T>.ExecuteAsync` (in Cosmos.Direct, not editable in this repo) honors the caller token before calling `ShouldRetryAsync`. See `ClientRetryPolicy.ShouldRetryAsync` lines 141-156 — on OCE it explicitly returns non-retry, but in the preempted path the policy is never consulted at all.

`PartitionKeyRangeCache` is not affected (its `BackoffRetryUtility.ExecuteAsync` call does not thread a caller cancellation token).

## Fix

See PR #5806. New internal `MetadataRetryHelper` runs a bespoke retry loop that:

- Always consults the retry policy on exception, before honoring the caller token.
- On caller-cancelled + retryable: grants ONE bounded grace window (default 10 s) for a cross-region retry attempt using a detached token.
- Surfaces the original exception (not grace-timeout OCE) if grace expires or the grace attempt fails.

Wired into `ClientCollectionCache.GetByRidAsync` and `GetByNameAsync`.

## Related

- Repro tests: #5787
- Fix PR: #5806

## Cross-SDK tracking

The same class of defect was investigated across all maintained Cosmos SDKs. Sibling issues:

- Python: Azure/azure-sdk-for-python#46471
- Rust: Azure/azure-sdk-for-rust#4253
- Go: Azure/azure-sdk-for-go#26649
- Java: not filed — Reactor's `retryWhen` operator structurally isolates subscription cancellation from the retry decision, so the defect does not reproduce.

NaluTripician:
## Summary

Fixes Azure/azure-cosmos-dotnet-v3#5805.

When the .NET Cosmos SDK has to warm a cold collection-metadata cache (`ClientCollectionCache.GetByRidAsync` / `GetByNameAsync`), the `ReadCollection` call goes through `TaskHelper.InlineIfPossible` → `BackoffRetryUtility<T>.ExecuteAsync`. That utility checks the caller's cancellation token **before** invoking `ShouldRetryAsync`. When the control-plane HTTP timeout policy burns 0.5 s → 5 s → 30 s (~36 s) against an unhealthy region and the caller's ~36.5 s timeout trips, `ClientRetryPolicy`'s "retry to next preferred region" decision is silently preempted and the operation surfaces `OperationCanceledException`.

Repro was landed in #5787.

## Fix

- New internal `MetadataRetryHelper.ExecuteAsync<T>(Func<CancellationToken, Task<T>>, IDocumentClientRetryPolicy, CancellationToken)`.
- Consults the retry policy on every exception **before** honoring caller cancellation.
- On caller-cancelled + retryable exception: grants one bounded grace window (default 10 s) for a cross-region retry attempt on a detached token. The grace token is threaded through the operation lambda, so the underlying HTTP call is not preempted by the (already cancelled) caller token.
- Surfaces the **original** exception (not a grace-timeout `OperationCanceledException`) when the grace window expires or the grace attempt fails.
- Wired into `ClientCollectionCache.GetByRidAsync` and `GetByNameAsync`.

## Scope

- `PartitionKeyRangeCache` is **not** affected by this defect (its call does not pass a caller CT into `BackoffRetryUtility`); no change needed.
- `AddressCache` is intentionally unchanged.
- Tuning the HTTP timeout escalation (0.5 s / 5 s / 30 s) is tracked separately.
- Extending the hedging availability strategy to metadata reads is tracked separately as a design doc.

## Tests

`Microsoft.Azure.Cosmos.Tests/MetadataRetryHelperTests.cs` — 10 tests:
- `ExecuteAsync_SucceedsFirstAttempt_NoRetry`
- `ExecuteAsync_RetriesOnTransient_WhenPolicyAllows`
- `ExecuteAsync_CrossRegionRetryExecutes_EvenWhenCallerTokenCancelled` *(primary fix validation)*
- `ExecuteAsync_GraceIsBounded_SurfacesOriginalExceptionOnSecondFailure`
- `ExecuteAsync_PolicyDeniesRetry_SurfacesOriginalExceptionEvenWhenCancelled`
- `ExecuteAsync_ZeroGrace_DoesNotRetryAfterCancellation`
- `ExecuteAsync_GraceExpires_SurfacesOriginalException`
- `LegacyBackoffRetryUtility_CrossRegionRetryIsPreempted_ByCancelledCallerToken` *(pins the pre-fix buggy behavior so accidental revert is caught)*
- `ExecuteAsync_FirstAttemptOCE_PolicyIsConsultedBeforeSurfacing`
- `ExecuteAsync_NegativeGrace_Throws`

All pass. The companion "legacy" test demonstrates that the pre-fix code path preempts the retry and asserts the buggy behavior — this is the proof that the defect exists and that the fix is required.

## Risk / rollout

- Caller's effective timeout extends by up to 10 s (`DefaultCrossRegionRetryGrace`) when the cross-region retry grace is invoked. This is opt-in per call via existing cancellation semantics: callers that want the previous behavior can continue to pass any CancellationToken; the grace only triggers when the retry policy says "retry" **and** the caller token is already cancelled.
- No API surface changes — new helper is internal.
- `PartitionKeyRangeCache` behavior unchanged.

## Review

This PR went through two rounds of agent review before opening. Round 1 surfaced three items (operation-token contract comment, first-attempt-OCE regression test, negative-grace validation) which are committed in the second commit. Round 2 returned `APPROVE` with no findings.

Closes #5805.

---


NaluTripician:
@sdkReviewAgent-2

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
```diff
@@ -0,0 +1,238 @@
...
+                    throw; // unreachable
+                }
+
+                if (shouldRetry == null || !shouldRetry.ShouldRetry)
+                {
+                    // Honor ShouldRetryResult.ExceptionToThrow if the policy has specified a
+                    // wrapper/translated exception (matches BackoffRetryUtility.ThrowIfDoneTrying).
+                    if (shouldRetry?.ExceptionToThrow != null)
+                    {
+                        throw shouldRetry.ExceptionToThrow;
```
xinlian12, line 152: 🟢 **Suggestion — Correctness: ExceptionToThrow stack-trace divergence from ThrowIfDoneTrying**

The mirrored `ShouldRetryResult.ThrowIfDoneTrying` that `BackoffRetryUtility` uses has a `ReferenceEquals` guard: when the policy's `ExceptionToThrow` is the same object reference as the caught exception, it calls `capturedException.Throw()` to preserve the original stack trace. Here, `throw shouldRetry.ExceptionToThrow` always resets the stack.

```csharp
// ThrowIfDoneTrying (BackoffRetryUtility):
if (ReferenceEquals(this.ExceptionToThrow, capturedException.SourceException))
    capturedException.Throw(); // preserves stack
else
    throw this.ExceptionToThrow; // new stack — translated exception

```

**Why it matters:** If a future retry policy calls `ShouldRetryResult.NoRetry(originalException)` with the same reference, diagnostics lose the original throw site. No current policy in the collection-read chain does this (I confirmed zero `ExceptionToThrow` usage in `src/*RetryPolicy*.cs`), so the practical impact today is nil — but the contract difference could surprise a future author.

**Suggested fix:**

```csharp
if (shouldRetry?.ExceptionToThrow != null)
{
    if (object.ReferenceEquals(shouldRetry.ExceptionToThrow, capturedException.SourceException))
    {
        capturedException.Throw();
    }

    throw shouldRetry.ExceptionToThrow;
}

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/MetadataRetryHelperTests.cs
```diff
@@ -0,0 +1,466 @@
...
+                "Helper must surface ShouldRetryResult.ExceptionToThrow (policy-translated), not the original exception.");
+        }
+
+        /// <summary>
+        /// Negative grace windows are invalid and must be rejected at the API boundary rather
+        /// than silently collapsed to "disabled". This prevents subtle misuse.
+        /// </summary>
+        [TestMethod]
+        [Owner("ntripician")]
+        public async Task ExecuteAsync_NegativeGrace_Throws()
```
xinlian12, line 522: 🟡 **Recommendation — Testing: MaxAttemptsHardCap (20) is untested**

This defensive hard cap prevents infinite loops from misconfigured policies — but no test verifies it fires. The commit message for `6c51bc3` explicitly lists "attempt cap" as a review finding that was addressed, yet the safeguard itself has no test.

**Why it matters:** If someone refactors this check (off-by-one, moves it below the catch, or removes it), no test catches the regression. A misconfigured policy that always returns `ShouldRetry=true` would silently spin.

**Suggested test:**

```csharp
[TestMethod]
public async Task ExecuteAsync_ExceedsMaxAttemptsHardCap_ThrowsInvalidOperationException()
{
    Mock<IDocumentClientRetryPolicy> policy = new();
    policy.Setup(p => p.ShouldRetryAsync(It.IsAny<Exception>(), It.IsAny<CancellationToken>()))
        .ReturnsAsync(ShouldRetryResult.RetryAfter(TimeSpan.Zero));

    int attempts = 0;
    InvalidOperationException thrown = await Assert.ThrowsExceptionAsync<InvalidOperationException>(
        () => MetadataRetryHelper.ExecuteAsync<int>(
            (_) => { attempts++; throw new Exception("always fail"); },
            policy.Object,
            CancellationToken.None));

    Assert.AreEqual(20, attempts);
    StringAssert.Contains(thrown.Message, "defensive attempt cap");
}

```

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
```diff
@@ -0,0 +1,238 @@
...
+                // metadata reads when the caller's timeout trips during the HTTP
+                // timeout policy escalation.
+                ShouldRetryResult shouldRetry;
+                try
+                {
+                    shouldRetry = await retryPolicy
+                        .ShouldRetryAsync(capturedException.SourceException, CancellationToken.None)
+                        .ConfigureAwait(false);
+                }
+                catch (Exception policyException)
```
xinlian12, line 128: 🟡 **Recommendation — Testing: ShouldRetryAsync-throws path is untested**

The `catch (Exception policyException)` block (lines 128–142) rethrows the **original** operation exception when the retry policy itself throws — not the policy's internal exception. This is a deliberate and correct design choice, but no test covers it.

**Why it matters:** A future refactor could accidentally surface the policy exception instead of the original, or remove the `capturedException.Throw()` call. The behavior (original exception surfaces, policy error is logged) is important for debuggability.

**Suggested test:**

```csharp
[TestMethod]
public async Task ExecuteAsync_PolicyThrowsDuringShouldRetry_SurfacesOriginalException()
{
    Mock<IDocumentClientRetryPolicy> policy = new();
    policy.Setup(p => p.ShouldRetryAsync(It.IsAny<Exception>(), It.IsAny<CancellationToken>()))
        .ThrowsAsync(new InvalidOperationException("policy internal error"));

    DocumentClientException thrown = await Assert.ThrowsExceptionAsync<DocumentClientException>(
        () => MetadataRetryHelper.ExecuteAsync<int>(
            (_) => throw new DocumentClientException("original 503",
                HttpStatusCode.ServiceUnavailable, SubStatusCodes.Unknown),
            policy.Object,
            CancellationToken.None));

    StringAssert.Contains(thrown.Message, "original 503");
}

```

xinlian12:
✅ **Review complete** (48:55)

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
NaluTripician: Addressed in 5bdae9f4f. Updated `MetadataRetryHelper` to mirror `ShouldRetryResult.ThrowIfDoneTrying`: when `ExceptionToThrow` references the same instance as the captured exception, we now rethrow via `ExceptionDispatchInfo.Throw()` to preserve the original stack trace; otherwise we throw the policy-translated exception. Used `==` rather than `object.ReferenceEquals` to avoid the CDX1000 boxing analyzer (which forbids casting `Exception` to `object`); reference equality is preserved since `Exception` does not override `op_Equality`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/MetadataRetryHelperTests.cs
NaluTripician: Addressed in 5bdae9f4f. Added `ExecuteAsync_ExceedsMaxAttemptsHardCap_ThrowsInvalidOperationException` which uses a misconfigured policy that always returns `ShouldRetry=true` and asserts (a) `attempts == 20` and (b) the thrown message contains `defensive attempt cap`.

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
NaluTripician: Addressed in 5bdae9f4f. Added `ExecuteAsync_PolicyThrowsDuringShouldRetry_SurfacesOriginalException` which sets the policy to throw `InvalidOperationException` and asserts the helper surfaces the original `DocumentClientException("original 503")`.

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
```diff
@@ -0,0 +1,246 @@
...
+                            capturedException.Throw();
+                        }
+
+                        throw shouldRetry.ExceptionToThrow;
+                    }
+
+                    capturedException.Throw();
+                }
+
+                // If the caller has not cancelled, honor any backoff and continue the loop.
```
xinlian12, line 158: 🟢 **Suggestion — Testing: Mid-backoff cancellation → grace transition is untested**

Lines 159–178 handle a real-world scenario: the caller's token cancels *during* `Task.Delay(shouldRetry.BackoffTime, cancellationToken)`. The `OperationCanceledException` is caught, the second `IsCancellationRequested` check (line 174) evaluates true, and execution falls through to the grace section.

This is arguably the most common production trigger for the grace path — the caller's ~36.5s timeout fires while the helper waits out a retry backoff after a 503 from an unhealthy region. Yet no test covers this transition. All cancellation tests use a pre-cancelled token (`cts.Cancel()` before the first attempt), which means line 159's outer check is immediately false and the backoff + catch block is never exercised.

**Why it matters:** A refactor that accidentally removes the `catch (OperationCanceledException)` block or the second `IsCancellationRequested` check would either surface a spurious OCE to callers or skip the grace retry — both regressions that no test would catch.

**Suggested test:**

```csharp
[TestMethod]
public async Task ExecuteAsync_CancellationDuringBackoff_TransitionsToGraceRetry()
{
    Mock<IDocumentClientRetryPolicy> policy = new();
    policy.Setup(p => p.ShouldRetryAsync(It.IsAny<Exception>(), It.IsAny<CancellationToken>()))
        .ReturnsAsync(ShouldRetryResult.RetryAfter(TimeSpan.FromSeconds(5)));

    using CancellationTokenSource cts = new();
    int attempts = 0;

    int result = await MetadataRetryHelper.ExecuteAsync<int>(
        (ct) =>
        {
            attempts++;
            if (attempts == 1)
            {
                // Cancel mid-backoff: token is live now, will fire during delay
                cts.CancelAfter(TimeSpan.FromMilliseconds(50));
                throw new DocumentClientException("503",
                    HttpStatusCode.ServiceUnavailable, SubStatusCodes.Unknown);
            }
            return Task.FromResult(99);
        },
        policy.Object,
        MetadataRetryHelper.DefaultCrossRegionRetryGrace,
        cts.Token);

    Assert.AreEqual(99, result);
    Assert.AreEqual(2, attempts);
}

```

xinlian12:
✅ **Review complete** (40:03)

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
NaluTripician: Good catch — added `ExecuteAsync_CancellationDuringBackoff_TransitionsToGraceRetry` in 306e025. The token is live at attempt start (so the outer `IsCancellationRequested` check on line 151 is false), then `cts.CancelAfter(50ms)` fires during the 5s `Task.Delay`, exercising the `OperationCanceledException` catch (159-163) and the inner check (166). Also re-asserts the detached-grace-token contract on this code path. Test passes locally.

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
```diff
@@ -0,0 +1,246 @@
...
+                // Always consult the retry policy BEFORE honoring caller cancellation.
+                // This is the correctness fix: BackoffRetryUtility honors cancellation
+                // first, which silently swallows cross-region failover decisions for
+                // metadata reads when the caller's timeout trips during the HTTP
+                // timeout policy escalation.
+                ShouldRetryResult shouldRetry;
+                try
+                {
+                    shouldRetry = await retryPolicy
+                        .ShouldRetryAsync(capturedException.SourceException, CancellationToken.None)
```
kundadebdatta, line 125: `ShouldRetryAsync` called with `CancellationToken.None`: This ensures the policy isn't preempted, but it also means a policy that internally relies on the caller's token for its own timeout won't get it. 

Thinking out loud: Could this have any implication on any unbounded retry attempts ?

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
```diff
@@ -0,0 +1,246 @@
...
+
+                // Caller token is cancelled. Grant a single bounded grace window for
+                // a cross-region retry so availability-critical failover is not
+                // silently preempted. The grace token is passed to the operation so
+                // the underlying HTTP call is not itself preempted by the caller's
+                // already-cancelled token. Subsequent cancellations collapse to the
+                // original exception so we do not extend the caller's timeout unbounded.
+                if (graceAttempted || crossRegionRetryGrace <= TimeSpan.Zero)
+                {
+                    DefaultTrace.TraceInformation(
```
kundadebdatta, line 188: `TraceVerbose` here and below ?

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
NaluTripician: Bounded by two structural guards: `MaxAttemptsHardCap = 20` (covered by `ExecuteAsync_ExceedsMaxAttemptsHardCap_ThrowsInvalidOperationException`) caps total iterations even if a misconfigured policy returns `ShouldRetry=true` indefinitely, and the single-shot `graceAttempted` flag (line 193) ensures at most one cross-region attempt under the bounded `crossRegionRetryGrace` window after caller cancellation.

The `CancellationToken.None` is deliberate — `ShouldRetryAsync` is the policy *decision* call, not a network op. Threading the caller token into it would re-introduce the exact defect this PR fixes (caller cancellation preempts the cross-region failover decision). None of the in-tree retry policies use `ShouldRetryAsync`'s CT for an internal timeout; they use it only for incidental awaits.

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
NaluTripician: Addressed in 29844bdf1 — downgraded the entry-event "granting Nms grace" log to `TraceVerbose` since it's a mid-flow event.

Kept the terminal paths at `TraceInformation`/`TraceWarning`:
- `caller token cancelled; cross-region retry grace already used or disabled. Surfacing original exception.` — terminal failure, stays at Information
- `grace window expired during cross-region retry. Surfacing original exception.` — already TraceWarning
- grace-attempt-failed path — already TraceWarning

These are the events you want in default logs during incident triage when a customer reports a cancellation/hang on metadata reads. Verbose logs are typically off in production.

kundadebdatta (approved):

ananth7592 (commented):
Is 10s hardcoded grace not fundamentally changing the timeout contract? When a customer sets a CancellationToken with a 10s timeout, they
  expect the operation to terminate within that window. With this change, it can silently extend to ~20s.

On the flip side, customers who have intentional fail-fast patterns, cannot opt out of this.

The change looks good but the contract is where my comment is

NaluTripician:
>   Is 10s hardcoded grace not fundamentally changing the timeout contract? When a customer sets a CancellationToken with a 10s timeout, they
>   expect the operation to terminate within that window. With this change, it can silently extend to ~20s.
> 
> On the flip side, customers who have intentional fail-fast patterns, cannot opt out of this.
> 
> The change looks good but the contract is where my comment is

The users request will timeout and cancel with their cancellation token, the metadata request to populate the cache will still go on in the background here. This is because of the customer has an aggressive fail fast policy and there is some gateway delay, the caches will never be populated.

ananth7592 (approved):

Microsoft.Azure.Cosmos/src/MetadataRetryHelper.cs
```diff
@@ -0,0 +1,246 @@
...
+                }
+
+                ExceptionDispatchInfo capturedException;
+                try
+                {
+                    return await operation(cancellationToken).ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    capturedException = ExceptionDispatchInfo.Capture(ex);
```
xinlian12, line 113: 🟢 **Suggestion — Resilience: Add `Task.Yield()` after catch to match BackoffRetryUtility pattern**

`BackoffRetryUtility.ExecuteRetryAsync` includes `await Task.Yield()` immediately after catching the operation exception, with the comment:

> *"this Yield is to 'reset' the stack to avoid stack overflows in Framework and to keep the total size of the StackTrace down if we fail"*

`MetadataRetryHelper` omits this yield. With `MaxAttemptsHardCap = 20` and `TimeSpan.Zero` backoffs, up to 20 async state machine frames could accumulate in the stack trace if the operation throws synchronously on each attempt — making failure diagnostics noisier than they need to be.

**Why it matters:** In practice the risk is minimal — `ReadCollectionAsync` always involves async HTTP I/O, so natural yield points exist. But aligning with the pattern of the exact code being replaced is low-cost insurance, and it prevents a subtle regression if this helper is ever reused for a synchronous operation path in the future.

**Suggested fix:**

```csharp
catch (Exception ex)
{
    await Task.Yield();
    capturedException = ExceptionDispatchInfo.Capture(ex);
}

```

xinlian12:
✅ **Review complete** (25:48)

NaluTripician:
Superseded by #5844, which ships the detached-executor approach as the production fix from a fresh worktree, with magic numbers replaced by documented derivations + ConfigurationManager clamps, the full unit-test matrix per the master work-item template, and a Phase 3b fresh-eyes review (.coding-harness/review-feedback-1.json) addressed.

PR #2491, opened 2021-05-21
Change Feed Processor: Adds support for Graph API accounts

ealsur:
## Description

Graph API accounts are allowed to perform operations through the SQL API and as such, can also read the Change Feed.

Change Feed Processor uses a Lease Container that has, by default, a Partition Key Definition of `/id` . But `/id` is not allowed in Graph API accounts, so Graph API users could not leverage Change Feed Processor even though they could do any SQL API operation.

https://github.com/Azure/azure-documentdb-changefeedprocessor-dotnet/pull/158 added Graph API support to the V2 Change Feed Processor.

This PR aims to port the same capability to V3 SDK to allow for users to migrate from V2 to V3 without any blocking issues.

The PR does not modify the public surface but just enables users to use Lease Containers that are partitioned by `/id` (current behavior) or partitioned by `/partitionKey` (valid in Graph API accounts). The internal implementations will detect the path and generate the extra attribute on the leases in case it is needed.

## Strategic value

Since we are also migrating Azure Functions to V3 SDK (https://github.com/Azure/azure-webjobs-sdk-extensions/pull/717) this will also mean that it would enable Azure Functions Triggers to be used on Graph API accounts, which was not previously possible.

## Key changes

The PR introduces a new implementation of `RequestOptionsFactory`, so based on the Lease Container partition key definition, the CFP builder uses a different Factory implementation. The current `PartitionedByIdCollectionRequestOptionsFactory` or the new `PartitionedByPartitionKeyCollectionRequestOptionsFactory` and the difference is one uses the value of the `id` as `PartitionKey` and the other, the value of the `partitionKey`, and one does not add the `PartitionKey` property to lease documents (default), while the other does.

## Misc

Besides adding new sets of tests to cover the new behavior, the PR adds new tests for pre-existing APIs that had no test coverage and improves test run time (decreasing Emulator tests time) by using ManualResetEvent instead of waiting pre-defined amount of time on CFP tests.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Closing issues

Closes #2291

PR #105, opened 2019-04-03
Stand By Feed and ChangeFeed pull model

ealsur:
# Poll model iterator for Change Feed

This PR introduces a `GetStandByFeedIterator` to `CosmosItems` that will sequentially read the Change Feed for a particular Container.

The sequential read will be done at the `PartitionKeyRange` level, reading each range until a `304` is obtained and continuing with the next range across `FetchNextAsync` calls.

In each iteration, a `ContinuationToken` will be returned that is a composite of a list of continuation tokens for all existing ranges.

The `HasMoreResults` variable is expected to be always true and is on the caller's responsibility to break out of the loop based on a particular logic (ie. when all ranges have been visited or when # amount of ranges have been visited).

Internally, the Iterator will handle the following error scenarios:

1. If the current range being read splits, a refresh of the children will be done and added to the continuation token for continuous read. Current range will be updated for a next call to `FetchNextAsync`.
2. If the current request exceeds Cosmos DB max response size (4Mb) due to the MaxItemSize parameter, the caller will be signaled to retry with `HasMoreResults` true in the response and internally, updating MaxItemSize to half.
3. If the current range does not exist anymore, it will be discarded from the token, and the next one will be available to read through a `FetchNextAsync` call (`HasMoreResults` will be true).

Additionally, when the user provides it's own ContinuationToken (obtained from a previous usage), the following validations will be done:
1. If the first range to read is invalid, it will be discarded, and the next valid one will be updated in the response's Continuation, ready for a `FetchNextAsync`.
2. If all ranges in the provided token are invalid/do not exist, a new token will be built based on the current PK Range distribution in the container will be generated and used.

## Sample usage

    CosmosItemsCore itemsCore = (CosmosItemsCore)this.Container.Items;
    CosmosFeedResultSetIterator setIterator =
        itemsCore.GetStandByFeedIterator(new CosmosChangeFeedRequestOptions() {
           StartTime = DateTime.MinValue
    });
    while (setIterator.HasMoreResults)
    {
        using (CosmosResponseMessage iterator =
            await setIterator.FetchNextSetAsync(this.cancellationToken))
        {
            Trace.WriteLine(iterator.Header.Continuation); // Global Continuation Token
            if (iterator.Content != null)
            {
                Trace.WriteLine(response.Count);
            }
        }
    }

PR #5807, opened 2026-04-22
ChangeFeedProcessor: Fixes lease de-duplication for /partitionKey-partitioned lease containers

kirankumarkolli:
**Research report (gist):** https://gist.github.com/kirankumarkolli/8bf1cbb9c3ee45dfa64511c4a0cb35ad

## Problem statement

In V3 SDK 3.20.0+, when the lease container is partitioned on `/partitionKey` (used for Gremlin/Graph API accounts and any customer-chosen layout), `DocumentServiceLeaseManagerCosmos.CreateLeaseIfNotExistAsync` sets the new lease's partition-key **value** to `Guid.NewGuid().ToString()`. Because Cosmos DB's `id` uniqueness is scoped per logical partition, two racing hosts that compute the same deterministic lease `id` land in different logical partitions and both succeed with `201 Created` instead of the `409 Conflict` the dedup logic relies on.

**Root cause:** V3 PR #2491 (2021-05-27) and its V2 predecessor PR #158 silently violated the original V2 PR #105 (2018) invariant that *"lease collection PK must be a function of monitored collection PK Range. We already have that function as lease id."*

### New env var

| Variable | Default | Purpose |
|---|---|---|
| `AZURE_COSMOS_CHANGE_FEED_LEASE_ID_AS_PARTITION_KEY_ENABLED` | `true` | When `true`, lease docs in a `/partitionKey`-partitioned lease container use the deterministic lease `id` as the partition-key value (restores the 409-based dedup invariant). When `false`, falls back to legacy `Guid.NewGuid()` behavior. |

### Code changes

**`ConfigurationManager.cs`** — New env-var constant and accessor `IsChangeFeedLeaseIdAsPartitionKeyEnabled()`, mirroring the established `AZURE_COSMOS_*_ENABLED` pattern.

**`DocumentServiceLeaseManagerCosmos.cs`** — New `useDeterministicPartitionKey` field snapshotted in the constructor (ensures consistency for the manager's lifetime). New `GetLeasePartitionKeyValue(leaseId)` helper used at both `CreateLeaseIfNotExistAsync` call sites, replacing the `Guid.NewGuid().ToString()` calls.

### Why this is the right fix

| Property | Status after fix |
|---|---|
| PK path | Unchanged (`/partitionKey`) — Gremlin still works |
| Public API surface | Unchanged |
| Default behavior | **New deterministic PK** (fix is on by default) |
| Operational escape hatch | `AZURE_COSMOS_CHANGE_FEED_LEASE_ID_AS_PARTITION_KEY_ENABLED=false` reverts per-process |
| Dedup invariant (PR #105) | **Restored** for all three lease-container layouts |
| Back-compat read path | Unchanged — `TryGetLeaseAsync` uses stored PK value |

### Not required (explicitly)

- No schema migration — old GUID-PK leases coexist with new `id`-PK leases; both read/update paths route through the stored `PartitionKey` field.
- No change to `PartitionedByPartitionKeyCollectionRequestOptionsFactory` (already routes by stored PK value).

## Risks and mitigations

| # | Risk | Likelihood | Mitigation |
|---|---|---|---|
| 1 | Existing customers carry a mix of GUID-PK (old) and `id`-PK (new) leases | Certain | Safe by design — `TryGetLeaseAsync` reads stored `lease.PartitionKey`; old leases remain readable/updatable |
| 2 | Duplicate leases already written under old code persist | Certain | Out of scope — not self-clearing; manual cleanup needed |
| 3 | Mixed-fleet window during rolling upgrade | Certain during rollout | Dedup fully restored only after all hosts upgraded. Not a new failure mode. |
| 4 | Hot-partition risk | Very low | Each lease has a different PK value (per-leaseToken), so distribution is similar to `/id` layout |

### Upgrade semantics — mixed-fleet behavior

During rolling upgrade (old + new SDK hosts coexist):
- Read/update/delete/checkpoint of **existing** leases: works on both sides (routes through stored `lease.PartitionKey`).
- Concurrent **create** of the same missing lease: still produces duplicates if one host is old SDK (GUID PK) and another is new (id PK). **Dedup is fully restored only after rollout completes.**

### Known limitation — pre-existing duplicates not auto-cleaned

Duplicate leases written by older SDK versions persist until the corresponding PKRange undergoes a split/merge. Manual cleanup (`SELECT * FROM c WHERE c.id = @id` + `DELETE` extras) is needed for existing duplicates.

### Unit tests — `DocumentServiceLeaseManagerCosmosTests.cs`

| Test | What it proves |
|---|---|
| `CreateLeaseIfNotExistAsync_PartitionKeyBehavior` (4 DataRows) | Parameterized across PKRange/EPK × default/opt-out: captures the `PartitionKey` sent in the create request and asserts deterministic (PK = lease.Id) or legacy (PK = GUID ≠ lease.Id) behavior |
| `CreateLeaseIfNotExistAsync_DuplicatePkId_Returns409` (2 DataRows) | Simulates Cosmos DB's (PK, id) uniqueness constraint via `ConcurrentDictionary` — first create succeeds, second with same (PK, id) gets 409 |
| `AcquireCompletes_WithPreExistingGuidPartitionKey` | Back-compat: lease with old random-GUID PK can still be acquired (stored PK preserved) |
| `RenewCompletes_WithPreExistingGuidPartitionKey` | Back-compat: lease with old random-GUID PK can still be renewed (stored PK preserved) |
| Existing `CreatesEPKBasedLease` / `CreatesPartitionKeyBasedLease` | **Strengthened:** added `Assert.AreEqual(lease.Id, lease.PartitionKey)` for `/partitionKey` factory |

Infrastructure: `[DoNotParallelize]` + `[TestInitialize]`/`[TestCleanup]` for env var safety. Shared helpers eliminate duplication across overload types.

### Emulator integration test — `GremlinSmokeTests.cs`

| Test | What changed |
|---|---|
| `Schema_DefaultsToHavingPartitionKey` | Converted to `[DataTestMethod]` with two rows: **(a)** default env → asserts `partitionKey == lease.id` (dedup invariant), **(b)** env var `false` → asserts `partitionKey` is a GUID ≠ `lease.id` (legacy behavior) |

Infrastructure: `[DoNotParallelize]` added (env var mutation is process-global).

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
@@ -130,10 +130,9 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                 Mode = this.GetChangeFeedMode()
             };
 
-            string partitionKeyValue = this.useDeterministicPartitionKey
-                ? documentServiceLease.LeaseId
-                : Guid.NewGuid().ToString();
-            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, partitionKeyValue);
+            this.requestOptionsFactory.AddPartitionKeyIfNeeded(
+                (string pk) => documentServiceLease.LeasePartitionKey = pk,
+                this.GetLeasePartitionKeyValue(documentServiceLease.LeaseId));
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
@@ -158,12 +157,18 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                 Mode = this.GetChangeFeedMode()
             };
 
-            string partitionKeyValue = this.useDeterministicPartitionKey
-                ? documentServiceLease.LeaseId
-                : Guid.NewGuid().ToString();
-            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, partitionKeyValue);
+            this.requestOptionsFactory.AddPartitionKeyIfNeeded(
+                (string pk) => documentServiceLease.LeasePartitionKey = pk,
+                this.GetLeasePartitionKeyValue(documentServiceLease.LeaseId));
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
+        }
+
+        private string GetLeasePartitionKeyValue(string leaseId)
+        {
+            return this.useDeterministicPartitionKey
+                ? leaseId
+                : Guid.NewGuid().ToString();
         }
 
         private string GetChangeFeedMode()
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
@@ -132,7 +132,7 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
 
             this.requestOptionsFactory.AddPartitionKeyIfNeeded(
                 (string pk) => documentServiceLease.LeasePartitionKey = pk,
-                GetLeasePartitionKeyValue(IsChangeFeedLeaseIdAsPartitionKeyEnabled, documentServiceLease.LeaseId));
+                DocumentServiceLeaseManagerCosmos.GetLeasePartitionKeyValue(documentServiceLease.LeaseId));
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
@@ -159,14 +159,14 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
 
             this.requestOptionsFactory.AddPartitionKeyIfNeeded(
                 (string pk) => documentServiceLease.LeasePartitionKey = pk,
-                GetLeasePartitionKeyValue(IsChangeFeedLeaseIdAsPartitionKeyEnabled, documentServiceLease.LeaseId));
+                DocumentServiceLeaseManagerCosmos.GetLeasePartitionKeyValue(documentServiceLease.LeaseId));
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
 
-        private static string GetLeasePartitionKeyValue(bool isLeaseIdAsPartitionKeyEnabled, string leaseId)
+        private static string GetLeasePartitionKeyValue(string leaseId)
         {
-            return isLeaseIdAsPartitionKeyEnabled
+            return DocumentServiceLeaseManagerCosmos.IsChangeFeedLeaseIdAsPartitionKeyEnabled
                 ? leaseId
                 : Guid.NewGuid().ToString();
         }
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
         private readonly DocumentServiceLeaseStoreManagerOptions options;
         private readonly RequestOptionsFactory requestOptionsFactory;
         private readonly AsyncLazy<TryCatch<string>> lazyContainerRid;
+        private readonly bool useDeterministicPartitionKey;
         private PartitionKeyRangeCache partitionKeyRangeCache;
 
         public DocumentServiceLeaseManagerCosmos(
@@ -44,6 +45,7 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
             this.options = options;
             this.requestOptionsFactory = requestOptionsFactory;
             this.lazyContainerRid = new AsyncLazy<TryCatch<string>>(valueFactory: (trace, innerCancellationToken) => this.TryInitializeContainerRIdAsync(innerCancellationToken));
+            this.useDeterministicPartitionKey = ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled();
         }
 
         public override async Task<DocumentServiceLease> AcquireAsync(DocumentServiceLease lease)
@@ -128,7 +130,10 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                 Mode = this.GetChangeFeedMode()
             };
 
-            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, Guid.NewGuid().ToString());
+            string partitionKeyValue = this.useDeterministicPartitionKey
+                ? documentServiceLease.LeaseId
+                : Guid.NewGuid().ToString();
+            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, partitionKeyValue);
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
@@ -153,7 +158,10 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                 Mode = this.GetChangeFeedMode()
             };
 
-            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, Guid.NewGuid().ToString());
+            string partitionKeyValue = this.useDeterministicPartitionKey
+                ? documentServiceLease.LeaseId
+                : Guid.NewGuid().ToString();
+            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, partitionKeyValue);
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
@@ -29,8 +29,9 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
         private readonly DocumentServiceLeaseStoreManagerOptions options;
         private readonly RequestOptionsFactory requestOptionsFactory;
         private readonly AsyncLazy<TryCatch<string>> lazyContainerRid;
-        private readonly bool isChangeFeedLeaseIdAsPartitionKeyEnabled;
         private PartitionKeyRangeCache partitionKeyRangeCache;
+
+        internal static bool IsChangeFeedLeaseIdAsPartitionKeyEnabled = ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled();
 
         public DocumentServiceLeaseManagerCosmos(
             ContainerInternal monitoredContainer,
@@ -45,7 +46,6 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
             this.options = options;
             this.requestOptionsFactory = requestOptionsFactory;
             this.lazyContainerRid = new AsyncLazy<TryCatch<string>>(valueFactory: (trace, innerCancellationToken) => this.TryInitializeContainerRIdAsync(innerCancellationToken));
-            this.isChangeFeedLeaseIdAsPartitionKeyEnabled = ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled();
         }
 
         public override async Task<DocumentServiceLease> AcquireAsync(DocumentServiceLease lease)
@@ -132,7 +132,7 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
 
             this.requestOptionsFactory.AddPartitionKeyIfNeeded(
                 (string pk) => documentServiceLease.LeasePartitionKey = pk,
-                GetLeasePartitionKeyValue(this.isChangeFeedLeaseIdAsPartitionKeyEnabled, documentServiceLease.LeaseId));
+                GetLeasePartitionKeyValue(IsChangeFeedLeaseIdAsPartitionKeyEnabled, documentServiceLease.LeaseId));
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
@@ -159,7 +159,7 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
 
             this.requestOptionsFactory.AddPartitionKeyIfNeeded(
                 (string pk) => documentServiceLease.LeasePartitionKey = pk,
-                GetLeasePartitionKeyValue(this.isChangeFeedLeaseIdAsPartitionKeyEnabled, documentServiceLease.LeaseId));
+                GetLeasePartitionKeyValue(IsChangeFeedLeaseIdAsPartitionKeyEnabled, documentServiceLease.LeaseId));
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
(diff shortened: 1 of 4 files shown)


NaluTripician:
@sdkReviewAgent

kirankumarkolli:
@Copilot follow issue agent fix instrucitons and implement it.

xinlian12:
##  Plan Review  ChangeFeed Lease De-duplication Fix

**Overall assessment: The root-cause analysis is excellent and the proposed code fix is correct.** The bug at lines 131 and 156 of `DocumentServiceLeaseManagerCosmos.cs` is real, well-researched, and the fix (using deterministic `LeaseId` as the partition key value) restores the V2 PR #105 invariant. All read/update/delete paths use the stored `lease.PartitionKey`, so old GUID-PK leases remain fully operational. The `DocumentServiceLeaseStoreCosmos` marker/lock documents already use this exact pattern  the fix is restoring consistency, not inventing something new.

However, the **test plan has significant design issues** that should be addressed before implementation begins.

---

###  Finding 1: Proposed emulator test will not trigger the duplicate lease bug

The proposed emulator test starts two `ChangeFeedProcessor` instances concurrently via `Task.WhenAll(p1.StartAsync(), p2.StartAsync())` and expects them to race on `CreateLeaseIfNotExistAsync`. **This will not work** because `BootstrapperCore.InitializeAsync()` serializes lease creation via a lock document:

```
Host A: AcquireInitializationLockAsync()  201 (lock acquired)
Host A: CreateMissingLeasesAsync()  creates all leases
Host A: MarkInitializedAsync() + ReleaseInitializationLockAsync()
Host B: AcquireInitializationLockAsync()  409 (lock held)  sleeps 15s
Host B: IsInitializedAsync()  true  breaks without creating any leases
```

Only one host ever calls `CreateLeaseIfNotExistAsync` during bootstrap. The test will pass on both pre-fix and post-fix code, providing **false confidence**.

**Suggested alternative:** Test the invariant directly at the lease-manager level  construct two `DocumentServiceLeaseManagerCosmos` instances pointing at the same real `/partitionKey`-partitioned lease container, call `CreateLeaseIfNotExistAsync` concurrently for the same `PartitionKeyRange`, then query the container and assert exactly one lease document exists.

---

###  Finding 2: No unit test proves the full 409 dedup chain with the new PK value

The proposed tests verify that the partition key *value* is deterministic and equals the lease ID. This is necessary but **not sufficient**. No proposed test mocks a second concurrent create returning 409 to verify the dedup chain end-to-end. The existing tests at lines 68154 only exercise the happy path (200 OK).

**Suggested addition:** Add a unit test where the mock returns OK on the first `CreateItemStreamAsync` and 409 on the second call for the same lease token. Assert the first returns a lease and the second returns null.

---

###  Finding 3: Missing back-compat test for pre-existing GUID-PK leases

The plan claims old leases with GUID partition keys "remain readable/updatable" because all downstream operations use the stored `lease.PartitionKey`. This is a critical upgrade-safety claim with **no proposed test coverage**.

**Suggested addition:** A unit test that creates a lease with a known GUID PK value (simulating a pre-fix lease), then verifies `TryGetLeaseAsync`, `RenewAsync`, and `AcquireAsync` still route correctly through the stored GUID value.

---

###  Finding 4: Strengthen existing `GremlinSmokeTests` as a cheap regression guard

`GremlinSmokeTests.Schema_DefaultsToHavingPartitionKey` (line 98) already exercises a `/partitionKey`-partitioned lease container but only asserts `Assert.IsNotNull(lease.partitionKey)`. After the fix, add `Assert.AreEqual((string)lease.id, (string)lease.partitionKey)`  a one-line change that makes the existing emulator test validate the dedup invariant. This is more reliable than the concurrent-startup test.

---

###  Finding 5: Emulator test comment about `instanceName` is incorrect

The plan says `.WithInstanceName("host-A") // same host name forces same deterministic lease id`. This is wrong  `instanceName` maps to `options.HostName` which determines **lease ownership**, not the lease ID. Lease ID = `prefix + leaseToken` where prefix includes `processorName`, `cosmosHost`, `dbRid`, `containerRid`  **not** `instanceName`.

---

###  Observation: Env-var toggle for a correctness fix

The env var preserves the ability to re-enable known-broken behavior. The mixed-fleet window during rollout is unavoidable regardless (old/new hosts still race with different PK strategies). The env var only helps if reverting *all* hosts, which means reverting to the bug. The complexity cost is modest but perpetual  the broken code path must be maintained and tested indefinitely. Worth deciding explicitly before implementation.

---

 AI-generated review  may be incorrect. Agree?  resolve the conversation. Disagree?  reply with your reasoning.

xinlian12:
✅ **Review complete** (33:23)

No new comments — existing review coverage is sufficient.

NaluTripician:
Minor readability/refactor nit on the two `CreateLeaseIfNotExistAsync` overloads: the PK-value selection block is duplicated verbatim. Since both sites must stay in lockstep to preserve the dedup invariant, consider extracting a small helper so they can't drift:

```csharp
private string GetLeasePartitionKeyValue(string leaseId)
{
    return this.useDeterministicPartitionKey
        ? leaseId
        : Guid.NewGuid().ToString();
}
```

Each call site then collapses to:

```csharp
this.requestOptionsFactory.AddPartitionKeyIfNeeded(
    (string pk) => documentServiceLease.LeasePartitionKey = pk,
    this.GetLeasePartitionKeyValue(documentServiceLease.LeaseId));
```

Keeps the snapshotted-flag guarantee intact and gives a single source of truth for the "PK value = lease id when deterministic" invariant.

Optional bigger refactor (probably out of scope for this PR): push the policy onto `RequestOptionsFactory` itself as a virtual `GetCreatePartitionKeyValue(string leaseId)`, defaulting to `Guid.NewGuid().ToString()` and overridden by `PartitionedByPartitionKeyCollectionRequestOptionsFactory` to return `leaseId`. That would localize the per-layout behavior where it belongs and remove the env-var branch from the lease manager.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
```diff
@@ -29,6 +29,7 @@ internal sealed class DocumentServiceLeaseManagerCosmos : DocumentServiceLeaseMa
         private readonly DocumentServiceLeaseStoreManagerOptions options;
         private readonly RequestOptionsFactory requestOptionsFactory;
         private readonly AsyncLazy<TryCatch<string>> lazyContainerRid;
+        private readonly bool useDeterministicPartitionKey;
```
kirankumarkolli: make it consistent with the method its initialized with.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
```diff
@@ -153,11 +157,20 @@ public override Task<DocumentServiceLease> CreateLeaseIfNotExistAsync(
...
 
-            this.requestOptionsFactory.AddPartitionKeyIfNeeded((string pk) => documentServiceLease.LeasePartitionKey = pk, Guid.NewGuid().ToString());
+            this.requestOptionsFactory.AddPartitionKeyIfNeeded(
+                (string pk) => documentServiceLease.LeasePartitionKey = pk,
+                this.GetLeasePartitionKeyValue(documentServiceLease.LeaseId));
 
             return this.TryCreateDocumentServiceLeaseAsync(documentServiceLease);
         }
 
+        private string GetLeasePartitionKeyValue(string leaseId)
```
kirankumarkolli: Can it be static?

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
```diff
@@ -44,6 +45,7 @@ public DocumentServiceLeaseManagerCosmos(
             this.options = options;
             this.requestOptionsFactory = requestOptionsFactory;
             this.lazyContainerRid = new AsyncLazy<TryCatch<string>>(valueFactory: (trace, innerCancellationToken) => this.TryInitializeContainerRIdAsync(innerCancellationToken));
+            this.isChangeFeedLeaseIdAsPartitionKeyEnabled = ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled();
```
kirankumarkolli: How about initialize with static ctor instead? Or will it break unit tests which are relying the per instance based initialization?

kirankumarkolli: Static ctor would break unit tests. Tests toggle the env var between runs via \[TestInitialize]\/\[TestCleanup]\ and create fresh \DocumentServiceLeaseManagerCosmos\ instances per test. A static field would read the value once per process, so the opt-out tests (\nvVar="false"\) would fail since the static field retains the value from first initialization.

Per-instance ctor initialization is intentional: it ensures each manager is internally consistent for its lifetime while allowing tests (and operators) to change the env var between manager lifetimes.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/DocumentServiceLeaseManagerCosmosTests.cs
```diff
@@ -18,8 +19,21 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Tests
 
     [TestClass]
     [TestCategory("ChangeFeed")]
+    [DoNotParallelize]
     public class DocumentServiceLeaseManagerCosmosTests
     {
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            DocumentServiceLeaseManagerCosmos.IsChangeFeedLeaseIdAsPartitionKeyEnabled = true;
```
kirankumarkolli: Use ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled() for default in both initialize, cleanup and related sections

kirankumarkolli:
@sdkReviewAgent

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/DocumentServiceLeaseManagerCosmosTests.cs
```diff
@@ -105,6 +119,10 @@ public async Task CreatesEPKBasedLease(int factoryType)
             Assert.AreEqual(feedRangeEpk.Range.Min, ((FeedRangeEpk)epkBasedLease.FeedRange).Range.Min);
             Assert.AreEqual(feedRangeEpk.Range.Max, ((FeedRangeEpk)epkBasedLease.FeedRange).Range.Max);
             ValidateRequestOptionsFactory(requestOptionsFactory, epkBasedLease);
+            if (requestOptionsFactory is PartitionedByPartitionKeyCollectionRequestOptionsFactory)
```
xinlian12: 🟡 **Recommendation — Testing: These assertions are environment-sensitive and redundant**

```csharp
if (requestOptionsFactory is PartitionedByPartitionKeyCollectionRequestOptionsFactory)
{
    Assert.AreEqual(epkBasedLease.Id, epkBasedLease.PartitionKey);
}

```

`TestInitialize` resets `IsChangeFeedLeaseIdAsPartitionKeyEnabled` by re-reading the env var via `ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled()`. If someone runs tests with `AZURE_COSMOS_CHANGE_FEED_LEASE_ID_AS_PARTITION_KEY_ENABLED=false` set in their environment, this assertion fails because `PartitionKey` would be a random GUID.

The same assertions (and more thorough ones) are already covered by the dedicated `CreateLeaseIfNotExistAsync_PartitionKeyBehavior` test, which explicitly pins the flag value before asserting.

**Suggested fix:** Either:
1. Explicitly set `IsChangeFeedLeaseIdAsPartitionKeyEnabled = true` at the start of these tests (if you want to keep the assertions), or
2. Remove these assertions since `CreateLeaseIfNotExistAsync_PartitionKeyBehavior` already covers the deterministic-PK behavior parameterically for both flag states.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
```diff
@@ -31,6 +31,8 @@ internal sealed class DocumentServiceLeaseManagerCosmos : DocumentServiceLeaseMa
         private readonly AsyncLazy<TryCatch<string>> lazyContainerRid;
         private PartitionKeyRangeCache partitionKeyRangeCache;
 
+        internal static bool IsChangeFeedLeaseIdAsPartitionKeyEnabled = ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled();
```
xinlian12, line 34: 🟡 **Recommendation — Design: Consider eliminating the static field in favor of inline reads**

```csharp
internal static bool IsChangeFeedLeaseIdAsPartitionKeyEnabled = ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled();

```

This `internal static` non-`readonly` field is the only ConfigurationManager-backed cached flag in the codebase that is both mutable and `internal`. The two siblings (`PartitionRoutingHelper.IsLengthAwareComparisonEnabled`, `QueryRangeUtils.IsLengthAwareComparisonEnabled`) are both `private static readonly`. The test pattern (direct field mutation) is also unique — 30+ other test files use `Environment.SetEnvironmentVariable` + `try/finally` instead.

**Why it matters:**
- Being `internal` (not `private`), any SDK code or `InternalsVisibleTo` assembly can write to it at runtime, silently reverting the fix process-wide.
- `[DoNotParallelize]` is per-assembly. The unit tests and emulator tests both mutate this same static field from different assemblies — if they ever share a test process, they race.
- Once mutated by a test, the field diverges from the env var — `ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled()` and the field can return different values.

**Suggested alternative — remove the field entirely:**

```csharp
private static string GetLeasePartitionKeyValue(string leaseId)
{
    return ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled()
        ? leaseId
        : Guid.NewGuid().ToString();
}

```

This reads the env var inline each time (matching `IsThinClientEnabled`, `IsBinaryEncodingEnabled`, and similar patterns that read live). `CreateLeaseIfNotExistAsync` involves a Cosmos DB network call, so one `Environment.GetEnvironmentVariable` per creation is negligible overhead. Tests would use `Environment.SetEnvironmentVariable` + `try/finally` — the established pattern.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GremlinSmokeTests.cs
```diff
@@ -75,39 +77,64 @@ public async Task WritesTriggerDelegate_WithLeaseContainer()
...
+                        Assert.IsNull(lease.PartitionId);
                     }
-
-                    Assert.IsNotNull(lease.LeaseToken);
-                    Assert.IsNull(lease.PartitionId);
                 }
-            }
 
-            await processor.StopAsync();
+                await processor.StopAsync();
```
xinlian12, line 131: 🟡 **Recommendation — Testing: `processor.StopAsync()` should be in the `finally` block**

```csharp
await processor.StopAsync();

```

If any assertion or `ReadNextAsync()` throws inside the `try` block, `processor.StopAsync()` is skipped. The `finally` block restores the static field but does not stop the processor. A leaked processor continues polling the emulator, potentially polluting subsequent test runs in the same process.

**Suggested fix:**

```csharp
try
{
    // ... existing code ...
}
finally
{
    await processor.StopAsync();
    DocumentServiceLeaseManagerCosmos.IsChangeFeedLeaseIdAsPartitionKeyEnabled =
        ConfigurationManager.IsChangeFeedLeaseIdAsPartitionKeyEnabled();
}

```

The risk is bounded (container cleanup in `TestCleanup` would orphan the processor), but defensively stopping it in `finally` prevents any background activity between test methods.

xinlian12:
✅ **Review complete** (31:09)

kushagraThapar (commented):
Overall looks good, requesting changes on missing changelog entry.
I am also curious, is there any way to fix the existing leases for existing customers? Maybe some guidance?

kirankumarkolli:
> Overall looks good, requesting changes on missing changelog entry. I am also curious, is there any way to fix the existing leases for existing customers? Maybe some guidance?

.NET changelog is included part of release not with PR. 
No real need to update existing entries and its still functional with no side-affect

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
kirankumarkolli: Reading environments on every call is very expensive.

kirankumarkolli: Setting value directly can be worked-around but marking it readonly will make different process for testing which is not worth for this effort

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/GremlinSmokeTests.cs
kirankumarkolli: Unrelated to this change its an existing pattern. Yes its possible that new asserts might increase the probability.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/DocumentServiceLeaseManagerCosmosTests.cs
kirankumarkolli: @copilot remove this change

kirankumarkolli:

NaluTripician (approved):

jeet1995 (approved):

xinlian12 (approved):
LGTM, thanks

PR #5815, opened 2026-04-28
Read Consistency Strategy: Adds hub region header for LastCommittedSingleWriteRegion strategy.

aavasthy:
# Pull Request Template

## Description
Direct package update to 4.43.1:
Merged PRs:

Update logic for ReadConsistencyStrategy header parsing and fallback improvements
https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB/pullrequest/2075563

Add LastCommittedSingleWriteRegion Read Consistency Strategy
https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB/pullrequest/2089383

DTC Status Codes and SubStatus Codes Generation
https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB/pullrequest/2030762?

This PR introduces a new LastCommittedWriteRegion value to the ReadConsistencyStrategy enum, enabling reads to be routed to the hub (write) region so they always reflect the most recent writes regardless of which region the client is connected to. When this strategy is set either at the client or request level and the operation is a read (point read, query, change feed, or ReadMany), the SDK additionally sets the ShouldProcessOnlyInHubRegion header so the backend routes the request to the write region. The header is only applied for read operations and is not set for writes.

Hedging request with hub region header:
When CrossRegionHedgingAvailabilityStrategy is active, the primary request may discover the hub region mid-flight
(after 2× 404/1002 triggers addHubRegionProcessingOnlyHeader = true in ClientRetryPolicy). Hedged requests are
clones of the original and run with their own ClientRetryPolicy instance, so they would normally repeat the entire
hub discovery cycle independently. To avoid this redundant retry overhead, we introduce a 
CrossRegionAvailabilityContext — a lightweight shared object with a volatile bool
ShouldAddHubRegionProcessingOnlyHeader flag. This context is injected into RequestMessage.Properties before the
clone loop in CrossRegionHedgingAvailabilityStrategy. Since Clone() performs a shallow dictionary copy, all clones
(primary + hedges) share the same CrossRegionAvailabilityContext reference. When the primary's ClientRetryPolicy
sets the hub flag after 2× 404/1002, it also sets the flag on the shared context. Each hedge's 
ClientRetryPolicy.OnBeforeSendRequest reads this shared flag on every attempt and attaches the 
x-ms-cosmos-hub-region-processing-only header immediately — without needing to go through its own 404/1002
discovery.

## Type of change

## Closing issues


NaluTripician:
@sdkReviewAgent-2

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -529,6 +529,10 @@ private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMe
                 requestMessage.Headers.Set(
                     HttpConstants.HttpHeaders.ReadConsistencyStrategy,
                     readConsistencyStrategy.Value.ToString());
+
+                requestMessage.Headers.Set(
```
kushagraThapar: can this get set even for write operations? If so, it might create a side effect.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -529,6 +529,10 @@ private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMe
...
+                    HttpConstants.HttpHeaders.ShouldProcessOnlyInHubRegion,
```
xinlian12: 🟡 **Recommendation — Hub header applies to all operation types (including writes)**

The new hub region header is set unconditionally whenever `ReadConsistencyStrategy` has a value — regardless of the request's `OperationType`. When `ReadConsistencyStrategy` is configured at the **client level** (`CosmosClientOptions.ReadConsistencyStrategy`), every request flowing through `SendAsync` gets this header — including writes (`CreateItemAsync`, `ReplaceItemAsync`, `UpsertItemAsync`, `DeleteItemAsync`, `PatchItemAsync`).

The sibling method `ValidateAndSetConsistencyLevelAsync` explicitly validates against `OperationType` and `ResourceType` via:

```csharp
ValidationHelpers.IsValidConsistencyLevelOverwrite(
    operationType: requestMessage.OperationType,
    resourceType: requestMessage.ResourceType)

```

This method performs no such check. Since `ReadConsistencyStrategy` is semantically a **read** concept (the name, docs, and enum description all say "read and query operations"), routing writes to the hub region is likely unnecessary overhead — and in multi-region write scenarios, it could add latency by forcing writes away from the nearest write region.

**Concrete scenario:** A customer sets `CosmosClientOptions { ReadConsistencyStrategy = Session }` for consistent reads. Every `CreateItemAsync` call now sends `x-ms-cosmos-hub-region-processing-only: True`, potentially routing writes to a distant hub region instead of the nearest one.

**Suggestion:** Either:
1. Add an operation-type guard (skip hub header for write operations), or
2. Add a comment/test documenting that writes intentionally get hub routing when `ReadConsistencyStrategy` is active, and confirm the backend ignores the header on writes.

Note: This is also a cross-SDK divergence — the Java SDK's `getEffectiveReadConsistencyStrategy()` explicitly returns `DEFAULT` for write operations, filtering them out. The .NET approach (proactive hub routing on first request) is already a deliberate divergence from Java's reactive approach (hub routing only after 404/1002 retry), so this may be intentional — but the write-operation case deserves explicit confirmation.

xinlian12:
✅ **Review complete** (55:02)

kushagraThapar (commented):
Now that we always route requests to hub region, what happens if customers set this in case of multi-writer account?

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
FabianMeiswinkel: Hubregion processing should only be triggered by a single ReadConsistencyStartegy enum value - like LastCommittedWriteRegion or similar. Not for all ReadConsistencyStartegies for sure!

FabianMeiswinkel (changes requested):
Hubregion processing should only be triggered by a single ReadConsistencyStartegy enum value - like LastCommittedWriteRegion or similar. Not for all ReadConsistencyStartegies for sure!

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: Hub region header is now only set when ReadConsistencyStrategy == LastCommittedWriteRegion AND the operation is a read (OperationTypeExtensions.IsReadOperation()).
 IsReadOperation includes: Read, ReadFeed, Query, SqlQuery, Head, HeadFeed and MetadataCheckAccess

aavasthy: Hub region header is now only set when ReadConsistencyStrategy == LastCommittedWriteRegion AND the operation is a read (OperationTypeExtensions.IsReadOperation()).
Write operations will never get this header regardless of the strategy configured.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: Hub region header is now only set when ReadConsistencyStrategy == LastCommittedWriteRegion AND the operation is a read (OperationTypeExtensions.IsReadOperation()).
Write operations will never get this header regardless of the strategy configured.

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -48,6 +48,13 @@ enum ReadConsistencyStrategy
         /// Quorum read with GCLSN barrier - returns the latest version across all regions.
         /// Only valid for accounts configured with Strong consistency.
         /// </summary>
-        GlobalStrong = 4
+        GlobalStrong = 4,
+
+        /// <summary>
```
xinlian12, line 53: 🟡 **Recommendation — Documentation: XML doc describes SDK plumbing, not user-facing guarantee**

Every sibling enum value describes **what the user sees** (a consistency guarantee):
- `Eventual`: "reads from a single random secondary replica"
- `Session`: "monotonic reads, writes and read-your-writes guarantees within a session"
- `LatestCommitted`: "returns the latest committed version in the current read region"
- `GlobalStrong`: "returns the latest version across all regions"

The new value describes **how the SDK routes the request** — internal headers, 403/3 status codes, and transparent retries. Users of this public enum shouldn't need to know about `WriteForbidden` retries to understand what they get.

**Why it matters:** This creates an inconsistent API surface. A user scanning the enum sees four values about data freshness and one about HTTP plumbing.

**Suggestion:** Rewrite to describe the guarantee:

```xml
/// <summary>
/// Returns the latest committed version from the hub (write) region, ensuring reads
/// reflect the most recent writes regardless of which region the client is connected to.
/// The SDK handles hub-region routing transparently.
/// </summary>

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
```diff
@@ -4422,6 +4422,328 @@ public async Task ReadItemAsync_ShouldAddHubHeader_OnRetryAfter_404_1002()
...
+            {
+                RequestCallBack = (request, cancellationToken) =>
+                {
+                    if (request.Method == HttpMethod.Get
+                        && request.RequestUri != null
+                        && request.RequestUri.AbsolutePath.Contains("/docs/"))
+                    {
+                        docReadRequestCount++;
+
+                        bool hasHubHeader = request.Headers.TryGetValues(HubRegionHeader, out IEnumerable<string> values)
```
xinlian12: 🟡 **Recommendation — Testing: Hardcoded header string decoupled from production constant**

The emulator tests reference the hub region header via a class-level string literal:

```csharp
private const string HubRegionHeader = "x-ms-cosmos-hub-region-processing-only";

```

Meanwhile, the unit tests in `HandlerTests.cs` correctly use the production constant `HttpConstants.HttpHeaders.ShouldProcessOnlyInHubRegion`.

**Why it matters:** If the production constant's value differs from this string (the constant is named `ShouldProcessOnlyInHubRegion`, which could map to `"x-ms-cosmos-should-process-only-in-hub-region"` — a different string), the emulator tests silently become no-ops: they look for a header that's never set and conclude "not present" which might match certain test assertions. Even if they match today, a future change to the constant won't be caught.

**Suggestion:** Replace the literal with `HttpConstants.HttpHeaders.ShouldProcessOnlyInHubRegion` to stay coupled to the production constant.

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -48,6 +48,13 @@ enum ReadConsistencyStrategy
...
+        /// When set, the SDK automatically adds the hub region processing header so that
+        /// the backend directs the request to the hub region. Non-hub regions respond with
+        /// 403/3 (WriteForbidden), which the SDK retries transparently.
+        /// </summary>
+        LastCommittedWriteRegion = 5
```
xinlian12: 🔴 **Blocking — Contract: Missing preview API contract update**

The `ReadConsistencyStrategy` enum is `public` in `#if PREVIEW` builds, and this PR adds `LastCommittedWriteRegion = 5`. However, the preview API contract file `DotNetPreviewSDKAPI.net6.json` still only lists 4 enum values (lines 1484–1503):

```
Eventual, GlobalStrong, LatestCommitted, Session

```

`LastCommittedWriteRegion` is absent. The `PreviewContractChanges` test compares the compiled DLL's public API surface against this baseline — it will fail on PREVIEW CI builds.

**Why it matters:** This is a CI gate. The PR cannot merge without updating the contract file.

**Fix:** Run `UpdateContracts.ps1` or manually add the `LastCommittedWriteRegion` field entry to `DotNetPreviewSDKAPI.net6.json`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/HandlerTests.cs
```diff
@@ -867,15 +880,17 @@ public async Task ReadConsistencyStrategyClientLevelApplied()
             // Verify client-level ReadConsistencyStrategy is applied when no request-level is set.
             using CosmosClient client = MockCosmosUtil.CreateMockCosmosClient(
                 accountConsistencyLevel: Cosmos.ConsistencyLevel.Strong,
-                customizeClientBuilder: builder => builder.WithReadConsistencyStrategy(Cosmos.ReadConsistencyStrategy.LatestCommitted));
+                customizeClientBuilder: builder => builder.WithReadConsistencyStrategy(Cosmos.ReadConsistencyStrategy.LastCommittedWriteRegion));
```
xinlian12: 🟡 **Recommendation — Testing: Existing test replaced rather than extended**

Per review rules: *"new feature tests should NOT replace existing test coverage for related features."*

This test previously validated `LatestCommitted` at the client level. It now tests `LastCommittedWriteRegion` instead — meaning the only unit test for `LatestCommitted` client-level propagation is gone.

While `LatestCommitted` is still covered at the **request level** via the `DataRow("LatestCommitted")` in `ReadConsistencyStrategyRequestOptionSetsHeaders`, the client-builder → handler fallback path (`this.RequestedClientReadConsistencyStrategy`) is no longer exercised for it.

**Suggestion:** Either parameterize this test over multiple strategies or add back a `LatestCommitted` variant alongside the new `LastCommittedWriteRegion` one:

```csharp
[TestMethod]
[DataRow("LatestCommitted")]
[DataRow("LastCommittedWriteRegion")]
public async Task ReadConsistencyStrategyClientLevelApplied(string strategyName) { ... }

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
```diff
@@ -4422,6 +4422,328 @@ public async Task ReadItemAsync_ShouldAddHubHeader_OnRetryAfter_404_1002()
...
+        /// present from the very first request and persist through all 403/3 retries.
+        /// Only LastCommittedWriteRegion triggers the hub header on reads.
+        /// Parameterized over forbidden count to cover single and multiple 403/3 retries.
+        /// </summary>
+        [TestMethod]
+        [Owner("aavasthy")]
+        [DataRow(1, DisplayName = "LastCommittedWriteRegion request-level: single 403/3 → success")]
+        [DataRow(3, DisplayName = "LastCommittedWriteRegion request-level: multiple 403/3 → success")]
+        [Description("Request-level LastCommittedWriteRegion sets hub header on reads with 403/3 retry flow.")]
+        public async Task ReadItemAsync_WithLastCommittedWriteRegion_RequestLevel_403_3_ThenSuccess(int forbiddenCount)
```
xinlian12: 🟡 **Recommendation — Testing: Feature tests placed in generic file instead of dedicated test class**

A dedicated `ReadConsistencyStrategyTests.cs` already exists in the emulator tests with 6 focused tests covering the ReadConsistencyStrategy feature (point reads, queries, change feed, read-many, override, dual-header rejection). The 4 new emulator tests in this file belong to the same feature area.

**Why it matters:** Feature tests become fragmented across files. A developer extending `ReadConsistencyStrategy` later will find `ReadConsistencyStrategyTests.cs` but miss these tests in `CosmosItemTests.cs`. The dedicated file is also where the `ReadItemWithReadConsistencyStrategy` test is parameterized — `LastCommittedWriteRegion` should be added as a DataRow there too for basic E2E coverage.

**Suggestion:** Move these 4 tests to `ReadConsistencyStrategyTests.cs` and add `[DataRow("LastCommittedWriteRegion")]` to the existing `ReadItemWithReadConsistencyStrategy` test there.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -529,6 +532,14 @@ private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMe
                 requestMessage.Headers.Set(
                     HttpConstants.HttpHeaders.ReadConsistencyStrategy,
                     readConsistencyStrategy.Value.ToString());
+
+                if (readConsistencyStrategy.Value == Cosmos.ReadConsistencyStrategy.LastCommittedWriteRegion
```
xinlian12: 🟡 **Recommendation — Design: Consider interaction with `CrossRegionHedgingAvailabilityStrategy`**

When `CrossRegionHedgingAvailabilityStrategy` and `LastCommittedWriteRegion` are both active, the hedging strategy clones the request (including the `ShouldProcessOnlyInHubRegion` header) and sends copies to multiple regions in parallel. Non-hub regions will return 403/3, which `IsFinalResult()` does NOT treat as final — so each hedged copy retries internally, generating cascading 403/3 traffic.

**Concrete scenario:** 3-region account with hedging enabled + `LastCommittedWriteRegion` read → 3 parallel requests → 2 get 403/3 and retry → generates ~5+ backend requests for one logical read, with wasted RUs and added latency.

**Why it matters:** The combination is semantically contradictory — hedging spreads across regions, but hub-only routing rejects all non-hub regions. This isn't a correctness bug (it eventually converges), but it's wasteful and could confuse diagnostics.

**Suggestion:** Either:
1. Add a check in `ShouldHedge` to skip hedging when `ShouldProcessOnlyInHubRegion` is set, or
2. Document this as an unsupported combination, or
3. At minimum, add a code comment noting this interaction for future maintainers.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/HandlerTests.cs
```diff
@@ -837,6 +838,18 @@ public async Task ReadConsistencyStrategyRequestOptionSetsHeaders(string strateg
                 Assert.AreEqual(strategy.ToString(), request.Headers[HttpConstants.HttpHeaders.ReadConsistencyStrategy]);
                 Assert.IsNull(request.Headers[HttpConstants.HttpHeaders.ConsistencyLevel],
                     "ConsistencyLevel header should not be set when ReadConsistencyStrategy is used");
+
+                if (strategy == Cosmos.ReadConsistencyStrategy.LastCommittedWriteRegion)
```
xinlian12: 🟡 **Recommendation — Testing: No coverage for Query/ChangeFeed/ReadMany with `LastCommittedWriteRegion`**

The PR description states this feature works with `QueryRequestOptions`, `ChangeFeedRequestOptions`, and `ReadManyRequestOptions`, and the handler gates on `OperationTypeExtensions.IsReadOperation()` which includes `Query`, `SqlQuery`, `ReadFeed`, etc. However, **all tests** (both unit and emulator) only exercise point reads (`ReadItemAsync` / `OperationType.Read`).

The existing `ReadConsistencyStrategyTests.cs` has tests for Query, ChangeFeed, and ReadMany — but only with `Session`/`Eventual`/`LatestCommitted`. None use `LastCommittedWriteRegion`.

**Why it matters:** If `IsReadOperation()` doesn't cover some operation type (or if the query pipeline handles headers differently), the hub header wouldn't be set for those operations. Without a test, this goes undetected.

**Suggestion:** Add at least one unit test in `HandlerTests.cs` with `OperationType.Query` + `LastCommittedWriteRegion` to verify the hub header is set. Ideally also add `LastCommittedWriteRegion` to the DataRows in `ReadConsistencyStrategyTests.cs`'s query/changefeed/readmany tests.

xinlian12:
✅ **Review complete** (07:10)

Posted **7** inline comment(s).

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/HandlerTests.cs
aavasthy: updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
aavasthy: updated

kushagraThapar (dismissed):
LGTM, thanks @aavasthy for the changes.

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -48,6 +48,12 @@ enum ReadConsistencyStrategy
         /// Quorum read with GCLSN barrier - returns the latest version across all regions.
         /// Only valid for accounts configured with Strong consistency.
         /// </summary>
-        GlobalStrong = 4
+        GlobalStrong = 4,
+
+        /// <summary>
+        /// Returns the latest committed version from the hub (write) region, ensuring reads
+        /// reflect the most recent writes regardless of which region the client is connected to.
```
FabianMeiswinkel, line 61: Add a comment what exactly is meant with write region (assuming only relevant for single master - so, what happens for multi master accounts when choosing this read consistency strategy)?

FabianMeiswinkel: Should this still work (and redirect to hub region (I think that could be an option - but then we might want to rename it - and also not really seeing the benefit for consistency in MM)

FabianMeiswinkel (changes requested):
Only waiting because I think we need to tighten the contract around what happens when choosing this value in MultMaster account - also, the fact that this is unsupported currently in Gateway mode should be documented/maybe even gated - like returning error when attempted to be used in Gateway mode?

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: Scoped it out specifically for single master accounts and added more comments around this.
Also made sure the LatestCommittedWriteRegion RCS fails with multi-master account.

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -526,9 +534,30 @@ private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMe
 
             if (readConsistencyStrategy.HasValue)
             {
+                // LastCommittedWriteRegion relies on hub-region routing which only applies
+                // to single-master accounts. In multi-master accounts every region is a write
+                // region and there is no partition-set level hub, so reject early with a clear error.
+                if (readConsistencyStrategy.Value == Cosmos.ReadConsistencyStrategy.LastCommittedWriteRegion
+                    && this.client.DocumentClient.UseMultipleWriteLocations)
```
xinlian12: 🔴 **Blocking — Correctness: Multi-master validation bypassed on first request**

The multi-master guard checks `this.client.DocumentClient.UseMultipleWriteLocations`, but this property defaults to `false` and is only set during client initialization inside `EnsureValidClientAsync` (line 87) — which runs **after** `ValidateAndSetReadConsistencyStrategyAsync` (line 82).

On the very first request to a multi-master account with `LastCommittedWriteRegion`:
1. `ValidateAndSetReadConsistencyStrategyAsync` runs → `UseMultipleWriteLocations == false` → guard passes → hub header set
2. `EnsureValidClientAsync` runs → initializes client → sets `UseMultipleWriteLocations = true`
3. Second request → `ValidateAndSetReadConsistencyStrategyAsync` → `UseMultipleWriteLocations == true` → **`ArgumentException` thrown**

The user sees: first request silently succeeds with hub-region routing on a multi-master account (undefined backend behavior), then all subsequent requests throw. This is the exact scenario the guard was added to prevent (per @kushagraThapar's question and @FabianMeiswinkel's feedback).

Compare with the sibling `ValidateAndSetConsistencyLevelAsync` (line 464) which correctly forces initialization via `await this.client.GetAccountConsistencyLevelAsync()` before validating. And `FillMultiMasterContext` (line 450) which reads the same property but runs after `EnsureValidClientAsync` — so it always gets the correct value.

**Fix:** Either move this validation to after `EnsureValidClientAsync`, or add an explicit initialization call before reading `UseMultipleWriteLocations`:

```csharp
if (readConsistencyStrategy.Value == Cosmos.ReadConsistencyStrategy.LastCommittedWriteRegion)
{
    await this.client.DocumentClient.EnsureValidClientAsync(/* trace */);
    if (this.client.DocumentClient.UseMultipleWriteLocations)
    {
        throw new ArgumentException(...);
    }
}

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AvailabilityStrategyUnitTests.cs
```diff
@@ -769,5 +769,48 @@ public async Task HedgeTriggered_HedgeContextContainsMultipleRegions()
...
         }
+
+        /// <summary>
+        /// Verifies that CrossRegionAvailabilityContext propagates the hub region header flag
+        /// across hedged request clones via the shared Properties dictionary.
+        /// This tests the core mechanism: shallow-copy of Properties preserves reference identity,
+        /// so volatile writes by one clone are visible to all others.
+        /// </summary>
+        [TestMethod]
+        public void CrossRegionAvailabilityContext_PropagatesHubHeaderFlagToHedgedRequests()
```
xinlian12, line 780: 🟡 **Recommendation — Testing: No `ClientRetryPolicy` unit tests for new `CrossRegionAvailabilityContext` integration**

This test validates that `Dictionary` shallow-copy preserves the shared reference — but it never instantiates `ClientRetryPolicy`, so it doesn't exercise any of the new code paths in `ClientRetryPolicy.OnBeforeSendRequest` (lines 231-249) or `ShouldRetryOnSessionNotAvailable` (lines 471-481).

The three pieces of conditional logic added to `ClientRetryPolicy` are only tested by the E2E emulator test `AvailabilityStrategy_HedgePicksUpHubHeaderAfter_404_1002_And_403_3`, which is infrastructure-heavy and doesn't isolate the mechanism.

Per review rules: *"DO check that retry policy changes correctly handle the modified code paths. Retry policies are critical infrastructure."*

**Missing unit test scenarios:**
1. When `request.Properties` contains a `CrossRegionAvailabilityContext` with `ShouldAddHubRegionProcessingOnlyHeader = true`, calling `OnBeforeSendRequest` should set the hub header immediately (simulates hedge picking up primary's flag)
2. After 2× 404/1002 on a single-master account, `ShouldRetryOnSessionNotAvailable` should set the shared context's flag to `true` (simulates primary propagating to hedges)
3. When `crossRegionAvailabilityContext` is `null` (non-hedging path), existing behavior is preserved — `addHubRegionProcessingOnlyHeader` alone drives the hub header

These would be additions to `ClientRetryPolicyTests.cs` alongside the existing hub region header tests (`ClientRetryPolicy_HubRegionHeader_AddedOn404_1002_BasedOnAccountType`, etc.).

xinlian12:
✅ **Review complete** (52:07)

NaluTripician (dismissed):
LGTM from hedging side

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: updated this check

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AvailabilityStrategyUnitTests.cs
aavasthy: added three dedicated unit tests in ClientRetryPolicyTests.cs that instantiate ClientRetryPolicy directly
and exercise all three new code paths: (1) hedge picks up hub header immediately from shared context flag, (2)
primary sets shared context flag after 2× 404/1002, and (3) null context (non-hedging) preserves existing local flag
behavior.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/HandlerTests.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
```diff
@@ -48,6 +48,18 @@ enum ReadConsistencyStrategy
...
+        /// This strategy is designed for single-master accounts only.
+        /// In single-master accounts, each physical partition has a designated hub (primary) replica
+        /// in the write region. When a read encounters a 404/1002 (ReadSessionNotAvailable), the SDK
+        /// retries by routing the request to the partition-set level hub region using the
+        /// <c>x-ms-cosmos-hub-region-processing-only</c> header. If the hub replica is in a different
+        /// region than expected (403/3 WriteForbidden), the SDK discovers and retries on the actual hub.
+        /// Returns the latest committed version from the hub (write) region, ensuring reads
+        /// reflect the most recent writes regardless of which region the client is connected to.
+        /// </summary>
+        LastCommittedWriteRegion = 5
```
FabianMeiswinkel: Just an idea - let the native english speaker decide :-) 

Would LastCommitedSingleWriteRegion better hint that there might be a problem fpr MultiMaster - switching form SM to MM would result in hard failures with ArgumentException now?

FabianMeiswinkel (dismissed):
LGTM - just please follow-up on the question regarding naming of the ReadConsistencyStrategy enum value

Microsoft.Azure.Cosmos/src/Resource/Settings/ReadConsistencyStrategy.cs
aavasthy: updated to LastCommitedSingleWriteRegion

FabianMeiswinkel (dismissed):
LGTM

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
```diff
@@ -1,19 +1,23 @@
...
+
 namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests
 {
     using System;
     using System.Collections.Generic;
     using System.Data;
     using System.Linq;
     using System.Net;
+    using System.Net.Http;
+    using System.Text;
```
kundadebdatta, line 10: nit: Do we need this import ?

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -529,11 +540,53 @@ private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMe
...
+            }
+
+            // LastCommitedSingleWriteRegion relies on hub-region routing which only applies
+            // to single-master accounts. In multi-master accounts every region is a write
+            // region and there is no partition-set level hub, so reject with a clear error.
+            if (readConsistencyStrategy.HasValue
+                && readConsistencyStrategy.Value == Cosmos.ReadConsistencyStrategy.LastCommitedSingleWriteRegion
+                && this.client.DocumentClient.UseMultipleWriteLocations)
+            {
+                throw new ArgumentException(
```
kundadebdatta: So, for multi master accounts, we are throwing argument exception, if `LastCommitedSingleWriteRegion` strategy is used and this will fail the entire request correct ?

we already have a `ValidateAndSetReadConsistencyStrategyAsync` method which validates and set the ConsistencyStrategy. Why not leverage the same method to do the validation for multi master accounts ? That way, this method `ValidateReadConsistencyStrategyAccountConstraints` becomes redudant and we can safely remove it.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
aavasthy: yes we use it for HttpClientFactory = () => new HttpClient(httpHandler)

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: updated the code and combined these two: ValidateAndSetReadConsistencyStrategyAsync and ValidateReadConsistencyStrategyAccountConstraints.


And yes if LastCommitedSingleWriteRegion is combined with multi-master accounts we are throwing exception

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -526,9 +530,41 @@ private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMe
 
             if (readConsistencyStrategy.HasValue)
             {
-                requestMessage.Headers.Set(
-                    HttpConstants.HttpHeaders.ReadConsistencyStrategy,
-                    readConsistencyStrategy.Value.ToString());
+                // LastCommittedSingleWriteRegion relies on hub-region routing which only applies
+                // to single-master accounts. In multi-master accounts every region is a write
+                // region and there is no partition-set level hub, so reject with a clear error.
+                if (readConsistencyStrategy.Value == Cosmos.ReadConsistencyStrategy.LastCommittedSingleWriteRegion
```
Meghana-Palaparthi: nit: It might be better to move this check within the `ResourceType.Document` block so that it only throws for document resource types, when the strategy would actually take effect, and doesn't throw for other types.

aavasthy: updated

Meghana-Palaparthi (dismissed):

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -526,9 +530,40 @@ private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMe
 
             if (readConsistencyStrategy.HasValue)
             {
-                requestMessage.Headers.Set(
-                    HttpConstants.HttpHeaders.ReadConsistencyStrategy,
-                    readConsistencyStrategy.Value.ToString());
+                if (requestMessage.ResourceType == ResourceType.Document)
```
NaluTripician, line 533: This looks like it drops the ReadConsistencyStrategy header for all strategies on non document operations. Is this behavior intended?

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
```diff
@@ -526,9 +530,40 @@ private Task ValidateAndSetReadConsistencyStrategyAsync(RequestMessage requestMe
...
+                {
+                    if (readConsistencyStrategy.Value == Cosmos.ReadConsistencyStrategy.LastCommittedSingleWriteRegion)
+                    {
+                        // LastCommittedSingleWriteRegion relies on hub-region routing which only applies
+                        // to single-master accounts. In multi-master accounts every region is a write
+                        // region and there is no partition-set level hub, so reject with a clear error.
+                        if (this.client.DocumentClient.UseMultipleWriteLocations)
+                        {
+                            throw new ArgumentException(
```
NaluTripician, line 542: Would CosmosException with BadRequest be more fitting here?

Directory.Build.props
```diff
@@ -3,7 +3,7 @@
 		<ClientOfficialVersion>3.59.0</ClientOfficialVersion>
 		<ClientPreviewVersion>3.60.0</ClientPreviewVersion>
 		<ClientPreviewSuffixVersion>preview.0</ClientPreviewSuffixVersion>
-		<DirectVersion>3.43.0</DirectVersion>
+		<DirectVersion>3.43.1</DirectVersion>
```
NaluTripician, line 6: Add this to the PR description to make it clear a direct package upgrade is happening, or have this in a different PR.

NaluTripician (changes requested):
A few changes might be needed but other than that LGTM

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: yes this header is for document resource type and read operation

Directory.Build.props
aavasthy: updated PR description

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: same in java where they set it as null or default for non-document operations 
https://github.com/Azure/azure-sdk-for-java/blob/0928217816fbf212221ad5f23a508ba2250c4161/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ReadConsistencyStrategy.java#L113

Microsoft.Azure.Cosmos/src/Handler/RequestInvokerHandler.cs
aavasthy: Going with ArgumentException to stay consistent with ValidateAndSetConsistencyLevelAsync in this same file, which already throws ArgumentException for the same kind of caller-side consistency validation

NaluTripician (dismissed):

kushagraThapar (changes requested):
Requesting changelog entry.

aavasthy:
> Requesting changelog entry.

updated changelog

kushagraThapar (approved):
LGTM, thanks @aavasthy for adding the changelog entry :)

FabianMeiswinkel (approved):
LGTM

Issue #5496, opened 2025-11-14
[Fundamentals] HttpTimeoutPolicy‎: Update the HttpTimeoutPolicyForThinClient to Use More Relaxed Timeouts


kundadebdatta:
Acceptance Criteria:

- Currently the Http Timeout Policy for thin client uses much aggressive timeouts for retry duration:

```
        private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
        {
            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
            (TimeSpan.FromSeconds(1), TimeSpan.Zero),
            (TimeSpan.FromSeconds(5), TimeSpan.Zero),
        };
```

Goal is to update this policy to use much relaxed value. We can start with 1s, 6s and 6s.

- Split the two instances to use for `InstanceShouldThrow503OnTimeoutForNonPointReads` and `InstanceShouldThrow503OnTimeoutForPointReads`.
- Create an environment variable and allow the default 1 sec timeout to override with the new environment variable (if set) if it is aggressive.

PR #5816, opened 2026-04-28
HttpTimeoutPolicy: Fixes aggressive 500ms first-attempt timeout in HttpTimeoutPolicyControlPlaneRetriableHotPath

### LINKED ISSUE #5642: HttpTimeoutPolicyControlPlaneRetriableHotPath 500ms first-attempt timeout causes recurring failures on .NET 10
### Describe the bug

After upgrading from .NET 9 to .NET 10 (no NuGet package changes), we observe that some requests are being cancelled every now and then due to cancellations/timeouts.
  This is happening for several microservices running in Azure App Services - several times daily.
  
  We are using OpenTelemetry in combination with Azure Monitor, and we see these TaskCanceledException's in Azure Monitor with no verifiable source in the StackTrace.
  
  The pattern for these TaskCanceledException's are that they are all cancelled just above the 500ms mark.
  
  The SDK retries successfully, but the initial cancellation is wasteful and produces error noise.
  
  We can see the HttpTimeoutPolicyControlPlaneRetriableHotPath uses a hard-coded 500ms first-attempt timeout.

  The 500ms timeout has existed for a long time, but changes in .NET 10's HTTP connection pool behavior cause requests to exceed 500ms more frequently than on .NET 9.

  ### Environment

  - **SDK Version**: 3.46.0 (timeout value unchanged on main branch)
  - **.NET version**: 10.0.x (works fine on 9.0.x with identical code and packages)
  - **OS**: Linux (also reproduced on Windows)

  ### Reproduction

  1. Run a Cosmos DB Emulator
  2. Insert artifical delay towards Emulator (e.g. traffic shaping using tc on linux)
  3. Create a Cosmos DB client with SDK 3.46.0+ on .NET 10
  4. Run continuous read/write operations - e.g. meta data control plane operations related to changefeed processing or health checks
  5. Observe that some requests are being cancelled and then retried successfully every few seconds
  6. Downgrade to .NET 9
  7. Observe that the cancellation exceptions disappear
  
  ### Verification for root cause

  Cloning the SDK and changing the first-attempt timeout from 500ms to 5s in `HttpTimeoutPolicyControlPlaneRetriableHotPath` eliminates the issue entirely.
  
  See these attached screenshots for verification
  
#### No artifical delay with normal latency and no fix applied - no errors logged
    
Image

 #### Introduced artifical 600ms delay and observe errors occuring
  
Image

  #### Trace detail from error with observed TaskCanceledException and url for the Cosmos operation

Image

##### Stack trace

	System.Threading.Tasks.TaskCanceledException: The operation was canceled.
	 ---> System.IO.IOException: Unable to read data from the transport connection: Operation canceled.
	 ---> System.Net.Sockets.SocketException (125): Operation canceled
	   --- End of inner exception stack trace ---
	   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
	   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource<System.Int32>.GetResult(Int16 token)
	   at System.Net.Security.SslStream.EnsureFullTlsFrameAsync[TIOAdapter](CancellationToken cancellationToken, Int32 estimatedSize)
	   at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
	   at System.Net.Security.SslStream.ReadAsyncInternal[TIOAdapter](Memory`1 buffer, CancellationToken cancellationToken)
	   at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
	   at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
	   --- End of inner exception stack trace ---
	   at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
	   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
	   at System.Net.Http.Metrics.MetricsHandler.SendAsyncWithMetrics(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
	   at System.Net.Http.DiagnosticsHandler.SendAsyncCore(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)

  #### Introduced artifical 600ms delay and change HttpTimeoutPolicyControlPlaneRetriableHotPath to use 5s for first try and observe no errors being logged
  
Image

  ### Related issues

  The SDK team has already relaxed aggressive first-attempt timeouts in other timeout policies:

  - **#5496** — `HttpTimeoutPolicyForThinClient` was updated from `0.5s, 1s, 5s` to `1s, 6s, 6s` because the timeouts were too aggressive
  - **#5484** — `HttpTimeoutPolicyForPartitionFailover` was similarly updated to use more relaxed timeouts for query workloads.

  `HttpTimeoutPolicyControlPlaneRetriableHotPath` still uses the original aggressive `0.5s, 5s, 65s` sequence and should receive the same treatment.

  ### Expected behavior

  The first-attempt timeout for requests should be either:
  1. Long enough to accommodate real-world network conditions (e.g. 6s, matching `HttpTimeoutPolicyForThinClient`)
  2. Configurable via `CosmosClientOptions` or similar Environment Variable

  The current 500ms value is insufficient for .NET 10's connection establishment characteristics and any environment with moderate network latency.
  
  Internally we have a zero tolerance policy for exceptions, so having these TaskCanceledException's eratically trigger our internal triage/alarm systems, without any current posibilities to mitigate is a real headache.

NaluTripician:
## Summary

Raises the first-attempt timeout for `HttpTimeoutPolicyControlPlaneRetriableHotPath` from **500ms to 1s**, aligning with the precedent set by `HttpTimeoutPolicyForThinClient` (#5496) and `HttpTimeoutPolicyForPartitionFailover` (#5484). This eliminates the spurious `TaskCanceledException` retries customers see on .NET 10 and in environments with moderate network latency.

## Issue

Fixes #5642

## Root Cause

`HttpTimeoutPolicyControlPlaneRetriableHotPath.TimeoutsAndDelays` used `(0.5s, 5s, 65s)` as its retry timeout sequence. The 500ms first-attempt budget is too tight for:

1. **.NET 10 HttpConnectionPool changes**: connection establishment / TLS handshake / SslStream initialization frequently takes longer than 500ms on .NET 10 even when the network is healthy.
2. **Real-world latency**: any deployment with moderate network latency (e.g., cross-region, traffic-shaped environments, or Azure App Services with cold connection pools) routinely exceeds 500ms on the very first attempt.

The result is a `TaskCanceledException` on attempt 1, an immediate (successful) retry on attempt 2, and noisy telemetry / wasted work even though nothing is actually failing. The reporter independently verified that raising the first-attempt timeout eliminates the issue.

The same class of bug was previously fixed in two sibling policies:

- **#5496** — `HttpTimeoutPolicyForThinClient`: `(0.5s, 1s, 5s)` → `(1s, 6s, 6s)`
- **#5484** — `HttpTimeoutPolicyForPartitionFailover`: relaxed similarly

`HttpTimeoutPolicyControlPlaneRetriableHotPath` was missed by those PRs and still carries the original aggressive value.

## Fix

Change the first-attempt timeout from `0.5s` to `1s`. Leave the 5s and 65s tail attempts unchanged so the overall retry budget for genuinely slow control-plane operations is preserved.

## Alternatives Considered

- **`(6s, 6s, 10s)` matching `HttpTimeoutPolicyForPartitionFailover`** — rejected because it shrinks the total retry budget from ~70s to ~22s. Hot-path control-plane calls (address resolution under high load) can legitimately need the 65s tail.
- **`(6s, 6s, 65s)` per the reporter''s suggestion** — rejected as a larger behavior change than necessary. The SDK team''s most recent precedent (#5496) settled on `1s` for the first-attempt timeout in similar circumstances. Bumping to 1s removes the spurious cancellations while keeping the change surgical.
- **Make the value configurable** — out of scope for a bug fix; would require API surface changes.

## How the Fix Works

`HttpTimeoutPolicyControlPlaneRetriableHotPath.Instance` is consumed by `GatewayAddressCache` for control-plane address-resolution calls. On each attempt, `CosmosHttpClient.SendHttpAsync` reads the next `(requestTimeout, delayForNextRequest)` tuple from the policy and uses `requestTimeout` as the per-request `CancellationTokenSource` budget. Raising the first tuple''s `requestTimeout` from 500ms to 1s gives the very first request a more realistic budget; if it still times out, the existing retry path runs unchanged.

## Testing

- [x] Existing tests pass: `CosmosHttpClientCoreTests` (17/17 passing locally, 42s)
- [x] Updated `RetryTransientIssuesTestAsync` so the mock delay for the first hot-path attempt (2s) still exceeds the new 1s timeout, preserving the original test intent (verifying that the cancellation token actually fires for each attempt).
- [x] Repro scenario from the issue (artificial 600ms latency producing spurious 500ms cancellations) is covered by the new first-attempt budget.

## Risk Assessment

- **Breaking change:** No
- **Performance impact:** None in the steady state. Minor positive impact: fewer wasted HTTP attempts and retries when the network has moderate latency.
- **Behavior change for existing users:** When the gateway is genuinely slow on the first attempt, callers will now wait up to 1s before the SDK retries, instead of 500ms. Total retry budget is unchanged (~70s).

## Cross-SDK Impact

This is a .NET-only timeout policy class. No equivalent class with the same 500ms hard-coded value was found in the Java/Python/Go/JS/Rust Cosmos SDKs in prior triage of #5496 and #5484. No companion PRs needed.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
@@ -20,9 +20,14 @@ namespace Microsoft.Azure.Cosmos
             this.shouldThrow503OnTimeout = shouldThrow503OnTimeout;
         }
 
+        // The first-attempt timeout was raised from 500ms to 1s to align with HttpTimeoutPolicyForThinClient
+        // (see issue #5642). The original 500ms value caused spurious TaskCanceledException retries on
+        // .NET 10 due to changes in HttpConnectionPool behavior and any environment with moderate network
+        // latency. The 5s and 65s tail attempts are preserved to keep the existing retry budget for slow
+        // control-plane operations.
         private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
+            (TimeSpan.FromSeconds(1), TimeSpan.Zero),
             (TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(1)),
             (TimeSpan.FromSeconds(65), TimeSpan.Zero),
         };
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Azure.Cosmos.Tests
                 }},
                 {HttpTimeoutPolicyControlPlaneRetriableHotPath.Instance,  new List<TimeSpan>()
                 {
-                    TimeSpan.FromSeconds(1),
+                    TimeSpan.FromSeconds(2),
                     TimeSpan.FromSeconds(6),
                     TimeSpan.FromSeconds(66)
                 }},


Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicyControlPlaneRetriableHotPath.cs
```diff
@@ -20,9 +20,14 @@ private HttpTimeoutPolicyControlPlaneRetriableHotPath(bool shouldThrow503OnTimeo
...
 
+        // The first-attempt timeout was raised from 500ms to 1s to align with HttpTimeoutPolicyForThinClient
+        // (see issue #5642). The original 500ms value caused spurious TaskCanceledException retries on
+        // .NET 10 due to changes in HttpConnectionPool behavior and any environment with moderate network
+        // latency. The 5s and 65s tail attempts are preserved to keep the existing retry budget for slow
+        // control-plane operations.
         private readonly IReadOnlyList<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> TimeoutsAndDelays = new List<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)>()
         {
-            (TimeSpan.FromSeconds(.5), TimeSpan.Zero),
+            (TimeSpan.FromSeconds(1), TimeSpan.Zero),
```
Praveen-Msft, line 30: Is it the only place where .5s needs update?

NaluTripician: Yes, this is the only remaining `HttpTimeoutPolicy` with a 0.5s first-attempt timeout. I checked all other policy classes in `Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy*.cs`:

- `HttpTimeoutPolicyDefault`: (65s, 65s, 65s)
- `HttpTimeoutPolicyControlPlaneRead`: (5s, 10s, 20s)
- `HttpTimeoutPolicyForThinClient`: (6s, 6s, 10s) — already updated by #5496
- `HttpTimeoutPolicyForPartitionFailover`: (6s, 6s, 10s) — already updated by #5484
- `HttpTimeoutPolicyNoRetry`: (65s)
- `HttpTimeoutPolicyControlPlaneRetriableHotPath`: (0.5s, 5s, 65s) → (1s, 5s, 65s) **(this PR)**

I also grepped the rest of `Microsoft.Azure.Cosmos/src` for any other `FromMilliseconds(500)` or `FromSeconds(0.5)` usages — the only matches are in XML-doc code examples (e.g. `ConnectionPolicy.cs`, `CosmosClientOptions.cs`), not in any actual timeout configuration.

kundadebdatta (approved):
LGTM.

aavasthy (approved):

PR #5817, opened 2026-04-28
RNTBD Dispatcher: Fixes idle timer thread pool starvation (#4393)

### LINKED ISSUE #4393: [FEATURE REQ] Avoid blocking Asynchronous calls as it leads to thread starvation.

NaluTripician:
Fixes #4393.

Converts `Dispatcher.OnIdleTimer` to async and wires it via
`.ContinueWith(...).Unwrap()` so idle timer callbacks no longer
block thread-pool workers on `Task.Wait()`.

Also includes a small follow-up fix for a pre-existing
`Debug.Assert` race on the Dispose-vs-timer cancellation path
(see "Pre-existing Debug.Assert race fix" section below).

Marked draft because this branch is the staging copy; the
canonical fix will land in msdata/direct via a separate PR
informed by this validation work.

## Validation summary

**The fix (Dispatcher.cs, three edits + one follow-up):**
- `OnIdleTimer` → `OnIdleTimerAsync` (sync → async `Task`)
- New `WaitTaskAsync` alongside the still-present sync `WaitTask`
- `ScheduleIdleTimer` updated to `.ContinueWith(OnIdleTimerAsync, ...).Unwrap()` with safety comment explaining why `.Unwrap()` is essential (without it the timer task completes early and disposal can race the still-running continuation)
- `OnIdleTimerAsync` early-cancellation return now nulls `this.idleTimer` and `this.idleTimerTask` before returning (see "Pre-existing Debug.Assert race fix" below)

The body of `OnIdleTimerAsync` is otherwise identical to the pre-fix `OnIdleTimer` body: lock blocks remain synchronous, no `await` inside any lock, and the existing `Debug.Assert(!Monitor.IsEntered(...))` guards on `callLock` and `connectionLock` are preserved.

**Why this bug is hard to test:** The Dispatcher's idle timer arms with a 750-second minimum (`StoreClientFactory.cs` 600s floor + `Connection.cs` `2*(sendHang+receiveHang)=150s` race buffer). Production starvation requires sustained scale and timing conditions. This is why issue #4393 sat from 2024 to 2026 without a CI catch — single-client tests cannot reliably synthesize the production trigger.

**Evidence:**

| Test | N | Max latency | Thread delta | Result |
|------|---|-------------|--------------|--------|
| Unit test (sync path) | n/a | (blocks calling thread) | n/a | Demonstrates pre-fix behavior |
| Unit test (async path) | n/a | (yields calling thread) | n/a | 5/5 PASS |
| Integration baseline (single dramatic run) | 50 | 48,093ms | 46 | Bug reproduced |
| Integration baseline (6 subsequent runs) | 50, 200 | ~1001ms | 0 | Bug not reproduced |
| Integration fix (sanity run) | 50 | 900ms | 0 | PASS |

- **Unit test** `DispatcherIdleTimerFixTests` (canonical evidence): uses `InternalsVisibleTo` + reflection to exercise `WaitTaskAsync` and `WaitTask` directly with a `TaskCompletionSource` that is never completed. Probe-based thread observation: async path yields the calling thread; sync path blocks it. 5/5 passing runs.
- **Integration test** `RntbdIdleTimerStarvationTests` (regression guard): asserts wiring stays intact (`timersFired > 0` via reflection-based `TraceListener` on `DefaultTrace.TraceSource`, observing the new `TraceInformation` line at `OnIdleTimerAsync` entry) and thread delta stays bounded (`< 10`). Latency assertion was dropped because the ~1001ms ceiling observed across 6 clean runs is a probe-implementation artifact, not a Dispatcher signal.
- **Bug-exists-in-environment evidence:** one integration run during investigation reproduced the production pathology end-to-end on the unpatched baseline at N=50: 48,093ms max probe latency, 46-thread pool growth, 2 probes never executed, `.Wait()` durations median 12.5s / mean 21.2s / p95 64.1s / max 67.6s, 67/110 calls blocked ≥1s. Six subsequent runs (3 at N=50, 3 at N=200) did not reproduce — bug is real in this environment but not reliably triggerable from a single test client at the scales we tested.

## Pre-existing `Debug.Assert` race fix

While reviewing the primary fix, @xinlian12 surfaced a pre-existing race on the Dispose-vs-timer cancellation path ([discussion_r3190324685](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5817#discussion_r3190324685)).

**The race:**
1. Idle timer fires; `OnIdleTimerAsync` continuation is queued but has not yet acquired `connectionLock`.
2. `Dispose()` acquires `connectionLock` first, calls `StartConnectionShutdown()` → `cancellation.Cancel()`, then `StopIdleTimer()` calls `idleTimer.CancelTimer()` which returns `false` (timer already fired). `StopIdleTimer` returns the still-running `idleTimerTask` without nulling `this.idleTimer` / `this.idleTimerTask`.
3. `Dispose()` releases the lock and blocks on `WaitTask(idleTimerTaskCopy, "idle timer")`.
4. `OnIdleTimerAsync` finally acquires the lock, observes `cancellation.IsCancellationRequested == true`, and returns immediately — without clearing the fields.
5. The continuation completes; `Dispose()` unblocks, re-acquires the lock, and trips:
   ```csharp
   Debug.Assert(this.idleTimer == null);
   Debug.Assert(this.idleTimerTask == null);
   ```

**Severity:** Release builds unaffected (`Debug.Assert` is a no-op outside `DEBUG`). Debug builds emit spurious assertion noise during a tight shutdown race window. Pre-existing — same defect existed in the synchronous `OnIdleTimer`.

**Fix:** Three lines in `OnIdleTimerAsync`'s early-cancellation path, mirroring the idle-close path at `Dispatcher.cs:572-573`:
```csharp
if (this.cancellation.IsCancellationRequested)
{
    this.idleTimer = null;
    this.idleTimerTask = null;
    return;
}
```

**Safety:** Both fields are guarded by `connectionLock`, which is held at this point. `Dispose()` already captured `idleTimerTaskCopy = this.idleTimerTask` inside `StopIdleTimer` before releasing the lock — mutating `this.idleTimerTask = null` here doesn't affect what `Dispose()` is awaiting. `cancellation.Cancel()` has only two callers (`Dispose()` and `OnIdleTimerAsync` path D); path D already nulls the fields itself before cancelling, so this null-on-cancel path runs only in the Dispose-initiated race.

**What this PR does NOT do:**
- Does not add `DisposeAsync` / `IAsyncDisposable`
- Does not modify `Channel`, `ChannelDictionary`, `LoadBalancingChannel`, `LoadBalancingPartition`, `LbChannelState`, or `IChannel`
- Does not change `bool disposed` to `int disposed` or add `Interlocked.CompareExchange` guards
- Does not modify any public API surface
- Does not change behavior of the existing sync `WaitTask` (preserved untouched in case other callers exist)
- Only modifies `Dispatcher.cs` plus one new unit test plus one reshaped integration test (and supporting Linux test-run scaffolding under `scripts/`)

**Limitations:**
- Tested only on Linux (matches production bug environment; Windows not tested in this round)
- Test endpoint was a live Azure Cosmos DB account; emulator not validated
- Integration test does not gate on starvation behavior because starvation could not be reliably reproduced at N=50 or N=200 from a single test client
- The single dramatic baseline reproduction is one data point; we did not pursue why that run reproduced and others did not (likely Azure-side timing, network conditions, or backend replica state — outside our control)
- The async disposal path described as Path 2 in issue #4393 is not addressed by this PR and remains a separate follow-up. The `Debug.Assert` race fix above is the only Path-2-adjacent change included here.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
@@ -522,8 +522,12 @@ namespace Microsoft.Azure.Documents.Rntbd
             }
         }
 
-        private void OnIdleTimer(Task precedentTask)
+        private async Task OnIdleTimerAsync(Task precedentTask)
         {
+            DefaultTrace.TraceInformation(
+                "[RNTBD Dispatcher {0}][{1}] Idle timer fired.",
+                this.ConnectionCorrelationId, this);
+
             Task receiveTaskCopy = null;
 
             Debug.Assert(!Monitor.IsEntered(this.connectionLock));
@@ -572,7 +576,8 @@ namespace Microsoft.Azure.Documents.Rntbd
                 receiveTaskCopy = this.CloseConnection();
             }
 
-            this.WaitTask(receiveTaskCopy, "receive loop");
+            await this.WaitTaskAsync(receiveTaskCopy, "receive loop")
+                .ConfigureAwait(false);
         }
 
         // this.connectionLock must be held.
@@ -580,7 +585,16 @@ namespace Microsoft.Azure.Documents.Rntbd
         {
             Debug.Assert(Monitor.IsEntered(this.connectionLock));
             this.idleTimer = this.idleTimerPool.GetPooledTimer((int)timeToIdle.TotalSeconds);
-            this.idleTimerTask = this.idleTimer.StartTimerAsync().ContinueWith(this.OnIdleTimer, TaskContinuationOptions.OnlyOnRanToCompletion);
+            // IMPORTANT: .Unwrap() is essential here. Without it, idleTimerTask
+            // would be Task<Task> and would complete when OnIdleTimerAsync
+            // returns its inner Task (at the first await), not when it
+            // finishes. StopIdleTimer() waits on idleTimerTask during
+            // shutdown; if idleTimerTask completes early, shutdown proceeds
+            // while OnIdleTimerAsync is still running, causing
+            // use-after-dispose on the connection. Do not remove .Unwrap().
+            this.idleTimerTask = this.idleTimer.StartTimerAsync()
+                .ContinueWith(this.OnIdleTimerAsync, TaskContinuationOptions.OnlyOnRanToCompletion)
+                .Unwrap();
             this.idleTimerTask.ContinueWith(
                 failedTask =>
                 {
@@ -676,6 +690,30 @@ namespace Microsoft.Azure.Documents.Rntbd
                 DefaultTrace.TraceWarning(
                     "[RNTBD Dispatcher {0}][{1}] Parallel task failed: {2}. Exception: {3}",
                     this.ConnectionCorrelationId, this, description, e.Message);
+                // Intentionally swallowing the exception. The caller can't
+                // do anything useful with it.
+            }
+        }
+
+        private async Task WaitTaskAsync(Task t, string description)
+        {
+            if (t == null)
+            {
+                return;
+            }
+            try
+            {
+                Debug.Assert(!Monitor.IsEntered(this.callLock));
+                Debug.Assert(!Monitor.IsEntered(this.connectionLock));
+                await t.ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                DefaultTrace.TraceWarning(
+                    "[RNTBD Dispatcher {0}][{1}] Parallel task failed: {2}. " +
+                    "Exception: {3}: {4}",
+                    this.ConnectionCorrelationId, this, description,
+                    e.GetType().Name, e.Message);
                 // Intentionally swallowing the exception. The caller can't
                 // do anything useful with it.
             }
--- a/Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
+++ b/Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
@@ -535,6 +535,16 @@ namespace Microsoft.Azure.Documents.Rntbd
             {
                 if (this.cancellation.IsCancellationRequested)
                 {
+                    // Dispose() raced ahead and StopIdleTimer() lost the
+                    // CancelTimer() race (returned the still-running
+                    // idleTimerTask instead of nulling the fields).
+                    // Clear the fields here under connectionLock so the
+                    // post-wait Debug.Assert in Dispose() sees a clean
+                    // state. Safe: Dispose() captured a local copy of
+                    // idleTimerTask before releasing the lock, so it is
+                    // not affected by this assignment.
+                    this.idleTimer = null;
+                    this.idleTimerTask = null;
                     return;
                 }
 
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/RntbdIdleTimerStarvationTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/RntbdIdleTimerStarvationTests.cs
@@ -1,0 +1,450 @@
+//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Net;
+    using System.Reflection;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    /// <summary>
+    /// Integration tests for the RNTBD idle-timer thread-pool starvation
+    /// fix (issue #4393). Reads connection info from the COSMOS_ENDPOINT
+    /// and COSMOS_KEY environment variables so the tests can run against
+    /// a live account from any Linux/Docker pipeline without touching
+    /// App.config.
+    /// </summary>
+    [TestClass]
+    [TestCategory("RntbdIdleTimer")]
+    public class RntbdIdleTimerStarvationTests
+    {
+        private const string EndpointEnvVar = "COSMOS_ENDPOINT";
+        private const string KeyEnvVar = "COSMOS_KEY";
+
+        private const string TestDatabaseId = "rntbd-starvation-test";
+        private static string GetContainerId(int n) => $"items-{n}";
+
+        private static (string Endpoint, string Key) ReadCredentialsOrSkip()
+        {
+            string endpoint = Environment.GetEnvironmentVariable(EndpointEnvVar);
+            string key = Environment.GetEnvironmentVariable(KeyEnvVar);
+
+            if (string.IsNullOrWhiteSpace(endpoint) || string.IsNullOrWhiteSpace(key))
+            {
+                Assert.Inconclusive(
+                    $"{EndpointEnvVar} and/or {KeyEnvVar} are not set; skipping test that requires a live Cosmos DB account.");
+            }
+
+            return (endpoint, key);
+        }
+
+        [TestMethod]
+        public async Task SmokeTest_CanReachEndpoint()
+        {
+            (string endpoint, string key) = ReadCredentialsOrSkip();
+
+            CosmosClientOptions options = new CosmosClientOptions
+            {
+                ConnectionMode = ConnectionMode.Direct,
+            };
+
+            using CosmosClient client = new CosmosClient(endpoint, key, options);
+
+            AccountProperties account = await client.ReadAccountAsync();
+
+            Assert.IsNotNull(account, "ReadAccountAsync returned null AccountProperties.");
+            Assert.IsFalse(string.IsNullOrWhiteSpace(account.Id), "AccountProperties.Id was empty.");
+        }
+
+        // REGRESSION GUARD for the idle timer fix wiring (issue #4393, PR #5722).
+        //
+        // What this test validates:
+        //   - Idle timers arm and fire correctly when the SDK runs against a
+        //     real Cosmos DB account.
+        //   - The .Unwrap() in Dispatcher.ScheduleIdleTimer continues to track
+        //     OnIdleTimerAsync completion correctly. If .Unwrap() is removed
+        //     in the future, the timer-fire-count assertion catches it (the
+        //     production trace at OnIdleTimerAsync entry stops being emitted
+        //     because the continuation chain breaks).
+        //   - No thread-count explosion or unhandled exceptions during
+        //     idle-fire at N=50 over an ~13-minute window.
+        //
+        // What this test does NOT validate:
+        //   - That the fix prevents thread pool starvation. This was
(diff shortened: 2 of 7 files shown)


NaluTripician:
@sdkReviewAgent-2

scripts/LinuxSmoke/LinuxSmoke.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <!--
+      Cosmos.csproj marks Microsoft.Azure.Cosmos.Direct and Microsoft.HybridRow
+      with PrivateAssets="All" so the native/managed side-assemblies
+      (Microsoft.Azure.Cosmos.Direct.dll, Microsoft.Azure.Cosmos.Core.dll,
+      Microsoft.Azure.Cosmos.Serialization.HybridRow.dll) do not flow
+      transitively through a ProjectReference. Re-reference them here so
+      they land in the test output dir and CosmosClient can load them at
+      runtime in the container.
+    -->
+    <PackageReference Include="Microsoft.Azure.Cosmos.Direct" Version="3.41.3" />
```
xinlian12, line 49: 🟡 **Recommendation — Maintainability: Hardcoded NuGet versions bypass centralized version management**

This project pins `Microsoft.Azure.Cosmos.Direct 3.41.3`, `Microsoft.HybridRow 1.1.0-preview3`, and MSTest `3.6.3` as literal version strings. The main repo uses `Directory.Build.props` with MSBuild properties like `$(DirectVersion)` for centralized version management — this project bypasses all of that.

As `$(DirectVersion)` evolves (currently `3.42.4`), this project silently falls behind, creating version mismatches between what CI tests and what the Linux pipeline tests. The `net6.0` target is also worth revisiting as .NET 6 is past end-of-support.

**Mitigating factor:** The PR description states this branch is a staging copy and the canonical fix will land via a separate PR. If this project is temporary scaffolding that will be removed after the canonical PR merges, this is acceptable. If it persists, it should import `Directory.Build.props` version properties.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/RntbdIdleTimerStarvationTests.cs
```diff
@@ -0,0 +1,450 @@
...
+            Assert.IsNotNull(account, "ReadAccountAsync returned null AccountProperties.");
+            Assert.IsFalse(string.IsNullOrWhiteSpace(account.Id), "AccountProperties.Id was empty.");
+        }
+
+        // REGRESSION GUARD for the idle timer fix wiring (issue #4393, PR #5722).
+        //
+        // What this test validates:
+        //   - Idle timers arm and fire correctly when the SDK runs against a
+        //     real Cosmos DB account.
+        //   - The .Unwrap() in Dispatcher.ScheduleIdleTimer continues to track
```
xinlian12: 🟡 **Recommendation — Test Accuracy: `.Unwrap()` regression claim is incorrect**

The comment block at lines 72–76 states:

> If .Unwrap() is removed in the future, the timer-fire-count assertion catches it (the production trace at OnIdleTimerAsync entry stops being emitted because the continuation chain breaks).

This is not accurate. Removing `.Unwrap()` does **not** prevent `OnIdleTimerAsync` from being invoked. `ContinueWith(this.OnIdleTimerAsync, OnlyOnRanToCompletion)` still calls the delegate regardless of `.Unwrap()`. The `TraceInformation("Idle timer fired.")` at the top of `OnIdleTimerAsync` runs *synchronously* before any `await` point, so it fires in both cases. The `timersFired > 0` assertion would still pass with `.Unwrap()` removed.

**What `.Unwrap()` actually controls** is what `idleTimerTask` tracks: the full async chain (correct) vs. just the outer `Task<Task>` that completes at the first `await` (incorrect — causes a use-after-dispose race during shutdown). The inline comment in `ScheduleIdleTimer` at Dispatcher.cs:588–594 explains this correctly.

**Why this matters:** A future maintainer reading this test might conclude that `.Unwrap()` correctness is guarded by CI. It is not — this assertion guards continuation-chain *reachability*, not `.Unwrap()` *presence*. The real protection against accidental `.Unwrap()` removal is the code comment in `ScheduleIdleTimer`, not this test.

**Suggested fix:** Revise the comment to accurately describe what the assertion validates:

```
//   - The continuation chain from ScheduleIdleTimer still reaches
//     OnIdleTimerAsync. This does NOT validate .Unwrap() correctness
//     (the trace fires before any await point, so it would fire even
//     without .Unwrap()). The .Unwrap() safety is guarded by the
//     inline comment in Dispatcher.ScheduleIdleTimer — see there.

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/RntbdIdleTimerStarvationTests.cs
```diff
@@ -0,0 +1,450 @@
...
+        //   yields thread pool threads where the pre-fix sync path blocked
+        //   them.
+        //
+        // References:
+        //   Issue #4393 (production bug report with stack trace)
+        //   PR #5722 (this fix)
+        //   DispatcherIdleTimerFixTests.cs (canonical fix evidence)
+        [DataTestMethod]
+        [DataRow(50)]
+        public async Task IdleTimerFire_WiringStillFunctional(int connectionCount)
```
xinlian12, line 118: 🟡 **Recommendation — Test Isolation: Process-global thread pool mutation without isolation attributes**

`ThreadPool.SetMinThreads(2, 2)` is a **process-global** mutation that persists for the entire 810+ second test duration. If MSTest runs this class in parallel with other emulator tests, the artificially constrained thread pool could cause unrelated tests to time out or exhibit false starvation symptoms.

No existing tests in this repo call `ThreadPool.SetMinThreads`, so this is a first. The `finally` block restores original values, but the window is ~14 minutes.

Additionally, there is no `[Timeout]` attribute — if the test hangs (e.g., the 810-second `Task.Delay` never completes due to an exception path), it blocks the test runner indefinitely.

**Suggested fix:** Add isolation and safety attributes to the test class or method:

```csharp
[DoNotParallelize]
[Timeout(900_000)] // 15 minutes — generous ceiling for the 810s idle wait

```

The `Assert.Inconclusive` guard mitigates the risk in standard CI (where `COSMOS_ENDPOINT` isn't set), but any pipeline that configures credentials would be affected.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/RntbdIdleTimerStarvationTests.cs
```diff
@@ -0,0 +1,450 @@
...
+
+            using CosmosClient client = new CosmosClient(endpoint, key, options);
+
+            AccountProperties account = await client.ReadAccountAsync();
+
+            Assert.IsNotNull(account, "ReadAccountAsync returned null AccountProperties.");
+            Assert.IsFalse(string.IsNullOrWhiteSpace(account.Id), "AccountProperties.Id was empty.");
+        }
+
+        // REGRESSION GUARD for the idle timer fix wiring (issue #4393, PR #5722).
```
xinlian12: 🟢 **Suggestion — Comment Accuracy: PR number references #5722 but this is PR #5817**

Lines 67 and 109 reference "PR #5722" as "this fix":

```
// REGRESSION GUARD for the idle timer fix wiring (issue #4393, PR #5722).
// ...
//   PR #5722 (this fix)

```

This PR is #5817. If #5722 was an earlier draft that was closed, the references should be updated to point to the current PR number to avoid confusion for future readers.

Microsoft.Azure.Cosmos/src/direct/Dispatcher.cs
```diff
@@ -522,8 +522,12 @@ private void StartIdleTimer()
             }
         }
 
-        private void OnIdleTimer(Task precedentTask)
+        private async Task OnIdleTimerAsync(Task precedentTask)
```
xinlian12, line 525: 💬 **Observation — Pre-existing: `Debug.Assert` race on Dispose-vs-timer cancellation path**

When `Dispose()` and the idle timer fire concurrently and `StopIdleTimer()` loses the cancel race (`CancelTimer()` → `false`), the following sequence occurs:

```mermaid
sequenceDiagram
    participant D as Dispose thread
    participant T as Timer thread (OnIdleTimerAsync)
    D->>D: lock(connectionLock)
    D->>D: StartConnectionShutdown() → cancellation.Cancel()
    D->>D: StopIdleTimer() → CancelTimer()=false → returns idleTimerTask
    D->>D: unlock(connectionLock)
    T->>T: lock(connectionLock)
    T->>T: cancellation.IsCancellationRequested → true
    T->>T: return (does NOT null idleTimer/idleTimerTask)
    T->>T: unlock(connectionLock)
    D->>D: WaitTask unblocks
    D->>D: lock(connectionLock)
    D->>D: Debug.Assert(idleTimer == null) ← FIRES
    D->>D: Debug.Assert(idleTimerTask == null) ← FIRES

```

The early-cancellation return at line 536–538 exits without clearing `this.idleTimer` and `this.idleTimerTask`. Only the idle-close path (lines 572–573) nulls them. `Dispose()` then hits the assertions at lines 475–476.

**Not a regression from this PR** — the same issue existed in the pre-fix `OnIdleTimer`. Release builds are unaffected (asserts are no-ops). Worth tracking for the async disposal follow-up (Path 2 in #4393), where the early-cancellation path could be updated to null out both fields before returning.

xinlian12:
✅ **Review complete** (28:09)

ananth7592 (approved):
LGTM, Except for the few things discussed offline

* First of all ,the docker setup is awesome for local testing of non windows platform features. Ofcourse, it is only for testing this PR and can't be checked in.
* BatchAsyncStreamer has the same missing .Unwrap() bug - Worth noting it in a different bug and fix it later (its latent and not correlated to this behavior )
* The Unit tests plus code changes are the one which will be checked in into msdata PR.

NaluTripician:
> LGTM, Except for the few things discussed offline
> 
> * First of all ,the docker setup is awesome for local testing of non windows platform features. Ofcourse, it is only for testing this PR and can't be checked in.
> * BatchAsyncStreamer has the same missing .Unwrap() bug - Worth noting it in a different bug and fix it later (its latent and not correlated to this behavior )
> * The Unit tests plus code changes are the one which will be checked in into msdata PR.

Adding the docker setup to our tooling repo. 
We can add this to the follow up
Unit tests will be added to msdata repo with that PR

xinlian12 (approved):
LGTM, thanks

NaluTripician:
Msdata PR out, will refer to this PR once Direct package update is installed

PR #5818, opened 2026-04-29
[Internal] DTS: Refactors JSON property names into shared constant and changes etag constant

Meghana-Palaparthi:
# Pull Request Template

## Description

Extract all JSON field name string literals from DistributedTransactionSerializer into `internal const string` constants at the top of the class

Replace the literals in the serializer body with the new constants.

Update all test files that previously hardcoded these same strings in `GetProperty`/`TryGetProperty` calls to reference `DistributedTransactionSerializer.<Constant>`
 instead — ensuring any future rename of a property automatically propagates to tests without a separate search-and-replace:
 - `DistributedTransactionOperationResult.cs`
 - `DistributedTransactionSerializerTests.cs`
 - `DistributedTransactionCommitterTests.cs`
 - `DistributedTransactionTests.cs` (emulator)

 No behavioral changes. All 58 unit tests pass.

Additionally, updates the constant `etag` to `ifMatch` in the request body to align with the wire contract update.

## Type of change

- [✓] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Closing issues

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionSerializer.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionSerializer.cs
@@ -16,6 +16,20 @@ namespace Microsoft.Azure.Cosmos
     /// </summary>
     internal static class DistributedTransactionSerializer
     {
+        internal const string Operations = "operations";
+        internal const string DatabaseName = "databaseName";
+        internal const string CollectionName = "collectionName";
+        internal const string Id = "id";
+        internal const string CollectionResourceId = "collectionResourceId";
+        internal const string DatabaseResourceId = "databaseResourceId";
+        internal const string PartitionKey = "partitionKey";
+        internal const string Index = "index";
+        internal const string ResourceBody = "resourceBody";
+        internal const string SessionToken = "sessionToken";
+        internal const string ETag = "ifMatch";
+        internal const string OperationType = "operationType";
+        internal const string ResourceType = "resourceType";
+
         /// <summary>
         /// Serializes a distributed transaction request body to a JSON stream.
         /// The body contains only the operations array. Other metadata like idempotencyToken,
@@ -32,7 +46,7 @@ namespace Microsoft.Azure.Cosmos
                 jsonWriter.WriteStartObject();
 
                 // operations array
-                jsonWriter.WriteStartArray("operations");
+                jsonWriter.WriteStartArray(Operations);
 
                 foreach (DistributedTransactionOperation operation in operations)
                 {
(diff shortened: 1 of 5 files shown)


kirankumarkolli (approved):

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionSerializer.cs
```diff
@@ -16,6 +16,20 @@ namespace Microsoft.Azure.Cosmos
...
+        internal const string DatabaseName = "databaseName";
+        internal const string CollectionName = "collectionName";
+        internal const string Id = "id";
+        internal const string CollectionResourceId = "collectionResourceId";
+        internal const string DatabaseResourceId = "databaseResourceId";
+        internal const string PartitionKey = "partitionKey";
+        internal const string Index = "index";
+        internal const string ResourceBody = "resourceBody";
+        internal const string SessionToken = "sessionToken";
+        internal const string ETag = "ifMatch";
```
kushagraThapar, line 29: This is interesting, below in the tests we replaced the old test which was literally checking `etag` keyword but this constant replaces it with `IfMatch` keyword. What should be its value?

kushagraThapar (changes requested):
I think the reason the issue I mentioned is not caught in the tests is because the tests are just checking with the constants that were introduced in this file. 
How about updating the tests to actually check for our newly introduced constants with the values being returend from the actual response from the wire?
If we cannot do it yet, maybe checking the constants agains the values we were checking them before?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionSerializer.cs
kushagraThapar: My bad on this, didn't realize the PR fixes the etag value as it was incorrect before.

Meghana-Palaparthi: The wire contract originally required "etag" in the request body, so the tests were using "etag". Now the contract has been updated to "ifMatch". So, `ifMatch` is now the correct value

kushagraThapar:
> I think the reason the issue I mentioned is not caught in the tests is because the tests are just checking with the constants that were introduced in this file. How about updating the tests to actually check for our newly introduced constants with the values being returend from the actual response from the wire? If we cannot do it yet, maybe checking the constants agains the values we were checking them before?

Synced up on this one, we are in process of adding new tests to our service and then we should be able to do end to end integration tests on these constants. 
@Meghana-Palaparthi can you please add a todo on these tests to convert them to integration tests once your service tests are merged in and ready to be used, thanks!

kushagraThapar (approved):

PR #5585, opened 2026-02-01
LINQ: Fixes array.Contains() for .NET 10 MemoryExtensions

kirankumarkolli:
## Description

🤖 **This PR was authored by GitHub Copilot** as part of an automated issue triage and resolution workflow.

Fixes #5518

In .NET 10, `array.Contains(x)` may resolve to `MemoryExtensions.Contains(ReadOnlySpan<T>, T)` instead of `Enumerable.Contains`. Since `ReadOnlySpan<T>` doesn't implement `IEnumerable`, the existing check in `BuiltinFunctionVisitor` fails with `NotSupportedException`.

---

## Issue Summary

| Property | Value |
|----------|-------|
| Issue | #5518 |
| Area | LINQ |
| SDK Version (reported) | 3.x with .NET 10 Preview |
| Severity | P2 |

---

### Code Path
```
BuiltinFunctionVisitor.cs:108 - VisitMethodCall()
  → Checks declaringType.IsEnumerable()
    → ReadOnlySpan<T> returns false (doesn't implement IEnumerable)
      → Falls through to NotSupportedException
```

### Root Cause
In .NET 10, the C# compiler resolves `array.Contains(item)` to `MemoryExtensions.Contains<T>(ReadOnlySpan<T>, T)` instead of `Enumerable.Contains<T>(IEnumerable<T>, T)`. 

The LINQ translator checks `declaringType.IsEnumerable()` to route extension methods, but `ReadOnlySpan<T>` (the first parameter type of `MemoryExtensions.Contains`) doesn't implement `IEnumerable<T>`, causing the method to be unrecognized.

This is a **forward compatibility issue** - existing code works in .NET 8/9 but fails in .NET 10+ previews.

---

### Files Modified

| File | Change |
|------|--------|
| `BuiltinFunctionVisitor.cs` | Added `IsMemoryExtensionsMethod()` helper and routing logic |
| `LinqMemoryExtensionsTests.cs` | Added comprehensive tests for MemoryExtensions detection |

### Code Changes
1. **Added `IsMemoryExtensionsMethod()` helper** - Detects when method is from `System.MemoryExtensions` by checking `DeclaringType.FullName`
2. **Added routing for MemoryExtensions.Contains** - Routes to `ArrayBuiltinFunctions` which already handles both 1-arg and 2-arg Contains signatures
3. **Added 6 unit tests** - Validate MemoryExtensions detection and ensure Enumerable.Contains is unaffected

---

## Generated Output (Before/After)

**Before (incorrect):**
[...]

PR #5477, opened 2025-11-06
Query: Fixes LINQ.Contains to handle Memory Span implicit conversion introduced in C#13/DotNet9

leminh98:
# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed. Include samples if adding new API, and include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] This change requires a documentation update

## Closing issues

To automatically close an issue: closes #IssueNumber

PR #5819, opened 2026-04-29
Linq: Fixes .NET 10 MemoryExtensions.Contains breaking change in LINQ queries

### LINKED ISSUE #5518: No support for Linq expressions using ReadOnlySpan methods

**Describe the bug**
After upgrading to .NET 10 noticed that some filters started throwing `NotSupportedException`. After investigation I narrowed it down to change in the `Contains` method - it's now an extension method on the `ReadOnlySpan` struct, which does not implement `IEnumerable` interface. As it does not implement `IEnumerable` interface, the [BuiltinFunctionVisitor](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/9edd18b149a76cae74b029d8ba4ea71f9e8a4124/Microsoft.Azure.Cosmos/src/Linq/BuiltinFunctions/BuiltinFunctionVisitor.cs#L106C1-L109C14) no longer interprets it as an array.

**To Reproduce**
```csharp
using Microsoft.Azure.Cosmos; // 3.56.0

var client = new CosmosClient("AccountEndpoint=https://test.url;AccountKey=test");

var container = client.GetContainer("a", "a");

var enumerable = new[] { "a" }.AsEnumerable();
var linq1 = container.GetItemLinqQueryable<string>()
    .Where(x => enumerable.Contains(x));

Console.WriteLine(linq1.ToString()); // Prints {"query":"SELECT VALUE root FROM root WHERE (root IN (\"a\"))"}

var array = new[] { "a" };
var linq2 = container.GetItemLinqQueryable<string>()
    .Where(x => array.Contains(x));

Console.WriteLine(linq2.ToString()); // Throws NotSupportedException
```

**Expected behavior**
Should work the same way as with Enumerables.

**Actual behavior**
Throws `NotSupportedException`.

**Environment summary**
SDK Version: 3.56.0
OS Version (e.g. Windows, Linux, MacOSX) - irrelevant.

**Additional context**
Noticed during migration from .NET 8 to .NET 10. `ReadOnlySpan` received some optimizations and methods were rewritten to work with it. Therefore, the `Contains` method was affected.

adityasa:
## Summary
Fixes #5518 - .NET 10 (C# 14) changed `array.Contains(x)` to use `MemoryExtensions.Contains` with `ReadOnlySpan<T>` instead of `Enumerable.Contains`. This breaks the LINQ-to-SQL translator.

## Architecture / Code Flow

```mermaid
flowchart TD
    A["<b>User LINQ Query</b><br/>.Where(x => array.Contains(x))"] --> B["<b>C# Compiler</b>"]
    
    B -->|".NET 6-9"| C["Enumerable.Contains&lt;T&gt;(array, x)"]
    B -->|".NET 10+"| D["MemoryExtensions.Contains&lt;T&gt;(<br/>ReadOnlySpan&lt;T&gt;.op_Implicit(array), x)"]
    
    C --> E["<b>SQLTranslator.TranslateExpression()</b>"]
    D --> E
    
    E --> F["<b>Step 1: ConstantEvaluator.PartialEval</b><br/>Collapse evaluable sub-expressions to constants"]
    F --> G["<b>Step 2: ExpressionToSql</b><br/>Translate expression tree to SQL AST"]
    
    subgraph step1["Step 1: ConstantEvaluator (Nominator - bottom-up)"]
        direction TB
        F1["Visit op_Implicit(array)<br/><i>child node, visited first</i>"]
        F2{"DeclaringType is<br/>ReadOnlySpan or Span?"}
        F3["Block: return false<br/><i>ref struct cannot be boxed</i>"]
        F4["Propagates up:<br/>parent Contains also<br/>not nominated"]
        F1 --> F2
        F2 -->|"Yes"| F3
        F3 --> F4
    end
    
    F --> step1
    step1 --> G
    
    subgraph step2["Step 2: BuiltinFunctionVisitor (routing)"]
        direction TB
        G1{"IsMemoryExtensionsMethod?<br/>DeclaringType == typeof MemoryExtensions<br/>AND Name == Contains"}
        G2["Route to<br/>ArrayBuiltinFunctions"]
        G3["Not matched<br/>(IndexOf, SequenceEqual, etc.)"]
        G4["Throws DocumentQueryException<br/><i>Method X not supported</i>"]
        G1 -->|"Yes"| G2
        G1 -->|"No"| G3
        G3 --> G4
    end
    
    G --> step2
    
    subgraph step3["Step 3: ArrayBuiltinFunctions.VisitImplicit"]
        direction TB
        H1["UnwrapSpanImplicitConversion<br/>Detects op_Implicit on Span types"]
        H2["Extracts original array constant<br/>from args[0].Arguments[0]"]
        H3["Translates to SQL IN clause"]
        H1 --> H2
        H2 --> H3
    end
    
    G2 --> step3
    step3 --> I["<b>Output SQL</b><br/>SELECT * FROM c WHERE c.Name IN ('a', 'b', 'c')"]
```

### Without this fix (on .NET 10)
The `ConstantEvaluator` attempts to evaluate `op_Implicit(array)` which invokes `ReadOnlySpan<T>.op_Implicit(T[])` and tries to box the ref struct result into `Expression.Constant`. This throws `NotSupportedException` which surfaces as a **"Method not supported"** error to the user.

## End-to-End Verification (.NET 10)

Verified with a standalone .NET 10 (`net10.0`, `LangVersion=preview`) console app that references the SDK via `ProjectReference`. This app could not be checked in because CI pipelines do not yet support .NET 9/10, but the verification was run locally against the Cosmos DB Emulator.

**All 8 scenarios pass:**

| # | Scenario | Generated SQL |
|---|----------|--------------|
| 1 | `string[].Contains(x.Field)` | `root["Name"] IN ("Alice", "Bob")` |
| 2 | `int[].Contains(x.Field)` | `root["Score"] IN (42, 99)` |
| 3 | `Guid[].Contains(x.Field)` | `root["Tag"] IN ("aaaa...")` |
| 4 | Full round-trip (insert + query + read) | 1 item returned  |
| 5 | `x.Children.Contains("value")` (array field on doc) | `ARRAY_CONTAINS(root["Children"], "child1")` |
| 6 | `x.Children.Contains` round-trip | 1 item returned  |
| 7 | Customer exact repro: `array.Contains(x)` on simple type | `root IN ("a")` |
| 8 | `enumerable.Contains(x)` still works | `root IN ("a")` |

<details>
<summary>Verification app code (click to expand)</summary>

```csharp
// Temporary .NET 10 verification app for issue #5518 fix
// Target: net10.0, LangVersion=preview
// References Microsoft.Azure.Cosmos.csproj via ProjectReference

using Microsoft.Azure.Cosmos;

const string EndpointUrl = "https://localhost:443";
const string PrimaryKey = "<emulator-key>";
const string DatabaseId = "VerifyLinqNet10";
const string ContainerId = "Items";

int passed = 0, failed = 0;

using CosmosClient client = new(
    accountEndpoint: EndpointUrl,
    authKeyOrResourceToken: PrimaryKey,
    new CosmosClientOptions { ConnectionMode = ConnectionMode.Gateway });

Database database = await client.CreateDatabaseIfNotExistsAsync(DatabaseId);
Container container = await database.CreateContainerIfNotExistsAsync(ContainerId, "/pk");

var item = new TestItem
{
    id = "test-1", pk = "p1", Name = "Alice",
    Score = 42, Tag = Guid.Parse("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"),
    Children = ["child1", "child2", "child3"]
};
await container.UpsertItemAsync(item, new PartitionKey("p1"));

// Test 1: string[].Contains  variable-side, triggers MemoryExtensions path on .NET 10
string[] names = ["Alice", "Bob"];
var q1 = container.GetItemLinqQueryable<TestItem>().Where(x => names.Contains(x.Name));
Console.WriteLine(q1.ToQueryDefinition().QueryText);
// Output: SELECT VALUE root FROM root WHERE (root["Name"] IN ("Alice", "Bob"))

// Test 5: item.Children.Contains  document-field-side
var q5 = container.GetItemLinqQueryable<TestItem>().Where(x => x.Children.Contains("child1"));
Console.WriteLine(q5.ToQueryDefinition().QueryText);
// Output: SELECT VALUE root FROM root WHERE ARRAY_CONTAINS(root["Children"], "child1")

// Test 7: Customer exact repro from issue #5518
var array = new[] { "a" };
var q7 = container.GetItemLinqQueryable<string>().Where(x => array.Contains(x));
Console.WriteLine(q7.ToQueryDefinition().QueryText);
// Output: SELECT VALUE root FROM root WHERE (root IN ("a"))

// Test 8: Enumerable.Contains still works (regression check)
var enumerable = new[] { "a" }.AsEnumerable();
var q8 = container.GetItemLinqQueryable<string>().Where(x => enumerable.Contains(x));
Console.WriteLine(q8.ToQueryDefinition().QueryText);
// Output: SELECT VALUE root FROM root WHERE (root IN ("a"))

await database.DeleteAsync();

class TestItem
{
    public string id { get; set; } = "";
    public string pk { get; set; } = "";
    public string Name { get; set; } = "";
    public int Score { get; set; }
    public Guid Tag { get; set; }
    public string[] Children { get; set; } = [];
}
```
</details>

## Changes
- **ConstantEvaluator.cs**: Exclude `op_Implicit` on `ReadOnlySpan<>/Span<>` types from partial evaluation (ref structs cannot be boxed)
- **ArrayBuiltinFunctions.cs**: `UnwrapSpanImplicitConversion()` unwraps `op_Implicit` to extract the underlying array before translating to SQL
- **BuiltinFunctionVisitor.cs**: Route `MemoryExtensions.Contains` to the array translator (scoped to `Contains` only; other methods produce a clear "method not supported" error)
- **LinqMemoryExtensionsTests.cs**: 12 unit tests covering:
  - ConstantEvaluator exclusions (op_Implicit on Span types blocked, non-Span still evaluable)
  - Supported translations (string/int/Guid arrays, Enumerable.Contains regression)
  - Unsupported methods produce clear errors (IndexOf, SequenceEqual)

## Supersedes
This PR supersedes both #5585 and #5477 with a more targeted fix.

## .NET Breaking Change Context

This issue stems from a **by-design breaking change** in C# 14 / .NET 10 where overload resolution now prefers Span<T>/ReadOnlySpan<T> extension methods ("first-class span" support). The .NET runtime team's mitigation is [OverloadResolutionPriority] on Span overloads so the compiler prefers IEnumerable<T> versions. However, LINQ providers (EF Core, Cosmos DB, etc.) still need to handle the case where the compiler emits the Span-based call.

**Related discussions:**
- [dotnet/runtime#109757](https://github.com/dotnet/runtime/issues/109757)  "First class span's break EFC" (main discussion)
- [dotnet/runtime#109549](https://github.com/dotnet/runtime/issues/109549)  API Proposal: Apply [OverloadResolutionPriority] to Span-based overloads
- [dotnet/efcore#35547](https://github.com/dotnet/efcore/issues/35547)  "Some usages of Contains no longer compile"
- [dotnet/docs#43952](https://github.com/dotnet/docs/issues/43952)  Breaking change documentation
- [Official docs](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/csharp-overload-resolution)  C# 14 overload resolution with span parameters


Microsoft.Azure.Cosmos/src/Linq/ConstantEvaluator.cs
```diff
@@ -53,6 +53,24 @@ private static bool CanBeEvaluated(Expression expression)
                 {
                     return false;
                 }
+
+                // In .NET 10+, array.Contains() resolves to MemoryExtensions.Contains(ReadOnlySpan<T>, T)
+                // which involves an op_Implicit conversion from T[] to ReadOnlySpan<T>.
+                // ReadOnlySpan<T> is a ref struct and cannot be boxed into Expression.Constant,
+                // so we must prevent evaluation of both the MemoryExtensions call and the op_Implicit conversion.
+                if (type != null && type.FullName == "System.MemoryExtensions")
```
adityasa: Can this be rewritten to use typeof check instead of doing name comparison( like in L52) ?  Also merge with that condition?

Microsoft.Azure.Cosmos/src/Linq/BuiltinFunctions/BuiltinFunctionVisitor.cs
```diff
@@ -119,6 +121,16 @@ public static SqlScalarExpression VisitBuiltinFunctionCall(MethodCallExpression
...
         }
 
+        /// <summary>
+        /// Checks if the method is from MemoryExtensions class (e.g., Contains on ReadOnlySpan).
+        /// In .NET 10+, array.Contains() resolves to MemoryExtensions.Contains(ReadOnlySpan, T).
+        /// </summary>
+        private static bool IsMemoryExtensionsMethod(MethodCallExpression methodCallExpression)
+        {
+            Type declaringType = methodCallExpression.Method.DeclaringType;
+            return declaringType != null && declaringType.FullName == "System.MemoryExtensions";
```
adityasa: >declaringType.FullName == "System.MemoryExtensions" [](http://example.com/codeflow?start=44&length=51)

Don't check names, prefer typeof comparsion.

adityasa: Should we only limit this to "supported" memoryextension methods like Contains?

Microsoft.Azure.Cosmos/src/Linq/ConstantEvaluator.cs
```diff
@@ -53,6 +53,24 @@ private static bool CanBeEvaluated(Expression expression)
...
+                }
+
+                if (methodCallExpression.Method.Name == "op_Implicit"
+                    && type != null
+                    && type.IsGenericType
+                    && (type.GetGenericTypeDefinition() == typeof(ReadOnlySpan<>)
+                        || type.GetGenericTypeDefinition() == typeof(Span<>)))
+                {
+                    return false;
+                }
```
adityasa, line 70: Should we limit this check when calling method name is Contains.

Microsoft.Azure.Cosmos/src/Linq/ConstantEvaluator.cs
```diff
@@ -53,6 +53,27 @@ private static bool CanBeEvaluated(Expression expression)
...
+                // op_Implicit converting T[] to ReadOnlySpan<T>/Span<T> cannot be evaluated
+                // because ref structs cannot be boxed. This is the argument to MemoryExtensions.Contains.
+                if (methodCallExpression.Method.Name == "op_Implicit"
+                    && type != null
+                    && type.IsGenericType
+                    && (type.GetGenericTypeDefinition() == typeof(ReadOnlySpan<>)
+                        || type.GetGenericTypeDefinition() == typeof(Span<>)))
+                {
+                    return false;
+                }
```
adityasa, line 70: Do we need this check? Since the wrapped Contains function will be rejected on L62 anyways. Asking since the outer method could be any other method.

adityasa: Yes, this check is required. The Nominator (see Nominator.cs) works **bottom-up**: it visits children first via ase.Visit(expression), then checks CanBeEvaluated on the current node. So the evaluation order is:

1. Nominator visits op_Implicit(array) first (child node)
2. If CanBeEvaluated returns true for it, it gets added to the nomination set
3. SubtreeEvaluator then tries to evaluate the nominated op_Implicit node
4. This invokes ReadOnlySpan<T>.op_Implicit(T[]) and attempts to box the result into Expression.Constant  **NotSupportedException** (ref struct cannot be boxed)

The L62 Contains check would never be reached because the crash happens at step 4 when evaluating the *child* op_Implicit. The parent Contains node is never even nominated (since its other argument x is a parameter, making it non-evaluable by the L47 check in Nominator).

NaluTripician:
@sdkReviewAgent-2

TRStarr (approved):

Microsoft.Azure.Cosmos/src/Linq/ConstantEvaluator.cs
```diff
@@ -53,6 +53,21 @@ private static bool CanBeEvaluated(Expression expression)
...
+                // In .NET 10+, array.Contains() resolves to MemoryExtensions.Contains(ReadOnlySpan<T>, T)
+                // which involves an op_Implicit conversion from T[] to ReadOnlySpan<T>.
+                // ReadOnlySpan<T> is a ref struct and cannot be boxed into Expression.Constant,
+                // so we must prevent evaluation of the op_Implicit argument.
+                // The Nominator processes bottom-up (children first), so blocking op_Implicit
+                // also prevents the parent MemoryExtensions.Contains from being nominated.
+                if (methodCallExpression.Method.Name == "op_Implicit"
+                    && type != null
+                    && type.IsGenericType
+                    && (type.GetGenericTypeDefinition() == typeof(ReadOnlySpan<>)
```
xinlian12, line 66: 🟢 **Suggestion — Hardening: Use `IsByRefLike` for broader ref-struct protection**

The current check matches only `ReadOnlySpan<>` and `Span<>` by generic type definition:

```csharp
&& (type.GetGenericTypeDefinition() == typeof(ReadOnlySpan<>)
    || type.GetGenericTypeDefinition() == typeof(Span<>)))

```

`Expression.Constant()` cannot box **any** ref struct — this is a CLR invariant, not specific to Span types. If a future .NET version introduces another ref struct with an `op_Implicit` that appears in LINQ expression trees, the same `NotSupportedException` crash would recur and need another fix.

Consider using:

```csharp
if (methodCallExpression.Method.Name == "op_Implicit"
    && type != null
    && type.IsByRefLike)
{
    return false;
}

```

This is simpler, correct for all ref structs, and eliminates the need to enumerate specific types. The `IsGenericType` guard can also be dropped since `IsByRefLike` works for both generic and non-generic ref structs.

Note: this suggestion applies specifically to `ConstantEvaluator`. In `UnwrapSpanImplicitConversion`, the narrower type check makes sense because you need to know the inner expression is an array.

Not blocking — the current code is correct for the known problem. This is a hardening improvement.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Linq/LinqMemoryExtensionsTests.cs
```diff
@@ -0,0 +1,464 @@
...
+            Assert.IsTrue(sql.Contains("\"a\""), $"Expected array element 'a' in SQL: {sql}");
+            Assert.IsTrue(sql.Contains("\"b\""), $"Expected array element 'b' in SQL: {sql}");
+            Assert.IsTrue(sql.Contains("\"c\""), $"Expected array element 'c' in SQL: {sql}");
+        }
+
+        /// <summary>
+        /// End-to-end: MemoryExtensions.Contains with int[] produces correct SQL IN clause.
+        /// </summary>
+        [TestMethod]
+        public void Translate_MemoryExtensionsContains_IntArray_ProducesInClause()
```
xinlian12, line 249: 🟢 **Suggestion — Tests: Use shared helper for consistency**

The int[] and Guid[] translation tests manually construct the expression tree (resolving `MemoryExtensions.Contains` method, calling `Expression.Call(opImplicit, ...)`, etc.), while the string[] test uses the shared `BuildNet10ContainsExpression<T>()` helper.

Using the helper consistently would:
- Reduce ~15 lines of duplicated reflection logic per test
- Ensure all translation tests exercise the same expression construction path — if the helper has a subtle bug, it would be caught by all three tests rather than only the string test

```csharp
// Instead of manual construction:
MethodInfo containsMethod = typeof(MemoryExtensions)...
MethodCallExpression spanConversion = Expression.Call(opImplicit, arrayConst);
MethodCallExpression containsCall = Expression.Call(containsMethod, spanConversion, paramX);

// Use:
MethodCallExpression containsCall = BuildNet10ContainsExpression<int>(arrayConst, paramX);

```

Minor code quality nit — not blocking.

xinlian12:
✅ **Review complete** (38:00)

Microsoft.Azure.Cosmos/src/Linq/ConstantEvaluator.cs
adityasa: By design, since I am intending to keep the fix _very_ scoped for now.

kirankumarkolli (dismissed):

neildsh (dismissed):

neildsh (approved):

sboshra (approved):

khdang (approved):

PR #5823, opened 2026-04-30
LocationCache: Fixes read fallback to use WriteEndpoints[0] when PPAF enabled and all regions excluded

### LINKED ISSUE #5821: Reads routed via defaultEndpoint do not failover after write region switch when ExcludeRegions filters all preferred regions
## Summary

When `ApplicationPreferredRegions == ExcludeRegions`, the SDK falls back to the region-agnostic `defaultEndpoint` for reads. After a write region switch, writes failover correctly (they use `AvailableWriteLocations[0]` directly), but reads remain pinned to the old region's address resolution because the `GlobalAddressResolver`'s cached `EndpointCache` for the default endpoint has a **stale `location` string** frozen at client init time.

## Root Cause

There are two layers to this bug:

### Layer 1: `LocationCache.GetApplicableEndpoints` uses static `defaultEndpoint` as read fallback

[LocationCache.cs#L372-L388](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/71d317ffc6b6d62199b35c06a372825849e406fc/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L372-L388)

```csharp
public ReadOnlyCollection<Uri> GetApplicableEndpoints(DocumentServiceRequest request, bool isReadRequest)
{
    // ...
    return GetApplicableEndpoints(
        isReadRequest ? this.locationInfo.AvailableReadEndpointByLocation : this.locationInfo.AvailableWriteEndpointByLocation,
        effectivePreferredLocations,
        this.defaultEndpoint,    // ← BUG: Static, region-agnostic, never updated
        request.RequestContext.ExcludeRegions);
}
```

The private static helper at [L416-L449](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/71d317ffc6b6d62199b35c06a372825849e406fc/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L416-L449) uses this fallback when no endpoints survive filtering:

```csharp
if (applicableEndpoints.Count == 0)
{
    applicableEndpoints.Add(fallbackEndpoint); // ← "myaccount.documents.azure.com"
}
```

**Key insight**: The fix already exists in `UpdateLocationCache`! At [L756-L760](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/71d317ffc6b6d62199b35c06a372825849e406fc/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L756-L760), `ReadEndpoints` correctly uses `WriteEndpoints[0]` as fallback:

```csharp
nextLocationInfo.ReadEndpoints = this.GetPreferredAvailableEndpoints(
    endpointsByLocation: nextLocationInfo.AvailableReadEndpointByLocation,
    orderedLocations: nextLocationInfo.AvailableReadLocations,
    expectedAvailableOperation: OperationType.Read,
    fallbackEndpoint: nextLocationInfo.WriteEndpoints[0]);  // ← Dynamic, correct!
```

But `GetApplicableEndpoints` (the ExcludeRegions path) bypasses this and uses `this.defaultEndpoint` instead.

### Layer 2: `GlobalAddressResolver.GetOrAddEndpoint` caches stale `AddressResolver.location`

[GlobalAddressResolver.cs#L327-L336](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/71d317ffc6b6d62199b35c06a372825849e406fc/Microsoft.Azure.Cosmos/src/Routing/GlobalAddressResolver.cs#L327-L336) — Once an endpoint is cached at init, `TryGetValue` returns immediately without validating location:

```csharp
if (this.addressCacheByEndpoint.TryGetValue(endpoint, out EndpointCache existingCache))
{
    return existingCache; // ← Never checks if location drifted
}
```

[AddressResolver.cs#L34](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/71d317ffc6b6d62199b35c06a372825849e406fc/Microsoft.Azure.Cosmos/src/Routing/AddressResolver.cs#L34) — `location` is `private readonly string`, frozen at creation.

[AddressResolver.cs#L72](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/71d317ffc6b6d62199b35c06a372825849e406fc/Microsoft.Azure.Cosmos/src/Routing/AddressResolver.cs#L72) — Propagates stale value: `request.RequestContext.RegionName = this.location`

## Proposed Fix

**Change `GetApplicableEndpoints` to use `WriteEndpoints[0]` as the read fallback instead of `this.defaultEndpoint`.**

This aligns with:
- The existing pattern in `UpdateLocationCache` (L760) which already uses `WriteEndpoints[0]` for `ReadEndpoints`
- Java SDK: [`LocationCache.java#L266`](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/routing/LocationCache.java#L266) — `writeRegionalRoutingContexts.get(0)`
- Python SDK: [`_location_cache.py#L241`](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/cosmos/azure-cosmos/azure/cosmos/_location_cache.py#L241) — `get_write_regional_routing_contexts()[0]`

```csharp
public ReadOnlyCollection<Uri> GetApplicableEndpoints(DocumentServiceRequest request, bool isReadRequest)
{
    if (request.RequestContext.ExcludeRegions == null || request.RequestContext.ExcludeRegions.Count == 0)
    {
        return isReadRequest ? this.ReadEndpoints : this.WriteEndpoints;
    }

    DatabaseAccountLocationsInfo databaseAccountLocationsInfoSnapshot = this.locationInfo;
    ReadOnlyCollection<string> effectivePreferredLocations = databaseAccountLocationsInfoSnapshot.EffectivePreferredLocations;

    Uri fallbackEndpoint = isReadRequest
        ? databaseAccountLocationsInfoSnapshot.WriteEndpoints[0]  // Dynamic: tracks current write region
        : this.defaultEndpoint;

    return GetApplicableEndpoints(
        isReadRequest ? this.locationInfo.AvailableReadEndpointByLocation : this.locationInfo.AvailableWriteEndpointByLocation,
        effectivePreferredLocations,
        fallbackEndpoint,
        request.RequestContext.ExcludeRegions);
}
```

## Why Read-Path Only

- **Writes** in single-master bypass `ExcludeRegions` entirely at [L347-348](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/71d317ffc6b6d62199b35c06a372825849e406fc/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L347-L348) — they use `AvailableWriteLocations` directly
- **Reads** hit `GetApplicableEndpoints` → all filtered → fallback to `defaultEndpoint` → stale `AddressResolver.location`

## Impact

| What | Before Fix | After Fix |
|------|-----------|-----------|
| Read fallback endpoint | `defaultEndpoint` (region-agnostic, static) | `WriteEndpoints[0]` (region-specific, dynamic) |
| `AddressResolver.location` | Stale after hub switch | Correct (cache keyed by regional URI, recreated) |
| `request.RequestContext.RegionName` | Reports wrong region | Reports correct region |
| Diagnostics | Wrong region in traces | Correct region |
| Per-partition routing | May make wrong decisions | Correct decisions |

## Cross-SDK Comparison

| SDK | Read Fallback When All Excluded | Correct? |
|-----|-------------------------------|----------|
| **.NET (current)** | `this.defaultEndpoint` | ❌ Static, stale |
| **.NET (proposed)** | `WriteEndpoints[0]` | ✅ Dynamic |
| **Java** | `writeRegionalRoutingContexts.get(0)` | ✅ Dynamic |
| **Python** | `get_write_regional_routing_contexts()[0]` | ✅ Dynamic |
| **Rust** | `self.default_endpoint` (gateway-only, no AddressResolver) | ⚠️ [azure-sdk-for-rust#4322](https://github.com/Azure/azure-sdk-for-rust/issues/4322) |

## Reproduction Scenario

1. Configure client with `ApplicationPreferredRegions = ["East US"]`
2. Send reads with `ExcludeRegions = ["East US"]`
3. Reads fall back to `defaultEndpoint` → `AddressResolver.location = "East US"` (init-time write region)
4. Trigger write region failover: East US → West US
5. `WriteEndpoints[0]` now points to West US regional endpoint (correctly updated)
6. **But** `GetApplicableEndpoints` still returns `defaultEndpoint` → stale cache hit in `GlobalAddressResolver`
7. `request.RequestContext.RegionName` incorrectly reports "East US"

ananth7592:
## Problem

When ApplicationPreferredRegions == ExcludeRegions, LocationCache.GetApplicableEndpoints falls back to 	his.defaultEndpoint — a static, region-agnostic URI set once at init and never updated. After a write region (hub) switch, the GlobalAddressResolver's cached EndpointCache for this default endpoint has a stale AddressResolver.location, causing incorrect region tracking in diagnostics, per-partition routing, and retry logic.

## Fix

When PPAF (IsPartitionLevelFailoverEnabled) is enabled, GetApplicableEndpoints now uses WriteEndpoints[0] (dynamic, tracks current write region) as the read fallback instead of 	his.defaultEndpoint.

This aligns with:
- UpdateLocationCache (L756-760) which already uses WriteEndpoints[0] for ReadEndpoints fallback
- Java SDK: writeRegionalRoutingContexts.get(0)
- Python SDK: get_write_regional_routing_contexts()[0]

### PPAF Gating
The fix is gated behind Func<bool> isPartitionLevelFailoverEnabled wired from ConnectionPolicy.EnablePartitionLevelFailover through GlobalEndpointManager, supporting dynamic enablement per PR #5310. When PPAF is disabled, original behavior (defaultEndpoint fallback) is preserved.

## Changes
- **LocationCache.cs**: Added isPartitionLevelFailoverEnabled parameter; gated read fallback behind it
- **GlobalEndpointManager.cs**: Wires ConnectionPolicy.EnablePartitionLevelFailover into LocationCache
- **LocationCacheTests.cs**: 3 new tests covering PPAF on/off/dynamic toggle scenarios

## Testing
All 94 LocationCacheTests pass.

Fixes #5821

### DIFF
--- a/Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
@@ -79,7 +79,7 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// List of regions to be excluded routing the request to.
         /// This can be used to route a request to a specific region by excluding all other regions.
-        /// If all regions are excluded, then the request will be routed to the primary/hub region.
+        /// If all regions are excluded, the SDK will route the request on a best-effort basis to maintain availability.
         /// </summary>
         public List<string> ExcludeRegions { get; set; }
 
--- a/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Azure.Cosmos.Routing
         private readonly int connectionLimit;
         private readonly ConcurrentDictionary<Uri, LocationUnavailabilityInfo> locationUnavailablityInfoByEndpoint;
         private readonly RegionNameMapper regionNameMapper;
+        private readonly Func<bool> isPartitionLevelFailoverEnabled;
 
         private DatabaseAccountLocationsInfo locationInfo;
         private DateTime lastCacheUpdateTimestamp;
@@ -39,13 +40,15 @@ namespace Microsoft.Azure.Cosmos.Routing
             Uri defaultEndpoint,
             bool enableEndpointDiscovery,
             int connectionLimit,
-            bool useMultipleWriteLocations)
+            bool useMultipleWriteLocations,
+            Func<bool> isPartitionLevelFailoverEnabled = null)
         {
             this.locationInfo = new DatabaseAccountLocationsInfo(preferredLocations, defaultEndpoint);
             this.defaultEndpoint = defaultEndpoint;
             this.enableEndpointDiscovery = enableEndpointDiscovery;
             this.useMultipleWriteLocations = useMultipleWriteLocations;
             this.connectionLimit = connectionLimit;
+            this.isPartitionLevelFailoverEnabled = isPartitionLevelFailoverEnabled;
 
             this.lockObject = new object();
             this.locationUnavailablityInfoByEndpoint = new ConcurrentDictionary<Uri, LocationUnavailabilityInfo>();
@@ -380,10 +383,18 @@ namespace Microsoft.Azure.Cosmos.Routing
 
             ReadOnlyCollection<string> effectivePreferredLocations = databaseAccountLocationsInfoSnapshot.EffectivePreferredLocations;
 
+            // For reads when PPAF is enabled, use WriteEndpoints[0] as fallback (dynamic,
+            // tracks current write region) instead of this.defaultEndpoint (static, region-agnostic,
+            // never updated after init). This aligns with UpdateLocationCache which already uses
+            // WriteEndpoints[0] as the ReadEndpoints fallback, and matches Java/Python SDK behavior.
+            Uri fallbackEndpoint = (isReadRequest && this.isPartitionLevelFailoverEnabled?.Invoke() == true)
+                ? databaseAccountLocationsInfoSnapshot.WriteEndpoints[0]
+                : this.defaultEndpoint;
+
             return GetApplicableEndpoints(
-                isReadRequest ? this.locationInfo.AvailableReadEndpointByLocation : this.locationInfo.AvailableWriteEndpointByLocation,
+                isReadRequest ? databaseAccountLocationsInfoSnapshot.AvailableReadEndpointByLocation : databaseAccountLocationsInfoSnapshot.AvailableWriteEndpointByLocation,
                 effectivePreferredLocations,
-                this.defaultEndpoint,
+                fallbackEndpoint,
                 request.RequestContext.ExcludeRegions);
         }
 
--- a/Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
@@ -58,7 +58,8 @@ namespace Microsoft.Azure.Cosmos.Routing
                 owner.ServiceEndpoint,
                 connectionPolicy.EnableEndpointDiscovery,
                 connectionPolicy.MaxConnectionLimit,
-                connectionPolicy.UseMultipleWriteLocations);
+                connectionPolicy.UseMultipleWriteLocations,
+                isPartitionLevelFailoverEnabled: () => connectionPolicy.EnablePartitionLevelFailover);
 
             this.owner = owner;
             this.defaultEndpoint = owner.ServiceEndpoint;
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/LocationCacheTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/LocationCacheTests.cs
@@ -1649,6 +1649,68 @@ namespace Microsoft.Azure.Cosmos.Client.Tests
         }
 
         [TestMethod]
+        [Description("Validates that when a PPAF partition-level override (LocationEndpointToRoute) is set, " +
+            "ResolveServiceEndpoint returns it directly, bypassing ExcludeRegions filtering entirely.")]
+        public void ValidateResolveServiceEndpoint_PPAFOverride_WinsOverExcludeRegions()
+        {
+            // Arrange: PPAF enabled, single preferred region "location1"
+            Collection<AccountRegion> writeLocations = new Collection<AccountRegion>()
+            {
+                new AccountRegion { Name = "location1", Endpoint = LocationCacheTests.Location1Endpoint.ToString() },
+            };
+
+            Collection<AccountRegion> readLocations = new Collection<AccountRegion>()
+            {
+                new AccountRegion { Name = "location1", Endpoint = LocationCacheTests.Location1Endpoint.ToString() },
+                new AccountRegion { Name = "location2", Endpoint = LocationCacheTests.Location2Endpoint.ToString() },
+            };
+
+            AccountProperties account = new AccountProperties
+            {
+                ReadLocationsInternal = readLocations,
+                WriteLocationsInternal = writeLocations,
+                EnableMultipleWriteLocations = false,
+            };
+
+            Uri defaultEndpoint = new Uri("https://myaccount.documents.azure.com");
+
+            LocationCache cache = new LocationCache(
+                preferredLocations: new List<string> { "location1" }.AsReadOnly(),
+                defaultEndpoint: defaultEndpoint,
+                enableEndpointDiscovery: true,
+                connectionLimit: 10,
+                useMultipleWriteLocations: false,
+                isPartitionLevelFailoverEnabled: () => true);
+
+            cache.OnDatabaseAccountRead(account);
+
+            // Simulate PPAF partition-level override: partition failed over to location2
+            Uri ppafOverrideEndpoint = LocationCacheTests.Location2Endpoint;
+
+            using (DocumentServiceRequest readRequest = DocumentServiceRequest.Create(
+                OperationType.Read,
+                ResourceType.Document,
+                AuthorizationTokenType.PrimaryMasterKey))
+            {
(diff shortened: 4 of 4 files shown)


Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -380,10 +383,18 @@ public ReadOnlyCollection<Uri> GetApplicableEndpoints(DocumentServiceRequest req
 
             ReadOnlyCollection<string> effectivePreferredLocations = databaseAccountLocationsInfoSnapshot.EffectivePreferredLocations;
 
+            // For reads when PPAF is enabled, use WriteEndpoints[0] as fallback (dynamic,
+            // tracks current write region) instead of this.defaultEndpoint (static, region-agnostic,
+            // never updated after init). This aligns with UpdateLocationCache which already uses
+            // WriteEndpoints[0] as the ReadEndpoints fallback, and matches Java/Python SDK behavior.
+            Uri fallbackEndpoint = (isReadRequest && this.isPartitionLevelFailoverEnabled?.Invoke() == true)
+                ? databaseAccountLocationsInfoSnapshot.WriteEndpoints[0]
```
kushagraThapar, line 391: Just curious, reading wirteEndpoint here from snapshot, but down below, we are reading AvailableWriteEndpointByLocation from this.locationInfo, can we use the same snapshot instead for both the reads, so that we don't get any stale / updated values in case the locationInfo gets updated in between (race condition)?

ananth7592: Good catch, Eliminated this TOCTOU and updated the PR>

kundadebdatta:
Shall we also update [the verbaige](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.requestoptions.excluderegions?view=azure-dotnet) in the `RequestOptions.ExcludeRegions` property to set the correct expectation here ?

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
kundadebdatta, line 391: Shall we always pick the `databaseAccountLocationsInfoSnapshot.WriteEndpoints[0]` or if we should condition it to check for the PPAF cache override as well? If the override is present, then always pick that endpoint as the hub endpoint.

ananth7592: Good question. The partition-level override (LocationEndpointToRoute) is set by GlobalPartitionEndpointManagerCore before 
   ResolveServiceEndpoint is called. When it's present, we short-circuit at L341 and never reach this fallback — ExcludeRegions 
  filtering
   doesn't even run.
   
   This fallback only fires when PPAF is enabled but the partition hasn't individually failed over (no override set). In that case, the
   partition still lives on the hub, so WriteEndpoints[0] is the correct endpoint — same as what GetHubUri() returns and what 
   ClientRetryPolicy uses for hub routing.
   
   So both paths land correctly:
   
   
    - Override present → override wins (L341, skips ExcludeRegions entirely)
    - No override → partition is on the hub → WriteEndpoints[0]
   
   No additional condition needed here. Let me know if this aligns

kundadebdatta (approved):

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -380,10 +383,18 @@ public ReadOnlyCollection<Uri> GetApplicableEndpoints(DocumentServiceRequest req
...
+                : this.defaultEndpoint;
```
NaluTripician, line 392: Do we still want to have `defaultEndpoint` here since that will point towards the global most of the time?

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
```diff
@@ -74,12 +74,12 @@ public class RequestOptions
...
-        public CosmosThresholdOptions CosmosThresholdOptions { get; set; }
-
+        public CosmosThresholdOptions CosmosThresholdOptions { get; set; }
+
         /// <summary>
         /// List of regions to be excluded routing the request to.
-        /// This can be used to route a request to a specific region by excluding all other regions.
-        /// If all regions are excluded, then the request will be routed to the primary/hub region.
+        /// This can be used to route a request to a specific region by excluding all other regions.
+        /// If all regions are excluded, the SDK will route the request on a best-effort basis to maintain availability.
```
NaluTripician, line 82: This sounds a bit vague IMO, maybe somethings like: 
`If every applicable region is excluded, the SDK falls back to the client's default (account) endpoint.` or `set primary write region`

kushagraThapar (approved):
LGTM, thanks @ananth7592

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
kundadebdatta: I think to keep the code in sync with the existing behavior and to avoid any breaking change, keeping the `defaultEndpoint` would still be applicable. This is the reason we wanted to ship it behind the PPAF flag to minimize any impact.

The ideal solution for "Routing the reads to follow the writes" will still be achievable by using the ReadConsistencyStrategy.

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
ananth7592: The intention is to actually discourage using this pattern for hub region routing and instead use Read consistency strategy APIs instead.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
ananth7592: This is done because we want to scope the radius of the fix to PPAF users only but eventually defaultEndpoint will always be WriteEndpoints[0]

NaluTripician (approved):

PR #4619, opened 2024-08-01
CFP AVAD: Fixes issue where customers are allowed to use WithStartTime and WithStartFromBeginning with CFP AVAD.

philipthomas-MSFT:
# Pull Request Template

Details contained at #4604 
## Description

Please include a summary of the change and which issue is fixed. Include samples if adding new API, and include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] This change requires a documentation update

## Closing issues

To automatically close an issue: closes #4604

PR #4370, opened 2024-03-25
ChangeFeedProcessor: Adds AllVersionsAndDeletes support to ChangeFeedProcessor

philipthomas-MSFT:
# Pull Request Template

## Description

Adding **AllVersionsAndDeletes** to `ChangeFeedProcessor`.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [] This change requires a documentation update

## Closing issues

To automatically close an issue: closes #IssueNumber

Issue #12, opened 2018-12-08
Why is index utilization 0.00 %?


erickbp:
Hi,

Can you please help me figure out why index utilization is 0.00 %?

I'm including this path in the indexing policy:
```
{
	"path": "/receivedDateTime/?",
	"indexes": [
		{
			"kind": "Range",
			"dataType": "Number",
			"precision": -1
		},
		{
			"kind": "Hash",
			"dataType": "String",
			"precision": 3
		}
	]
}
```
and this is an example of my documents:
```
{
    "id": "",
    "userId": "XYZ",
    ...
    "receivedDateTime": 635903262620000000,
    ...
}
```

When I run the following query I get zero index utilization.
Notes:
1. It does not happen with any other property/path that I included in the indexing policy
2. UserId is the partition key

```
_client.CreateDocumentQuery<Message>(UriFactory.CreateDocumentCollectionUri(_databaseId, _collectionId),
                    new FeedOptions
                    {
                        PopulateQueryMetrics = true,
                        MaxItemCount = maxItemCount
                    })
                .Where(item => item.UserId == Guid.Parse("XYZ"))
                .OrderBy(m => m.ReceivedDateTime)
                .AsDocumentQuery();
```
![image](https://user-images.githubusercontent.com/10678433/49682162-2eedea00-fa7d-11e8-9217-aa3d0e18a183.png)

On the other hand if I add `item.ReceivedDateTime >= 0` to the where clause I get 98.02 % of index utilization even when `item.ReceivedDateTime >= 0` is true for all the documents.
```
_client.CreateDocumentQuery<Message>(UriFactory.CreateDocumentCollectionUri(_databaseId, _collectionId),
                    new FeedOptions
                    {
                        PopulateQueryMetrics = true,
                        MaxItemCount = maxItemCount
                    })
                .Where(item => item.UserId == Guid.Parse("XYZ") && item.ReceivedDateTime >= 0)
                .OrderBy(m => m.ReceivedDateTime)
                .AsDocumentQuery();
```
![image](https://user-images.githubusercontent.com/10678433/49682190-fac6f900-fa7d-11e8-9368-3246d49d69f1.png)

Thank you

j82w:
This is related to the v2 SDK, and there is already a duplicate question in that repo. https://github.com/Azure/azure-cosmos-dotnet-v2/issues/646

PR #5825, opened 2026-05-01
ChangeFeed: Adds Full Fidelity Change Feed APIs to GA surface

### LINKED ISSUE #5814: ChangeFeed: Adds GA promotion for AllVersionsAndDeletes (Full Fidelity Change Feed) APIs
## Summary

The Full Fidelity Change Feed (AllVersionsAndDeletes) feature has been in Preview since SDK version 3.18.0-preview (PR #2122). After multiple iterations of hardening, contract refinements, and CFP integration, the feature is ready for GA promotion.

This issue tracks **removing the `#if PREVIEW` guards** from 7 types/members to make them unconditionally `public`, along with updating the corresponding API contract files.

## APIs to Promote

| # | Type / Member | File | Line | Description |
|---|---------------|------|------|-------------|
| 1 | `ChangeFeedMode.AllVersionsAndDeletes` | [`src/ChangeFeed/ChangeFeedMode.cs`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedMode.cs#L56) | L56 | Core entry point — the mode selector for FFCF |
| 2 | `ChangeFeedItem<T>` | [`src/Resource/FullFidelity/ChangeFeedItem.cs`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs#L52) | L52 | Wrapper for AVAD change items (current + previous + metadata) |
| 3 | `ChangeFeedMetadata` | [`src/Resource/FullFidelity/ChangeFeedMetadata.cs`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedMetadata.cs#L18) | L18 | Metadata: operation type, LSN, conflict resolution timestamp, previous LSN |
| 4 | `ChangeFeedOperationType` | [`src/Resource/FullFidelity/ChangeFeedOperationType.cs`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedOperationType.cs#L12) | L12 | Enum: Create, Replace, Delete |
| 5 | `ChangeFeedPolicy` | [`src/Resource/Settings/ChangeFeedPolicy.cs`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Resource/Settings/ChangeFeedPolicy.cs#L34) | L34 | Retention policy configuration for the AVAD log |
| 6 | `ChangeFeedPolicyDefinition` | [`src/Fluent/Settings/ChangeFeedPolicyDefinition.cs`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Fluent/Settings/ChangeFeedPolicyDefinition.cs#L12) | L12 | Fluent builder for setting ChangeFeedPolicy on a container |
| 7 | `ContainerProperties.ChangeFeedPolicy` | [`src/Resource/Settings/ContainerProperties.cs`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Resource/Settings/ContainerProperties.cs#L382) | L382 | Exposes ChangeFeedPolicy on ContainerProperties |

## Required Changes

For each file listed above:

1. Remove the `#if PREVIEW` / `public` / `#else` / `internal` / `#endif` block and replace with unconditional `public`.
2. Regenerate the GA API contract file (GenAPI) to include the newly public surface.
3. Verify that preview contract files are updated accordingly (remove the entries that moved to GA).

## Notes

- `Container.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>()` is already unconditionally `public` (no `#if PREVIEW` guard) — no changes needed there.
- The pull-model `GetChangeFeedIterator<T>()` accepting `ChangeFeedMode` is also already public, but passing `AllVersionsAndDeletes` is effectively blocked without Preview since the static property itself is `internal` without the flag.

## Related PRs / Issues

- #2122 — Initial FFCF preview support
- #3197 — Core SDK changes for FFCF
- #3596 — Renamed FullFidelity → AllVersionsAndDeletes
- #4370 — Added AVAD support to ChangeFeedProcessor
- #4380 — Renamed ChangeFeedItemChange → ChangeFeedItem
- #4618 — Fixed AVAD deserialization (STJ + Newtonsoft)
- #5470 — Added id/pk to ChangeFeedMetadata for deletes

ananth7592:
## Summary

Promotes Full Fidelity Change Feed (FFCF) APIs from preview to the GA surface area, per #5814.

### Changes
- Removed preview/beta attributes from FFCF APIs:
  - \\ChangeFeedMode.AllVersionsAndDeletes\\
  - \\ChangeFeedItem<T>\\
  - \\ChangeFeedMetadata\\
  - \\ChangeFeedOperationType\\
  - \\ChangeFeedPolicy.FullFidelityRetention\\
  - \\ChangeFeedPolicyDefinition\\
  - \\ContainerBuilder.WithChangeFeedPolicy()\\
  - \\ContainerProperties.ChangeFeedPolicy\\
- Updated contract file (\\DotNetSDKAPI.net6.json and 
DotNetPreviewSDKAPI.net6.json\\) to include promoted APIs
- Updated preview contract JSON to reflect removal from preview surface

### Testing
- Build passes with 0 warnings and 0 errors
- Existing FFCF tests continue to pass unchanged

### Local testing with ChangeFeed

Setup a container with AVAD support

Setup CFP to capture CRUDs. Following example shows "Inserts" and "Deletes"

Screenshots
Screenshot 2026-05-04 155922
Screenshot 2026-05-04 155847

#5826 

Resolves #5814


NaluTripician (commented):
# Deep Review

**Stated intent:** Promote FFCF / "AllVersionsAndDeletes" Change Feed APIs (`ChangeFeedMode.AllVersionsAndDeletes`, `ChangeFeedItem<T>`, `ChangeFeedMetadata`, `ChangeFeedOperationType`, `ChangeFeedPolicy.FullFidelityRetention`, `ChangeFeedPolicyDefinition`, `ContainerBuilder.WithChangeFeedPolicy()`, `ContainerProperties.ChangeFeedPolicy`, `Container.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>()`) from `#if PREVIEW` to GA.

## Overall Assessment

The mechanical "flip `#if PREVIEW` to plain `public`" work is correct: every type listed is unwrapped, the `Container.cs` reshuffle is sound, the duplicate `ContainerInternal` abstract is correctly removed, the contract JSONs are in sync (preview removals ↔ GA additions), and `ContainerCore.Items.cs` exposes the override publicly.

**However**, this PR is **more than a mechanical promotion**:

1. It silently bundles a **behavioral change** to `ChangeFeedProcessorCore.StartAsync` that carves AVAD out of the first-change-skip anchor introduced by #5617 — undocumented in the PR description, no test, and an unanswered design question carried forward from #4619.
2. It ships **doc bugs to GA** that were tolerable in preview (stale type name in XML example, self-referential `<see cref>`, a `ManualResetEvent` that's never `Set()`).
3. **No changelog entry** for the largest customer-facing GA addition in this release window.
4. **The downstream `Microsoft.Azure.Cosmos.Encryption` package will break** — its override is gated by `#if ENCRYPTIONPREVIEW` and throws `NotImplementedException` even when present.
5. `ChangeFeedItem<T>` ships with **public setters** on every property — inconsistent with sibling `ChangeFeedMetadata` (`internal set`), and once GA that mutability cannot be narrowed.

**Recommendation: hold for the two 🔴 items, plus the encryption-package break (#3), AVAD test (#4), and changelog (#7) before merge.**

---

## Findings (prioritized)

### 1. 🔴 Blocking · Hidden behavior change in `ChangeFeedProcessorCore.StartAsync`
**File:** `Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs:61-69`

```csharp
if (!this.changeFeedProcessorOptions.StartFromBeginning
    && this.changeFeedProcessorOptions.StartTime == null
    && this.changeFeedProcessorOptions.Mode != ChangeFeedMode.AllVersionsAndDeletes   // ← NEW
    && string.IsNullOrEmpty(this.changeFeedProcessorOptions.StartContinuation))
{
    this.changeFeedProcessorOptions.StartTime = DateTime.UtcNow.AddSeconds(-1);
}
```

This is a **behavioral change disguised as a GA-promotion PR**:

- It is **not mentioned in the PR description**, only obliquely referenced by the second commit message ("Fixes PREVIEW guard…", which actually fixes something else).
- It carves AVAD out of the anchor added by **PR #5617** ("Fixes first-change skip during initial startup by anchoring StartTime"). PR #5617's own validation showed **7/10 emulator tests fail without that fix**; this PR now exempts AVAD from the same fix.
- The exemption is plausibly correct — AVAD's iterator (`ChangeFeedPartitionKeyResultSetIteratorCore`) precedence is `existing continuation > StartTime > StartFromBeginning > Now`, so leaving `StartTime` null falls through to `ChangeFeedStartFrom.Now(feedRange)` which emits `IfNoneMatch: *` — but the PR provides **no justification, no analysis, and no test demonstrating that AVAD does *not* suffer the same first-change-skip race** that #5617 documented for incremental mode.
- The unanswered review question from PR #4619 ("@kirankumarkolli: *What's backend behavior in these cases?*") is still the right question to ask here.

**Why it matters:** if AVAD does share the timing-gap-during-async-lease-acquisition that #5617 fixed, this PR silently re-introduces that bug for AVAD users right as they become a GA audience. A real-world repro: customer calls `Build().StartAsync()`, then immediately writes an item — depending on how long lease acquisition takes, that write may or may not be observed by the AVAD processor on first read.

**Suggested action:**
- Document the change in the PR description and changelog.
- Add a symmetric unit test in `ChangeFeedProcessorCoreTests.cs` mirroring `StartAsync_SetsStartTime_WhenNoStartOptionsProvided` for AVAD: `StartAsync_DoesNotSetStartTime_WhenAllVersionsAndDeletes`, asserting `options.StartTime` remains `null`.
- Add an AVAD-mode emulator test mirroring `TestWithRunningProcessor_ImmediateWriteAfterStart` proving the first-change-skip race does NOT reproduce on AVAD.
- Explicitly answer in the PR description: *what is the resulting backend start position for an AVAD processor that reaches `StartAsync` with `StartTime == null`, `StartFromBeginning == false`, and `StartContinuation` empty?*

---

### 2. 🔴 Blocking · Stale type name `ChangeFeedItemChanges<T>` in GA-bound XML doc example
**File:** `Microsoft.Azure.Cosmos/src/Resource/FullFidelity/ChangeFeedItem.cs:27`

```xml
/// using (FeedIterator<ChangeFeedItem<ToDoActivity>> feedIterator = container.GetChangeFeedIterator<ChangeFeedItemChanges<ToDoActivity>>(
```

The XML doc example for the now-GA-public `ChangeFeedItem<T>` type uses **the old preview type name `ChangeFeedItemChanges<ToDoActivity>`** as the generic parameter to `GetChangeFeedIterator`. That type was renamed to `ChangeFeedItem<T>` (the rename was an explicit follow-up item from PR #4370 per @jcocchi: *"I'd like to see `ChangeFeedItemChange` simplified to `ChangeFeedItem`, but we agreed to handle that in a separate PR"*).

**Why it matters:** copy-paste-from-IntelliSense customers — the canonical first audience for a newly-GA API — will get **a CS0246 "type not found"** compile error. This is the exact GA-facing IntelliSense surface; a doc bug here is a support-ticket factory.

**Fix:** change line 27 to `container.GetChangeFeedIterator<ChangeFeedItem<ToDoActivity>>(`.

---

### 3. 🟡 Recommendation · `EncryptionContainer.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes` throws `NotImplementedException`
**File:** `Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainer.cs:772-775` (inside `#if ENCRYPTIONPREVIEW`)

```csharp
public override ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(string processorName, ChangeFeedHandler<ChangeFeedItem<T>> onChangesDelegate)
{
    throw new NotImplementedException();
}
```

This was a **known TODO at preview merge time** (PR #4370): @philipthomas-MSFT explicitly flagged *"After build and merged, custom encryption container can have the actual implementation?"* — never resolved.

Two consequences once `Container.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>` becomes a public unconditional abstract:

1. **Runtime:** Any GA customer using `Microsoft.Azure.Cosmos.Encryption` (Always Encrypted) who wraps a container and calls FFCF will get `NotImplementedException` from a *publicly documented GA method*. That is a worse failure mode than "this method does not exist" — there is no way for the caller to discover the limitation short of running the code.
2. **Build:** The override is gated by `#if ENCRYPTIONPREVIEW`. When `Microsoft.Azure.Cosmos.Encryption` rebuilds against a Cosmos NuGet that contains the new GA abstract method, the non-PREVIEW build of `EncryptionContainer` will be **missing the required override** — yielding either a compile error (CS0534) or making `EncryptionContainer` implicitly abstract. This is a downstream-package break that this PR causes but does not fix.

**Suggested action:**
- Either implement the override in `EncryptionContainer` (preferred — even if it just delegates to the inner container the way `Microsoft.Azure.Cosmos.Encryption.Custom.EncryptionContainer.cs:991` already does), or
- Keep the new GA abstract `#if PREVIEW`-only on `Container` for one more release cycle while the Encryption package catches up.

Note: the Custom package (`Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainer.cs:991-996`) *does* implement this correctly via delegation. Mirror that pattern.

---

### 4. 🟡 Recommendation · Missing test coverage for the AVAD carve-out
**File:** `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedProcessorCoreTests.cs`

PR #5617 added three unit tests + one emulator test specifically validating the `StartTime` auto-anchor behavior:
- `StartAsync_SetsStartTime_WhenNoStartOptionsProvided`
- `StartAsync_DoesNotOverrideExplicitStartTime`
- `StartAsync_DoesNotSetStartTime_WhenStartFromBeginning`
- `TestWithRunningProcessor_ImmediateWriteAfterStart` (emulator)

This PR adds a fourth condition to the same `if` but **adds no test for it**. The new branch is uncovered. The existing `StartAsync_SetsStartTime_WhenNoStartOptionsProvided` test passes only by virtue of defaulting `Mode = LatestVersion`, so it provides no regression protection if a future refactor changes the AVAD precedence.

**Suggested action:** add `StartAsync_DoesNotSetStartTime_WhenAllVersionsAndDeletes` (unit) and an AVAD-mode `TestWithRunningProcessor_ImmediateWriteAfterStart` (emulator) — see finding #1 for rationale.

---

### 5. 🟡 Recommendation · Reference-equality comparison on abstract `ChangeFeedMode`
**File:** `ChangeFeedProcessorCore.cs:63` (and the same pattern in `ChangeFeedProcessorBuilder.cs:136, 158`)

```csharp
this.changeFeedProcessorOptions.Mode != ChangeFeedMode.AllVersionsAndDeletes
```

`ChangeFeedMode` is an `abstract class` with no `==`/`!=` operator overloads, no `Equals` override, and `AllVersionsAndDeletes` is a static property returning the singleton `ChangeFeedModeFullFidelity.Instance`. This works **only** because every code path that produces an `AllVersionsAndDeletes` mode goes through the same singleton.

**Why it matters now:** as a GA feature, `ChangeFeedMode` becomes a frozen public surface that future maintainers will extend (e.g., a `ChangeFeedMode.AllVersionsAndDeletes(retention: …)` overload, a deserialized mode, or a test mock). Any of those would silently invert this condition. A future regression test of the form "does AVAD with no explicit start use NoStartTime?" will quietly stop guarding the invariant.

The pattern is now established in **three call sites** in code paths going GA — the cost of cleanup grows with each future copy.

**Suggested action:** introduce an internal `Mode.IsAllVersionsAndDeletes` helper (or compare by type — `Mode is ChangeFeedModeFullFidelity`) and use it consistently in all three sites.

---

### 6. 🟡 Recommendation · `ChangeFeedItem<T>` ships with public setters on a server-emitted DTO
**File:** `ChangeFeedItem.cs:60, 67, 74`

```csharp
public T Current { get; set; }
public ChangeFeedMetadata Metadata { get; set; }
public T Previous { get; set; }
```

Compare with sibling `ChangeFeedMetadata` which uses `internal set` on every property (`Lsn`, `OperationType`, `PreviousLsn`, `IsTimeToLiveExpired`, `Id`, `PartitionKey`) — the conservative GA pattern.

The class doc remarks (`ChangeFeedItem.cs:51`) say: *"is an optional helper class that uses Newtonsoft serialization libraries. Users are welcome to create their own custom helper class."* That suggests the setters are intentional so users can construct instances for tests or custom serialization. If so, **say so explicitly in the XML doc** so the inconsistency with `ChangeFeedMetadata`'s setter visibility is visible to API reviewers.

**Why it matters:** narrowing a `public set` to `internal set` or `init` after GA is a binary-breaking change. If the intent is "users can construct this for tests", document it; if not, change to `internal set` (or `init` with `[JsonInclude]`) before GA.

---

### 7. 🟡 Recommendation · Missing changelog entry for the GA promotion
**File:** `changelog.md`

The 3.59.0 section lists 15 PRs but **no entry for #5825** — despite this being the largest customer-facing surface addition in the release: a feature graduating to GA with ~9 new public types/members.

The PR doesn't touch `changelog.md` at all.

**Why it matters:** customers reading the changelog to decide whether to upgrade will see no mention that AVAD/FFCF is now GA. The new abstract method on `Container` is a contract change that anyone subclassing `Container` (rare but real — e.g., test doubles, custom encryption wrappers) needs to know about.

**Suggested action:** add an entry like:
> [5825](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5825) ChangeFeed: Promotes Full Fidelity Change Feed (AllVersionsAndDeletes) APIs to GA. Also exempts AVAD from the implicit StartTime back-off introduced in #5617.

Place it in the `#### Added` section (and arguably also a one-liner under `#### Fixed` for the implicit-`StartTime` carve-out, since that's a behavioral correction).

---

### 8. 🟡 Recommendation · Stale doc example on GA-bound `Container.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes`
**File:** `Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs:1804, 1816, 1867`

Two issues in the same XML doc that is graduating to GA IntelliSense:

1. **Line 1804** — Self-referential `<see cref>`:
   ```xml
   /// Initializes a <see cref="GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes"/> for change feed processing ...
   ```
   The cref points back at the method itself rather than the type it returns (`ChangeFeedProcessorBuilder`). This was tolerable in preview but will produce a self-loop in docs.microsoft.com / IntelliSense pop-ups once GA.

2. **Lines 1816 & 1867** — `ManualResetEvent` declared but never `Set()`:
   ```csharp
   /// ManualResetEvent allProcessedDocumentsEvent = new ManualResetEvent(false);
   /// ... // delegate never calls .Set()
   /// allProcessedDocumentsEvent.WaitOne(10 * 1000);
   ```
   A customer copy-pasting the canonical sample will see their app block for a full 10 seconds on every run. Worse, they'll likely blame the SDK before realizing the sample is wrong.

**Suggested action:**
- Change `<see cref="GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes"/>` → `<see cref="ChangeFeedProcessorBuilder"/>`.
- Either delete the `ManualResetEvent` entirely (replace `WaitOne` with `await Task.Delay(...)`) **or** add `allProcessedDocumentsEvent.Set();` inside the delegate after the `Console.WriteLine`.

---

### 9. 🟡 Recommendation · No front-door validation for AVAD + unsupported start options inside `StartAsync`
**File:** `ChangeFeedProcessorCore.cs:57-77`

PR #4619 added builder-level validation (`ChangeFeedProcessorBuilder.WithStartTime` and `WithStartFromBeginning`) that throws `InvalidOperationException` when these are called on an AVAD builder. But the *internal* code path in `StartAsync` now silently no-ops for AVAD instead of validating.

Two asymmetries that should be reconciled before GA:

- A user-facing `WithStartTime(t)` on AVAD throws a loud exception.
- The SDK-internal auto-set `StartTime = UtcNow - 1s` on AVAD now silently skips.
- `StartContinuation` is **not validated at all** for AVAD in either path. If a user serializes an incremental-mode continuation token and passes it to an AVAD builder, this PR doesn't catch it.

**Why it matters:** if "AVAD + StartTime" is invalid (which is what #4619 codified), then the SDK should never reach `StartAsync` with that combination. Conversely, if AVAD silently no-oping is the intended GA behavior, then the loud throw in `WithStartTime` should be downgraded. They can't both be right.

**Suggested action:** decide which contract holds. Either:
- Tighten `StartAsync` to throw if `Mode == AVAD && (StartFromBeginning || StartTime.HasValue || !string.IsNullOrEmpty(StartContinuation))` — covering paths that bypass the builder, OR
- Loosen the builder to match `StartAsync`'s silent-skip behavior, with an XML doc remark.

---

### 10. 🟢 Suggestion · Dead `#if PREVIEW` in `SettingsContractTests.ContainerSettingsDefaults`
**File:** `Microsoft.Azure.Cosmos.Tests/SettingsContractTests.cs:774-810`

After this PR, the `#if PREVIEW` and `#else` branches of `ContainerSettingsDefaults` contain **the same 16 property names**, just in different order — `ChangeFeedPolicy` appears mid-list in PREVIEW and appended at the end in GA. The conditional compilation is now functionally a no-op.

**Why it matters:** dead conditional compilation hides intent and the ordering discrepancy will confuse the next person adding a preview-only property — they'll have to figure out why two identical lists exist.

**Suggested fix:** collapse to a single unconditional `TypeAccessorGuard(...)` call with one canonical ordering, dropping the `#if PREVIEW/#else/#endif`.

---

### 11. 🟢 Suggestion · Refactor the 4-clause boolean with named local + comment
**File:** `ChangeFeedProcessorCore.cs:61-69`

The condition now answers a non-obvious question: *"do we need to apply the StartTime back-off compensation?"* The combined four-clause boolean buries the AVAD exclusion, and the existing comment explains what the back-off does but **not why FFCF is exempted**.

**Suggested refactor:**
```csharp
bool shouldAnchorStartTime =
    !this.changeFeedProcessorOptions.StartFromBeginning
    && this.changeFeedProcessorOptions.StartTime == null
    && string.IsNullOrEmpty(this.changeFeedProcessorOptions.StartContinuation)
    // AVAD uses LSN-based continuation (not RFC1123 IfModifiedSince),
    // so the seconds-precision back-off does not apply.
    && this.changeFeedProcessorOptions.Mode != ChangeFeedMode.AllVersionsAndDeletes;

if (shouldAnchorStartTime)
{
    this.changeFeedProcessorOptions.StartTime = DateTime.UtcNow.AddSeconds(-1);
}
```

---

### 12. 💬 Observation · PR description says `API_3.59.0.txt` was updated; it was not
**File:** `Microsoft.Azure.Cosmos/contracts/API_3.59.0.txt`

The PR description states: *"Updated GA contract file (API_3.59.0.txt) to include promoted APIs"*. The file at PR head is unchanged from the released 3.59.0 baseline.

The actual GA contract update is in `Microsoft.Azure.Cosmos.Tests/Contracts/DotNetSDKAPI.net6.json` (+317 lines, matching the −335 from `DotNetPreviewSDKAPI.net6.json`).

**Suggested action:** edit the PR description to say *"Updated GA contract JSON (`DotNetSDKAPI.net6.json`) to include promoted APIs"*.

---

## Summary

| # | Severity | Category | One-line |
|---|---|---|---|
| 1 | 🔴 Blocking | Hidden Behavior Change | Undocumented AVAD carve-out from #5617's first-change-skip fix, no test, unanswered design question |
| 2 | 🔴 Blocking | Doc Correctness | `ChangeFeedItem.cs:27` references the deleted type `ChangeFeedItemChanges<T>` — copy-paste will not compile |
| 3 | 🟡 Recommendation | Downstream Break | `Microsoft.Azure.Cosmos.Encryption.EncryptionContainer` throws `NotImplementedException` and is gated by `#if ENCRYPTIONPREVIEW` |
| 4 | 🟡 Recommendation | Test Coverage | No symmetric test for the AVAD `StartTime` carve-out |
| 5 | 🟡 Recommendation | Design Fragility | Reference-equality on abstract `ChangeFeedMode` singleton |
| 6 | 🟡 Recommendation | API Surface Lock | `ChangeFeedItem<T>` ships with public setters; inconsistent with `ChangeFeedMetadata` |
| 7 | 🟡 Recommendation | Release Hygiene | No changelog entry |
| 8 | 🟡 Recommendation | Doc Bugs Going GA | Self-referential `<see cref>` and unsignaled `ManualResetEvent` in `Container.cs` sample |
| 9 | 🟡 Recommendation | Validation Asymmetry | Builder throws on AVAD+StartTime, but `StartAsync` silently skips |
| 10 | 🟢 Suggestion | Dead Code | Identical `#if PREVIEW`/`#else` branches in `SettingsContractTests.cs` |
| 11 | 🟢 Suggestion | Readability | Extract 4-clause boolean to named local with explanation |
| 12 | 💬 Observation | Doc Accuracy | PR description claims `API_3.59.0.txt` was updated; it wasn't (the JSON contract was) |

The PR is ~80% mechanical and that work is correct. The remaining ~20% (the `StartAsync` change, the GA-bound doc examples, the encryption-package gap, the missing tests) is what GA promotion PRs typically miss and what reviewers should focus on.

*— Posted by Copilot on behalf of @NaluTripician*

ananth7592:
## Addressing Deep Review Findings

Thanks for the thorough review! Here's what has been addressed in the latest commit:

### �� Blockers

**#1 — Hidden behavioral change (AVAD StartTime carve-out)**
✅ Fixed. Refactored the 4-clause boolean into a named local `shouldAnchorStartTime` with a detailed comment explaining why AVAD is exempt (LSN-based continuation via `IfNoneMatch: *`, not RFC1123 `IfModifiedSince`). Also added a defensive guard in `StartAsync` that throws `InvalidOperationException` if AVAD mode is combined with `StartFromBeginning` or explicit `StartTime`.

**#2 — Stale type name in XML doc**
✅ Fixed. Changed `ChangeFeedItemChanges<ToDoActivity>` → `ChangeFeedItem<ToDoActivity>` in the XML doc example.

### 🟡 Recommendations

**#3 — EncryptionContainer downstream break**
✅ Fixed. Moved the `GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes` override out of `#if ENCRYPTIONPREVIEW` and implemented it via delegation to the inner container (mirroring the Custom encryption package pattern).

**#4 — No unit test for AVAD exemption**
✅ Fixed. Added `StartAsync_DoesNotSetStartTime_WhenAllVersionsAndDeletes` test that verifies `StartTime` remains null for AVAD mode.

**#5 — Reference equality on ChangeFeedMode**
⏭️ Kept as-is. `ChangeFeedMode.AllVersionsAndDeletes` returns a singleton (`ChangeFeedModeFullFidelity.Instance`), and all usage is internal. Reference equality is consistent with the existing codebase pattern.

**#6 — ChangeFeedItem<T> public setters**
✅ Fixed. Added an XML `<remarks>` block explaining that public setters are intentional — required for deserialization by both Newtonsoft.Json and System.Text.Json, and to allow users to construct instances for testing.

**#7 — Missing changelog entry**
✅ Fixed. Added entries under 3.60.0-preview.0: one under `Added` for the GA promotion and one under `Fixed` for the AVAD StartTime exemption.

**#8 — Container.cs XML doc bugs**
✅ Fixed. Corrected the self-referential `<see cref>` to point to `ChangeFeedProcessorBuilder` and added the missing `allProcessedDocumentsEvent.Set()` call in the sample code.

**#9 — Validation asymmetry**
✅ Fixed. Added a guard in `StartAsync` that throws `InvalidOperationException` if AVAD mode is used with `StartFromBeginning` or `StartTime`. Also added two corresponding unit tests (`StartAsync_ThrowsWhenAllVersionsAndDeletes_WithStartFromBeginning` and `StartAsync_ThrowsWhenAllVersionsAndDeletes_WithStartTime`).

### 🟢 Suggestions

**#10 — Dead `#if PREVIEW` in SettingsContractTests**
⏭️ Kept as-is. The `#if PREVIEW` / `#else` branches have a deliberate ordering difference for `ChangeFeedPolicy` in the property list.

**#11 — Extract 4-clause boolean**
✅ Fixed. Done as part of #1 — extracted to `shouldAnchorStartTime`.

### 💬 Observations

**#12 — PR description inaccuracy**
📝 Will update the PR description separately.

kushagraThapar (commented):
Thanks for the thorough turnaround on the deep-review feedback! I re-walked everything against the latest commit (`907d5a4`) and almost all of it looks great. A few small items worth tracking before this merges:

## ✅ Verified addressed

- **`StartAsync` refactor** (`ChangeFeedProcessorCore.cs:79-95`): the named local + multi-line comment reads cleanly, and the AVAD carve-out rationale (LSN-based vs RFC1123) is now self-documenting.
- **AVAD guards in `StartAsync`** (lines 64-77): matches the builder behavior for `StartFromBeginning` and `StartTime`.
- **Three new unit tests** (`StartAsync_DoesNotSetStartTime_WhenAllVersionsAndDeletes`, `..._ThrowsWhenAllVersionsAndDeletes_WithStartFromBeginning`, `..._ThrowsWhenAllVersionsAndDeletes_WithStartTime`) — coverage looks good.
- **`ChangeFeedItem.cs`**: type-name fix at line 27 + public-setter `<remarks>` block at lines 51-55.
- **`Container.cs` XML doc**: `<see cref>` corrected and `allProcessedDocumentsEvent.Set()` added inside the example delegate.
- **Changelog**: both entries present under 3.60.0-preview.0.
- **API contract**: all 8 promoted symbols present in `DotNetSDKAPI.net6.json`; the simple `#if PREVIEW` removals across the other 6 files are clean.

## ⚠️ Items worth tracking before/after merge

### 1. EncryptionContainer guard is a tactical workaround

Current state at `Microsoft.Azure.Cosmos.Encryption/src/EncryptionContainer.cs:773-779`:

```csharp
#if ENCRYPTIONPREVIEW || SDKPROJECTREF
    public override ChangeFeedProcessorBuilder GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(...)
    {
        return this.Container.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes(...);
    }
#endif
```

This works for the current CI because the existing GA Cosmos NuGet doesn't have the abstract method yet. **However**, once this PR ships and the new `Microsoft.Azure.Cosmos` NuGet is published with the abstract method on `Container`, the Encryption package's NuGet GA build will fail with **CS0534** because neither `ENCRYPTIONPREVIEW` nor `SDKPROJECTREF` is set in that configuration.

For comparison, `Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionContainer.cs:991` already has the override **unconditionally** — that's the eventual end state for the regular Encryption package too.

**Suggestion**: track this with a follow-up issue (and ideally a `// TODO:` comment on the guard) so the guard removal happens in lockstep with the next Encryption NuGet bump.

### 2. `StartContinuation` is still not validated for AVAD

Pre-existing gap from #4619, not introduced by this PR — but worth calling out since you've already added the sibling guards for `StartFromBeginning` and `StartTime` in this PR. A user can still hand an incremental-mode continuation token to an AVAD builder/processor and get undefined behavior. Either a small follow-up to add the third guard, or a tracking issue, would close the loop.

### 3. PR description references `API_3.59.0.txt`

Confirmed not updated in this PR — the actual contract delta lives in `DotNetSDKAPI.net6.json` (verified). Just a PR-description refresh, no code change needed.

### Optional nit (no action required)

The new throw-tests use `[ExpectedException(typeof(InvalidOperationException))]` which doesn't pin the message text. Asserting the message contains the option name + `AllVersionsAndDeletes` would be a tiny robustness win, but functional today.

---

Other than the EncryptionContainer follow-up, this looks merge-ready from my side. Nice work on the thorough turnaround! 🚀

kushagraThapar (changes requested):
Requesting changes as this will break encryption package.

ananth7592:
@kushagraThapar  
#1 EncryptionContainer guard is a tactical workaround

 Acknowledged — this is intentionally temporary. The #if ENCRYPTIONPREVIEW || SDKPROJECTREF guard exists solely because today's shipped
  GA NuGet (3.51.0) doesn't have the abstract method. The moment we publish the next GA SDK NuGet (3.59.0) with 
  GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes unconditional, the Encryption csproj's <PackageReference Version="[3.51.0,)" />
  will resolve to the new version, and the GA build will fail with CS0534 unless we remove the guard.

  The fix is mechanical and must happen in lockstep:

   1. Bump Microsoft.Azure.Cosmos.Encryption.csproj line 32: [3.51.0,) → [3.59.0,)
   2. Remove the #if ENCRYPTIONPREVIEW || SDKPROJECTREF / #endif around the override
   3. This matches what Encryption.Custom already does (unconditional override at line 991)

  Action: I've added a // TODO: comment on the guard referencing a tracking issue, and file the follow-up issue to remove it during the
  next Encryption release. This is standard practice — same pattern was used when GetChangeFeedProcessorBuilder (the non-AVAD variant)
  was promoted to GA.

 2. StartContinuation not validated for AVAD

  Good catch. This PR added guards for StartFromBeginning and StartTime on the AVAD builder (preventing users from mixing
  incremental-mode options with AllVersionsAndDeletes). However, WithStartContinuation (passing an incremental-mode continuation token)
  is not guarded.

  This is a pre-existing gap from #4619, not a regression introduced here. The risk: a user passes an incremental-mode continuation
  token to an AVAD processor → backend returns confusing errors or silently returns wrong data.

  Options:

   - Option A (small follow-up in this PR): Add validation in the AVAD builder that rejects non-AVAD continuation tokens. Low-risk, ~5 
  lines of code.
   - Option B (separate issue): Track as a follow-up since it's pre-existing and requires defining what "AVAD continuation token" looks 
  like for validation.

  I'd lean toward Option B — it's pre-existing behavior and the continuation token format validation needs careful design. Happy to file
  the issue.

Updated the PR Description to address #3 PR description references API_3.59.0.txt
Confirmed not updated in this PR — the actual contract del

ananth7592:
Created a separate issue for tracking #2 from above https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5863

kushagraThapar (dismissed):
LGTM, thanks @ananth7592 !

kundadebdatta (dismissed):
LGTM

kushagraThapar (dismissed):
Re-walked everything against HEAD `7610750`. All three concerns from my prior review look fully addressed:

1. ✅ **EncryptionContainer guard** — explicit `// TODO:` comment now lives at `EncryptionContainer.cs:773-776` naming the trigger (next Encryption NuGet bump), the action (remove the `#if` and unconditionalize the override), and pointing at `Encryption.Custom/src/EncryptionContainer.cs` as the reference end-state. Today's 4 follow-up commits iterated on the `#if` boundaries to make the guard work under all build configs (PREVIEW/GA × ProjectRef/NuGet × Encryption/Encryption.Custom).
2. ✅ **`StartContinuation` AVAD validation** — tracked in well-scoped follow-up issue #5863, cross-linked back to this PR and to #4619. Good as a separate change since it needs its own continuation-token-format design.
3. ✅ **PR description** — `API_3.59.0.txt` reference removed; description correctly points at `DotNetSDKAPI.net6.json` for the contract delta.

Thanks for the careful turnaround on the Encryption guard, @ananth7592 — happy to approve. The 5 Encryption pipelines being green will be the empirical confirmation that the guard works in all configurations; I'll trust those over my static reading. 🚀

kundadebdatta (approved):
LGTM!

NaluTripician (approved):

Issue #11, opened 2018-12-06
FormatException : Input string was not in a correct format.


RockyBalboa2018:
i try this very simple code sample, but getting a FormatException : Input string was not in a correct format.
Database and collection are created, but exception occurs on container.Items.CreateItemAsync
**SDK :**
Microsoft.Azure.Cosmos 3.0.0.1-preview (via nuget)
**Emulator :** 
Azure Cosmos DB Emulator 2.1.3.0

 ```csharp
string databaseId = "TodoDatabase";
string containerId = "TodoContainer";
 string partitionKey = "/partitionKey";

// Create new instance of CosmosClient
using (CosmosClient cosmosClient = new CosmosClient("https://localhost:5002", "C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw=="))
            {
                // Create new database
                CosmosDatabase database = await cosmosClient.Databases.CreateDatabaseIfNotExistsAsync(databaseId);

                // Create new container
                CosmosContainer container = await database.Containers.CreateContainerIfNotExistsAsync(containerId, partitionKey);

                // Add item to container
                var todoItem = new TodoItem()
                {
                    Id = Guid.NewGuid().ToString(),
                    PartitionKey = Guid.NewGuid().ToString(),
                    Task = "Get started with Azure Cosmos DB!"
                };
                var todoItemResponse = await container.Items.CreateItemAsync<TodoItem>(todoItem.PartitionKey, todoItem);
            }

j82w:
I was not able to repro the issue using the below TodoItem in a .net core 2.1 console app. All items are required to have a "id" property and the matching partitionkey property. The properties are case sensitive.

```
public class TodoItem
    {
        [JsonProperty(PropertyName = "id")]
        public string Id { get; set; }
        [JsonProperty(PropertyName = "partitionKey")]
        public string PartitionKey { get; set; }
        [JsonProperty(PropertyName = "task")]
        public string Task { get; set; }
    }
```

1. What project and version are you using?
2. What version of Newtonsoft.Json?
3. Can you please add the full exception message? It has additional information that would make it easier to pinpoint the issue.

PR #5826, opened 2026-05-04
Retry: Adds client-side 449 (RetryWith) retry with jittered exponential backoff

yash2710:
## Description

The proxy will no longer retry 449 for external SDK requests to avoid noisy-neighbor amplification on shared proxy resources. This adds client-side 449 retry handling in ClientRetryPolicy with jittered exponential backoff using the existing RetryWithConfiguration parameters:
- Initial backoff: 10ms (configurable)
- Max backoff: 1000ms (configurable)
- Jitter: random salt up to 5ms (configurable)
- Total wait time budget: 30s (configurable)
- Backoff multiplier: 2x

The retry is triggered for both exception-based (DocumentClientException) and exceptionless (ResponseMessage) 449 responses in Gateway mode.

## Type of change


NaluTripician:
@sdkReviewAgent-2

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -24,21 +25,39 @@ internal sealed class ClientRetryPolicy : IDocumentClientRetryPolicy
...
         private const int MaxServiceUnavailableRetryCount = 1;
 
+        // Default backoff parameters for 449 (RetryWith) - matches direct-mode GoneAndRetryWithRequestRetryPolicy defaults
+        private const int DefaultRetryWithInitialBackoffMilliseconds = 10;
+        private const int DefaultRetryWithMaxBackoffMilliseconds = 1000;
+        private const int DefaultRetryWithRandomSaltMilliseconds = 5;
+        private const int DefaultRetryWithTotalWaitTimeMilliseconds = 30000;
+        private const int RetryWithBackoffMultiplier = 2;
+
+        private static readonly Random RetryWithRandom = new Random();
```
xinlian12, line 46: 🔴 **Blocking — Correctness: Thread-unsafe `static Random` on `netstandard2.0`**

`System.Random` is [not thread-safe](https://learn.microsoft.com/en-us/dotnet/api/system.random#thread-safety) on .NET Framework and .NET Standard 2.0. Since `ClientRetryPolicy` is instantiated **per-request** (via `RetryPolicy.GetRequestPolicy()`), multiple concurrent requests all share this single `static Random` instance and call `RetryWithRandom.Next()` from different thread pool threads without synchronization.

**Concrete failure scenario:** Under high concurrent load with many 449 responses (the exact operational scenario this PR targets), two threads call `Next()` simultaneously. The internal `Random` state corrupts, causing it to permanently return 0 for all subsequent calls. All jitter disappears, and every concurrent client retries with identical timing — creating precisely the thundering herd the jitter is designed to prevent.

This is silent — no exception, no trace, just degraded behavior under load.

The project targets `netstandard2.0` (confirmed in `Microsoft.Azure.Cosmos.csproj:19`), so `Random.Shared` (.NET 6+) is unavailable. The Java SDK correctly uses `ThreadLocalRandom.current()` for the equivalent code.

**Suggested fix** — simplest option with minimal contention:

```csharp
[ThreadStatic]
private static Random t_retryWithRandom;
private static Random RetryWithRandom => t_retryWithRandom ??= new Random();

```
Or use a lock (contention is minimal — one `Next()` per retry):

```csharp
private static readonly object s_randomLock = new object();
private static readonly Random RetryWithRandom = new Random();

// In ShouldRetryOnRetryWithStatus:
int salt;
lock (s_randomLock) { salt = RetryWithRandom.Next(1, saltMax); }

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -664,8 +664,172 @@ private static GlobalPartitionEndpointManagerCore.PartitionKeyRangeFailoverInfo
...
+            ShouldRetryResult second = await retryPolicy.ShouldRetryAsync(retryWithException, CancellationToken.None);
+            ShouldRetryResult third = await retryPolicy.ShouldRetryAsync(retryWithException, CancellationToken.None);
+
+            Assert.IsTrue(first.ShouldRetry);
+            Assert.IsTrue(second.ShouldRetry);
+            Assert.IsTrue(third.ShouldRetry);
+
+            // Backoff should generally increase (accounting for jitter)
+            // First: ~10ms + salt, Second: ~20ms + salt, Third: ~40ms + salt
+            Assert.IsTrue(
```
xinlian12: 🟡 **Recommendation — Testing: Exponential backoff assertion is nearly vacuous**

The `||` (OR) condition makes this assertion trivially satisfiable. With base doubling (10→20→40ms) and salt range [1,4]ms:
- `second` is always in [21, 24]ms, `first` is always in [11, 14]ms → `second >= first` is **always true**
- So the OR is never exercised, and the test tells you nothing about the third step

Worse: if a regression broke exponential growth (making all backoffs flat at 10ms + jitter = [11,14]ms), the test would **still pass ~75% of the time** because `P(second ≥ first) ≈ 0.5` and `P(third ≥ second) ≈ 0.5`, giving `P(either) = 0.75`.

**Suggested fix** — assert the exponential structure directly:

```csharp
// With base 10ms * 2^n and salt [1,4]ms:
Assert.IsTrue(first.BackoffTime.TotalMilliseconds >= 11 && first.BackoffTime.TotalMilliseconds <= 14,
    $"First backoff should be 10+[1,4]ms, was {first.BackoffTime.TotalMilliseconds}ms");
Assert.IsTrue(second.BackoffTime.TotalMilliseconds >= 21 && second.BackoffTime.TotalMilliseconds <= 24,
    $"Second backoff should be 20+[1,4]ms, was {second.BackoffTime.TotalMilliseconds}ms");
Assert.IsTrue(third.BackoffTime.TotalMilliseconds >= 41 && third.BackoffTime.TotalMilliseconds <= 44,
    $"Third backoff should be 40+[1,4]ms, was {third.BackoffTime.TotalMilliseconds}ms");

```

Or at minimum, use `&&` (AND) instead of `||`:

```csharp
Assert.IsTrue(
    second.BackoffTime > first.BackoffTime && third.BackoffTime > second.BackoffTime, ...);

```
This always holds with correct exponential growth (21 > 14 and 41 > 24) and would fail reliably with flat backoff.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -664,8 +664,172 @@ private static GlobalPartitionEndpointManagerCore.PartitionKeyRangeFailoverInfo
...
+            DocumentClientException retryWithException = new DocumentClientException(
+                message: "RetryWith",
+                innerException: new Exception(),
+                statusCode: (HttpStatusCode)449,
+                substatusCode: SubStatusCodes.Unknown,
+                requestUri: request.RequestContext.LocationEndpointToRoute,
+                responseHeaders: headers.Object);
+
+            // First retry should succeed (stopwatch just started)
+            ShouldRetryResult firstResult = await retryPolicy.ShouldRetryAsync(retryWithException, CancellationToken.None);
```
xinlian12, line 1334: 🟡 **Recommendation — Testing: Missing assertion on first retry result**

`firstResult` is captured but never asserted. If a bug caused `ShouldRetryOnRetryWithStatus()` to always return `NoRetry()` (rejecting ALL retries including the first), this test would still pass — it only checks that `secondResult.ShouldRetry == false`.

The test should verify that the 449 handler accepts at least one retry before the budget runs out:

```csharp
Assert.IsTrue(firstResult.ShouldRetry, "First retry should succeed with budget remaining.");
Assert.IsTrue(firstResult.BackoffTime.TotalMilliseconds <= 1,
    "First backoff should be capped to remaining 1ms budget.");

```

This distinguishes "retry works then stops after budget" from "retry never works at all."

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -664,8 +664,172 @@ private static GlobalPartitionEndpointManagerCore.PartitionKeyRangeFailoverInfo
...
+                statusCode: (HttpStatusCode)449,
+                substatusCode: SubStatusCodes.Unknown,
+                requestUri: request.RequestContext.LocationEndpointToRoute,
+                responseHeaders: headers.Object);
+
+            ShouldRetryResult result = await retryPolicy.ShouldRetryAsync(retryWithException, CancellationToken.None);
+
+            Assert.IsTrue(result.ShouldRetry, "First 449 should trigger a retry.");
+            Assert.IsTrue(result.BackoffTime.TotalMilliseconds > 0, "Backoff should be greater than zero.");
+            Assert.IsTrue(result.BackoffTime.TotalMilliseconds <= 1000, "First backoff should not exceed max backoff of 1000ms.");
```
xinlian12: 🟡 **Recommendation — Testing: First backoff upper bound too loose (1000ms)**

With initial backoff = 10ms and max jitter = 4ms (`Random.Next(1, 5)`), the first backoff should be in [11, 14]ms. Asserting `<= 1000` (the max cap) doesn't catch regressions where the initial backoff accidentally uses a much larger value — e.g., a bug setting initial backoff to 999ms would pass this test.

**Suggested fix:**

```csharp
Assert.IsTrue(result.BackoffTime.TotalMilliseconds >= 11, "First backoff should be at least 10ms + 1ms salt.");
Assert.IsTrue(result.BackoffTime.TotalMilliseconds <= 14, "First backoff should be at most 10ms + 4ms salt.");

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -664,8 +664,172 @@ private static GlobalPartitionEndpointManagerCore.PartitionKeyRangeFailoverInfo
...
+                message: "RetryWith",
+                innerException: new Exception(),
+                statusCode: (HttpStatusCode)449,
+                substatusCode: SubStatusCodes.Unknown,
+                requestUri: request.RequestContext.LocationEndpointToRoute,
+                responseHeaders: headers.Object);
+
+            ShouldRetryResult first = await retryPolicy.ShouldRetryAsync(retryWithException, CancellationToken.None);
+            ShouldRetryResult second = await retryPolicy.ShouldRetryAsync(retryWithException, CancellationToken.None);
+            ShouldRetryResult third = await retryPolicy.ShouldRetryAsync(retryWithException, CancellationToken.None);
```
xinlian12, line 1383: 🟡 **Recommendation — Testing: No test for max backoff cap**

The exponential backoff test exercises only 3 retries (reaching ~40ms). There's no test verifying that backoff actually caps at `retryWithMaxBackoffMilliseconds` (default 1000ms) after sufficient doublings (10→20→40→80→160→320→640→1000→1000…). This is a key invariant of the algorithm — if the `Math.Min` cap were removed or broken, no test would catch it.

**Suggested:** Add a test with a small `MaximumRetryForRetryWithMilliseconds` (e.g., 25ms) and `RandomSaltForRetryWithMilliseconds = 0`, then retry enough times (4+) to exceed the cap. Verify backoff plateaus:

```csharp
// After 3 retries (10→20→25 capped), the 4th should also be capped at 25ms
Assert.AreEqual(25, fourthResult.BackoffTime.TotalMilliseconds);

```

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -542,6 +579,67 @@ private ShouldRetryResult ShouldRetryOnUnavailableEndpointStatusCodes()
...
+            }
+
+            // Initialize or advance exponential backoff
+            if (this.retryWithCurrentBackoffMilliseconds == null)
+            {
+                this.retryWithCurrentBackoffMilliseconds = this.retryWithInitialBackoffMilliseconds;
+            }
+            else
+            {
+                this.retryWithCurrentBackoffMilliseconds = Math.Min(
```
xinlian12: 🟢 **Suggestion — Robustness: Guard against integer overflow in backoff multiplication**

`retryWithCurrentBackoffMilliseconds.Value * RetryWithBackoffMultiplier` is computed in unchecked `int` arithmetic before the `Math.Min` cap is applied. With default values (max=1000), this is safe — capping occurs at 640→1000 well before overflow. However, if a future config change allows larger `MaximumRetryForRetryWithMilliseconds` values, after ~27 doublings from 10 the multiplication overflows `int`, wraps negative, and `Math.Min(negative, max)` returns a negative value. `Task.Delay(negative TimeSpan)` throws `ArgumentOutOfRangeException`, crashing the retry loop.

A defensive fix would be to clamp before multiplying:

```csharp
this.retryWithCurrentBackoffMilliseconds = Math.Min(
    Math.Min(this.retryWithCurrentBackoffMilliseconds.Value, this.retryWithMaxBackoffMilliseconds / ClientRetryPolicy.RetryWithBackoffMultiplier)
        * ClientRetryPolicy.RetryWithBackoffMultiplier,
    this.retryWithMaxBackoffMilliseconds);

```

Low urgency since current validation prevents the extreme values needed, but worth considering for future-proofing.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -542,6 +579,67 @@ private ShouldRetryResult ShouldRetryOnUnavailableEndpointStatusCodes()
...
+                    this.retryWithCurrentBackoffMilliseconds.Value * ClientRetryPolicy.RetryWithBackoffMultiplier,
+                    this.retryWithMaxBackoffMilliseconds);
+            }
+
+            int backoffMilliseconds = this.retryWithCurrentBackoffMilliseconds.Value;
+
+            // Add jitter
+            if (this.retryWithRandomSaltMilliseconds.HasValue && this.retryWithRandomSaltMilliseconds.Value > 0)
```
xinlian12: 💬 **Observation — Design: Additive jitter provides diminishing relative spread at higher backoff levels**

`Random.Next(1, 5)` adds [1, 4]ms of fixed jitter. At the initial 10ms backoff, this is meaningful (~10-40% variance). But at higher levels — 640ms or the 1000ms cap — it's <0.5% spread, which provides negligible thundering herd avoidance across many concurrent clients.

This matches the existing direct-mode `GoneAndRetryWithRequestRetryPolicy` defaults, so consistency is a valid reason to keep it. But if the team wants stronger thundering herd protection for the gateway path (which may see higher concurrency than direct mode), consider proportional "equal jitter":

```csharp
// Equal jitter: half deterministic + half random
int halfBackoff = backoffMilliseconds / 2;
backoffMilliseconds = halfBackoff + RetryWithRandom.Next(0, halfBackoff + 1);

```

Also note: when `retryWithRandomSaltMilliseconds == 1`, `Random.Next(1, 1)` always returns 1 (deterministic), so salt=1 produces no randomness despite passing the `> 0` guard.

No action needed if consistency with direct-mode is the priority.

xinlian12:
✅ **Review complete** (58:08)

Posted **7** inline comment(s).

yash2710:
@sdkReviewAgent-2

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -542,6 +580,70 @@ private ShouldRetryResult ShouldRetryOnUnavailableEndpointStatusCodes()
...
+                    "ClientRetryPolicy: RetryWith (449) total wait time exceeded. Elapsed: {0}ms, TotalWaitTime: {1}ms. Location: {2}; ResourceAddress: {3}",
+                    elapsedMilliseconds,
+                    this.retryWithTotalWaitTimeMilliseconds,
+                    this.documentServiceRequest?.RequestContext?.LocationEndpointToRoute?.ToString() ?? string.Empty,
+                    this.documentServiceRequest?.ResourceAddress ?? string.Empty);
+                return ShouldRetryResult.NoRetry();
+            }
+
+            // Initialize or advance exponential backoff
+            if (this.retryWithCurrentBackoffMilliseconds == null)
```
xinlian12, line 683: 🟡 **Recommendation — Behavioral divergence: First retry applies delay, unlike direct-mode's penalty-free first retry**

The direct-mode `GoneAndRetryWithRequestRetryPolicy` explicitly grants a penalty-free first retry with this guard:

```csharp
// Don't penalise first retry with delay.
if (this.attemptCount++ > 1)
{
    // backoff computation only runs from the second retry onward
}

```

I verified this on the `msdata/direct` branch — `attemptCount` is initialized to `1`, so the first call evaluates `1 > 1 = false` and skips the entire backoff block. The `SdkDesign.md` (line 158) corroborates: *"There is no delay for the first retry."*

The PR code applies the full initial backoff (10ms + jitter → [11, 14]ms) from the very first retry. This means:

| Retry # | Direct mode | Gateway mode (this PR) |
|---------|------------|----------------------|

| 1st | 0ms | ~12ms |

| 2nd | ~12ms | ~22ms |

| 3rd | ~22ms | ~42ms |

This may be an intentional design choice for gateway mode — a small initial delay gives the contention time to resolve. But the code comment on line 28 says *"matches direct-mode GoneAndRetryWithRequestRetryPolicy defaults"* which could be misread as claiming behavioral parity.

**Suggestion:** Either align with direct-mode by adding a penalty-free first retry:

```csharp
if (this.retryWithCurrentBackoffMilliseconds == null)
{
    this.retryWithCurrentBackoffMilliseconds = this.retryWithInitialBackoffMilliseconds;
    return ShouldRetryResult.RetryAfter(TimeSpan.Zero);
}

```
Or update the code comment to clarify: *"Default values match direct-mode; first retry intentionally includes initial backoff for gateway path."*

xinlian12:
✅ **Review complete** (26:09)

yash2710:
@sdkReviewAgent-2

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -542,6 +581,77 @@ private ShouldRetryResult ShouldRetryOnUnavailableEndpointStatusCodes()
...
+                    "ClientRetryPolicy: Retrying on 449 (RetryWith) — first attempt, no delay. Location: {0}; ResourceAddress: {1}",
+                    this.documentServiceRequest?.RequestContext?.LocationEndpointToRoute?.ToString() ?? string.Empty,
+                    this.documentServiceRequest?.ResourceAddress ?? string.Empty);
+
+                return ShouldRetryResult.RetryAfter(TimeSpan.Zero);
+            }
+            else
+            {
+                this.retryWithCurrentBackoffMilliseconds = (int)Math.Min(
+                    (long)this.retryWithCurrentBackoffMilliseconds.Value * ClientRetryPolicy.RetryWithBackoffMultiplier,
```
xinlian12: 🔴 **Blocking — Correctness: Backoff doubles before it's used, shifting the entire exponential sequence by one step**

The `else` branch updates `retryWithCurrentBackoffMilliseconds` (line 625-627) **before** it's read on line 630. This means the second call uses `initial × 2` instead of `initial`:

```
Call 1: null → set to 10, return 0ms          ✓
Call 2: 10 → else: min(10×2, max)=20 → read 20 + jitter  → [21,24]ms
Call 3: 20 → else: min(20×2, max)=40 → read 40 + jitter  → [41,44]ms

```

But the tests assert `[11,14]ms` for the second call and `[21,24]ms` for the third. **Three tests will fail at runtime:**

- `BackoffIncreasesExponentially`: expects `[11,14]` for second, gets `[21,24]`
- `BackoffCapsAtMaximum` (salt=0, max=25): `Assert.AreEqual(10, second)` gets 20
- `ShouldRetryWithBackoff`: same second-call mismatch

**Fix** — read before doubling (for the *next* call):

```csharp
int backoffMilliseconds = this.retryWithCurrentBackoffMilliseconds.Value;
this.retryWithCurrentBackoffMilliseconds = (int)Math.Min(
    (long)this.retryWithCurrentBackoffMilliseconds.Value * ClientRetryPolicy.RetryWithBackoffMultiplier,
    this.retryWithMaxBackoffMilliseconds);

```

This produces the intended sequence: `[0, 10+j, 20+j, 40+j, ...]`.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -356,6 +387,14 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
                 return this.ShouldRetryOnUnavailableEndpointStatusCodes();
             }
 
+            // Received 449 (RetryWith) - transient concurrency error from the store.
+            // The proxy no longer retries 449 for external SDK requests to avoid noisy-neighbor amplification.
+            // The SDK handles retries with jittered exponential backoff.
+            if (statusCode.HasValue && (int)statusCode.Value == 449)
```
xinlian12: 🟡 **Recommendation — Connectivity Mode: No guard for Gateway-only — risk of double-retry in Direct mode**

The PR description says this retry targets **"Gateway mode"**, but the status code check here is unconditional — it fires for *any* 449 regardless of connectivity mode.

In Direct mode, `GoneAndRetryWithRequestRetryPolicy` (in the closed-source `Microsoft.Azure.Cosmos.Direct` package) already handles 449 at the RNTBD transport layer with its own 30s jittered exponential backoff. When that inner policy exhausts its budget, `TransportHandler.SendAsync` (line 39-42) catches the resulting `DocumentClientException(449)` and converts it to `ResponseMessage(449)`. That response then flows back to `AbstractRetryHandler`'s retry loop, which calls `ClientRetryPolicy.ShouldRetryAsync(ResponseMessage)` — hitting this new 449 branch and starting **another** 30s retry cycle.

**Consequence:** In Direct mode with persistent 449s, total retry time could be ~60s (inner 30s + outer 30s) instead of the configured 30s budget, with multiplicative amplification of backend calls — the exact noisy-neighbor problem this PR aims to prevent in Gateway mode.

The review rules for this repo require: *"DO check that region-routing changes consider all connectivity modes (Gateway and Direct)."*

**Suggested fix:** Add a connectivity mode guard, e.g.:

```csharp
// Only retry 449 in Gateway mode. In Direct mode, GoneAndRetryWithRequestRetryPolicy
// already handles 449 at the transport layer.
if (statusCode.HasValue && (int)statusCode.Value == 449
    && this.documentServiceRequest?.RequestContext?.UseGatewayMode != false)
{
    return this.ShouldRetryOnRetryWithStatus();
}

```

Or if the double-retry is intentional for Direct mode, document that in the PR description and code comments.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -542,6 +581,77 @@ private ShouldRetryResult ShouldRetryOnUnavailableEndpointStatusCodes()
...
+            if (this.retryWithRandomSaltMilliseconds.HasValue && this.retryWithRandomSaltMilliseconds.Value > 1)
+            {
+                lock (ClientRetryPolicy.RetryWithRandomLock)
+                {
+                    backoffMilliseconds += ClientRetryPolicy.RetryWithRandom.Next(1, this.retryWithRandomSaltMilliseconds.Value);
+                }
+            }
+
+            // Cap to remaining time and max backoff
+            backoffMilliseconds = Math.Min(backoffMilliseconds, Math.Min(remainingMilliseconds, this.retryWithMaxBackoffMilliseconds));
```
xinlian12, line 712: 🟢 **Suggestion — Design: Jitter is completely eliminated at the max backoff plateau**

When `retryWithCurrentBackoffMilliseconds` reaches `retryWithMaxBackoffMilliseconds` (1000ms by default), jitter adds [1,4]ms → `backoffMilliseconds` becomes [1001,1004]. Then this `Math.Min` cap clamps it back to `min(1001-1004, min(remaining, 1000))` = **1000** — since `maxBackoff` is always ≤ the jittered value.

This means **all concurrent clients retry with identical 1000ms delays** at the plateau, defeating the purpose of jitter for thundering herd prevention. The earlier observation comment noted diminishing jitter, but this is the specific mechanism: the cap line makes jitter exactly zero once backoff reaches the max.

A simple fix (if desired): cap `retryWithCurrentBackoffMilliseconds` to `maxBackoff` but apply jitter *after* the cap so the final value can exceed `maxBackoff` by up to `salt`:

```csharp
backoffMilliseconds = Math.Min(this.retryWithCurrentBackoffMilliseconds.Value, this.retryWithMaxBackoffMilliseconds);
// Add jitter AFTER the max cap
backoffMilliseconds += jitterValue;
// Only cap to remaining time (not max backoff)
backoffMilliseconds = Math.Min(backoffMilliseconds, remainingMilliseconds);

```

No urgency if this matches the direct-mode behavior intentionally.

xinlian12:
✅ **Review complete** (52:49)

yash2710:
@sdkReviewAgent-2

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -1,752 +1,867 @@
...
+                return this.ShouldRetryOnUnavailableEndpointStatusCodes();
+            }
+
+            // Received 449 (RetryWith) - transient concurrency error from the store.
+            // The proxy no longer retries 449 for external SDK requests to avoid noisy-neighbor amplification.
+            // The SDK handles retries with jittered exponential backoff.
+            // Only retry in Gateway mode or ThinClient mode. In Direct mode, GoneAndRetryWithRequestRetryPolicy
+            // already handles 449 at the transport layer — retrying here would cause double-retry amplification.
+            if (statusCode.HasValue && (int)statusCode.Value == 449
+                && (this.isGatewayClientMode || this.isThinClientEnabled))
```
xinlian12: 🟡 **Recommendation — Correctness: DTX 449/5352 requests in Gateway mode bypass DTX-specific retry handler**

The new 449 check at line 420 fires **before** the DTX check at line 426. For a DTX request running in Gateway mode that receives 449/5352 (`DtcCoordinatorRaceConflict`), the generic handler intercepts it — the DTX handler is never reached.

```csharp
// Line 420: catches ANY 449 in Gateway mode, regardless of substatus or isDtxRequest
if (statusCode.HasValue && (int)statusCode.Value == 449
    && (this.isGatewayClientMode || this.isThinClientEnabled))
{
    return this.ShouldRetryOnRetryWithStatus();  // ← returns here
}

// Line 426: never reached for 449 in Gateway/ThinClient mode
if (this.isDtxRequest)
{
    return this.ShouldRetryDtxRequest(statusCode, subStatusCode, retryAfter, hasResponseBody);
}

```

**Why it matters:** The DTX handler has critical logic the generic handler lacks:
- **Body-presence check:** DTX defers to the outer `DistributedTransactionCommitter` loop when `hasResponseBody && isCoordinatorRetriable` to avoid inner×outer retry amplification. The generic handler retries regardless of response body.
- **Budget model:** DTX uses count-based cap (`MaxDtxRetryCount=10`) with server-directed `Retry-After` delay. The generic handler uses a 30s Stopwatch budget with exponential backoff starting at 0ms.

If DTX currently only runs in Direct mode, this is not a live regression — but it's a latent trap. Any future enablement of DTX in Gateway mode would silently hit the wrong retry path.

**Suggested fix** — add a `!this.isDtxRequest` exclusion:

```csharp
if (statusCode.HasValue && (int)statusCode.Value == 449
    && !this.isDtxRequest
    && (this.isGatewayClientMode || this.isThinClientEnabled))

```

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -1038,8 +1038,302 @@ private static GlobalPartitionEndpointManagerCore.PartitionKeyRangeFailoverInfo
...
-            return null;
-        }
+            return null;
+        }
+
+        /// <summary>
+        /// Tests that 449 (RetryWith) triggers a retry with backoff via the exception path.
+        /// </summary>
+        [TestMethod]
+        public async Task ShouldRetryAsync_On449RetryWith_ShouldRetryWithBackoff()
```
xinlian12, line 1224: 🟡 **Recommendation — Testing: Missing negative test for Direct-mode 449 non-retry**

All 6 new tests use `isGatewayClientMode: true`. There is no test verifying that 449 is **not** retried when `isGatewayClientMode: false` and `isThinClientEnabled: false` (Direct mode).

**Why it matters:** The guard `(this.isGatewayClientMode || this.isThinClientEnabled)` is the critical safety mechanism that prevents double-retry amplification with the transport-layer `GoneAndRetryWithRequestRetryPolicy` in Direct mode. This invariant has zero test coverage — a future refactoring that accidentally removes or inverts the guard would pass all existing tests while silently enabling retry storms in Direct mode.

**Suggested fix** — add a parameterized test covering all three modes:

```csharp
[DataTestMethod]
[DataRow(false, false, false, DisplayName = "Direct mode — should NOT retry 449")]
[DataRow(true, false, true, DisplayName = "Gateway mode — should retry 449")]
[DataRow(false, true, true, DisplayName = "ThinClient mode — should retry 449")]
public async Task ShouldRetryAsync_On449_ModeGating(
    bool isGatewayClientMode, bool isThinClientEnabled, bool expectRetry)
{
    // ... create policy with isGatewayClientMode/isThinClientEnabled ...
    // Assert result.ShouldRetry == expectRetry
}

```

This also covers the `isThinClientEnabled: true` branch, which is currently untested.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -1,752 +1,867 @@
...
+            if (this.retryWithRandomSaltMilliseconds.HasValue && this.retryWithRandomSaltMilliseconds.Value > 1)
+            {
+                lock (ClientRetryPolicy.RetryWithRandomLock)
+                {
+                    backoffMilliseconds += ClientRetryPolicy.RetryWithRandom.Next(1, this.retryWithRandomSaltMilliseconds.Value);
+                }
+            }
+
+            // Cap to remaining time and max backoff
+            backoffMilliseconds = Math.Min(backoffMilliseconds, Math.Min(remainingMilliseconds, this.retryWithMaxBackoffMilliseconds));
```
xinlian12: 🟢 **Suggestion — Design: Jitter is fully suppressed once backoff reaches the max plateau**

When `retryWithCurrentBackoffMilliseconds` reaches `retryWithMaxBackoffMilliseconds` (1000ms default), jitter adds [1,4]ms producing e.g. 1003ms. Then this `Math.Min` clamps it back: `min(1003, min(remaining, 1000))` = **1000** — because `maxBackoff` is always ≤ the jittered value.

**Why it matters:** At the plateau, all concurrent clients retry with exactly identical 1000ms delays, defeating jitter's thundering herd prevention. The existing @xinlian12 comment (#11) already noted this mechanism. Since this matches the direct-mode `GoneAndRetryWithRequestRetryPolicy` behavior, consistency is a valid reason to keep it — but if gateway mode sees higher concurrency than direct mode, consider applying jitter **after** the max cap by only capping to `remainingMilliseconds`, not `maxBackoffMilliseconds`:

```csharp
backoffMilliseconds = Math.Min(backoffMilliseconds, remainingMilliseconds);

```

No urgency if consistency with direct-mode is the priority.

xinlian12:
✅ **Review complete** (08:32)

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -1,752 +1,867 @@
-﻿//------------------------------------------------------------
```
Meghana-Palaparthi, line 1: NIT: revert the unnecessary whitespace changes

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
Meghana-Palaparthi: Agree. DTS always only runs in Gateway mode, so adding a defensive check `!this.isDtxRequest` to prevent retry amplification is a better approach

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
yash2710: This change happened due to a merge commit. Getting it fixed

yash2710:
@sdkReviewAgent-2

yash2710:
@sdkReviewAgent-2

xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -392,6 +422,19 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
...
 
+            // Received 449 (RetryWith) - transient concurrency error from the store.
+            // The proxy no longer retries 449 for external SDK requests to avoid noisy-neighbor amplification.
+            // The SDK handles retries with jittered exponential backoff.
+            // Only retry in Gateway mode or ThinClient mode. In Direct mode, GoneAndRetryWithRequestRetryPolicy
+            // already handles 449 at the transport layer — retrying here would cause double-retry amplification.
+            // Exclude DTX requests: DTX 449/5352 is handled by ShouldRetryDtxRequest with its own budget.
+            if (statusCode.HasValue && (int)statusCode.Value == 449
+                && (this.isGatewayClientMode || this.isThinClientEnabled)
+                && !this.isDtxRequest)
```
xinlian12, line 433: 🟡 **Recommendation — Testing: `!isDtxRequest` guard has zero test coverage in Gateway mode**

The `!isDtxRequest` exclusion on this line is production-critical — @Meghana-Palaparthi confirmed that DTX always runs in Gateway mode. This means every DTX 449/5352 in production flows through this guard to reach `ShouldRetryDtxRequest()` instead of `ShouldRetryOnRetryWithStatus()`.

However, no test exercises this guard. The existing DTX 449 tests (`DtxRequest_449_5352_ShouldRetry_WithDefaultRetryInterval` at line 991, `DtxRequest_449_5352_ShouldRetry_HonorsRetryAfterHeader` at line 1015) all construct the policy without `isGatewayClientMode`, which defaults to `false` (Direct mode):

```csharp
// Line 1000 — existing DTX test
ClientRetryPolicy policy = new ClientRetryPolicy(
    endpointManager, this.partitionKeyRangeLocationCache, new RetryOptions(), enableEndpointDiscovery, false);
    // ← isGatewayClientMode defaults to false (Direct mode)

```

In Direct mode, the 449 check at line 431 evaluates `(false || false)` → `false`, so the **mode guard** skips the new 449 handler — the `!isDtxRequest` guard is never reached. If someone later removes `!this.isDtxRequest` from line 433, all existing tests still pass, but in production (Gateway mode) DTX requests would silently hit the wrong retry handler — losing the DTX-specific count-based budget, `Retry-After` header honoring, and body-presence checks.

**Suggested fix** — add a test that exercises the DTX guard in Gateway mode:

```csharp
[TestMethod]
public async Task DtxRequest_449_InGatewayMode_ShouldUseDtxHandler_NotRetryWithHandler()
{
    // ... setup ...
    ClientRetryPolicy policy = new ClientRetryPolicy(
        endpointManager, this.partitionKeyRangeLocationCache, new RetryOptions(),
        enableEndpointDiscovery, isThinClientEnabled: false, isGatewayClientMode: true);

    DocumentServiceRequest request = CreateDtxRequest();
    policy.OnBeforeSendRequest(request);

    ResponseMessage response = new ResponseMessage((HttpStatusCode)StatusCodes.RetryWith);
    response.Headers.SubStatusCodeLiteral =
        DistributedTransactionConstants.DtcCoordinatorRaceConflict.ToString();

    ShouldRetryResult result = await policy.ShouldRetryAsync(response, CancellationToken.None);

    // DTX handler uses server-directed Retry-After, not jittered exponential backoff.
    // Verify the retry uses DTX semantics (default 10ms interval), not the
    // generic 449 handler's penalty-free first retry (0ms).
    Assert.IsTrue(result.ShouldRetry);
    Assert.IsTrue(result.BackoffTime.TotalMilliseconds >= 10,
        "DTX retry should use DTX handler's default interval, not the generic 449 handler's 0ms first retry.");
}

```

This also validates that the existing DTX 449 tests (which run in Direct mode) represent a scenario that doesn't occur in production. Consider updating them to pass `isGatewayClientMode: true` to match actual DTX behavior.

xinlian12:
✅ **Review complete** (46:39)

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
yash2710: @Meghana-Palaparthi can you take this up as a followup task?

Meghana-Palaparthi: I'll take this up in a follow-up PR

kushagraThapar:
**Missing `x-ms-noretry-449` proxy-coordination header**

Java's equivalent change ([azure-sdk-for-java#49332](https://github.com/Azure/azure-sdk-for-java/pull/49332), commit `85fadfc5046`) injects `x-ms-noretry-449: true` on every Gateway request to tell the proxy *"the SDK owns 449 retry — proxy, don't."* `ThinClientStoreModel` overrides to no-op since ThinClient doesn't go through the Gateway V1 retry loop.

This PR sends no such header. Two cases:
- If the proxy's 449-drop is gated on `x-ms-noretry-449`, .NET Gateway clients running this code would still get proxy-side 449 retry on top of the new SDK-side retry — at odds with the noisy-neighbor mitigation this PR is designed to provide.
- If the proxy's 449-drop is unconditional, this is a parity/consistency-with-Java item only.

References:
- Java header constant: `HttpConstants.java:122` → `NO_RETRY_449 = "x-ms-noretry-449"`
- Java injection site: `RxGatewayStoreModel.java:322-324` → `applyGatewayRetryWithHeaders`
- Java ThinClient override: `ThinClientStoreModel.java:81-83` (no-op)
- .NET worktree: `grep -rn "noretry-449\|NoRetry449" Microsoft.Azure.Cosmos/src/` → 0 matches

Can the proxy team confirm which case we're in? If gated, we'd need to mirror Java's header injection in `GatewayStoreClient` / `RequestInvokerHandler`, plus a ThinClient no-op override.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -1214,8 +1214,302 @@ private static GlobalPartitionEndpointManagerCore.PartitionKeyRangeFailoverInfo
...
 
-            return null;
-        }
+            return null;
+        }
+
+        /// <summary>
+        /// Tests that 449 (RetryWith) triggers a retry with backoff via the exception path.
+        /// </summary>
+        [TestMethod]
```
jeet1995, line 1223: Can we add fault injection-based tests? It might help to confirm the StopWatch doesn't get reset between retries.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -392,6 +422,19 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
                 return this.ShouldRetryOnUnavailableEndpointStatusCodes();
             }
 
+            // Received 449 (RetryWith) - transient concurrency error from the store.
+            // The proxy no longer retries 449 for external SDK requests to avoid noisy-neighbor amplification.
+            // The SDK handles retries with jittered exponential backoff.
+            // Only retry in Gateway mode or ThinClient mode. In Direct mode, GoneAndRetryWithRequestRetryPolicy
+            // already handles 449 at the transport layer — retrying here would cause double-retry amplification.
+            // Exclude DTX requests: DTX 449/5352 is handled by ShouldRetryDtxRequest with its own budget.
+            if (statusCode.HasValue && (int)statusCode.Value == 449
```
jeet1995, line 431: Curious why we decided to make this change here and not say in GatewayStoreClient where you wouldn't require to check (this.isGatewayClientMode || this.isThinClientEnabled)?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
jeet1995: @Meghana-Palaparthi we should track this in a GitHub issue if not already.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -31,14 +32,29 @@ internal sealed class ClientRetryPolicy : IDocumentClientRetryPolicy
...
+        // ----- RetryWith (449) jittered exponential backoff constants -----
+        private const int DefaultRetryWithInitialBackoffMilliseconds = 10;
+        private const int DefaultRetryWithMaxBackoffMilliseconds = 1000;
+        private const int DefaultRetryWithRandomSaltMilliseconds = 5;
+        private const int DefaultRetryWithTotalWaitTimeMilliseconds = 30000;
+        private const int RetryWithBackoffMultiplier = 2;
+
         private static readonly TimeSpan DtxInfraFailureBaseBackoff = TimeSpan.FromMilliseconds(100);
         private static readonly TimeSpan DtxInfraFailureMaxBackoff = TimeSpan.FromSeconds(5);
+        private static readonly object RetryWithRandomLock = new object();
```
jeet1995, line 45: Why do we have a static lock? That would be 1-lock behind which all operations queue up per deployment unit?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
xinlian12, line 431: 🟢 **Suggestion — Consistency: Use `StatusCodes.RetryWith` instead of magic number `449`**

The DTX handler in this same file uses the named constant at line 808:

```csharp
statusCodeValue == (int)StatusCodes.RetryWith

```

But this guard uses the raw integer literal `449`. Using the same named constant here improves grep-ability and makes it immediately clear these two code paths handle the same status code:

```csharp
if (statusCode.HasValue && (int)statusCode.Value == (int)StatusCodes.RetryWith
    && (this.isGatewayClientMode || this.isThinClientEnabled)
    && !this.isDtxRequest)

```

Minor consistency nit — no behavioral impact.

xinlian12:
✅ **Review complete** (54:00)

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -392,6 +422,19 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
                 return this.ShouldRetryOnUnavailableEndpointStatusCodes();
             }
 
+            // Received 449 (RetryWith) - transient concurrency error from the store.
+            // The proxy no longer retries 449 for external SDK requests to avoid noisy-neighbor amplification.
```
xinlian12, line 426: 🟡 **Recommendation — Cross-SDK: Java SDK sends `x-ms-noretry-449` header to disable proxy-side retry**

The Java SDK's `RxGatewayStoreModel` explicitly sends `x-ms-noretry-449: true` on every gateway request via `applyGatewayRetryWithHeaders()`, which tells the proxy to skip its server-side 449 retry. This .NET PR relies entirely on the proxy removing its 449 retry behavior unconditionally.

**Why it matters:** If the proxy's behavioral change is rolled out progressively (region-by-region or via gateway instance deployment), there is a window where:
1. Some gateway instances still retry 449 server-side (old behavior)
2. The SDK also retries 449 client-side (this PR's new behavior)
3. Result: the exact noisy-neighbor amplification this PR aims to prevent

The Java SDK avoids this transition-period risk by explicitly opting out of proxy retry via the header — a defense-in-depth approach that works regardless of proxy deployment state.

**Recommendation:** Confirm with the service team whether the proxy's 449 retry removal is unconditional for all SDK traffic, or whether it requires the `x-ms-noretry-449` header to opt in. If the latter, add header transmission in the gateway request path (e.g., in `GatewayStoreModel` or `GatewayStoreClient`).

xinlian12:
✅ **Review complete** (47:57)

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -31,14 +32,29 @@ internal sealed class ClientRetryPolicy : IDocumentClientRetryPolicy
         private const int MaxDtxRetryCount = 10;
         private const int MaxDtxInfraFailureRetryCount = 9;
         private const int DtxInfraFailureMaxExponent = 6;
+
+        // ----- RetryWith (449) jittered exponential backoff constants -----
+        private const int DefaultRetryWithInitialBackoffMilliseconds = 10;
+        private const int DefaultRetryWithMaxBackoffMilliseconds = 1000;
+        private const int DefaultRetryWithRandomSaltMilliseconds = 5;
+        private const int DefaultRetryWithTotalWaitTimeMilliseconds = 30000;
```
xinlian12, line 40: 💬 **Observation — Cross-SDK: No extended retry budget for strong consistency requests**

The Java SDK uses a **60-second** total wait time for 449 retries on strong consistency requests (`STRONG_GATEWAY_RETRY_WITH_TIMEOUT_IN_SECONDS = 60` in `GatewayRetryWithRetryPolicy.java`), compared to 30s for other consistency levels. The .NET direct-mode transport layer (`GoneAndRetryWithRequestRetryPolicy`) has a similar `StrongGoneAndRetryWithRetryTimeoutInSeconds = 60`.

This gateway-mode implementation uses a flat 30s regardless of consistency level. Strong consistency operations require quorum acknowledgment, which can increase contention and extend the window during which 449s occur. If the team observes elevated 449 exhaustion rates under strong consistency, this would be the lever to adjust.

No action needed for this PR — documenting as a future consideration since the direct-mode parity claim in the code comment (line 682) is accurate for the default case, but the strong-consistency extension is not carried over to Gateway mode.

xinlian12:
✅ **Review complete** (52:56)

PR #5827, opened 2026-05-04
ChangeFeedEstimator: Adds support for inmemory lease container

yash2710:
## Description

Change feed estimator threw `ArgumentNullException` when an inmemory lease container was being used. Update validations so in-memory lease containers work with change feed estimator

## Type of change


NaluTripician:
@sdkReviewAgent-2

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorRunner.cs
```diff
@@ -72,10 +72,16 @@ public void ApplyBuildConfiguration(
         {
             if (leaseContainer == null && customDocumentServiceLeaseStoreManager == null) throw new ArgumentNullException(nameof(leaseContainer));
 
+            DocumentServiceLeaseContainer documentServiceLeaseContainer = customDocumentServiceLeaseStoreManager?.LeaseContainer;
```
xinlian12, line 75: 🟢 **Suggestion — Consistency: Propagate `LeaseContainer` null-check to `ChangeFeedProcessorCore`**

This new validation (lines 75-79) is a good defense-in-depth pattern — it catches a buggy `DocumentServiceLeaseStoreManager` that returns null from `LeaseContainer` at config time rather than letting it NRE later during `InitializeLeaseStoreAsync`.

However, the sibling `ChangeFeedProcessorCore.ApplyBuildConfiguration` (lines 44-47) does **not** perform this same inner-null check — it only validates `customDocumentServiceLeaseStoreManager == null && leaseContainer == null` and then trusts the manager's properties. If a custom implementation provided a null `LeaseContainer`, the processor would fail later with a less helpful NRE during `InitializeAsync`.

Consider propagating this validation to `ChangeFeedProcessorCore` for symmetry:

```csharp
// In ChangeFeedProcessorCore.ApplyBuildConfiguration, after the existing null check:
if (leaseContainer == null && customDocumentServiceLeaseStoreManager?.LeaseContainer == null)
{
    throw new ArgumentNullException(nameof(customDocumentServiceLeaseStoreManager), 
        "The provided DocumentServiceLeaseStoreManager has a null LeaseContainer.");
}

```

Not blocking — the in-tree `DocumentServiceLeaseStoreManagerInMemory` always initializes a non-null `LeaseContainer`, so this guard is unreachable with current implementations. But making both paths consistent avoids future surprises.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedEstimatorRunnerTests.cs
```diff
@@ -258,6 +283,45 @@ static Task estimationDelegate(long estimation, CancellationToken token)
...
+            ChangeFeedEstimatorRunner estimator = new ChangeFeedEstimatorRunner(estimationDelegate, TimeSpan.FromSeconds(5));
+            estimator.ApplyBuildConfiguration(
+                leaseStoreManager.Object,
+                null, // No physical lease container (in-memory scenario)
+                "instanceName",
+                new ChangeFeedLeaseOptions() { LeasePrefix = "estimatorTest" },
+                new ChangeFeedProcessorOptions(),
+                ChangeFeedEstimatorRunnerTests.GetMockedContainer("monitored"));
+
+            try
```
xinlian12, line 317: 🟢 **Suggestion — Testing: Consider verifying a full estimation cycle completes**

This test validates that the object graph (`ChangeFeedEstimatorRunner` → `ChangeFeedEstimatorCore` → `ChangeFeedEstimatorIterator`) is constructed without throwing — which is the main fix this PR delivers. Good.

However, `StartAsync()` launches `FeedEstimatorRunner.RunAsync` asynchronously and `StopAsync()` is called immediately after, so the estimation loop may be cancelled before `ReadNextAsync` on the `ChangeFeedEstimatorIterator` ever executes. This means the iterator's `InitializeLeaseStoreAsync` skip-path (the `if (this.documentServiceLeaseContainer == null)` guard at line 338) is never exercised by this test.

Consider either:
- Adding a brief delay or `TaskCompletionSource`-based signal before `StopAsync()` to ensure one estimation cycle runs, or
- Adding a separate test that directly calls `ChangeFeedEstimatorCore.GetCurrentStateIterator().ReadNextAsync()` with a null `leaseContainer` and non-null `documentServiceLeaseContainer`

This would ensure that if someone later modifies the iterator's `InitializeLeaseStoreAsync` to unconditionally dereference `this.leaseContainer`, a test catches it.

xinlian12:
✅ **Review complete** (29:15)

yash2710:
@sdkReviewAgent-2

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorCore.cs
```diff
@@ -46,6 +46,11 @@ public void ApplyBuildConfiguration(
                 throw new ArgumentNullException(nameof(leaseContainer));
             }
 
+            if (leaseContainer == null && customDocumentServiceLeaseStoreManager?.LeaseContainer == null)
```
xinlian12, line 49: 🟡 **Recommendation — Testing: Two existing `ChangeFeedProcessorCoreTests` now silently validate the wrong code path**

This new validation at line 49 fires **before** the `instanceName` (line 56) and `monitoredContainer` (line 59) null checks. Two existing tests are affected:

**`ApplyBuildConfiguration_ValidatesNullInstance`** (line 42) and **`ApplyBuildConfiguration_ValidatesNullMonitoredContainer`** (line 56) both pass `Mock.Of<DocumentServiceLeaseStoreManager>()` with `leaseContainer: null`. Since `Mock.Of<>()` returns `null` for `.LeaseContainer` by default, this new check evaluates `leaseContainer == null && LeaseContainer == null` → `true` and throws `ArgumentNullException` before reaching the `instanceName` or `monitoredContainer` validations. Both tests still pass (they assert `[ExpectedException(typeof(ArgumentNullException))]` without checking the parameter name) but now test this lease-container guard instead of what they claim to test.

**Why it matters:** If someone later breaks the `instanceName` or `monitoredContainer` null validation, no test would catch it when a custom store manager is provided. The test names suggest coverage that doesn't actually exist.

**Fix:** Update both tests to set up a non-null `LeaseContainer` on their mocks, exactly as was done for `ApplyBuildConfiguration_ValidCustomStore`:

```csharp
// In both ValidatesNullInstance and ValidatesNullMonitoredContainer:
Mock<DocumentServiceLeaseStoreManager> leaseStoreManager = new Mock<DocumentServiceLeaseStoreManager>();
leaseStoreManager.Setup(l => l.LeaseContainer).Returns(Mock.Of<DocumentServiceLeaseContainer>);

// Then pass leaseStoreManager.Object instead of Mock.Of<DocumentServiceLeaseStoreManager>()

```

Additionally, consider adding a `ApplyBuildConfiguration_ValidatesCustomStoreWithNullLeaseContainer` test to `ChangeFeedProcessorCoreTests` (matching the one added to `ChangeFeedEstimatorRunnerTests`) to explicitly cover this new negative validation path.

xinlian12:
✅ **Review complete** (23:29)

yash2710:
@sdkReviewAgent-2

xinlian12:
✅ **Review complete** (25:26)

No new comments — existing review coverage is sufficient.

yash2710:
@sdkReviewAgent-2

xinlian12:
✅ **Review complete** (33:30)

No new comments — existing review coverage is sufficient.

kushagraThapar (commented):
Overall looks good, thanks @yash2710 , can you please add the changelog entry.

yash2710:
> Overall looks good, thanks @yash2710 , can you please add the changelog entry.

@Copilot add a change log entry for this change

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedEstimatorRunnerTests.cs
```diff
@@ -258,6 +283,45 @@ static Task estimationDelegate(long estimation, CancellationToken token)
             }
         }
 
+        [TestMethod]
+        public async Task StartAsync_WithInMemoryLeaseContainer_BuildsEstimatorSuccessfully()
```
ananth7592, line 287: Hey Yash, — thanks for taking care of this regression! I went through the new test in ChangeFeedEstimatorRunnerTests.cs and wanted to share some thoughts on the coverage.

The new test StartAsync_WithInMemoryLeaseContainer_BuildsEstimatorSuccessfully does the important work of catching the original bug:

   1. StartAsync → InitializeLeaseStoreAsync skip-branch (correctly skipped when documentServiceLeaseContainer != null)
   2. BuildFeedEstimatorRunner constructs a real ChangeFeedEstimatorCore with leaseContainer: null, dsc: mock — the exact ctor that used to throw

  So as a regression test for this bug, it works — if anyone reverts your ChangeFeedEstimatorCore ctor change, it will fail here. Nice.

  Where I think we have a coverage gap:

   - StartAsync is fire-and-forget at this.runAsync = this.feedEstimatorRunner.RunAsync(...) — the StopAsync() in finally cancels the token almost immediately, so the iterator never actually runs
   - The ChangeFeedEstimatorIterator is therefore never instantiated during the test
   - There's no explicit assertion, so if StartAsync ever becomes a no-op, the delegate is never invoked, or RunAsync swallows an exception, this test will silently keep passing
   - The runner→iterator wiring boundary isn't really exercised — your separate iterator test (ShouldWorkWithNullLeaseContainer_WhenDocumentServiceLeaseContainerProvided) is great for the iterator in isolation, but a future refactor at the runner that breaks how it composes the   iterator would slip through

  Suggestion: the pattern from StartAsync_TriggersDelegate is already a great template — it wires up mockedIterator → ReadNextAsync → FeedResponse, awaits StartAsync, polls receivedEstimation, then Assert.AreEqual(remainingWork, estimationDelegateValue). Could we add a sibling   test that combines that pattern with your in-memory lease setup? I'd keep the existing "doesn't throw during build" test as well (maybe add a small Assert.IsNotNull(estimator) so it has at least one explicit assertion) and add the behavioral one alongside.

  This was @xinlian12's original concern too — you addressed it at the iterator level (which is good!), this would just close the loop at the runner→iterator boundary.

changelog.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 
 - [5783](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5783) Container: Fixes SemanticRerankAsync TypeLoadException in derived classes
+- [5827](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5827) Change Feed Estimator: Fixes `ArgumentNullException` when using an in-memory lease container
```
kushagraThapar: The changelog structure has changed, worth updating your PR with the main branch to reflect the latest bits.

kushagraThapar (dismissed):

yash2710:
Addressed in 58845301c (on top of the merge with `main`):

- **xinlian12 / ananth7592 (runner→iterator coverage):** Added `Assert.IsNotNull(estimator)` to `StartAsync_WithInMemoryLeaseContainer_BuildsEstimatorSuccessfully` so the build-only test has an explicit assertion, and added a new `StartAsync_TriggersDelegate_WithInMemoryLeaseContainer` test modeled on `StartAsync_TriggersDelegate`. It composes the runner with a custom `DocumentServiceLeaseStoreManager` (non-null `LeaseContainer`) and `leaseContainer: null`, mocks `GetCurrentStateIterator(...).ReadNextAsync(...)`, and asserts the estimation delegate receives the mocked remaining work — exercising the runner → iterator wiring at the in-memory lease boundary.
- **kushagraThapar (changelog):** Merged `main` into the branch and moved the entry to the new `### Unreleased` → `#### Features Added` section as `[5827] ChangeFeedEstimator: Adds support for in-memory lease container.` (the prior entry had landed under the already-released `3.61.0-preview.0` section).
- xinlian12''s earlier suggestions to propagate the null-check to `ChangeFeedProcessorCore` and to update the `ChangeFeedProcessorCoreTests` mocks (plus add `ApplyBuildConfiguration_ValidatesCustomStoreWithNullLeaseContainer`) were already covered in the prior `Address comments` commits.

All 42 tests in `ChangeFeedEstimatorRunnerTests` + `ChangeFeedProcessorCoreTests` + `ChangeFeedEstimatorIteratorTests` pass locally.

kushagraThapar (dismissed):

ananth7592 (dismissed):
LGTM, Thanks Yash

changelog.md
```diff
@@ -20,12 +20,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Features Added
 
 - [5838](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5838) EmbeddingGenerator: Adds ICosmosEmbeddingGenerator client-wide configuration (preview)
-- [5827](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5827) ChangeFeedEstimator: Adds support for in-memory lease container.
 
 #### Breaking Changes
 
 #### Bugs Fixed
 
+- [5827](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5827) ChangeFeedEstimator: Change feed estimator threw `ArgumentNullException` when an inmemory lease container was being used. Update validations so in-memory lease containers work with change feed estimator
```
kushagraThapar, line 28: 🥇

kushagraThapar (approved):

ananth7592 (approved):

Issue #5837, opened 2026-05-05
[Embedding V0] Emulator tests + baseline regeneration


ananth7592:
# Emulator tests + baseline regeneration

Part of: #5830 (V0 embedding generation).

## Scope

1. Extend `Microsoft.Azure.Cosmos.EmulatorTests/Query/HybridSearchQueryTests.cs` with at least one end-to-end case that uses an `IEmbeddingGenerator` and exercises the new `embeddingParameterMap` flow.
   - A simple `FakeEmbeddingGenerator` returning deterministic vectors (e.g. `text.Length`-derived) is fine.
   - Cover both single-component and multi-component (RRF) hybrid queries.
2. Verify ServiceInterop bypass: when `EmbeddingGenerator` is configured, the gateway plan path is taken even on Windows-x64 (where SI is normally used).
3. Regenerate `EndToEndTraceWriterBaselineTests` baselines for the hybrid-search test cases that newly contain the "Embedding Generation" trace child.
   - Inspect the diff to confirm only the expected new node appears and it does not contain customer data (no raw text, no vectors).

## Acceptance criteria

- New emulator test passes against the local emulator.
- Regenerated baselines reviewed and committed.
- Existing trace baselines for non-embedding queries unchanged.

## Files likely touched

- `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Query/HybridSearchQueryTests.cs`
- `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/Tracing/EndToEndTraceWriterBaselineTests.cs` and its baseline `.json`/`.xml` files.

## Dependencies

- `plumb-into-pipeline` (#5834).
- `diagnostics` (#5835).

Issue #5832, opened 2026-05-05
[Embedding V0] DTO additions + new QueryFeatures.EmbeddingGeneration bit


ananth7592:
# DTO additions + new `QueryFeatures.EmbeddingGeneration` bit

Part of: #5830 (V0 embedding generation).

## Scope

1. Add field to both plan DTOs:

   ```csharp
   [JsonProperty("embeddingParameterMap")]
   public IReadOnlyDictionary<string, string> EmbeddingParameterMap { get; set; }
   ```

   in:
   - `Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryInfo.cs`
   - `Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/HybridSearchQueryInfo.cs`

   The example XML from the design discussion shows the map under the leaf `<queryInfo>`, but pure-vector ORDER BY queries can also produce it — covering both DTOs is safer.

2. Add the new flag in `Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryFeature.cs`:

   ```csharp
   EmbeddingGeneration = 1 << 17,
   ```

3. In `QueryPlanRetriever.cs`, make the supported-features string **per-request** (the way `SupportedQueryFeaturesWithHybridSearchQueryPlanOptimizationDisabledString` is built today) and only OR-in the new bit when the request actually has an `IEmbeddingGenerator` configured. This keeps gateway behavior unchanged for existing customers.

## Open items to confirm before merging

- Exact wire shape of `embeddingParameterMap` in the GW HTTP JSON response (object vs array of `{key,value}`). The XML shown in the design doc is an internal representation.
- Final bit name and value confirmed with the gateway team.

## Acceptance criteria

- Plan DTO JSON deserializes the new field correctly (round-trip unit test).
- Supported-features string contains the new bit only when `EmbeddingGenerator` is configured.
- Existing supported-features tests still pass (no regression for default code path).

## Files likely touched

- `Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryInfo.cs`
- `Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/HybridSearchQueryInfo.cs`
- `Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryFeature.cs`
- `Microsoft.Azure.Cosmos/src/Query/Core/QueryPlan/QueryPlanRetriever.cs`

## Dependencies

[...]

PR #5838, opened 2026-05-05
EmbeddingGenerator: Adds ICosmosEmbeddingGenerator client-wide configuration (preview)

ananth7592:
Adds the **public surface only** for V0 query-time embedding generation in the .NET SDK (preview). No runtime behavior is wired up yet — pipeline resolver and binding land in a follow-up PR.

### Interface and result type

```csharp
public interface ICosmosEmbeddingGenerator
{
    Task<CosmosEmbeddingResult> GenerateEmbeddingsAsync(
        IReadOnlyList<string> texts,
        string endpoint,
        string deploymentName,
        int dimensions,
}

public sealed class CosmosEmbeddingResult
{
    public IReadOnlyList<ReadOnlyMemory<float>> Vectors { get; }
    public int? TotalTokens { get; }
    public TimeSpan? Latency { get; }
}
```

Signature mirrors [Python SDK PR #46902](https://github.com/Azure/azure-sdk-for-python/pull/46902).

### Client-wide configuration

```csharp
var client = new CosmosClientBuilder(endpoint, cred)
    .WithEmbeddingGenerator(myGenerator)
    .Build();

ICosmosEmbeddingGenerator current = client.EmbeddingGenerator;
```

Or via options:

```csharp
var client = new CosmosClient(endpoint, key, new CosmosClientOptions
{
    EmbeddingGenerator = myGenerator,
});
```

### Source (new)
- `Microsoft.Azure.Cosmos/src/ICosmosEmbeddingGenerator.cs`
- `Microsoft.Azure.Cosmos/src/CosmosEmbeddingResult.cs`

### Source (modified)
- `CosmosClientOptions.cs` — adds `EmbeddingGenerator` (`[JsonIgnore]`)
- `Fluent/CosmosClientBuilder.cs` — adds `WithEmbeddingGenerator(...)`
- `CosmosClient.cs` — adds `EmbeddingGenerator` read-only accessor

### Tests
- `CosmosClientOptionsUnitTests.cs`:
  - `VerifyEmbeddingGeneratorBuilderProperties` — builder round-trip + null-arg guard
  - `CosmosClient_EmbeddingGenerator_ReturnsConfiguredInstance` — read-only accessor surfaces builder- and options-set instances
  - `CosmosClientOptions_Clone_PreservesEmbeddingGenerator` — locks down `Clone()` preservation
- `Contracts/DotNetPreviewSDKAPI.net6.json` — regenerated for the new surface

## Validation

- ✅ PREVIEW build (`/p:IsPreview=true`): 0 errors
- ✅ Non-PREVIEW build: 0 errors
- ✅ `ContractEnforcement` + `SettingsContractTests` + `CosmosClientOptionsUnitTests`: PREVIEW 220/220 ; non-PREVIEW 221/221
- ✅ Baseline regenerated with LF line endings preserved

### DIFF
--- a/Microsoft.Azure.Cosmos/src/CosmosClient.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClient.cs
@@ -607,6 +607,15 @@ namespace Microsoft.Azure.Cosmos
         /// <remarks>This property is read-only. Modifying any options after the client has been created has no effect on the existing client instance.</remarks>
         public virtual CosmosClientOptions ClientOptions => this.ClientContext.ClientOptions;
 
+#if PREVIEW
+        /// <summary>
+        /// Gets the client-wide <see cref="ICosmosEmbeddingGenerator"/>, or <c>null</c> if none was set.
+        /// Set via <see cref="CosmosClientOptions.EmbeddingGenerator"/> or
+        /// <see cref="Fluent.CosmosClientBuilder.WithEmbeddingGenerator"/>.
+        /// </summary>
+        public virtual ICosmosEmbeddingGenerator EmbeddingGenerator => this.ClientContext.ClientOptions.EmbeddingGenerator;
+#endif
+
         /// <summary>
         /// The response factory used to create CosmosClient response types.
         /// </summary>
--- a/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
@@ -406,6 +406,18 @@ namespace Microsoft.Azure.Cosmos
         ReadConsistencyStrategy? ReadConsistencyStrategy { get; set; }
 
         /// <summary>
+        /// Gets or sets the client-wide default <see cref="ICosmosEmbeddingGenerator"/> used to generate
+        /// query-time vector embeddings for hybrid and vector-search queries.
+        /// </summary>
+        [JsonIgnore]
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        ICosmosEmbeddingGenerator EmbeddingGenerator { get; set; }
+
+        /// <summary>
         /// Sets the priority level for requests created using cosmos client.
         /// </summary>
         /// <remarks>
--- a/Microsoft.Azure.Cosmos/src/CosmosEmbeddingResult.cs
+++ b/Microsoft.Azure.Cosmos/src/CosmosEmbeddingResult.cs
@@ -1,0 +1,69 @@
+//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// The result of a call to <see cref="ICosmosEmbeddingGenerator.GenerateEmbeddingsAsync"/>.
+    /// Carries the generated float32 vectors plus optional diagnostic fields (token usage,
+    /// latency) the SDK surfaces through <c>CosmosDiagnostics</c>.
+    /// </summary>
+#if PREVIEW
+    public
+#else
+    internal
+#endif
+    sealed class CosmosEmbeddingResult
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="CosmosEmbeddingResult"/>.
+        /// </summary>
+        /// <param name="vectors">
+        /// The generated float32 embedding vectors, one per input string supplied to the
+        /// originating <see cref="ICosmosEmbeddingGenerator.GenerateEmbeddingsAsync"/> call,
+        /// in the same order as the inputs.
+        /// </param>
+        /// <param name="totalTokens">
+        /// Optional total token count consumed by the embedding service to produce these vectors.
+        /// Pass <c>null</c> when the underlying service does not report token usage.
+        /// </param>
+        /// <param name="latency">
+        /// Optional duration the implementation observed for the embedding service call (for
+        /// example, the wall-clock time around the underlying HTTP request). Surfaced through
+        /// <c>CosmosDiagnostics</c> for query-time observability. Pass <c>null</c> when the
+        /// implementation does not measure latency.
+        /// </param>
+        public CosmosEmbeddingResult(
+            IReadOnlyList<ReadOnlyMemory<float>> vectors,
+            int? totalTokens = null,
+            TimeSpan? latency = null)
+        {
+            this.Vectors = vectors ?? throw new ArgumentNullException(nameof(vectors));
+            this.TotalTokens = totalTokens;
+            this.Latency = latency;
+        }
+
+        /// <summary>
+        /// Gets the generated float32 embedding vectors, one per input string, in the same
+        /// order as the inputs supplied to <see cref="ICosmosEmbeddingGenerator.GenerateEmbeddingsAsync"/>.
+        /// </summary>
+        public IReadOnlyList<ReadOnlyMemory<float>> Vectors { get; }
+
+        /// <summary>
+        /// Gets the total number of tokens the embedding service consumed to generate
+        /// <see cref="Vectors"/>, or <c>null</c> when the underlying service does not report it.
+        /// </summary>
+        public int? TotalTokens { get; }
+
+        /// <summary>
+        /// Gets the duration the implementation observed for the underlying embedding service
+        /// call, or <c>null</c> when the implementation does not measure it. Surfaced through
+        /// <c>CosmosDiagnostics</c> for query-time observability.
+        /// </summary>
+        public TimeSpan? Latency { get; }
+    }
+}
--- a/Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
+++ b/Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
@@ -869,5 +869,23 @@ namespace Microsoft.Azure.Cosmos.Fluent
             this.clientOptions.ReadConsistencyStrategy = readConsistencyStrategy;
             return this;
         }
+
+        /// <summary>
+        /// Sets the client-wide default <see cref="ICosmosEmbeddingGenerator"/> used to generate
+        /// query-time vector embeddings for hybrid and vector-search queries.
+        /// </summary>
+        /// <param name="embeddingGenerator">The embedding generator to use as the client-wide default.</param>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="embeddingGenerator"/> is <c>null</c>.</exception>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        CosmosClientBuilder WithEmbeddingGenerator(ICosmosEmbeddingGenerator embeddingGenerator)
+        {
+            this.clientOptions.EmbeddingGenerator = embeddingGenerator ?? throw new ArgumentNullException(nameof(embeddingGenerator));
+            return this;
+        }
     }
 }
--- a/Microsoft.Azure.Cosmos/src/ICosmosEmbeddingGenerator.cs
+++ b/Microsoft.Azure.Cosmos/src/ICosmosEmbeddingGenerator.cs
@@ -1,0 +1,113 @@
+//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Defines a contract for generating float32 vector embeddings from input text strings
+    /// supplied by the Azure Cosmos DB query pipeline.
+    /// The SDK invokes this when a query plan contains <c>GenerateEmbeddings(...)</c> literals
+    /// (for example <c>VectorDistance(GenerateEmbeddings("big brown cat"), c.embedding)</c>).
+    /// Set a client-wide default via <c>CosmosClientOptions.EmbeddingGenerator</c> or
+    /// <c>CosmosClientBuilder.WithEmbeddingGenerator</c>. Implementations MUST be thread-safe and are
+    /// responsible for any caching, retries, and authentication required to call the underlying
+    /// embedding service.
+    /// </summary>
+    /// <remarks>
+    /// <para><b>Preview surface.</b> The SDK call site that invokes this method is delivered
+    /// in a follow-up release. Setting an instance via
+    /// <see cref="CosmosClientOptions.EmbeddingGenerator"/> or
+    /// <see cref="Fluent.CosmosClientBuilder.WithEmbeddingGenerator"/> has no runtime effect
+    /// today; the surface is shipped in this preview so customers can author and test
+    /// implementations against the contract ahead of the resolver landing.</para>
+    /// <para><b>Lifecycle and disposal.</b> The customer owns the generator instance. The SDK
+    /// keeps a reference for the lifetime of the configured <see cref="CosmosClient"/> (or the
+    /// <see cref="Container"/> reference it was bound to) but never disposes it. If the
+    /// implementation holds disposable resources (for example an <c>HttpClient</c> or an
+    /// <c>EmbeddingClient</c>), the customer is responsible for disposing them when their
+    /// application tears down.</para>
+    ///
+    /// <para><b>Error semantics.</b> Implementations are responsible for handling transient
+    /// failures from the underlying embedding service (network errors, rate limiting, etc.)
+    /// via their own retry policy. The SDK does not retry calls to this method. Any exception
+    /// thrown by the implementation is wrapped into a <see cref="CosmosException"/> and
+    /// surfaced to the originating SDK caller.</para>
+    ///
+    /// <para><b>Cancellation.</b> Implementations should honor the supplied
+    /// <see cref="CancellationToken"/> cooperatively wherever feasible (typically by forwarding
+    /// it to the underlying HTTP call). Best-effort cancellation is acceptable; ignoring the
+    /// token entirely is discouraged because it defeats caller-side timeouts.</para>
+    ///
+    /// <para><b>Idempotency and concurrency.</b> The SDK may invoke this method multiple times
+    /// for the same inputs (for example during internal query retry) and may invoke it
+    /// concurrently from multiple threads. Implementations must be safe to call repeatedly
+    /// and from parallel callers, and must not assume per-call state. Note that each call
+    /// typically incurs cost at the underlying embedding service; implementations may cache
+    /// responses internally if they want to avoid duplicate billing for identical inputs.</para>
+    /// </remarks>
+#if PREVIEW
+    public
+#else
+    internal
+#endif
+    interface ICosmosEmbeddingGenerator
+    {
+        /// <summary>
+        /// Generates an embedding vector for each of the supplied input strings.
+        /// </summary>
+        /// <param name="texts">
+        /// The input strings to embed, in the order the implementation MUST preserve in the
+        /// returned <see cref="CosmosEmbeddingResult.Vectors"/> (one vector per input, same
+        /// index). Typed as <see cref="IReadOnlyList{T}"/> so implementations can size their
+        /// outbound batch without re-enumeration and so the 1:1 ordered contract is encoded
+        /// in the signature.
+        /// </param>
+        /// <param name="endpoint">
+        /// The embedding service endpoint to call (for example the Azure OpenAI account endpoint).
+        /// Sourced from the container's <c>EmbeddingSource.Endpoint</c> when configured.
+        /// </param>
+        /// <param name="deploymentName">
+        /// The model deployment name to invoke at <paramref name="endpoint"/>. Sourced from the
+        /// container's <c>EmbeddingSource.DeploymentName</c> when configured.
+        /// </param>
+        /// <param name="dimensions">
+        /// The vector dimensionality the produced embeddings must match. For models that support
+        /// dimensionality reduction (for example <c>text-embedding-3-small</c> /
+        /// <c>text-embedding-3-large</c>), implementations MUST forward this value to the
+        /// underlying service so the returned vectors have the expected length; otherwise the
+        /// service returns its default size, which may not match the container's
+        /// <see cref="VectorEmbeddingPolicy"/>.
+        /// </param>
+        /// <param name="cancellationToken">
+        /// A <see cref="CancellationToken"/> propagated from the originating SDK call
+        /// (for example <c>FeedIterator.ReadNextAsync</c>). Implementations should honor cancellation.
+        /// </param>
+        /// <returns>
+        /// A task that resolves to a <see cref="CosmosEmbeddingResult"/> whose
+        /// <see cref="CosmosEmbeddingResult.Vectors"/> contains one float32 vector per input,
+        /// each of length <paramref name="dimensions"/>, in the same order as
+        /// <paramref name="texts"/>.
+        /// <para>
+        /// Query-time vectors are sent to the Azure Cosmos DB gateway as float32 regardless of
+        /// the container's stored <see cref="VectorDataType"/>. Implementations targeting
+        /// containers configured for <see cref="VectorDataType.Uint8"/>,
+        /// <see cref="VectorDataType.Int8"/>, or <see cref="VectorDataType.Float16"/> storage
+        /// should still produce float32 vectors here; the Azure Cosmos DB service applies the
+        /// configured quantization at write time. This contract
+        /// covers all four <see cref="VectorDataType"/> storage configurations supported by
+        /// the container's <see cref="VectorEmbeddingPolicy"/>.
+        /// </para>
+        /// </returns>
+        Task<CosmosEmbeddingResult> GenerateEmbeddingsAsync(
+            IReadOnlyList<string> texts,
+            string endpoint,
+            string deploymentName,
+            int dimensions,
+            CancellationToken cancellationToken = default);
+    }
+}
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.net6.json
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.net6.json
@@ -42,6 +42,22 @@
       },
       "NestedTypes": {}
     },
+    "Microsoft.Azure.Cosmos.CosmosClient;System.Object;IsAbstract:False;IsSealed:False;IsInterface:False;IsEnum:False;IsClass:True;IsValueType:False;IsNested:False;IsGenericType:False;IsSerializable:False": {
+      "Subclasses": {},
+      "Members": {
+        "Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator EmbeddingGenerator": {
+          "Type": "Property",
+          "Attributes": [],
+          "MethodInfo": "Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator EmbeddingGenerator;CanRead:True;CanWrite:False;Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator get_EmbeddingGenerator();IsAbstract:False;IsStatic:False;IsVirtual:True;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+        },
+        "Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator get_EmbeddingGenerator()": {
+          "Type": "Method",
+          "Attributes": [],
+          "MethodInfo": "Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator get_EmbeddingGenerator();IsAbstract:False;IsStatic:False;IsVirtual:True;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+        }
+      },
+      "NestedTypes": {}
+    },
     "Microsoft.Azure.Cosmos.CosmosClientOptions;System.Object;IsAbstract:False;IsSealed:False;IsInterface:False;IsEnum:False;IsClass:True;IsValueType:False;IsNested:False;IsGenericType:False;IsSerializable:False": {
       "Subclasses": {},
       "Members": {
@@ -54,6 +70,20 @@
           "Type": "Method",
           "Attributes": [],
           "MethodInfo": "Boolean get_EnableRemoteRegionPreferredForSessionRetry();IsAbstract:False;IsStatic:False;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+        },
+        "Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator EmbeddingGenerator[Newtonsoft.Json.JsonIgnoreAttribute()]": {
+          "Type": "Property",
+          "Attributes": [
+            "JsonIgnoreAttribute"
+          ],
+          "MethodInfo": "Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator EmbeddingGenerator;CanRead:True;CanWrite:True;Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator get_EmbeddingGenerator();IsAbstract:False;IsStatic:False;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;Void set_EmbeddingGenerator(Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator);IsAbstract:False;IsStatic:False;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+        },
+        "Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator get_EmbeddingGenerator()[System.Runtime.CompilerServices.CompilerGeneratedAttribute()]": {
+          "Type": "Method",
+          "Attributes": [
+            "CompilerGeneratedAttribute"
+          ],
+          "MethodInfo": "Microsoft.Azure.Cosmos.ICosmosEmbeddingGenerator get_EmbeddingGenerator();IsAbstract:False;IsStatic:False;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
         },
         "System.Nullable`1[Microsoft.Azure.Cosmos.ReadConsistencyStrategy] get_ReadConsistencyStrategy()[System.Runtime.CompilerServices.CompilerGeneratedAttribute()]": {
           "Type": "Method",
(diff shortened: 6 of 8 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosQueryRequestOptionsUniTests.cs
```diff
@@ -22,5 +24,35 @@ public void StatelessTest()
...
+            Assert.IsNull(requestOptions.EmbeddingGenerator);
+        }
+
+        [TestMethod]
+        public void EmbeddingGenerator_RoundTripsAssignedInstance()
+        {
+            IEmbeddingGenerator generator = new TestEmbeddingGenerator();
+            QueryRequestOptions requestOptions = new QueryRequestOptions
+            {
+                EmbeddingGenerator = generator,
```
FabianMeiswinkel: Change to add this Property to QueryRequestOptions is missing?

Microsoft.Azure.Cosmos/src/RequestOptions/QueryRequestOptions.cs
```diff
@@ -247,6 +247,38 @@ public ConsistencyLevel? ConsistencyLevel
...
+        /// If the gateway returns a plan that requires embedding generation
+        /// but this property is <c>null</c>, the SDK throws an exception
+        /// describing how to set the generator.
+        /// </para>
+        /// </remarks>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
```
FabianMeiswinkel: Like many other configs there should be inheritance - RequestOptions preferred but if null and CosmosClientOptons has it defined that use that

FabianMeiswinkel: So exception would b ethrown when embedding is required if neither request options nor cleintoptions have it set?

FabianMeiswinkel (changes requested):
Client Options need this as well?

kushagraThapar (commented):
Thanks @ananth7592 , overall looks good, I have two non-blocking comments: 

1. **Thread-safety.** When set at the client level, the SDK will call `GenerateEmbeddingsAsync` concurrently from parallel queries/partition reads on a single
 instance. Please document on the interface that implementations must be safe to invoke concurrently — otherwise customers may wire up stateful implementations and may hit 
race conditions.
 
 2. **No per-request opt-out.** `QueryRequestOptions.EmbeddingGenerator = null` falls through to the client-level default; there's no way to say "use no 
generator for this call." Either document that as intentional or we can add a sentinel (cf. `NullEmbeddingGenerator` in `Microsoft.Extensions.AI`).

FabianMeiswinkel (dismissed):
LGTM

xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/src/IEmbeddingGenerator.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// Implementations may be invoked concurrently with other SDK operations
+    /// and should be thread-safe.
+    /// </para>
+    /// </remarks>
+#if PREVIEW
+    public
+#else
+    internal
+#endif
+    interface IEmbeddingGenerator
```
xinlian12: 🟡 **Recommendation — Name Collision With `Microsoft.Extensions.AI.IEmbeddingGenerator`**

`Microsoft.Extensions.AI` (shipping with .NET 10 and available as a NuGet package) exports `IEmbeddingGenerator<TInput, TEmbedding>`. While the generic arity differs (non-generic vs generic), any project referencing both `Microsoft.Azure.Cosmos` (preview) and `Microsoft.Extensions.AI` will get ambiguous reference errors requiring fully-qualified names or `using` aliases everywhere.

This is not hypothetical — the most common embedding implementations (Azure.AI.OpenAI, Ollama, etc.) implement `Microsoft.Extensions.AI.IEmbeddingGenerator<string, Embedding<float>>`. Users will likely have both in scope.

The SDK's existing convention is to prefix public types with `Cosmos`: `CosmosSerializer`, `CosmosLinqSerializer`, `CosmosDiagnostics`, `CosmosResponseFactory`. Renaming to `ICosmosEmbeddingGenerator` would eliminate the collision and follow the established pattern.

xinlian12: 🔴 **Blocking — Missing Contract Baseline Update**

The PR adds preview-visible public API surface (`IEmbeddingGenerator`, `EmbeddingGenerator` properties, `WithEmbeddingGenerator`), but `DotNetPreviewSDKAPI.net6.json` is not updated — I confirmed it contains zero `EmbeddingGenerator` or `IEmbeddingGenerator` entries.

The `ContractEnforcementTests.PreviewContractChanges()` test reflects over the built DLL, computes the preview-only delta, and compares against this baseline via strict JSON equality. When `IEmbeddingGenerator` appears as `public` in a PREVIEW build but is absent from the baseline, the test fails with `"Public API has changed."` The existing `ReadConsistencyStrategy` entries in this file confirm the precedent — all PREVIEW-gated APIs must be present.

The PR description mentions the file was "regenerated," but it is not in the changed files list (only 7 files changed, none are contract JSON).

**Action:** Run `UpdateContracts.ps1` and include the regenerated `DotNetPreviewSDKAPI.net6.json`.

Microsoft.Azure.Cosmos/src/IEmbeddingGenerator.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <param name="cancellationToken">
+        /// A <see cref="CancellationToken"/> propagated from the originating
+        /// SDK call (for example <c>FeedIterator.ReadNextAsync</c>).
+        /// Implementations should honor cancellation.
+        /// </param>
+        /// <returns>
+        /// A task that resolves to a sequence of embedding vectors with the
+        /// same cardinality and ordering as <paramref name="text"/>.
+        /// </returns>
+        Task<IEnumerable<ReadOnlyMemory<double>>> GenerateEmbeddingsAsync(
```
xinlian12: 🟡 **Recommendation — `double` Misaligns With SDK's Own Vector Data Model and Ecosystem**

The return type uses `ReadOnlyMemory<double>` (64-bit), but the SDK's own `VectorDataType` enum defines `Float32`, `Uint8`, `Int8`, `Float16` — there is no `Float64`/`double` variant. The broader ecosystem converges on `float`:

| Source | Element Type |
|--------|-------------|

| SDK `VectorDataType.Float32` | `float` (32-bit) |

| OpenAI API | `float` |

| Azure OpenAI SDK | `ReadOnlyMemory<float>` |

| `Microsoft.Extensions.AI` | `Embedding<float>` → `ReadOnlyMemory<float>` |

Using `double` means: every embedding provider will up-cast `float→double` on production, and the SDK will down-cast `double→float` when injecting parameters into `float32` vector indexes. This doubles memory per element for no benefit and creates unnecessary friction for implementors bridging to standard libraries.

Since this is preview-gated, now is the time to change to `ReadOnlyMemory<float>` before downstream PRs (#5832–#5837) build on this type.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -405,6 +405,23 @@ public ConnectionMode ConnectionMode
...
+        /// also set on a specific request, the request-level value takes precedence.
+        /// If the gateway returns a plan that requires embedding generation and neither the request-level
+        /// nor this property is set, the SDK throws an exception describing how to configure a generator.
+        /// </remarks>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        IEmbeddingGenerator EmbeddingGenerator { get; set; }
```
xinlian12: 🔴 **Blocking — Missing `[JsonIgnore]` Will Crash Diagnostic Serialization**

`CosmosClientOptions.GetSerializedConfiguration()` (line 1404) calls `JsonConvert.SerializeObject(this)`. In PREVIEW builds, `EmbeddingGenerator` is `public`, so JSON.NET will attempt to serialize whatever `IEmbeddingGenerator` implementation the user provides. A typical implementation holding an `HttpClient`, API keys, or other non-serializable members will throw a `JsonSerializationException`.

The two analogous non-serializable reference-type properties in this class both have `[JsonIgnore]`:
- `WebProxy` (line 687): `[JsonIgnore]`
- `HttpClientFactory` (line 818): `[JsonIgnore]`

**Concrete failure scenario:** User sets `CosmosClientOptions.EmbeddingGenerator` to an implementation wrapping an OpenAI client → any code path calling `GetSerializedConfiguration()` (e.g., `ClientConfigurationTraceDatum`) throws, crashing or corrupting diagnostic output.

**Fix:** Add `[JsonIgnore]` above this property.

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -869,5 +869,26 @@ CosmosClientBuilder WithReadConsistencyStrategy(Cosmos.ReadConsistencyStrategy r
...
+        /// <param name="embeddingGenerator">The embedding generator to use as the client-wide default.</param>
+        /// <returns>The current <see cref="CosmosClientBuilder"/>.</returns>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        CosmosClientBuilder WithEmbeddingGenerator(IEmbeddingGenerator embeddingGenerator)
+        {
+            this.clientOptions.EmbeddingGenerator = embeddingGenerator;
```
xinlian12: 🟡 **Recommendation — Missing Null Guard in Builder Method**

`WithEmbeddingGenerator` accepts a reference type without null validation:

```csharp
this.clientOptions.EmbeddingGenerator = embeddingGenerator; // null silently accepted

```

Compare with `WithHttpClientFactory` (line 667):

```csharp
this.clientOptions.HttpClientFactory = httpClientFactory 
    ?? throw new ArgumentNullException(nameof(httpClientFactory));

```

A `With*` builder method implies "set to this value." Passing `null` should be caught early rather than silently making the property no-op. If the intent is to allow clearing via `null`, that's worth documenting explicitly.

**Fix:** `this.clientOptions.EmbeddingGenerator = embeddingGenerator ?? throw new ArgumentNullException(nameof(embeddingGenerator));`

xinlian12:
✅ **Review complete** (39:17)

ananth7592:
@xinlian12  Have responded or resolved the comments . Please take a look, Thank you

@kushagraThapar  That was a good observation. I have updated the doc with concurrency message to the user. As for the null-fall-trhough. The null-falls-through behavior is intentional and consistent with how other client-level settings (PriorityLevel
, ReadConsistencyStrategy) work in the SDK — null on the request means "use the client default." We document this explicitly. A
sentinel/NullEmbeddingGenerator would add API surface complexity for an edge case; customers can easily provide a no-op implementation
if needed.

xinlian12 (dismissed):
LGTM, thanks

Microsoft.Azure.Cosmos/src/ICosmosEmbeddingGenerator.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <param name="cancellationToken">
+        /// A <see cref="CancellationToken"/> propagated from the originating
+        /// SDK call (for example <c>FeedIterator.ReadNextAsync</c>).
+        /// Implementations should honor cancellation.
+        /// </param>
+        /// <returns>
+        /// A task that resolves to a sequence of <see cref="float"/> embedding vectors
+        /// with the same cardinality and ordering as <paramref name="text"/>.
+        /// </returns>
+        Task<IEnumerable<ReadOnlyMemory<float>>> GenerateEmbeddingsAsync(
```
Pilchie: We support integer datatypes for vectors as well, right? Is there a way to use this feature with them?

Microsoft.Azure.Cosmos/src/RequestOptions/RequestOptions.cs
```diff
@@ -115,6 +115,11 @@ public class RequestOptions
         /// </summary>
         internal virtual ReadConsistencyStrategy? BaseReadConsistencyStrategy { get; set; }
 
+        /// <summary>
+        /// Gets or sets the embedding generator for the request.
+        /// </summary>
+        internal virtual ICosmosEmbeddingGenerator BaseEmbeddingGenerator { get; set; }
```
Pilchie: What's the reason for having this internal property instead of only having it on QueryRequestOptions?

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -405,6 +405,24 @@ public ConnectionMode ConnectionMode
...
+        /// If the gateway returns a plan that requires embedding generation and neither the request-level
+        /// nor this property is set, the SDK throws an exception describing how to configure a generator.
+        /// </remarks>
+        [JsonIgnore]
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        ICosmosEmbeddingGenerator EmbeddingGenerator { get; set; }
```
Pilchie, line 418: Should we expose a readonly version of this on `CosmosClient` itself? Is that planned later when we actually start consuming the `ICosmosEmbeddingGenerator`?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.net6.json
```diff
@@ -1,1825 +1,1892 @@
-{
```
Pilchie, line 1: What happened here?

Microsoft.Azure.Cosmos/src/ICosmosEmbeddingGenerator.cs
Pilchie: Great comment - note that `<remarks>` sections don't show in IDE tooltips, so customers may not see that though.

Microsoft.Azure.Cosmos/src/ICosmosEmbeddingGenerator.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <c>ContainerProperties.EmbeddingGenerator</c>. Implementations MUST be thread-safe and are
+    /// responsible for any caching, retries, and authentication required to call the underlying
+    /// embedding service.
+    /// </summary>
+#if PREVIEW
+    public
+#else
+    internal
+#endif
+    interface ICosmosEmbeddingGenerator
```
kushagraThapar, line 58: Could you please add a `<remarks>` block documenting the contract expectations for implementers? Specifically:

- **Lifecycle / disposal** — does the SDK own/dispose the generator if it implements `IDisposable`? (I'm guessing no, but worth being explicit.)
- **Error semantics** — what should impls throw on transient vs permanent failures, and does the SDK retry?
- **Cancellation** — should impls honor the `CancellationToken` cooperatively, or is best-effort fine?
- **Idempotency** — can the SDK call this multiple times for the same inputs (e.g., on retry)?

The downstream resolver/pipeline PRs (#5832–#5837) and any external implementers will benefit from a clear contract here — I am curious to learn how you're thinking about these so we keep impls aligned.

Microsoft.Azure.Cosmos/src/CosmosClient.cs
```diff
@@ -607,6 +607,16 @@ internal CosmosClient(
...
         public virtual CosmosClientOptions ClientOptions => this.ClientContext.ClientOptions;
 
+#if PREVIEW
+        /// <summary>
+        /// Gets the client-wide <see cref="ICosmosEmbeddingGenerator"/>, or <c>null</c> if none was set.
+        /// Set via <see cref="CosmosClientOptions.EmbeddingGenerator"/> or
+        /// <see cref="Fluent.CosmosClientBuilder.WithEmbeddingGenerator"/>. A per-container override
+        /// can be set via <see cref="ContainerProperties.EmbeddingGenerator"/>.
+        /// </summary>
+        public virtual ICosmosEmbeddingGenerator EmbeddingGenerator => this.ClientContext.ClientOptions.EmbeddingGenerator;
```
kushagraThapar, line 616: Could you please add a small direct test for `CosmosClient.EmbeddingGenerator`? The existing test reads `ClientOptions.EmbeddingGenerator`, which exercises the underlying storage but not the new accessor itself. A simple test that constructs a client with a generator and asserts `client.EmbeddingGenerator` returns the same instance would lock down the public surface.

Microsoft.Azure.Cosmos/src/CosmosEmbeddingResult.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// The result of a call to <see cref="ICosmosEmbeddingGenerator.GenerateEmbeddingsAsync"/>.
+    /// Carries the generated float32 vectors and an optional token-usage count for diagnostics
+    /// and billing purposes.
+    /// </summary>
+#if PREVIEW
+    public
+#else
+    internal
+#endif
+    sealed class CosmosEmbeddingResult
```
kushagraThapar, line 20: Could you please consider adding a few structured diagnostic fields to `CosmosEmbeddingResult` now (V0), before downstream PRs and external impls start shipping? Some candidates worth thinking about:

- `ModelName` (echo of the model the impl actually used — useful when the resolver has fallbacks)
- Latency / duration
- Cache hit/miss indicator (if the impl has a cache)
- Token usage is already there ✅

Adding fields after V0 ships is awkward — implementers will already be using the smaller constructor, and `CosmosDiagnostics` integration becomes harder. A `Properties`-style dictionary would also work if you'd rather keep the strongly-typed surface lean. I am curious to hear how you're planning to surface impl-side telemetry into `CosmosDiagnostics`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosQueryRequestOptionsUniTests.cs
```diff
@@ -4,9 +4,6 @@
 
 namespace Microsoft.Azure.Cosmos
 {
-    using System;
-    using System.Collections.Generic;
-    using System.Text;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
```
kushagraThapar, line 10: Could you please revert this file? The only change is removing 3 unused `using` statements, which looks like leftover cleanup from the design pivot away from `QueryRequestOptions.EmbeddingGenerator`. Unused-`using` cleanup is unrelated to the V0 surface and adds noise to the diff — would be cleaner to drop it from this PR (or split into a separate housekeeping PR if we also want to fix the pre-existing `UniTests` → `UnitTests` typo while we're there).

kushagraThapar (commented):
Hi @ananth7592, thanks for the iteration on this!

Two things I wanted to confirm with you before we land this:

1. **Per-request override** — In an earlier revision, this PR added `QueryRequestOptions.EmbeddingGenerator` for per-request scope, and the PR description still references that API. The HEAD diff has moved to client + container scope only, and I don't see a per-request layer anywhere. Could you please confirm that the per-request override is intentionally deferred to a follow-up (e.g., one of #5832–#5837), and not lost in the pivot? If so, a one-line callout in the PR description would help future readers understand the V0 scope.

2. **PR description vs HEAD** — Relatedly, the description still shows the old surface (`IEmbeddingGenerator` without the `Cosmos` prefix, `ReadOnlyMemory<double>`, property on `QueryRequestOptions`). Could you please refresh it to match the current HEAD before merge? Since the merge commit body is permanent in git history, it'd be great to have it reflect what actually shipped.

Microsoft.Azure.Cosmos/src/ICosmosEmbeddingGenerator.cs
ananth7592: Great questions:
Disposal-  No. The customer owns the generator instance and its lifecycle. The SDK never disposes it — even if it implements IDisposable. If the customer's generator wraps a HttpClient/EmbeddingClient, they should dispose it when they dispose their CosmosClient (or whenever their app's composition root tears down).

Error semantics - Only the provider will perform retries, any errors after that will be surefaced up to the cx as CosmosExcetion
Canalletion - In our own .AI provider, we will do it coopreateively , otherts i sbest effort 

Idempotency — can the SDK call this multiple times for the same inputs (e.g., on retry)? Yes,  this is stateless embeddings call


 1. Disposal — "Yes, typically, we clean up in the dispose, In the case of our own provider, yes"

  Issue: the reviewer's actual question is "does the SDK own/dispose the generator?" — i.e., they want to know if they need to keep their own reference alive or if the SDK takes ownership. Your answer talks about what your provider does on dispose, which is the opposite direction.

  Better: No. The customer owns the generator instance and its lifecycle. The SDK never disposes it — even if it implements IDisposable. If the customer's generator wraps a HttpClient/EmbeddingClient, they should dispose it when they dispose their CosmosClient (or whenever their app's composition root tears down).

  2. Error semantics  -  Implementations are responsible for their own retries (transient HTTP failures, rate-limit backoff, etc.) — the SDK does not retry the generator. Whatever the generator throws is wrapped by the SDK into a CosmosException and surfaced to the caller. So:

   - Transient — implementation should retry internally; surface as exception only after retries are exhausted.
   - Permanent (auth failure, invalid deployment, bad input) — throw immediately; SDK wraps and surfaces.

  3. Cancellation —  Implementations should honor the CancellationToken cooperatively wherever feasible (e.g., pass it through to the underlying HTTP call). Best-effort is acceptable, but ignoring it entirely is discouraged — long-running embedding calls without cancellation defeat caller-side timeouts.

  4. Idempotency —  Yes, the SDK may invoke the generator multiple times for the same inputs (e.g., on internal query retry). Implementations should be safe to call repeatedly — no per-call state assumptions. Note that each call typically incurs cost at the underlying embedding service; implementations may cache responses internally if they want to avoid duplicate billing.

Microsoft.Azure.Cosmos/src/ICosmosEmbeddingGenerator.cs
```diff
@@ -0,0 +1,95 @@
...
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Defines a contract for generating float32 vector embeddings from input text strings
+    /// supplied by the Azure Cosmos DB query pipeline.
+    /// The SDK invokes this when a query plan contains <c>GenerateEmbeddings(...)</c> literals
+    /// (for example <c>VectorDistance(GenerateEmbeddings("big brown cat"), c.embedding)</c>).
+    /// Set a client-wide default via <c>CosmosClientOptions.EmbeddingGenerator</c> or
+    /// <c>CosmosClientBuilder.WithEmbeddingGenerator</c>, and optionally override per-container via
+    /// <c>Database.GetContainer(id, embeddingGenerator)</c>. Implementations MUST be thread-safe and are
```
kushagraThapar: Could you please update this reference? `ContainerProperties.EmbeddingGenerator` doesn't actually exist on HEAD — the per-container override is provided via `Database.GetContainer(string, ICosmosEmbeddingGenerator)` (and `CosmosClient.GetContainer(db, c, generator)`). Suggest aligning with the wording already used on `CosmosClient.EmbeddingGenerator` (CosmosClient.cs L614-615), e.g.:

```suggestion
    /// <c>CosmosClientBuilder.WithEmbeddingGenerator</c>, and optionally override per-container via
    /// <see cref="Database.GetContainer(string, ICosmosEmbeddingGenerator)"/>. Implementations MUST be thread-safe and are
```

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -405,6 +405,22 @@ public ConnectionMode ConnectionMode
 #endif
         ReadConsistencyStrategy? ReadConsistencyStrategy { get; set; }
 
+        /// <summary>
+        /// Gets or sets the client-wide default <see cref="ICosmosEmbeddingGenerator"/> used to generate
+        /// query-time vector embeddings for hybrid and vector-search queries.
+        /// </summary>
+        /// <remarks>
+        /// A per-container override can be supplied via <c>Database.GetContainer(id, embeddingGenerator)</c>;
```
kushagraThapar: Same one — `ContainerProperties.EmbeddingGenerator` doesn't exist on HEAD. Could you please point this at the actual surface? Suggested wording (mirrors the xmldoc on `CosmosClient.EmbeddingGenerator`):

```suggestion
        /// A per-container override can be supplied via <see cref="Database.GetContainer(string, ICosmosEmbeddingGenerator)"/>;
```

Microsoft.Azure.Cosmos/src/Fluent/CosmosClientBuilder.cs
```diff
@@ -869,5 +869,27 @@ CosmosClientBuilder WithReadConsistencyStrategy(Cosmos.ReadConsistencyStrategy r
             this.clientOptions.ReadConsistencyStrategy = readConsistencyStrategy;
             return this;
         }
+
+        /// <summary>
+        /// Sets the client-wide default <see cref="ICosmosEmbeddingGenerator"/> used to generate
+        /// query-time vector embeddings for hybrid and vector-search queries.
+        /// </summary>
+        /// <remarks>
+        /// A per-container override can be supplied via <c>Database.GetContainer(id, embeddingGenerator)</c>;
```
kushagraThapar: Same fix here — `ContainerProperties.EmbeddingGenerator` doesn't exist on HEAD. Could you please align with the actual surface?

```suggestion
        /// A per-container override can be supplied via <see cref="Database.GetContainer(string, ICosmosEmbeddingGenerator)"/>;
```

kushagraThapar (dismissed):
Thanks @ananth7592 for iterating on this — looks great overall, approving!

One small documentation fix I'd like to land before merge: the latest commit's `<remarks>` on a few sites pivoted to talk about `ContainerProperties.EmbeddingGenerator`, but I couldn't find that property anywhere on the current HEAD. The actual per-container override surface is still `Database.GetContainer(string, ICosmosEmbeddingGenerator)` (and the equivalent `CosmosClient.GetContainer(db, c, generator)` overload). The xmldoc on `CosmosClient.EmbeddingGenerator` (CosmosClient.cs L614-615) already references the correct member, so we just need to align the other three sites with that same wording. Pinned them as inline comments below — they're tiny edits.

The rest looks good:
- `<remarks>` block now covers lifecycle/disposal, error semantics, cancellation, and concurrency — and concurrency is also called out in `<summary>`, which nicely picks up @Pilchie's IDE-tooltip note 👍
- `CosmosClient_EmbeddingGenerator_ReturnsConfiguredInstance` test covers the new accessor
- `CosmosEmbeddingResult` picked up `Latency` — thanks for that
- Unrelated `using`-cleanup in `CosmosQueryRequestOptionsUniTests.cs` is reverted
- PR description now matches HEAD
- Changelog entry added

Approving — please push the three docstring fixes before merging. Thank you!

Microsoft.Azure.Cosmos/src/Resource/Database/Database.cs
```diff
@@ -403,6 +403,44 @@ public abstract Task<ResponseMessage> DeleteStreamAsync(
...
+        /// </remarks>
+        /// <example>
+        /// <code language="c#">
+        /// <![CDATA[
+        /// Database db = this.cosmosClient.GetDatabase("myDatabaseId");
+        /// Container container = db.GetContainer("docs", myEmbeddingGenerator);
+        /// ]]>
+        /// </code>
+        /// </example>
+        public virtual Container GetContainer(string id, ICosmosEmbeddingGenerator embeddingGenerator)
```
FabianMeiswinkel: This sounds like a terrible idea - why do we need a per container concept - usually configs can have a default on client level - and overrides in RequestOptions - why is that not sufficent here?

FabianMeiswinkel (changes requested):
Why not follow established config (override) pattern - client-level + operation level

kushagraThapar (commented):
Thanks @ananth7592 for the productive iteration — the squash addressed a lot in one go:

- per-container layer removed (and the associated `ContainerProperties` / `Database.GetContainer(.., ICosmosEmbeddingGenerator)` / `SettingsContractTests` / `CosmosQueryRequestOptionsUniTests` churn went with it)
- comprehensive `<remarks>` block on `ICosmosEmbeddingGenerator` covering Lifecycle and disposal / Error semantics / Cancellation / Idempotency and concurrency
- rebased cleanly onto `da4efa8c8` (3.60.0)
- PR description rewritten to match HEAD (no more pointers at the old `IEmbeddingGenerator` / `double` / `QueryRequestOptions.EmbeddingGenerator` shape)
- two new tests landed (`CosmosClient_EmbeddingGenerator_ReturnsConfiguredInstance` covering the new accessor across builder-set + options-set paths, and `CosmosClientOptions_Clone_PreservesEmbeddingGenerator` covering direct-assignment + clone isolation)

That covers 8 of the 10 follow-ups I had after my prior pass. Three small things I'd love your take on before this lands — none are merge blockers from my side, and all three are easier to relax now than after preview ships.

---

**1. Could you please help me understand the V0 thinking on `EmbeddingSource` consolidation + `VectorDataType` widening on the interface signature?**

Two related observations on `Task<CosmosEmbeddingResult> GenerateEmbeddingsAsync(IReadOnlyList<string> texts, string endpoint, string deploymentName, int dimensions, CancellationToken cancellationToken = default)`:

- Adjacent merged PR #5848 already shipped `EmbeddingSource` on main with 5 fields (`SourcePaths`, `DeploymentName`, `ModelName`, `Endpoint`, `AuthType`). The new method takes 2 of those as flat strings (`endpoint`, `deploymentName`) and drops `AuthType` and `ModelName` on the floor — so implementations have no signal for how to authenticate, and no way to distinguish a deployment name from a model name when the underlying service cares about the difference. Future fields on `EmbeddingSource` would also become breaking-shape changes on the interface.

- On the data-type side, Cosmos supports four `VectorDataType` values (`Float32`, `Uint8`, `Int8`, `Float16`). The interface only surfaces `Float32` via `IReadOnlyList<ReadOnlyMemory<float>>` and the xmldoc explicitly says "float32" — a customer whose container is configured for `Uint8` (quantized vectors, cost optimization) silently cannot use this feature. This is also @Pilchie's still-unresolved P1 thread upthread.

Both of these are easier to relax in V0 than after preview ships — once shipped, broadening either dimension is a breaking change. A strongly-typed signature like

```csharp
Task<CosmosEmbeddingResult> GenerateEmbeddingsAsync(
    IReadOnlyList<string> texts,
    EmbeddingSource source,
    int dimensions,
    VectorDataType dataType,
```

would let the SDK pass the configured container's `EmbeddingSource` straight through (no field-by-field unpacking) and let implementations branch on `dataType` so all four `VectorDataType` values are reachable. The result type could stay `IReadOnlyList<ReadOnlyMemory<float>>` for now — or evolve to `ReadOnlyMemory<byte>`-per-vector once `dataType` selects the layout.

Could you please share whether this strongly-typed shape is on the table for V0, or whether there's a reason to keep the flat-string + float32-only contract for preview? If "float32-only V0" is the intentional answer, would it work to document that explicitly in `<remarks>` and link the tracking follow-up for the other three datatypes — that would let me close out the Pilchie thread cleanly.

---

**2. Could you please soften the forward-promising tense in the `ICosmosEmbeddingGenerator` `<summary>`?**

The first paragraph (`ICosmosEmbeddingGenerator.cs:14-15`) currently says:

> The SDK invokes this when a query plan contains `GenerateEmbeddings(...)` literals (for example `VectorDistance(GenerateEmbeddings("big brown cat"), c.embedding)`).

Today, though, no call site exists yet:

```
$ grep -rn 'GenerateEmbeddingsAsync\b' Microsoft.Azure.Cosmos/src/ --include='*.cs' | grep -v ICosmosEmbeddingGenerator.cs
(no output)
```

That's intentional — wiring is deferred to a follow-up — but the present-tense `<summary>` will make customers reading IDE tooltips believe the SDK calls this today. Could you please prepend a one-sentence note to the `<remarks>` so the IDE-tooltip story stays honest about V0 scope? Something like:

> Preview surface; the SDK call site is delivered in a follow-up release. Setting this generator has no effect today.

That keeps the existing forward-promising wording for the eventual GA story but flags the V0 gap clearly for anyone hovering on the type in 3.61.0-preview.0.

---

**3. Housekeeping: could you please leave a short reply on the prior inline threads?**

The force-push addressed (or made moot) 6 of the 7 inline comments I left earlier (the lifecycle/disposal `<remarks>` ask, the `CosmosClient.EmbeddingGenerator` direct-test ask, the `CosmosQueryRequestOptionsUniTests.cs` revert, and the three xmldoc-references-to-`ContainerProperties.EmbeddingGenerator` asks — all now addressed or moot since the per-container layer is gone). The threads on GitHub are still mechanically open, though, because the squash didn't post per-thread replies.

Could you please leave a one-line reply on each (or one summary comment listing fixed / moot status)? Saves the next reviewer pass and keeps the conversation tidy — purely housekeeping, not a merge blocker from my side.

---

To be explicit about scope: I am intentionally not re-raising @FabianMeiswinkel's CHANGES_REQUESTED ask about operation-level override here — that's a separate design conversation that belongs in a follow-up PR rather than this one. I am also setting aside the smaller minor / nit items from my prior pass (`[JsonIgnore]` regression-test guard, `ClientOptionJsonConverter` diagnostics surfacing, `#if PREVIEW` symmetry, `Microsoft.Extensions.AI` adapter sample, cross-SDK pioneering coordination, propose-design-doc-extension) — none of those need to land in this PR.

Thank you for the iteration!

ananth7592:
@kushagraThapar 
 Thanks — both are fair concerns. We chose this shape deliberately.

1. Could you please help me understand the V0 thinking on EmbeddingSource consolidation + VectorDataType widening on the interface signature?
  On EmbeddingSource consolidation. Two reasons we kept the flat (endpoint, deploymentName, dimensions) shape instead of passing EmbeddingSource:

   1. Auth is provider-owned, not SDK-owned. Customers bootstrap authentication in their ICosmosEmbeddingGenerator constructor — same pattern as OpenAIClient / EmbeddingClient / AzureOpenAIClient. Surfacing AuthType on the call is redundant signal since the credential decision has already been made at construction time, and EmbeddingSource.AuthType is a service-side enum, not a runtime credential.
   3. Interface mirrors the embedding SDKs customers already know. EmbeddingClient.GenerateEmbeddingsAsync(inputs, EmbeddingGenerationOptions { Dimensions }) — a customer can forward (texts, deploymentName, options { Dimensions = dimensions }) directly. ModelName is a service-side hint for the Cosmos write-path embedding generator, not consumed by client-side providers. SourcePaths and AuthType are similarly server-DTO concerns.

  If a future EmbeddingSource field genuinely needs to reach client-side providers, that's an additive parameter on the interface — a breaking change inside preview, which is what preview is for. We'd rather take that risk than leak the whole server DTO onto a client contract permanently.

  On VectorDataType widening. Float32-only by signature is intentional for V0:

   1. All embedding services produce float32. Azure OpenAI, OpenAI, Cohere, Voyage, Mistral — none expose a "give me int8" knob. Quantization is a Cosmos storage-side concern.
   2. The wire protocol is float32 regardless of stored type. VectorDistance takes float32 query vectors and the service quantizes at write time per VectorEmbeddingPolicy.DataType. Surfacing VectorDataType on the interface would imply the generator can produce non-float32, which the wire doesn't accept anyway.
   4. Customers pre-quantizing locally can still do so inside their generator and return the float32 representation.

  To close @Pilchie's thread cleanly, I'll add a <remarks> paragraph on GenerateEmbeddingsAsync:

   Query-time vectors are sent to the Azure Cosmos DB gateway as float32 regardless of the container's stored VectorDataType. Implementations targeting Uint8 / Int8 / Float16 storage should still produce float32 vectors; the service applies quantization at write time. This contract may be revisited if customer-side models that natively produce non-float32 vectors become common.

  Direct answer: the strongly-typed (EmbeddingSource, VectorDataType) shape is not on the table for V0 — flat float32 is deliberate, justified by industry convention, provider-owned auth, and the float32-on-the-wire protocol.

ananth7592:
2. Could you please soften the forward-promising tense in the ICosmosEmbeddingGenerator <summary>?
Done,

FabianMeiswinkel (approved):
LGTM

kushagraThapar (approved):
Thanks @ananth7592 for the thoughtful iteration — the new float32 wire-format `<para>` (lines 95-104) is a great architectural framing of the multi-datatype concern: clarifying that the SDK boundary stays float32 because the service applies quantization server-side at write time is cleaner than the "document V0 limitation" fallback I'd offered, and it cleanly covers all four `VectorDataType` storage configurations in one shot. The preview-surface `<para>` (lines 22-27) also reads exactly right for the IDE tooltip story.

Approving — looks great. Thank you!

Issue #5842, opened 2026-05-06
[Embedding V0] AzureOpenAIEmbeddingGenerator provider implementation


ananth7592:
# [Embedding V0] `AzureOpenAIEmbeddingGenerator` ΓÇö provider implementation

Part of: #5830 (V0 embedding generation).

## Background

This is the concrete `IEmbeddingGenerator` implementation that lives inside the `Microsoft.Azure.Cosmos.AI` package. It wraps `Azure.AI.OpenAI`'s `EmbeddingClient` to generate text embeddings, and is wired up from `VectorEmbeddingPolicy.Embedding.EmbeddingSource` (endpoint, deploymentName, authType) or from explicit constructor parameters.

Reference implementation exists in the CosmosDB monorepo at:
`/Product/Cosmos/EmbeddingGenerator/Service/Providers/AzureOpenAITextEmbeddingProvider.cs`
(adapt ΓÇö do not copy verbatim; the monorepo class has service-side concerns not relevant here).

## Public API

```csharp
namespace Microsoft.Azure.Cosmos.AI
{
    /// <summary>
    /// IEmbeddingGenerator implementation backed by Azure OpenAI.
    /// Construct from an EmbeddingSource (read from VectorEmbeddingPolicy) or
    /// provide endpoint/deployment/credential directly.
    /// </summary>
    public sealed class AzureOpenAIEmbeddingGenerator : IEmbeddingGenerator, IDisposable
    {
        // Constructor 1: from VectorEmbeddingPolicy EmbeddingSource + ApiKey
        public AzureOpenAIEmbeddingGenerator(EmbeddingSource source, string apiKey);

        // Constructor 2: from VectorEmbeddingPolicy EmbeddingSource + TokenCredential (Entra)
        public AzureOpenAIEmbeddingGenerator(EmbeddingSource source, TokenCredential credential);

        // Constructor 3: explicit params + ApiKey (for customers not using VectorEmbeddingPolicy)
        public AzureOpenAIEmbeddingGenerator(string endpoint, string deploymentName, int dimensions, string apiKey);

        // Constructor 4: explicit params + TokenCredential (Entra)
        public AzureOpenAIEmbeddingGenerator(string endpoint, string deploymentName, int dimensions, TokenCredential credential);

        public Task<IEnumerable<ReadOnlyMemory<double>>> GenerateEmbeddingsAsync(
            IEnumerable<string> text,
            CancellationToken cancellationToken = default);

[...]

Issue #5830, opened 2026-05-05
[Tracking] Embedding Generation in the .NET SDK Query Path (V0)


ananth7592:
# [Tracking] Embedding Generation in the .NET SDK (V0)

## Summary

Add first-class support for SDK-side embedding generation for hybrid / vector queries. When a user writes:

```sql
SELECT TOP 10 *
FROM c
ORDER BY RANK VectorDistance(c.text, 'big brown cat')
```

The gateway rewrites the query to:

```sql
SELECT TOP 10 *
FROM c
ORDER BY RANK VectorDistance(c.embedding, @documentdb-hybridsearchquery-embedding-0)
```

ΓÇªand returns an `embeddingParameterMap` (`@documentdb-hybridsearchquery-embedding-0` ΓåÆ `"big brown cat"`). The SDK calls the registered `IEmbeddingGenerator`, injects the returned vectors as `SqlParameter`s, and executes the rewritten query per-partition. Latency surfaces as a first-class diagnostics item.

### Interface (core SDK ΓÇö `Microsoft.Azure.Cosmos`)

```csharp
public interface IEmbeddingGenerator
{
    Task<IEnumerable<ReadOnlyMemory<double>>> GenerateEmbeddingsAsync(
        IEnumerable<string> text,
        CancellationToken cancellationToken = default);
}
```

Registered on `QueryRequestOptions.EmbeddingGenerator`.

### Provider package (`Microsoft.Azure.Cosmos.AI`)

Separate optional NuGet ΓÇö analogous to `Microsoft.Azure.Cosmos.Encryption`. Ships `AzureOpenAIEmbeddingGenerator`, the concrete Azure OpenAI / Foundry-backed implementation. Customers who use a different provider implement `IEmbeddingGenerator` directly. The core SDK has **no dependency** on this package.

### VectorEmbeddingPolicy extension

The `embeddingSource` block (new in V0) is added to each `vectorEmbeddings` entry in the container policy. It carries the endpoint, deployment name, auth type, and source paths ΓÇö the values used to construct an `AzureOpenAIEmbeddingGenerator`.

```json
"embeddingSource": {
  "sourcePaths": ["/title", "/abstract"],
  "deploymentName": "text-embedding-3-small",
  "modelName": "text-embedding-3-small",
  "endpoint": "https://embedding-south-central.cognitiveservices.azure.com/",
  "authType": "ApiKey"
}
```

### Flow

[...]

PR #5845, opened 2026-05-06
Embedding: Adds Microsoft.Azure.Cosmos.AI package scaffold

### LINKED ISSUE #5841: [Embedding V0] New Microsoft.Azure.Cosmos.Embedding package — scaffold
# [Embedding V0] New `Microsoft.Azure.Cosmos.AI` package ΓÇö scaffold

Part of: #5830 (V0 embedding generation).

## Background

The concrete embedding provider implementation (Azure OpenAI / Foundry) ships as a **separate optional package**, analogous to `Microsoft.Azure.Cosmos.Encryption`. This keeps the core SDK free of Azure AI SDK dependencies for the vast majority of customers who are not using vector search.

- Core SDK: `Microsoft.Azure.Cosmos` ΓÇö defines `IEmbeddingGenerator` interface only.
- Provider package: `Microsoft.Azure.Cosmos.AI` ΓÇö ships `AzureOpenAIEmbeddingGenerator` (and any future providers).

## Scope

Scaffold the new project structure (no business logic yet ΓÇö that is in the next sub-issue):

```
Microsoft.Azure.Cosmos.AI/
  Directory.Build.props          # versioning (CosmosAIOfficialVersion, CosmosAIPreviewVersion)
  src/
    Microsoft.Azure.Cosmos.AI.csproj
    AssemblyInfo.cs
  tests/
    Microsoft.Azure.Cosmos.AI.Tests/
      Microsoft.Azure.Cosmos.AI.Tests.csproj
```

### csproj requirements

- **Target frameworks**: `net8.0;netstandard2.0` (mirror what core SDK targets for broad compatibility).
- **PackageId**: `Microsoft.Azure.Cosmos.AI`
- **NuGet dependencies**:
  - `Microsoft.Azure.Cosmos` (project reference when `SdkProjectRef=True`, otherwise NuGet pin)
  - `Azure.AI.OpenAI` (the Azure OpenAI client SDK)
  - `Azure.Core` (for `TokenCredential`)
- **Strong-name signing**: use the same `35MSSharedLib1024.snk` as the core SDK.
- Follow the exact same `SdkProjectRef` / `IsPreview` conditional-reference pattern as `Microsoft.Azure.Cosmos.Encryption.csproj`.

### Solution file

Add the new `src` and `tests` projects to `Microsoft.Azure.Cosmos.sln`.

### Directory.Build.props

Add version properties alongside Encryption's entries:
```xml
<CosmosAIOfficialVersion>1.0.0</CosmosAIOfficialVersion>
<CosmosAIPreviewVersion>1.0.0</CosmosAIPreviewVersion>
<CosmosAIPreviewSuffixVersion>preview.0</CosmosAIPreviewSuffixVersion>
```

## Acceptance criteria

- `dotnet build Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj` succeeds with zero warnings.
- Test project (`dotnet test`) runs zero tests (empty) but exits cleanly.
- `SdkProjectRef=True` project reference path works.
- Package metadata (description, tags, license, icon) is correct.

## Files likely touched

- `Microsoft.Azure.Cosmos.AI/` (new folder tree)
- `Microsoft.Azure.Cosmos.sln` (new project entries)
- `Directory.Build.props` (version properties)

## Dependencies

- #5831 ΓÇö needs `IEmbeddingGenerator` to be defined in the core SDK.

ananth7592:
## Summary

Scaffolds the new `Microsoft.Azure.Cosmos.AI` package — an optional, provider-style extension library for the Azure Cosmos DB .NET SDK that delivers AI-powered capabilities to applications built on Cosmos DB.

  This PR is infrastructure only — no business logic. The first capability implementation follows in #5842.

---

## Motivation

  AI is becoming a first-class requirement for modern data applications. Developers increasingly want their database layer to participate in AI workflows — powering semantic search, retrieval-augmented generation (RAG), and intelligent query augmentation — without having to build and maintain the plumbing themselves.

  Microsoft.Azure.Cosmos.AI is the home for these capabilities in the .NET SDK. It is a dedicated, optional package that extends the core Azure Cosmos DB SDK with AI-powered features, keeping them modular and independently evolvable.

  Why a separate package?

   - Zero cost for non-AI workloads. Applications that don't use AI features pull in no AI dependencies. Azure.AI.OpenAI and related packages are only present for customers who explicitly opt in.
   - Independent versioning. The AI package ships on its own cadence, allowing rapid iteration on AI capabilities without coupling to the core SDK release cycle.
   - Established extensibility model. This follows the exact pattern already proven by Microsoft.Azure.Cosmos.Encryption — a separate, optional, provider-style package that extends the core SDK without polluting it.
   - Clear separation of concerns. The core SDK owns the data plane. The AI package owns AI service integration. Each can evolve, be tested, and be reviewed independently.

  Over time, Microsoft.Azure.Cosmos.AI will be the single, consistent destination for AI-powered Cosmos DB capabilities in .NET — whether that is automatic vector embedding generation, semantic ranking, AI-assisted query construction, or future capabilities as the AI landscape
  evolves.

---

## Changes

**New project scaffold**
- `Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj` - `netstandard2.0`, SNK signing, `SdkProjectRef` pattern, `Azure.AI.OpenAI 2.1.0`, explicit `Newtonsoft.Json 10.0.2`
- `Microsoft.Azure.Cosmos.AI/src/AssemblyInfo.cs` - `InternalsVisibleTo` for test project with strong-name public key
- `Microsoft.Azure.Cosmos.AI/tests/.../Microsoft.Azure.Cosmos.AI.Tests.csproj` - Empty MSTest project wired to main project
- `Microsoft.Azure.Cosmos.AI/Directory.Build.props` - Delegates up to repo root props

**Versioning**
- `Directory.Build.props` - Added `CosmosAIOfficialVersion`, `CosmosAIPreviewVersion`, `CosmosAIPreviewSuffixVersion`
- `Microsoft.Azure.Cosmos.sln` - Added `AI` solution folder with both projects

**Publishing pipelines**
- `azure-pipelines-cosmosai-official.yml` - Top-level release pipeline with Release Gates stage and Build/Pack/Publish stage
- `templates/static-tools-cosmosai.yml` - Static analysis with three build variants (NuGet ref, SdkProjectRef, parity check) plus BinSkim, CredScan, PoliCheck, AntiMalware, Component Governance
- `templates/cosmosai-nuget-pack.yml` - Build, `.nupkg`, `.snupkg` symbols, SBOM manifest, Azure blob drop, artifact publish

---

## Design Decisions

**Why `netstandard2.0`?**
Matches `Microsoft.Azure.Cosmos` and `Microsoft.Azure.Cosmos.Encryption` - broadest framework reach without sacrificing modern API access.

**Why `Newtonsoft.Json` with `PrivateAssets="All"`?**
The Cosmos SDK enforces an explicit `Newtonsoft.Json >= 10.0.2` reference on all dependents via an MSBuild check. `PrivateAssets="All"` satisfies the check at build time while keeping the package off the public `.nuspec` dependency graph for consumers - identical to the Encryption package.

**Why pipelines in this PR?**
Infrastructure is reviewable independently of business logic. The YAML files can merge and be registered in Azure DevOps now; the pipeline should be activated (queued for release) only after #5842 lands and provides real content to gate on.

---

## Verification

- `dotnet build` (official + SdkProjectRef variants) -> **0 warnings, 0 errors**
- `dotnet test` -> **No tests yet (empty), exits 0**

---

## Related

- #5830 - Epic: AI-powered querying for Cosmos DB .NET SDK
- #5838 - Core SDK: `ICosmosEmbeddingGenerator` interface + `QueryRequestOptions`/`CosmosClientOptions` surface
- #5842 - Implementation: `AzureOpenAIEmbeddingGenerator` provider (follows this PR)

Closes #5841
Part of #5830

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -9,11 +9,14 @@
 		<EncryptionOfficialVersion>2.0.5</EncryptionOfficialVersion>
 		<EncryptionPreviewVersion>2.1.0</EncryptionPreviewVersion>
 		<EncryptionPreviewSuffixVersion>preview5</EncryptionPreviewSuffixVersion>
+		<CosmosAIOfficialVersion>1.0.0</CosmosAIOfficialVersion>
+		<CosmosAIPreviewVersion>1.0.0</CosmosAIPreviewVersion>
+		<CosmosAIPreviewSuffixVersion>preview.0</CosmosAIPreviewSuffixVersion>
 		<CustomEncryptionVersion>1.0.0-preview09</CustomEncryptionVersion>
 		<HybridRowVersion>1.1.0-preview3</HybridRowVersion>
 		<LangVersion>10.0</LangVersion>
 		<AboveDirBuildProps>$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))</AboveDirBuildProps>
-		<DefineConstants Condition=" '$(IsNightly)' == 'true' or '$(IsPreview)' == 'true' ">$(DefineConstants);PREVIEW;ENCRYPTIONPREVIEW</DefineConstants>
+		<DefineConstants Condition=" '$(IsNightly)' == 'true' or '$(IsPreview)' == 'true' ">$(DefineConstants);PREVIEW;ENCRYPTIONPREVIEW;COSMOSAIPREVIEW</DefineConstants>
 	</PropertyGroup>
 	<Import Project="$(AboveDirBuildProps)" Condition=" '$(AboveDirBuildProps)' != '' " />
 </Project>
--- a/Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
+++ b/Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
@@ -14,11 +14,13 @@
     <Authors>Microsoft</Authors>
     <Description>AI extensions library for Azure Cosmos DB for NoSQL. Provides an Azure OpenAI / Foundry-backed embedding provider for automatic vector embedding generation in hybrid and vector search queries. Designed to be extensible for broader AI scenarios including semantic re-ranking, OCR, and model integrations. For more information, refer to https://aka.ms/CosmosDB</Description>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
-    <Title>Microsoft Azure Cosmos DB AI extensions library</Title>
+    <Title>Microsoft Azure Cosmos DB AI extension library</Title>
     <PackageId>Microsoft.Azure.Cosmos.AI</PackageId>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/Azure/azure-cosmos-dotnet-v3</PackageProjectUrl>
+    <PackageReleaseNotes>The change log for this SDK is made available at https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.AI/changelog.md at the time of release.</PackageReleaseNotes>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
@@ -27,6 +29,7 @@
 
   <ItemGroup>
     <AdditionalFiles Include="..\..\Microsoft.Azure.Cosmos\src\stylecop.json" Link="stylecop.json" />
+    <None Include="..\README.md" Pack="true" PackagePath="\" />
   </ItemGroup>
 
   <!-- Core SDK reference: project ref when building from source, otherwise NuGet pin -->
--- a/Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
+++ b/Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
@@ -1,0 +1,80 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AssemblyName>Microsoft.Azure.Cosmos.AI</AssemblyName>
+    <RootNamespace>Microsoft.Azure.Cosmos.AI</RootNamespace>
+    <LangVersion>$(LangVersion)</LangVersion>
+    <CosmosAIVersion Condition=" '$(IsPreview)' != 'true' ">$(CosmosAIOfficialVersion)</CosmosAIVersion>
+    <CosmosAIVersion Condition=" '$(IsPreview)' == 'true' ">$(CosmosAIPreviewVersion)</CosmosAIVersion>
+    <CosmosAIVersionSuffix Condition=" '$(IsPreview)' == 'true' ">$(CosmosAIPreviewSuffixVersion)</CosmosAIVersionSuffix>
+    <Version Condition=" '$(CosmosAIVersionSuffix)' == '' ">$(CosmosAIVersion)</Version>
+    <Version Condition=" '$(CosmosAIVersionSuffix)' != '' ">$(CosmosAIVersion)-$(CosmosAIVersionSuffix)</Version>
+    <CurrentDate>$([System.DateTime]::Now.ToString(yyyyMMdd))</CurrentDate>
+    <Company>Microsoft Corporation</Company>
+    <Authors>Microsoft</Authors>
+    <Description>AI extensions library for Azure Cosmos DB for NoSQL. Provides an Azure OpenAI / Foundry-backed embedding provider for automatic vector embedding generation in hybrid and vector search queries. Designed to be extensible for broader AI scenarios including semantic re-ranking, OCR, and model integrations. For more information, refer to https://aka.ms/CosmosDB</Description>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <Title>Microsoft Azure Cosmos DB AI extensions library</Title>
+    <PackageId>Microsoft.Azure.Cosmos.AI</PackageId>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageProjectUrl>https://github.com/Azure/azure-cosmos-dotnet-v3</PackageProjectUrl>
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <PackageTags>microsoft;azure;cosmos;cosmosdb;documentdb;nosql;azureofficial;dotnetcore;netcore;netstandard;ai;embedding;vector;openai;semanticranking</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="..\..\Microsoft.Azure.Cosmos\src\stylecop.json" Link="stylecop.json" />
+  </ItemGroup>
+
+  <!-- Core SDK reference: project ref when building from source, otherwise NuGet pin -->
+  <ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' != 'True' ">
+    <PackageReference Include="Microsoft.Azure.Cosmos" Version="[3.59.0,)" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' == 'True' ">
+    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.60.0-preview.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(SdkProjectRef)' == 'True' ">
+    <ProjectReference Include="..\..\Microsoft.Azure.Cosmos\src\Microsoft.Azure.Cosmos.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="..\..\Icon.png" Pack="true" PackagePath=""/>
+  </ItemGroup>
+
+  <!-- Required by Microsoft.Azure.Cosmos SDK; PrivateAssets=All keeps it off the public dependency graph -->
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.2" NoWarn="NU1903" PrivateAssets="All" />
+  </ItemGroup>
+
+  <!-- Azure OpenAI provider dependency -->
+  <ItemGroup>
+    <PackageReference Include="Azure.AI.OpenAI" Version="2.1.0" />
+  </ItemGroup>
+
+  <!-- Analyzers -->
+  <ItemGroup>
+    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="16.0.102" PrivateAssets="All" />
+  </ItemGroup>
+
+  <!-- Strong-name signing -->
+  <PropertyGroup>
+    <SigningType>Product</SigningType>
+    <SignAssembly>true</SignAssembly>
+    <DelaySign>true</DelaySign>
+    <AssemblyOriginatorKeyFile>..\..\35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(SdkProjectRef)' == 'True' ">
+    <DefineSdkProjectRefSymbol Condition=" '$(DefineSdkProjectRefSymbol)' == '' ">true</DefineSdkProjectRefSymbol>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(DefineSdkProjectRefSymbol)' == 'true' ">
+    <DefineConstants>$(DefineConstants);SDKPROJECTREF</DefineConstants>
+  </PropertyGroup>
+</Project>
--- a/Microsoft.Azure.Cosmos.AI/src/Mirrored/AssemblyKeys.cs
+++ b/Microsoft.Azure.Cosmos.AI/src/Mirrored/AssemblyKeys.cs
@@ -1,0 +1,12 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.AI
+{
+    internal static class AssemblyKeys
+    {
+        /// <summary>TestPublicKey is an unsupported strong key for testing and internal use only.</summary>
+        internal const string TestPublicKey = ", PublicKey=0024000004800000940000000602000000240000525341310004000001000100197c25d0a04f73cb271e8181dba1c0c713df8deebb25864541a66670500f34896d280484b45fe1ff6c29f2ee7aa175d8bcbd0c83cc23901a894a86996030f6292ce6eda6e6f3e6c74b3c5a3ded4903c951e6747e6102969503360f7781bf8bf015058eb89b7621798ccc85aaca036ff1bc1556bb7f62de15908484886aa8bbae";
+    }
+}
--- a/Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Contracts/ContractEnforcementTests.cs
+++ b/Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Contracts/ContractEnforcementTests.cs
@@ -1,0 +1,32 @@
+namespace Microsoft.Azure.Cosmos.AI.Tests
+{
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [TestCategory("Windows")]
+    [TestCategory("UpdateContract")]
+    [TestClass]
+    public class ContractEnforcementTests
+    {
+        /// <summary>
+        /// Locks the public API surface of <c>Microsoft.Azure.Cosmos.AI</c> against a
+        /// versioned baseline so accidental renames, visibility changes, or removals
+        /// surface at PR review time instead of shipping to a preview NuGet.
+        ///
+        /// IMPORTANT: Because tests can run on multiple .NET versions, the contract
+        /// validation uses framework-specific baselines to ensure consistency:
+        ///
+        /// - When running on net8.0: validates against DotNetSDKCosmosAIAPI.net8.json
+        ///
+        /// To update baselines, run: UpdateContracts.ps1 from the repository root.
+        /// </summary>
+        [TestMethod]
+        public void ContractChanges()
+        {
+            Cosmos.Tests.Contracts.ContractEnforcement.ValidateContract(
+                dllName: "Microsoft.Azure.Cosmos.AI",
+                contractType: Cosmos.Tests.Contracts.ContractType.Standard,
+                baselinePattern: "DotNetSDKCosmosAIAPI",
+                breakingChangesPattern: "DotNetSDKCosmosAIAPIChanges");
+        }
+    }
+}
--- a/Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Contracts/DotNetSDKCosmosAIAPI.net8.json
+++ b/Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Contracts/DotNetSDKCosmosAIAPI.net8.json
@@ -1,0 +1,44 @@
+{
+  "Subclasses": {},
+  "Members": {
+    "Boolean Equals(System.Object)": {
+      "Type": "Method",
+      "Attributes": [],
+      "MethodInfo": "Boolean Equals(System.Object);IsAbstract:False;IsStatic:False;IsVirtual:True;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+    },
+    "Int32 GetHashCode()": {
+      "Type": "Method",
+      "Attributes": [],
+      "MethodInfo": "Int32 GetHashCode();IsAbstract:False;IsStatic:False;IsVirtual:True;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+    },
+    "System.Boolean System.Object.Equals(System.Object, System.Object)": {
+      "Type": "Method",
+      "Attributes": [],
+      "MethodInfo": "System.Boolean System.Object.Equals(System.Object, System.Object);IsAbstract:False;IsStatic:True;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+    },
+    "System.Boolean System.Object.ReferenceEquals(System.Object, System.Object)": {
+      "Type": "Method",
+      "Attributes": [],
+      "MethodInfo": "System.Boolean System.Object.ReferenceEquals(System.Object, System.Object);IsAbstract:False;IsStatic:True;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+    },
+    "System.String ToString()": {
+      "Type": "Method",
+      "Attributes": [],
+      "MethodInfo": "System.String ToString();IsAbstract:False;IsStatic:False;IsVirtual:True;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+    },
+    "System.Type GetType()[System.Runtime.CompilerServices.NullableContextAttribute((Byte)1)]-[System.Runtime.CompilerServices.IntrinsicAttribute()]": {
+      "Type": "Method",
+      "Attributes": [
+        "IntrinsicAttribute",
+        "NullableContextAttribute"
+      ],
+      "MethodInfo": "System.Type GetType();IsAbstract:False;IsStatic:False;IsVirtual:False;IsGenericMethod:False;IsConstructor:False;IsFinal:False;"
+    },
+    "Void .ctor()": {
+      "Type": "Constructor",
+      "Attributes": [],
+      "MethodInfo": "Void .ctor()"
+    }
+  },
+  "NestedTypes": {}
+}
--- a/Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
+++ b/Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
@@ -1,0 +1,38 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <IsTestProject>true</IsTestProject>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <Platform>AnyCPU</Platform>
+    <TargetFramework>net8.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <RootNamespace>Microsoft.Azure.Cosmos.AI.Tests</RootNamespace>
+    <LangVersion>$(LangVersion)</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
+    <PackageReference Include="Moq" Version="4.8.3" />
+    <PackageReference Include="MSTest.TestAdapter" Version="1.2.0" />
+    <PackageReference Include="MSTest.TestFramework" Version="1.2.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageReference Include="coverlet.collector" Version="3.1.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Azure.Cosmos.AI.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <SignAssembly>true</SignAssembly>
+    <DelaySign>true</DelaySign>
+    <AssemblyOriginatorKeyFile>..\..\..\testkey.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <PlatformTarget>x64</PlatformTarget>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+</Project>
--- a/Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
+++ b/Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
@@ -12,16 +12,27 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
+    <Compile Include="..\..\..\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.Tests\Contracts\ContractEnforcement.cs" Link="Contracts\ContractEnforcement.cs" />
+    <Compile Include="..\..\..\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.Tests\Contracts\ContractType.cs" Link="Contracts\ContractType.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
     <PackageReference Include="Moq" Version="4.8.3" />
-    <PackageReference Include="MSTest.TestAdapter" Version="1.2.0" />
-    <PackageReference Include="MSTest.TestFramework" Version="1.2.0" />
+    <PackageReference Include="MSTest.TestAdapter" Version="3.7.3" />
+    <PackageReference Include="MSTest.TestFramework" Version="3.7.3" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageReference Include="coverlet.collector" Version="3.1.2" />
+    <PackageReference Include="coverlet.collector" Version="6.0.2" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.Azure.Cosmos.AI.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="Contracts\DotNetSDKCosmosAIAPI*.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
   </ItemGroup>
 
   <PropertyGroup>
--- a/Microsoft.Azure.Cosmos.sln
+++ b/Microsoft.Azure.Cosmos.sln
@@ -39,6 +39,12 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FaultInjectionTests", "Micr
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests", "Microsoft.Azure.Cosmos.Encryption.Custom\tests\Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests\Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests.csproj", "{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "AI", "AI", "{701B153E-380B-4452-9120-3975C59DC296}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Cosmos.AI", "Microsoft.Azure.Cosmos.AI\src\Microsoft.Azure.Cosmos.AI.csproj", "{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Cosmos.AI.Tests", "Microsoft.Azure.Cosmos.AI\tests\Microsoft.Azure.Cosmos.AI.Tests\Microsoft.Azure.Cosmos.AI.Tests.csproj", "{80F63801-FFB4-4D8E-9466-A44C1F0C4946}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Cover|Any CPU = Cover|Any CPU
@@ -193,6 +199,30 @@ Global
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Release|Any CPU.Build.0 = Release|Any CPU
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Release|x64.ActiveCfg = Release|Any CPU
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}.Release|x64.Build.0 = Release|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Cover|Any CPU.ActiveCfg = Debug|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Cover|Any CPU.Build.0 = Debug|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Cover|x64.ActiveCfg = Debug|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Cover|x64.Build.0 = Debug|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Debug|x64.Build.0 = Debug|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Release|Any CPU.Build.0 = Release|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Release|x64.ActiveCfg = Release|Any CPU
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343}.Release|x64.Build.0 = Release|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Cover|Any CPU.ActiveCfg = Debug|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Cover|Any CPU.Build.0 = Debug|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Cover|x64.ActiveCfg = Debug|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Cover|x64.Build.0 = Debug|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Debug|x64.Build.0 = Debug|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Release|Any CPU.Build.0 = Release|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Release|x64.ActiveCfg = Release|Any CPU
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946}.Release|x64.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -205,6 +235,8 @@ Global
 		{F87719DB-BB52-4B12-9D9C-F6AE30BAB3D7} = {51F858D8-707E-4F21-BCC6-4D6123832E4F}
 		{B5B3631D-AC2F-4257-855D-D6FE12F20B60} = {51F858D8-707E-4F21-BCC6-4D6123832E4F}
 		{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC} = {51F858D8-707E-4F21-BCC6-4D6123832E4F}
+		{60C02F10-7DE4-4BF0-BA28-732ABAF0A343} = {701B153E-380B-4452-9120-3975C59DC296}
+		{80F63801-FFB4-4D8E-9466-A44C1F0C4946} = {701B153E-380B-4452-9120-3975C59DC296}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {C6A1D820-CB03-4DE6-87D1-46EF476F0040}
--- a/azure-pipelines-cosmosai.yml
+++ b/azure-pipelines-cosmosai.yml
@@ -1,0 +1,33 @@
+# CI pipeline for Microsoft.Azure.Cosmos.AI — runs on every PR touching the AI package.
+# Validates the three-build parity strategy (NuGet ref, SdkProjectRef, parity check)
+# before changes reach main, complementing the manual release-gate in
+# azure-pipelines-cosmosai-official.yml.
+trigger: none
+
+pr:
+  branches:
+    include:
+    - main
+    - releases/*
+  paths:
+    include:
+    - 'Microsoft.Azure.Cosmos.AI/**'
+    - 'templates/cosmosai-nuget-pack.yml'
+    - 'templates/static-tools-cosmosai.yml'
+    - 'azure-pipelines-cosmosai.yml'
+    - 'azure-pipelines-cosmosai-official.yml'
+    - 'Directory.Build.props'
+
+variables:
+  BuildConfiguration: Release
+
+jobs:
+- template: templates/static-tools-cosmosai.yml
+  parameters:
+    BuildConfiguration: '${{ variables.BuildConfiguration }}'
+
+- template: templates/cosmosai-nuget-pack.yml
+  parameters:
+    BuildConfiguration: '${{ variables.BuildConfiguration }}'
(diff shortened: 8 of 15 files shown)


NaluTripician:
@sdkReviewAgent

templates/cosmosai-nuget-pack.yml
```diff
@@ -0,0 +1,76 @@
...
+        inputs:
+          SourcePath: '$(Build.ArtifactStagingDirectory)/bin/AnyCPU/$(BuildConfiguration)/Microsoft.Azure.Cosmos.AI/**'
+          azureSubscription: azuresdkpartnerdrops
+          Destination: AzureBlob
+          storage: azuresdkpartnerdrops
+          ContainerName: 'drops'
+          BlobPrefix: 'cosmosdb/csharp/cosmosai/$(BlobVersion)'
+          CleanTargetBeforeCopy: ${{ parameters.CleanupFolder }}
+
+    - task: PublishBuildArtifacts@1
```
xinlian12: 🟡 **Pipeline: `PublishBuildArtifacts` nested inside release-only conditional — CI builds skip artifact publishing**

```yaml
  - ${{ if eq(parameters.ReleasePackage, true) }}:
    ...
    - task: PublishBuildArtifacts@1          # ← 4-space indent = inside conditional

```

In the sibling `encryption-nuget-pack.yml` (line 76), `PublishBuildArtifacts@1` is at the **step root level** (2-space indent), meaning it runs unconditionally for both CI and release builds:

```yaml
  - task: PublishBuildArtifacts@1            # ← 2-space indent = outside conditional

```

**Why this matters:** The CI pipeline (`azure-pipelines-cosmosai.yml`) calls this template with `ReleasePackage: false`. Because the publish step is inside the conditional, PR/CI builds will silently skip artifact publishing. Reviewers and testers won't be able to download the `.nupkg` from Azure DevOps to inspect the package output — only the release pipeline produces downloadable artifacts. This appears to be a copy-paste indentation error from mirroring the encryption template.

**Suggested fix:** Outdent `PublishBuildArtifacts@1` to the `steps` level (2-space indent) to match the encryption template:

```yaml

    # end of ${{ if eq(parameters.ReleasePackage, true) }} block

  - task: PublishBuildArtifacts@1
    displayName: 'Publish Artifacts: Microsoft.Azure.Cosmos.AI'
    inputs:
      artifactName: Microsoft.Azure.Cosmos.AI

```

azure-pipelines-cosmosai.yml
```diff
@@ -0,0 +1,32 @@
...
+# azure-pipelines-cosmosai-official.yml.
+trigger: none
+
+pr:
+  branches:
+    include:
+    - main
+    - releases/*
+  paths:
+    include:
```
xinlian12, line 13: 🟡 **Pipeline: `Directory.Build.props` missing from CI trigger paths**

```yaml
paths:
  include:
  - 'Microsoft.Azure.Cosmos.AI/**'
  - 'templates/cosmosai-nuget-pack.yml'
  - 'templates/static-tools-cosmosai.yml'
  - 'azure-pipelines-cosmosai.yml'
  - 'azure-pipelines-cosmosai-official.yml'

  # Missing: 'Directory.Build.props'

```

**Why this matters:** The root `Directory.Build.props` contains `CosmosAIOfficialVersion`, `CosmosAIPreviewVersion`, and `CosmosAIPreviewSuffixVersion`. If someone bumps the AI package version in `Directory.Build.props` without touching any file under `Microsoft.Azure.Cosmos.AI/`, this CI pipeline won't trigger. The version change lands in `main` without running the AI-specific three-build parity validation.

The main `azure-pipelines.yml` does trigger on `Directory.Build.props` changes, but it doesn't run the AI-specific static analysis or pack steps.

**Suggested fix:** Add `'Directory.Build.props'` to the path includes.

Directory.Build.props
```diff
@@ -9,6 +9,9 @@
 		<EncryptionOfficialVersion>2.0.5</EncryptionOfficialVersion>
 		<EncryptionPreviewVersion>2.1.0</EncryptionPreviewVersion>
 		<EncryptionPreviewSuffixVersion>preview5</EncryptionPreviewSuffixVersion>
+		<CosmosAIOfficialVersion>1.0.0</CosmosAIOfficialVersion>
+		<CosmosAIPreviewVersion>1.0.0</CosmosAIPreviewVersion>
+		<CosmosAIPreviewSuffixVersion>preview.0</CosmosAIPreviewSuffixVersion>
```
xinlian12, line 14: 🟢 **Versioning: Consider adding `COSMOSAIPREVIEW` define constant**

The existing `DefineConstants` line adds `PREVIEW` and `ENCRYPTIONPREVIEW` when `IsPreview` or `IsNightly` is set, but this PR does not add a `COSMOSAIPREVIEW` equivalent:

```xml
<DefineConstants Condition=" '$(IsNightly)' == 'true' or '$(IsPreview)' == 'true' ">
    $(DefineConstants);PREVIEW;ENCRYPTIONPREVIEW
</DefineConstants>

```

**Why this matters:** Zero impact today (no `#if` blocks exist in the scaffold), but the first time someone adds preview-gated AI code (e.g., `#if COSMOSAIPREVIEW`), they'll discover the symbol doesn't exist. Adding it now while the versioning pattern is being established is one fewer thing to retrofit later.

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
+    <PackageReference Include="Moq" Version="4.8.3" />
+    <PackageReference Include="MSTest.TestAdapter" Version="1.2.0" />
+    <PackageReference Include="MSTest.TestFramework" Version="1.2.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageReference Include="coverlet.collector" Version="3.1.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Azure.Cosmos.AI.csproj" />
```
xinlian12, line 29: 💬 **Testing: No API contract enforcement test infrastructure (track for follow-up)**

The Encryption test projects include `ContractEnforcement.cs` linked compilation and baseline JSON files (`DotNetSDKEncryptionAPI*.json`) that automatically detect public API surface changes and fail CI if the baseline isn't updated. The AI test project has none of this:

```xml
<!-- Encryption.Tests has these, AI.Tests does not: -->
<Compile Include="...\ContractEnforcement.cs" Link="Contracts\ContractEnforcement.cs" />
<Compile Include="...\ContractType.cs" Link="Contracts\ContractType.cs" />
<None Include="Contracts\DotNetSDKEncryptionAPI*.json">
    <CopyToOutputDirectory>Always</CopyToOutputDirectory>
</None>

```

This is completely expected for a scaffold with no public APIs. However, it should be added **before PR #5842 merges** to prevent accidental API surface drift once real public types land. Per the repo's `.pragentrules`, public/preview API changes must have corresponding contract baseline updates.

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
```diff
@@ -0,0 +1,80 @@
...
+    <AdditionalFiles Include="..\..\Microsoft.Azure.Cosmos\src\stylecop.json" Link="stylecop.json" />
+  </ItemGroup>
+
+  <!-- Core SDK reference: project ref when building from source, otherwise NuGet pin -->
+  <ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' != 'True' ">
+    <PackageReference Include="Microsoft.Azure.Cosmos" Version="[3.59.0,)" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(SdkProjectRef)' != 'True' AND '$(IsPreview)' == 'True' ">
+    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.60.0-preview.0" />
```
xinlian12, line 41: 🟡 **Pipeline: Main CI `build-preview.yml` will NuGet-restore `3.60.0-preview.0` for this project**

```xml
<PackageReference Include="Microsoft.Azure.Cosmos" Version="3.60.0-preview.0" />

```

Adding the AI project to `Microsoft.Azure.Cosmos.sln` means it participates in **all** solution-level builds, including `build-preview.yml` Job 1 ("Preview Flag"), which builds the entire solution with `-p:IsPreview=true` but **without** `-p:SdkProjectRef=true`. This triggers the NuGet reference to `3.60.0-preview.0` rather than the `ProjectReference`.

**Why this matters:** If `3.60.0-preview.0` is not published on the NuGet feed configured in `NuGet.config`, the `build-preview.yml` Job 1 restore will fail. This breaks the main CI pipeline (`azure-pipelines.yml`) for **all** PRs — not just AI-related ones — since `azure-pipelines.yml` triggers on `include: '*'`.

The Encryption package has the same pattern (`3.52.0-preview.0`), but that's a long-published version. `3.60.0-preview.0` is the current in-development preview that may not yet be on the feed.

**Suggested action:** Verify that `3.60.0-preview.0` is available on the feed configured in `NuGet.config`. If not, either pin to the latest published preview, or add the AI package's parity builds to `build-preview.yml` (mirroring how Encryption's parity check is structured there) and use `SdkProjectRef` for the solution build.

xinlian12:
✅ **Review complete** (30:54)

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
ananth7592: Verified: `Microsoft.Azure.Cosmos 3.60.0-preview.0` is published on nuget.org (https://www.nuget.org/packages/Microsoft.Azure.Cosmos/3.60.0-preview.0), so the `build-preview.yml` Job 1 restore resolves cleanly. No change required.

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
ananth7592: Tracked as follow-up for PR #5842, which is the first PR introducing public AI types. ContractEnforcement.cs + DotNetSDKAIAPI.json baseline will land there before any public surface is added, mirroring the Encryption test project setup.

ananth7592:
**Why a separate `azure-pipelines-cosmosai-official.yml` rather than folding AI into the main SDK pipelines?**

A few concrete reasons, all of which mirror the established pattern in `azure-pipelines-encryption-official.yml`:

1. **Independent release cadence.** The AI package version is governed by its own properties in `Directory.Build.props` (`CosmosAIOfficialVersion`, `CosmosAIPreviewVersion`, `CosmosAIPreviewSuffixVersion`) — fully decoupled from `ClientOfficialVersion` / `ClientPreviewVersion`. We need to be able to ship an AI patch (e.g., a provider bug-fix) without cutting a new core SDK release, and vice versa. A separate official pipeline is the cleanest way to enforce that.

2. **Separate approval gate.** Official releases require a manual approval on the publish stage. Keeping the AI release on its own pipeline means the AI release approver list, retention policy, and partner-drop blob path (`cosmosdb/csharp/cosmosai/$(BlobVersion)`) are scoped independently from the core SDK approval workflow.

3. **Different consumer surface.** AI publishes to a distinct NuGet package id (`Microsoft.Azure.Cosmos.AI`) with its own symbol package, signing manifest, and partner-drop destination. Bundling it into `azure-pipelines.yml` would either pollute every core-SDK release with AI artifacts or require complex conditional gating inside a single mega-pipeline.

4. **Precedent.** `Microsoft.Azure.Cosmos.Encryption` has used this exact "separate official pipeline + shared template" structure (`templates/encryption-nuget-pack.yml`, `azure-pipelines-encryption-official.yml`) for years. The AI pipelines (`templates/cosmosai-nuget-pack.yml` + `azure-pipelines-cosmosai-official.yml`) deliberately mirror that layout so the maintenance pattern is consistent across the optional-extension packages.

CI coverage is **not** sacrificed by this split: `azure-pipelines-cosmosai.yml` runs on every PR touching the AI package (and now also `Directory.Build.props` after commit 3b1fdb0e0), so the AI package still gets full pack + parity validation on each change. The "official" pipeline only handles the release-time signing, symbol packaging, and partner-drop publishing.

Happy to consolidate into the main SDK pipelines if the team would rather move that direction across both Encryption and AI together — that would be a larger refactor that should land as its own change rather than be bundled into this scaffold.

Directory.Build.props
ananth7592: Done in 3b1fdb0e0 — added `COSMOSAIPREVIEW` to the existing `DefineConstants` line alongside `PREVIEW;ENCRYPTIONPREVIEW` so future preview-gated AI code can `#if COSMOSAIPREVIEW` without retrofitting the version pattern.

Microsoft.Azure.Cosmos.AI/src/Mirrored/AssemblyKeys.cs
```diff
@@ -0,0 +1,12 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.AI
+{
+    internal static class AssemblyKeys
+    {
+        /// <summary>TestPublicKey is an unsupported strong key for testing and internal use only.</summary>
+        internal const string TestPublicKey = ", PublicKey=0024000004800000940000000602000000240000525341310004000001000100197c25d0a04f73cb271e8181dba1c0c713df8deebb25864541a66670500f34896d280484b45fe1ff6c29f2ee7aa175d8bcbd0c83cc23901a894a86996030f6292ce6eda6e6f3e6c74b3c5a3ded4903c951e6747e6102969503360f7781bf8bf015058eb89b7621798ccc85aaca036ff1bc1556bb7f62de15908484886aa8bbae";
```
ananth7592, line 10: Will take it up in later versions to use the latest .NET Automatic build time generation in the Post GA timeline

azure-pipelines-cosmosai.yml
ananth7592: Done in 3b1fdb0e0 — added `'Directory.Build.props'` to `pr.paths.include` in `azure-pipelines-cosmosai.yml` so version bumps to `CosmosAIOfficialVersion` / `CosmosAIPreviewVersion` / `CosmosAIPreviewSuffixVersion` now trigger the AI-specific static-analysis + three-build parity validation, even when no file under `Microsoft.Azure.Cosmos.AI/` is touched.

ananth7592:
## Local validation on commit `f67098f96`

All review feedback applied. Branch is now a single squashed commit. Local build matrix:

### Three-build parity (AI src project)

| # | Configuration | Core SDK source | Result |
|---|---|---|---|
| 1 | Stable | `Microsoft.Azure.Cosmos` NuGet `[3.59.0,)` | ✅ 0 warn / 0 err (5.6 s) |
| 2 | Preview (`-p:IsPreview=true`) | `Microsoft.Azure.Cosmos` NuGet `3.60.0-preview.0` | ✅ 0 warn / 0 err (2.6 s) |
| 3 | SdkProjectRef (`-p:SdkProjectRef=true`) | In-tree `ProjectReference` | ✅ 0 warn / 0 err (17.6 s) |

Command:
```powershell
dotnet build Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj -c Release [flags] --nologo
```

### Pack

```powershell
dotnet pack Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj -c Release -p:SdkProjectRef=true -o .\artifacts
```

✅ Produced `Microsoft.Azure.Cosmos.AI.1.0.0.nupkg` (6,931 bytes).

### Test project build (net8.0 after Pilchie feedback)

```powershell
dotnet build Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj -c Release
```

✅ 0 warn / 0 err. Output: `bin/Release/net8.0/Microsoft.Azure.Cosmos.AI.Tests.dll`.

### Cross-package smoke test (core SDK still compiles with AI project present)

```powershell
dotnet build Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj -c Release
```

✅ 0 warn / 0 err.

### Review-comment status

| # | Comment | Status |
|---|---|---|
| 1 | `PublishBuildArtifacts` nested inside release-only conditional | ✅ Fixed — outdented to step root in `templates/cosmosai-nuget-pack.yml`, matching `encryption-nuget-pack.yml`. |
| 2 | `Directory.Build.props` missing from CI trigger paths | ✅ Fixed — added to `pr.paths.include` in `azure-pipelines-cosmosai.yml`. |
| 3 | `COSMOSAIPREVIEW` define constant missing | ✅ Fixed — added alongside `PREVIEW;ENCRYPTIONPREVIEW` in root `Directory.Build.props`. |
| 4 | No API contract enforcement test infrastructure | ✅ Deferred to #5849 (first PR introducing public AI surface). |
| 5 | Main CI will NuGet-restore `3.60.0-preview.0` | ✅ Verified — `3.60.0-preview.0` is published on nuget.org; restore resolves cleanly. |
| 6 (Pilchie, raised on #5849) | `net6.0` is out of support; target `net8.0` | ✅ Applied here too — test project bumped to `net8.0`. Library stays `netstandard2.0` (max consumer reach; matches the core SDK and Encryption library). |

### Note: pre-existing main-branch build break in Encryption

A solution-level build with `-p:SdkProjectRef=true` (no `-p:IsPreview=true`) fails on `origin/main` HEAD `da4efa8c8` with 5× CS0115 in `Microsoft.Azure.Cosmos.Encryption` / `Microsoft.Azure.Cosmos.Encryption.Custom`. Root cause is a `#if` guard mismatch where the base `Container.IsFeedRangePartOfAsync` / `GetPartitionKeyRangesAsync` are gated `#if PREVIEW` but the derived overrides use `#if SDKPROJECTREF` (or are unguarded). **Unrelated to this PR** — repro on clean `origin/main` confirmed.

Ready for re-review. CC @xinlian12.

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
```diff
@@ -0,0 +1,80 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
NaluTripician, line 1: The PR description says the `.sln` got an "AI" solution folder with both projects, but the diff doesn't show any change to `Microsoft.Azure.Cosmos.sln` and I can't find `Cosmos.AI` referenced anywhere in it. Was this dropped in a rebase?

This matters because `templates/build-internal.yml` and `templates/build-preview.yml` both build via `Microsoft.Azure.Cosmos.sln` on every PR — that's how core-SDK changes get caught against Encryption today (the static-tools-encryption template cites PR #5825 and PR #5783 as prior incidents of exactly this class). The new `azure-pipelines-cosmosai.yml` only triggers on `Microsoft.Azure.Cosmos.AI/**`, so any future core-SDK PR that breaks the AI package will sail through and only surface on the next AI-directory PR.

Adding both projects to the `.sln` under an "AI" solution folder fixes it. Sanity check: `dotnet build Microsoft.Azure.Cosmos.sln -p:IsPreview=true -p:SdkProjectRef=true` should build AI alongside Encryption.

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
```diff
@@ -0,0 +1,80 @@
...
+    <CosmosAIVersion Condition=" '$(IsPreview)' == 'true' ">$(CosmosAIPreviewVersion)</CosmosAIVersion>
+    <CosmosAIVersionSuffix Condition=" '$(IsPreview)' == 'true' ">$(CosmosAIPreviewSuffixVersion)</CosmosAIVersionSuffix>
+    <Version Condition=" '$(CosmosAIVersionSuffix)' == '' ">$(CosmosAIVersion)</Version>
+    <Version Condition=" '$(CosmosAIVersionSuffix)' != '' ">$(CosmosAIVersion)-$(CosmosAIVersionSuffix)</Version>
+    <CurrentDate>$([System.DateTime]::Now.ToString(yyyyMMdd))</CurrentDate>
+    <Company>Microsoft Corporation</Company>
+    <Authors>Microsoft</Authors>
+    <Description>AI extensions library for Azure Cosmos DB for NoSQL. Provides an Azure OpenAI / Foundry-backed embedding provider for automatic vector embedding generation in hybrid and vector search queries. Designed to be extensible for broader AI scenarios including semantic re-ranking, OCR, and model integrations. For more information, refer to https://aka.ms/CosmosDB</Description>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <Title>Microsoft Azure Cosmos DB AI extensions library</Title>
```
NaluTripician: Nit: `<Title>...AI extensions library</Title>` is plural — FaultInjection and Encryption both use singular ("Fault Injection library", "encryption library"). Title surfaces on nuget.org, so worth aligning with the sibling convention if not intentional.

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
```diff
@@ -0,0 +1,80 @@
...
+    <PackageId>Microsoft.Azure.Cosmos.AI</PackageId>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageProjectUrl>https://github.com/Azure/azure-cosmos-dotnet-v3</PackageProjectUrl>
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <PackageTags>microsoft;azure;cosmos;cosmosdb;documentdb;nosql;azureofficial;dotnetcore;netcore;netstandard;ai;embedding;vector;openai;semanticranking</PackageTags>
```
NaluTripician, line 27: Two artifacts worth seeding now, before the first preview push to nuget.org:

1. **`Microsoft.Azure.Cosmos.AI/changelog.md`** — both Encryption and FaultInjection ship a per-package changelog. FaultInjection's csproj also has `<PackageReleaseNotes>` pointing at it so customers can find release notes from nuget.org. Without one, the first preview push has no discoverable release notes and the implementation PR (#5842) ends up scope-creeping one in.
2. **`Microsoft.Azure.Cosmos.AI/README.md`** — FaultInjection has one, and it's what nuget.org renders when `<PackageReadmeFile>` is wired up. Trivial to seed with a stub.

Both are essentially free at the scaffold stage and load-bearing once the package is on nuget.org.

Microsoft.Azure.Cosmos.AI/src/Mirrored/AssemblyKeys.cs
```diff
@@ -0,0 +1,12 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos.AI
+{
+    internal static class AssemblyKeys
```
NaluTripician, line 7: This makes three copies of the same hard-coded test public key in the repo (Encryption, Encryption.Custom, and now AI), all under "Mirrored" folders with no canonical source. If the key is ever rotated, three files diverge.

Out of scope for this PR, but worth a follow-up: extract a single `AssemblyKeys.cs` (e.g., under `Microsoft.Azure.Cosmos/build/`) and `<Compile Include="..." Link="Mirrored\AssemblyKeys.cs" />` it from all three packages.

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <Platform>AnyCPU</Platform>
+    <TargetFramework>net8.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <RootNamespace>Microsoft.Azure.Cosmos.AI.Tests</RootNamespace>
+    <LangVersion>$(LangVersion)</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
```
NaluTripician: These pins are very stale to inherit from Encryption — Encryption is locked into them by accumulated test history, but a brand-new test project with zero tests doesn't have that constraint:

- `Microsoft.NET.Test.Sdk` **16.2.0** is from 2019 (current 17.13+)
- `MSTest.TestAdapter` / `MSTest.TestFramework` **1.2.0** is from 2016; MSTest is on 3.x and has consolidated under a single `MSTest` meta-package
- `Moq` **4.8.3** predates the 2023 SponsorLink supply-chain incident
- `coverlet.collector` **3.1.2** vs. current 6.x

Cheapest moment to land on the modern stack is right now, before #5842 lands real tests. Worth taking either the Azure SDK Engineering recommended test stack or whatever the newer test projects in this repo are using.

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <PackageReference Include="Moq" Version="4.8.3" />
+    <PackageReference Include="MSTest.TestAdapter" Version="1.2.0" />
+    <PackageReference Include="MSTest.TestFramework" Version="1.2.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageReference Include="coverlet.collector" Version="3.1.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Azure.Cosmos.AI.csproj" />
```
NaluTripician, line 29: `Microsoft.Azure.Cosmos.Encryption.Tests.csproj` links in `ContractEnforcement.cs` from the main SDK tests and ships `DotNetSDKEncryptionAPI.net6.json` / `.net8.json` baselines that lock the public surface. This project has neither.

A new GA-track package is the cheapest moment to wire this up — once #5842 lands `AzureOpenAIEmbeddingGenerator` and whatever options/builder types come with it, a single misplaced `public` modifier or accidental rename ships into a preview NuGet with nothing in the PR pipeline to catch it. Suggest adding:

```xml
<Compile Include="..\..\..\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.Tests\Contracts\ContractEnforcement.cs" Link="Contracts\ContractEnforcement.cs" />
<Compile Include="..\..\..\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.Tests\Contracts\ContractType.cs" Link="Contracts\ContractType.cs" />
```

plus an empty `DotNetSDKCosmosAIAPI.net8.json` baseline and a `[TestClass]` that calls `ContractEnforcement.ValidateContractContainBreakingChanges(...)`. Baseline stays empty for this PR; #5842 fills it.

templates/static-tools-cosmosai.yml
```diff
@@ -0,0 +1,111 @@
...
+  # This matches the exact preprocessor surface the shipped NuGet package is
+  # built with (PREVIEW on, SDKPROJECTREF off) while still linking against the
+  # latest SDK abstractions.
+  - task: DotNetCoreCLI@2
+    displayName: Build Microsoft.Azure.Cosmos.AI (NuGet-surface parity vs main SDK source)
+    inputs: 
+      command: build
+      nugetConfigPath: NuGet.config
+      projects: Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
+      arguments: '-p:Optimize=true -p:IsPreview=true -p:SdkProjectRef=true -p:DefineSdkProjectRefSymbol=false --configuration Release'
```
NaluTripician, line 51: The 4th build variant from `static-tools-encryption.yml` is missing — the one that builds against main SDK source with `IsPreview` *unset*, i.e., no `PREVIEW` / `COSMOSAIPREVIEW` symbol defined. Encryption added it explicitly after PR #5825 (FFCF GA promotion) had a miss the preview-only parity build let through; the comment block in `static-tools-encryption.yml` calls that out.

AI is in exactly the same position: `Directory.Build.props` now defines `COSMOSAIPREVIEW` as a preview-only constant, so any future `#if COSMOSAIPREVIEW` block creates the same GA/preview surface divergence. Suggest adding a 4th `DotNetCoreCLI@2` step right after this one with `arguments: '-p:Optimize=true -p:SdkProjectRef=true -p:DefineSdkProjectRefSymbol=false --configuration Release'` (no `IsPreview=true`), and lifting the explanatory comment from the Encryption template.

templates/static-tools-cosmosai.yml
```diff
@@ -0,0 +1,111 @@
...
+      versioningScheme: OFF
+
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-binskim.BinSkim@4
+    displayName: 'BinSkim'
+    condition: eq(1,2) #disabling as nuget repo failing
```
NaluTripician: The `condition: eq(1,2) #disabling as nuget repo failing` line (and the matching three on CredScan/PoliCheck/PostAnalysis below) is years-old text copied verbatim from `static-tools-encryption.yml`. The result is that the **official release-gate pipeline** for a brand-new package never runs BinSkim, CredScan, PoliCheck, or PostAnalysis — only Component Governance and (`continueOnError: true`) AntiMalware actually run.

CredScan in particular has real value here since AI scenarios typically take endpoint URIs and token credentials. Default for a new package should arguably be "on." Suggest either (a) removing the `condition: eq(1,2)` lines (the "nuget repo failing" rationale needs re-validation — it's stale), or (b) keeping them disabled with a `# TODO(#NNNN)` referencing a real tracking issue and a removal target, not a copy of a comment from a different package.

NaluTripician (commented):
Did a structural diff against `Microsoft.Azure.Cosmos.Encryption` since that's the pattern this scaffold is following.Most of it lines up cleanly; a few items below are worth sorting before this lands, since they get more expensive to fix once #5842 puts real code on top of the scaffold.

Headline items:
- 🔴 The `.sln` update mentioned in the PR description doesn't appear in the diff — leaves the AI projects off the per-PR build path that catches core-SDK breakages today.
- 🟡 Missing the 4th GA-surface parity build step from `static-tools-encryption.yml` (the one added after PR #5825 had a miss).
- 🟡 No per-package changelog, README, or contract baseline — easy to seed at scaffold time, ugly to scope-creep into the implementation PR.
- 🟡 Test project pins (`Microsoft.NET.Test.Sdk 16.2.0`, `MSTest 1.2.0`, `Moq 4.8.3`, `coverlet 3.1.2`) are 2016–2019 vintage. New test project = cheapest moment to land on the modern stack.
- 🟡 GDN tasks (BinSkim/CredScan/PoliCheck/PostAnalysis) inherit the disabled `condition: eq(1,2)` line verbatim from Encryption with a years-old comment — at minimum needs a tracking issue.

Inline comments below.

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
ananth7592: Done in db7ea2e09 — both Microsoft.Azure.Cosmos.AI and Microsoft.Azure.Cosmos.AI.Tests are now in Microsoft.Azure.Cosmos.sln under a new AI solution folder (mirrors the existing Encryption / Encryption.Custom folders). Verified locally with dotnet build Microsoft.Azure.Cosmos.sln -p:IsPreview=true -p:SdkProjectRef=true --configuration Release — 0 errors, 16 pre-existing warnings (all Microsoft.Azure.Cosmos.Performance.Tests net6.0 EOL warnings, unrelated). This now means every PR build that runs the solution will catch core-SDK changes that break the AI package, same as the Encryption guardrail.

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
ananth7592: Done in db7ea2e09 — Microsoft Azure Cosmos DB AI extension library (singular), to align with Microsoft Azure Cosmos DB Fault Injection library and Microsoft Azure Cosmos DB client-side encryption library.

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
ananth7592: Done in db7ea2e09:

- Seeded Microsoft.Azure.Cosmos.AI/changelog.md (Keep-a-Changelog format, mirrors Encryption) with an Unreleased entry describing the scaffold.
- Seeded Microsoft.Azure.Cosmos.AI/README.md describing the package, why it's separate, and planned capabilities — mirrors the FaultInjection README structure.
- Wired <PackageReleaseNotes> pointing at the changelog (FaultInjection pattern) and <PackageReadmeFile>README.md</PackageReadmeFile> so nuget.org renders the README.

#5842 inherits the changelog Unreleased bucket — its description should add a customer-facing entry under ### Added for AzureOpenAIEmbeddingGenerator, per the repo's changelog classifier.

Microsoft.Azure.Cosmos.AI/src/Mirrored/AssemblyKeys.cs
ananth7592: Agreed and noted — leaving the 3 Mirrored\AssemblyKeys.cs copies as-is in this PR per your out-of-scope call. I'll file a follow-up issue to extract a single canonical AssemblyKeys.cs under Microsoft.Azure.Cosmos/build/ and <Compile Link> it from all three packages (Encryption, Encryption.Custom, AI), and link the issue from this thread once filed.

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
ananth7592: Done in db7ea2e09 — modernized to match what the main SDK tests project (Microsoft.Azure.Cosmos.Tests.csproj) uses today:

| Pin | Before | After |
|---|---|---|
| Microsoft.NET.Test.Sdk | 16.2.0 | 17.12.0 |
| MSTest.TestAdapter | 1.2.0 | 3.7.3 |
| MSTest.TestFramework | 1.2.0 | 3.7.3 |
| coverlet.collector | 3.1.2 | 6.0.2 |
| Moq | 4.8.3 | 4.8.3 (unchanged — matches main SDK tests pin) |

Left Moq at 4.8.3 deliberately so we're not diverging from the rest of the repo on the SponsorLink question in this PR — happy to bump it across the repo (or swap to NSubstitute) as a separate change if you'd like. Tests project builds and runs clean (Passed 1/1).

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
ananth7592: Done in db7ea2e09:

- Linked ..\..\..\Microsoft.Azure.Cosmos\tests\Microsoft.Azure.Cosmos.Tests\Contracts\ContractEnforcement.cs and ContractType.cs into the AI test project (same <Compile Include … Link> pattern as Microsoft.Azure.Cosmos.Encryption.Tests.csproj).
- Added Contracts\ContractEnforcementTests.cs with a [TestClass] that calls ValidateContract(dllName: "Microsoft.Azure.Cosmos.AI", baselinePattern: "DotNetSDKCosmosAIAPI", breakingChangesPattern: "DotNetSDKCosmosAIAPIChanges").
- Added Contracts\DotNetSDKCosmosAIAPI.net8.json baseline. The AI assembly currently has no own public types, so the baseline only captures System.Object-inherited members on the assembly-level enumeration; #5842 fills it with real types via UpdateContracts.ps1.
- Wired the <None Include="Contracts\DotNetSDKCosmosAIAPI*.json"> copy-to-output entry so baselines flow to the test bin folder.

Test run: Passed 1/1 (ContractChanges).

templates/static-tools-cosmosai.yml
ananth7592: Done in db7ea2e09 — added the 4th DotNetCoreCLI@2 step right after the SDKPROJECTREF parity step, with rguments: '-p:Optimize=true -p:SdkProjectRef=true -p:DefineSdkProjectRefSymbol=false --configuration Release' (no IsPreview=true). Lifted the explanatory comment from static-tools-encryption.yml and updated the constant name to COSMOSAIPREVIEW (which Directory.Build.props defines as the preview-only constant for this package). The cross-reference to PR #5825 is preserved.

templates/static-tools-cosmosai.yml
ananth7592: Partial: in db7ea2e09 I replaced the verbatim `condition: eq(1,2) #disabling as nuget repo failing` comments on BinSkim, CredScan, PoliCheck, and PostAnalysis with `condition: eq(1,2) # TODO(#5845): re-enable …` markers, and added a block comment at the top of the GDN section explaining the intent (verify OneES NuGet feed access for these tasks, then remove the gates) and calling out that CredScan in particular matters for AI endpoint URIs / token credentials.

I stopped short of flipping the conditions on in this PR because I can't validate the OneES feed access from a local dev box, and a release-gate pipeline failing on its first run after this lands would block #5842. Happy to flip them on in a follow-up PR as soon as a release-gate dry run confirms green — or earlier in this PR if you'd prefer I just remove the conditions and let CI tell us. Your call.

ananth7592:
## Verification ✅

Re-ran the full test battery from a clean state on commit `db7ea2e09`. All 4 static-tools build variants (matching what `templates/static-tools-cosmosai.yml` runs in CI) and the test project are green.

| # | Variant | Command | Result |
|---|---|---|---|
| 1 | NuGet SDK reference (default customer-consumed build) | `dotnet build ... -p:Optimize=true --configuration Release` | ✅ 0 warnings, 0 errors |
| 2 | SDKREF main SDK source (preview) | `... -p:Optimize=true -p:IsPreview=true -p:SdkProjectRef=true --configuration Release` | ✅ 0 warnings, 0 errors |
| 3 | NuGet-surface parity (PREVIEW on, `SDKPROJECTREF` off) | `... -p:IsPreview=true -p:SdkProjectRef=true -p:DefineSdkProjectRefSymbol=false --configuration Release` | ✅ 0 warnings, 0 errors |
| 4 | **GA-surface parity (no `IsPreview`, `SDKPROJECTREF` off) — NEW step added in `db7ea2e09` per @NaluTripician** | `... -p:SdkProjectRef=true -p:DefineSdkProjectRefSymbol=false --configuration Release` | ✅ 0 warnings, 0 errors |
| 5 | `dotnet test` AI tests (includes new `ContractEnforcementTests`) | `dotnet test ...AI.Tests.csproj --configuration Release` | ✅ Passed 1/1 (`ContractChanges`, 69 ms) |

### Full sln build (validates `.sln` update)
```
dotnet build Microsoft.Azure.Cosmos.sln -p:IsPreview=true -p:SdkProjectRef=true --configuration Release
```
→ **0 errors**. 16 warnings, all pre-existing `net6.0` EOL warnings on `Microsoft.Azure.Cosmos.Performance.Tests` (unrelated to this PR; predate it on `main`).

This confirms `Microsoft.Azure.Cosmos.AI` and `Microsoft.Azure.Cosmos.AI.Tests` now compile as part of every per-PR build (`templates/build-internal.yml` / `templates/build-preview.yml`), as called for in the inline `.sln` thread.

### Updates the PR description's "Verification" section
The original PR description claimed:
- ~~`dotnet test` → No tests yet (empty), exits 0~~

Now reads:
- `dotnet test` → **1 passed (`ContractChanges`)** — the new contract-enforcement guardrail wired up per the review comment on `Microsoft.Azure.Cosmos.AI.Tests.csproj`.

Inline replies on each of @NaluTripician's 8 comments already posted (above on this PR).

xinlian12 (approved):
LGTM, thanks

Microsoft.Azure.Cosmos.sln
```diff
@@ -39,6 +39,12 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FaultInjectionTests", "Micr
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests", "Microsoft.Azure.Cosmos.Encryption.Custom\tests\Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests\Microsoft.Azure.Cosmos.Encryption.Custom.Performance.Tests.csproj", "{CE4D6DA8-148D-4A98-943B-D8C2D532E1DC}"
```
kundadebdatta, line 40: Do we need to check in this file ?

azure-pipelines-cosmosai.yml
```diff
@@ -0,0 +1,33 @@
...
+    - 'Microsoft.Azure.Cosmos.AI/**'
```
kundadebdatta, line 14: **🟡 Recommendation** · CI/Pipelines: Trigger Coverage

**Parity-build CI never runs on core SDK PRs that can break it**

The 4-variant parity build in `static-tools-cosmosai.yml` (NuGet-ref, SdkProjectRef, NuGet-surface parity, GA-surface parity) exists to catch core SDK source changes that would break the AI package's compiled surface against the next shipped SDK NuGet. But `pr.paths.include` here only triggers on changes under `Microsoft.Azure.Cosmos.AI/**` and the listed YAML/props — not on `Microsoft.Azure.Cosmos/**`. So a core-SDK PR that renames/removes a member the AI package relies on will not fire the AI parity builds.

The `.sln` addition (per @NaluTripician's feedback) helps the main solution build compile AI alongside the SDK, but the main solution build does not exercise the `SdkProjectRef`-off parity variants that catch `#if PREVIEW` / `#if SDKPROJECTREF` / `#if COSMOSAIPREVIEW` divergence — exactly the regression class that motivated PR #5825's fix in `static-tools-encryption.yml`.

Recommendation: add `'Microsoft.Azure.Cosmos/**'` to `pr.paths.include`, or accept the gap explicitly with a comment explaining why the solution-level main CI is considered sufficient.

---

templates/static-tools-cosmosai.yml
```diff
@@ -0,0 +1,137 @@
...
+  # TODO(#5845): The Guardian tasks below (BinSkim, CredScan, PoliCheck,
+  # PostAnalysis) inherit a years-old "condition: eq(1,2) #disabling as nuget
+  # repo failing" gate from static-tools-encryption.yml. Re-enable on this
+  # brand-new package once the OneES NuGet feed access for these tasks is
+  # confirmed green; CredScan in particular has real value here because AI
+  # scenarios commonly handle endpoint URIs and token credentials. Tracking
+  # work consolidated under PR #5845 follow-up; remove the eq(1,2) gates and
+  # this comment once validated.
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-binskim.BinSkim@4
+    displayName: 'BinSkim'
```
kundadebdatta, line 81: **🟡 Recommendation** · Security: Guardian Gates Disabled

**GDN tasks still gated `eq(1,2)`; needs an explicit decision before merge**

The release-gate pipeline of a brand-new package never runs BinSkim, CredScan, PoliCheck, or PostAnalysis — only Component Governance and a `continueOnError: true` AntiMalware actually execute.

CredScan is especially load-bearing for an AI package: AI scenarios are precisely where customers paste endpoint URIs and API keys into samples / docs / tests. A new package should enter the world with credential scanning on by default.

The `TODO` points at `#5845` itself, so once this merges the TODO has no real owner. Recommendation: flip them on now, let the first release-gate run tell us whether OneES feed access is actually broken, and gate them back off (with a real tracking issue link, not `#5845`) only if it fails.

---

templates/static-tools-cosmosai.yml
```diff
@@ -0,0 +1,137 @@
...
+    displayName: 'PoliCheck'
+    condition: eq(1,2) # TODO(#5845): re-enable once OneES nuget feed access verified
+    inputs:
+      targetType: F
+      optionsFC: 0
+
+  # AntiMalware scan
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-antimalware.AntiMalware@4
+    displayName: 'AntiMalware'
+    continueOnError: true # signature refresh failing resulting in tasks failures
```
kundadebdatta, line 113: **🟡 Recommendation** · Security: AntiMalware Soft-Failed

**`AntiMalware` step set to `continueOnError: true`**

Inherited verbatim from the Encryption template, with a justification ("signature refresh failing") that's a transient infra issue from another package. Combined with all four other Guardian tasks being `eq(1,2)`-gated (see separate comment), this means for the AI package **no security-scan task can actually break the release-gate build** — AntiMalware will swallow a real malware signature hit and let the package publish.

Defense in depth fails when every layer is "don't fail the build." Either drop `continueOnError: true` for the AI package (it's brand new — no historical infra excuse), or treat this and the Guardian gating as one decision: at least one Guardian gate must be hard-failing for the official release pipeline to be meaningfully gated.

---

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
```diff
@@ -0,0 +1,83 @@
...
+    <LangVersion>$(LangVersion)</LangVersion>
+    <CosmosAIVersion Condition=" '$(IsPreview)' != 'true' ">$(CosmosAIOfficialVersion)</CosmosAIVersion>
+    <CosmosAIVersion Condition=" '$(IsPreview)' == 'true' ">$(CosmosAIPreviewVersion)</CosmosAIVersion>
+    <CosmosAIVersionSuffix Condition=" '$(IsPreview)' == 'true' ">$(CosmosAIPreviewSuffixVersion)</CosmosAIVersionSuffix>
+    <Version Condition=" '$(CosmosAIVersionSuffix)' == '' ">$(CosmosAIVersion)</Version>
+    <Version Condition=" '$(CosmosAIVersionSuffix)' != '' ">$(CosmosAIVersion)-$(CosmosAIVersionSuffix)</Version>
+    <CurrentDate>$([System.DateTime]::Now.ToString(yyyyMMdd))</CurrentDate>
+    <Company>Microsoft Corporation</Company>
+    <Authors>Microsoft</Authors>
+    <Description>AI extensions library for Azure Cosmos DB for NoSQL. Provides an Azure OpenAI / Foundry-backed embedding provider for automatic vector embedding generation in hybrid and vector search queries. Designed to be extensible for broader AI scenarios including semantic re-ranking, OCR, and model integrations. For more information, refer to https://aka.ms/CosmosDB</Description>
```
kundadebdatta, line 15: **🟢 Suggestion** · Packaging: Public Description Accuracy

**Description on the NuGet listing over-promises capabilities**

Two issues with text that ships verbatim to nuget.org search results & package page:

1. **"Foundry-backed"** — the only AI dependency is `Azure.AI.OpenAI 2.1.0`; there is no Azure AI Foundry SDK dependency. If "Foundry" here means "Azure AI Foundry Models endpoint reached via the OpenAI SDK," that subtle distinction is obscured.
2. **"OCR"** — the README's Planned Capabilities lists embedding generation, semantic re-ranking, and AI-assisted query construction. OCR is not mentioned in the README or in the PR description. Setting that expectation on the public listing for a package that ships zero APIs sets up a "where's the OCR?" disconnect.

Easier to scope it down now (drop "Foundry-backed" and "OCR" until they have a real shipping plan) than to walk it back once customers see it.

---

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Contracts/ContractEnforcementTests.cs
```diff
@@ -0,0 +1,32 @@
...
+    [TestCategory("UpdateContract")]
+    [TestClass]
+    public class ContractEnforcementTests
+    {
+        /// <summary>
+        /// Locks the public API surface of <c>Microsoft.Azure.Cosmos.AI</c> against a
+        /// versioned baseline so accidental renames, visibility changes, or removals
+        /// surface at PR review time instead of shipping to a preview NuGet.
+        ///
+        /// IMPORTANT: Because tests can run on multiple .NET versions, the contract
```
kundadebdatta, line 15: **🟢 Suggestion** · Maintainability: Misleading Documentation

**`ContractEnforcementTests.cs` — misleading multi-target comment**

The doc comment frames this as if the project is multi-targeted, but the test csproj is single-targeted to `net8.0`. There's one baseline file and one target framework. The "Because tests can run on multiple .NET versions—" preamble will mislead the next reader into hunting for non-existent `.net6.json` / `.net9.json` baselines.

Either trim the comment ("Locks the public API surface against `DotNetSDKCosmosAIAPI.net8.json`. Update via `UpdateContracts.ps1` from the repo root."), or multi-target the test project to `net6.0;net8.0` and seed the additional baseline (which the doc currently implies is the intent).

---

Microsoft.Azure.Cosmos.AI/src/Microsoft.Azure.Cosmos.AI.csproj
```diff
@@ -0,0 +1,83 @@
...
+  </ItemGroup>
+
+  <!-- Required by Microsoft.Azure.Cosmos SDK; PrivateAssets=All keeps it off the public dependency graph -->
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.2" NoWarn="NU1903" PrivateAssets="All" />
+  </ItemGroup>
+
+  <!-- Azure OpenAI provider dependency -->
+  <ItemGroup>
+    <PackageReference Include="Azure.AI.OpenAI" Version="2.1.0" />
```
kundadebdatta, line 59: **🟢 Suggestion** · Packaging: Dependency Version Range

**`Azure.AI.OpenAI` pinned to exact version may conflict with customer apps**

`Azure.AI.OpenAI` is a fast-moving package customers commonly already reference directly. A bare `Version="2.1.0"` is NuGet shorthand for `[2.1.0,)` — fine for resolution, but it does flow to consumers (no `PrivateAssets="All"`), which means any breaking change in `Azure.AI.OpenAI` between 2.x minors or a future 3.x will surface to AI-package customers.

Worth stating an intentional supported range (e.g., `[2.1.0,3.0.0)`) so the API contract is explicit, mirroring how dependent ranges are stated for the core SDK above. Pick a deliberate compatibility posture for the package's only public, customer-visible dependency before #5842 ships and customers start adopting.

---

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Microsoft.Azure.Cosmos.AI.Tests.csproj
```diff
@@ -0,0 +1,49 @@
...
+
+  <PropertyGroup>
+    <SignAssembly>true</SignAssembly>
+    <DelaySign>true</DelaySign>
+    <AssemblyOriginatorKeyFile>..\..\..\testkey.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <PlatformTarget>x64</PlatformTarget>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
kundadebdatta, line 46: **🟢 Suggestion** · Maintainability: Inherited csproj Cruft

**`tests.csproj` carries inherited `AllowUnsafeBlocks` / `PlatformTarget x64` cruft**

The test project has zero unsafe code today and no x64-specific dependencies. The `Debug|AnyCPU` `PropertyGroup` with `PlatformTarget=x64` + `AllowUnsafeBlocks=true` looks copy-pasted from a sibling test csproj that did need it.

For a brand-new test project both flags are dead config; `AllowUnsafeBlocks=true` in particular is something analyzers and supply-chain audits flag, and silently enabling it on a test project is unhelpful.

@NaluTripician's earlier comment about modernizing the package pins already led to a refresh; this is the same class of "drop inherited cruft at scaffold time" change. Trivial to delete now, cheaper than explaining later why the AI test project allows unsafe code.

---

Microsoft.Azure.Cosmos.AI/tests/Microsoft.Azure.Cosmos.AI.Tests/Contracts/DotNetSDKCosmosAIAPI.net8.json
```diff
@@ -0,0 +1,44 @@
+{
+  "Subclasses": {},
+  "Members": {
+    "Boolean Equals(System.Object)": {
```
kundadebdatta, line 4: **💬 Observation** · Maintainability: Contract Baseline Sanity Check

**Contract baseline contains only `System.Object` members at root**

Not a bug — author confirmed this was generated by `UpdateContracts.ps1` and the test passes — but worth a quick sanity check. The baseline's root-level `Members` listing the inherited `System.Object` methods looks odd; one would more typically expect the root to be empty (zero public types) and the `System.Object` members to surface only as the implicit base when a real public type is added.

If this is in fact what the contract tool emits for an empty assembly, fine. If it's actually capturing some unintended public surface (e.g., the assembly attribute path), it'd be good to confirm now — the contract diff produced when #5842 adds the first real public type is going to be wide either way, and starting from an unusual baseline could make that diff harder to review.

Low-effort sanity check: re-run `UpdateContracts.ps1` from clean and confirm the produced baseline matches what's checked in.

---

PR #5850, opened 2026-05-09
[WIP] Routing: Fixes HPK partition key query returning empty results during partition split

ananth7592:
## Problem

During or shortly after a partition split, HPK (Hierarchical Partition Key) point queries (full 3-component partition key) can silently return zero results with no backend contact and no exception.

Observed in production with SDK 3.58.0. Diagnostics confirmed:
- Successful sibling query at 23:10: contacts backend, returns 1 doc, RequestCharge=6.46
- Failing query at 23:21: no MoveNextAsync, no backend, 0 docs, 0 RU

### Primary bug: LengthAware vs ordinal mismatch in GetOverlappingRanges

After a partition split, the routing map has a 128-char boundary (H1+H2+H3 + 32 trailing zeros). A full 3-component HPK point query produces a 96-char EPK = H1+H2+H3 (no trailing zeros).

GetOverlappingRanges uses two different comparison semantics:

1. LengthAwareMinComparer.BinarySearch -- trims trailing zeros before comparing. Sees TrimEnd(boundary_128) == EPK_96, finds exact match at PKRange2 (index 1), minIndex=1.
2. Range.CheckOverlapping -- uses ordinal string comparison. Checks if EPK_96 falls in [boundary_128, ...). Ordinal: EPK_96 < boundary_128 because a shorter string that is a prefix of a longer one sorts before it. Result: no overlap -> 0 ranges returned.

The document actually lives in PKRange1 because EPK_96 < boundary_128 ordinal, but GetOverlappingRanges never checks PKRange1. Routes to EmptyQueryPipelineStage -> silent empty result.

### Secondary bug: forceRefresh not plumbed through

CosmosQueryClientCore.GetTargetPartitionKeyRangesAsync did not pass forceRefresh to TryGetOverlappingRangesAsync, so the stale-cache retry path was a no-op.

### Primary fix: CollectionRoutingMap.GetOverlappingRanges fallback

After the main loop, if no ranges were added AND LengthAwareMinComparer found an exact match (minIndexRaw >= 0) AND minIndex > 0, fall back to check orderedRanges[minIndex - 1] with ordinal CheckOverlapping. This returns PKRange1 without affecting other query shapes:

- Range queries [EPK, EPK+FF]: main loop already adds PKRange2 because EPK+FF > boundary_128 ordinal -> fallback not triggered
- Normal point queries not at a LengthAware boundary: minIndexRaw < 0 -> fallback not triggered
- All existing CollectionRoutingMap tests continue to pass

### Secondary fix: CosmosQueryClientCore + CosmosQueryExecutionContextFactory

1. Pass forceRefresh through to TryGetOverlappingRangesAsync.
2. When EffectiveRangesForPartitionKey is non-null but returns 0 ranges, retry once with forceRefresh=true as defense-in-depth.

## Tests Added

Five new test methods in CollectionRoutingMapTest.cs covering the exact production scenario (128-char boundary, 96-char EPK). These previously FAILED with useLengthAwareRangeComparer=true. All 18 routing-map tests now pass.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/Core/Pipeline/CosmosQueryExecutionContextFactory.cs
@@ -639,6 +639,18 @@ namespace Microsoft.Azure.Cosmos.Query.Core.ExecutionContext
                     containerQueryProperties.EffectiveRangesForPartitionKey,
                     forceRefresh: false,
                     trace);
+
+                // If no overlapping ranges are found, the routing map might be stale (e.g., during/after a partition split).
+                // Retry with forceRefresh=true to get an up-to-date routing map before returning an empty result.
+                if (targetRanges.Count == 0)
+                {
+                    targetRanges = await queryClient.GetTargetPartitionKeyRangesAsync(
+                        resourceLink,
+                        containerQueryProperties.ResourceId,
+                        containerQueryProperties.EffectiveRangesForPartitionKey,
+                        forceRefresh: true,
+                        trace);
+                }
             }
             else if (TryGetEpkProperty(properties, out string effectivePartitionKeyString))
             {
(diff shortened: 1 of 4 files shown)


ananth7592:
Not a right fix to do

PR #5855, opened 2026-05-11
[Internal] DTS: Refactors Create/Upsert item ops in DTx to extract id from resource body

Meghana-Palaparthi:
# Pull Request Template

## Description

> Extension of PR [[Internal] DTS: Adds required document id to all DTS operations
](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5793).
To ensure consistency with Create/Upsert operations in other SDK flows, this PR implements a solution that extracts id from the resource body instead of having users provide it explicitly.

This pull request refactors how item IDs are handled in distributed transaction operations for Azure Cosmos DB. The main change is that the `id` for Create and Upsert operations is now automatically extracted from the resource body (the serialized JSON), rather than being passed explicitly as a parameter. This affects both the implementation and the public API, and is validated with new and updated tests.

**API and Behavior Changes:**

* The `CreateItem` and `UpsertItem` methods in `DistributedWriteTransaction` and their corresponding stream overloads no longer require the `id` parameter. Instead, the item's `id` is read directly from the serialized resource body or stream payload. [[1]](diffhunk://#diff-f9ec0d1444fc185a1ddc561a0187fd6c980277b524f5449eb50fd95ef9fe3887L27-L35) [[2]](diffhunk://#diff-f9ec0d1444fc185a1ddc561a0187fd6c980277b524f5449eb50fd95ef9fe3887L45-L53) [[3]](diffhunk://#diff-f9ec0d1444fc185a1ddc561a0187fd6c980277b524f5449eb50fd95ef9fe3887L154-L162) [[4]](diffhunk://#diff-f9ec0d1444fc185a1ddc561a0187fd6c980277b524f5449eb50fd95ef9fe3887L173-L181)

* The implementation in `DistributedTransactionOperation` now includes logic to extract the `id` property from the top-level JSON object in the resource body for Create and Upsert operations. If the `id` is missing or empty, an exception is thrown. [[1]](diffhunk://#diff-30fceb3d3e008bbde11f8dd7453a6dcef707a5452f9e235a858068b93dd5e1dbR9-R10) [[2]](diffhunk://#diff-30fceb3d3e008bbde11f8dd7453a6dcef707a5452f9e235a858068b93dd5e1dbR20-R21) [[3]](diffhunk://#diff-30fceb3d3e008bbde11f8dd7453a6dcef707a5452f9e235a858068b93dd5e1dbR82-R163)

**Codebase Refactoring:**

* All usages and overrides of `CreateItem` and `UpsertItem` in `DistributedWriteTransactionCore` and related classes have been updated to remove the `id` parameter and associated validation. [[1]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL30-L35) [[2]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL45) [[3]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL55-L60) [[4]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL73) [[5]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL223-L228) [[6]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL238) [[7]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL248-L253) [[8]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL266)

**Testing:**

* A new test verifies that the `id` is correctly extracted from the resource body for both Create and Upsert operations.

* Existing tests have been updated to use the new overloads (without the `id` parameter) and to validate the correct serialization of operations. [[1]](diffhunk://#diff-ab125ecaef08e5023deceb8459c3ae5e9dcb9b09d6291254200b8822f6c7fc9fL72-R104) [[2]](diffhunk://#diff-ab125ecaef08e5023deceb8459c3ae5e9dcb9b09d6291254200b8822f6c7fc9fL98-R129) [[3]](diffhunk://#diff-ab125ecaef08e5023deceb8459c3ae5e9dcb9b09d6291254200b8822f6c7fc9fL125-R157) [[4]](diffhunk://#diff-ab125ecaef08e5023deceb8459c3ae5e9dcb9b09d6291254200b8822f6c7fc9fL152-R183)

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperation.cs
@@ -6,6 +6,8 @@ namespace Microsoft.Azure.Cosmos
 {
     using System;
     using System.IO;
+    using System.Text;
+    using System.Text.Json;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Azure.Documents;
@@ -15,6 +17,8 @@ namespace Microsoft.Azure.Cosmos
     /// </summary>
     internal class DistributedTransactionOperation
     {
+        private static readonly byte[] IdPropertyNameUtf8 = Encoding.UTF8.GetBytes("id");
+
         protected Memory<byte> body;
 
         public DistributedTransactionOperation(
@@ -75,11 +79,88 @@ namespace Microsoft.Azure.Cosmos
             {
                 this.body = await BatchExecUtils.StreamToMemoryAsync(this.ResourceStream, cancellationToken);
             }
+
+            this.ReconcileIdWithBody();
+        }
+
+        /// <summary>
+        /// For Create/Upsert operations the item id is always derived from the resource body.
+        /// Extracts the top-level <c>id</c> string property and assigns it to <see cref="Id"/>.
+        /// Throws <see cref="ArgumentException"/> if the body has no non-empty <c>id</c> string.
+        /// </summary>
+        private void ReconcileIdWithBody()
+        {
+            if (this.OperationType != Documents.OperationType.Create
+                && this.OperationType != Documents.OperationType.Upsert)
+            {
+                return;
+            }
+
+            if (this.body.IsEmpty)
+            {
+                return;
+            }
+
+            string bodyId = TryReadIdFromJsonObject(this.body.Span);
+
+            if (string.IsNullOrWhiteSpace(bodyId))
+            {
+                throw new ArgumentException(
+                    $"The resource body for the {this.OperationType} operation at index {this.OperationIndex} does not contain a non-empty 'id' string property. Provide 'id' in the resource body.");
+            }
+
+            this.Id = bodyId;
+        }
+
+        /// <summary>
+        /// Scans the top level of a JSON object for an "id" string property and returns its value.
+        /// Returns null if the body is not a JSON object, has no "id" property, or the value is not a string.
+        /// Allocation is limited to the returned id string itself.
+        /// </summary>
+        private static string TryReadIdFromJsonObject(ReadOnlySpan<byte> body)
+        {
+            try
+            {
+                Utf8JsonReader reader = new Utf8JsonReader(body);
+                if (!reader.Read() || reader.TokenType != JsonTokenType.StartObject)
+                {
+                    return null;
+                }
+
+                while (reader.Read() && reader.TokenType == JsonTokenType.PropertyName)
+                {
+                    if (reader.ValueTextEquals(IdPropertyNameUtf8))
+                    {
+                        return reader.Read() && reader.TokenType == JsonTokenType.String ? reader.GetString() : null;
+                    }
+
+                    reader.Skip();
+                }
+
+                return null;
+            }
+            catch (JsonException)
+            {
+                return null;
+            }
         }
     }
 
     internal class DistributedTransactionOperation<T> : DistributedTransactionOperation
     {
+        public DistributedTransactionOperation(
+            Documents.OperationType operationType,
+            int operationIndex,
+            string database,
+            string container,
+            PartitionKey partitionKey,
+            T resource,
+            DistributedTransactionRequestOptions requestOptions = null)
+            : base(operationType, operationIndex, database, container, partitionKey, id: null, requestOptions)
+        {
+            this.Resource = resource;
+        }
+
         public DistributedTransactionOperation(
             Documents.OperationType operationType,
             int operationIndex,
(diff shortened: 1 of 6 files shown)


kushagraThapar:
## Review

Nice follow-up to #5793 — the `Utf8JsonReader` extraction is implemented well (correctly handles nested objects via `Skip()`, escape-aware `ValueTextEquals`, cached UTF-8 byte array, single allocation). No bugs found. Below are diagnosability / documentation suggestions, none are merge blockers.

### 🟡 Suggestions

1. **Document & enforce the UTF-8 (no BOM) contract.** `Utf8JsonReader` requires UTF-8 without a BOM. The typed path is safe with default Cosmos serializers, but a customer-plugged-in `CosmosSerializer` or a customer-supplied `Stream` could be UTF-16 or include a BOM (`EF BB BF`). In that case `ReadStartObject()`/`Read()` throws `JsonException`, the broad `catch` swallows it, and the user gets *"does not contain a non-empty 'id' string property"* — a misleading message whose real cause is encoding. Consider (a) adding an XML `<remarks>` to each new overload stating UTF-8 (no BOM) is required, and (b) optionally detecting & stripping a leading UTF-8 BOM before reading.

2. **Flag the behavior change for typed `<T>` callers using PascalCase `Id`.** With the default Json.NET resolver (no `[JsonProperty("id")]`, no camelCase contract), `public string Id { get; set; }` serializes as `"Id"`, not `"id"`, so extraction returns null and the op throws `ArgumentException` at commit. This aligns DTS with `Container.CreateItem<T>` semantics — good design — but customers porting from #5793 will hit it at runtime. The `TestItem` in this PR was already updated to use `[Newtonsoft.Json.JsonProperty("id")]`, which proves the point. Add an XML `<remarks>` to the typed overloads: *"The resource type must serialize a top-level lowercase `id` JSON property of type string, matching `Container.CreateItem<T>` requirements."*

3. **Empty body silently no-ops for Create/Upsert.** `if (this.body.IsEmpty) { return; }` leaves `Id = null` and the request fails downstream with a confusing error. For Create/Upsert specifically, an empty body is a programmer error and should throw the same `ArgumentException` the missing-id case throws.

4. **One error message conflates four failure modes.** *"...does not contain a non-empty 'id' string property..."* is produced for: (a) body isn't a JSON object, (b) no `id` property, (c) `id` value isn't a string (numeric / array / object / null), (d) `id` is empty/whitespace, plus (e) the `JsonException` path (malformed JSON / wrong encoding). Customers can't tell which without a debugger. Consider returning a richer result from `TryReadIdFromJsonObject` (enum or out param) so the thrown message is specific. QoL only.

5. **Test coverage gaps.** The data-driven `InvalidBodyId_ThrowsAtCommit` covers missing/empty/whitespace `id`. Worth locking down with additional rows / cases:
   - `{"id": 42}` — numeric id (non-string)
   - `{"id": null}` — null id
   - `{"id": ["x"]}` and `{"id": {}}` — non-string id
   - `[1, 2, 3]` — top-level array, not object
   - `{"foo": 1, "id": "x"}` — **id NOT first property** (the most important — exercises the loop's `Skip` path that the current tests don't)
   - Malformed JSON (e.g., truncated `{"id": "x"` — exercises the swallowed `JsonException` path)

   The implementation handles all of these correctly today, but tests would lock the contract.

### 🟢 Nits

- `DistributedTransactionOperation` base ctor still takes `string id` even though Create/Upsert paths now overwrite it (typed wrapper passes `id: null`, stream path relies on default value). Vestigial — fine to leave for this PR, but worth a follow-up cleanup.
- `Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransaction.cs` no longer ends with a trailing newline (`\ No newline at end of file` in the diff) — tiny style nit.

### 👍 Positives

- Excellent edge-case test `UpsertItemStream_NestedIdIgnored` proves only top-level `id` matches
- `CreateAndUpsert_IdlessOverloads_IdExtractedFromBody` exercises end-to-end serialization
- All 18+ call sites in existing tests updated consistently
- PR description is detailed with diff-hunk anchors and explicitly calls out the API/behavior change
- Internal-only surface, no changelog entry needed (consistent with team convention)

kushagraThapar (approved):

Meghana-Palaparthi:
Discussed with @kirankumarkolli and @FabianMeiswinkel and decided to close this PR - Distributed transactions will always require users to provide `id` in `CreateItem` and `UpsertItem` operations. This also prevents the overhead of having to extract id from resource body

PR #5856, opened 2026-05-12
[Internal] DTS: Adds partitionKeyRangeId to operation result to assemble valid session tokens in DTS

Meghana-Palaparthi:
### Problem

Distributed transaction (DTC) operation-level session tokens are sent by the coordinator as **raw LSN vectors** (e.g., `"-1#425344#1=12345"`) without the `{pkRangeId}:` prefix. `SessionContainer.SetSessionToken()` unconditionally does `Split(':')[1]` on the token — an LSN-only token has no colon, so `tokenParts` has length 1, causing an **`IndexOutOfRangeException`** that escapes and surfaces a committed transaction as failed to the caller.

### Fix

Add a `partitionKeyRangeId` field to `DistributedTransactionOperationResult` and assemble the canonical `{pkRangeId}:{lsn}` session token in `FromJson()` before it reaches `SessionContainer`.

### Design Decisions

1. **Assembly happens in `FromJson()`, not `MergeSessionTokens()`**
   The token is assembled at parse time so that `result.SessionToken` is always in canonical form (or null) by the time any downstream code reads it. This keeps `MergeSessionTokens` simple — it just sets the header.

2. **Warn + null (skip merge) when `partitionKeyRangeId` is absent/blank — no throw**
   The DTC coordinator currently **never sends `partitionKeyRangeId`** (tracked by [#5857](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5857)). Once the coordinator update ships, both fields will always be present and the null-path will be removed.

   We considered throwing `InvalidOperationException` on blank `partitionKeyRangeId`, but rejected it because:
   - The server-side `SetPartitionKeyRangeId()` has **zero validation** — it can legitimately be null or empty (e.g., master resource operations, partition startup, aborted transactions where `SDKResponseBuilder.cpp:193` explicitly sets empty string).
   - All non-DTC SDK modes handle null/empty pkRangeId **without throwing**: Gateway mode omits the header entirely; Direct mode falls back to the request's token prefix or hardcodes `"0"`.
   - Throwing would transform the exact crash class this PR fixes (committed tx → exception) into a different but equally bad failure.

3. **Known limitation: stale-read window when pkRangeId is absent**
   When `partitionKeyRangeId` is absent (current coordinator behavior), `SessionToken` is set to null → session token merge is skipped → `SessionContainer` retains the previous LSN for that partition. A subsequent session-consistent read may be served by a replica that hasn't yet replicated the DTC commit. This is a **strict improvement** over the pre-fix crash (committed tx surfacing as failed) and is transient — replication lag is typically milliseconds, and any other write to the same partition advances the token past the gap. This will be fully resolved when the coordinator always sends `partitionKeyRangeId` ([#5857](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5857)).

### Changes

| File | Change |
|---|---|
| `DistributedTransactionOperationResult.cs` | Add `PartitionKeyRangeId` JSON property. `FromJson()` assembles `{pkRangeId}:{lsn}` when valid, else warns and sets `SessionToken = null`. Removed unused `using System.ComponentModel`. |
| `DistributedTransactionSerializer.cs` | Add `PartitionKeyRangeId` constant for the JSON field name. |
| `DistributedTransactionCommitter.cs` | Widen try/catch to wrap entire per-operation body including `Operations[result.Index]`. Improve catch diagnostics. Token is already canonical from `FromJson()` — `MergeSessionTokens` just sets the header. |
| `DistributedTransactionResponseTests.cs` | Add 5 parser tests: pkRangeId absent, blank (empty/whitespace ×3 via DataRow), sessionToken field absent. |
| `DistributedTransactionCommitterTests.cs` | Add end-to-end committer tests: blank pkRangeId → commit succeeds, session container not updated. |
| `DistributedTransactionTests.cs` (emulator) | Update `ValidateSessionTokenMergedIntoDtcClient` to use new wire format. Add `ValidateSessionTokenSkipped_WhenPartitionKeyRangeIdAbsent`. Both tests use matching PKs for partition alignment. |

### Wire Contract (Transitional)

```
Current coordinator behavior:
  sessionToken:        always sent (raw LSN vector, e.g., "-1#425344#1=12345")
  partitionKeyRangeId: never sent

Future coordinator behavior (issue #5857):
  sessionToken:        always sent (raw LSN vector)
  partitionKeyRangeId: always sent (e.g., "3")

SDK assembly:
  Both present + valid  → "{pkRangeId}:{lsn}" (canonical)
  pkRangeId absent/blank → warn + null (skip merge)
  sessionToken absent    → no assembly needed (no merge)
```

## Type of change

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
@@ -204,12 +204,6 @@ namespace Microsoft.Azure.Cosmos
                         continue;
                     }
 
-                    if (result.StatusCode == HttpStatusCode.NotFound
-                        && result.SubStatusCode == SubStatusCodes.ReadSessionNotAvailable)
-                    {
-                        continue;
-                    }
-
                     // SessionToken is already in canonical {pkRangeId}:{lsn} format, assembled by FromJson.
                     // Note: each SetSessionToken call acquires a write lock on the SessionContainer.
                     // For a future optimization, consider a batch-update API on ISessionContainer to
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
@@ -193,10 +193,6 @@ namespace Microsoft.Azure.Cosmos
             for (int i = 0; i < response.Count; i++)
             {
                 DistributedTransactionOperationResult result = response[i];
-                if (result == null)
-                {
-                    continue;
-                }
 
                 DistributedTransactionOperation operation = null;
                 try
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
@@ -181,10 +181,8 @@ namespace Microsoft.Azure.Cosmos
             // so that subsequent Session-consistency reads on the affected collections can use the latest token
             // without getting ReadSessionNotAvailable.
             //
-            // DTC spans multiple collections so the server embeds per-operation session
-            // tokens in the JSON body; those are already parsed into DistributedTransactionOperationResult.SessionToken,
-            // but we must explicitly push them into the SessionContainer.
-
+            // DTC spans multiple collections so the server embeds per-operation session tokens in the JSON body.
+            // DistributedTransactionOperationResult.FromJson assembles each token into canonical {pkRangeId}:{lsn} form.
             if (response == null || response.Count == 0 || serverRequest == null || sessionContainer == null)
             {
                 return;
@@ -195,49 +193,45 @@ namespace Microsoft.Azure.Cosmos
             for (int i = 0; i < response.Count; i++)
             {
                 DistributedTransactionOperationResult result = response[i];
-                DistributedTransactionOperation operation = serverRequest.Operations[result.Index];
-
-                if (string.IsNullOrEmpty(result.SessionToken) || string.IsNullOrEmpty(operation.CollectionResourceId))
+                DistributedTransactionOperation operation = null;
+                try
                 {
-                    continue;
-                }
+                    operation = serverRequest.Operations[result.Index];
 
-                if (result.StatusCode == HttpStatusCode.NotFound
-                    && result.SubStatusCode == SubStatusCodes.ReadSessionNotAvailable)
-                {
-                    continue;
-                }
-
-                // SessionContainer.SetSessionToken expects the token in {pkRangeId}:{lsn} format.
-                // The backend may send pkRangeId separately or the token may already be assembled.
-                string tokenForHeader = result.SessionToken;
-                if (tokenForHeader.IndexOf(':') < 0)
-                {
-                    // Token is LSN-only; we need pkRangeId to assemble the full format.
-                    if (string.IsNullOrWhiteSpace(result.PartitionKeyRangeId))
+                    if (string.IsNullOrEmpty(result.SessionToken) || string.IsNullOrEmpty(operation.CollectionResourceId))
                     {
-                        // Cannot form a valid session token without pkRangeId; silently skip merging
-                        // this operation's token but continue processing the rest of the response.
-                        DefaultTrace.TraceWarning(
-                            "DTC operation index {0} (collection {1}) returned LSN-only session token without partitionKeyRangeId; skipping session token merge.",
-                            result.Index,
-                            operation.CollectionResourceId);
                         continue;
                     }
 
-                    tokenForHeader = result.PartitionKeyRangeId + ":" + tokenForHeader;
+                    if (result.StatusCode == HttpStatusCode.NotFound
+                        && result.SubStatusCode == SubStatusCodes.ReadSessionNotAvailable)
+                    {
+                        continue;
+                    }
+
+                    // SessionToken is already in canonical {pkRangeId}:{lsn} format, assembled by FromJson.
+                    // Note: each SetSessionToken call acquires a write lock on the SessionContainer.
+                    // For a future optimization, consider a batch-update API on ISessionContainer to
+                    // reduce lock acquisitions when multiple operations target the same collection.
+                    headers.Clear();
+                    headers[HttpConstants.HttpHeaders.SessionToken] = result.SessionToken;
+
+                    sessionContainer.SetSessionToken(
+                        operation.CollectionResourceId,
+                        DistributedTransactionConstants.GetCollectionFullName(operation.Database, operation.Container),
+                        headers);
                 }
-
-                // Note: each SetSessionToken call acquires a write lock on the SessionContainer.
-                // For a future optimization, consider a batch-update API on ISessionContainer to
-                // reduce lock acquisitions when multiple operations target the same collection.
-                headers.Clear();
-                headers[HttpConstants.HttpHeaders.SessionToken] = tokenForHeader;
-
-                sessionContainer.SetSessionToken(
-                    operation.CollectionResourceId,
-                    DistributedTransactionConstants.GetCollectionFullName(operation.Database, operation.Container),
-                    headers);
+                catch (Exception ex)
+                {
+                    // Session-token bookkeeping must never fail a transaction the server already committed.
+                    // Log and continue so the remaining operations' tokens are still attempted.
+                    DefaultTrace.TraceWarning(
+                        "DTC session token merge failed for operation index {0} (collection {1}): [{2}] {3}",
+                        result.Index,
+                        operation?.CollectionResourceId ?? "<unknown>",
+                        ex.GetType().Name,
+                        ex.Message);
+                }
             }
         }
     }
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
@@ -208,11 +208,31 @@ namespace Microsoft.Azure.Cosmos
                     continue;
                 }
 
+                // SessionContainer.SetSessionToken expects the token in {pkRangeId}:{lsn} format.
+                // The backend may send pkRangeId separately or the token may already be assembled.
+                string tokenForHeader = result.SessionToken;
+                if (tokenForHeader.IndexOf(':') < 0)
+                {
+                    // Token is LSN-only; we need pkRangeId to assemble the full format.
+                    if (string.IsNullOrWhiteSpace(result.PartitionKeyRangeId))
+                    {
+                        // Cannot form a valid session token without pkRangeId; silently skip merging
+                        // this operation's token but continue processing the rest of the response.
+                        DefaultTrace.TraceWarning(
+                            "DTC operation index {0} (collection {1}) returned LSN-only session token without partitionKeyRangeId; skipping session token merge.",
+                            result.Index,
+                            operation.CollectionResourceId);
+                        continue;
+                    }
+
+                    tokenForHeader = result.PartitionKeyRangeId + ":" + tokenForHeader;
+                }
+
                 // Note: each SetSessionToken call acquires a write lock on the SessionContainer.
                 // For a future optimization, consider a batch-update API on ISessionContainer to
                 // reduce lock acquisitions when multiple operations target the same collection.
                 headers.Clear();
-                headers[HttpConstants.HttpHeaders.SessionToken] = result.SessionToken;
+                headers[HttpConstants.HttpHeaders.SessionToken] = tokenForHeader;
 
                 sessionContainer.SetSessionToken(
                     operation.CollectionResourceId,
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
@@ -193,6 +193,11 @@ namespace Microsoft.Azure.Cosmos
             for (int i = 0; i < response.Count; i++)
             {
                 DistributedTransactionOperationResult result = response[i];
+                if (result == null)
+                {
+                    continue;
+                }
+
                 DistributedTransactionOperation operation = null;
                 try
                 {
@@ -221,7 +226,7 @@ namespace Microsoft.Azure.Cosmos
                         DistributedTransactionConstants.GetCollectionFullName(operation.Database, operation.Container),
                         headers);
                 }
-                catch (Exception ex)
+                catch (Exception ex) when (ex is not OperationCanceledException)
                 {
                     // Session-token bookkeeping must never fail a transaction the server already committed.
                     // Log and continue so the remaining operations' tokens are still attempted.
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Azure.Cosmos
     using System.Net;
     using System.Text.Json;
     using System.Text.Json.Serialization;
+    using Microsoft.Azure.Cosmos.Core.Trace;
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Documents;
 
@@ -91,9 +92,7 @@ namespace Microsoft.Azure.Cosmos
         public virtual string SessionToken { get; internal set; }
 
         /// <summary>
-        /// Gets the partition key range ID associated with the operation result.
-        /// When present, it is combined with <see cref="SessionToken"/> to form the
-        /// full session token in the format {partitionKeyRangeId}:{lsn}.
+        /// Gets the raw partition key range ID emitted by the server.
         /// </summary>
         [JsonInclude]
         [JsonPropertyName("partitionKeyRangeId")]
@@ -145,7 +144,7 @@ namespace Microsoft.Azure.Cosmos
         /// Creates a <see cref="DistributedTransactionOperationResult"/> from a JSON element.
         /// </summary>
         /// <param name="json">The JSON element containing the operation result.</param>
-        /// <returns>The deserialized operation result.</returns>
+        /// <returns>The deserialized operation result with a canonical session token.</returns>
         internal static DistributedTransactionOperationResult FromJson(JsonElement json)
         {
             DistributedTransactionOperationResult result = JsonSerializer.Deserialize<DistributedTransactionOperationResult>(json, DistributedTransactionOperationResult.CaseInsensitiveOptions);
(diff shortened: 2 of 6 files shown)


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -208,11 +208,31 @@ internal static void MergeSessionTokens(
...
                 }
 
+                // SessionContainer.SetSessionToken expects the token in {pkRangeId}:{lsn} format.
+                // The backend may send pkRangeId separately or the token may already be assembled.
+                string tokenForHeader = result.SessionToken;
+                if (tokenForHeader.IndexOf(':') < 0)
+                {
+                    // Token is LSN-only; we need pkRangeId to assemble the full format.
+                    if (string.IsNullOrWhiteSpace(result.PartitionKeyRangeId))
+                    {
```
FabianMeiswinkel: Why this hack? If the PkRangeId is missing it is an invalid session token? Should be fixed instead in the service!

FabianMeiswinkel (changes requested):
SDK should not attempt to fix broken sesison tokens.

NaluTripician:
## Deep Review (local) — findings

Sharing findings from a local deep review. Not posting as a formal "Request Changes" because @FabianMeiswinkel already has `CHANGES_REQUESTED` on file — these notes are meant to feed into that iteration.

### Summary

This correctly identifies a real latent crash from #5705: `SessionContainer.SetSessionToken` does `tokenParts[1]` unconditionally, so any LSN-only token coming back in a `CommitDistributedTransaction` response throws `IndexOutOfRangeException` out of `MergeSessionTokens` → bubbles to `CommitTransactionAsync`'s outer `catch` → **the SDK reports a committed DTS as failed.** Good find.

However: the architectural direction is contested, and the workaround is incomplete — other malformed shapes will still convert a committed DTS into a thrown exception.

---

### 🔴 Blocker

**B1 — Architectural pushback unresolved.** Aligning with @FabianMeiswinkel: the session token shape is a server-emitted contract; the right fix is server-side, not a parsing escape hatch in the SDK. Once dual-shape parsing ships, the SDK has to tolerate both shapes indefinitely for interop with older clients, which effectively makes the "transitional" code permanent. PR #5705's earlier review from @kirankumarkolli made the same point ("Validations right place is upstream when response is composed.").

If a client-side patch must land as a stop-gap, gate it behind a feature flag / version check with a tracking issue and an explicit deletion target.

---

### 🟡 Major

**M2 — The fix is partial.** `SessionContainer.SetSessionToken` (private overload, `SessionContainer.cs:306-321`) still throws on:
- `":"` → `tokenParts = ["", ""]`, `SessionTokenHelper.Parse("")` throws
- `"0:"` → same parse failure
- `"abc:def"` (non-numeric LSN) → `FormatException`
- `result.PartitionKeyRangeId` containing a colon → produces `"0:abc:lsn..."`, then `tokenParts[1] = "abc"` fails to parse

Each of these still escapes `MergeSessionTokens` → caught by `CommitTransactionAsync`'s `catch (Exception ex) when (ex is not OperationCanceledException)` (line 68) → **server-committed DTS surfaces as a thrown exception to the caller.** That's exactly the failure mode this PR is trying to fix; it just plugs one of several holes.

Suggestion: wrap the inner merge loop body (or the `SetSessionToken` call) in `try/catch`-and-trace. Nothing in post-commit session-token bookkeeping should be able to fail a transaction that the server already committed.

**M3 — Silent skip creates a mystery 404/1002 for customers.** When pkRangeId is missing, the SessionContainer never learns the LSN advance. The next session-consistency read against a replica that hasn't caught up returns `ReadSessionNotAvailable`. From the customer's perspective: "I committed a transaction, then immediately read and got a 404." The only signal of the real cause is a `DefaultTrace.TraceWarning` on the writing client, with no link to the response the customer holds.

Consider attaching a diagnostic note onto `DistributedTransactionResponse` so the skip is visible on the result the caller already has.

**M4 — No operational signal for the Cosmos team.** `DefaultTrace` is invisible in most deployments. The service team can't tell if this branch fires zero times per day or millions. That's exactly the kind of "compensating client code that ages badly" — no metric ever lights up, so no one notices the server was never fixed.

Suggestion: route this through `CosmosDbEventSource` and/or a counter (e.g. `DtsMalformedSessionTokenCount`) so it shows up in standard ETW ingestion. Add a `// TODO(issue#…)` deletion gate.

**M5 — Comment vs. behavior contradict each other.** The block at `DistributedTransactionCommitter.cs:212-222` reads:

```csharp
// Cannot form a valid session token without pkRangeId; silently skip merging
// this operation's token but continue processing the rest of the response.
DefaultTrace.TraceWarning(...);
```

The comment (and PR description) say "silently skip"; the code emits a warning. Pick one — given M4 I'd keep and upgrade the trace, then reword the comment.

---

### 🟢 Minor

- **m6 — `IndexOf(':') < 0` is a weak shape heuristic.** `":"` and `"0:"` slip through as "pre-assembled". A stricter check (both halves non-empty after split) or a structured assembly from the two fields is more defensible.
- **m7 — Trace-log spam risk.** If the server regresses universally, every op of every transaction emits a warning. Sample, or fold into the counter from M4.
- **m8 — Test gap: multi-operation mixed-shape responses.** All five new tests use a single operation. The `continue` path was added so the loop keeps processing surrounding operations, but no test exercises that (e.g. op 0 pre-assembled, op 1 LSN+pkRangeId, op 2 LSN-only-missing-pkRangeId → assert ops 0 and 1 still merge correctly).
- **m9 — Tests don't assert the `TraceWarning`.** Given M3/M4 depend on that diagnostic, the trace is behavior, not incidental. Subscribe a listener and lock it down.
- **m10 — `PartitionKeyRangeId` becomes part of the `INTERNAL`-public surface.** It also serializes outbound via `[JsonInclude]` + public getter. If this is provisional, mark it `[JsonIgnore]` for serialization and XML-doc it as transitional so consumers don't take a dependency.

---

### 💬 Suggestions

- **s11** — Prefer `.Contains(':', StringComparison.Ordinal)` over `IndexOf(':') < 0`.
- **s12** — `// TODO(issue#NNNN): Remove once the gateway emits {pkRangeId}:{lsn}` deletion gate.
- **s13** — Per @kirankumarkolli's prior guidance on #5705, do the assembly in `DistributedTransactionOperationResult.FromJson` (where both fields are right there) and produce a canonical `{pkRangeId}:{lsn}` `SessionToken` to the rest of the SDK. That keeps the transitional knowledge in one place and removes the need to expose `PartitionKeyRangeId` separately at all (resolving m10).

---

### ✅ Positives

- The crash analysis on `SessionContainer.SetSessionToken`'s `tokenParts[1]` is non-obvious and a real find.
- Test-helper extension (4-tuple overload alongside the 2/3-tuple shims) keeps 30+ existing tests working unchanged — clean refactor.
- Deliberate edge-case coverage on the new branch (`""`, `" "`, `"   "`, and the precedence test for pkRangeId present *and* token already containing `:`).
- Copy constructor correctly updated to propagate `PartitionKeyRangeId` — easy to miss.
- Trace message includes operation index and collection rid — good diagnostic context if anyone does read the trace.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -208,11 +208,31 @@ internal static void MergeSessionTokens(
                     continue;
                 }
 
+                // SessionContainer.SetSessionToken expects the token in {pkRangeId}:{lsn} format.
+                // The backend may send pkRangeId separately or the token may already be assembled.
+                string tokenForHeader = result.SessionToken;
```
jeet1995: For debugging purposes, what do we do if SessionToken is malformed?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
Meghana-Palaparthi: Coordinator still hasn't made a change to always send pkRangeId. Once that change is made, I'll remove this check and require it in SDK

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
Meghana-Palaparthi: I have wrapped it in a try block, which would log if it encountered a malformed session token and continue

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -195,29 +193,45 @@ internal static void MergeSessionTokens(
...
+                    // reduce lock acquisitions when multiple operations target the same collection.
+                    headers.Clear();
+                    headers[HttpConstants.HttpHeaders.SessionToken] = result.SessionToken;
+
+                    sessionContainer.SetSessionToken(
+                        operation.CollectionResourceId,
+                        DistributedTransactionConstants.GetCollectionFullName(operation.Database, operation.Container),
+                        headers);
+                }
+                catch (Exception ex)
```
NaluTripician: `catch (Exception ex)` here swallows `OperationCanceledException`. The outer catch in `CommitTransactionAsync` deliberately uses `when (ex is not OperationCanceledException)` so cancellations propagate; this inner catch breaks that contract.

Even though `ISessionContainer.SetSessionToken` doesn't take a `CancellationToken` today, anyone wiring one through later will have OCEs silently downgraded to a `TraceWarning`. Please mirror the outer pattern:

```csharp
catch (Exception ex) when (ex is not OperationCanceledException)
```

Or, more disciplined, catch only the expected types (`IndexOutOfRangeException`, `FormatException`, `ArgumentException`, `InvalidOperationException`) so an unrelated bug elsewhere doesn't get masked.

A paired test that throws `OperationCanceledException` from the mock `SetSessionToken` and asserts it propagates (sibling to `CommitTransactionAsync_SwallowsSetSessionTokenException`) would lock the contract down.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -195,29 +193,45 @@ internal static void MergeSessionTokens(
...
+                {
+                    // Session-token bookkeeping must never fail a transaction the server already committed.
+                    // Log and continue so the remaining operations' tokens are still attempted.
+                    DefaultTrace.TraceWarning(
+                        "DTC session token merge failed for operation index {0} (collection {1}): [{2}] {3}",
+                        result.Index,
```
NaluTripician, line 225: `result.Index` is dereferenced without a null-guard on `result`. If `response[i]` is ever `null` (deserializer quirk on a malformed payload, or an explicit `null` element in the JSON array), the assignment on line 195 succeeds, the try body's first line NREs on `result.Index`, then this catch's format-args path NREs *again* on `result.Index` — which escapes `MergeSessionTokens` and defeats the safety net you just built.

You already use `operation?.CollectionResourceId ?? "<unknown>"` defensively on the next line — please mirror that here:

```csharp
result?.Index ?? -1,
```

Ideally with a `continue` if `result` is null at the top of the loop body so the catch path doesn't get hit for that case at all.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -154,6 +163,22 @@ internal static DistributedTransactionOperationResult FromJson(JsonElement json)
                 result.ResourceStream = new MemoryStream(bytes, 0, bytes.Length, writable: false, publiclyVisible: true);
             }
 
+            if (!string.IsNullOrEmpty(result.SessionToken))
```
NaluTripician: Asymmetric null-checks: `!string.IsNullOrEmpty(result.SessionToken)` (here) vs `!string.IsNullOrWhiteSpace(result.PartitionKeyRangeId)` (line 168).

The outer guard admits a whitespace `sessionToken` like `"   "`. With a valid pkRangeId `"0"` you'd assemble `"0:   "` — has a colon, so it gets past `MergeSessionTokens`' `IsNullOrEmpty` check too, reaches `SessionContainer.SetSessionToken`, and `SessionTokenHelper.Parse("   ")` throws. The new try/catch will swallow it — but the assembly is doing actively wrong work and we lose any signal that the wire payload is malformed.

Please make both checks `IsNullOrWhiteSpace`:

```csharp
if (!string.IsNullOrWhiteSpace(result.SessionToken))
```

And add a `DataRow` in `DistributedTransactionResponseTests` for whitespace `sessionToken` symmetric to the existing whitespace-pkRangeId rows.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -154,6 +163,22 @@ internal static DistributedTransactionOperationResult FromJson(JsonElement json)
...
+            {
+                if (!string.IsNullOrWhiteSpace(result.PartitionKeyRangeId))
+                {
+                    result.SessionToken = result.PartitionKeyRangeId + ":" + result.SessionToken;
```
NaluTripician, line 176: This branch is load-bearing on the assumption that **every** deployed coordinator emits LSN-only `sessionToken` (per the PR description). If any coordinator version — current or in-flight — still emits the canonical `{pkRangeId}:{lsn}` shape directly in `sessionToken` (and no separate `partitionKeyRangeId` field), the `else` on line 172 silently nulls a valid token → SessionContainer never advances → next session-consistent read against a lagging replica returns `ReadSessionNotAvailable`. That's a silent regression on a path that previously worked.

Please add a defensive "already canonical" pass-through so the SDK tolerates both wire shapes:

```csharp
if (result.SessionToken.IndexOf(':') >= 0)
{
    // Already canonical {pkRangeId}:{lsn} — leave as-is.
}
else if (!string.IsNullOrWhiteSpace(result.PartitionKeyRangeId))
{
    result.SessionToken = result.PartitionKeyRangeId + ":" + result.SessionToken;
}
else
{
    DefaultTrace.TraceWarning(...);
    result.SessionToken = null;
}
```

Alternatively, link an explicit written confirmation from the service team in the PR description that no deployed coordinator (any region/SKU) emits canonical-prefixed tokens in this field. The one-line passthrough is cheaper.

A unit test for `sessionToken="0:12345"` + `partitionKeyRangeId=null` asserting the token survives untouched would cover this.

NaluTripician (changes requested):
Following up on my earlier deep-review notes — the crash fix and overall shape look right, but I'd like these specific code asks resolved before I sign off. None are large; all are inline below.

**Code asks (gating my sign-off):**
1. Narrow the new outer `catch (Exception)` in `MergeSessionTokens` to exclude `OperationCanceledException` so it matches `CommitTransactionAsync`'s outer-catch contract.
2. Null-guard `result` in the new catch's `TraceWarning` format-args path so a null array element can't NRE *inside* the safety net and escape the loop.
3. Make the SessionToken / PartitionKeyRangeId null-checks in `FromJson` symmetric (both `IsNullOrWhiteSpace`).
4. Add an "already canonical" pass-through in `FromJson` so the SDK is defensive against any coordinator version that still emits `{pkRangeId}:{lsn}` directly in `sessionToken` — silently nulling a valid token is a worse regression than the original crash.

**Process gates (not code, but worth calling out):**
- @FabianMeiswinkel still has `CHANGES_REQUESTED` on file — please get an explicit re-review / dismissal rather than relying on the text reply.
- CI on `6fb31fb` is still in-progress for most `dotnet-v3-ci` jobs (Microsoft.Azure.Cosmos.Tests, Static Analysis, EmulatorTests suites, packaging). Want to see those green before I approve.

Once the four inline items are addressed and CI/Fabian unblock, this is good to go from my side.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
Meghana-Palaparthi: That's a valid concern. I'll include a null check within the `FromJson` method in `DistributedTransactionOperationResult` rather than adding it downstream here.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
Meghana-Palaparthi: The current contract states that DTC will only send the raw LSN value in sessionToken. Even for existing non-DTX flows, the backend service does not send a canonical session token. If, for DTX, the backend starts sending a canonical `{pkRangeId}:{lsn}` session token, it would be violating the contract. I can still add a check to handle this defensively; in case it ever happens.

NaluTripician (approved):
## Approve with nits ✅

Thanks for the iterations — ~80% of my prior asks are addressed, the crash fix is real, and the test coverage (parser-level + committer-level + emulator + `DelegatingTraceListener` for the trace contract) is genuinely thorough. Signing off; Fabian is updating his review asynchronously so I'm not gating on that here.

Nits below are non-blocking — happy to take any/all as follow-ups.

### Nits

1. **`result?.Index` in the catch's format args.** `FromJson`'s `?? throw new JsonException(...)` guard closes the realistic path, but the catch at line 224 is the last line of defense against bookkeeping bugs. A null-conditional access (`result?.Index ?? -1`) costs nothing and keeps the warning robust if anything upstream regresses.

2. **404 / `ReadSessionNotAvailable` skip removal in `6708805` is undefended.** The rationale ("unreachable" because DTC ops are writes) is plausible, but the deletion has no inline comment and the prior test `CommitTransactionAsync_SkipsMerge_When404ReadSessionNotAvailable` was removed rather than replaced. Either restore the skip (cheap defensive depth) or add a 2-line comment + a regression test asserting `MergeSessionTokens` no-ops gracefully on a per-op 404/1002 — relies on the new try/catch.

3. **Edge-colon assembly produces double-colon tokens.** When `pkRangeId="X"` and `sessionToken=":Y"`, the assembled output is `"X::Y"`. The new try/catch in `MergeSessionTokens` turns the downstream throw into a `TraceWarning`, but `FromResponseMessage_OperationResult_SessionToken_AssembledWhenColonIsAtEdge` actively *locks in* the malformed output as intended behavior. Suggest only assembling when the raw token contains no colon at all — and update that test to assert the token is nulled / skipped instead.

4. **TODO deletion gates for the transitional null-pkRangeId branch are only in tests, not at the `FromJson` branch sites.** When the coordinator update ships, the person ripping this out has to discover the three FromJson branches by hand. A `// TODO(#NNNN, delete after coordinator emits pkRangeId)` next to each branch would make that work mechanical. Pair with a tracking issue.

5. **`TraceWarning` spam risk is unbounded.** If the coordinator never updates for a given account, every op of every DTS commit logs. Not urgent, but worth a sampling guard or first-N-per-process throttle before this is in customers' hands at scale.

6. **Brittle trace assertion.** `CommitTransactionAsync_EmitsTraceWarning_WhenPartitionKeyRangeIdIsAbsent` matches on `.Contains("partitionKeyRangeId")` — if anyone reformats the warning and drops that exact substring the test goes silently green. Consider asserting on the operation index or a more stable token like `"DTC operation index"`.

7. **Customer/operational diagnostic gap (M3/M4 from my prior review).** `DefaultTrace.TraceWarning` is invisible in nearly all customer deployments and gives the Cosmos team no production telemetry to size the urgency of the coordinator fix. Even one of these as a follow-up would help materially: a `SessionMergeSkipped` flag on `DistributedTransactionResponse`, or routing the warning through `CosmosDbEventSource` so it's observable.

### Positives worth saying out loud
- `OperationCanceledException` carve-out + the new test that locks the contract is exactly right.
- Already-canonical pass-through branch + the `colonIndex > 0 && colonIndex < length - 1` guard is a clean way to reject `":"` / `"0:"` without breaking real tokens.
- `IsNullOrWhiteSpace` symmetry + the whitespace-to-null normalization branch and the parameterized tests covering `" "`, `"   "` are nice.
- Copy constructor propagating `PartitionKeyRangeId` — easy to miss.
- Moving the `Operations[result.Index]` access inside the try/catch is a genuine hardening fix on top of the headline change.
- `?? throw new JsonException` on `FromJson`'s `Deserialize` result quietly turns "NRE somewhere later" into a boundary-level exception.

LGTM 🚢

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -195,29 +193,40 @@ internal static void MergeSessionTokens(
...
+                    // Note: each SetSessionToken call acquires a write lock on the SessionContainer.
+                    // For a future optimization, consider a batch-update API on ISessionContainer to
+                    // reduce lock acquisitions when multiple operations target the same collection.
+                    headers.Clear();
+                    headers[HttpConstants.HttpHeaders.SessionToken] = result.SessionToken;
+
+                    sessionContainer.SetSessionToken(
+                        operation.CollectionResourceId,
+                        DistributedTransactionConstants.GetCollectionFullName(operation.Database, operation.Container),
+                        headers);
```
FabianMeiswinkel, line 217: This is quite dangerous - teh headers collection is mutating in every iteration of teh for loop - it is safe as long as SetSessionToken is only usign the Collection while the method is executed and not maintaining any reference - but there could be subtle race conditions if that behaviro changes and  contracts are not clearly called out.

FabianMeiswinkel: It seems to be safe here right now thought - it only accesses the collection to grab certain headers during the method call - not keeping a reference

FabianMeiswinkel (approved):
LGTM

PR #5798, opened 2026-04-21
Encryption Pipelines: Adds SDK-surface parity build to catch TypeLoad gaps

NaluTripician:
## Problem

The Encryption and Encryption.Custom official release pipelines previously built each project only against the pinned `Microsoft.Azure.Cosmos` NuGet declared in the project file. That does not detect abstract members added to `Container` on master that lack an unconditional override in `EncryptionContainer` — the exact class of defect fixed by #5783 (`SemanticRerankAsync` `TypeLoadException`).

The customer-facing symptom from #5783:

> `System.TypeLoadException`: Method `SemanticRerankAsync` in type `Microsoft.Azure.Cosmos.Encryption.Custom.EncryptionContainer` does not have an implementation.

## Why a naive SDKREF build is not enough

The override was gated `#if PREVIEW && SDKPROJECTREF`. Any build that defines **both** symbols — which is what the existing PR-validation build (`build-preview.yml`) does, and what an obvious "build with `SdkProjectRef=true`" addition would also do — compiles the override back in and the missing-implementation condition never manifests:

| Build mode | `PREVIEW` | `SDKPROJECTREF` | Override present? | Catches the bug? |
|---|---|---|---|---|
| Current default (NuGet ref, pinned `3.41.0-preview.0`) | ❌ | ❌ | — (no abstract either) | ❌ |
| `-p:IsPreview=true -p:SdkProjectRef=true` | ✅ | ✅ | ✅ | ❌ (masks the defect) |
| **New parity build** (`DefineSdkProjectRefSymbol=false`) | ✅ | ❌ | ❌ | ✅ |

The bug only manifests when `PREVIEW` is on and `SDKPROJECTREF` is off, linked against an SDK that contains the new abstract member. That is the exact surface of the shipped Encryption.Custom NuGet.

## Change

1. **Decouple the `SDKPROJECTREF` preprocessor define from the `SdkProjectRef` MSBuild property** via a new `DefineSdkProjectRefSymbol` knob in both:
   - `Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj`
   - `Microsoft.Azure.Cosmos.Encryption/src/Microsoft.Azure.Cosmos.Encryption.csproj`

   Default behaviour is unchanged: `SdkProjectRef=true` still sets `SDKPROJECTREF`. Pipelines can now opt out with `-p:DefineSdkProjectRefSymbol=false`.

[...]

PR #5861, opened 2026-05-13
Pipelines: Adds Encryption SDK GA + Preview parity to PR, nightly, and rolling

NaluTripician:
## Problem

PR #5798 added the Encryption / Encryption.Custom SDK-source project-ref + NuGet-surface parity builds to `templates/build-preview.yml` so the PR-validation pipeline catches any newly-added abstract member on `Container` under `#if PREVIEW` that lacks an unconditional override in `EncryptionContainer` (the class of defect behind #5783).

Those parity steps have **two gaps**:

### Gap 1 — They do not run on nightly or rolling

| Pipeline | File | Currently runs `build-preview.yml`? |
|---|---|---|
| PR validation | `azure-pipelines.yml` | ✅ (added in #5798) |
| Nightly | `azure-pipelines-nightly.yml` | ❌ |
| Rolling | `azure-pipelines-rolling.yml` | ❌ |

So a defect introduced via a path that bypasses PR validation (force-merge, retroactive SDK change, build-config drift) would not be caught until the next encryption release-gate run.

### Gap 2 — The Encryption parity build does not cover the GA NuGet surface

PR #5798's parity step always sets `-p:IsPreview=true`. Per `Directory.Build.props:16`, that defines both `PREVIEW` **and** `ENCRYPTIONPREVIEW`. So `#if ENCRYPTIONPREVIEW`-gated overrides in `EncryptionContainer` get compiled in — even though they do **not** ship in the GA NuGet of `Microsoft.Azure.Cosmos.Encryption` (which uses package reference `[3.51.0,)` when `IsPreview` is unset).

**Concrete miss — PR #5825**: that PR promotes `Container.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>` from `#if PREVIEW` to GA-unconditional. The Encryption override is gated `#if ENCRYPTIONPREVIEW || SDKPROJECTREF`. The Preview parity build defined `ENCRYPTIONPREVIEW` and went green; the author then had to push three follow-up "fix the guard" commits (`c0b56b14e`, `a0347ebb7`, `907d5a41d`) once the build broke downstream.

Reproduced at PR #5825 head (`907d5a41d`):

```
dotnet build Microsoft.Azure.Cosmos.Encryption.csproj -c Release \
  -p:SdkProjectRef=true -p:DefineSdkProjectRefSymbol=false

error CS0534: 'EncryptionContainer' does not implement inherited abstract
member 'Container.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<T>(...)'
Build FAILED.
```

The same command on `main` succeeds (0 errors), so the new step does not regress current `main`.

### Commit 1 — wire `build-preview.yml` into the off-PR pipelines

* `azure-pipelines-nightly.yml` — new `Preview parity build` stage invoking `templates/build-preview.yml`.
* `azure-pipelines-rolling.yml` — additional `templates/build-preview.yml` job alongside the existing `build-test.yml` runs.

### Commit 2 — add a GA NuGet-surface parity step for `Microsoft.Azure.Cosmos.Encryption`

Adds a second parity step alongside the existing Preview one in:

* `templates/build-preview.yml` (Encryption Project Ref SDK Preview job)
* `templates/static-tools-encryption.yml`

The new step omits `-p:IsPreview`, so neither `PREVIEW` nor `ENCRYPTIONPREVIEW` is defined — matching the exact preprocessor surface of the shipped GA NuGet — while still linking against the latest SDK source via `-p:SdkProjectRef=true`.

No GA parity step is added for `Microsoft.Azure.Cosmos.Encryption.Custom`: its csproj package-references `Microsoft.Azure.Cosmos 3.41.0-preview.0` unconditionally, so it has no GA NuGet surface to validate. A documenting note is added in both Custom templates.

## Type of change

## Closing issues

Follow-up to #5798. Related to #5783, #5825.

### DIFF
--- a/azure-pipelines-nightly.yml
+++ b/azure-pipelines-nightly.yml
@@ -47,3 +47,19 @@ stages:
         OutputPath: '$(Build.ArtifactStagingDirectory)/bin/AnyCPU/$(BuildConfiguration)/Microsoft.Azure.Cosmos'
         BlobVersion: 'nightly-preview'
         CleanupFolder: true
+
+# Run the preview-flag build (which includes the Encryption / Encryption.Custom
+# SDK-source project-ref + NuGet-surface parity builds added in PR #5798) on
+# the nightly schedule so any newly-added abstract member on Container under
+# #if PREVIEW that lacks an unconditional override in EncryptionContainer is
+# caught off-PR as well.
+- stage:
+  displayName: Preview parity build
+  variables:
+    Codeql.BuildIdentifier: preview_parity
+  jobs:
+    - template:  templates/build-preview.yml
+      parameters:
+        BuildConfiguration: Release
+        Arguments: ' '
+        VmImage: $(VmImage)
--- a/azure-pipelines-rolling.yml
+++ b/azure-pipelines-rolling.yml
@@ -44,3 +44,14 @@ jobs:
     IncludePerformance: true
     IncludeCoverage: true
 
+# Run the preview-flag build (which includes the Encryption / Encryption.Custom
+# SDK-source project-ref + NuGet-surface parity builds added in PR #5798) on
+# the rolling schedule so any newly-added abstract member on Container under
+# #if PREVIEW that lacks an unconditional override in EncryptionContainer is
+# caught off-PR as well.
+- template:  templates/build-preview.yml
+  parameters:
+    BuildConfiguration: Release
+    Arguments: ' '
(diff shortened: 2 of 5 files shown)


kushagraThapar (approved):

kundadebdatta (approved):
LGTM

PR #5868, opened 2026-05-14
Diagnostics: Adds Hedging Detection API (`HedgingStarted`, `GetRequestedRegions` and `GetRespondedRegions`)

### LINKED ISSUE #5867: [Feature] Hedging Detection API — public accessors on CosmosDiagnostics
## Summary

Add a public **Hedging Detection API** to `Microsoft.Azure.Cosmos.CosmosDiagnostics` so customers can post-hoc determine whether a successful or failed point/feed operation went through cross-region hedging, which regions were dispatched against, and which regions responded. The new surface is strictly additive, non-breaking, and is reachable from both `ItemResponse<T>.Diagnostics` (success path) and `CosmosException.Diagnostics` (error path).

This is part of a cross-SDK feature being implemented in parallel across .NET, Java, and Python (with a spec-only deliverable for Rust). See cross-SDK design doc *"Hedging Detection API - Technical Design - Final"* (2026-05) for the contract.

## Public API additions

```csharp
namespace Microsoft.Azure.Cosmos
{
    public readonly struct RequestedRegion : IEquatable<RequestedRegion>
    {
        public string RegionName { get; }
        public RequestedRegionReason Reason { get; }
        public RequestedRegion(string regionName, RequestedRegionReason reason);
        public override bool Equals(object obj);
        public bool Equals(RequestedRegion other);
        public override int GetHashCode();
        public override string ToString(); // "{regionName}:{reason}"
    }

    public enum RequestedRegionReason : byte
    {
        Initial,
        OperationRetry,
        TransportRetry,        // reserved — not populated today (SE-001: signal lives in closed-source Direct package)
        Hedging,
        RegionFailover,
        CircuitBreakerProbe,
    }

    public abstract class CosmosDiagnostics
    {
        // Existing members unchanged.
        public virtual bool HedgingStarted() => false;
        public virtual IReadOnlyList<RequestedRegion> GetRequestedRegions() => Array.Empty<RequestedRegion>();
        public virtual IReadOnlyList<string> GetRespondedRegions() => Array.Empty<string>();
    }
}
```

`CosmosTraceDiagnostics` (internal sealed) overrides all three to return live state. New state is populated at orchestrator dispatch sites (hedging strategy, retry policy, region failover, PPCB probe) and response-handling sites; appends are thread-safe via a private `object regionLock`. JSON shape of `Diagnostics.ToString()` is **unchanged** — new state is NOT serialized into the trace tree.

## Acceptance criteria (testable)

- [ ] **AC1** Single-region client, ReadItem success, no retries → `HedgingStarted() == false`; `GetRequestedRegions().Count == 1` with reason `Initial`; `GetRespondedRegions().Count == 1`.
- [ ] **AC2** Multi-region client, hedging enabled, primary responds under threshold → `HedgingStarted() == false`; `GetRequestedRegions().Count == 1` (**no phantom `Hedging` entry**); `GetRespondedRegions().Count == 1`.
- [ ] **AC3** Multi-region client, hedging enabled, primary slow, hedge arm wins → `HedgingStarted() == true`; `GetRequestedRegions()` has ≥2 entries: `(primary, Initial)` and `(hedgeRegion, Hedging)`; `GetRespondedRegions()` has ≥1 entry.
- [ ] **AC4** 410 Gone retry on same region → `GetRequestedRegions()` includes consecutive entries `(region, Initial)` then `(region, OperationRetry)`.
- [ ] **AC5** Region failover after 503 → `GetRequestedRegions()` includes `(originalRegion, Initial)` then `(secondaryRegion, RegionFailover)`.
- [ ] **AC6** PPCB probe after a region was previously circuit-broken → `GetRequestedRegions()` includes an entry with reason `CircuitBreakerProbe`. (Requires new `RequestContext` PPCB flag.)
- [ ] **AC7** All-regions-down error → `CosmosException.Diagnostics.GetRequestedRegions()` non-empty; `GetRespondedRegions()` may be empty.
- [ ] **AC8** `CosmosException.Diagnostics` exposes same data as the would-be-successful-response `Diagnostics`.
- [ ] **AC9** Backward compat — existing customer subclass of `CosmosDiagnostics` that does NOT override the new methods returns the safe defaults and does not throw.
- [ ] **AC10** Contract baselines (`DotNetSDKAPI.net6.json`, `DotNetSDKAPI.net8.json`) regenerated via `Microsoft.Azure.Cosmos\tools\UpdateContracts.ps1`; diffs committed; only the new public members appear.
- [ ] **AC11** `CosmosDiagnostics.ToString()` JSON for a known fixed diagnostic input matches the baseline byte-for-byte (golden-file test).
- [ ] **AC12** `RegionContactedInDiagnosticsBenchmark` mean latency does not regress by >5% vs baseline.
- [ ] **AC13** Hedge arm registered then cancelled **before threshold delay elapses** → `GetRequestedRegions()` does NOT contain a `Hedging` entry for the cancelled region (uses fault injection; primary completes in <50% of threshold so `Task.Delay(threshold, ct)` throws `OperationCanceledException`).
- [ ] **AC14** `GetRespondedRegions()` allows duplicates (same region producing two responses); XML doc warns that `Count > 1` does NOT imply multi-region success.
- [ ] **AC15** Reason coverage — `TransportRetry` value never appears in v1 production paths (covered by a fault-injection AC test that enumerates the union of reasons across the full matrix; documents SE-001).
- [ ] **AC16** Live multi-region smoke test (≥1) — runs against the team's existing multi-region test account with hedging enabled, injects primary-slow latency, and asserts `HedgingStarted() == true`, `GetRequestedRegions()` includes both regions, `GetRespondedRegions()` includes the secondary region.

## Files in scope

- New: `Microsoft.Azure.Cosmos\src\Diagnostics\RequestedRegion.cs`, `RequestedRegionReason.cs`
- Modify: `Microsoft.Azure.Cosmos\src\Diagnostics\CosmosDiagnostics.cs`, `CosmosTraceDiagnostics.cs`, `AvailabilityStrategy.cs` / `ParallelHedgingAvailabilityStrategy.cs`, `ClientRetryPolicy.cs`, `GlobalEndpointManager.cs`, `PartitionKeyRangeCache` response pipeline, `RequestContext` (PPCB flag), `changelog.md`
- Tests: new `RequestedRegionTests.cs`, new `HedgingDetectionEmulatorTests.cs` (under `Microsoft.Azure.Cosmos.EmulatorTests\FaultInjection\`), live-account multi-region test
- Contracts: `DotNetSDKAPI.net6.json`, `DotNetSDKAPI.net8.json` (regenerated)
- OpenSpec: refresh `openspec/changes/hedging-detection-api/` (predecessor PR #5741 left artifacts on closed branch `feature/hedging-detection-api`)

## Out of scope

- Wiring into OpenTelemetry / OTLP exporter — separate work item.
- Changing the existing JSON `ToString()` shape — explicitly preserved.
- Live wire-format changes — none; this is local diagnostics only.
- Cross-SDK parity automation — handled at the cross-SDK design layer.

## Cross-SDK companion issues

- **Java**: forthcoming companion issue in `Azure/azure-sdk-for-java` (`sdk/cosmos`).
- **Python**: forthcoming companion issue in `Azure/azure-sdk-for-python` (`sdk/cosmos`).
- **Rust**: forthcoming spec-only issue + draft PR in `Azure/azure-sdk-for-rust` (sequenced after PR #4330).

## Notes for the implementer

- Full internal spec, landscape research, plan, risk register (`side-effects.json`), questions+answers, and an `.docx` review bundle are available from the workflow author (`@NaluTripician`) on request — they are team-only and not linked here.
- Phase 1 review gate completed on 2026-05-14 with three substantive decisions: (Q3=B) Python collapsed to single Option B path; (Q5=B) Rust accepts `Retry → OperationRetry` JSON wire-format break with `#[serde(alias = "retry")]`; (Q10) at least one live multi-region smoke test required per implementing SDK.
- This issue is being dispatched to the **Coding Agent Harness** for end-to-end implementation; reviewers may receive a draft PR shortly.

NaluTripician:
# Diagnostics: Adds hedging detection API (`HedgingStarted` / `GetRequestedRegions` / `GetRespondedRegions`)

Closes #5867 (partial — covers .NET; Java and Python tracked separately).

## Summary

Adds a focused, hot-path-friendly public surface on `CosmosDiagnostics` so callers can observe per-operation hedging behavior without parsing the `ToString()` JSON:

- `virtual bool HedgingStarted()` — `true` iff the SDK dispatched ≥ 1 hedge arm.
- `virtual IReadOnlyList<RequestedRegion> GetRequestedRegions()` — every region the SDK dispatched to, in observed order, tagged with a `RequestedRegionReason`.
- `virtual IReadOnlyList<string> GetRespondedRegions()` — every region that produced a response, in arrival order, duplicates allowed.
- New types: `RequestedRegion` (readonly struct, `IEquatable<>`, case-insensitive name equality) and `RequestedRegionReason` enum (`Initial`, `OperationRetry`, `RegionFailover`, `Hedging`, `CircuitBreakerProbe`, `TransportRetry`).

All members are virtual with safe defaults (`false` / empty list); custom subclasses of `CosmosDiagnostics` are unaffected.

## Why this design

- **State lives on `TraceSummary.HedgingDetectionState`**, not as a `TraceDatum`. This means the `Diagnostics.ToString()` JSON shape is unchanged, while the API remains O(1). Mirrors the existing `RegionsContacted` pattern.
- **Cross-handler signaling via `RequestMessage.Properties[__CosmosDB_HedgingDetection_NextDispatchReason]`** — `Properties` is already deep-cloned per hedge arm by `RequestMessage.Clone()` and is shared with `DocumentServiceRequest.Properties`, so the upstream sites (`ClientRetryPolicy`, `CrossRegionHedgingAvailabilityStrategy`) can signal the downstream dispatch site (`TransportHandler`) without adding a new internal contract.
- **Region resolution at dispatch via `GlobalEndpointManager.GetLocation(locationEndpointToRoute)`** — `DocumentServiceRequest.RequestContext.RegionName` is not populated until after the append site.
- **No phantom Hedging entries** — hedge arms (`requestNumber > 0`) are only tagged inside `CloneAndSendAsync`, which is invoked only after the previous threshold delay elapses without primary cancellation.

See `openspec/changes/hedging-detection-api/design.md` for the full design write-up.

## Wiring

| Site | Change |
| ---- | ------ |
| `CrossRegionHedgingAvailabilityStrategy.CloneAndSendAsync` | For `requestNumber > 0` set `Properties[KEY] = Hedging`. |
| `ClientRetryPolicy.OnBeforeSendRequest` | After `RouteToLocation`, when `retryContext != null` set `Properties[KEY]` to `RegionFailover` or `OperationRetry`. |
| `TransportHandler.ProcessMessageAsync` | After `ToDocumentServiceRequest`, append a `RequestedRegion`, then remove the property. |
| `ClientSideRequestStatisticsTraceDatum` | At all 3 response-record sites, also append the region to `GetRespondedRegions`. |

## AC coverage

| AC   | Status | Mechanism |
| ---- | ------ | --------- |
| AC1  | ✅ | `HedgingDetectionState.HedgingStarted` flag, flipped on first `Hedging` append. |
| AC2  | ✅ | Hedge arm tag set only inside `requestNumber > 0`; threshold delay gates dispatch. |
| AC3  | ✅ | Single-region happy path: `Initial` requested, region recorded as responded. |
| AC4  | ✅ | `ClientRetryPolicy` tags `OperationRetry` for same-region retries. |
| AC5  | ✅ | `ClientRetryPolicy` tags `RegionFailover` when `RetryRequestOnPreferredLocations`. |
| AC6  | ⏸ Deferred | `CircuitBreakerProbe` reserved in enum but not populated — PPCB rerouting decision happens inside `Microsoft.Azure.Cosmos.Direct`'s `storeProxy.ProcessMessageAsync`, after the pre-dispatch append site. Tracked as follow-up requiring a Direct-package change. |
| AC7  | ⏸ Deferred | `TransportRetry` reserved but per-channel transport retries inside Direct are not surfaced at this layer in v1. |
| AC8  | ✅ | Hedge winner appears both in `GetRequestedRegions` (`Hedging`) and `GetRespondedRegions`. |
| AC9  | ✅ | Default virtuals on `CosmosDiagnostics` return `false` / empty — `CosmosDiagnosticsBackwardCompatTests`. |
| AC10 | ✅ | `IReadOnlyList<T>` returns immutable snapshot arrays. |
| AC11 | ✅ | State off the trace tree → `ToString()` JSON shape unchanged. |
| AC12 | ✅ | `RequestedRegion` is a readonly struct with case-insensitive name equality, `==`/`!=`, custom `GetHashCode`, `ToString`. |
| AC13 | ✅ | No phantom Hedging entries — see design §5. |
| AC14 | ✅ | Duplicates in `GetRespondedRegions` allowed and preserved. |
| AC15 | ✅ | Lock-protected snapshots; concurrency test (8 writers × 500 iterations). |
| AC16 | ⏸ Tracked | Live multi-region smoke test deferred to a follow-up emulator/live-account test; requires an environment outside the per-commit CI matrix. |

## Files changed

**Public surface (`src/Diagnostics`)**
- `CosmosDiagnostics.cs` — 3 new virtual methods with SE-013 `<remarks>`.
- `CosmosTraceDiagnostics.cs` — overrides delegate to `HedgingDetectionState`.
- `RequestedRegion.cs` (new) — readonly struct.
- `RequestedRegionReason.cs` (new) — enum.

**Internal state (`src/Tracing`)**
- `HedgingDetectionState.cs` (new) — lock-protected per-trace state.
- `TraceSummary.cs` — eager `HedgingDetectionState` property.

**Wiring**
- `Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs`
- `ClientRetryPolicy.cs`
- `Handler/TransportHandler.cs`
- `Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs`

**Tests**
- `Diagnostics/RequestedRegionTests.cs` (6 tests)
- `Diagnostics/HedgingDetectionStateTests.cs` (8 tests incl. concurrency)
- `Diagnostics/CosmosDiagnosticsBackwardCompatTests.cs` (3 tests, AC9)

**Contract / changelog / OpenSpec**
- `tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetSDKAPI.net6.json`
- `changelog.md`
- `openspec/changes/hedging-detection-api/{proposal.md,design.md,tasks.md,specs/hedging-detection/spec.md}`

## Validation

- `dotnet build Microsoft.Azure.Cosmos\src\Microsoft.Azure.Cosmos.csproj -c Debug` — clean ✅
- `dotnet build Microsoft.Azure.Cosmos\src\Microsoft.Azure.Cosmos.csproj -c Release` — clean ✅
- New unit tests — 17 / 17 ✅
- GA contract enforcement test — 107 / 107 ✅

## Open questions / follow-ups

- **AC6 / `CircuitBreakerProbe`** — full PPCB tagging needs a small change in `Microsoft.Azure.Cosmos.Direct`'s `DocumentServiceRequestContext` to flag probe requests. Suggested follow-up: surface an `IsPpcbProbe` flag and append a `CircuitBreakerProbe`-tagged entry at the rerouting site.
- **AC7 / `TransportRetry`** — same Direct-package consideration for per-channel transport retries.
- **AC16** — live multi-region smoke test; can either reuse the existing `Microsoft.Azure.Cosmos.EmulatorTests` infra or add a single live-account test under an environment-gated test category.
- Java / Python issue threads should be opened separately referencing issue #5867 and the cross-SDK invariants in `internal-spec.md`.

cc @kirankumarkolli — reviewer of record per the internal spec.


NaluTripician:
### Diagnostics sample — what callers will see

Attaching a sample of how the new surface looks at call-sites and what the values look like for a few realistic dispatch shapes. The `Diagnostics.ToString()` JSON shape is **unchanged** (state lives off the trace tree on `TraceSummary.HedgingDetectionState`), so the only observable difference is the three new accessors on `CosmosDiagnostics`.

#### Caller usage

```csharp
ItemResponse<Book> response = await container.ReadItemAsync<Book>(id, pk);
CosmosDiagnostics diagnostics = response.Diagnostics;

bool hedged = diagnostics.HedgingStarted();
IReadOnlyList<RequestedRegion> requested = diagnostics.GetRequestedRegions();
IReadOnlyList<string>          responded = diagnostics.GetRespondedRegions();

logger.LogInformation(
    "Hedging started: {Hedged}; requested=[{Req}]; responded=[{Resp}]",
    hedged,
    string.Join(", ", requested),                 // uses RequestedRegion.ToString() → "East US:Initial"
    string.Join(", ", responded));

// Non-exhaustive enum — always include a default arm
foreach (RequestedRegion r in requested)
{
    switch (r.Reason)
    {
        case RequestedRegionReason.Initial:             /* first dispatch */              break;
        case RequestedRegionReason.OperationRetry:      /* same-region retry */           break;
        case RequestedRegionReason.RegionFailover:      /* cross-region failover */       break;
        case RequestedRegionReason.Hedging:             /* hedge arm dispatched */        break;
        case RequestedRegionReason.CircuitBreakerProbe: /* PPCB probe (reserved in v1) */ break;
        case RequestedRegionReason.TransportRetry:      /* reserved in v1 */              break;
        default:                                        /* future-proof */                break;
    }
}
```

#### Scenarios

##### 1. Single-region happy path (no hedging configured)

```text
HedgingStarted()      = false
GetRequestedRegions() = [ East US:Initial ]
GetRespondedRegions() = [ East US ]
```

##### 2. Same-region retry (e.g. 410 Gone → SDK retries primary)

```text
HedgingStarted()      = false
GetRequestedRegions() = [ East US:Initial, East US:OperationRetry ]
GetRespondedRegions() = [ East US, East US ]
```

##### 3. Region fail-over (preferred-locations list exhausts primary, fails over)

```text
HedgingStarted()      = false
GetRequestedRegions() = [ East US:Initial, West US:RegionFailover ]
GetRespondedRegions() = [ East US, West US ]
```

##### 4. Hedging configured, primary wins under threshold (no hedge arm dispatched)

```text
HedgingStarted()      = false           // ← AC1: registered ≠ dispatched
GetRequestedRegions() = [ East US:Initial ]
GetRespondedRegions() = [ East US ]
```

> Note: `HedgingStarted() == false` here does **not** mean hedging was disabled — the primary just won under the configured threshold so no hedge arm ever fan-out. To check configuration, inspect `CosmosClientOptions.AvailabilityStrategy` / `RequestOptions.AvailabilityStrategy` directly.

##### 5. Hedging configured, hedge arm wins (primary slow / failed)

```text
HedgingStarted()      = true
GetRequestedRegions() = [ East US:Initial, West US:Hedging ]
GetRespondedRegions() = [ West US, East US ]   // hedge winner first; late primary response preserved (AC14)
```

##### 6. Hedging configured, two hedge arms, last responds (worst case)

```text
HedgingStarted()      = true
GetRequestedRegions() = [ East US:Initial, West US:Hedging, North Europe:Hedging ]
GetRespondedRegions() = [ North Europe, West US, East US ]
```

#### Equality / formatting on `RequestedRegion`

```csharp
RequestedRegion a = new RequestedRegion("East US", RequestedRegionReason.Hedging);
RequestedRegion b = new RequestedRegion("east us", RequestedRegionReason.Hedging);

a == b           // true (case-insensitive on name, exact on reason)
a.ToString()     // "East US:Hedging"
```

#### Backward-compat note (AC9)

Any customer-authored subclass of `CosmosDiagnostics` that does **not** override the new virtuals returns:

```text
HedgingStarted()      = false
GetRequestedRegions() = []   // Array.Empty<RequestedRegion>()
GetRespondedRegions() = []   // Array.Empty<string>()
```

This is also what the SDK returns for diagnostics objects produced before this version's plumbing populated them.

#### `Diagnostics.ToString()` JSON shape

Unchanged. State is on `TraceSummary.HedgingDetectionState`, not in the trace tree, so existing dashboards / log parsers that consume the JSON are unaffected.

Microsoft.Azure.Cosmos/src/Diagnostics/RequestedRegionReason.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// <c>Microsoft.Azure.Cosmos.Direct</c> package and is not observable from the SDK layer.
+    /// </para>
+    /// </remarks>
+    public enum RequestedRegionReason : byte
+    {
+        /// <summary>
+        /// The first dispatch of the operation. Appears exactly once per operation at the
+        /// beginning of the dispatch sequence.
+        /// </summary>
+        Initial = 0,
```
kundadebdatta: **🔴 Critical (F1)** `default(RequestedRegion)` is indistinguishable from a real `Initial` dispatch.

`RequestedRegion` is a `readonly struct`, so `default(RequestedRegion)` bypasses the ctor null-guard and yields `RegionName == null, Reason == Initial` — a perfectly valid-looking `Initial` entry with no region name. Any `new RequestedRegion[N]`, uninitialized field, or deserialized value will produce phantom `Initial` entries that aren't bugs callers can attribute to themselves.

After GA this is an API-shape problem we can't fix. **Suggested fix:** reserve `Unknown = 0` (or `None = 0`) as the default sentinel and shift `Initial = 1`. Then `default(RequestedRegion).Reason` is unambiguously the "no real dispatch" value and callers can detect it. Document the sentinel in the non-exhaustive `<remarks>` block.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosDiagnostics.cs
```diff
@@ -97,5 +97,100 @@ public virtual ServerSideCumulativeMetrics GetQueryMetrics()
...
+        /// <c>RequestOptions.AvailabilityStrategy</c> directly.
+        /// </para>
+        /// <para>
+        /// O(1). Safe to call on both the success path (<c>ItemResponse&lt;T&gt;.Diagnostics</c>)
+        /// and the error path (<c>CosmosException.Diagnostics</c>). Returns <c>false</c> for
+        /// diagnostics objects created outside the SDK pipeline (e.g. customer-authored
+        /// subclasses of <see cref="CosmosDiagnostics"/>).
+        /// </para>
+        /// </remarks>
+        public virtual bool HedgingStarted()
```
kundadebdatta, line 125: **🟡 Recommendation (F2)** No end-to-end tests cover the wiring this PR adds.

The 17 added tests cover `HedgingDetectionState` and `RequestedRegion` in isolation plus AC9 backward-compat defaults — but **nothing** verifies the integration that is the entire ask of the PR:

- `ClientRetryPolicy.OnBeforeSendRequest` actually writes `OperationRetry` vs `RegionFailover` for the two branches of `RetryRequestOnPreferredLocations`.
- `CrossRegionHedgingAvailabilityStrategy.CloneAndSendAsync` writes `Hedging` only on `requestNumber > 0`.
- `TransportHandler.AppendDispatchedRegion` resolves a region, appends an entry, and removes the property.
- An end-to-end hedge fan-out via the existing mock pipeline produces the expected `HedgingStarted()`/`GetRequestedRegions()`/`GetRespondedRegions()` sequence.

AC1–AC5 are essentially asserted by inspection only. `CrossRegionHedgingAvailabilityStrategyTests` is the natural home for the integration coverage, and the gap is what would have caught F3 (hedge-arm retry overwrite) before merge.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -277,6 +277,20 @@ public void OnBeforeSendRequest(DocumentServiceRequest request)
...
 
             request.RequestContext.RouteToLocation(this.locationEndpoint);
+
+            // Hedging-Detection API: tag the upcoming dispatch reason on Properties so that
+            // the downstream dispatch site (TransportHandler / GatewayStoreModel) can append
+            // a RequestedRegion entry with the correct reason. Only override when this is a
+            // genuine retry attempt — first-attempt dispatches default to Initial (set by
+            // the dispatch site), and hedge-arm dispatches have their Hedging reason set by
+            // CrossRegionHedgingAvailabilityStrategy before reaching this policy.
+            if (this.retryContext != null && request.Properties != null)
```
kundadebdatta, line 295: **🟡 Recommendation (F3)** Hedge-arm internal retries silently re-tag the dispatch reason from `Hedging` to `OperationRetry`/`RegionFailover`.

On the first attempt of a hedge arm, `retryContext == null`, so the `Hedging` value set by `CrossRegionHedgingAvailabilityStrategy.CloneAndSendAsync` survives — good. But when that hedge arm itself triggers a retry (e.g., 410 Gone or 449), the same cloned `RequestMessage` re-enters `OnBeforeSendRequest` with `retryContext != null` and this block **unconditionally overwrites** the dispatch reason. `TransportHandler` has already `Remove`'d the prior `Hedging` value, so the retry of a hedge arm is recorded as `OperationRetry` — the hedge origin is silently lost from the `GetRequestedRegions()` sequence.

Arguably correct semantics (it *is* a retry now), but:

1. The XML-doc comment immediately above this block reads as if the author assumed `Hedging` would survive into `TransportHandler` ("hedge-arm dispatches have their Hedging reason set by CrossRegionHedgingAvailabilityStrategy *before reaching this policy*") — which is true on the first attempt but not on subsequent retries of the same arm.
2. The spec is silent on this corner.
3. No test pins the behavior.

**Suggested fix:** Make an explicit decision — either preserve `Hedging` if already present (check the property before overwriting), introduce a separate `HedgedRetry` reason, or accept the current behavior and document it + add a test that drives a 410 on a hedge arm and asserts the recorded sequence.

Microsoft.Azure.Cosmos/src/Handler/TransportHandler.cs
```diff
@@ -183,5 +193,70 @@ internal static ResponseMessage AggregateExceptionConverter(AggregateException a
...
+                reason = resolvedReason;
+                serviceRequest.Properties.Remove(HedgingDetectionState.DispatchReasonPropertyKey);
+            }
+
+            // Resolve region name from the routing endpoint URI.
+            string regionName = null;
+            Uri endpoint = serviceRequest.RequestContext?.LocationEndpointToRoute;
+            if (endpoint != null && globalEndpointManager != null)
+            {
+                regionName = globalEndpointManager.GetLocation(endpoint);
```
kundadebdatta, line 258: **🟡 Recommendation (F4)** Thin-client / PPAF / per-partition routed dispatches produce **no entry** in `GetRequestedRegions()`.

`GlobalEndpointManager.GetLocation(uri)` walks `AvailableWriteEndpointByLocation` and `AvailableReadEndpointByLocation` and returns `null` for any URI not in those maps. That includes:

- Thin-client endpoints (`ResolveThinClientEndpoint` in `ClientRetryPolicy.cs:274-277` returns endpoints from `ThinClientReadEndpoints`/`ThinClientWriteEndpoints`, which `GetLocation` does not consult).
- PPAF / PPCB-rerouted URIs.
- Any per-partition routed endpoint.

For those operations `AppendDispatchedRegion` bails silently — but `ClientSideRequestStatisticsTraceDatum.RecordResponse`/`RecordHttpResponse` still call `AppendResponded` because the responded region is derived from a different source. Net result: **`GetRespondedRegions()` has entries but `GetRequestedRegions()` is empty** for thin-client and PPCB scenarios — a surprising asymmetry that the XML docs do not disclose.

**Suggested fix (lowest cost):** add a remark to the XML doc on `CosmosDiagnostics.GetRequestedRegions` and `HedgingStarted` calling out that thin-client / PPAF dispatches are not currently included, and cross-reference AC6/AC7. **Better:** fall back to `serviceRequest.RequestContext.RegionName` when `GetLocation` returns null. **Best:** make `GlobalEndpointManager` aware of thin-client endpoints.

Microsoft.Azure.Cosmos/src/Handler/TransportHandler.cs
```diff
@@ -183,5 +193,70 @@ internal static ResponseMessage AggregateExceptionConverter(AggregateException a
...
+            }
+
+            // Resolve reason; consume the property so retries don't carry a stale value.
+            RequestedRegionReason reason = RequestedRegionReason.Initial;
+            if (serviceRequest.Properties != null
+                && serviceRequest.Properties.TryGetValue(HedgingDetectionState.DispatchReasonPropertyKey, out object reasonObj)
+                && reasonObj is RequestedRegionReason resolvedReason)
+            {
+                reason = resolvedReason;
+                serviceRequest.Properties.Remove(HedgingDetectionState.DispatchReasonPropertyKey);
```
kundadebdatta: **🟡 Recommendation (F5)** Property is consumed **before** the region-resolution check, so a failed resolution silently swallows the dispatch-reason signal.

Order of operations in `AppendDispatchedRegion`:

1. Read `reason` from `Properties[KEY]` (line 238-241).
2. **Remove the property** (this line).
3. Try to resolve `regionName` from `LocationEndpointToRoute` (lines 246-251).
4. If region resolution fails, `return` at line 256 without appending.

The dispatch-reason signal has now been consumed but never recorded. If anything downstream re-enters this dispatch path on the same `serviceRequest` (or if a future code path also reads the key), the original reason is gone and the next entry will default to `Initial` even though the upstream caller intended a different reason.

**Suggested fix:** Only `Remove` the property *after* a successful `state.AppendRequested(...)` call (i.e., move the `Remove` to immediately before or after the append on line 259). Otherwise leave the signal in place so a re-dispatch can still consume it.

Microsoft.Azure.Cosmos/src/Handler/TransportHandler.cs
```diff
@@ -183,5 +193,70 @@ internal static ResponseMessage AggregateExceptionConverter(AggregateException a
...
+                regionName = globalEndpointManager.GetLocation(endpoint);
+            }
+
+            // Skip if we couldn't resolve a name; better empty than misleading "unknown".
+            if (string.IsNullOrEmpty(regionName))
+            {
+                return;
+            }
+
+            state.AppendRequested(regionName, reason);
```
kundadebdatta, line 268: **🟡 Recommendation (F8)** No test verifies the cross-handler property-cleanup invariant.

A core design invariant of this PR is: `TransportHandler.AppendDispatchedRegion` reads and **removes** the `Properties[DispatchReasonPropertyKey]` after consumption so that a subsequent dispatch on the same request defaults to `Initial` unless an upstream site re-tags it. A regression here (e.g., a future refactor that drops the `Remove` call, or that re-reads stale state) would produce silent reason corruption — exactly the kind of bug that won't show up in any of the current in-isolation unit tests.

**Suggested fix:** Add a focused unit test that invokes the dispatch site twice on the same `RequestMessage` without re-tagging between calls, and asserts that the second recorded entry is `Initial`. Pair it with a test that drives `ClientRetryPolicy.OnBeforeSendRequest` then the dispatch site, asserting the recorded reason matches the policy's intent. These two tests would also lock the design decision behind F3 (hedge-arm retry overwrite).

Microsoft.Azure.Cosmos/src/Tracing/TraceSummary.cs
```diff
@@ -95,5 +95,16 @@ public void AddRegionContacted(string regionName, Uri locationEndpoint)
...
+        /// <summary>
+        /// Per-operation state backing the Hedging Detection API surface (HedgingStarted /
+        /// GetRequestedRegions / GetRespondedRegions on <see cref="CosmosDiagnostics"/>).
+        /// </summary>
+        /// <remarks>
+        /// Lives on <see cref="TraceSummary"/> so that the entire trace tree for a single
+        /// operation shares one state instance. Populated at orchestrator dispatch sites
+        /// and response-handling sites; never serialized into the trace tree.
+        /// </remarks>
+        public HedgingDetectionState HedgingDetectionState { get; } = new HedgingDetectionState();
```
kundadebdatta, line 107: **🟡 Recommendation (F6)** `HedgingDetectionState` is eagerly allocated for **every** `TraceSummary`.

Every operation now allocates one `HedgingDetectionState` + one `object regionLock` even when no `AvailabilityStrategy` is configured (i.e., for the vast majority of customers). The inner `requestedRegions`/`respondedRegions` lists are already lazy, so the wrapper is the only mandatory cost — but on hot read paths at high QPS this is two small-object allocations per operation that most callers will never query.

All wired consumers already use null-safe access (`?.HedgingDetectionState?.AppendResponded(...)`, `?.HedgingDetectionState?.AppendRequested(...)`), so making the wrapper lazy is a drop-in change:

```csharp
private HedgingDetectionState hedgingDetectionState;
public HedgingDetectionState HedgingDetectionState
    => LazyInitializer.EnsureInitialized(ref this.hedgingDetectionState);
```

Matches the existing `regionContactedInternal` lazy pattern in this same class.

Microsoft.Azure.Cosmos/src/Tracing/HedgingDetectionState.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                this.respondedRegions.Add(regionName);
+            }
+        }
+
+        /// <summary>
+        /// Returns <c>true</c> if at least one <see cref="RequestedRegionReason.Hedging"/>
+        /// entry has been appended for this operation.
+        /// </summary>
+        internal bool HedgingStarted
```
kundadebdatta: **🟡 Recommendation (F7)** `HedgingStarted` getter takes the lock to read a write-once-true `bool`.

The public `CosmosDiagnostics.HedgingStarted()` virtual is documented as "O(1). Safe to call on both the success path and the error path." But the implementation acquires a Monitor on every read of a monotonic flag (set once to true inside `AppendRequested`, never reset). This adds an avoidable `Monitor.Enter` per customer call on the hot diagnostics path.

**Suggested fix:** Mark the backing field `volatile bool hedgingStarted` and read it without the lock. Keep the write under the lock (or switch to `Volatile.Write`/`Interlocked.Exchange`) to preserve the memory-ordering guarantee with the list mutation.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -329,6 +329,16 @@ private async Task<HedgingResponse> CloneAndSendAsync(
...
+
+                    // Hedging-Detection API: this code path is only reached AFTER the
+                    // previous loop iteration's threshold delay elapsed without primary-wins
+                    // cancellation. Tag the upcoming dispatch as Hedging so the downstream
+                    // dispatch site records it with the correct reason. If this method
+                    // is never invoked for a given requestNumber (e.g., primary wins under
+                    // the threshold), no phantom Hedging entry is produced — see AC2/AC13
+                    // and design doc §12 "no phantom entries".
+                    clonedRequest.Properties[HedgingDetectionState.DispatchReasonPropertyKey] =
+                        RequestedRegionReason.Hedging;
```
kundadebdatta, line 341: **💬 Observation (F9)** New API surface overlaps with the existing `HedgeContext`/`HedgeConfig`/`ResponseRegion` trace datums.

This same strategy already writes `HedgeContext`, `HedgeConfig`, and `ResponseRegion` datums to diagnostics (see `ExecuteAvailabilityStrategyAsync` lines 227-280; `HedgeContext` semantics were specifically refined by PR #5665 to mean "hedging actually occurred"). The new `HedgingStarted()` / `GetRequestedRegions()` / `GetRespondedRegions()` accessors convey overlapping information through a separate path — for example, the `HedgingStarted()` `true` condition is equivalent to "`HedgeContext` datum present" per #5665's contract.

The PR doesn't document the relationship or a migration/deprecation plan for the existing datums. Two consumers reading the same trace might disagree because one parses `ToString()` JSON (existing datums) and one calls the new API.

**Suggested action:** Either explicitly mark the existing datums as superseded (with a deprecation timeline) or document that they remain authoritative for JSON serialization while the new API is the recommended programmatic surface. Worth a paragraph in the changelog either way.

Microsoft.Azure.Cosmos/src/Diagnostics/RequestedRegion.cs
```diff
@@ -0,0 +1,103 @@
...
+        public override int GetHashCode()
+        {
+            int regionHash = this.RegionName == null
+                ? 0
+                : StringComparer.OrdinalIgnoreCase.GetHashCode(this.RegionName);
+            // Combine without depending on System.HashCode (net6+) to keep this constant
+            // across TFMs.
+            unchecked
+            {
+                return (regionHash * 397) ^ (byte)this.Reason;
```
kundadebdatta, line 73: **💬 Observation (F10)** `GetHashCode` mixes `Reason` into only the low 8 bits.

`(regionHash * 397) ^ (byte)this.Reason` casts the enum to `byte` first, so two `RequestedRegion` instances differing only in `Reason` collide on the upper 24 bits of the hash. Equality is unaffected; hashed collections (`HashSet<RequestedRegion>`, `Dictionary`) degrade modestly when callers store many `(sameRegion, differentReason)` pairs — exactly the shape this PR is going to produce in hedge + retry scenarios.

The comment correctly justifies avoiding `System.HashCode` for TFM constancy. Trivial improvement that preserves that:

```csharp
return unchecked((regionHash * 397) ^ ((int)this.Reason * 31));
```

Microsoft.Azure.Cosmos/src/Diagnostics/RequestedRegionReason.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// for this operation and chosen a different region from the preferred-locations list.
+        /// </summary>
+        RegionFailover = 4,
+
+        /// <summary>
+        /// A probe dispatch to a previously circuit-broken region driven by the SDK's
+        /// per-partition automatic failover / per-partition per-region circuit breaker
+        /// (PPCB) health check.
+        /// </summary>
+        CircuitBreakerProbe = 5,
```
kundadebdatta: **💬 Observation (F11)** `CircuitBreakerProbe` is missing the "Reserved but not populated" doc note that `TransportRetry` has.

`TransportRetry` correctly states "**Reserved but not populated by the .NET SDK in this release.**" `CircuitBreakerProbe` has no such note, yet a `grep` over `src/` confirms it is also **never written** anywhere in this PR or in existing PPCB code. Callers writing a switch arm for it will never have it hit, and there is no indication that's the expected current-version behavior.

**Suggested fix:** Add the same "Reserved but not populated by the .NET SDK in this release" sentence to this XML doc (or wire PPCB probes to set the property at the dispatch site).

Microsoft.Azure.Cosmos/src/Tracing/HedgingDetectionState.cs
```diff
@@ -0,0 +1,153 @@
...
+    {
+        /// <summary>
+        /// Well-known key used to stash a pending dispatch-reason override on
+        /// <see cref="RequestMessage.Properties"/> (or
+        /// <see cref="Microsoft.Azure.Documents.DocumentServiceRequestContext"/>'s
+        /// properties bag) so an upstream site (e.g. ClientRetryPolicy or the hedging
+        /// orchestrator) can signal the downstream dispatch site why the next attempt
+        /// is happening. The value must be a <see cref="RequestedRegionReason"/>.
+        /// </summary>
+        internal const string DispatchReasonPropertyKey = "__CosmosDB_HedgingDetection_NextDispatchReason";
```
kundadebdatta, line 44: **💬 Observation (F12)** New `__CosmosDB_*` Properties-key prefix is a new convention in this repo.

A `gh search code "__CosmosDB_"` returns no other usages in the codebase. Existing cross-handler property-bag keys — for example `CrossRegionAvailabilityContext.PropertyKey` (consumed just above the new code in `ClientRetryPolicy.OnBeforeSendRequest`) and the various `WFConstants.BackendHeaders.*` keys — use bare class-scoped string constants without a leading prefix.

The encapsulation via `HedgingDetectionState.DispatchReasonPropertyKey` is fine; only the **string value** introduces the new prefix. It's defensible (collision-proof, easy to grep) but is a stylistic divergence worth either:

- A one-line code comment justifying the prefix (e.g., "internal-only sentinel key, prefixed to avoid collision with customer-supplied keys"), or
- Alignment to the existing bare-constant style.

Microsoft.Azure.Cosmos/src/Diagnostics/RequestedRegionReason.cs
kundadebdatta: This is now addressed. Thanks!

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
kundadebdatta: Fixed and added test to cover the scenario.

Microsoft.Azure.Cosmos/src/Handler/TransportHandler.cs
kundadebdatta: Addressed.

Microsoft.Azure.Cosmos/src/Tracing/HedgingDetectionState.cs
kundadebdatta: Thanks! This is a good catch. Addressed.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosDiagnostics.cs
```diff
@@ -97,5 +97,100 @@ public virtual ServerSideCumulativeMetrics GetQueryMetrics()
...
+        {
+            return false;
+        }
+
+        /// <summary>
+        /// Returns the regions the SDK actually dispatched this operation to, in observed
```
jeet1995, line 131: nit - thoughts on using verbiage like `CosmosClient dispatched this operation to ...`

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
```diff
@@ -158,5 +158,25 @@ public override int GetFailedRequestCount()
...
+        /// <inheritdoc/>
+        public override bool HedgingStarted()
+        {
+            return this.Value?.Summary?.HedgingDetectionState?.HedgingStarted ?? false;
+        }
+
+        /// <inheritdoc/>
+        public override IReadOnlyList<RequestedRegion> GetRequestedRegions()
+        {
+            return this.Value?.Summary?.HedgingDetectionState?.GetRequestedRegionsSnapshot()
```
jeet1995, line 171: How would debugging work when the result is empty? When all is empty outcome expected?

Microsoft.Azure.Cosmos/src/Diagnostics/RequestedRegionReason.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// <see cref="Hedging"/>.
+        /// </summary>
+        Hedging = 4,
+
+        /// <summary>
+        /// An endpoint-failure-driven retry to a different region (write conflict re-route,
+        /// region marked unavailable). The SDK has marked the previous endpoint as failed
+        /// for this operation and chosen a different region from the preferred-locations list.
+        /// </summary>
+        RegionFailover = 5,
```
jeet1995, line 94: Does it help to disambiguate various failovers - partition-level vs account-level?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -37,7 +37,7 @@ internal sealed class ClientRetryPolicy : IDocumentClientRetryPolicy
         private readonly IDocumentClientRetryPolicy throttlingRetry;
         private readonly GlobalEndpointManager globalEndpointManager;
         private readonly GlobalPartitionEndpointManager partitionKeyRangeLocationCache;
-        private readonly bool enableEndpointDiscovery;
+        private readonly bool enableEndpointDiscovery;
```
jeet1995, line 40: nit: Looks like difference in LF handling is sneaking through.

Microsoft.Azure.Cosmos/src/Diagnostics/CosmosTraceDiagnostics.cs
kundadebdatta: When `HedgingStarted() == false`, `GetRequestedRegions().Count == 0`, and `GetRespondedRegions().Count == 0`, the SDK did not dispatch this operation against any backend region in this client's session — either the diagnostics object was synthesized outside the pipeline (customer subclass / default(CosmosDiagnostics)-equivalent path), or the operation failed before reaching the dispatch site (pre-flight validation, serializer crash, etc.). For the latter, inspecting the `CosmosException` (if any) and `Diagnostics.ToString()` for the trace prefix would help uncover.

Microsoft.Azure.Cosmos/src/Diagnostics/RequestedRegionReason.cs
kundadebdatta: I don't think it does provides a clear indication on whether the failover was due to PPAF or Account level. If we are looking for that detail, then I think it's worth to follow up on a separate PR.

Cc: @NaluTripician can you track this from your end ?

jeet1995 (approved):

kushagraThapar (approved):
LGTM, thanks @NaluTripician and @kundadebdatta

PR #5869, opened 2026-05-15
[Internal] DTS: Adds coordinator diagnosticString, CosmosDiagnostics, and OpenTelemetry support

Meghana-Palaparthi:
# Pull Request Template

## Description

This pull request enhances distributed transaction support in the Cosmos DB SDK with improved diagnostics, tracing, and error handling. The changes introduce comprehensive diagnostic collection for distributed transaction commits, ensure that diagnostic information is consistently propagated and logged, and add safeguards for session token management. These improvements make distributed transactions more observable and robust, especially in failure and retry scenarios.

**Diagnostics and Tracing Enhancements:**

* Added `DiagnosticString` and `Diagnostics` properties to `DistributedTransactionResponse`, ensuring detailed coordinator and client-side diagnostics are available to callers. The diagnostic string is now surfaced in error messages for failed transactions. [[1]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87R126-R137) [[2]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87R322) [[3]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87R347-R352) [[4]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87R398-R422)
* Integrated OpenTelemetry tracing and improved operation helper usage in both `DistributedReadTransactionCore` and `DistributedWriteTransactionCore`, capturing richer telemetry and associating traces with distributed transaction commits. [[1]](diffhunk://#diff-6163326bb8cc03946b105c2171cea56bb35233a11e1c4134084a30117b7e4619R11-R12) [[2]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eR13-R14) [[3]](diffhunk://#diff-6163326bb8cc03946b105c2171cea56bb35233a11e1c4134084a30117b7e4619L47-R68) [[4]](diffhunk://#diff-8a48a060b9de00cc81904d4a3f5c9fb0d7085b2688b4744a5269db93ff072b6eL274-R294) [[5]](diffhunk://#diff-dcbc1967f14470915aa8905f983909e89d64a162c0b60cd3e0a6b57f213483cdR15-R18)
* Updated `DistributedTransactionCommitter` to accept an `ITrace` for commit operations, accumulate diagnostics across retries, and ensure diagnostics are attached to the final response. [[1]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074R14) [[2]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074L51-R54) [[3]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074L66-R69) [[4]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074R80-R104)

**Error Handling and Logging Improvements:**

* Added truncation for potentially large diagnostic strings before logging to prevent log bloat and interleaving, with a dedicated utility method.
* Improved session token merging logic to handle out-of-range indices and exceptions gracefully, logging warnings without failing already-committed transactions. [[1]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074R219-R226) [[2]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074R246-R261)

**Serialization and Deserialization Robustness:**

* Ensured that deserialization of `DistributedTransactionOperationResult` fails fast with a clear exception if the result is null, improving reliability.
* Added support for the `diagnosticString` property in distributed transaction serialization and deserialization.

**Response Construction Consistency:**

* Ensured that the preserved diagnostic string is propagated when constructing new `DistributedTransactionResponse` instances from response messages. [[1]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87R239) [[2]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L237-R253)

These changes collectively make distributed transactions more transparent, easier to troubleshoot, and safer in production environments.

## Type of change

## Closing issues

## Changelog

> This feature is still not accessible to the public

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
@@ -123,6 +123,19 @@ namespace Microsoft.Azure.Cosmos
         /// </summary>
         public virtual bool IsRetriable { get; }
 
+        /// <summary>
+        /// Gets the diagnostic string from the coordinator describing the transaction outcome
+        /// </summary>
+        public virtual string DiagnosticString { get; private set; }
+
+        /// <summary>
+        /// Gets the client-side diagnostics for the distributed transaction, covering the full
+        /// retry loop and per-attempt spans (address resolution, network, retries, latency).
+        /// </summary>
+        /// <remarks>Non-null when the response is returned from <c>CommitTransactionAsync</c>. The SDK sets this property
+        /// before returning; callers should treat it as non-null in normal usage but guard defensively in edge cases.</remarks>
+        public virtual CosmosDiagnostics Diagnostics { get; internal set; }
+
         internal virtual SubStatusCodes SubStatusCode { get; }
 
         internal virtual CosmosSerializerCore SerializerCore { get; }
@@ -224,17 +237,22 @@ namespace Microsoft.Azure.Cosmos
 
                         if (responseMessage.IsSuccessStatusCode)
                         {
+                            string preservedDiagnosticString = response.DiagnosticString;
+                            SubStatusCodes wireSubStatusCode = responseMessage.Headers.SubStatusCode;
                             response.Dispose();
 
                             return new DistributedTransactionResponse(
                                 HttpStatusCode.InternalServerError,
-                                SubStatusCodes.Unknown,
+                                wireSubStatusCode,
                                 ClientResources.InvalidServerResponse,
                                 responseMessage.Headers,
                                 serverRequest.Operations,
                                 serializer,
                                 createResponseTrace,
-                                idempotencyToken);
+                                idempotencyToken)
+                            {
+                                DiagnosticString = preservedDiagnosticString,
+                            };
                         }
 
                         response.CreateAndPopulateResults(serverRequest.Operations, createResponseTrace);
@@ -303,6 +321,7 @@ namespace Microsoft.Azure.Cosmos
         {
             List<DistributedTransactionOperationResult> results = new List<DistributedTransactionOperationResult>();
             bool isRetriable = false;
+            string diagnosticString = null;
 
             JsonDocument responseJson;
             try
@@ -325,6 +344,12 @@ namespace Microsoft.Azure.Cosmos
                     isRetriableElement.ValueKind == JsonValueKind.True)
                 {
                     isRetriable = true;
+                }
+
+                if (root.TryGetProperty(DistributedTransactionSerializer.DiagnosticString, out JsonElement diagnosticStringElement) &&
+                    diagnosticStringElement.ValueKind == JsonValueKind.String)
+                {
+                    diagnosticString = diagnosticStringElement.GetString();
                 }
 
                 // Parse operation results from "operationResponses" array.
@@ -372,10 +397,22 @@ namespace Microsoft.Azure.Cosmos
                 }
             }
 
+            // Incorporate the coordinator's diagnosticString into the error message so it
+            // surfaces in response.ErrorMessage and any exception message the caller builds.
+            // Only merge on error responses — success responses must keep ErrorMessage null.
+            string effectiveErrorMessage = responseMessage.ErrorMessage;
+            bool isSuccessStatus = (int)finalStatusCode >= 200 && (int)finalStatusCode <= 299;
+            if (!isSuccessStatus && !string.IsNullOrWhiteSpace(diagnosticString))
+            {
+                effectiveErrorMessage = string.IsNullOrWhiteSpace(effectiveErrorMessage)
+                    ? diagnosticString
+                    : $"{effectiveErrorMessage} ({diagnosticString})";
+            }
+
             return new DistributedTransactionResponse(
                 finalStatusCode,
                 finalSubStatusCode,
-                responseMessage.ErrorMessage,
+                effectiveErrorMessage,
                 responseMessage.Headers,
                 serverRequest.Operations,
                 serializer,
@@ -383,7 +420,8 @@ namespace Microsoft.Azure.Cosmos
                 idempotencyToken,
                 isRetriable)
             {
-                results = results
+                results = results,
+                DiagnosticString = diagnosticString
             };
         }
 
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
@@ -10,6 +10,8 @@ namespace Microsoft.Azure.Cosmos
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos.Telemetry.OpenTelemetry;
+    using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Documents;
 
     internal class DistributedWriteTransactionCore : DistributedWriteTransaction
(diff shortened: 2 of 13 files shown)


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedWriteTransactionCore.cs
```diff
@@ -271,13 +273,25 @@ public override DistributedWriteTransaction UpsertItemStream(
...
+                task: (trace) =>
+                {
+                    DistributedTransactionCommitter committer = new DistributedTransactionCommitter(
+                        operations: this.operations,
+                        clientContext: this.clientContext);
 
-            return await committer.CommitTransactionAsync(cancellationToken);
+                    return committer.CommitTransactionAsync(trace, cancellationToken);
+                },
+                openTelemetry: new (OpenTelemetryConstants.Operations.CommitDistributedWriteTransaction,
```
kushagraThapar, line 292: Could you please add a write-side wiring test that mirrors `DistributedReadTransactionCoreTests.CommitTransactionAsync_RoutesThroughOperationHelper_WithExpectedWiring` (currently asserts `CommitDistributedReadTransaction`, `OperationType.CommitDistributedTransaction`, and `TraceComponent.Batch` for the read flow)?

The wiring here is symmetric to the read side, but a copy-paste regression in the OTel constant on this line — e.g. accidentally using `CommitDistributedReadTransaction` for the write path — wouldn't be caught by anything we have today. A 30-line test parity with the read one would close the gap.

Non-blocking.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/OpenTelemetryResponse.cs
```diff
@@ -29,6 +29,21 @@ internal OpenTelemetryResponse(TransactionalBatchResponse responseMessage)
         {
         }
 
+        internal OpenTelemetryResponse(DistributedTransactionResponse response)
+           : this(
+                  statusCode: response.StatusCode,
```
kushagraThapar: Tiny defensive nit — `response.StatusCode` / `response.Count` are unguarded, so this constructor would NPE if a future caller (or a refactor of `OperationHelperAsync`) ever invokes the OTel `GetAttributes` callback with a `null` response. The two other constructors in this file have the same shape so they're not really worse off, but since this one is brand new it would be nice to either:

- Add a `response is null` short-circuit at the top, or
- Throw `ArgumentNullException(nameof(response))` explicitly so the failure mode is unambiguous.

Not a blocker — `OperationHelperAsync` won't pass null today — just future-proofing.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -224,17 +237,22 @@ internal static async Task<DistributedTransactionResponse> FromResponseMessageAs
 
                         if (responseMessage.IsSuccessStatusCode)
                         {
+                            string preservedDiagnosticString = response.DiagnosticString;
+                            SubStatusCodes wireSubStatusCode = responseMessage.Headers.SubStatusCode;
```
kushagraThapar, line 241: Nice fix to preserve `responseMessage.Headers.SubStatusCode` here instead of hardcoding `SubStatusCodes.Unknown`. I am curious to learn whether you'd be willing to add a tiny test assertion to lock this in — something like extending `FromResponseMessage_*InvalidServerResponse*` (or adding a new case) to verify the rebuilt response's `SubStatusCode` matches the wire header. It's a small behavior change that's easy to regress silently otherwise.

Non-blocking.

kushagraThapar (dismissed):
Approving — really nice work on this one. A few small polish items below as non-blocking follow-ups.

## What I liked

- **Trace propagation is done right.** Threading the caller's `parentTrace` into `ExecuteCommitWithRetryAsync` (instead of allocating a sibling root) means the per-attempt spans nest properly under the operation's root trace, and `CosmosTraceDiagnostics` is allocated once and reused across retries. `response.Diagnostics` is set on every return path that returns a response (success early-return AND retry-budget-exhausted), which is exactly what we want.
- **The `ErrorMessage` merge is carefully guarded** — `IsNullOrWhiteSpace` checks on both pieces, success-status check to keep `ErrorMessage` null on 2xx, and the MultiStatus-promotion case (207→409 via per-op promotion) is covered. The edge-case tests for empty / whitespace-only / mixed-whitespace diagnosticString are great.
- **OTel compatibility check** updated to remove `DistributedTransactionResponse` from the skip list and actually exercise the new constructor — keeps us honest going forward.
- **`SubStatusCodes.Unknown` → `wireSubStatusCode`** in the invalid-body rebuild path is a nice small fix to preserve server-provided context.

## Polish suggestions (non-blocking)

1. Could you add a write-side wiring test mirroring `DistributedReadTransactionCoreTests.CommitTransactionAsync_RoutesThroughOperationHelper_WithExpectedWiring`? See the inline comment.
2. A defensive null-guard on the new `OpenTelemetryResponse(DistributedTransactionResponse)` constructor would protect against a future caller passing `null` — see inline.
3. The PR description mentions "Ensured that deserialization of `DistributedTransactionOperationResult` fails fast with a clear exception if the result is null", but I couldn't find that change in this diff. Could you either remove it from the description or point me at the corresponding code so I know what to look for? Just want to be sure nothing's missed.
4. The `SubStatusCodes.Unknown` → `wireSubStatusCode` change deserves a one-line test assertion since it's a subtle behavior change — see inline.

Again, these are polish items and not blockers for approval. Thank you!

kushagraThapar (approved):

kushagraThapar (approved):
Thanks @Meghana-Palaparthi — really appreciate the quick turnaround on all three follow-ups. Re-verified against HEAD:

- ✅ Write-side wiring test added at `DistributedWriteTransactionTests.cs:393` (`CommitTransactionAsync_RoutesThroughOperationHelper_WithExpectedWiring`) — locks in operation name, `OperationType.CommitDistributedTransaction`, `TraceComponent.Batch`, and the `CommitDistributedWriteTransaction` OTel constant. That'll catch any copy-paste from the read path cleanly.
- ✅ `OpenTelemetryResponse(DistributedTransactionResponse)` now does `(response ?? throw new ArgumentNullException(nameof(response)))` — unambiguous failure mode.
- ✅ New `FromResponseMessage_CountMismatch_SuccessStatus_PreservesWireSubStatusCode` test pins down the synthetic-500 sub-status preservation. Nice extra touch preserving `DiagnosticString` across the rebuild too.

Approving. 🙏

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -123,6 +123,19 @@ public virtual DistributedTransactionOperationResult this[int index]
         /// </summary>
         public virtual bool IsRetriable { get; }
 
+        /// <summary>
+        /// Gets the diagnostic string from the coordinator describing the transaction outcome
+        /// </summary>
+        public virtual string DiagnosticString { get; private set; }
```
jeet1995, line 129: Could this be plugged into `CosmosDiagnostics`? That way you don't expose an extra property (especially of a service-side telemetry property).

jeet1995 (approved):

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
Meghana-Palaparthi: DiagnosticString is a server-side coordinator payload, while CosmosDiagnostics is a client-side structured trace object.

So, it might be possible to be surfaced alongside CosmosDiagnostics, but it doesn’t naturally belong inside it without mixing server and client diagnostics. I'll explore this more.

PR #5864, opened 2026-05-13
[Internal] Changelog: Adds per-PR Unreleased pattern; retires in-repo release-copilot-agent

NaluTripician:
## Summary

Moves the .NET Cosmos SDK and the FaultInjection package from release-time PR-title-derived
changelogs to per-PR manual entries under a persistent `### Unreleased` section, matching the
Java and Python SDK pattern.

After this PR:
- Every PR that touches `Microsoft.Azure.Cosmos/src/**` or `Microsoft.Azure.Cosmos/FaultInjection/src/**` must add (or justify the absence of) a changelog entry under `### Unreleased` in the corresponding `changelog.md`.
- The release agent (now exclusively in the toolkit) reads that section, reconciles it, and closes it with a version+date at release time.

**This PR must be reviewed as a pair with the cosmos-sdk-copilot-toolkit PR:**
**[azure-data-database-platform/cosmos-sdk-copilot-toolkit#68](https://github.com/azure-data-database-platform/cosmos-sdk-copilot-toolkit/pull/68)**
Do not approve one without the other.

Background: [Issue #5858](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5858),
[PR #5860](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5860) (retroactive fix-up, separate in-flight).

### Changelogs (Unreleased section structure)

| File | Action | Notes |
|---|---|---|
| `changelog.md` | Edit | Adds `### Unreleased` block with four subsections. Includes PR #5851 (Change Feed Estimator) pre-staged under `#### Bugs Fixed`. |
| `Microsoft.Azure.Cosmos/FaultInjection/changelog.md` | Edit | Prepends `### Unreleased` block (anchor: `unreleased-faultinjection`). Existing beta sections untouched. |

### Contributor documentation

| File | Action | Notes |
|---|---|---|
| `CONTRIBUTING.md` | Edit | Inserts `## Changelog entry` section after Contribution flow. Includes: where to add, how to write, verb→subsection table, [Internal] definition, preview-feature carve-out, reviewer checklist, merge conflict guidance. |
| `PULL_REQUEST_TEMPLATE.md` | Edit | Appends `## Changelog` block with two checkboxes. |

### AI / Copilot guidance

| File | Action | Notes |
|---|---|---|
[...]

PR #5870, opened 2026-05-15
CrossRegionHedgingAvailabilityStrategy: Fixes `StackOverflow` in `CrossRegionHedgingAvailabilityStrategy` Observed in .NET Framework 4.7.2

kundadebdatta:
## Problem statement

Customer using the latest .NET SDK on **.NET Framework 4.7.2** hit a `System.StackOverflowException`
during a regional outage. The dump shows the failing thread's top frames in
`CrossRegionHedgingAvailabilityStrategy`:

```
RequestSenderAndResultCheckAsync     (deepest hedging frame; calls into the request pipeline)
  → ExceptionDispatchInfo.Throw
  → TaskAwaiter.ThrowForNonSuccess
  → TaskAwaiter.HandleNonSuccessAndDebuggerNotification
CloneAndSendAsync                    (middle hedging frame)
  → ExceptionDispatchInfo.Throw
  → TaskAwaiter.ThrowForNonSuccess
ExecuteAvailabilityStrategyAsync     (topmost hedging frame)
  → ExceptionDispatchInfo.Throw
  → TaskAwaiter.ThrowForNonSuccess
```

Every `CosmosOperationCanceledException` thrown deep in the request pipeline (e.g. when
`hedgeRequestsCancellationTokenSource` cancels in-flight hedged calls) propagates *synchronously*
back through the entire pipeline of awaiting async methods. On .NET Framework 4.7.2 each async
method's exception path consumes ~10KB of stack (`ExceptionDispatchInfo.Throw` →
`TaskAwaiter.ThrowForNonSuccess` → `HandleNonSuccessAndDebuggerNotification`). With the request
pipeline (handlers, retry policies, store clients, etc.) layered under hedging, the cumulative
stack consumption can blow the 1MB managed thread stack.

This is **not reproducible on .NET Core / .NET 5+** — those runtimes reset the synchronous
exception stack across `await` boundaries automatically. Only .NET Framework 4.x consumers
are affected.

## Approach

Apply the documented workaround: **insert `await Task.Yield();` in the catch block of
middle-layer async methods** along the propagation chain. `Task.Yield()` schedules the
continuation on the threadpool, so the rethrown exception unwinds onto a fresh stack
instead of accumulating frames.

Guidance:
- **Never** add it at the deepest frame (it just delays without resetting useful stack).
- **Never** add it at the absolute topmost frame (caller has nowhere left to unwind into).
- **Do** add it on a middle layer — once is enough per long chain to reset the stack.

### P0 — Fix the hedging stack overflow (this PR)

`Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs`

1. **`CloneAndSendAsync` (middle layer)** — currently has no try/catch. Wrap the awaited call
   in `try { ... } catch { await Task.Yield(); throw; }`. This is the recommended insertion
   point per the runtime guidance — middle-layer, between the deepest hedging frame
   (`RequestSenderAndResultCheckAsync`) and the topmost (`ExecuteAvailabilityStrategyAsync`).
2. **`RequestSenderAndResultCheckAsync` — `throw ex;` bug fix** — line 372 uses `throw ex;`
   which resets the captured stack trace. Change to `throw;` (a tightly-coupled bug discovered
   while reading the same catch block).
3. **Unit test** — extend `AvailabilityStrategyUnitTests` with a regression test that wires up
   a deep, fake handler chain whose deepest stage throws `OperationCanceledException`, drives
   it through hedging, and asserts the call returns/throws normally without a `StackOverflowException`.
   We can't directly assert "no SO" (SO terminates the process), but we can assert that the
   continuation after the catch runs on a different thread (proof the yield happened) and that
   the original exception type and message are preserved.

### P1 — Audit other deep async chains in the SDK (follow-up)

Same pattern is *latent* anywhere a long async/await chain rethrows from a catch on
.NET Framework. Candidates worth auditing in a follow-up PR:

| File | Location | Notes |
| --- | --- | --- |
| `Handler/AbstractRetryHandler.cs` | `SendAsync` catches (`DocumentClientException`, `CosmosException`, `AggregateException`, `OperationCanceledException`) and `ExecuteHttpRequestAsync` retry loop | Sits in the request pipeline; long await chain below it |
| `ChangeFeed/ChangeFeedIteratorCore.cs` | Two `catch (OperationCanceledException) → throw new CosmosOperationCanceledException` blocks (lines 243, 287) | Async iterator — deep pipeline below |
| `ReadFeed/ReadFeedIteratorCore.cs` | Same pattern (line 257) | |
| `Query/v3Query/QueryIterator.cs` | Same pattern (line 225) | |
| `Resource/ClientContextCore.cs` | `OperationHelperWithRootTraceAsync` `TryTransformException` rethrow (line 615) | Top-of-pipeline; arguably "topmost", but worth verifying |

Approach for P1:
- Triage each by stack depth below the catch (deepest pipelines = highest risk).
- For each confirmed risk, add `await Task.Yield()` in **one** middle-layer catch on the
  propagation path (typically the iterator/handler that sits 1 layer above the request
  pipeline, not the very top).
- Do **not** add `await Task.Yield()` everywhere — it has a small perf cost (extra threadpool
  hop) on every exception. One per long chain is sufficient.

### Out of scope

- Refactoring hedging concurrency model.
- Changing the `Task.WhenAny` accumulation pattern (the leaked-task issue is separate).
- Adding `Task.Yield` on the success path.

## Notes

- `Task.Yield()` cost: one threadpool dispatch per exception. Negligible vs. the alternative
  (process crash) and only paid on the exception path, which is already slow.
- Behavior on .NET Core / .NET 5+: identical observable behavior; one extra threadpool hop
  on exceptions. Acceptable.
- The fix is netstandard2.0 friendly (target framework of `Microsoft.Azure.Cosmos`).

## Validation plan

1. Build `Microsoft.Azure.Cosmos.csproj`.
2. Run `AvailabilityStrategyUnitTests` and any other existing hedging tests.
3. Ensure the new regression test passes.
4. Confirm contracts (`API_*.txt`) are unchanged — no public API surface changes.

## Status

- Branch: `users/kundadebdatta/fix_stackoverflow_error_in_availability_strategy`
- Commit: `d04487cef` — "Fix StackOverflowException in CrossRegionHedgingAvailabilityStrategy on .NET Framework"
- Build: clean, 0 warnings.
- Tests: 103/103 related tests pass (AvailabilityStrategy + ClientRetryPolicy + RetryHandler + CrossRegion + PartitionKeyRangeFailover).
- Push: **not pushed yet** — awaiting user confirmation per stored preference.

## Next steps

- Get user confirmation, then push the branch and open a PR.
- Schedule the P1 audit (per the table above) as a follow-up PR.


FabianMeiswinkel (dismissed):
LGTM

ananth7592 (dismissed):
LGTM except for one item

Task.Yield() runs on all runtimes including .NET 6+ where SO doesn't exist; comment says "no-op" but it's a threadpool dispatch. The threadpool dispatch adds pressure but since it is much miniscule in impact compared to bottleneck of network timeouts happening during outage /hedge scenario

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AvailabilityStrategyUnitTests.cs
```diff
@@ -812,5 +812,149 @@ public void CrossRegionAvailabilityContext_PropagatesHubHeaderFlagToHedgedReques
...
+        ///     SynchronizationContext during exception propagation, demonstrating the synchronous
+        ///     propagation chain was broken.
+        ///
+        /// NOTE on test target framework: this test project (Microsoft.Azure.Cosmos.Tests) only
+        /// targets net6.0, where the underlying StackOverflowException does NOT reproduce — .NET
+        /// Core / .NET 5+ already optimize the synchronous exception-propagation path. The test
+        /// therefore asserts the proximate cure (the yield occurred + stack trace was preserved)
+        /// rather than the absence of an SO. That is sufficient regression coverage: removing the
+        /// production fix in CloneAndSendAsync's catch block makes the PostCount assertion below
+        /// fail, and removing the throw-ex -> throw fix in RequestSenderAndResultCheckAsync makes
```
NaluTripician, line 844: **Major — this claim is empirically false; the test does not actually cover the `throw ex;` → `throw;` change.**

The docstring (lines 832-833) and this comment (lines 843-844) both assert the test exercises the `throw ex;` → `throw;` fix in `RequestSenderAndResultCheckAsync`. Verified by reverting just that line back to `throw ex;` (keeping the `Task.Yield()` wrapper intact) and re-running this test:

```
Passed!  - Failed: 0, Passed: 1, Skipped: 0
```

Reason: with the application CT pre-cancelled and the sender throwing `OperationCanceledException`, control inside `RequestSenderAndResultCheckAsync` goes through the **filtered OCE catch** (`catch (OperationCanceledException oce) when (hedgeRequestsCancellationTokenSource.IsCancellationRequested)`), which wraps in `CosmosOperationCanceledException(oce, trace)`. The generic `catch (Exception ex)` at line 390-394 — where the `throw ex;` → `throw;` change actually lives — is never entered.

The stack-trace assertion still passes because `CosmosOperationCanceledException.StackTrace` is overridden to return `this.originalException.StackTrace` (see `Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosOperationCanceledException.cs:98`), so the original OCE's stack is preserved regardless of what happens at the generic-catch rethrow site.

**Suggested fix:** add a second test case (or parameterize this one) that drives the generic catch — uncancelled CTS, sender throws a non-OCE (e.g. `InvalidOperationException`), assert the surfaced exception's stack trace contains the sender's frame. Without that, future authors can silently revert `throw;` back to `throw ex;` with no signal.

(For contrast, the `postCountDelta > 0` assertion **is** honest: removing the `Task.Yield()` catch makes it fail deterministically with `Observed delta: 0`.)

changelog.md
```diff
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Bugs Fixed
 
+- [5870](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5870) CrossRegionHedgingAvailabilityStrategy: Fixes StackOverflow in CrossRegionHedgingAvailabilityStrategy Observed in .NET Framework 4.7.2.
```
NaluTripician, line 29: **Major (merge-time blocker) — wrong changelog section; needs rebase.**

This branch was forked from `2e69650f`, well before commit `8b03a5af0` on `main` ([#5864](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5864) — "[Internal] Changelog: Adds per-PR Unreleased pattern"), which restructured this file. Upstream `main` now uses:

```
### <a name="unreleased"/> Unreleased Preview
#### Fixed
#### Added
```

The branch still appends under the old `### Unreleased` → `#### Bugs Fixed` heading, which no longer exists on `main`. Merging as-is will either produce a conflict on this file or — depending on how it's resolved — silently drop the entry under a heading that no longer exists.

**Suggested fix:** rebase on current `main` and re-place this bullet under `#### Fixed` in the `Unreleased Preview` section.

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -284,7 +284,11 @@ internal override async Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
 
                     if (lastException != null)
                     {
-                        throw lastException;
+                        // Use ExceptionDispatchInfo to preserve the original throwing-frame stack
+                        // trace. `throw lastException;` would reset the StackTrace property to the
+                        // current frame, which defeats the throw-vs-throw-ex preservation work in
+                        // CloneAndSendAsync / RequestSenderAndResultCheckAsync.
+                        System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(lastException).Throw();
```
NaluTripician, line 291: **Minor — undeclared third source change with no test coverage.**

The PR description lists two production changes (the `Task.Yield`/rethrow wrapper in `CloneAndSendAsync` and `throw ex;` → `throw;` in `RequestSenderAndResultCheckAsync`). This `throw lastException;` → `ExceptionDispatchInfo.Capture(lastException).Throw()` is a third change that isn't called out anywhere.

The change itself is correct (better stack preservation than `throw lastException;`), but:

1. The new regression test doesn't exercise this branch. With `applicationProvidedCancellationToken` pre-cancelled in the test, control exits through `await (Task<HedgingResponse>)completedTask;` at line ~216, never reaching the `lastException != null` branch at lines 285-292. A future revert here would have zero coverage.
2. The inline comment says this protects "the throw-vs-throw-ex preservation work in `CloneAndSendAsync` / `RequestSenderAndResultCheckAsync`" — but `lastException` here is sourced from `completedTask.Exception.Flatten().InnerExceptions.FirstOrDefault()`, so the hazard being addressed is actually `throw lastException;` itself at this site, not anything in the catches below.

**Suggested fix:** either add a small test that drives the all-hedge-tasks-faulted path (uncancelled `applicationProvidedCancellationToken`, all hedges throw, `while (requestTasks.Any())` accumulates `lastException`, assert sender's frame survives in the surfaced stack), or mention in the PR description that this third change is intentionally untested glue.

NaluTripician (commented):
Deep review pass. The core fix (single `Task.Yield()` in `CloneAndSendAsync`'s catch) is correctly placed at the middle layer of the hedging seam, the untyped catch scope is intentional, and the compiler-stored `ExceptionDispatchInfo` across the await preserves the original stack on the subsequent `throw;`. Three findings below — two are notable, one is a minor disclosure/coverage gap.

1. **Major** — the new regression test's claim of covering the `throw ex;` → `throw;` change in `RequestSenderAndResultCheckAsync` is empirically false (verified by reverting just that line and watching the test still pass). The pre-cancelled-CTS path routes through the OCE-filtered catch, not the generic catch where the fix lives, and `CosmosOperationCanceledException.StackTrace` delegates to the inner OCE's stack trace so the preservation is incidental, not caused by `throw;`. See inline comment.
2. **Major (merge-time)** — `changelog.md` targets `#### Bugs Fixed` under `### Unreleased`, but `main` (commit 8b03a5af0, PR #5864) replaced that with `#### Fixed` under `### Unreleased Preview`. Rebase needed or the entry lands in a heading that no longer exists. See inline comment.
3. **Minor** — the `throw lastException;` → `ExceptionDispatchInfo.Capture(...).Throw()` change in `ExecuteAvailabilityStrategyAsync` is correct but isn't called out in the PR description and isn't exercised by the new test. See inline comment.

Everything else checked out: scope is exactly the 3 advertised files, PR title matches the prlint regex, `postCountDelta > 0` is empirically robust (not flaky), and no disposal/cancellation races introduced.

changelog.md
kundadebdatta: On the changelog section: I think the rebase direction got inverted. Looking at #5864's diff, it actually went from Unreleased Preview/Fixed/Added to Unreleased/Features Added/Bugs Fixed/Other Changes — so the current main uses #### Bugs Fixed, which is where this entry already sits. The branch was also merged with latest main on top (commit 763866c8b). I think we're good as-is, but happy to re-check if I'm reading the diff wrong.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AvailabilityStrategyUnitTests.cs
kundadebdatta: Thanks for the catch on #1 and #3 — added a new test `SenderException_NonOce_AllHedgesFault_PreservesStackTraceThroughGenericCatchAndDispatchInfo` that drives both undertested paths (verified by reverting each fix individually — both produce a deterministic failure with the sentinel frame missing).

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
kundadebdatta: This is now addressed.

ananth7592 (approved):

NaluTripician (approved):

FabianMeiswinkel (approved):
LGTM

PR #5880, opened 2026-05-19
Client Encryption: Adds JsonProcessor knob and Newtonsoft/Stream test variations

adamnova:
## Summary
- reintroduce a typed `EncryptionOptions.JsonProcessor` opt-in while keeping the default `JsonProcessor.Newtonsoft`
- keep `RequestOptions.Properties["encryption-json-processor"]` working as the cross-cutting override, with explicit precedence over the typed property
- validate Newtonsoft and Stream side-by-side in unit tests and emulator roundtrip coverage, while keeping legacy encryption tests Newtonsoft-only by design
- update Encryption.Custom public API contract baselines for both net6 and net8

## Rationale
PR #5454 removed `EncryptionOptions.JsonProcessor` because the processor could already be supplied through `RequestOptions.Properties`, and that property-bag path still matters for decrypt/query request options that do not carry `EncryptionOptions`.

This change restores a typed, opt-in encrypt-side knob on `EncryptionOptions` for callers that want an explicit per-operation choice without changing the production default. The property defaults to `Newtonsoft`, and the existing property-bag override still wins when both are set so the existing decrypt/query override path remains the single consistent escape hatch across request option types.

## Test coverage
- `Microsoft.Azure.Cosmos.Encryption.Custom.Tests`
  - restored/added typed JsonProcessor coverage and precedence validation
  - fixed legacy test discovery so the Newtonsoft-only legacy roundtrips execute again
- `Microsoft.Azure.Cosmos.Encryption.Custom.EmulatorTests`
  - parameterized key create/read/RUD/transactional batch roundtrip tests across `Newtonsoft` and `Stream`
  - read/query Stream selection still uses `RequestOptions.Properties` so decrypt-side coverage stays aligned with the current public surface

## Validation
- `dotnet restore Microsoft.Azure.Cosmos.sln`
- `dotnet build Microsoft.Azure.Cosmos.Encryption.Custom\src\Microsoft.Azure.Cosmos.Encryption.Custom.csproj -c Release`
- `dotnet build Microsoft.Azure.Cosmos.Encryption.Custom\tests\Microsoft.Azure.Cosmos.Encryption.Custom.Tests\Microsoft.Azure.Cosmos.Encryption.Custom.Tests.csproj -c Release`
- `dotnet test Microsoft.Azure.Cosmos.Encryption.Custom\tests\Microsoft.Azure.Cosmos.Encryption.Custom.Tests\Microsoft.Azure.Cosmos.Encryption.Custom.Tests.csproj -c Release --nologo`
- `dotnet build Microsoft.Azure.Cosmos.Encryption.Custom\tests\EmulatorTests\Microsoft.Azure.Cosmos.Encryption.Custom.EmulatorTests.csproj -c Release`

## Notes
- Emulator tests were built but not executed locally.
- No failing Stream-pipeline variations were observed in the exercised unit coverage.

### DIFF
--- a/Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
+++ b/Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionOptions.cs
@@ -35,5 +35,17 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom
         /// Example of a path specification: /sensitive
         /// </summary>
         public IEnumerable<string> PathsToEncrypt { get; set; }
+
+        /// <summary>
+        /// Gets or sets the JSON processor to use for encryption of this operation.
+        /// </summary>
+        /// <remarks>
+        /// This controls the encrypt-side processor for the operation.
+        /// If <see cref="global::Microsoft.Azure.Cosmos.RequestOptions.Properties"/> contains the key "encryption-json-processor",
+        /// that value takes precedence over this setting.
+        /// Defaults to <see cref="JsonProcessor.Newtonsoft"/>.
+        /// <c>JsonProcessor.Stream</c> is available only in the .NET 8 package (net8.0 target framework).
+        /// </remarks>
+        public JsonProcessor JsonProcessor { get; set; } = JsonProcessor.Newtonsoft;
     }
 }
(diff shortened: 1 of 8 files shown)


adamnova:
Closing because the opt-in knob already exists via `RequestOptions.Properties["encryption-json-processor"]` (per PR #5454), and the side-by-side test pattern is already present in `MdeCustomEncryptionTests`, `EncryptionProcessorTests`, and `EncryptionBenchmark`. Re-scoping to add the missing parameterized variations using the existing mechanism. See the follow-up PR.

PR #5883, opened 2026-05-19
Direct: Fixes Direct/TCP connection limit validation

### LINKED ISSUE #5878: Direct: MaxTcpConnectionsPerEndpoint does not validate documented minimum value constraint
﻿## Description

`ConnectionPolicy.MaxTcpConnectionsPerEndpoint` documents that the value must be greater than or equal to 16 (default 65,535), but the setter is a plain auto-property with no validation:

`csharp
// ConnectionPolicy.cs:499-503
public int? MaxTcpConnectionsPerEndpoint
{
    get;
    set;
}
`

Similarly, `CosmosClientOptions.MaxTcpConnectionsPerEndpoint` just assigns and calls `ValidateDirectTCPSettings()`, which only checks that Direct-mode settings aren't used in Gateway mode - it does not validate the range.

This means values less than 16, zero, or negative can be set without any error, contradicting the documented contract.

## Expected behavior

The setter should validate that the value is greater than or equal to 16 (or whatever the actual minimum is) and throw `ArgumentOutOfRangeException` for invalid values, consistent with the documented contract.

## Locations

- `Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs:497-503`
- `Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs:659-666`

## Related

- The same pattern (documented constraint without code enforcement) may apply to `MaxRequestsPerTcpConnection` which documents a minimum of 4 but also appears to be unchecked.

orange-dot:
## Summary
- Validates `MaxRequestsPerTcpConnection` rejects values below the documented minimum of 4.
- Validates `MaxTcpConnectionsPerEndpoint` rejects values below the documented minimum of 16.
- Covers both `CosmosClientOptions` and `ConnectionPolicy` paths.

## Fixes
  - Fixes #5878

## Tests
- `DOTNET_ROLL_FORWARD=Major dotnet test Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Microsoft.Azure.Cosmos.Tests.csproj --filter "VerifyCosmosClientOptionsDirectTcpConnectionLimitsValidateMinimumValues|VerifyConnectionPolicyDirectTcpConnectionLimitsValidateMinimumValues" --no-restore /p:EnableSourceControlManagerQueries=false`
- `DOTNET_ROLL_FORWARD=Major dotnet test Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Microsoft.Azure.Cosmos.Tests.csproj --filter "CosmosClientOptionsUnitTests" --no-restore /p:EnableSourceControlManagerQueries=false`

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
+++ b/Microsoft.Azure.Cosmos/src/ConnectionPolicy.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Azure.Cosmos
         private const int defaultMediaRequestTimeout = 300;
         private const int defaultMaxConcurrentFanoutRequests = 32;
         private const int defaultMaxConcurrentConnectionLimit = 50;
+        private const int minimumMaxRequestsPerTcpConnection = 4;
+        private const int minimumMaxTcpConnectionsPerEndpoint = 16;
 
         internal UserAgentContainer UserAgentContainer;
         private static ConnectionPolicy defaultPolicy;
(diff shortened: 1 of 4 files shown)


orange-dot:

PR #5872, opened 2026-05-15
[Internal] DTS: Adds AbortDistributedTransaction signal on commit cancellation

Meghana-Palaparthi:
# Pull Request Template

## Description

This pull request improves the reliability and correctness of distributed transaction abort handling in the Cosmos DB SDK. It introduces a best-effort abort signal when a commit is canceled, ensures abort requests are handled consistently throughout the stack, and adds comprehensive tests to verify these behaviors.

When CommitTransactionAsync is cancelled while the in-flight HTTP request to the coordinator is in progress, the SDK now sends a AbortDistributedTransaction request using the same idempotency token so the coordinator can roll back any in-progress changes. The original OperationCanceledException always propagates to the caller; the abort signal is fire-and-forget with a configurable 30-second timeout.

**Distributed Transaction Abort Handling:**

* Added a best-effort abort signal in `DistributedTransactionCommitter`: when a commit operation is canceled, the code now sends an abort request to the coordinator to roll back any in-progress changes. This includes a new `TrySendAbortSignalAsync` method, a configurable abort timeout, and robust logging for abort outcomes. [[1]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074R25-R32) [[2]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074L43-R52) [[3]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074L132-R139) [[4]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074R151-R158) [[5]](diffhunk://#diff-8da7da32691191fb49279546f3936019368f6db95a04a7992f6fb1e065ffa074R180-R246)

**Core SDK and Protocol Updates:**

* Updated HTTP method and request handling to support `AbortDistributedTransaction` operations, ensuring they are recognized as distributed transaction requests and mapped to HTTP POST like commit operations. [[1]](diffhunk://#diff-2184809dc755140d4323b7008d49e34a21a962cfd634a8f990c66f307f4dbd6aR286) [[2]](diffhunk://#diff-3f93aff3daf641a0dd88add0fd676ec71e1af50c59404f98753a716e5901bb21R402) [[3]](diffhunk://#diff-142016342192e6f7886247b7e3c0ea6d1b92400a322a3c4b336b5a34867aa765R19-R25)

[...]

PR #5884, opened 2026-05-19
[Internal] DTS: Adds double-commit guardrail to prevent idempotency token bypass

Meghana-Palaparthi:
# Pull Request Template

## Description

Adds an Interlocked-based execute-once guard to DistributedWriteTransactionCore.CommitTransactionAsync() that throws InvalidOperationException on subsequent calls. Each CommitTransactionAsync() generates a new idempotency token (Guid), so calling it twice on the same instance would bypass server-side duplicate detection and risk a double-commit. The guard ensures only the first call proceeds.

Changes:
- DistributedWriteTransactionCore: Added 'commitCalled' field with Interlocked.CompareExchange guard
- DistributedTransaction: Updated CommitTransactionAsync xmldoc with remarks and exception documentation
- DistributedWriteTransactionTests: Added 3 tests covering sequential double-call, post-failure double-call, and concurrent race condition scenarios

## Type of change

## Closing issues

## Changelog

> Feature not public yet

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
@@ -23,20 +23,14 @@ namespace Microsoft.Azure.Cosmos
         /// </summary>
         /// <remarks>
         /// This method is single-use: it can only be called once per transaction instance.
-        /// Each call generates a unique idempotency token that the server uses for duplicate
-        /// detection during the SDK's internal retries. A second call would generate a new
-        /// token and bypass that server-side duplicate detection, risking a double-commit.
-        /// <para>
-        /// If the call fails for any reason (including transient network failures, cancellation,
-        /// or non-retriable server errors), the transaction instance is permanently consumed.
-        /// To retry, construct a new transaction with the same operations. When the previous
-        /// commit's outcome is unknown, verify the resulting state before retrying to avoid
-        /// duplicate writes.
-        /// </para>
+        /// If the call fails for any reason (including transient network failures or cancellation),
+        /// the transaction instance is permanently consumed. To retry, construct a new transaction
+        /// with the same operations.
         /// </remarks>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> to observe while waiting for the task to complete.</param>
         /// <returns>A <see cref="Task{TResult}"/> containing a <see cref="DistributedTransactionResponse"/> that represents the result of the transaction.</returns>
         /// <exception cref="InvalidOperationException">Thrown if <see cref="CommitTransactionAsync"/> has already been called on this instance.</exception>
+        /// <exception cref="OperationCanceledException">Thrown if <paramref name="cancellationToken"/> is cancelled before or during the commit.</exception>
         public abstract Task<DistributedTransactionResponse> CommitTransactionAsync(CancellationToken cancellationToken = default);
     }
 }
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
@@ -21,8 +21,22 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// Commits the distributed transaction.
         /// </summary>
+        /// <remarks>
+        /// This method is single-use: it can only be called once per transaction instance.
+        /// Each call generates a unique idempotency token that the server uses for duplicate
+        /// detection during the SDK's internal retries. A second call would generate a new
+        /// token and bypass that server-side duplicate detection, risking a double-commit.
+        /// <para>
+        /// If the call fails for any reason (including transient network failures, cancellation,
+        /// or non-retriable server errors), the transaction instance is permanently consumed.
+        /// To retry, construct a new transaction with the same operations. When the previous
+        /// commit's outcome is unknown, verify the resulting state before retrying to avoid
+        /// duplicate writes.
+        /// </para>
+        /// </remarks>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> to observe while waiting for the task to complete.</param>
         /// <returns>A <see cref="Task{TResult}"/> containing a <see cref="DistributedTransactionResponse"/> that represents the result of the transaction.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if <see cref="CommitTransactionAsync"/> has already been called on this instance.</exception>
         public abstract Task<DistributedTransactionResponse> CommitTransactionAsync(CancellationToken cancellationToken = default);
     }
 }
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionConstants.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionConstants.cs
@@ -8,6 +8,10 @@ namespace Microsoft.Azure.Cosmos
 
     internal static class DistributedTransactionConstants
     {
+        // Commit guard: values used with Interlocked.CompareExchange to enforce single-use semantics.
+        internal const int CommitNotStarted = 0;
+        internal const int CommitStarted = 1;
+
         // DTX envelope sub-status codes. The full status/body matrix and the inner-vs-outer retry ownership
         // is documented at the call site in ClientRetryPolicy.ShouldRetryDtxRequest.
         internal const int DtcCoordinatorRaceConflict = 5352; // 449 sub-status: coordinator ETag race
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedWriteTransactionTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedWriteTransactionTests.cs
@@ -387,6 +387,209 @@ namespace Microsoft.Azure.Cosmos.Tests
             Assert.IsFalse(response.IsSuccessStatusCode);
         }
 
+        // Double-commit guard
+
+        [TestMethod]
+        public async Task CommitAsync_CalledTwice_ThrowsInvalidOperationException()
+        {
+            Mock<CosmosClientContext> contextMock = this.BuildContextSetup();
+            contextMock
+                .Setup(c => c.ProcessResourceOperationStreamAsync(
+                    It.IsAny<string>(),
+                    It.IsAny<ResourceType>(),
+                    It.IsAny<OperationType>(),
+                    It.IsAny<RequestOptions>(),
+                    It.IsAny<ContainerInternal>(),
+                    It.IsAny<PartitionKey?>(),
+                    It.IsAny<string>(),
+                    It.IsAny<Stream>(),
+                    It.IsAny<Action<RequestMessage>>(),
+                    It.IsAny<ITrace>(),
+                    It.IsAny<CancellationToken>()))
+                .ReturnsAsync(BuildSuccessResponse(1));
+
+            DistributedWriteTransaction tx = new DistributedWriteTransactionCore(contextMock.Object)
+                .CreateItem(Database, Container, new PartitionKey("pk"), "item-id", new TestItem());
+
+            // First commit should succeed
+            DistributedTransactionResponse response = await tx.CommitTransactionAsync(CancellationToken.None);
+            Assert.IsTrue(response.IsSuccessStatusCode);
+
+            // Second commit must throw
+            InvalidOperationException ex = await Assert.ThrowsExceptionAsync<InvalidOperationException>(
+                () => tx.CommitTransactionAsync(CancellationToken.None));
+            Assert.IsTrue(ex.Message.Contains("already been called"));
+            Assert.IsTrue(ex.Message.Contains("construct a new DistributedWriteTransaction"),
+                "Error message should direct callers to the correct retry pattern.");
+        }
+
+        [TestMethod]
+        public async Task CommitAsync_CalledAfterFailedCommit_ThrowsInvalidOperationException()
+        {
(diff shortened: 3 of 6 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedWriteTransactionTests.cs
```diff
@@ -387,6 +387,209 @@ public async Task CommitAsync_ErrorResponse_ReturnsIsSuccessStatusCodeFalse()
...
+                        Interlocked.Increment(ref invocationCount);
+                        await Task.Delay(50, ct);
+                        return BuildSuccessResponse(1);
+                    });
+
+            DistributedWriteTransaction tx = new DistributedWriteTransactionCore(contextMock.Object)
+                .CreateItem(Database, Container, new PartitionKey("pk"), "item-id", new TestItem());
+
+            Task<DistributedTransactionResponse> task1 = tx.CommitTransactionAsync(CancellationToken.None);
+            Task<DistributedTransactionResponse> task2 = tx.CommitTransactionAsync(CancellationToken.None);
```
kushagraThapar: Could you please help me understand whether this test actually races the `Interlocked.CompareExchange`? Both `task1` and `task2` are created on the same thread, and since `Interlocked.CompareExchange(ref this.commitCalled, 1, 0)` is the very first statement of `CommitTransactionAsync` (before any `await`), `task1`'s `CompareExchange` completes **synchronously** before `task2` is even constructed. So the second invocation always sees `commitCalled == 1` regardless of atomicity — the test would still pass if the guard were replaced with a non-atomic `if (commitCalled == 1) throw; commitCalled = 1;`, which is exactly the bug class the test name promises to defend against.

A true race needs N worker tasks all blocked on a barrier and released simultaneously, so multiple threads genuinely hit `CompareExchange` concurrently. There's a nice in-repo precedent in `AvailabilityStrategyUnitTests.cs:560-602` using `ManualResetEventSlim`. Sketch:

```csharp
[TestMethod]
public async Task CommitAsync_ConcurrentCalls_OnlyOneSucceeds()
{
    int invocationCount = 0;
    Mock<CosmosClientContext> contextMock = this.BuildContextSetup();
    contextMock
        .Setup(c => c.ProcessResourceOperationStreamAsync(/* ... */))
        .Returns<...>(async (...) =>
        {
            Interlocked.Increment(ref invocationCount);
            await Task.Delay(50, ct);
            return BuildSuccessResponse(1);
        });

    DistributedWriteTransaction tx = new DistributedWriteTransactionCore(contextMock.Object)
        .CreateItem(Database, Container, new PartitionKey("pk"), "item-id", new TestItem());

    const int RacerCount = 16;
    using ManualResetEventSlim gate = new(initialState: false);
    Task<DistributedTransactionResponse>[] tasks = Enumerable.Range(0, RacerCount)
        .Select(_ => Task.Run(async () =>
        {
            gate.Wait();
            return await tx.CommitTransactionAsync(CancellationToken.None);
        }))
        .ToArray();

    gate.Set(); // release all racers simultaneously

    int successCount = 0;
    int rejectedCount = 0;
    foreach (Task<DistributedTransactionResponse> t in tasks)
    {
        try
        {
            await t;
            successCount++;
        }
        catch (InvalidOperationException)
        {
            rejectedCount++;
        }
    }

    Assert.AreEqual(1, successCount, "exactly one racer should win the CompareExchange");
    Assert.AreEqual(RacerCount - 1, rejectedCount, "all other racers should observe the gate as already-set");
    Assert.AreEqual(1, invocationCount, "the underlying commit pipeline should only fire once");
}
```

The `Assert.AreEqual(1, invocationCount, ...)` is the part that actually proves atomicity — without `Interlocked`, the test would flake on a torn read.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
```diff
@@ -21,8 +21,22 @@ abstract class DistributedTransaction
         /// <summary>
         /// Commits the distributed transaction.
         /// </summary>
+        /// <remarks>
```
kushagraThapar, line 24: Curious to learn — would you consider scoping this `<remarks>` block to `DistributedWriteTransaction` only, or alternatively enforcing the same single-use contract on `DistributedReadTransactionCore` too?

The new xmldoc here on the abstract base says *"This method is single-use … `<exception cref="InvalidOperationException">` Thrown if … already been called"*, and that's true for `DistributedWriteTransactionCore.CommitTransactionAsync`. But `DistributedReadTransactionCore.CommitTransactionAsync` (line 47 of that file) currently silently allows N invocations — a caller writing a helper against the abstract base type gets one behavior for write and another for read, with no compile-time signal.

Three resolutions in preference order — happy to go whichever way fits your design intent:

**(a) Template-method (most future-proof)** — seal `CommitTransactionAsync` on `DistributedTransaction`, run the guard centrally there, have subclasses override a protected `CommitTransactionCoreAsync` (same pattern `HttpClient.SendAsync` / `SendCoreAsync` uses). The uniform contract then auto-applies to read, write, and any future DTS subclass (abort, multi-region, batched commit), so `#5884`-style guards don't need to be added per-subclass in future PRs.

**(b) Add the same `Interlocked` guard to `DistributedReadTransactionCore`** — cheap; removes the abstract-base-vs-implementation asymmetry even if reads don't strictly need semantic protection.

**(c) Move the `<remarks>` and `<exception>` xmldoc down from `DistributedTransaction.cs` → `DistributedWriteTransactionCore.cs` (override-level)** — preserves the current behavior; just makes the contract truthful per concrete type.

If reads genuinely never benefit from the guard (e.g., the read path is naturally idempotent because no fresh token is generated per call), then (c) is the smallest and most honest change. If the team is planning to add `AbortTransactionAsync` or other per-call-token DTS methods soon (and #5872 suggests yes), then (a) is the most economical investment.

Not blocking — just curious which way the team is thinking.

kushagraThapar (dismissed):
Thanks @Meghana-Palaparthi — really nice, focused guardrail PR. The `Interlocked.CompareExchange(ref this.commitCalled, 1, 0) != 0` pattern is exactly right (returns the original value), the exception message is excellent (states what / why / how-to-recover / when-not-to-blindly-retry — this is template-worthy for the rest of the DTS surface), and the four "permanently-consume" branches (success, error response, network exception, cancellation) are all covered in the test matrix.

Approving — two suggestions inline, neither blocking:

1. The "ConcurrentCalls" test doesn't actually race the `CompareExchange` (both tasks run sequentially on the same thread before either hits `await`) — would you mind hardening it so it actually exercises concurrent racers?
2. The new `<remarks>` lives on the abstract `DistributedTransaction.CommitTransactionAsync`, but `DistributedReadTransactionCore.CommitTransactionAsync` doesn't enforce single-use. A caller writing a helper against the abstract base gets one behavior for write and another for read with no compile-time signal — could you please help me understand how you'd like to resolve this asymmetry?

Also wanted to flag a few small positives for the record:

- The "fail-once-and-done" semantics (the gate is NOT reset on failure) is the deliberate conservative choice, well-documented in both the exception message and the `<remarks>`. Different from `HttpClient.SendAsync` / `DbCommand.ExecuteNonQuery` (which let you retry after failure), but the right call here because the idempotency-token contract makes "did it succeed?" unknowable from the client side on a timeout/cancellation.
- The existing `CommitTransaction_SameBodyAndTokenSentOnEveryRetryAttempt` test in `DistributedTransactionCommitterTests.cs` already establishes the threat model this PR closes (one token reused across SDK-internal retries). Nice tie-in.
- No cross-SDK parity gap — Java and Python don't have DTS yet, so this is the canonical place to set the contract.

NaluTripician (dismissed):
Approving. The `Interlocked.CompareExchange` execute-once guard is correctly implemented — right field type (`int`, no `volatile` needed), right architectural layer above the committer's internal retry loop (which reuses one `IdempotencyToken` across attempts, so the guard doesn't interfere with legitimate retries), correct CAS semantics. `InvalidOperationException` matches the Cosmos lifecycle-guard convention (`ChangeFeedProcessorBuilder`, `ItemBatchOperation`, etc.). The threat model is real and well-explained in the xmldoc: each `CommitTransactionAsync` calls `new DistributedTransactionServerRequest(...)` which unconditionally sets `IdempotencyToken = Guid.NewGuid()`, so a naïve caller retry would bypass server-side duplicate detection. Consuming the transaction even on a failed first attempt is the correct call given the SDK can't distinguish "request never sent" from "response lost." Changelog skip is policy-compliant — the type is `internal` in the shipping NuGet and absent from all `API_*.txt` contracts.

A few non-blocking follow-ups worth considering in a later PR:

1. **Asymmetric enforcement of the documented contract.** The new `<remarks>` and `<exception cref="InvalidOperationException">` tags sit on the abstract `DistributedTransaction.CommitTransactionAsync` base method, so they apply universally — but only `DistributedWriteTransactionCore` enforces the guard. `DistributedReadTransactionCore.CommitTransactionAsync` has no `commitCalled` field and no CAS, so a read transaction called twice will issue two reads with two different idempotency tokens, contradicting the documented contract. Reads are idempotent at the data layer today so there's no immediate data-loss risk, but the asymmetry will silently become a real bug the moment the read path picks up any side effect. Two clean fixes: lift the guard into the abstract base via a template-method split (`sealed override` + `protected abstract …InternalAsync`) so the contract is honored universally, or narrow the xmldoc down to the Write override only.

2. **`CommitAsync_ConcurrentCalls_OnlyOneSucceeds` isn't actually concurrent.** Both `CommitTransactionAsync` calls run on the test thread; the first synchronous prefix (the CAS + `new DistributedTransactionCommitter(...)`) completes before the second call even starts, so the two CAS operations serialize deterministically through the `async` state machine. The whole purpose of `Interlocked.CompareExchange` over a plain `if (commitCalled == 0) { commitCalled = 1; … }` is correctness under a real cross-thread race — the current test would pass against a naïve non-atomic implementation just as easily. Wrap each call in `Task.Run` with a `Barrier(2).SignalAndWait()` to actually exercise the CAS. While you're in there, assert which task won and that exactly one succeeded (`task.Status == TaskStatus.RanToCompletion && task.Result.IsSuccessStatusCode`) vs. faulted with IOE — today the test only checks that *some* IOE was caught and that `invocationCount == 1`, so a regression where the first call threw and the second faulted with IOE would still pass.

3. **Operation-builder methods remain mutable after commit.** `CreateItem`, `UpsertItem`, `ReplaceItem`, `DeleteItem`, `PatchItem` (and the `*Stream` variants) don't check `commitCalled`, so a consumed transaction will happily accept new operations and return `this`. The next commit attempt throws, but the object visibly behaves like a fresh transaction in the meantime. Given the new xmldoc explicitly tells users to "construct a new `DistributedWriteTransaction`", consider gating the builders too — or at least calling this out in the docs so it isn't a surprise.

Minor nits:
- The exception-message substring assertions (`ex.Message.Contains("already been called")`, etc.) are brittle relative to team precedent in #5793, which asserts exception type only. Either drop the substring asserts or extract the message to a `private const string` in the SUT and reference it from both sides.
- Magic `0`/`1` in `Interlocked.CompareExchange(ref this.commitCalled, 1, 0)` would read more cleanly with named constants (`CommitNotCalled = 0`, `CommitCalled = 1`).
- `commitCalled` is a past-participle; the file right next door (`DistributedTransactionResponse.cs`) uses `isDisposed`, and Cosmos lifecycle flags generally follow `is*` (e.g., `isBuilt`, `isStarted`). `isCommitInvoked` would be more consistent.
- The new `<exception>` xmldoc documents only `InvalidOperationException`, but `CommitTransactionAsync` clearly also surfaces `OperationCanceledException` (one of the new tests asserts this) plus transport/`CosmosException` failures. Worth adding for intellisense honesty.
- PR description says "Added 3 tests" but the diff adds 5 (cancellation + transient-exception tests are missing from the body summary).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedWriteTransactionTests.cs
Meghana-Palaparthi: Fixed this to actually test a race consition.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransaction.cs
Meghana-Palaparthi: Refactored this. Also, there was no double-commit blocking on read transactions. That was a miss. I added it to be consistent across distributed transactions.

kushagraThapar (dismissed):

Meghana-Palaparthi:

kushagraThapar (approved):

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionConstants.cs
```diff
@@ -8,6 +8,10 @@ namespace Microsoft.Azure.Cosmos
 
     internal static class DistributedTransactionConstants
     {
+        // Commit guard: values used with Interlocked.CompareExchange to enforce single-use semantics.
+        internal const int CommitNotStarted = 0;
```
jeet1995, line 12: Curious why `int` was chosen and not `bool`?

jeet1995 (approved):

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionConstants.cs
Meghana-Palaparthi: `int` was chosen because it's used with I`nterlocked.CompareExchange`. That API doesn't have an overload for `bool`. If I wanted to use bool, I'd have to use `lock` instead. In this commit guard, `Interlocked.CompareExchange` is best because it guarantees only one caller can flip “not started” to “started” in a single atomic action, so duplicate commits are blocked without taking a lock.

PR #5885, opened 2026-05-19
[Internal] DTS: Adds session-token support and typed response deserialization for transactions

Meghana-Palaparthi:
# Pull Request Template

## Description

This pull request introduces several improvements and new features to the distributed transactions API in the Azure Cosmos DB SDK, with a focus on enhancing typed deserialization of operation results, streamlining session token handling, and improving documentation and usability. The most important changes are outlined below:

**Typed Deserialization and Result Handling:**
- Added a generic `DistributedTransactionOperationResult<T>` class, enabling easy access to deserialized resources from distributed transaction operations. This is returned by the new `GetOperationResultAtIndex<T>(int index)` method on `DistributedTransactionResponse`, which safely deserializes the resource body into the requested type without affecting the underlying stream. [[1]](diffhunk://#diff-31f1a4f3f77bbf984d36bb6ba94c05c0013b4f52e38334787ed61f2418c60334R246-R279) [[2]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87R81-R121)
- Introduced a `CreateSnapshot` utility method to produce independent copies of resource streams, ensuring that repeated deserialization or direct stream access remain safe and side-effect free.

**Session Token Management:**
- Added a `SessionToken` property to `DistributedTransactionRequestOptions`, allowing each operation in a distributed transaction to specify its own session token for partition-level consistency. This is now set when constructing a `DistributedTransactionOperation`. [[1]](diffhunk://#diff-5c56bac65a7d03185857db755486c5cd5fde48f373e0d28354c450a28b7af30eR18-R33) [[2]](diffhunk://#diff-30fceb3d3e008bbde11f8dd7453a6dcef707a5452f9e235a858068b93dd5e1dbR36)
- Updated documentation to clarify how session tokens should be obtained and used for distributed transactions.

**API Usability and Documentation:**
- Improved XML documentation for distributed read transactions, clarifying snapshot isolation semantics and providing guidance on result deserialization with the new APIs.
- Added remarks to `ResourceStream` property documentation to clarify ownership and correct usage patterns.

**Internal Consistency and Plumbing:**
- Ensured that the serializer is consistently propagated and available for deserialization in all relevant operation result and response objects. [[1]](diffhunk://#diff-31f1a4f3f77bbf984d36bb6ba94c05c0013b4f52e38334787ed61f2418c60334R43) [[2]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87R384) [[3]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L402-R445)

**Minor Improvements:**
- Made the `CommitTransactionAsync` method's `CancellationToken` parameter optional for improved developer ergonomics.

## Type of change

## Closing issues

## Changelog

> Feature not public yet

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
@@ -113,6 +113,7 @@ namespace Microsoft.Azure.Cosmos
         public virtual Stream ResourceStream { get; internal set; }
 
         /// <summary>
+        /// Gets the number of request units consumed by this operation.
         /// </summary>
         [JsonInclude]
         [JsonPropertyName("requestCharge")]
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
@@ -40,6 +40,7 @@ namespace Microsoft.Azure.Cosmos
             this.RequestCharge = other.RequestCharge;
             this.ActivityId = other.ActivityId;
             this.Trace = other.Trace;
+            this.SerializerCore = other.SerializerCore;
         }
 
         /// <summary>
@@ -102,6 +103,12 @@ namespace Microsoft.Azure.Cosmos
         /// Gets the resource stream associated with the operation result.
         /// The stream contains the raw response payload returned by the operation.
         /// </summary>
+        /// <remarks>
+        /// <b>Lifetime / ownership:</b> This stream is owned by the enclosing
+        /// <see cref="DistributedTransactionResponse"/> and will be disposed when that response is disposed.
+        /// Do not dispose it directly. To deserialize to a typed object, use
+        /// <see cref="DistributedTransactionResponse.GetOperationResultAtIndex{T}"/>.
+        /// </remarks>
         [JsonIgnore]
         public virtual Stream ResourceStream { get; internal set; }
 
@@ -135,10 +142,53 @@ namespace Microsoft.Azure.Cosmos
         [JsonIgnore]
         internal ITrace Trace { get; set; }
 
+        [JsonIgnore]
+        internal CosmosSerializerCore SerializerCore { get; set; }
+
         private static readonly JsonSerializerOptions CaseInsensitiveOptions = new JsonSerializerOptions
         {
             PropertyNameCaseInsensitive = true,
         };
+
+        /// <summary>
+        /// Returns a fresh <see cref="Stream"/> that exposes the same bytes as <paramref name="source"/>
+        /// without affecting the source stream's position or lifetime.
+        /// </summary>
+        internal static Stream CreateSnapshot(Stream source)
+        {
+            // Fast path for standard MemoryStream.
+            if (source is MemoryStream ms
+                && ms.GetType() == typeof(MemoryStream)
+                && ms.TryGetBuffer(out ArraySegment<byte> buffer))
+            {
+                return new MemoryStream(
+                    buffer: buffer.Array,
+                    index: buffer.Offset,
+                    count: buffer.Count,
+                    writable: false,
+                    publiclyVisible: true);
+            }
+
+            // Fallback for other stream types. For seekable streams the position is reset
+            // before and after copying. Non-seekable streams are copied from their current
+            // position, which is acceptable here because the only producer of ResourceStream
+            // is a freshly-created MemoryStream at position 0 (see FromJson).
+            if (source.CanSeek)
+            {
+                source.Position = 0;
+            }
+
+            MemoryStream copy = new MemoryStream();
+            source.CopyTo(copy);
+            copy.Position = 0;
+
+            if (source.CanSeek)
+            {
+                source.Position = 0;
+            }
+
+            return copy;
+        }
 
         /// <summary>
         /// Creates a <see cref="DistributedTransactionOperationResult"/> from a JSON element.
@@ -192,5 +242,39 @@ namespace Microsoft.Azure.Cosmos
 
             return result;
         }
+    }
+
+    /// <summary>
+    /// Represents a typed result for a specific operation that was part of a <see cref="DistributedTransaction"/> request.
+    /// </summary>
+    /// <typeparam name="T">The type to which the resource body was deserialized.</typeparam>
+#pragma warning disable SA1402 // File may only contain a single type
+#if INTERNAL
+    public
+#else
+    internal
+#endif
+    class DistributedTransactionOperationResult<T> : DistributedTransactionOperationResult
+#pragma warning restore SA1402 // File may only contain a single type
+    {
+        internal DistributedTransactionOperationResult(DistributedTransactionOperationResult result, T resource)
+            : base(result)
+        {
+            this.Resource = resource;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DistributedTransactionOperationResult{T}"/> class.
+        /// </summary>
+        protected DistributedTransactionOperationResult()
+        {
+        }
+
+        /// <summary>
+        /// Gets the resource deserialized from the operation response body.
+        /// Returns <c>default</c> when the server did not return a resource body for the operation
+        /// (for example, a delete operation or a failed operation).
+        /// </summary>
+        public virtual T Resource { get; set; }
     }
 }
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
@@ -113,25 +113,17 @@ namespace Microsoft.Azure.Cosmos
         public virtual Stream ResourceStream { get; internal set; }
 
         /// <summary>
-        /// Request charge in request units for the operation.
         /// </summary>
         [JsonInclude]
         [JsonPropertyName("requestCharge")]
         public virtual double RequestCharge { get; internal set; }
 
-        [JsonIgnore]
-        internal virtual SubStatusCodes SubStatusCode { get; set; }
-
         /// <summary>
-        /// Gets the sub-status code value as an unsigned integer.
+        /// Gets the sub-status code for the operation.
         /// </summary>
         [JsonInclude]
         [JsonPropertyName("subStatusCode")]
-        public virtual uint SubStatusCodeValue
-        {
-            get => (uint)this.SubStatusCode;
-            internal set => this.SubStatusCode = (SubStatusCodes)value;
-        }
+        public virtual SubStatusCodes SubStatusCode { get; internal set; }
 
         /// <summary>
         /// ActivityId related to the operation.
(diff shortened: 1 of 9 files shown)


NaluTripician (dismissed):
Looks good — approving. Implementation of the outbound session-token plumbing and the typed `GetOperationResultAtIndex<T>` / `DistributedTransactionOperationResult<T>` is correct, well-scoped, and well-tested. Nothing blocking from me.

A few non-blocking nits / follow-ups to consider:

- **Read-op session-token coverage.** All four new committer tests use `OperationType.Create`, but the PR description leads with Read consistency as the primary use case. A single `DistributedReadTransactionCore` test that sets `SessionToken` on a Read op and asserts it reaches the request body would close the most visible coverage gap.
- **Deserialization-failure path on `GetOperationResultAtIndex<T>`.** Happy path, no-body, no-serializer, OOB index, post-Dispose, and called-twice are all covered, but there's no test (or `<exception>` doc) for the case where the body JSON cannot be deserialized to `T`. Worth pinning the exception type with one test and an `<exception>` element on the XML doc.
- **Dispose hygiene in the new `DistributedTransactionResponseTests`.** Most of the new tests create a `DistributedTransactionResponse` and never dispose it, which is inconsistent with the dispose-specific tests in the same file and with `IDisposable` hygiene in general. Prefer `using DistributedTransactionResponse response = ...` across all the new ones.
- **`TestDocument` attribute mismatch.** The test type uses `[System.Text.Json.Serialization.JsonPropertyName]`, but the runtime serializer is `CosmosJsonDotNetSerializer` (Newtonsoft). The attributes are silently ignored — tests pass only because property casing happens to match. Either drop the attributes or switch to `[Newtonsoft.Json.JsonProperty]` so the intent is honest.
- **`DistributedTransactionOperationResult<T>.Resource` setter.** Currently `{ get; set; }`. Every other property on the base class is `{ get; internal set; }` — tightening this to `internal set` would match the existing pattern.
- **Redundant `ThrowIfDisposed()` in `GetOperationResultAtIndex<T>`.** The indexer call on the next line already does this check. Harmless, but worth either removing or leaving a one-line comment noting the outer call is deliberate.
- **`SessionToken` doc on `DistributedTransactionRequestOptions`.** Doc requires `{partitionKeyRangeId}:{lsn}` format, but the SDK doesn't validate — would be helpful to call out that an invalid format will surface as a server-side error.

None of these block. Nice work.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionRequestOptions.cs
```diff
@@ -14,5 +15,21 @@ namespace Microsoft.Azure.Cosmos
...
+        /// <remarks>
+        /// Because a distributed transaction spans multiple partitions, each operation must supply its own session token
+        /// rather than a single commit-level token.  The token is serialized into the request body alongside the operation
+        /// and used by the server to enforce read-your-own-writes consistency for that specific partition.
+        /// <para>
+        /// Obtain the token from a prior <see cref="DistributedTransactionOperationResult.SessionToken"/> or from
+        /// the session token surface of another SDK response that targeted the same partition.
+        /// </para>
+        /// </remarks>
+        public string SessionToken { get; set; }
```
kushagraThapar, line 34: **M1 — DTS session-token plumbing as a parallel track**

I'm curious to learn how you and the team are thinking about long-term convergence here.

This PR establishes a new outbound contract: per-operation session tokens travel in the JSON body, where the rest of the SDK (point ops, batch, query) sends them in the `x-ms-session-token` header. On the inbound side, DTS uses `SessionContainer.SetSessionToken(collectionRid, fullName, headers)` — the overload that **bypasses** `ShouldUpdateSessionToken` — while the request-path uses the `(DSR, headers)` overload that gates on it. The bypass overload is now used by exactly one caller: DTS.

The MAX-merge invariant from `cosmosdb-design-docs/16-session-tokens.md#token-merge-logic` is preserved transitively, but the 2-3-year maintenance bill is real: any future hardening on the request path (new `ShouldUpdateSessionToken` rules, telemetry, distributed cache integration) will need to be plumbed twice unless the two paths converge.

Three options, in increasing-now-cost / decreasing-later-cost order:
1. Pre-GA: route DTS per-op tokens through per-operation headers (multi-valued or structured) so the request-path flow handles ingestion uniformly. Needs server cooperation.
2. Pre-GA: refactor `MergeSessionTokens` to construct a synthetic DSR per operation and call `(DSR, headers)`, so `ShouldUpdateSessionToken` applies and the bypass overload can eventually be removed.
3. Document the divergence explicitly so the parallel track is at least visible to the next maintainer (cheapest now, costliest later).

Could you please help me understand the team's thinking on which of these (or which alternative) is the target for the GA cut? Non-blocking — purely a supportability question for the record.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -135,11 +142,54 @@ public virtual uint SubStatusCodeValue
         [JsonIgnore]
```
kushagraThapar, line 142: **M2 + M3 — public surface of `DistributedTransactionOperationResult` diverges from `TransactionalBatchOperationResult` peer; `SubStatusCodeValue` is the only `public uint` sub-status in the SDK**

I wanted to flag two coupled public-surface decisions that lock in together once `#if INTERNAL` lifts. Both are non-blocking; raising now because they're cheapest-to-fix before any non-preview customer takes a dependency:

**M2 — Visibility divergence from the batch peer.** On `TransactionalBatchOperationResult`, `SessionToken` (L99), `PartitionKeyRangeId` (L104), `RequestCharge` (L89), and `SubStatusCode` (L94) are all **`internal virtual`**. On DTS, all four are **`public virtual { get; internal set; }`**. This PR establishes the convention that per-operation session tokens, physical-partition IDs, and request charges are part of the public response surface. Once a single non-preview customer takes a dependency, none of these can be narrowed. Would you mind sharing whether the divergence from batch is deliberate (e.g., customer scenarios DTS needs to enable that batch doesn't)? If yes, it'd be great to capture that in an XML `<remarks>` block so the choice is durable; if no, aligning to `internal virtual` for now keeps options open.

**M3 — `public virtual uint SubStatusCodeValue` is the only `uint` sub-status surface in the SDK.** `CosmosException.SubStatusCode` (line 101) is `int`; `TransactionalBatchOperationResult.SubStatusCode` (line 94) is `SubStatusCodes`. Customer code flowing DTS results into existing logging / exception-correlation pipelines will need a cast at every site, and `uint` doesn't compose with `(int)SubStatusCodes.XYZ` constants without a cast either. Cross-SDK lens: Rust uses `u32` (aligns); Java + Python both use signed int (diverges). Two pre-GA options that are non-breaking now but breaking after GA:
1. Expose `public virtual int SubStatusCode` (matching `CosmosException`) and keep the `uint`-backed `SubStatusCodeValue` `internal` for STJ binding only.
2. Expose `public virtual SubStatusCodes SubStatusCode` directly (the enum already round-trips from `int` via `(SubStatusCodes)value`).

Either is one-line; both become breaking changes after the preview lifts. `uint` is the shape that's hardest to back out of.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionRequestOptions.cs
```diff
@@ -14,5 +15,21 @@ namespace Microsoft.Azure.Cosmos
 #endif
     class DistributedTransactionRequestOptions : RequestOptions
     {
+        /// <summary>
+        /// Gets or sets the session token for this individual operation.
+        /// </summary>
+        /// <value>
+        /// The session token, in <c>{partitionKeyRangeId}:{lsn}</c> format, for the partition targeted by this operation.
```
kushagraThapar: **M4 — XML doc claims V1 session-token wire format; design-doc canonical is V2 (vector)**

Could you please help me understand the choice to document the wire format as V1 (`{partitionKeyRangeId}:{lsn}`)?

The design-doc canonical (`cosmosdb-design-docs/16-session-tokens.md#token-format` + `cosmosdb-design-docs/05-replication-consistency.md#vectorsessiontoken--deep-dive`) is V2: `pkRangeId:version#globalLSN#region=lsn…`, which is what `SessionTokenHelper.Parse` actually accepts (verified at `SessionContainer.cs:314`). Once customers internalize "the SDK wants `{pkRangeId}:{lsn}`" and pass V1 strings on a multi-master account, the V1 token loses all per-region progress on merge — and correcting the doc later doesn't undo the bad mental model.

Two cheap pre-GA fixes:
1. Reword to "an opaque session token previously returned by the SDK; do not parse or construct manually" — matches the spirit of `TransactionalBatchRequestOptions.SessionToken` which doesn't promise any format.
2. Document both V1 and V2 explicitly.

(Same V1 claim is echoed at `DistributedTransactionOperationResult.cs:222` and on the commit-side comments at `DistributedTransactionCommitter.cs:185, 207` — would be great to align all four.)

kushagraThapar (commented):
Hi @Meghana-Palaparthi — thanks for this thorough PR. The implementation is correct, the snapshot-based multi-call story on `GetOperationResultAtIndex<T>` is genuinely better than the peer batch behavior, and @NaluTripician's prior approval covers the bulk of the correctness side cleanly.

I wanted to layer in one extra angle that we're explicitly trying to call out on every Cosmos PR going forward: **supportability and long-term breaking-change risk for surfaces that lock in once `#if INTERNAL` lifts**. Below are the three highest-leverage things that I think are worth thinking about before this preview eventually GAs — all of them are non-blocking, and several are one-line fixes today that would otherwise be breaking changes after release.

1. **M1 — DTS / non-DTS session-token plumbing convergence** (inline on `DistributedTransactionRequestOptions.cs`) — single highest 2-3-year cost decision in this PR.
2. **M2 + M3 — public surface of `DistributedTransactionOperationResult` diverges from `TransactionalBatchOperationResult` peer; `SubStatusCodeValue` is the only `public uint` sub-status in the SDK** (inline on `DistributedTransactionOperationResult.cs`).
3. **M4 — XML doc claims V1 session-token wire format; design-doc canonical is V2 (vector)** (inline on `DistributedTransactionRequestOptions.cs`).

Leaving this as a `COMMENT` review rather than an approval — @NaluTripician already has the correctness side covered; I'm just adding the supportability lens to the public record.

(We also have 8 minor + 9 nit items in our internal review notes — happy to surface any inline if helpful. Mainly: a typed `IEnumerable<DistributedTransactionOperationResult<T>>` projection, a `[JsonExtensionData]` forward-compat catch-all, decoupling C# names from wire field names via a private DTO, and a proposed extension to `cosmosdb-design-docs/16-session-tokens.md` covering DTS as the first surface using body-based session tokens.)

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
Meghana-Palaparthi: All the properties are public for STJ deserialization to work. It throws `InvalidOperationException` when included properties are not `public`. I agree this isn't enough reason to expose these properties to the customer. I'll change these to `internal` and move to manual deserialization in a follow-up PR

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionRequestOptions.cs
Meghana-Palaparthi: It does make sense to use the SetSessionToken overload with (DSR, headers) to avoid maintenance at two places. I'll update this in a follow-up PR

kushagraThapar (dismissed):
Thanks @Meghana-Palaparthi — really appreciate the quick turnaround on the doc fix and for opening #5896 to cleanly handle the public-surface / manual-deserialization work in a separate PR. Approving. 🙏

jeet1995 (dismissed):

Meghana-Palaparthi:

kushagraThapar (approved):

jeet1995 (approved):

PR #5890, opened 2026-05-19
[Internal] Thin Client Integration: Add tests supporting Read consistency strategy in thinclient mode.

aavasthy:
# Pull Request Template

## Description

ReadConsistencyStrategy was previously supported in Gateway and Direct connectivity modes. This PR extends that support to requests routed through the Thin Client proxy, so applications using Gateway mode with the Thin Client endpoint get the same per-request and per-client control over read behavior (Eventual, Session, LatestCommitted, GlobalStrong).

The x-ms-cosmos-read-consistency-strategy header set by RequestInvokerHandler flows through the existing pipeline: GatewayStoreClient.PrepareRequestMessageAsync copies it from the DocumentServiceRequest onto the outbound HttpRequestMessage (allowed via the x-ms-* filter), and ThinClientTransportSerializer.SerializeProxyRequestAsync rebuilds a RequestNameValueCollection from those HTTP headers — which populates the strongly-typed  eadConsistencyStrategy property — before calling the Direct package's TransportSerialization.BuildRequestForProxy. 

## Type of change

## Changelog

>


changelog.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Features Added
 
 - [5838](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5838) EmbeddingGenerator: Adds ICosmosEmbeddingGenerator client-wide configuration (preview)
+- [5890](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5890) ThinClient: Adds Read Consistency Strategy support for requests routed through the Thin Client proxy.
```
kushagraThapar: nit - this changelog entry should be in the previously released version 3.60.0 (as discussed offline). 
Worth adding the TCP keep alive changelog entry as well as you pointed out, thanks!

kushagraThapar (dismissed):
LGTM with one suggestion, thanks!

kushagraThapar (approved):
LGTM, thanks!

kundadebdatta (approved):
LGTM

PR #5893, opened 2026-05-20
[Internal] PRLint: Refactors regex failure comment format in prlint.yml

yash2710:
## Description

Updated the comment format for regex failure messages and examples in the PR lint workflow.

## Type of change

## Changelog

> Internal CI comment changes for build error clarity

### DIFF
--- a/.github/workflows/prlint.yml
+++ b/.github/workflows/prlint.yml
@@ -18,11 +18,13 @@ jobs:
         on-failed-regex-fail-action: true
         on-failed-regex-request-changes: false
         on-failed-regex-create-review: true
-        on-failed-regex-comment: >
-         Please follow the required format: \"[Internal] Category: (Adds|Fixes|Refactors|Removes) Description\"<br /><br />
+        on-failed-regex-comment: |
+         Please follow the required format: \"[Internal] Category: (Adds|Fixes|Refactors|Removes) Description\"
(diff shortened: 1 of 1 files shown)


kushagraThapar (commented):
Nice cleanup — switching from `>` + `<br />` to `|` is much more readable. Two small follow-ups I'd suggest folding into this PR, since they're in the same lines and would deliver the full "comment now renders cleanly on GitHub" outcome:

1. **Drop the backslash escapes in `\"[Internal] Category: ...\"`.** Backslash escapes (`\"`) are only interpreted inside YAML *double-quoted* strings — in `|` (or `>`) block scalars they pass through literally, so the rendered comment shows `\"…\"` with visible backslashes. Either remove the backslashes (`"[Internal] Category: ..."`) or wrap the format spec in backticks for inline-code rendering.

2. **Wrap the regex in backticks** so GitHub Markdown doesn't try to interpret characters like `|`, `?`, `.`, `[`, `]`:
   ```yaml
   Regex: `(\[Internal\]|\[v4\] )?.{3}.+: (Adds|Fixes|Refactors|Removes) .{3}.+`
   ```

Nit: the new blank line between the regex and the `[Internal]` paragraph has 9 trailing spaces — harmless in YAML, but some lint configs flag it.

Otherwise LGTM 👍

kushagraThapar (approved):
LGTM, thanks @yash2710 !

ananth7592 (approved):
LGTM

PR #5895, opened 2026-05-20
[Internal] Distributed Transaction: Refactors APIs to accept Container reference

Meghana-Palaparthi:
## Description

Replaces the `(string database, string collection)` parameter pair on every `DistributedWriteTransaction` / `DistributedReadTransaction` abstract method with a single `Container` reference  consistent with the rest of the v3 SDK where operations target a `Container` instance.

This is a hard signature swap (no overloads). Safe because all DTC APIs are gated behind `#if INTERNAL` and have no shipped customer surface.

## Changes

**Source (4 files):**
- `DistributedWriteTransaction`: 8 abstract methods updated
- `DistributedReadTransaction`: `ReadItem` updated
- `*Core` overrides extract `container.Database.Id` / `container.Id` at the entry point; downstream pipeline (`DistributedTransactionOperation`, committer, serializer, wire format) intentionally untouched
- `ValidateContainerReference(Container)` checks: `container != null`, `container.Id` non-empty, `container.Database != null`, `container.Database.Id` non-empty

**Tests (4 files):**
- `DistributedWriteTransactionTests`, `DistributedReadTransactionCoreTests`: rebuilt with `BuildMockContainer()` Moq helper; 14 new null/empty/whitespace validation test cases added
- `DistributedTransactionSerializerTests`, emulator `DistributedTransactionTests`: call sites updated to pass `Container` directly

## Validation
- Full solution build: 0 errors
- Unit tests: 148/148 DTC tests pass (was 134 before validation additions)
- Emulator tests: compile clean; not run locally (require Windows Cosmos Emulator)

## Changelog

No changelog entry required  the affected APIs are `#if INTERNAL` and are not part of the customer-visible public surface in any released SDK build.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedReadTransaction.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedReadTransaction.cs
@@ -24,15 +24,13 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// Adds a read operation to the distributed transaction.
         /// </summary>
-        /// <param name="database">The name of the database containing the container.</param>
-        /// <param name="collection">The name of the container where the item exists.</param>
+        /// <param name="container">The <see cref="Container"/> reference where the item exists.</param>
         /// <param name="partitionKey">The partition key for the item.</param>
         /// <param name="id">The unique identifier of the item to read.</param>
         /// <param name="requestOptions">Options for the read operation.</param>
         /// <returns>The current <see cref="DistributedReadTransaction"/> instance for method chaining.</returns>
         public abstract DistributedReadTransaction ReadItem(
-            string database,
-            string collection,
+            Container container,
             PartitionKey partitionKey,
             string id,
             DistributedTransactionRequestOptions requestOptions = null);
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedReadTransactionCore.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedReadTransactionCore.cs
@@ -27,15 +27,15 @@ namespace Microsoft.Azure.Cosmos
             string id,
             DistributedTransactionRequestOptions requestOptions = null)
         {
-            DistributedReadTransactionCore.ValidateContainerReference(container);
+            (string databaseId, string containerId) = DistributedTransactionConstants.ValidateAndUnpackContainer(container, this.clientContext.Client);
             DistributedReadTransactionCore.ValidateItemId(id);
 
             this.operations.Add(
                 new DistributedTransactionOperation(
                     operationType: OperationType.Read,
                     operationIndex: this.operations.Count,
-                    database: container.Database.Id,
-                    container: container.Id,
+                    database: databaseId,
+                    container: containerId,
                     partitionKey: partitionKey,
                     id: id,
                     requestOptions: requestOptions));
@@ -51,35 +51,6 @@ namespace Microsoft.Azure.Cosmos
                 clientContext: this.clientContext);
 
             return await committer.CommitTransactionAsync(cancellationToken);
-        }
-
-        private static void ValidateContainerReference(Container container)
-        {
-            if (container == null)
-            {
-                throw new ArgumentNullException(nameof(container));
-            }
-
-            if (string.IsNullOrWhiteSpace(container.Id))
-            {
-                throw new ArgumentException(
-                    "Container reference must have a non-empty Id.",
-                    nameof(container));
-            }
-
-            if (container.Database == null)
-            {
-                throw new ArgumentException(
-                    "Container reference must expose a non-null Database.",
-                    nameof(container));
-            }
-
-            if (string.IsNullOrWhiteSpace(container.Database.Id))
-            {
-                throw new ArgumentException(
-                    "Container reference must have a non-empty Database.Id.",
-                    nameof(container));
-            }
         }
 
         private static void ValidateItemId(string id)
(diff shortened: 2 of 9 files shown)


kushagraThapar (dismissed):
## ✅ Approving — clean execution of the prior design-doc feedback

This PR lands the **#1 critical recommendation** from the review of design doc #5877: swap `(string database, string collection)` for `Container`. The shape now matches the rest of the v3 SDK (`Container.CreateItemAsync`, `TransactionalBatch`), the hard signature swap is the right call given `#if INTERNAL` gating, the validation pattern is BCL-correct (`ArgumentNullException` only for `container == null`; `ArgumentException` with `paramName = nameof(container)` for property-level problems), and the 14 new validation tests + `BuildMockContainer()` helper give the contract solid coverage. Approving.

Leaving two suggestions below that I'd love to see folded into this PR — both are small, contained, and prevent us from cementing a slightly misleading contract. Everything else (XML doc wording, factory location revisit, idempotency token, payload limits, atomic read-modify-write, etc.) is being tracked separately as follow-up work.

---

### 🔴 Suggestion 1 — Validate that the `Container` belongs to the same `CosmosClient` (high-value, ~5 lines)

The pipeline below the API surface is transaction-owner-scoped, not container-scoped:
- `CosmosClient.CreateDistributedWriteTransaction()` binds the transaction to **that** client's `ClientContext`.
- The committer uses `this.clientContext.GetCachedContainerPropertiesAsync(...)` (`DistributedTransactionCommitterUtils.cs:30`), `this.clientContext.SerializerCore`, and `this.clientContext.DocumentClient?.sessionContainer`.
- The new `Container` parameter is reduced to `container.Database.Id` + `container.Id` at the entry point; the rest is discarded.

So today:
```csharp
DistributedWriteTransaction tx = clientA.CreateDistributedWriteTransaction();
tx.CreateItem(containerFromClientB, pk, id, body);  // silently routed through clientA
```
…the SDK uses `clientA`'s credentials/account/serializer/session/encryption without warning. This is arguably worse than the string version because the `Container` object visually implies the customer's client/pipeline is honored.

Fix is small. Add to `ValidateContainerReference` (both `DistributedWriteTransactionCore` and `DistributedReadTransactionCore`), and pass `this.clientContext.Client` in:

```csharp
// Verified accessibility: Database.Client is public abstract (Database.cs:32);
// CosmosClientContext.Client is internal abstract (CosmosClientContext.cs:26).
// Both reachable from *Core (same assembly).
if (!object.ReferenceEquals(container.Database.Client, this.clientContext.Client))
{
    throw new ArgumentException(
        "Container must belong to the same CosmosClient instance that created this distributed transaction.",
        nameof(container));
}
```

Use **instance** equality (`ReferenceEquals`) — not endpoint equality — because serializer, handlers, retry config, session state, and encryption are all client-instance-scoped.

Suggested test additions (per op family):
- same `CosmosClient` → succeeds
- different `CosmosClient` → throws `ArgumentException`
- `container.Database.Client == null` → throws `ArgumentException`

---

### 🟡 Suggestion 2 — Snapshot container properties once, then validate + use the locals

`Container` is abstract, so wrappers / decorators / mocks can return different values across successive property reads. Today the code reads `container.Database` in `ValidateContainerReference`, then reads `container.Database.Id` and `container.Id` again at the operation construction site. Cheap defensive change that also makes the same-client check above plug in naturally:

```csharp
private static (string databaseId, string containerId, CosmosClient owner) ValidateAndUnpack(Container container)
{
    if (container == null)
    {
        throw new ArgumentNullException(nameof(container));
    }

    string containerId = container.Id;
    Database database = container.Database;

    if (string.IsNullOrWhiteSpace(containerId))
    {
        throw new ArgumentException(
            "Container reference must have a non-empty Id.",
            nameof(container));
    }

    if (database == null)
    {
        throw new ArgumentException(
            "Container reference must expose a non-null Database.",
            nameof(container));
    }

    string databaseId = database.Id;

    if (string.IsNullOrWhiteSpace(databaseId))
    {
        throw new ArgumentException(
            "Container reference must have a non-empty Database.Id.",
            nameof(container));
    }

    return (databaseId, containerId, database.Client);
}
```

Then at each call site:

```csharp
(string databaseId, string containerId, CosmosClient owner) = ValidateAndUnpack(container);

if (!object.ReferenceEquals(owner, this.clientContext.Client))
{
    throw new ArgumentException(
        "Container must belong to the same CosmosClient instance that created this distributed transaction.",
        nameof(container));
}

this.operations.Add(
    new DistributedTransactionOperation<T>(
        operationType: OperationType.Create,
        operationIndex: this.operations.Count,
        databaseId,
        containerId,
        partitionKey,
        id,
        resource,
        requestOptions));
```

This avoids any chance of a TOCTOU-style inconsistency between validation and use, and centralizes the (currently duplicated) validation across both `*Core` classes.

---

Nice work overall — this is the right shape for the API, and these two tweaks lock in the safety story before the surface starts getting reviewed by more consumers. 👍

_Reviewed via Copilot CLI with prior context from design-doc review of #5877._

kushagraThapar (dismissed):

ananth7592:
Overall LGTM. One thing worth flagging on the API shape: we take a Container but immediately unpack it to (databaseId, containerId) and the rest of the DTX pipeline only sees those strings ( as before). The PR description calls this out, but it's worth being explicit about what we're giving up — anything attached to the Container instance (custom CosmosSerializer, container-scoped ClientEncryptionPolicy, decoratorwrappers for telemetry/retries, future per-container pipeline customizations) is silently discarded once we hit downstream DTX stuff DistributedTransactionOperation / the committer.etc.

The signature implies "this Container's behavior is used," but really we only use it for name extraction + the ReferenceEquals owner check — so the first consumer who wraps a Container in an encryptiondecorator will be very surprised when encryption is bypassed. 

Can we add an XML doc comment on the abstract methods (and onValidateAndUnpackContainer) saying DTC bypasses the per-Container request pipeline and only the name strings are honored

kushagraThapar (approved):

ananth7592 (approved):
Spoke offline for adding required xml tags in a follow up PR. 
LGTM, Approving it Thanks Meghana.

PR #5896, opened 2026-05-20
[Internal] DTS: Removes dead code and refactors to manual JSON deserialization

Meghana-Palaparthi:
# Pull Request Template

## Description

This pull request refactors the distributed transaction response and operation result handling in the Cosmos DB SDK. The main focus is on removing dependencies on `System.Text.Json.Serialization` attributes and reflection-based deserialization, instead implementing explicit JSON parsing logic. Additionally, error message and tracing fields are removed from the public API surface and internal data flow, simplifying the object model.

### Serialization and Deserialization Refactoring

* Removed all `System.Text.Json.Serialization` attributes (such as `[JsonInclude]`, `[JsonPropertyName]`, `[JsonIgnore]`, and `[JsonConstructor]`) from `DistributedTransactionOperationResult` and replaced reflection-based deserialization with a manual, case-insensitive property lookup and assignment in the new `FromJson` method. (`DistributedTransactionOperationResult.cs`)
* Added a new helper method `TryGetProperty` for case-insensitive property lookup when parsing JSON elements. (`DistributedTransactionOperationResult.cs`)
* Updated the deserialization logic in `DistributedTransactionResponse` to use the new explicit parsing methods instead of relying on `System.Text.Json` attributes. (`DistributedTransactionResponse.cs`) [[1]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L324-R314) [[2]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L341-L342)

### Error Message and Tracing Cleanup

* Removed the `ErrorMessage` property and associated logic from `DistributedTransactionResponse` and related constructors, as well as the corresponding test assertions. (`DistributedTransactionResponse.cs`, `DistributedTransactionResponseTests.cs`) [[1]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L35-L49) [[2]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L106-L110) [[3]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L203-L215) [[4]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L232-R224) [[5]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L378-L382)
* Removed the `Trace` and `ActivityId` fields from both `DistributedTransactionResponse` and `DistributedTransactionOperationResult`, and eliminated their propagation through constructors and methods. (`DistributedTransactionResponse.cs`, `DistributedTransactionOperationResult.cs`) [[1]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L35-L49) [[2]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L132-L133) [[3]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L341-L342) [[4]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L391-R370) [[5]](diffhunk://#diff-2115107c88c6a1f6a4b4bce64e055f821eb64b82ba0813f4e55ad60a015aae87L401-L402)

### Public API Simplification

* Changed several properties in `DistributedTransactionOperationResult` (such as `SessionToken`, `PartitionKeyRangeId`, and `SubStatusCode`) from public to internal, further restricting their visibility and clarifying the public API. (`DistributedTransactionOperationResult.cs`)

### Test Adjustments

* Updated tests to match the new deserialization logic and removed checks for properties that no longer exist (such as `SubStatusCodeValue`). (`DistributedTransactionResponseTests.cs`) [[1]](diffhunk://#diff-60339e9720594260887240840c9612ed875596345fa3fb21714827d34511a59aL495-R495) [[2]](diffhunk://#diff-60339e9720594260887240840c9612ed875596345fa3fb21714827d34511a59aL511-R512)

These changes collectively improve maintainability, reduce reliance on reflection and serialization attributes, and clarify the SDK's public and internal APIs.

## Type of change

## Closing issues

## Changelog

> Feature isn't public yet.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
@@ -75,44 +75,49 @@ namespace Microsoft.Azure.Cosmos
         /// <returns>The deserialized operation result with a canonical session token.</returns>
         internal static DistributedTransactionOperationResult FromJson(JsonElement json)
         {
+            if (json.ValueKind != JsonValueKind.Object)
+            {
+                throw new JsonException($"DTC operation result must be a JSON object, but was '{json.ValueKind}'.");
+            }
+
             DistributedTransactionOperationResult result = new DistributedTransactionOperationResult();
 
-            if (TryGetProperty(json, "index", out JsonElement indexEl) && indexEl.TryGetInt32(out int index))
+            if (TryGetInt32Property(json, DistributedTransactionSerializer.Index, out int index))
             {
                 result.Index = index;
             }
 
-            if (TryGetProperty(json, "statusCode", out JsonElement statusCodeEl) && statusCodeEl.TryGetInt32(out int statusCode))
+            if (TryGetInt32Property(json, DistributedTransactionSerializer.StatusCode, out int statusCode))
             {
                 result.StatusCode = (HttpStatusCode)statusCode;
             }
 
-            if (TryGetProperty(json, "subStatusCode", out JsonElement subStatusEl) && subStatusEl.TryGetUInt32(out uint subStatus))
+            if (TryGetUInt32Property(json, DistributedTransactionSerializer.SubStatusCode, out uint subStatus))
             {
                 result.SubStatusCode = (SubStatusCodes)subStatus;
             }
 
-            if (TryGetProperty(json, "etag", out JsonElement etagEl) && etagEl.ValueKind == JsonValueKind.String)
+            if (TryGetProperty(json, DistributedTransactionSerializer.ResponseETag, out JsonElement etagEl) && etagEl.ValueKind == JsonValueKind.String)
             {
                 result.ETag = etagEl.GetString();
             }
 
-            if (TryGetProperty(json, "sessionToken", out JsonElement sessionTokenEl) && sessionTokenEl.ValueKind == JsonValueKind.String)
+            if (TryGetProperty(json, DistributedTransactionSerializer.SessionToken, out JsonElement sessionTokenEl) && sessionTokenEl.ValueKind == JsonValueKind.String)
             {
                 result.SessionToken = sessionTokenEl.GetString();
             }
 
-            if (TryGetProperty(json, "partitionKeyRangeId", out JsonElement pkRangeIdEl) && pkRangeIdEl.ValueKind == JsonValueKind.String)
+            if (TryGetProperty(json, DistributedTransactionSerializer.PartitionKeyRangeId, out JsonElement pkRangeIdEl) && pkRangeIdEl.ValueKind == JsonValueKind.String)
             {
                 result.PartitionKeyRangeId = pkRangeIdEl.GetString();
             }
 
-            if (TryGetProperty(json, "requestCharge", out JsonElement requestChargeEl) && requestChargeEl.TryGetDouble(out double requestCharge))
+            if (TryGetDoubleProperty(json, DistributedTransactionSerializer.RequestCharge, out double requestCharge))
             {
                 result.RequestCharge = requestCharge;
             }
 
-            if (TryGetProperty(json, DistributedTransactionSerializer.ResourceBody, out JsonElement resourceBody)
+            if (TryGetPropertyOrdinal(json, DistributedTransactionSerializer.ResourceBody, out JsonElement resourceBody)
                 && resourceBody.ValueKind != JsonValueKind.Undefined
                 && resourceBody.ValueKind != JsonValueKind.Null)
             {
@@ -157,6 +162,12 @@ namespace Microsoft.Azure.Cosmos
 
         internal static bool TryGetProperty(JsonElement element, string propertyName, out JsonElement value)
         {
+            if (element.ValueKind != JsonValueKind.Object)
+            {
+                value = default;
+                return false;
+            }
+
             foreach (JsonProperty prop in element.EnumerateObject())
             {
                 if (string.Equals(prop.Name, propertyName, StringComparison.OrdinalIgnoreCase))
@@ -164,6 +175,80 @@ namespace Microsoft.Azure.Cosmos
                     value = prop.Value;
                     return true;
                 }
+            }
+
+            value = default;
+            return false;
+        }
+
+        internal static bool TryGetPropertyOrdinal(JsonElement element, string propertyName, out JsonElement value)
+        {
+            if (element.ValueKind != JsonValueKind.Object)
+            {
+                value = default;
+                return false;
+            }
+
+            return element.TryGetProperty(propertyName, out value);
+        }
+
+        private static bool TryGetInt32Property(JsonElement element, string propertyName, out int value)
+        {
+            if (TryGetProperty(element, propertyName, out JsonElement propertyElement))
+            {
+                if (propertyElement.ValueKind != JsonValueKind.Number)
+                {
+                    throw new JsonException($"'{propertyName}' must be a JSON number, but was '{propertyElement.ValueKind}'.");
+                }
+
+                if (!propertyElement.TryGetInt32(out value))
+                {
+                    throw new JsonException($"'{propertyName}' must be a 32-bit integer JSON number.");
+                }
+
+                return true;
+            }
+
+            value = default;
+            return false;
+        }
+
+        private static bool TryGetUInt32Property(JsonElement element, string propertyName, out uint value)
+        {
+            if (TryGetProperty(element, propertyName, out JsonElement propertyElement))
+            {
+                if (propertyElement.ValueKind != JsonValueKind.Number)
+                {
+                    throw new JsonException($"'{propertyName}' must be a JSON number, but was '{propertyElement.ValueKind}'.");
+                }
+
+                if (!propertyElement.TryGetUInt32(out value))
+                {
+                    throw new JsonException($"'{propertyName}' must be a 32-bit unsigned integer JSON number.");
+                }
+
+                return true;
+            }
+
+            value = default;
+            return false;
+        }
+
+        private static bool TryGetDoubleProperty(JsonElement element, string propertyName, out double value)
+        {
+            if (TryGetProperty(element, propertyName, out JsonElement propertyElement))
+            {
+                if (propertyElement.ValueKind != JsonValueKind.Number)
+                {
+                    throw new JsonException($"'{propertyName}' must be a JSON number, but was '{propertyElement.ValueKind}'.");
+                }
+
+                if (!propertyElement.TryGetDouble(out value))
+                {
+                    throw new JsonException($"'{propertyName}' must be a double-precision JSON number.");
+                }
+
+                return true;
             }
 
             value = default;
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
@@ -8,7 +8,6 @@ namespace Microsoft.Azure.Cosmos
     using System.IO;
     using System.Net;
     using System.Text.Json;
-    using System.Text.Json.Serialization;
     using Microsoft.Azure.Cosmos.Core.Trace;
     using Microsoft.Azure.Cosmos.Tracing;
     using Microsoft.Azure.Documents;
@@ -46,14 +45,7 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// Initializes a new instance of the <see cref="DistributedTransactionOperationResult"/> class.
         /// </summary>
-        /// <remarks>
-        /// Must be <c>public</c> for System.Text.Json reflection-based deserialization.
-        /// System.Text.Json 6.x only scans <c>BindingFlags.Public</c> constructors when resolving
-        /// <see cref="JsonConstructorAttribute"/>; non-public constructors are not found.
-        /// Support for non-public constructors was added in System.Text.Json 7.0.
-        /// </remarks>
-        [JsonConstructor]
-        public DistributedTransactionOperationResult()
+        internal DistributedTransactionOperationResult()
         {
         }
 
@@ -88,7 +80,6 @@ namespace Microsoft.Azure.Cosmos
         /// Do not dispose it directly. To deserialize to a typed object, use
         /// <see cref="DistributedTransactionResponse.GetOperationResultAtIndex{T}"/>.
         /// </remarks>
-        [JsonIgnore]
         public virtual Stream ResourceStream { get; internal set; }
 
         /// <summary>
@@ -105,13 +96,10 @@ namespace Microsoft.Azure.Cosmos
         /// <summary>
         /// ActivityId related to the operation.
         /// </summary>
-        [JsonIgnore]
         internal virtual string ActivityId { get; set; }
 
-        [JsonIgnore]
         internal ITrace Trace { get; set; }
 
-        [JsonIgnore]
         internal CosmosSerializerCore SerializerCore { get; set; }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -192,5 +154,20 @@ internal static DistributedTransactionOperationResult FromJson(JsonElement json)
 
             return result;
         }
+
+        internal static bool TryGetProperty(JsonElement element, string propertyName, out JsonElement value)
+        {
+            foreach (JsonProperty prop in element.EnumerateObject())
```
kushagraThapar, line 245: Could you please help me understand the failure mode for non-object JSON elements here? `JsonElement.EnumerateObject()` throws `InvalidOperationException` when `ValueKind != Object`, but the only `catch` in the call chain (`DistributedTransactionResponse.cs:327`) catches `JsonException`. So a payload like `{"operationResponses":[null]}` or `{"operationResponses":[42]}` looks like it would bubble `InvalidOperationException` out of `FromJson` → past the recovery branch → out of `FromResponseMessageAsync` to the caller, whereas the old `JsonSerializer.Deserialize<T>(...)` path would have fallen back to `InternalServerError`.

Would you mind adding a guard at the top of `TryGetProperty` plus a regression test in a follow-up PR? Something like:

```csharp
if (element.ValueKind != JsonValueKind.Object)
{
    value = default;
    return false;
}
```

Happy to be wrong about the contract — if per-op elements are guaranteed to be objects upstream, a comment to that effect would also work.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -147,10 +75,44 @@ public virtual uint SubStatusCodeValue
...
-                ?? throw new JsonException($"Failed to deserialize DTC operation result: Deserialize returned null. JSON element kind: '{json.ValueKind}'.");
+            DistributedTransactionOperationResult result = new DistributedTransactionOperationResult();
+
+            if (TryGetProperty(json, "index", out JsonElement indexEl) && indexEl.TryGetInt32(out int index))
+            {
+                result.Index = index;
+            }
 
-            if (json.TryGetProperty(DistributedTransactionSerializer.ResourceBody, out JsonElement resourceBody)
+            if (TryGetProperty(json, "statusCode", out JsonElement statusCodeEl) && statusCodeEl.TryGetInt32(out int statusCode))
```
kushagraThapar: I am curious about the behavior change for the numeric fields here (`statusCode`, `subStatusCode` at line 90, `requestCharge` at line 110). Before this PR, a malformed payload like `{"statusCode":"abc"}` would throw `JsonException` and get caught at `DistributedTransactionResponse.cs:327`, which clears `results` and lets the count-mismatch branch promote to `InternalServerError`. After this PR, `TryGetInt32` simply returns `false` and `StatusCode` silently stays at `0` — the operation gets reported as "failed with status 0" with no diagnostic signal. Same shape for `requestCharge` (silently `0` → RU tracking off) and `subStatusCode` (silently `Unknown`).

I notice the string-typed fields (`etag` / `sessionToken` / `partitionKeyRangeId`) intentionally skip-on-wrong-type per the PR description — that reads as a sensible null-safety guard. Was the same silent-default intended for the numeric fields, or would you consider throwing `JsonException` on present-but-wrong-type so the existing recovery path still kicks in? For example:

```csharp
if (TryGetProperty(json, "statusCode", out JsonElement statusCodeEl))
{
    if (!statusCodeEl.TryGetInt32(out int statusCode))
    {
        throw new JsonException($"'statusCode' must be a JSON number, but was '{statusCodeEl.ValueKind}'.");
    }
    result.StatusCode = (HttpStatusCode)statusCode;
}
```

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -321,14 +304,14 @@ private static async Task<DistributedTransactionResponse> PopulateFromJsonConten
             {
                 JsonElement root = responseJson.RootElement;
 
-                if (root.TryGetProperty("isRetriable", out JsonElement isRetriableElement) &&
+                if (DistributedTransactionOperationResult.TryGetProperty(root, "isRetriable", out JsonElement isRetriableElement) &&
```
kushagraThapar: Could you please help me understand whether the case-sensitivity change at these top-level lookups is intentional? Before this PR, `root.TryGetProperty("isRetriable", ...)` here and the matching `operationResponses` lookup at line 314 (plus `resourceBody` at `DistributedTransactionOperationResult.cs:115`) went through the BCL's case-**sensitive** `TryGetProperty`. The new `DistributedTransactionOperationResult.TryGetProperty` is `OrdinalIgnoreCase`, so the SDK now silently accepts `"IsRetriable"` / `"OperationResponses"` / `"ResourceBody"`.

If the wire contract is strict camelCase from the server, this is an over-permissive loosening worth tightening (one option: a separate `TryGetPropertyOrdinal` helper for the top-level lookups). If it's intentional defensive parsing, a one-line comment explaining the choice would help future maintainers — and ideally a `[DataRow]` covering the PascalCase variant. Either way is fine with me, I'd just like to make sure it's a conscious decision rather than an accidental side-effect of the helper consolidation.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedTransactionResponseTests.cs
```diff
@@ -492,7 +492,7 @@ public async Task Count_ReturnsNumberOfParsedResults()
         // Operation result property deserialization
 
         [TestMethod]
-        [Description("SubStatusCodeValue deserializes from the 'subStatusCode' JSON property and is accessible as both uint and SubStatusCodes enum.")]
+        [Description("SubStatusCode deserializes from the 'subStatusCode' JSON property.")]
```
kushagraThapar: Quick observation on test coverage in this file: I went through the 28 `[TestMethod]`s and they all feed exact camelCase keys with well-typed numeric values. That means the behavior changes this PR introduces aren't actually exercised in tests, specifically:

1. Case-insensitive top-level lookups (the inline on `DistributedTransactionResponse.cs:307` above) — easy `[DataRow]` with PascalCase keys.
2. Numeric type mismatch on `statusCode` / `requestCharge` / `subStatusCode` (the inline on `DistributedTransactionOperationResult.cs:85` above) — assertion shape depends on which way you go there.
3. Non-object element inside `operationResponses` (the inline at line 160 of the helper) — a payload like `{"operationResponses":[null]}` should yield `InternalServerError`, not throw.

Could you please add one test per axis in the follow-up PR? They're cheap to write and would pin down the contract before DTS goes public to customers.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -147,10 +75,44 @@ public virtual uint SubStatusCodeValue
...
+            {
+                result.SubStatusCode = (SubStatusCodes)subStatus;
+            }
+
+            if (TryGetProperty(json, "etag", out JsonElement etagEl) && etagEl.ValueKind == JsonValueKind.String)
+            {
+                result.ETag = etagEl.GetString();
+            }
+
+            if (TryGetProperty(json, "sessionToken", out JsonElement sessionTokenEl) && sessionTokenEl.ValueKind == JsonValueKind.String)
```
kushagraThapar: Small consistency observation: PR #5818 introduced `DistributedTransactionSerializer.SessionToken` / `PartitionKeyRangeId` / `Index` / `ResourceBody` constants precisely so the deserializer doesn't drift from the serializer on a rename. Here the new code uses the constant for `ResourceBody` at line 115 but hard-codes the literal `"sessionToken"` here, and similarly for `"index"` (line 80), `"statusCode"` (85), `"subStatusCode"` (90), `"etag"` (95), `"partitionKeyRangeId"` (105), `"requestCharge"` (110).

Could you please route these lookups through the existing constants and add the missing ones (`StatusCode`, `SubStatusCode`, `RequestCharge`, `IsRetriable`, `OperationResponses`, plus a separate response-side `Etag` since `DistributedTransactionSerializer.ETag = "ifMatch"` is a different field) in the follow-up? Keeps the principle from #5818 holding end-to-end.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -32,21 +32,17 @@ class DistributedTransactionResponse : IReadOnlyList<DistributedTransactionOpera
         private DistributedTransactionResponse(
```
kushagraThapar, line 32: I am curious about the rationale for dropping `errorMessage` from this constructor (and the corresponding public `ErrorMessage` property). Looking at sibling shapes across the SDKs:

- .NET `TransactionalBatchResponse.cs:113` — `public virtual string ErrorMessage { get; internal set; }`
- Java `azure-cosmos/src/main/java/com/azure/cosmos/models/CosmosBatchResponse.java` — `public String getErrorMessage()`
- Python `azure-cosmos/azure/cosmos/exceptions.py` — `self.http_error_message = message` on `CosmosBatchOperationError`

I also notice the synthetic `InternalServerError` branch at lines 213-220 (previously `ClientResources.InvalidServerResponse`) and the JSON parse fallbacks now have no error-text on the response itself — operators only get a `DefaultTrace.TraceWarning`, which is in-process and not surfaced to callers.

If this is deliberate (e.g. `responseMessage.ErrorMessage` already covers the customer-visible surface) no objection — but it is a divergence from three peer surface contracts at the moment DTS goes public, so could you please confirm the intent with Fabian (#5856 reviewer) and either restore the property in a follow-up or capture the conscious omission in a DTS readiness doc? On the design-docs side, a small extension in `tvaron3/cosmosdb-design-docs` would be a great place to anchor the error-surface decision — would you mind filing that as well?

kushagraThapar (dismissed):
Thanks for the cleanup, @Meghana-Palaparthi — the manual JSON deserializer is the right shape and aligns nicely with what `TransactionalBatchOperationResult`, Java's `BatchResponseParser`, and Python's batch-response paths already do (no STJ attributes, direct property walks). The public→internal demotions for `SessionToken` / `PartitionKeyRangeId` / `SubStatusCode` / `ActivityId` / `Trace` also bring DTS in line with `TransactionalBatchOperationResult.cs:94,99,104`, which I think is the right consolidation.

Approving since this is `[Internal]` and there are no shipped customers yet. I've left a few inline observations on correctness fidelity vs the old `JsonSerializer.Deserialize<T>` path that would be worth picking up in a follow-up PR — especially the `InvalidOperationException` leak from `EnumerateObject()` and the silent-default behavior on numeric type mismatches. These are subtle behavior changes customers won't see today, but the DTS team is likely to hit them during internal testing.

None of the inline items block this PR — they're all opportunities for a small fast-follow.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
Meghana-Palaparthi: Thanks, this is valid. I updated the local parser to guard non-object operation entries before property enumeration. FromJson now throws JsonException for non-object elements (for example null/number), which routes through the existing recovery path and results in the expected InternalServerError behavior instead of leaking InvalidOperationException.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
Meghana-Palaparthi: Thanks for calling out. I dropped it earlier because we weren't receiving it in response from DTC. I restored it now and using it for SDK error messages - to provide details of failures

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
Meghana-Palaparthi: Great point, agreed. I changed numeric parsing (index, statusCode, subStatusCode, requestCharge) so present-but-invalid values throw JsonException rather than silently defaulting to 0/Unknown. That preserves the original failure semantics and keeps malformed payloads from looking like valid operation results.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
Meghana-Palaparthi: Thanks for calling this out. I confirmed this distinction and aligned behavior with main: top-level keys (isRetriable, operationResponses) and resourceBody are now case-sensitive again. The previous case-insensitive behavior on main came from JsonSerializer for operation-result properties only, not these top-level lookups.

kushagraThapar (dismissed):

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -314,21 +306,27 @@ private static async Task<DistributedTransactionResponse> PopulateFromJsonConten
...
+
                 return null;
             }
 
             using (responseJson)
             {
                 JsonElement root = responseJson.RootElement;
 
-                if (root.TryGetProperty("isRetriable", out JsonElement isRetriableElement) &&
+                if (DistributedTransactionOperationResult.TryGetPropertyOrdinal(root, DistributedTransactionSerializer.IsRetriable, out JsonElement isRetriableElement) &&
```
jeet1995: **Question:** This uses `TryGetPropertyOrdinal` (case-sensitive) for top-level keys (`isRetriable`, `operationResponses`), while per-operation fields (`index`, `statusCode`, `Etag`, etc.) use the custom `TryGetProperty` with `OrdinalIgnoreCase`. `resourceBody` also uses case-sensitive matching.

Is the mixed case sensitivity intentional? What's the rationale for case-insensitive matching on per-operation fields but case-sensitive on the envelope and `resourceBody`?

The old code had the same split (top-level via `JsonElement.TryGetProperty`, per-op via `PropertyNameCaseInsensitive = true`), so this PR preserves the existing behavior — but was that a deliberate design choice or just inherited?

Meghana-Palaparthi: There isn't a strong rationale for this. Originally, everything was case-sensitive. We made per-op fields case-insensitive because they changed frequently in the coordinator. While there's still a chance that casing could change at the per-op level, the risk is lower for top-level keys. We could make everything case-insensitive to minimize the risk of breaking the flow if there are any silent changes in the coordinator.

Meghana-Palaparthi: To be cohesive across, made all the properties case-insensitive and updated the tests accordingly

kushagraThapar (approved):

jeet1995 (approved):

PR #5902, opened 2026-05-24
[Client encryption]: Adds Stream fast-path bypass for legacy Newtonsoft peek in DecryptAsync

adamnova:
## Summary
When a caller explicitly opts into `JsonProcessor.Stream` via the property-bag override (`encryption-json-processor=Stream` on `RequestOptions.Properties`), the `EncryptionProcessor.DecryptAsync(input, encryptor, diagCtx, requestOptions, ct)` overload was still parsing the response payload through Newtonsoft's `JObject` before delegating to `MdeEncryptionProcessor`. That `JObject` parse exists only to detect legacy `AEAes256CbcHmacSha256Randomized` documents and route them to the Newtonsoft fallback. For MDE documents (which Stream callers exclusively use) the parse is wasted work and dominates per-request allocations.

This adds a fast path that skips directly to `MdeEncryptionProcessor.DecryptAsync`. If the document turns out to be legacy AE-AES, `MdeEncryptionProcessor` throws `NotSupportedException`, which is caught and the call falls through to the existing JObject peek + legacy decrypt path. The documented "Stream is a hint; legacy documents transparently fall back to Newtonsoft" contract is preserved — `EncryptionProcessorTests.Decrypt_StreamSelection_LegacyAlgorithm_FallsBackToNewtonsoft` continues to pass.

## Why
Managed allocation profiling (`dotnet-trace` + TraceEvent) showed the Newtonsoft `JObject` peek accounting for 38% of allocations on 1 KB documents and 52% on 100 KB documents. That is the largest single attributable cost in the Stream/MDE path, so removing it for MDE callers is the highest-leverage single optimization.

## Measurement
Processor-level Read benchmark (no Cosmos network; through `EncryptionContainer` wrapping an in-memory fake `Container`; BenchmarkDotNet MediumRun, `InProcessNoEmitToolchain`, `MemoryDiagnoser`):

| Doc | Before allocated | After allocated | Δ KB | Δ % | Before mean | After mean | Latency Δ |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 KB | 41.87 KB/op | 25.07 KB/op | -16.80 | **-40 %** | 67 µs | 40 µs | **-40 %** |
| 10 KB | 92.88 KB/op | 52.50 KB/op | -40.38 | **-44 %** | 121 µs | 88 µs | **-27 %** |
| 100 KB | 533.87 KB/op | 252.99 KB/op | -280.88 | **-53 %** | 961 µs | 594 µs | **-38 %** |

The Newtonsoft column was unchanged in the same matrix (control, fast path does not fire).

## Change scope
- 1 file, +14 lines
- `Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs` only
- `#if NET8_0_OR_GREATER` gated (matches `JsonProcessor.Stream` availability)
- No public API change
- No new dependency

## Test coverage
- All 420 `Microsoft.Azure.Cosmos.Encryption.Custom.Tests` (`net8.0`) pass.
- Key regression test for the legacy fallback contract:
  - `Decrypt_StreamSelection_LegacyAlgorithm_FallsBackToNewtonsoft` — passes after the change (NotSupportedException catch routes back through the legacy path).
- Cross-compatibility tested by encrypting with Newtonsoft and decrypting with Stream (and vice versa) in a parameterized component-test suite.

### DIFF
--- a/Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
+++ b/Microsoft.Azure.Cosmos.Encryption.Custom/src/EncryptionProcessor.cs
@@ -182,6 +182,30 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom
             Debug.Assert(encryptor != null);
             Debug.Assert(diagnosticsContext != null);
 
+            // Fast path: when the caller explicitly opts into JsonProcessor.Stream via the property-bag
+            // override, skip the Newtonsoft JObject peek that exists only to detect legacy AE-AES documents.
+            // The Stream/MDE path will throw NotSupportedException if the document turns out to be legacy
+            // AE-AES; fall back to the JObject peek path in that case to preserve the documented
+            // "Stream is a hint; legacy documents transparently fall back to Newtonsoft" contract
+            // (covered by EncryptionProcessorTests.Decrypt_StreamSelection_LegacyAlgorithm_FallsBackToNewtonsoft).
+#if NET8_0_OR_GREATER
+            if (requestOptions != null
+                && requestOptions.TryReadJsonProcessorOverride(out JsonProcessor overrideProcessor)
+                && overrideProcessor == JsonProcessor.Stream)
+            {
+                try
+                {
+                    return await MdeEncryptionProcessor.DecryptAsync(input, encryptor, diagnosticsContext, requestOptions, cancellationToken);
+                }
+                catch (NotSupportedException)
+                {
+                    // Document is legacy AE-AES (or otherwise unsupported by the Stream processor).
+                    // Rewind and fall through to the JObject peek path below.
+                    input.Position = 0;
+                }
+            }
+#endif
+
             // Try to peek at the content to check if it's legacy encryption algorithm
             // Some streams (e.g., those that only support async reads or contain malformed JSON) may throw exceptions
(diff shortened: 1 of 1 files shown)


adamnova:
Superseded by #5903.

#5903 replaces the `try` / `catch(NotSupportedException)` Stream-opt-in fast path introduced here with a purpose-built `LegacyAlgorithmDetector` (`Utf8JsonReader` peek of `_ei._ea`). Same MDE-double-parse fix as this PR, plus:

- No wasted `MdeEncryptionProcessor.DecryptAsync` deserialize on legacy documents (this PR did a full STJ parse + base64-decode of `_ed` before catching `NotSupportedException`).
- No exception-driven control flow.
- Non-opt-in path is bit-for-bit unchanged.
- 37 parity tests in `EncryptionProcessorStreamParityTests` assert opt-in and non-opt-in produce identical bytes and `DecryptionContext` on every shape (legacy, MDE, unencrypted, malformed, missing properties, case variants, async-only streams).

Benchmark vs this PR (legacy 1KB `−58%` time / `−13%` alloc; MDE neutral by design — both already avoid the master double-parse).

Closing in favour of #5903.

adamnova:
Closing — superseded by #5903 (see comment above for details).

PR #5903, opened 2026-05-25
Encryption.Custom: Refactors Stream-opt-in legacy detection to Utf8JsonReader peek

adamnova:
## Description

Follow-up to #5902. Replaces the exception-driven legacy AE-AES detection (try/catch over `NotSupportedException` from `SystemTextJsonStreamAdapter`) with a purpose-built `LegacyAlgorithmDetector` (`Utf8JsonReader` scan) for `JsonProcessor.Stream` opt-in callers.

The original detector triggered a full `JsonSerializer.DeserializeAsync<EncryptionPropertiesWrapper>` (base64-decoding the entire `_ed` payload) before throwing, so legacy documents paid for two full parses. The new detector classifies the document in a single pass over the top-level `_ei._ea` property without allocating either a `JObject` or an `EncryptionPropertiesWrapper`.

## Strict behavioural parity

Stream-opt-in callers see byte-for-byte the same output as the non-opt-in (Newtonsoft) path on every input that does not match the MDE fast-path classification. Two correctness fixes landed alongside the detector to make this guarantee hold:

1. `LegacyAlgorithmDetector` exact-matches both the legacy AND the MDE algorithm names. Empty strings, unknown algorithms, case variants, prefixes, and suffixes all classify as `Unknown` (routed through the JObject peek path) instead of being optimistically forwarded to the streaming MDE adapter.
2. `DecryptViaJObjectPeekAsync` gained `stripStreamOverrideOnSuccessFallthrough`: when the synchronous JObject parse succeeded, the post-peek MDE fallthrough strips the `JsonProcessor.Stream` override (parity with non-opt-in). When the parse threw (async-only stream, malformed JSON), the override is preserved so the streaming MDE adapter still gets a shot - preserves the pre-existing `Decrypt_Cancellation_Aborts` contract on `SlowCancelableStream`.

## Coverage

- 44 direct `LegacyAlgorithmDetector` unit tests (empty/whitespace/non-object inputs, `_ei` shape variants, `_ea` position variants, JSON escapes, case-sensitivity, prefix/suffix collisions, MDE exact-match enforcement).
- 10 Stream-opt-in legacy round-trip integration tests in `EncryptionProcessorTests` (full property parity, `DecryptionContext` path lists, large docs, repeated round-trips, null sensitive values, non-MemoryStream wrappers, byte-for-byte parity with Newtonsoft, unknown-algorithm propagation).
- **37 strict-parity tests in `EncryptionProcessorStreamParityTests`** that run every input through BOTH paths and assert agreement on: thrown vs succeeded, exception family, output bytes, `PathsDecrypted`. Sections: round-trip, legacy, edge-case payloads (empty/null/non-string `_ea`, `_ei` shape variants, truncated/empty/whitespace JSON, deep nesting), input-stream variants, cancellation, repeated calls.
- **All 285 tests pass on net8.0** (up from 194 on master). Zero warnings under `TreatWarningsAsErrors=true`.

## Benchmarks

New `LegacyDetectionBenchmark` with `[MemoryDiagnoser]` measures the detector against an inlined replica of the PR #5902 try/catch baseline (MediumRun, InProcessEmit, 15 iters x 2 launches):

| Scenario | Doc | New (us / KB alloc) | Baseline (us / KB alloc) | Time | Alloc |
|---|---|---|---|---|---|
| **Legacy** | 1 KB   |  62 /  45.3 | 146 /  52.3 | **-58%** | **-13%** |
| Legacy     | 10 KB  | 316 / 215.4 | 401 / 232.1 | -21% | -7% |
| Legacy     | 100 KB | 2847 / 1859 | 3266 / 1974 | -13% | -6% |
| **MDE**    | 1 KB   |  25 /  12.3 |  24 /  12.3 | neutral | +-0 |
| MDE        | 10 KB  |  60 /  28.7 |  58 /  28.7 | neutral | +-0 |
| MDE        | 100 KB | 677 / 225.3 | 680 / 225.2 | neutral | +-0 |

Strictly equal-or-better on allocations in every configuration; removes the exception-throw + wasted MDE deserialize cost from the legacy path entirely.

Also fixed the perf project so BenchmarkDotNet can load the existing `EncryptionBenchmark` (BDN refuses `[Params]` on non-public members like the internal `JsonProcessor` enum). Added a public `JsonProcessorOption` surrogate enum + `BenchmarkSwitcher` in `Program.cs`.

## Gating

The detector and routing changes are gated behind `NET8_0_OR_GREATER` (the framework the Stream opt-in path requires).

## Type of change

- [x] Refactor (no functional changes, no api changes)


adamnova:
Closing - superseded. The real parity fixes were absorbed into #5985: the verbatim-escape fix (same defect as #5959), numeric-string _ef handling, and stream-input disposal. The Utf8JsonReader-peek legacy-detection optimization was deferred (niche perf on a cold path; conflicts with the container-default threading in #5985) and can be revisited as a focused follow-up; the large audit-probe/[Ignore] test files were not taken.

PR #5904, opened 2026-05-26
[Internal] DTS: Fixes per-operation status code loss on error responses

Meghana-Palaparthi:
# Pull Request Template

## Description

When the service returns a non-success status (e.g., 452) for a distributed transaction batch with a JSON body containing per-operation results, the response body was being discarded:

  - `GatewayStoreClient.ParseResponseAsync` threw `DocumentClientException` because 452 is not covered by `IsValidStatusCodeForExceptionlessRetry`, which disposed the response body.
  - `Extensions.ToCosmosResponseMessage` (`DocumentServiceResponse` overload) routed non-success responses through `CosmosExceptionFactory.Create`, whose `GetErrorFromStream` also disposes the body.

As a result, `DistributedTransactionResponse.FromResponseMessageAsync` saw null Content, skipped JSON parsing, and stamped every operation with the envelope status code (e.g., 452), losing the real per-op statuses (e.g., 412 PreconditionFailed).

Fix:
- `GatewayStoreClient.ParseResponseAsync`: for `DistributedTransactionBatch` requests, return a `DocumentServiceResponse `with the buffered body preserved when the body is non-empty. Falls back to the existing throw path when there is no body so detailed error context is retained.
- `Extensions.ToCosmosResponseMessage`: for `DistributedTransactionBatch` responses, bypass the `CosmosExceptionFactory` path and construct a `ResponseMessage` that keeps Content set. A lightweight `CosmosException` carrying status code, sub-status code, activity id, and request charge is still attached so consumers retain error context. Trace is propagated from `requestMessage.Trace` (with a `NoOpTrace` fallback to avoid NRE when no trace is set).

Tests:
- 6 new unit tests in `GatewayStoreClientTests` covering body preservation for 452/409/500/449, the null-body fallback, the Extensions DTX path (including `CosmosException` population), and a non-DTX regression check.

## Type of change

## Closing issues

## Changelog

> Feature not public yet

### DIFF
--- a/Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
+++ b/Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
@@ -137,7 +137,6 @@ namespace Microsoft.Azure.Cosmos
                 {
                     // For DTC responses, preserve the response body (operationResponses JSON) so the caller can parse per-op statuses.
                     // If there is no response body, fall back to throwing to preserve detailed error context.
-                    if (responseMessage.Content == null
                     INameValueCollection headers = GatewayStoreClient.ExtractResponseHeaders(responseMessage);
                     Stream contentStream = await GatewayStoreClient.BufferContentIfAvailableAsync(responseMessage);
 
@@ -145,12 +144,7 @@ namespace Microsoft.Azure.Cosmos
                     {
                         throw await GatewayStoreClient.CreateDocumentClientExceptionAsync(responseMessage, requestStatistics);
                     }
-                    {
-                        throw await GatewayStoreClient.CreateDocumentClientExceptionAsync(responseMessage, requestStatistics);
-                    }
 
-                    INameValueCollection headers = GatewayStoreClient.ExtractResponseHeaders(responseMessage);
-                    Stream contentStream = await GatewayStoreClient.BufferContentIfAvailableAsync(responseMessage);
                     return new DocumentServiceResponse(
                         body: contentStream,
                         headers: headers,
--- a/Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
+++ b/Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
@@ -132,6 +132,26 @@ namespace Microsoft.Azure.Cosmos
                         clientSideRequestStatistics: requestStatistics,
                         serializerSettings: serializerSettings);
                 }
+                else if (request != null
+                    && request.ResourceType == ResourceType.DistributedTransactionBatch)
+                {
+                    // For DTC responses, preserve the response body (operationResponses JSON) so the caller can parse per-op statuses.
+                    // If there is no response body, fall back to throwing to preserve detailed error context.
+                    if (responseMessage.Content == null
+                        || responseMessage.Content.Headers.ContentLength.GetValueOrDefault() == 0)
+                    {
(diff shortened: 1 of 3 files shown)


Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -132,6 +132,26 @@ internal static async Task<DocumentServiceResponse> ParseResponseAsync(HttpRespo
                         clientSideRequestStatistics: requestStatistics,
                         serializerSettings: serializerSettings);
                 }
+                else if (request != null
+                    && request.ResourceType == ResourceType.DistributedTransactionBatch)
```
jeet1995, line 136: I am curious if we can expand `IsValidStatusCodeForExceptionlessRetry` to include 449s, 452s or 500s (reuse code path from L126-L134)?

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
```diff
@@ -132,6 +132,26 @@ internal static async Task<DocumentServiceResponse> ParseResponseAsync(HttpRespo
...
+                {
+                    // For DTC responses, preserve the response body (operationResponses JSON) so the caller can parse per-op statuses.
+                    // If there is no response body, fall back to throwing to preserve detailed error context.
+                    if (responseMessage.Content == null
+                        || responseMessage.Content.Headers.ContentLength.GetValueOrDefault() == 0)
```
jeet1995: **Suggestion:** The `ContentLength.GetValueOrDefault()` check can false-negative when `Content-Length` header is absent (e.g., chunked transfer encoding) — `GetValueOrDefault()` returns 0 for null, causing a fallback to throw even when body content exists.

Consider pushing this check into the already-buffered stream instead:

```suggestion
                    INameValueCollection headers = GatewayStoreClient.ExtractResponseHeaders(responseMessage);
                    Stream contentStream = await GatewayStoreClient.BufferContentIfAvailableAsync(responseMessage);

                    if (contentStream == null || contentStream.Length == 0)
                    {
                        throw await GatewayStoreClient.CreateDocumentClientExceptionAsync(responseMessage, requestStatistics);
                    }
```

`BufferContentIfAvailableAsync` already handles null content (returns null), and checking `contentStream.Length` validates **actual buffered bytes** rather than the header value. This is also consistent with the success and exceptionless paths above, which don't check `Content-Length`.

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
Meghana-Palaparthi: Yeah, we can add 452's to `IsValidStatusCodeForExceptionlessRetry`. Currently, that status code is being used only by DTx, so it's safe to add it there. I need to explore whether I'll be breaking any existing flows by adding 449s and 500s. 

I'll take that as a follow-up item as that requires changes in direct package and a direct package release. Currently we need this change to get unblocked for E2E testing

jeet1995 (approved):

Microsoft.Azure.Cosmos/src/GatewayStoreClient.cs
jeet1995: @Meghana-Palaparthi - could you create a tracking item if needed?

kushagraThapar (approved):

PR #4249, opened 2024-01-11
Documentation: Fixes API comment for `RequestOptions.IfMatchEtag`

kundadebdatta:
# Pull Request Template

## Description

This PR fixes the API documentation for `RequestOptions.IfMatchEtag`.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Closing issues

To automatically close an issue: closes #4250, closes #3976

PR #5905, opened 2026-05-27
ReadMany: Adds point-read fast path for single-tuple partitions

### LINKED ISSUE #4369: ReadMany: Dynamically choose between point read and query

**Describe the bug**
Currently when there is only one PK+id tuple for a physical partition in readMany always a query is issued. The query cover-chargeof 2.82 RU makes this very expensive - instead a point read operation should be used hwen there is a single tuple for a Partition.

**To Reproduce**
Create a container with 2 thousand partitions, issue readMany calls with 100 random tuples of pk+id - you will see that readMany will use much more RU (at least 2x) compared to sequentilly doing point reads (or using custom logic to concurrently issue point reads)

**Expected behavior**
The SDK shoudl dynamically choose between point read (single tuple of smallish docs (<= 32 KB)) and query (large docs >32 KB or more than 1 tuple)

**Actual behavior**
Query is always used - resulting in unnecessary query cover-charge of 2..82 

**Additional context**
Java SDK has the optimization to choose point read for single tuple of PK+id already.

ananth7592:
## Summary

Adds the per-physical-partition point-read fast path to `ReadManyItems[Stream]Async`. When a request resolves to a physical partition that contains exactly one `(id, partitionKey)` tuple, the SDK now issues a point read for that tuple instead of a parameterized `SELECT * FROM c WHERE c.id IN (...)` (or OR-of-AND for non-`/id` PK) query. The decision is made independently per `PartitionKeyRange`, so a single `ReadMany` call may dispatch a mix of point reads and queries.

Closes #4369.

## Why

Point reads on small (<32 KB) documents are ~1 RU; the query path pays a fixed cover charge (~2.82 RU minimum) on top of the per-item cost. For workloads that read a few items spread across many partitions (a common cache-fill / fan-out shape), this is a >2x RU and latency win.

## Cross-SDK references

| SDK | File | Method | Status |
| --- | ---- | ------ | ------ |
| Java | `RxDocumentClientImpl.java` | `pointReadsForReadMany` / `queryForReadMany` / `getRangeQueryMap` | Already shipped via [Azure/azure-sdk-for-java#31723](https://github.com/Azure/azure-sdk-for-java/pull/31723); 404 swallowing tuned by PRs #34966 and #35513. |
| Python | `_read_items_helper.py` | `_execute_query_chunk_worker` (`if len(items_for_query) == 1:`) | Shipped. |
| .NET | `ReadManyQueryHelper.cs` | `ReadManyTaskHelperAsync` | **Previously missing** — addressed in this PR. |

## What changed

- `Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs`
  - `ReadManyTaskHelperAsync`: branch on `entry.Value.Count == 1` to dispatch a point read instead of building a `QueryDefinition`.
  - New `GeneratePointReadResponseForPartitionAsync`: issues `ContainerCore.ReadItemStreamAsync`, swallows 404/0 as "item not present" (preserving the RU charge in the aggregate), short-circuits other failures via `QueryResponse.CreateFailure`, and wraps successes as a synthetic single-element `QueryResponse` so the existing `CombineStreamsFromQueryResponses` / `CombineFeedResponseFromQueryResponses<T>` aggregation works unchanged.
  - Tracing: new child trace scope `"Point read for a partitionkeyrange"` (Query / Info), paralleling the existing `"Execute query for a partitionkeyrange"`.
- `Microsoft.Azure.Cosmos/src/ReadManyRequestOptions.cs`
  - New `internal ItemRequestOptions ConvertToItemRequestOptions()` mirroring the existing `ConvertToQueryRequestOptions()`. Maps `ConsistencyLevel`, `ReadConsistencyStrategy`, `SessionToken`, `IfMatchEtag`, `IfNoneMatchEtag`, `Properties`, `AddRequestHeaders`, `ExcludeRegions`.

## Tests added

In `Microsoft.Azure.Cosmos.EmulatorTests/CosmosReadManyItemsTests.cs`:

- **`ReadMany_AllSingleTuplePartitions_UsesPointReadsOnWire`** — uses a counting `RequestHandler` to assert zero queries and at least N point reads are issued when every PK is distinct.
- **`ReadMany_MixedPartitionLayout_UsesBothPaths`** — asserts both branches execute and the merged result is correct when a request mixes single- and multi-tuple partitions.
- **`ReadMany_SingleTuplePartitionMissingItem_SilentlyOmitted`** — asserts a 404 on a point-read branch is treated as missing-item (matches Java `pointReadsForReadMany` + Java PRs #34966/#35513) and does not fail the overall call.

## Verification

- `dotnet build Microsoft.Azure.Cosmos/src/Microsoft.Azure.Cosmos.csproj -c Release` — 0 warnings, 0 errors.
- `dotnet build Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/...` — 0 warnings, 0 errors.
- `dotnet build Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/...` — 0 warnings, 0 errors.
- Unit-test run: 2781 passed / 16 pre-existing failures (all `BinaryEncodingOverTheWireTests` socket errors from external dataset downloads + one unrelated parsing-baseline test). None ReadMany-related.
- Emulator tests not run locally (emulator unavailable on the build agent); will run in PR CI.

## Out of scope

- Doc-size-based dispatch heuristic (Java/Python don't do this either).
- Public API changes — no new methods or option flags.
- Python's additional "all tuples share one logical PK -> `pk = @p AND c.id IN (...)`" optimization (Java lacks it too — tracking separately if needed).
- Trace baseline XML for `EndToEndTraceWriterBaselineTests.ReadManyAsync` — a one-time regeneration (under review) may be needed to capture the new child trace nodes; flagged for follow-up commit if CI surfaces a diff.

## Risk / behavior change

Customer-observable: RU and latency profile improves on the sparse-multi-partition workload. The only externally-visible semantic change is on a path that previously returned a query failure: a 404 on a missing single-tuple item is now treated as "item not present" (consistent with Java and Python; previously this scenario already returned an empty result set via the query path because `SELECT * WHERE c.id IN ('missing')` returns 0 documents -- so end-to-end behavior is unchanged in practice).

---


Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -397,6 +441,157 @@ private async Task<List<ResponseMessage>> GenerateStreamResponsesForPartitionAsy
...
+            }
+            catch
+            {
+                this.CancelCancellationToken(cancellationToken);
+                throw;
+            }
+
+            using (pointReadResponse)
+            {
+                string containerRid = await this.container.GetCachedRIDAsync(forceRefresh: false, trace, cancellationToken);
```
FabianMeiswinkel: NIT - you do not need the Container RID from Container metadata cache - each document response has _rid - which for a document conatins the Container Rid in the Uri

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -397,6 +441,157 @@ private async Task<List<ResponseMessage>> GenerateStreamResponsesForPartitionAsy
...
+                        result: elements,
+                        count: elements.Count,
+                        responseHeaders: responseHeaders,
+                        serializationOptions: null,
+                        trace: trace)
+                };
+            }
+        }
+
+        private static async Task<CosmosElement> ReadStreamAsCosmosElementAsync(
```
FabianMeiswinkel: This should probably be internal and thoroughly unit tested with various stream types and edge cases (different positions in stream, allow vs. disallow getBuffer etc.)

The .Net Stream API is very powerful but full of subtle gotchas - so, testing the heck out of this method with unit tests will help codifying expectations/contracts.

FabianMeiswinkel (dismissed):
LGTM exceptr for the missing unit tests on the stream processing

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
ananth7592: Fixed in 1d5e51524. `GeneratePointReadResponseForPartitionAsync` now parses the response body once and derives the parent container RID from the document's `_rid` field via a small `TryGetContainerRidFromDocument` helper, mirroring the canonical pattern used by `CollectionCache` and `NetworkAttachedDocumentContainer`. The `GetCachedRIDAsync` round-trip is gone.

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
ananth7592: Done in 1d5e51524.

* Visibility flipped to `internal static`.
* New `Microsoft.Azure.Cosmos.Tests/ReadManyQueryHelperUnitTests.cs` adds **31 invocations** that pin the contract end-to-end:
  - **Null / empty inputs:** `null` stream short-circuits; `MemoryStream` with `Length == 0` returns null; non-`MemoryStream` that yields zero bytes returns null.
  - **`MemoryStream` variants:** publicly-visible (default ctor + `Write` → `TryGetBuffer` fast path) **and** private-buffer (`new MemoryStream(byte[])` → `ToArray()` fallback) — each sanity-asserted via `TryGetBuffer` first so the test fixture itself can't drift.
  - **Position semantics, explicitly asymmetric:** `MemoryStream` at `Position == Length` still returns full content (TryGetBuffer/ToArray ignore position); non-`MemoryStream` at non-zero position reads from current position to EOF (CopyToAsync contract). Both pinned as separate tests so a future "unify the branches" refactor can't silently regress either side.
  - **Chunked-read reassembly:** parametric across 1-byte, 8-byte, and single-shot chunk sizes via a custom `ChunkedReadOnlyStream : Stream` fixture.
  - **Cancellation:** pre-cancelled token throws `TaskCanceledException` on the non-`MemoryStream` branch (CopyToAsync honors the token) **but** does not throw on the `MemoryStream` branch (no async ops, token never consulted) — captured as two tests so any future "honor cancellation everywhere" change is forced to be deliberate.
  - **JSON shape parametric:** object / array / number / string / bool(true&false) / null — each asserted to surface as the right `Cosmos*` element subtype.

Teeth-checked: sabotaging the helper (skipping the `Length == 0` short-circuit + truncating the buffer) trips 17 of the stream tests and the canonical valid-RID test as well — so the suite is sensitive to the contract, not just to surface presence.

xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -85,8 +86,51 @@ internal async Task<List<ResponseMessage>[]> ReadManyTaskHelperAsync(IDictionary
...
+            // When a caller sets IfMatchEtag / IfNoneMatchEtag on the ReadMany call, the
+            // multi-id query path effectively ignores them (the per-document precondition
+            // has no place to land on a multi-doc SELECT), but the single-doc point read
+            // honors them strictly. To avoid same-input/different-outcome behavior driven
+            // purely by hash-bucketing (one (id, pk) per physical partition would hit the
+            // fast path and surface a 304/412 that the query path would never produce),
+            // skip the fast path whenever either ETag header is set and let everything
+            // route through the query path uniformly. Java sidesteps the issue at the API
+            // boundary by not exposing ETag on CosmosReadManyRequestOptions at all.
+            bool readManyEtagSet = !string.IsNullOrEmpty(readManyRequestOptions?.IfMatchEtag)
```
xinlian12: have few questions on this: 

1. does IfMatchEtag and ifNoneMatchEtag really make a difference for readItem option? See the discussion https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4249 and also a local test shows that these are not being honored for readItem seems, please verify 
2. Even if IfMatchEtag and ifNoneMatchEtag works for read operation, but why not just ignore it for read operation for now to match the existing behavior (changing to query also ignores these) 
3. If IfMatchEtag works, and we want to expose it, then exposing one IfMatchEtag on the read many request option level feels a little weird, I feel the contract will need to change

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -397,6 +441,205 @@ private async Task<List<ResponseMessage>> GenerateStreamResponsesForPartitionAsy
...
+        /// <summary>
+        /// Shapes the synthetic <see cref="CosmosQueryResponseMessageHeaders"/> that the
+        /// point-read fast path hands to <see cref="QueryResponse.CreateSuccess"/> /
+        /// <see cref="QueryResponse.CreateFailure"/>. Mirrors the wire response's
+        /// SubStatusCode so non-404 failures (e.g. 400/410/503) surface the same
+        /// diagnostic signal a caller would see on a direct ReadItemStreamAsync call;
+        /// without this, the stream overload's combined ResponseMessage would surface
+        /// SubStatusCode=Unknown for those failures.
+        /// </summary>
+        private static CosmosQueryResponseMessageHeaders BuildSyntheticQueryResponseHeaders(
```
xinlian12: 🟡 **Recommendation — Defensive Guard: `containerRid` can be null on the 404 and failure paths, creating a latent `ArgumentNullException` in the `Lazy<MemoryStream>`**

On lines 500 and 515, `BuildSyntheticQueryResponseHeaders` receives `containerRid: null`. This null flows into `QueryResponse.CreateSuccess` (line 86 of QueryResponse.cs), which captures it in a `Lazy<MemoryStream>`:

```csharp
Lazy<MemoryStream> memoryStream = new Lazy<MemoryStream>(() => CosmosElementSerializer.ToStream(
    responseHeaders.ContainerRid,  // null
    result, ...));

```

If `.Content` on this `QueryResponse` is ever accessed, `CosmosElementSerializer.ToStream` calls `jsonWriter.WriteStringValue(containerRid)` → `Encoding.UTF8.GetByteCount(null)` → **`ArgumentNullException`**.

**Why it's safe today:** Both `CombineStreamsFromQueryResponses` and `CombineFeedResponseFromQueryResponses<T>` read `queryResponse.CosmosElements` directly and never access `.Content`, so the Lazy is never realized.

**Why it's worth guarding:** Any future code path that accesses `.Content` (e.g., a diagnostic enhancement, logging, or middleware) would crash with an NRE on a missing-item response. The same risk applies on the success path if `TryGetContainerRidFromDocument` returns null (malformed `_rid`).

**Suggestion:**

```csharp
containerRid: containerRid ?? string.Empty

```
in `BuildSyntheticQueryResponseHeaders`, or guard `containerRid` at the call sites. This makes the Lazy safe to realize regardless of caller.

Microsoft.Azure.Cosmos/src/ReadManyRequestOptions.cs
```diff
@@ -94,5 +94,20 @@ internal QueryRequestOptions ConvertToQueryRequestOptions()
                 ExcludeRegions = this.ExcludeRegions
             };
         }
+
+        internal ItemRequestOptions ConvertToItemRequestOptions()
```
xinlian12, line 98: 🟢 **Suggestion — Test Coverage: Add a unit test for `ConvertToItemRequestOptions` property mapping fidelity**

Both `ConvertToQueryRequestOptions()` (line 83) and the new `ConvertToItemRequestOptions()` (line 98) map the same 8 properties. Neither has a unit test that asserts every property transfers correctly.

**Why it matters:** When a new property is added to `ReadManyRequestOptions` (e.g., `ReadConsistencyStrategy` was added in PR #5685), it must be mapped in *both* conversion methods. Without a test, a copy-paste miss would silently drop the property on the point-read path — the kind of bug that's hard to catch at runtime because it degrades behavior (e.g., wrong consistency level) rather than crashing.

**Suggestion:** Add a unit test that constructs a `ReadManyRequestOptions` with every property set to a distinguishable non-default value, calls `ConvertToItemRequestOptions()`, and asserts each field on the returned `ItemRequestOptions`. Consider adding the same test for `ConvertToQueryRequestOptions()` as a follow-up.

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -397,6 +441,205 @@ private async Task<List<ResponseMessage>> GenerateStreamResponsesForPartitionAsy
...
+            try
+            {
+                pointReadResponse = await this.container.ReadItemStreamAsync(
+                    id: itemId,
+                    partitionKey: partitionKey,
+                    trace: trace,
+                    requestOptions: itemRequestOptions,
+                    cancellationToken: cancellationToken);
+            }
+            catch
```
xinlian12, line 468: 💬 **Observation — Pre-existing: `CancelCancellationToken` is a no-op, so sibling tasks are never cancelled on failure**

The `CancelCancellationToken` call here (line 479) and on the non-success branch (line 511) follows the pre-existing pattern from the query path (lines 430, 436). However, this method creates a *new* linked `CancellationTokenSource`, cancels it, then immediately disposes it:

```csharp
using (var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
{
    cts.Cancel(); // cancels this linked CTS, NOT the original token
}

```

Cancelling a linked CTS does **not** propagate cancellation back to the parent token. The intent is clearly to cancel sibling partition tasks on failure, but the original `cancellationToken` remains unaffected — sibling tasks in `Task.WhenAll` continue to completion (consuming RUs unnecessarily on transient failures).

**Not a regression** — the PR faithfully replicates the existing query-path pattern. To actually cancel siblings, `ReadManyTaskHelperAsync` would need a shared `CancellationTokenSource` passed to all tasks. Worth tracking as a follow-up since it affects both old and new code paths.

xinlian12:
✅ **Review complete** (40:01)

xinlian12 (dismissed):
LGTM, thanks

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
```diff
@@ -85,8 +86,51 @@ internal async Task<List<ResponseMessage>[]> ReadManyTaskHelperAsync(IDictionary
...
+            // honors them strictly. To avoid same-input/different-outcome behavior driven
+            // purely by hash-bucketing (one (id, pk) per physical partition would hit the
+            // fast path and surface a 304/412 that the query path would never produce),
+            // skip the fast path whenever either ETag header is set and let everything
+            // route through the query path uniformly. Java sidesteps the issue at the API
+            // boundary by not exposing ETag on CosmosReadManyRequestOptions at all.
+            bool readManyEtagSet = !string.IsNullOrEmpty(readManyRequestOptions?.IfMatchEtag)
+                                 || !string.IsNullOrEmpty(readManyRequestOptions?.IfNoneMatchEtag);
+
             foreach (KeyValuePair<PartitionKeyRange, List<(string, PartitionKey)>> entry in partitionKeyRangeItemMap)
```
xinlian12, line 89: have a follow up question - how collection re-create is being handled currently? will it cause the partitionKeyRangeItemMap to be re-caluclated etc? or it is handled per inner-partition level?

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
ananth7592: Done in 9a92800ac.

**What changed:**
- Removed `IfMatchEtag` / `IfNoneMatchEtag` from `ConvertToItemRequestOptions`. Added a comment block explaining why this method deliberately diverges from `ConvertToQueryRequestOptions`.
- Removed the `readManyEtagSet` fast-path gate in `ReadManyTaskHelperAsync` (no longer needed once the new mapper strips the headers — the divergence the gate was preventing cannot occur).
- Deleted the three emulator tests that were guarding the now-removed gate (`ReadMany_IfMatchEtagSet_BypassesFastPath`, `ReadMany_IfNoneMatchEtagSet_BypassesFastPath`, `ReadMany_NoEtagOptions_FastPathStillEngages_Sanity`).

**Why this design**, matching your three questions:
1. Confirmed — per #4249 doc clarification, `IfMatchEtag` / `IfNoneMatchEtag` apply only to mutating ops (Replace/Delete/Upsert). They are inert headers on point Reads, so passing them through the new fast path accomplishes nothing on the wire.
2. Agreed — the query path silently ignores them today, so the new path matching that silent-ignore preserves end-to-end behavior exactly. No caller observes a change.
3. Agreed — per-item ETags at the ReadMany level are semantically broken by construction. Rather than perpetuate the footgun in the new mapper, the new method is deliberately ETag-blind and the comment documents why.

**Left alone in this PR:** `ConvertToQueryRequestOptions` still copies the two ETag properties (pre-existing behavior). Deprecating `IfMatchEtag` / `IfNoneMatchEtag` on `ReadManyRequestOptions` itself feels like a separate cleanup that should be done on its own PR with proper API-change review. Happy to file a follow-up issue for that if useful.

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
ananth7592: Good catch, thanks. Confirmed it is pre-existing — the linked-CTS-and-throw-away pattern shows up at four call sites in `ReadManyQueryHelper` and predates this PR. PR #5905 faithfully replicates it for symmetry between the two execution branches.

Tracking the fix as a follow-up so this PR can stay scoped to the single-physical-partition optimization: **#5935** ([ReadMany: sibling partition tasks not cancelled on first failure (CancelCancellationToken is a no-op)](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5935)). The issue captures the root cause, impact, and a suggested fix (replace the per-call linked CTS with a shared CTS that lives for `ReadManyTaskHelperAsync` and gets cancelled on any partition failure so siblings actually observe it).

Microsoft.Azure.Cosmos/src/ReadManyRequestOptions.cs
ananth7592: Done in 44fa6e80d.

Added `ReadManyRequestOptionsTests` with five unit tests covering both mappers:

1. **`ConvertToItemRequestOptions_CopiesAllMappedPropertiesWithDistinguishableValues`** — explicit per-property assertion using a fully-populated `ReadManyRequestOptions` (every value chosen to be distinguishable so a wrong-target copy would fail).
2. **`ConvertToItemRequestOptions_DoesNotCopyETagHeaders_PolicyAssertion`** — codifies the design decision from the prior commit (9a92800ac, in reply to comment #3382347442) so a future reintroduction of `IfMatchEtag` / `IfNoneMatchEtag` on the point-read mapper trips a test rather than silently changing wire behavior.
3. **`ConvertToQueryRequestOptions_CopiesAllMappedPropertiesWithDistinguishableValues`** — symmetric coverage on the legacy mapper, addressing your follow-up suggestion.
4. **`EveryPublicSettableProperty_OnReadManyRequestOptions_IsClassified`** — reflection-based completeness guard. Every public-settable property on `ReadManyRequestOptions` (including those inherited from `RequestOptions`) must be classified into one of three allowlists:
   - `MappedOnBothMappers` (the expected baseline — 5 properties)
   - `MappedOnQueryOnly` (deliberate divergence — `IfMatchEtag`, `IfNoneMatchEtag`)
   - `NotMappedByEither_PreExistingGap` (`PriorityLevel`, `CosmosThresholdOptions`, `AvailabilityStrategy` — documented gap, separate scope)
   
   Any new property added to either type that isn't deliberately classified fails this test. That's exactly the "silently dropped on the point-read path" failure mode you flagged.
5. **`ClassifiedProperties_MustExist_OnReadManyRequestOptions_OrInheritedFromRequestOptions`** — guards the other direction so renames/removals don't leave stale allowlists.

**Verification:**
- All 5 tests pass.
- All 38 `ReadMany*` unit tests in the project pass (33 prior + 5 new).
- Teeth-checked test 2 by re-adding `IfMatchEtag` to the mapper → `ConvertToItemRequestOptions_DoesNotCopyETagHeaders_PolicyAssertion` fails as expected → reverted.
- Teeth-checked test 4 by removing `ExcludeRegions` from the allowlist → `EveryPublicSettableProperty_OnReadManyRequestOptions_IsClassified` correctly reports `Unclassified properties (likely newly added): ExcludeRegions` → reverted.

**Note on the pre-existing gap:** the reflection test surfaces that `PriorityLevel`, `CosmosThresholdOptions`, and `AvailabilityStrategy` are inherited from `RequestOptions` but copied by neither mapper today. Mapping each of those is a separate behavior change on top of pre-existing code — out of scope for this PR, but now visible to future maintainers via the explicit `NotMappedByEither_PreExistingGap` bucket.

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
ananth7592: "Will it cause partitionKeyRangeItemMap to be re-calculated?" → No. The map is built once in ExecuteReadManyRequestAsync before fan-out and is never rebuilt within a single ReadMany call. This is unchanged from pre-existing behavior.
 - "Or is it handled per inner-partition level?" → Yes. Each per-partition sub-request (query OR point read) flows through the standard client pipeline → NamedCacheRetryHandler → InvalidPartitionExceptionRetryPolicy. On 410 + NameCacheIsStale (collection-recreate) the policy sets ForceNameCacheRefresh = true, clears routing hints, and retries once. The new point-read branch inherits this for free because container.ReadItemStreamAsync traverses the same RequestInvokerHandler root as container.GetItemQueryStreamIterator.

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
ananth7592: Addressed in [`a3d3fc3a9`](https://github.com/Azure/azure-cosmos-dotnet-v3/commit/a3d3fc3a9).

**Fix** — single insertion point inside `BuildSyntheticQueryResponseHeaders`:
```diff
- containerRid: containerRid
+ containerRid: containerRid ?? string.Empty
```
Covers all three callers (404/Unknown branch L491, !IsSuccess branch L506, and the success-with-malformed-_rid case at L540) with no behavior change on the happy path where a real RID flows through.

**Tests** — promoted the helper from `private static` → `internal static` (matching the visibility pattern of `ReadStreamAsCosmosElementAsync` and `TryGetContainerRidFromDocument`) and added 6 unit tests in `ReadManyQueryHelperUnitTests.cs`:

| Test | What it pins |
| --- | --- |
| `_NullContainerRid_IsCoercedToEmptyString` | The fix itself — direct contract |
| `_NullContainerRid_QueryResponseContentMaterializationDoesNotThrow` | End-to-end repro: builds the synthetic header with `containerRid: null`, hands it to `QueryResponse.CreateSuccess`, forces `.Content` realization, and asserts the JSON envelope is well-formed with `"_rid":""` and `"_count":0` |
| `_NonNullContainerRid_IsPreservedExactly` | No over-coercion on the success path |
| `_EmptyContainerRid_IsPreservedExactly` | Pins the **null-only** coalesce so a future refactor cannot silently widen it to `IsNullOrWhiteSpace` |
| `_ForwardsRequestChargeActivityIdAndSubStatusCode_FromUpstream` | Pins the helper's original raison-d'être (mirroring upstream `SubStatusCode` so non-404 failures surface their real diagnostic signal — e.g. 410/`PartitionKeyRangeGone`) |
| `_MissingUpstreamSubStatusCode_DefaultsToUnknown` | Regression on the `?? SubStatusCodes.Unknown` belt-and-suspenders coalesce |

**Teeth-checked** — reverted the `?? string.Empty` and re-ran the materialization test. It fails with the exact stack you predicted:
```
System.ArgumentNullException: Value cannot be null. (Parameter 'chars')
  at System.Text.UTF8Encoding.GetByteCount(String chars)
  at Microsoft.Azure.Cosmos.Json.JsonWriter.WriteStringValue(String value)
  at Microsoft.Azure.Cosmos.Serializer.CosmosElementSerializer.ToStream(...)
  at Microsoft.Azure.Cosmos.QueryResponse.<>c__DisplayClass17_0.<CreateSuccess>b__0()
  at System.Lazy`1.get_Value()
  at Microsoft.Azure.Cosmos.QueryResponse.get_Content()
```
Reverted. Build clean (0 warnings, 0 errors); **44/44 ReadMany unit tests pass** (38 prior + 6 new).

Microsoft.Azure.Cosmos/src/ReadManyQueryHelper.cs
xinlian12: would not this cause issue? since after re-create, the documents might lands on different partitions?

xinlian12 (approved):
LGTM, thanks

kundadebdatta (approved):
LGTM

PR #5912, opened 2026-05-28
Pagination + RetryPolicies: Fixes CancellationToken not being honored across PartitionKeyRangeCache, query pagination, and retry/backoff paths

### LINKED ISSUE #5906: ClientRetryPolicy.ShouldRetryAsync does not honor operation CancellationToken — FeedIterator can run far past the user's timeout under 410/1022 LeaseNotFound storms (partition merge / lease handoff)
### Summary

During a partition merge / lease handoff on a cross-partition `FeedIterator` query, the SDK can run for **many multiples** of the user's `CancellationToken` deadline before surfacing `TaskCanceledException`. We observed a 100-second wall-clock under a 30-second token. The retry plumbing for `410/1022 LeaseNotFound` is correct functionally (cross-region failover happens, request eventually succeeds elsewhere), but it is **not budgeted against the operation-level `CancellationToken`**.

This is reproducible during routine partition merge cutovers (and lease handoffs more generally) on any account where prefetch is active and the cross-partition fan-out is more than one.

### Repro context

- SDK: `Microsoft.Azure.Cosmos` **3.56.0**, Direct mode, .NET 8, Windows.
- Account: multi-region with a single write region, Strong consistency.
- Operation: `Typed FeedIterator ReadNextAsync` against a cross-partition collection during a partition merge cutover.
- User-supplied `CancellationToken` budget: **30 seconds**.
- Observed wall-clock until `TaskCanceledException`: **99,602 ms**.
- `CosmosDiagnostics.Summary` showed:
  - `DirectCalls`: `(200,0)=6`, `(204,0)=267`, `(410,1022)=6`.
  - `GatewayCalls`: `(200,0)=1`.
- The customer's in-flight prefetched pages were addressed to the source partition during the merge handoff and received `410/1022 LeaseNotFound`.

Customer-facing impact: connections held open for ~2 minutes → socket exhaustion and thread starvation in their downstream service, even though the SDK eventually returned (or, in this case, surfaced the cancellation).

### Where in the code

The retry path for `410/1022 LeaseNotFound` today:

1. **Direct package**: `GoneAndRetryWithRequestRetryPolicy` does same-region replica retries on `410/1022` (closed-source; exposed via symbols `IsGoneWithLeaseNotFound`, `LeaseNotFoundException` in `Microsoft.Azure.Cosmos.Direct.dll`).
2. **v3 SDK**: `Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs` `ShouldRetryAsync` → `ShouldRetryInternalAsync` does cross-region failover.
   - Added in #5108 (shipped 3.50.0).
   - Hardened for BarrierRequests in #5469 (shipped 3.55.0).
   - Both are present in 3.56.0, so the functional retry path is correct.

The gap is here:

```csharp
// Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs:71
public async Task<ShouldRetryResult> ShouldRetryAsync(
    Exception exception,
    CancellationToken cancellationToken)   // <-- received
{
    // ...
    if (exception is DocumentClientException clientException)
    {
        // ...
        ShouldRetryResult shouldRetryResult = await this.ShouldRetryInternalAsync(
            clientException?.StatusCode,
            clientException?.GetSubStatus());   // <-- CT not passed
        if (shouldRetryResult != null)
        {
            return shouldRetryResult;
        }
    }
    // ...
    return await this.throttlingRetry.ShouldRetryAsync(exception, cancellationToken);
}

// Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs:220
private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
    HttpStatusCode? statusCode,
    SubStatusCodes? subStatusCode)   // <-- no CT in signature
{
    // ... returns ShouldRetryResult.RetryAfter(...) without ever
    // observing cancellationToken.IsCancellationRequested
}
```

`ShouldRetryAsync` receives the user's `CancellationToken` but never calls `cancellationToken.ThrowIfCancellationRequested()` and does not forward it into `ShouldRetryInternalAsync`. For any status code it handles itself (`410/1022`, `503`, `403/3`, `404/1002`, etc.), it returns `ShouldRetryResult.RetryAfter(...)` based purely on policy state. If the CT has already expired when the retry decision is made, the policy still answers "retry now" — the next attempt is scheduled and only stops at the next physical socket-read await.

Combined with `FeedIterator` prefetch fanning out multiple pages in flight, a single user operation can spawn **N concurrent same-region + cross-region retry chains**, each consuming the full local + cross-region retry budget independently. The operation-level CT does not preempt them.

### What we'd like

Some combination of:

1. **CT-aware retry decisions.** `ClientRetryPolicy.ShouldRetryAsync` should call `cancellationToken.ThrowIfCancellationRequested()` (or otherwise return `NoRetry()`) before returning a `RetryAfter` result. Plumb `CancellationToken` into `ShouldRetryInternalAsync` and check it there too.
2. **CT propagation into prefetched in-flight pages.** When the operation-level CT fires, `FeedIterator`'s prefetched/in-flight pages should be cancelled rather than each chain running its retry budget to completion before its `await` finally observes the cancellation.
3. **Optional per-operation cumulative retry budget.** A `CosmosClientOptions` knob (or `QueryRequestOptions`) that caps wall-clock retry latency for an operation, independent of the user's CT. This gives customers a deterministic upper bound on "how long can a partition event delay my query?" without having to size their CT for the worst case.

(1) is the smallest change and would, by itself, fix the reported "99.6 s under a 30 s CT" symptom. (2) closes the gap that prefetch creates. (3) is a defense-in-depth option for customers who can't tightly bound the CT.

### Acceptance

A unit test or emulator test that:
- Configures `FeedIterator` over a partitioned collection with prefetch enabled.
- Injects synthetic `410/1022` responses on N pages concurrently.
- Asserts the operation surfaces `OperationCanceledException` within `userCt.Cancel()` + `tolerance` (e.g., ≤ 1 s) rather than running each retry chain to completion.

tvaron3:
## Description

Fixes Azure/azure-cosmos-dotnet-v3#5906.

When a customer's operation-level `CancellationToken` expires during a cross-partition `FeedIterator` page (or any retry-driven `ReadItem` / point-write), the SDK was continuing to do work past the deadline. The two main offenders:

1. **`PartitionKeyRangeCache` lookups** used by query pagination (cross-partition fan-out, split / merge handling, and PKRange cache refresh between pages) ran `do { ... await ProcessMessageAsync(...) ... } while(...)` against `forceRefresh = true` without checking the caller's CT between iterations. A 410/1022 storm during a partition merge cutover could keep the loop spinning until the per-attempt RNTBD timeout fired, well past the customer's deadline.
2. **Retry policies** (`ClientRetryPolicy`, `ResourceThrottleRetryPolicy`, `WebExceptionRetryPolicy`, `BulkExecutionRetryPolicy`) did not short-circuit when the operation CT was already cancelled, so each policy would schedule another backoff `Task.Delay` against the same expired token, surfacing as a stalled operation rather than `OperationCanceledException`.

The end-to-end symptom was the same as #5906: ~99–133 s wall-clock under a 30 s `CancellationToken`, with the entire stall sitting inside `Prefetching` / `PartitionKeyRangeCache` refresh paths and no forward progress on transport requests.

This PR plumbs the operation-level `CancellationToken` end-to-end and adds `ThrowIfCancellationRequested` short-circuits inside the retry loops, so cancellation is observed within the next backoff slot (not after the full retry budget elapses).

### Retry policies — early CT short-circuit

All four retry policies now check `cancellationToken.IsCancellationRequested` *before* scheduling the next backoff `Task.Delay`, so a cancelled operation stops within the current attempt rather than burning the full retry budget:

- **`ClientRetryPolicy.ShouldRetryAsync`** — cross-region retries on `Gone (410)` / `LeaseNotFound (410/1022)` and other transient backend status codes. The pre-cancelled path also avoids `RefreshLocationAsync` round-trips that would otherwise hold the operation open.
- **`ResourceThrottleRetryPolicy.ShouldRetryAsync`** — `429` throttling backoffs (up to 60 s by default, which is what dominated the original stall).
- **`WebExceptionRetryPolicy.ShouldRetryAsync`** — exponential backoff on retriable HTTP / socket exceptions.
- **`BulkExecutionRetryPolicy.ShouldRetryAsync` (both overloads) and `ShouldRetryInternalAsync`** — defense-in-depth `ThrowIfCancellationRequested` at the top of the internal method covers the race where the token cancels between the public guard and the cache call.

### `PartitionKeyRangeCache` — CT inside the refresh loop

- `TryGetOverlappingRangesAsync`, `TryGetPartitionKeyRangeByIdAsync`, and `TryLookupAsync` now accept a `CancellationToken` parameter.
- `GetRoutingMapForCollectionAsync` checks the token at the top of every iteration of its `do { ... } while(...)` refresh loop, surfacing `OperationCanceledException` instead of issuing another gateway round-trip when the caller's deadline has passed.
- The legacy 4-arg `TryLookupAsync` overload is retained for back-compat (callers like `AddressResolver` and `DocumentClient` that have no CT in scope are documented as a deferred follow-up).

### Query pagination + FeedRange plumbing — forward the operation CT end-to-end

Each of these methods previously either accepted no CT or accepted one and dropped it before reaching `PartitionKeyRangeCache`. They now thread the caller's CT all the way down:

- **`ContainerCore.GetFeedRangesAsync`** → `PartitionKeyRangeCache.TryGetOverlappingRangesAsync`
- **`FeedRange.GetEffectiveRangesAsync`** — abstract `FeedRangeInternal` plus all three overrides (`FeedRangeEpk`, `FeedRangePartitionKey`, `FeedRangePartitionKeyRange`), with `ThrowIfCancellationRequested` at the top of the synchronous overrides and CT forwarded into both `TryGetPartitionKeyRangeByIdAsync` calls in `FeedRangePartitionKeyRange`.
- **All `FeedRange` callers** updated to pass the operation CT: `FeedRangePartitionKeyRangeExtractor` (3 sites), `ContainerCore.Items` (2 sites), `CosmosQueryClientCore`.
- **`PartitionRoutingHelper`** — `TryGetTargetRangeFromContinuationTokenRangeAsync`, `GetReplacementRangesAsync`, and `TryAddPartitionKeyRangeToContinuationTokenAsync` gained CT parameters with `ThrowIfCancellationRequested` short-circuits, and forward the CT to every `TryGetOverlappingRangesAsync` and `TryGetRangeByEffectivePartitionKeyAsync` call — including the empty-`providedPartitionKeyRanges` (`WHERE false`) branch and the forward-continuation branch that had previously dropped the token.
- **`PartitionKeyRangeHandler`** — 3 caller updates to forward CT into the helper above.
- **v2 `DefaultDocumentQueryExecutionContext.ExecuteOnceAsync`** — private `TryGetTargetPartitionKeyRangeAsync` gained CT; 3 caller updates.
- **`FeedRangeCompositeContinuation.HandleSplitAsync`** — private `TryGetOverlappingRangesAsync` helper gained CT.

### Additional CT-honor paths

- **`CosmosHttpClientCore`** — gateway HTTP retry loop now honors CT between attempts.
- **`GlobalAddressResolver`** — open-connection paths honor the supplied CT.
- **`RequestInvokerHandler`** — operation CT is forwarded into the request pipeline.

### Unit tests pinning the contract at every plumbed layer

Each plumbing point above has a dedicated unit test that would fail if the contract regresses (CT dropped on the way down, OCE not surfaced, retry policy schedules another delay past the deadline). Pattern: `Mock<...>(MockBehavior.Strict)` plus `.Callback<...>(... => captured = ct)` so the assertion verifies the *same CT instance* reaches the underlying provider — not just that the method completes.

- **`PartitionKeyRangeCacheTest`**
  - `TryGetPartitionKeyRangeByIdAsync_WithCancelledToken` — short-circuit before any store-model call (`Times.Never`).
  - `GetRoutingMapForCollectionAsync_TokenCancelledMidRetryLoop` — cancels inside the first `ProcessMessageAsync` callback; asserts no second iteration of the `do { ... } while(...)` loop runs.
- **`FeedRangeTests`**
  - `FeedRangeEPK_GetEffectiveRangesAsync_WithCancelledToken_ThrowsOperationCanceledException`
  - `FeedRangePK_GetEffectiveRangesAsync_WithCancelledToken_ThrowsOperationCanceledException`
  - `FeedRangePKRangeId_GetEffectiveRangesAsync_PropagatesCancellationTokenToRoutingMapProvider`
- **`PartitionRoutingHelperTest`**
  - `GetReplacementRangesAsync_PropagatesCancellationTokenToRoutingMapProvider`
  - `TryGetTargetRangeFromContinuationTokenRangeAsync_WithCancelledToken_ThrowsOperationCanceledException`
  - `TryGetTargetRangeFromContinuationTokenRangeAsync_EmptyProvidedRanges_PropagatesCancellationTokenToRoutingMapProvider` — pins the `WHERE false` / empty-provided-ranges branch that the deep reviewer flagged.
- **`BulkExecutionRetryPolicyTests`** — cancelled-token short-circuit on both public overloads.
- **`ClientRetryPolicyTests`** — cancelled-token short-circuit covers the 410 / 1022 cross-region path.
- **`ResourceThrottleRetryPolicyTests`** — cancelled-token short-circuit on 429 backoff path; asserts no `Task.Delay` is scheduled.
- **`WebExceptionRetryPolicyTests`** — cancelled-token short-circuit on retriable HTTP / socket exception backoff.
- **`CosmosHttpClientCoreTests`** — gateway HTTP retry loop CT honor.
- **`NetworkAttachedDocumentContainerTests`** — CT propagation through the change-feed `MonadicRefreshProviderAsync` path.
- **`PartitionKeyRangeHandlerTests`** — Moq fixtures updated to bind the new CT parameter.

### Fault-injection end-to-end repro tests

Deterministic e2e tests that fail on `master` and pass on this branch. Each pairs a fault rule with a CT shorter than the SDK's default retry budget and asserts (a) `OperationCanceledException` is surfaced and (b) wall-clock stays inside the budget + a small grace period.

- **`FaultInjectionLeaseNotFound_FeedIteratorHonorsCancellationToken`** (Direct mode) — cross-partition `FeedIterator` running `SELECT TOP 1 ... ORDER BY`; injects `LeaseNotFound (410/1022)` on `QueryItem` / `ReadItem` against the primary region with `WithTimes(int.MaxValue)`; secondary region also injected with a deterministic delay; 5 s CT, ≤ 7 s wall-clock assertion. Includes a cold-start warm-up `ReadItemAsync` (catches `NotFound`) before enabling the rule and starting the stopwatch, so the boundary is not raced by the first-request channel-open / cache-build.
- **`FaultInjectionMetadataRequestDelay_HonorsCancellationToken`** (Direct mode) — Strong-consistency multi-region read with preferred regions ordered to force the consistency-barrier `Head/Collection` path; injects `ResponseDelay(20 s)` on every `MetadataRequest` in the primary region; 5 s CT; returns `Assert.Inconclusive` if the harness doesn't actually exercise the barrier path so it can't false-pass.
- **`FaultInjectionGateway_ResponseDelay_HonorsCancellationToken`** (Gateway mode) — Gateway-mode twin with a matching cold-start warm-up.

### Fault-injection primitives added to support these tests

- **`FaultInjectionServerErrorType.LeaseNotFound`** (`410 / 1022`) — Direct (`StoreResponse`) + Gateway (`HttpResponseMessage`) result paths, modeled on the existing `PartitionIsMigrating` (`410 / 1008`).
- **`FaultInjectionOperationType.MetadataRequest`** → `OperationType.Head` — lets fault rules target the consistency-barrier requests issued under Strong / Bounded-Staleness consistency.
- Gateway and Direct status-code `[DataRow]` smoke tables extended with `LeaseNotFound`.
- `Microsoft.Azure.Cosmos/FaultInjection/README.md` updated with the new error-type row and a `MetadataRequest` note.

## Type of change

## Closing issues

closes #5906

## Changelog

The changelog entry under `### Other Changes` also discloses that callers whose token fires after the SDK has already started a retry attempt or a metadata refresh may now observe the original retriable `CosmosException` (e.g. `429`, `410/1022`) instead of an `OperationCanceledException`, depending on which phase of the backoff / refresh was active. The contract that the operation no longer keeps running past the caller's deadline is unchanged.

### DIFF
--- a/Microsoft.Azure.Cosmos/FaultInjection/README.md
+++ b/Microsoft.Azure.Cosmos/FaultInjection/README.md
@@ -28,10 +28,13 @@ This result will return a server error to the customer: `FaultInjectionServerErr
 | `Timeout`               | 408:0        | The operation did not complete within the allocated time.                   |
 | `PartitionIsSplitting`  | 410:1007     | The partition is currently splitting.                                       |
 | `PartitionIsMigrating`  | 410:1008     | The partition is currently migrating.                                       |
+| `LeaseNotFound`         | 410:1022     | The replica's lease for the addressed documentServiceId is not found. Fires during partition merge / lease handoff. |
 | `SendDelay`             | n/a          | Will inject a delay to the request before it is sent to the backend.         |
 | `ResponseDelay`         | n/a          | Will inject a delay to the request after a response is received from the backend before returning the result. |
 | `ConnectionDelay`       | n/a          | Used to simulate high channel acquisition.                                  |
 | `ServiceUnavailable`    | 503:0        | The service is currently unavailable.                                       |
+
+> Note: `FaultInjectionConditionBuilder.WithOperationType(FaultInjectionOperationType.MetadataRequest)` targets metadata HEAD requests (e.g., the consistency-barrier `Head/Collection` requests the SDK issues under Strong / Bounded-Staleness consistency to verify `GlobalCommittedLSN` before returning a read).
 
 ##### `FaultInjectionConnectionErrorResult`
 
--- a/Microsoft.Azure.Cosmos/FaultInjection/src/FaultInjectionOperationType.cs
+++ b/Microsoft.Azure.Cosmos/FaultInjection/src/FaultInjectionOperationType.cs
@@ -79,6 +79,15 @@ namespace Microsoft.Azure.Cosmos.FaultInjection
         MetadataQueryPlan,
 
         /// <summary>
+        /// Metadata / barrier HEAD requests (e.g., consistency-barrier
+        /// <c>Head</c> against the <c>Collection</c> resource the SDK issues under
+        /// Strong / Bounded-Staleness consistency to verify <c>GlobalCommittedLSN</c>
+        /// on a replica before returning a read result).
+        /// Maps to <see cref="Microsoft.Azure.Documents.OperationType.Head"/>.
+        /// </summary>
+        MetadataRequest,
+
+        /// <summary>
         /// All operation types. Default value.
         /// </summary>
         All,
--- a/Microsoft.Azure.Cosmos/FaultInjection/src/implementation/FaultInjectionRuleProcessor.cs
+++ b/Microsoft.Azure.Cosmos/FaultInjection/src/implementation/FaultInjectionRuleProcessor.cs
@@ -354,6 +354,7 @@ namespace Microsoft.Azure.Cosmos.FaultInjection
                 FaultInjectionOperationType.MetadataPartitionKeyRange => OperationType.ReadFeed,
                 FaultInjectionOperationType.MetadataRefreshAddresses => OperationType.Invalid,
                 FaultInjectionOperationType.MetadataQueryPlan => OperationType.QueryPlan,
+                FaultInjectionOperationType.MetadataRequest => OperationType.Head,
                 _ => throw new ArgumentException($"FaultInjectionOperationType: {faultInjectionOperationType} is not supported"),
             };
         }
@@ -376,6 +377,7 @@ namespace Microsoft.Azure.Cosmos.FaultInjection
                 FaultInjectionOperationType.MetadataPartitionKeyRange => ResourceType.PartitionKeyRange,
                 FaultInjectionOperationType.MetadataRefreshAddresses => ResourceType.Address,
                 FaultInjectionOperationType.MetadataQueryPlan => ResourceType.Document,
+                FaultInjectionOperationType.MetadataRequest => ResourceType.Collection,
                 _ => throw new ArgumentException($"FaultInjectionOperationType: {faultInjectionOperationType} is not supported"),
             };
         }
--- a/Microsoft.Azure.Cosmos/FaultInjection/src/implementation/FaultInjectionServerErrorResultInternal.cs
+++ b/Microsoft.Azure.Cosmos/FaultInjection/src/implementation/FaultInjectionServerErrorResultInternal.cs
@@ -258,6 +258,20 @@ namespace Microsoft.Azure.Cosmos.FaultInjection
                     };
 
                     return storeResponse;
+
+                case FaultInjectionServerErrorType.LeaseNotFound:
+                    INameValueCollection leaseNotFoundHeaders = args.RequestHeaders;
+                    leaseNotFoundHeaders.Set(WFConstants.BackendHeaders.SubStatus, ((int)SubStatusCodes.LeaseNotFound).ToString(CultureInfo.InvariantCulture));
+                    leaseNotFoundHeaders.Set(WFConstants.BackendHeaders.LocalLSN, lsn);
+
+                    storeResponse = new StoreResponse()
+                    {
+                        Status = 410,
+                        Headers = leaseNotFoundHeaders,
+                        ResponseBody = new MemoryStream(FaultInjectionResponseEncoding.GetBytes($"Fault Injection Server Error: Lease Not Found, rule: {ruleId}"))
+                    };
+
+                    return storeResponse;
                 case FaultInjectionServerErrorType.ServiceUnavailable:
                     INameValueCollection serviceUnavailableHeaders = args.RequestHeaders;
                     serviceUnavailableHeaders.Set(WFConstants.BackendHeaders.LocalLSN, lsn);
--- a/Microsoft.Azure.Cosmos/FaultInjection/tests/FaultInjectionDirectModeTests.cs
+++ b/Microsoft.Azure.Cosmos/FaultInjection/tests/FaultInjectionDirectModeTests.cs
@@ -1799,25 +1799,23 @@ namespace Microsoft.Azure.Cosmos.FaultInjection.Tests
         }
 
         // -----------------------------------------------------------------------------
-        // Issue #5906 / ICM 21000001041367 — partition merge / lease handoff repro
+        // Partition merge / lease handoff repro.
         //
-        // Customer-observed symptom: under a partition merge / lease handoff the SDK
-        // returned 410/1022 (LeaseNotFound) to in-flight cross-partition FeedIterator
-        // pages. The retry plumbing (Direct GoneAndRetryWithRequestRetryPolicy ->
-        // ClientRetryPolicy cross-region retry, PR #5108 in 3.50.0 + #5469 in 3.55.0)
-        // eventually succeeded against the destination region, but each page-level
-        // retry chain ran past the operation-level CancellationToken. End-to-end
-        // wall-clock was ~100 s for a 30 s user CT.
+        // During a partition merge / lease handoff, the backend may return 410/1022
+        // (LeaseNotFound) to in-flight cross-partition FeedIterator pages. The retry
+        // plumbing (Direct GoneAndRetryWithRequestRetryPolicy -> ClientRetryPolicy
+        // cross-region retry) eventually succeeds against the destination region, but
+        // each page-level retry chain can run past the operation-level
+        // CancellationToken.
         //
         // This test installs a LeaseNotFound fault on the primary region only and
         // asserts the operation-level CancellationToken is honored within a small
-        // tolerance window. EXPECTED TO FAIL on master until #5906 ships a fix to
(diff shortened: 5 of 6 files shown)


tvaron3:
Superseded by #5913, which is opened from a branch on the upstream repository so internal CI lanes can run. Closing this fork-based PR to consolidate review.

PR #5916, opened 2026-06-01
Direct: Refactors OpenTcpConnectionTimeout negative-value warning trace to DocumentClient

yumnahussain:
Follow-up to #5873 addressing @kundadebdatta's comment: move the negative-`TimeSpan` warning trace out of the `CosmosClientOptions.OpenTcpConnectionTimeout` setter and into `DocumentClient`, where the value is consumed.

- `CosmosClientOptions`: remove the setter-side `DefaultTrace.TraceWarning`
- `DocumentClient`: emit the warning inside the existing `openTcpConnectionTimeout < TimeSpan.Zero` branch.
- `ConnectionPolicy` / `CosmosClientBuilder`: XML-doc wording sync.
- Tests: drop the obsolete setter-warning test; add `DocumentClientOpenTcpConnectionTimeoutWarningTests` covering negative-emits-warning, truncation-with-warning, setter-silent, non-negative-silent (parameterized), and once-per-construction.
- Changelog: `Unreleased` > `Other Changes` entry (PR #5873 already shipped in 3.61.0-preview.0).

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DocumentClient.cs
+++ b/Microsoft.Azure.Cosmos/src/DocumentClient.cs
@@ -929,11 +929,23 @@ namespace Microsoft.Azure.Cosmos
                 {
                     // Values in [TimeSpan.Zero, 1 second) become 0 (use RequestTimeout).
                     // Values >= 1 second round up to the nearest whole second, clamped to int.MaxValue.
-                    // Negative values are truncated via (int)TotalSeconds, preserving pre-PR behavior.
+                    // Negative values are truncated via (int)TotalSeconds, preserving pre-PR behavior,
+                    // and a warning trace is emitted (Direct mode only) because negative timeouts
+                    // cause the TransportClient to fall back to the configured RequestTimeout.
+                    // The warning is gated on Direct mode because openConnectionTimeoutInSeconds is
+                    // not consumed in Gateway mode, so the message would be misleading there.
                     TimeSpan openTcpConnectionTimeout = connectionPolicy.OpenTcpConnectionTimeout.Value;
 
                     if (openTcpConnectionTimeout < TimeSpan.Zero)
                     {
+                        if (connectionPolicy.ConnectionMode == ConnectionMode.Direct)
+                        {
+                            DefaultTrace.TraceWarning(
+                                "OpenTcpConnectionTimeout value {0} is negative. Negative values are not recommended; "
+                                + "the TransportClient will fall back to the configured RequestTimeout.",
+                                openTcpConnectionTimeout);
+                        }
+
                         this.openConnectionTimeoutInSeconds = (int)openTcpConnectionTimeout.TotalSeconds;
                     }
                     else if (openTcpConnectionTimeout < TimeSpan.FromSeconds(1))
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DocumentClientOpenTcpConnectionTimeoutWarningTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DocumentClientOpenTcpConnectionTimeoutWarningTests.cs
@@ -1,0 +1,270 @@
+﻿//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.Tests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using Microsoft.Azure.Cosmos.Core.Trace;
+    using Microsoft.Azure.Cosmos.Fluent;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    /// <summary>
+    /// Pins the behavior contracted by the follow-up to PR #5873: the
+    /// negative-<see cref="CosmosClientOptions.OpenTcpConnectionTimeout"/>
+    /// warning trace is emitted by <c>DocumentClient</c> at client construction
+    /// (the consumer of the value), not by the
+    /// <c>CosmosClientOptions.OpenTcpConnectionTimeout</c> property setter.
+    /// </summary>
+    [TestClass]
+    [DoNotParallelize]
+    public class DocumentClientOpenTcpConnectionTimeoutWarningTests
+    {
+        private const string AccountEndpoint = "https://localhost:8081/";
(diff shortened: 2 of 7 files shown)


Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -929,11 +929,17 @@ internal virtual void Initialize(Uri serviceEndpoint,
...
                     // Values >= 1 second round up to the nearest whole second, clamped to int.MaxValue.
-                    // Negative values are truncated via (int)TotalSeconds, preserving pre-PR behavior.
+                    // Negative values are truncated via (int)TotalSeconds, preserving pre-PR behavior,
+                    // and a warning trace is emitted because negative timeouts cause the TransportClient
+                    // to fall back to the configured RequestTimeout.
                     TimeSpan openTcpConnectionTimeout = connectionPolicy.OpenTcpConnectionTimeout.Value;
 
                     if (openTcpConnectionTimeout < TimeSpan.Zero)
                     {
+                        DefaultTrace.TraceWarning(
```
NaluTripician: One behavior change worth a conscious decision: this warning now also fires for the legacy path where someone constructs `DocumentClient` directly with a `ConnectionPolicy { ConnectionMode = Gateway, OpenTcpConnectionTimeout = <negative> }`. That path has no validation (unlike `CosmosClientOptions`, which blocks Direct-only settings in Gateway mode via `ValidateDirectTCPSettings`), so previously it emitted no warning. In Gateway mode `openConnectionTimeoutInSeconds` is never consumed, so the message about falling back to RequestTimeout is misleading there. Consider gating the warning on Direct mode, or confirm we're fine treating it as harmless log noise.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DocumentClientOpenTcpConnectionTimeoutWarningTests.cs
```diff
@@ -0,0 +1,219 @@
...
+            this.BuildClientWithOpenTcpConnectionTimeout(TimeSpan.FromSeconds(-2));
+
+            IReadOnlyList<CapturedTraceEvent> warnings = this.listener.SnapshotWarnings(WarningSubstring);
+            Assert.AreEqual(
+                2,
+                warnings.Count,
+                "Two DocumentClient constructions with negative timeouts should emit exactly two warnings (once per construction, not once per setter call).");
+        }
+
+        private CosmosClient BuildClientWithOpenTcpConnectionTimeout(TimeSpan openTcpConnectionTimeout)
```
NaluTripician, line 180: All the new tests route through this helper, which hardcodes `ConnectionMode.Direct`. Two coverage gaps worth pinning:

1. Gateway mode + negative `OpenTcpConnectionTimeout` via a raw `ConnectionPolicy` (the legacy direct-`DocumentClient` path). Whatever we decide about the Gateway-mode warning above, a test would lock the contract.
2. "Set negative, then overwrite with non-negative, then build -> 0 warnings." Moving the warning to construction time is a real improvement (the old setter approach false-positived on the transient negative value); a test would document that intended behavior.

changelog.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Other Changes
 
+- [5916](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5916) Direct: Moves the `OpenTcpConnectionTimeout` negative-value warning trace from `CosmosClientOptions` to `DocumentClient`, so the warning is emitted by the consumer at client construction instead of from a property setter.
```
NaluTripician: Accurate, but it slightly undersells the observable change: the warning now fires once per client construction instead of on each property assignment, and it no longer false-positives when a negative value is later overwritten. Might be worth a few extra words so readers catch the timing shift.

NaluTripician (commented):
Local review of this PR. Overall it's a clean, low-risk refactor that correctly moves the negative-`OpenTcpConnectionTimeout` warning to the point of consumption in `DocumentClient`, matching the direction from the #5873 review. No blocking issues from my side:

- The new tests genuinely exercise the production `DocumentClient.Initialize` negative branch (`MockDocumentClient` chains into the real internal ctor, and `ConnectionTimeout == -5` comes from real `(int)(-5.7)` logic, not a mock override).
- No double-warning: `ConnectionPolicy.OpenTcpConnectionTimeout` is a plain auto-property, so `DocumentClient` is now the single emission point.
- Removing the `Core.Trace` using in `CosmosClientOptions.cs` doesn't break the build (no remaining `DefaultTrace` references there).

Left a few non-blocking notes inline. Minor observations I didn't inline:
- The new tests mutate global `DefaultTrace` state, but they're correctly guarded with `[DoNotParallelize]` and have null-safe cleanup, so it's fine as-is.
- Asserting on the culture-formatted `TimeSpan` in the message is slightly brittle, but `TimeSpan.ToString("c")` is invariant and the listener formats with `InvariantCulture`, so it's safe.
- The added comments and XML-doc wording updates read accurately and are consistent with the implementation.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
yumnahussain: Gated the warning on `ConnectionMode == Direct` so the legacy Gateway path stays quiet. Kept the truncation assignment outside the gate to preserve legacy behavior on `openConnectionTimeoutInSeconds`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DocumentClientOpenTcpConnectionTimeoutWarningTests.cs
yumnahussain: Added a Gateway-mode test (raw `ConnectionPolicy` + negative timeout -> 0 warnings) and a set-negative-then-overwrite test (transient -30s overwritten with +5s before build -> 0 warnings). Extracted a `BuildClientFromConnectionPolicy` helper so the Gateway test bypasses `CosmosClientOptions`.

changelog.md
yumnahussain: Reworded to call out the timing shift (once per construction vs. per assignment) and the no-more-false-positive behavior when a negative value is overwritten or when running in Gateway mode.

kushagraThapar (approved):

NaluTripician (approved):

PR #5923, opened 2026-06-03
CrossRegionalHedging: Adds Metadata Hedging Support

### LINKED ISSUE #5917: [Cross Regional Hedging] Add Support to Hedge Metadata Requests
### Acceptance Criteria:

- Add Support to Hedge Metadata Requests

kundadebdatta:
## Summary

Adds **bounded cross-region hedging** for metadata cache reads so a slow primary region no longer dominates the latency tail of the `Collection` and `PartitionKeyRange` metadata cache reads. Hedging applies to **both** the first-population (cold-start) read **and** steady-state **refresh / force-refresh** reads of those two caches — it is restricted by **request type** (`Collection` `Read`, `PartitionKeyRange` `ReadFeed` first page), **not** by cold start.

The feature is **internal in Phase 1** — there is no public configuration API and no public API/contract surface. The effective opt-in follows the account's PPAF (Per-Partition Automatic Failover) state and can be force-enabled or suppressed through the `AZURE_COSMOS_METADATA_HEDGING_ENABLED` environment variable.

Closes #5917. Full design: [docs/PPAF_Metadata_Hedging_ColdStart_Design.md](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/users/kundadebdatta/5917_implement_metadata_hedging/docs/PPAF_Metadata_Hedging_ColdStart_Design.md).

## Enablement Criteria:

- Metadata Hedging is applicable only for the Read Collection and Get Pk Ranges metadata calls.
- It's applicable for both cold start and warm path scenarios.
- The metadata hedging threshold is chosen between the first HTTP client timeout (i.e. 500 ms) and the second one (i.e. 5 secs). By default, it's kept at 1.5 secs.
- Metadata Hedging is enabled by default in the preview package. In the GA package it is off by default (follows the account's PPAF state) and can be turned on with the `AZURE_COSMOS_METADATA_HEDGING_ENABLED` environment variable.

## Scope change in the latest revision — beyond cold start

Earlier revisions limited hedging to the **cold-start window only**. We decided hedging should also cover **steady-state refresh reads** of the same two caches, because a degraded primary inflates the latency tail on refresh reads just as it does on cold start.

- The `!IsColdStart` eligibility gate in `MetadataHedgingStrategy.EvaluateEligibility` is **removed**; both cold-start and refresh reads of a supported type are now eligible.
- The **request-type restriction stays**: only `Collection` `Read` and `PartitionKeyRange` `ReadFeed` (first page) hedge.
- `IsColdStart` is retained as a **diagnostics-only** signal; `MetadataHedgeSkipReason.NotColdStart` is retired (never produced).
- The opt-in environment variable is renamed `AZURE_COSMOS_METADATA_HEDGING_FOR_COLDSTART_ENABLED` → **`AZURE_COSMOS_METADATA_HEDGING_ENABLED`** to reflect the broader scope.
- Every amplification safeguard is unchanged (see "Not bombarding the Gateway" below).

## How it works

- A single `MetadataHedgingStrategy` is created per `CosmosClient` and consumed by `ClientCollectionCache` (Collection `Read`) and `PartitionKeyRangeCache` (`PartitionKeyRange` `ReadFeed`, **first page only**), for both first-population and refresh reads.
- On an eligible metadata read the strategy issues the **primary** request and starts a timer. If the primary has not produced an *acceptable* response within the threshold, it dispatches a **single hedged** request to a second preferred region and returns the first acceptable winner. The losing branch is cancelled and cleaned up in the background (response body + CTS disposed, budget released).
- **Threshold** is SDK-derived and **not customer-configurable**: `HttpTimeoutPolicyControlPlaneRetriableHotPath.FirstAttemptTimeout` (1 s) + `500 ms` = **1.5 s**.
- **Cross-region dedup**, **shared failure classification** (`MetadataRegionalFailureClassifier`), per-branch `DocumentServiceRequest.Clone()`, and the auth-reject (401/403) guard are unchanged from prior revisions.

## Not bombarding the Gateway (amplification safeguards)

Broadening to refresh reads does **not** increase steady-state secondary-region load unboundedly. A hedge is only ever dispatched for a read that is *both* eligible *and* slow, bounded by:

- **Latency threshold (1.5 s)** — a healthy primary never hedges.
- **One hedge per logical operation** (`HasHedgedThisOperation` latch).
- **Per-client concurrency budget (default 8)** — acquired with a non-blocking `Wait(TimeSpan.Zero)`; when exhausted the read falls back to primary-only (`BudgetExhausted`).
- **First page only** for `PartitionKeyRange` ReadFeed; subsequent pages pin to the winner.
- **Cross-region dedup** caps total attempts at the preferred-region count.
- **Single-region / excluded-region** reads never hedge.

## Eligibility gate

`EvaluateEligibility` falls back to primary-only for: opt-in disabled / Gateway kill-switch on / PPAF disabled (when opt-in `null`); already hedged this operation; unsupported resource type; `PartitionKeyRange` read that is not the first page; single-region account (or excluded regions leave no target); budget exhausted. **Cold start is no longer a gate.**

## Validation — tests added for the non-cold-start path

- `MetadataHedgingStrategyTests`: `EvaluateEligibility_NotColdStart_StillEligible`, `…_PartitionKeyRangeFirstPage_StillEligible`, `…_NotColdStart_UnsupportedResource_StillSkipsByType`, `ExecuteAsync_WarmRead_PrimarySlowHedgeFast_HedgeWins`, `ExecuteAsync_WarmRead_PrimaryWinsBeforeThreshold_DoesNotDispatchHedge`.
- `ClientCollectionCacheTests`: `…_WarmRead_PrimarySlow_HedgeFires_TwoSends`, `…_WarmRead_FastPrimary_SingleSend`.
- `PartitionKeyRangeCacheTests`: `WithStrategy_WarmRefresh_PrimarySlow_PageOneHedges`, `WithStrategy_WarmRefresh_FastPrimary_SingleSendPerPage`.
- `MetadataHedgingTelemetryTests`: the skipped-event/diagnostics test re-pointed to the still-valid `ResourceTypeNotSupported` skip reason.

Full metadata-hedging unit slice: **89/89 passing** (net6.0).

## Validation — real fault-injection on a live 5-region account (PR vs main)

Validated with the SDK's own **Fault Injection** framework so the slow hub is injected at the **real transport layer** and **real SDK operations** are measured end-to-end — not a `Task.Delay` simulation. A region-scoped `ResponseDelay` is applied to the **hub (West US 2)** Gateway metadata ops (`MetadataContainer` + `MetadataPartitionKeyRange`); secondary regions stay healthy so a hedge can win. The harness drives cold-start `ReadItemAsync` (populates the Collection + PkRange caches) and forced `PartitionKeyRangeCache.TryGetOverlappingRangesAsync(forceRefresh:true)` reads, so the strategy is exercised through its **real call sites** in `ClientCollectionCache` / `PartitionKeyRangeCache`. A/B: **PR (ON)** `AZURE_COSMOS_METADATA_HEDGING_ENABLED=true` vs **main (OFF)** `=false`, same injection + workload. Hedge counts are cross-checked against the SDK meter `Azure.Cosmos.Client.MetadataHedging`. Harness (test-only, no `src` changes, no changelog) + raw data + `realtime` (8 s, no compression) repro: [`docs/metadata-hedging-stress/faultinjection`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/nalutripician/metadata-hedging-stress-test/docs/metadata-hedging-stress/faultinjection).

Hub `ResponseDelay` = 3 s (compression of prod 5–10 s); the metadata HTTP timeout ladder is 1 s → 5 s → 65 s, so OFF lands at ~4 s while ON hedges at the 1.5 s threshold.

| Measurement point | main p50 | PR p50 | Δ p50 | meter fires PR / main |
| --- | ---: | ---: | ---: | ---: |
| Cold start (end-to-end `ReadItemAsync`) | 12,463 ms | 3,693 ms | **−70 %** | 16 / 0 |
| Refresh — low contention | 4,033 ms | 1,592 ms | **−61 %** | 30 / 0 |
| Refresh — saturating storm | 4,034 ms | 1,604 ms | **−60 %** | 32 / 0 |
| Mixed — refresh subset | 3,921 ms | 1,859 ms | **−53 %** | 1 / 0 |
| Mixed — **end-to-end** (70 % warm reads) | 48 ms | 34 ms | **~0** | 1 / 0 |

**Real fault-injected latency, PR vs main:**
latency

**Saturating storm CDF — the per-client budget bound, observed: ~67 % recover at ~1.6 s (hedged); the budget-exhausted ~⅓ (16/48) fall back to primary-only and match main at ~4 s:**
cdf

**Honest end-to-end framing — mixed workload p50 is unchanged (warm reads dominate); only the metadata-refresh tail (p95/p99) improves:**
mixed

**Conclusions:**
1. **Behaviour changes correctly** — hedges fire only with the PR (meter-confirmed 16 / 30 / 32); main fires **zero**.
2. **Real latency win on metadata reads** — cold start **−70 %** (12.5 s → 3.7 s) and refresh **−60 %** (4.0 s → 1.6 s) in real fault-injected wall-clock.
3. **Honest end-to-end** — for a warm-read-dominated mixed workload the **end-to-end p50 is unchanged**; hedging recovers the **metadata-refresh tail**, not every op (the refresh-subset win is **−53 % to −61 %**).
4. **Budget bound holds under load** — under a saturating storm of distinct containers, exactly the budget's worth hedge (32) and the remainder (**16 budget-exhausted**) fall back to primary-only, so the gateway is not flooded.
5. **Region fan-out** is a single secondary (East US); the other three regions are untouched.

A genuinely degraded-Azure-region drill (no synthetic delay) remains a pre-GA step (design §12). Two earlier simulated harnesses (`Task.Delay`-based) are kept under `docs/metadata-hedging-stress` but marked **superseded** by this fault-injection validation.

## Type of change

New feature (non-breaking). No public API surface in Phase 1; observable only through diagnostics/telemetry.


Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -857,67 +857,86 @@ public Func<HttpClient> HttpClientFactory
...
+        /// request for the first-time population of the Collection and PartitionKeyRange
+        /// metadata caches if the primary region's response has not arrived within the
+        /// SDK-derived default threshold — even when PPAF is disabled. When set to
+        /// <c>false</c>, metadata hedging is suppressed regardless of the PPAF state.
+        /// The Gateway-controlled <c>disableCrossRegionalHedging</c> account flag, when
+        /// set, takes precedence over this property. The hedging threshold and all other
+        /// tuning knobs are SDK-derived defaults and are not customer-configurable.
+        /// See <c>docs/PPAF_Metadata_Hedging_ColdStart_Design.md</c> §5.1.
+        /// </remarks>
+        internal bool? EnableMetadataHedgingForColdStart { get; set; }
```
NaluTripician: (A) This is the only real change in this file — but the diff shows +252/-233 because the whole file got re-saved with flipped line endings (CRLF↔LF). Every other removed/added pair is byte-identical. Suggest `git checkout origin/main -- CosmosClientOptions.cs` and re-applying just this property so the file shows ~14 real lines instead of ~470 phantom ones.

copilot_plan.md
```diff
@@ -0,0 +1,111 @@
+# Metadata Hedging — Staged Implementation Plan
```
NaluTripician: (A) This looks like an authoring/planning artifact rather than a deliverable, and it's already stale — it describes a public `CosmosClientOptions.MetadataHedgingOptions` property and `ConnectionPolicy` plumbing (Stage 1) that didn't ship. Recommend removing it from the PR.

Microsoft.Azure.Cosmos/src/MetadataHedgingOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <c>HttpTimeoutPolicy.FirstAttemptTimeout + 500&#160;ms</c> (today
+        /// 1.5&#160;s).
+        /// </summary>
+        public TimeSpan? Threshold { get; set; }
+
+        /// <summary>
+        /// Reserved for future use when <see cref="MaxHedgeBranchesPerAttempt"/>
+        /// &gt; 1. Default: 500&#160;ms.
+        /// </summary>
+        public TimeSpan? ThresholdStep { get; set; }
```
NaluTripician: (B) `ThresholdStep` (instance) is never read — only the static `DefaultThresholdStep` is used in the threshold formula. More broadly this whole class is speculative for Phase 1: it's `internal`, the production path always calls `CreateIfEnabled(..., options: null, ...)`, and it's only set in tests. Consider trimming to the consumed fields (`Threshold`, `PerClientConcurrencyBudget`) or dropping the class.

Microsoft.Azure.Cosmos/src/MetadataHedgingOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// Maximum simultaneous hedge branches per attempt. Default: 1.
+        /// Values &gt; 1 are reserved for a future release.
+        /// </summary>
+        public int MaxHedgeBranchesPerAttempt { get; set; } = 1;
```
NaluTripician: (B) `MaxHedgeBranchesPerAttempt` is never read anywhere (`Reserved for future use`). Recommend removing until the multi-branch path actually exists.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingContext.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    internal sealed class MetadataHedgingContext
+    {
+        private Uri winningEndpoint;
+        private string winningRegion;
+        private int hasHedgedThisOperation;
+
+        public bool IsColdStart { get; set; }
+
+        public ResourceType ResourceType { get; set; }
```
NaluTripician: (B) `ResourceType` is assigned by both caches but never read in `src` — the strategy derives resource type from `request.ResourceType`. Dead field; remove it + the two assignments.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingContext.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public bool IsFirstReadFeedPage { get; set; } = true;
+
+        public ConcurrentDictionary<string, byte> AttemptedEndpoints { get; }
+            = new ConcurrentDictionary<string, byte>();
+
+        public Uri WinningEndpoint => Volatile.Read(ref this.winningEndpoint);
+
+        public string WinningRegion => Volatile.Read(ref this.winningRegion);
```
NaluTripician: (B) `WinningRegion` / `winningRegion` / the `region` arg on `RecordWinner` are never read (only `WinningEndpoint` is consumed; region attribution for diagnostics lives on `MetadataHedgeDiagnostics`). Simplify `RecordWinner` to publish only the endpoint.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,608 @@
...
+            this.disposed = true;
+            this.hedgeBudget.Dispose();
+        }
+
+        /// <summary>
+        /// Per-branch acceptable-winner predicate composed over
+        /// <see cref="RetryUtility.IsRegionalFailure"/>. A 401 / plain 403
+        /// from the hedge branch is rejected — see design §5.13.
+        /// </summary>
+        internal static bool IsAcceptableWinner(DocumentServiceResponse response, HedgeBranch branch)
```
NaluTripician, line 513: (B) Minor: the two `if (response.StatusCode == Unauthorized)` / `Forbidden` blocks under `branch == HedgeBranch.Hedge` can collapse into one `if (branch == Hedge && (401 || 403)) return false;`.

NaluTripician (commented):
## Review focus: conciseness & unneeded changes

Thanks for this — the core hedging design is coherent and carefully written (CTS ownership transfer, volatile diagnostics, background loser cleanup, net472 stack discipline). My feedback below is concentrated on **trimming the diff** (the explicit ask) plus one real concurrency concern. Grouped as (A) unneeded/unrelated changes, (B) conciseness, (C) correctness. Inline comments mark the specific spots.

### (A) Unneeded / unrelated changes
- **`CosmosClientOptions.cs` (+252/-233) is ~all line-ending (CRLF↔LF) churn.** Only ~14 lines are real (the new property). Recommend `git checkout origin/main -- CosmosClientOptions.cs` then re-apply just `EnableMetadataHedgingForColdStart`. Removes ~470 phantom lines and keeps `git blame` intact.
- **`copilot_plan.md` (+111) is an authoring/planning artifact** and is already stale (it describes a public `CosmosClientOptions.MetadataHedgingOptions` + `ConnectionPolicy` plumbing that never shipped). Recommend dropping it from the PR.
- **Whitespace-only churn** in `PartitionKeyRangeCache.cs` and `DocumentClient.cs` on unrelated members (e.g. `endpointManager` field/param/assignment, `UseMultipleWriteLocations`). Revert so only hedging additions remain.
- **`docs/PPAF_Metadata_Hedging_ColdStart_Design.md` (+1379)** — optional: consider landing the design doc as a separate docs PR so this one stays code-focused.

### (B) Conciseness
- **`MetadataHedgingOptions` is speculative for Phase 1** — it's `internal`, always constructed with `options: null` in the production path, and only ever set in tests. `ThresholdStep` and `MaxHedgeBranchesPerAttempt` are dead even internally. Trim to the two consumed fields, or drop the class and pass threshold+budget as params.
- **`MetadataHedgingContext.ResourceType` and `WinningRegion`** are set but never read in `src` (resource type comes from `request.ResourceType`; region attribution lives on `MetadataHedgeDiagnostics`). Remove them + the `region` arg on `RecordWinner`.
- Minor: collapse the two identical `if` blocks in `IsAcceptableWinner`; add `using System.Collections.Generic;` to drop the repeated fully-qualified `KeyValuePair` in `MetadataHedgingMeter.cs`.

### (C) Correctness (one worth fixing)
- **Primary and hedge share one `DocumentServiceRequest`.** The caller's `sendToEndpoint` delegate calls `RouteToLocation` on the *same* request from both concurrent branches. The proven data-plane path (`CrossRegionHedgingAvailabilityStrategy`) clones the request per branch instead. If the primary re-resolves mid-flight (transient 410/redirect) after the hedge re-routes, both branches can hit the same region — defeating hedging and corrupting region telemetry. Recommend cloning per branch + a test asserting the two sends hit distinct endpoints.
- Minor: PK-range continuation pages (2..N) still enter `ExecuteAsync` and re-route redundantly; the loser-diagnostics write races the trace publish (memory-safe via volatile, but `LoserOutcome` may read null); and the changelog markets a feature whose opt-in is `internal` with no public API in Phase 1.

Nothing here is blocking the design — mostly diff hygiene. Happy to draft the exact reverts for (A) if useful.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/MetadataHedgingMeter.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private static readonly Histogram<double> HedgeFiredElapsedHistogram = Meter.CreateHistogram<double>(
+            name: CosmosDbClientMetrics.MetadataHedgingMetrics.Name.HedgeFiredElapsed,
+            unit: CosmosDbClientMetrics.MetadataHedgingMetrics.Unit.Sec,
+            description: CosmosDbClientMetrics.MetadataHedgingMetrics.Description.HedgeFiredElapsed);
+
+        internal static void RecordFire(string primaryRegion, string hedgeRegion, double elapsedMs)
+        {
+            FiresCounter.Add(1,
+                new System.Collections.Generic.KeyValuePair<string, object>("primary_region", primaryRegion ?? "unknown"),
```
NaluTripician: (B) `System.Collections.Generic.KeyValuePair<...>` is fully qualified ~10 times here. Add `using System.Collections.Generic;` (or a small `Tag(name,value)` helper) to cut the noise.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,608 @@
...
+                Stopwatch sw = Stopwatch.StartNew();
+                Task<DocumentServiceResponse> primaryTask = SendOneAsync(sendToEndpoint, request, primaryEndpoint, primaryCts.Token);
+                Task hedgeTimer = Task.Delay(this.threshold, timerCts.Token);
+
+                Task firstCompleted = await Task.WhenAny(primaryTask, hedgeTimer);
+
+                // ---- 5a. Primary genuinely won before threshold ----
+                if (firstCompleted == primaryTask
+                    && primaryTask.Status == TaskStatus.RanToCompletion
+                    && IsAcceptableWinner(primaryTask.Result, HedgeBranch.Primary))
```
NaluTripician, line 321: (C) Concurrency: primary and hedge share a single `DocumentServiceRequest` — the caller's `sendToEndpoint` delegate calls `RouteToLocation` on the same request from both branches. `CrossRegionHedgingAvailabilityStrategy` deliberately clones per branch instead. If the primary re-resolves after the hedge re-routes, both can target the same region (no hedge benefit + wrong region telemetry). Suggest cloning the request per branch and adding a test that races a slow primary vs hedge and asserts distinct endpoints.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
```diff
@@ -323,6 +368,26 @@ private async Task<DocumentServiceResponse> ExecutePartitionKeyRangeReadChangeFe
                     {
                         try
                         {
+                            if (this.metadataHedgingStrategy != null && hedgeContext != null)
+                            {
+                                MetadataHedgingResult hedgeResult = await this.metadataHedgingStrategy.ExecuteAsync(
```
NaluTripician: (C) Minor: on continuation pages (2..N) the request is already pinned to the winner above, but this still calls `ExecuteAsync`, which re-evaluates eligibility, hits `Skip(NotFirstReadFeedPage)`, and re-routes via `PrimaryOnlyAsync`. Consider gating on `hedgeContext.IsFirstReadFeedPage` and calling `storeModel.ProcessMessageAsync(request)` directly otherwise.

changelog.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [5838](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5838) EmbeddingGenerator: Adds ICosmosEmbeddingGenerator client-wide configuration (preview)
 - [5911](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5911) DistributedTransaction: Adds `DistributedReadTransaction` and `DistributedWriteTransaction` APIs (with `CosmosClient.CreateDistributedReadTransaction` / `CreateDistributedWriteTransaction`) for atomic read and write operations across partitions and containers (preview)
+- [5917](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5917) Metadata Hedging: Adds cold-start cross-region hedging for the Collection and PartitionKeyRange metadata caches in PPAF-enabled multi-region accounts. Reduces first-operation latency tails when the primary region is slow during cache population. Off by default in Phase 1; tracked via the new `Azure.Cosmos.Client.MetadataHedging` Meter and the `Metadata Hedge Context` diagnostics block. See [docs/PPAF_Metadata_Hedging_ColdStart_Design.md](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/main/docs/PPAF_Metadata_Hedging_ColdStart_Design.md).
```
NaluTripician: (C) Minor: this reads as customer-facing, but `EnableMetadataHedgingForColdStart` is `internal` — there's no public API to enable it in Phase 1. Consider wording it as internal/preview, or confirm the Meter+diagnostics are the intended observable surface.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
NaluTripician: Done in 0a8650c2 — reset the file to `origin/main` and re-applied only `EnableMetadataHedgingForColdStart`, so the diff is ~23 lines (the property plus `UseLengthAwareRangeComparer` going back to main's version). All ~470 phantom CRLF lines are gone, and the file now has zero line-ending churn.

copilot_plan.md
NaluTripician: Removed in 0a8650c2 — dropped `copilot_plan.md`, and also fixed the now-stale "introduced unwired by Stage 1/3/4" remark in `MetadataHedgingStrategy` that referenced it.

Microsoft.Azure.Cosmos/src/MetadataHedgingOptions.cs
NaluTripician: Removed the unused instance `ThresholdStep` in 0a8650c2. Kept `Threshold`, `PerClientConcurrencyBudget`, and the static `DefaultThresholdStep` (still used to derive the default hedge threshold in `CreateIfEnabled`).

Microsoft.Azure.Cosmos/src/MetadataHedgingOptions.cs
NaluTripician: Removed `MaxHedgeBranchesPerAttempt` in 0a8650c2 — it had no readers.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingContext.cs
NaluTripician: Removed the dead `ResourceType` field and both cache-side assignments in 0a8650c2; the strategy derives resource type from `request.ResourceType`.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingContext.cs
NaluTripician: Simplified `RecordWinner` to publish only the endpoint and dropped `WinningRegion`/`winningRegion` + the `region` arg in 0a8650c2 (region attribution stays on `MetadataHedgeDiagnostics`).

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
NaluTripician: Collapsed the two blocks into a single `if (branch == Hedge && (401 || 403))` condition in 0a8650c2.

Microsoft.Azure.Cosmos/src/Telemetry/OpenTelemetry/MetadataHedgingMeter.cs
NaluTripician: Added `using System.Collections.Generic;` and de-qualified the repeated `KeyValuePair` usages in 0a8650c2.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
NaluTripician: Fixed in 0a8650c2 — `SendOneAsync` now clones the `DocumentServiceRequest` per branch, so the concurrent primary and hedge sends never share routing state (matching the data-plane `CrossRegionHedgingAvailabilityStrategy` pattern). Verified `Clone()` yields an independent `RequestContext` and copies headers. Added `ExecuteAsync_PrimaryAndHedge_RouteIndependentClonesToDistinctEndpoints`, which races a slow primary against a fast hedge and asserts the two branches route distinct clones to distinct endpoints (it fails against the old shared-request code).

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
NaluTripician: Done in 0a8650c2 — gated the hedge call on `hedgeContext.IsFirstReadFeedPage`; continuation pages (2..N) now call `storeModel.ProcessMessageAsync(request)` directly (the request is already pinned to the winning endpoint), skipping the redundant eligibility/re-route round-trip.

changelog.md
NaluTripician: Reworded in 0a8650c2 — it now states the opt-in is `internal` (no public configuration API in Phase 1), that behavior follows the account's PPAF state, and that the observable surface is the new `Azure.Cosmos.Client.MetadataHedging` Meter + the `Metadata Hedge Context` diagnostics block. Also dropped the inaccurate "Off by default".

NaluTripician:
Addressed all of the review feedback in `0a8650c2`. Quick rundown against the grouping:

**(A) Unneeded / unrelated changes**
- `CosmosClientOptions.cs`: reset to main + re-applied only the new property. The whole-file CRLF flip is gone — the file diff dropped from ~485 to ~23 lines.
- `copilot_plan.md`: removed.
- **Line-ending churn**: fully eliminated across the PR now — `CosmosClientOptions.cs`, `DocumentClient.cs`, and `PartitionKeyRangeCache.cs` show **zero** carriage-return churn (verified with `--ignore-cr-at-eol`). Only the genuine hedging changes remain on those files.
- Design doc: I kept `PPAF_Metadata_Hedging_ColdStart_Design.md` in this PR for now, since ~10 code comments reference it by path/section — splitting it out would leave those references dangling until a follow-up docs PR merges. Happy to split it (doc + ref updates) as a fast-follow if you'd prefer a code-only PR.

**(B) Conciseness**: trimmed `MetadataHedgingOptions` to the consumed fields; removed `MetadataHedgingContext.ResourceType`/`WinningRegion` and simplified `RecordWinner`; collapsed the `IsAcceptableWinner` 401/403 blocks; added the `using` for `KeyValuePair`.

**(C) Correctness**: the shared-`DocumentServiceRequest` race is fixed — primary and hedge now each send an independent `request.Clone()`, with a new race test asserting distinct clones route to distinct endpoints. Continuation pages skip the hedge orchestration, the loser-diagnostics null window is closed, and the changelog now reflects the internal/observable-only surface.

Build is clean and all 72 hedging unit tests pass.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,871 @@
...
+                diag.Eligible = false;
+                diag.SkipReason = MetadataHedgeSkipReason.BudgetExhausted;
+                MetadataHedgingMeter.RecordBudgetExhausted(diag.ResourceType);
+                CosmosDbEventSource.MetadataHedgeSkipped(diag.SkipReason.ToString(), diag.ResourceType);
+                return await this.PrimaryOnlyAsync(request, primaryEndpoint, sendToEndpoint, hedgeContext, diag, trace, cancellationToken);
+            }
+
+            // ---- 3. Resolve hedge endpoint ----
+            ReadOnlyCollection<Uri> applicable = this.globalEndpointManager.GetApplicableEndpoints(request, isReadRequest: true);
+            Uri hedgeEndpoint = applicable.FirstOrDefault(u => u != null && !u.Equals(primaryEndpoint));
```
NaluTripician: The hedge budget permit is taken at line 262 (`hedgeBudget.Wait`), but this endpoint resolution runs before the `try`/`finally` (which starts at line 291). Two problems on this path:

1. `applicable.FirstOrDefault(...)` has no null-guard on `applicable` itself. If `GetApplicableEndpoints` returns null during a concurrent location-cache refresh/failover, this throws `NullReferenceException` (the `hedgeEndpoint == null` check below doesn't help).
2. If `GetApplicableEndpoints` throws (or the NRE above fires), the exception escapes before the `try`, so the `finally` that releases the permit never runs and a budget permit leaks permanently.

Repeated occurrences (exactly the cold-start-during-failover window where hedging matters most) drive `AvailableBudget` to 0, after which every future hedge on this client silently skips with `BudgetExhausted` for the lifetime of the `CosmosClient` â€” no error surfaced.

Suggest moving the `Wait` to the first statement inside the `try` (initialize `budgetReleased=false` first), and adding `if (applicable == null || hedgeEndpoint == null) { Release; return PrimaryOnly... }`. A regression test that makes `GetApplicableEndpoints` throw/return null after eligibility passes and asserts `AvailableBudget == PerClientConcurrencyBudget` would lock this down.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,871 @@
...
+                    if (branch == HedgeBranch.Hedge
+                        && finished.Status == TaskStatus.RanToCompletion
+                        && finished.Result != null
+                        && (finished.Result.StatusCode == HttpStatusCode.Unauthorized
+                            || finished.Result.StatusCode == HttpStatusCode.Forbidden))
+                    {
+                        diag.HedgeOutcome = finished.Result.StatusCode == HttpStatusCode.Unauthorized
+                            ? "Auth401"
+                            : "Auth403";
+                        MetadataHedgingMeter.RecordHedgeAuthReject(diag.HedgeRegion, (int)finished.Result.StatusCode);
```
NaluTripician: This auth-reject telemetry only records when the hedge task is `RanToCompletion` with a non-null `Result` (lines 363-365). For `GatewayStoreModel`, 401/403 are typically *thrown* as `DocumentClientException`, producing a faulted task with no `Result` â€” so the hedge is still correctly rejected, but `HedgeOutcome`, `RecordHedgeAuthReject` and `MetadataHedgeAuthReject` never fire and we lose the cross-region auth signal this is meant to capture. Consider also classifying the auth reject from a faulted task's `DocumentClientException.StatusCode`.

Microsoft.Azure.Cosmos/src/MetadataRequestThrottleRetryPolicy.cs
```diff
@@ -149,10 +159,11 @@ private Task<ShouldRetryResult> ShouldRetryInternalAsync(
                 return this.throttlingRetryPolicy.ShouldRetryAsync(exception, cancellationToken);
             }
 
-            if (statusCode == HttpStatusCode.ServiceUnavailable 
-                || statusCode == HttpStatusCode.InternalServerError
-                || (statusCode == HttpStatusCode.Gone && subStatus == SubStatusCodes.LeaseNotFound)
-                || (statusCode == HttpStatusCode.Forbidden && subStatus == SubStatusCodes.DatabaseAccountNotFound))
+            if (MetadataHedgingStrategy.IsRegionalFailure(
```
NaluTripician: Layering concern: this retry policy now depends on `MetadataHedgingStrategy` for its core failure classification (and holds a `MetadataHedgingStrategy.MetadataHedgingContext` field at line 69). The dependency should point hedging â†’ retry, but here it points retry â†’ hedging, so this lower-level retry component can no longer be reasoned about or unit-tested independently of hedging. The failure-class list is legitimately shared, but its home shouldn't be the hedging type â€” consider extracting `IsRegionalFailure` into a neutral static (e.g. `MetadataRegionalFailureClassifier`) that both the strategy and the retry policy depend on.

Microsoft.Azure.Cosmos/src/Routing/ClientCollectionCache.cs
```diff
@@ -230,27 +277,42 @@ private async Task<ContainerProperties> ReadCollectionAsync(
...
                         retryPolicyInstance?.OnBeforeSendRequest(request);
 
+                        if (this.metadataHedgingStrategy != null)
+                        {
+                            MetadataHedgingStrategy.MetadataHedgingContext hedgeContext = new MetadataHedgingStrategy.MetadataHedgingContext
+                            {
+                                IsColdStart = isColdStart,
+                            };
+
+                            (retryPolicyInstance as MetadataRequestThrottleRetryPolicy)?.AttachHedgeContext(hedgeContext);
```
NaluTripician: This `as`-cast side channel makes the dedup contract implicit and fragile. The same `MetadataHedgingContext` is written by `ExecuteAsync` (`AttemptedEndpoints` / `WinningEndpoint`) and read by the retry policy via `AttemptedEndpoints`. If the retry policy is ever wrapped or replaced, the `as` returns null and the dedup silently degrades to a no-op â€” the only symptom being an extra metadata retry into a region the hedge already burned. Suggest a narrow read-only interface (e.g. `IAttemptedEndpoints`) injected through the typed constructor instead of the cast, with field ownership/read-safety documented.

Microsoft.Azure.Cosmos/src/Routing/CollectionCache.cs
```diff
@@ -263,7 +286,7 @@ private Task<ContainerProperties> ResolveByRidAsync(
                 async () =>
                 {
                     DateTime currentTime = DateTime.UtcNow;
-                    ContainerProperties collection = await this.GetByRidAsync(apiVersion, collectionResourceId, trace, clientSideRequestStatistics, cancellationToken);
+                    ContainerProperties collection = await this.GetByRidAsync(apiVersion, collectionResourceId, trace, clientSideRequestStatistics, isColdStart: true, cancellationToken);
```
NaluTripician: `isColdStart` ends up meaning three different things across call sites. Here in `ResolveByRidAsync` it's hard-coded `true` â€” so even a *forced* refresh-by-RID is treated as cold-start and will hedge, which looks unintended. In `ResolveByNameAsync` (line 321) it's `!forceRefesh`, the request-based path passes `false`, and `PartitionKeyRangeCache` uses `previousRoutingMap == null`. Hedge eligibility keys off a flag whose meaning drifts by path. Suggest naming the concept precisely (e.g. `isFirstPopulation`), computing it once, and justifying (or fixing) this unconditional `true`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Routing/MetadataHedgingStrategyTests.cs
```diff
@@ -0,0 +1,841 @@
...
+                    return Task.FromResult(NewOkResponse());
+                },
+                ctx,
+                NoOpTrace.Singleton,
+                CancellationToken.None);
+
+            Assert.AreEqual(1, callCount);
+            Assert.IsFalse(result.HedgeFired);
+            Assert.AreEqual(PrimaryEndpoint, result.WinningEndpoint);
+            Assert.AreEqual(strategy.PerClientConcurrencyBudget, strategy.AvailableBudget);
```
NaluTripician, line 275: Coverage gap on the highest-risk fault paths. This is the only budget assertion and it's a happy path â€” there's no test asserting `AvailableBudget` returns to full after a send / endpoint-resolution fault (that's exactly what would catch a permit leak on the `ExecuteAsync` resolve path). Also missing: both-branches-regional-failure fallback (prefer the primary outcome, `TotalAttempts == 2`), and a faulted-task winner flowing through `ObserveWinningTaskAsync` (every test here returns responses or cancels the loser; none throws). Adding these three would meaningfully harden the strategy.

NaluTripician (commented):
Local deep-review of the metadata-hedging change. One blocking item â€” a budget-permit leak plus a latent `NullReferenceException` on the endpoint-resolve path in `ExecuteAsync` â€” and a few design/test recommendations are left inline. Overall the orchestration (budget accounting, loser-CTS hand-off, single-winner latch, hedged-this-operation latch) is carefully done and the legacy no-hedge retry path is preserved. Line numbers refer to the current head (5150b02).

Microsoft.Azure.Cosmos/src/MetadataRequestThrottleRetryPolicy.cs
NaluTripician: Fixed in 043d742c8 — extracted the failure-class list into a neutral `MetadataRegionalFailureClassifier.IsRegionalFailure` (its own file/type). Both `MetadataRequestThrottleRetryPolicy` and `MetadataHedgingStrategy.IsAcceptableWinner` now depend on the classifier, so the retry policy no longer points at the hedging type for its core classification. Repointed the existing `MetadataHedgingRegionalFailureTests` to the classifier.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Routing/MetadataHedgingStrategyTests.cs
NaluTripician: Fixed in 043d742c8 — added the three fault-path tests: `ExecuteAsync_HedgeEndpointResolutionFaults_ReleasesBudgetPermit` (budget returns to full after a resolution fault), `ExecuteAsync_BothBranchesRegionalFailure_PrefersPrimaryOutcome` (both 503 -> primary outcome preferred, `TotalAttempts == 2`, budget restored), and `ExecuteAsync_FaultedWinner_RethrowsThroughObserveWinningTaskAndReleasesBudget` (a faulted winner surfaces verbatim through `ObserveWinningTaskAsync` and the budget is released).

Microsoft.Azure.Cosmos/src/Routing/CollectionCache.cs
NaluTripician: Fixed in 043d742c8 — renamed the cache-layer flag from `isColdStart` to `isFirstPopulation` across the `CollectionCache` / `ClientCollectionCache` overloads and mapped it into `MetadataHedgingContext.IsColdStart` at the boundary. Each call site is now justified inline: `ResolveByRidAsync` passes `true` because the `AsyncCache` value factory only runs on a cache miss (there is no force-refresh path through it); `ResolveByNameAsync` passes `!forceRefesh`; `RefreshAsync` passes `false`.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
NaluTripician: Fixed in 043d742c8 — the hedge-endpoint resolution (`GetApplicableEndpoints` + `FirstOrDefault`) now runs *inside* the `try`, with `applicable?.FirstOrDefault(...)` null-guarding the result. Any throw (or null) now runs the `finally` and releases the permit instead of leaking it, so a concurrent location-cache refresh/failover can no longer drive `AvailableBudget` to 0 for the client lifetime. Added `ExecuteAsync_HedgeEndpointResolutionFaults_ReleasesBudgetPermit` (GetApplicableEndpoints throws on the resolution call after the permit is taken; asserts budget is fully restored) — it fails against the old pre-`try` code.

Microsoft.Azure.Cosmos/src/Routing/ClientCollectionCache.cs
NaluTripician: Fixed in 043d742c8 — replaced the concrete-type `as` cast with a narrow `IMetadataHedgeContextReceiver` seam. `MetadataRequestThrottleRetryPolicy` implements it and `ClearingSessionContainerClientRetryPolicy` forwards to its inner policy, so the dedup context reaches the consumer even when the policy is wrapped (the collection path wraps with `ClearingSessionContainerClientRetryPolicy`, where the old `as MetadataRequestThrottleRetryPolicy` was always null). The attach is now explicit rather than a silent type-guess.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
NaluTripician: Fixed in 043d742c8 — auth-reject classification now covers BOTH a completed task whose `DocumentServiceResponse` carries 401/403 AND a faulted task whose thrown `DocumentClientException` carries 401/403 (the GatewayStoreModel path). Extracted `TryGetHedgeAuthRejectStatus(response, faultException)` so `HedgeOutcome` / `RecordHedgeAuthReject` / `MetadataHedgeAuthReject` fire in both cases. Added unit coverage for the completed, faulted, and non-auth/null paths.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Contracts/DotNetPreviewSDKAPI.net6.json
```diff
@@ -238,6 +238,11 @@
           "Attributes": [],
           "MethodInfo": null
         },
+        "Microsoft.Azure.Cosmos.CosmosDbClientMetrics+MetadataHedgingMetrics": {
```
kundadebdatta: I wonder is this contract change needed @NaluTripician ?

kundadebdatta: Removed the contract since it isnot needed.

NaluTripician:
## 🔬 Worst-case stress test — metadata hedging under a degraded hub region

I ran a worst-case stress test of this PR against a **live 5-region account** to show the hedging code stays sound when the SDK goes from healthy → **hub (write) region gateway has a 5–10 s delay on all metadata calls** and reads return **PartitionKeyRange-Gone** (which spawns `PartitionKeyRange` ReadFeed *refresh* reads). Tracked: per-region call counts, low vs high partition counts (**100 / 10k / 50k**), and how many PkRange reads are spawned vs hedged.

> **No PR/src code was changed.** The harness only *adds* a test (`MetadataHedgingStressHarness.cs`) that drives the **real** `MetadataHedgingStrategy.ExecuteAsync` against a **real** `GlobalEndpointManager` from a live `CosmosClient`. Results reflect the current PR state. Full artifacts, CSVs, README, and the harness are on branch [`nalutripician/metadata-hedging-stress-test`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/nalutripician/metadata-hedging-stress-test/docs/metadata-hedging-stress).

### Methodology
- Live account `nalu-new`, 5 regions: **West US 2 (hub)** · East US · South Central US · Central US · North Central US. The harness resolves these through the real `GlobalEndpointManager` (real `ReadEndpoints` / `GetApplicableEndpoints` / `GetLocation`).
- The strategy's `sendToEndpoint` seam injects the degraded transport: **hub** = 5–10 s delay then `410 / PartitionKeyRangeGone`; **secondaries** = fast `200 OK`.
- Each PartitionKeyRange-Gone event = one `PartitionKeyRange` ReadFeed refresh read (`IsColdStart=false` — the steady-state path, worst case for amplification).
- Hedge counts cross-checked against the SDK's own meter `Azure.Cosmos.Client.MetadataHedging` via a `MeterListener` — **`meterFires` matched measured `hedgeFired` exactly in every scenario** (8 / 800 / 4,054 / 56).
- High-partition runs are **time-compressed 10×** (threshold 150 ms models prod 1.5 s; hub delay 0.5–1.0 s models 5–10 s) so a 50k-op storm finishes in ~90 s; one extra run uses **true production timing** (1.5 s / 5–10 s).

### Results

| Scenario | PkRange reads spawned | hedged (= meter fires) | budget-exhausted | max concurrent secondary | p50 | p99 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Healthy baseline (hub fast) | 2,000 | **0** | 1,920 | **0** | 16 ms | 32 ms |
| Degraded, P=100 | 100 | 8 | 92 | **8** | 729 ms | 992 ms |
| Degraded, P=10,000 | 10,000 | 800 | 9,200 | **8** | 734 ms | 1,002 ms |
| Degraded, P=50,000 | 50,000 | 4,054 | 45,946 | **8** | 737 ms | 1,001 ms |
| **P=100, REAL timing (1.5 s / 5–10 s)** | 100 | 56 | 44 | **8** | **1.57 s** | 9.91 s |

**Per-region calls:** the hub takes every primary send; **all hedges go to exactly one region (the 2nd preferred, East US)** — the other three regions are never touched.

### Graphs

**The gateway is NOT bombarded — secondary fan-out stays capped at the budget (8) whether 100 or 50,000 partitions storm at once:**
secondary in-flight cap

**Per-region metadata calls (hedge concentrates on East US, others untouched):**
calls per region

**PkRange refreshes spawned vs hedged vs budget-exhausted:**
spawned vs hedged

**Production-timing tail recovery — median PkRange refresh recovered to 1.57 s vs 5–10 s hub:**
realtime latency win

**Healthy hub ⇒ zero hedges, zero secondary calls (no amplification):**
healthy no amplification

### Conclusions
1. **No amplification when healthy** — a healthy hub wins before the 1.5 s threshold ⇒ 0 hedges, 0 secondary calls.
2. **Gateway not bombarded, even at 50k partitions** — concurrent secondary fan-out is hard-bounded by the per-client budget (**max 8 in flight**) independent of storm size. Ops beyond the budget fall back to primary-only (`BudgetExhausted`) rather than exceeding it.
3. **Real tail recovery** — at production timing the median refresh recovered to **1.57 s** (hedge wins from East US) vs the 5–10 s degraded hub. The budget-exhausted remainder intentionally stays primary-only, so the p99 tail equals the hub delay **by design** (bounded amplification), not a defect.
4. **Deterministic single hedge target** — all hedges go to the 2nd preferred region; no fan-out across all 5.

The code behaves exactly as the design intends under the worst case: it recovers the metadata tail when the hub is slow, while structurally capping secondary-region load. 🚦

NaluTripician:
### 🔄 Updated the PR description's validation section — phased PR-vs-main scenarios

Replaced the old `## Benchmarking` section with **`## Validation — worst-case scenarios on a live 5-region account (PR vs main)`**. It runs the **same workload** against the **real** `MetadataHedgingStrategy` + **real** `GlobalEndpointManager` of a live 5-region account, comparing **PR (metadata hedging)** vs **main (metadata reads primary-only)** across the four requested timelines:

1. **S1** — completely healthy → PR == main, **0 hedges / 0 secondary calls**.
2. **S2** — bad cold start → PR cuts cold-start metadata p50 **~28 %** (thundering-herd; budget caps how many hedge).
3. **S3** — bad cold start → OK → degraded (hub delay + `410 PkRange-Gone`), with normal hedging on both arms → PR cuts refresh-bearing-read p50 **~65 %**.
4. **S4** — same as S3 but healthy cold start → same **~65 %** degraded-phase win.

The S3 timeline graph makes it visual: on **main**, every read that hits `410 PkRange-Gone` stalls on the slow hub's metadata refresh even *with* data-plane normal hedging (the routing map must refresh first); **PR** recovers most of them via a single bounded hedge to East US. The p99 tail that remains is the budget-exhausted fallback — **by design**, so the gateway is never flooded.

Harness (test-only, no src changes) + raw CSVs + graphs: [`docs/metadata-hedging-stress/scenarios`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/nalutripician/metadata-hedging-stress-test/docs/metadata-hedging-stress/scenarios).

NaluTripician:
## ✅ Re-based the validation on REAL Fault Injection (transport-layer, real SDK read path)

Per review feedback that the earlier harnesses faked latency with `Task.Delay` and called `MetadataHedgingStrategy.ExecuteAsync` directly, I added a new **real fault-injection harness** — `Microsoft.Azure.Cosmos/FaultInjection/tests/MetadataHedgingFaultInjectionHarness.cs` (test-only, gated `RUN_HEDGE_FI=1`, `[TestCategory("StressHarness")]`, **no src changes → no changelog entry**).

### What's real now
- **Transport-layer delay** via `Microsoft.Azure.Cosmos.FaultInjection`: a region-scoped `ResponseDelay` on the **hub (West US 2)** Gateway metadata ops (`MetadataContainer` + `MetadataPartitionKeyRange`); secondaries healthy so a hedge can win.
- **Real SDK read path** — cold-start `ReadItemAsync` (populates Collection + PkRange caches) and forced `PartitionKeyRangeCache.TryGetOverlappingRangesAsync(forceRefresh:true)` (real PkRange ReadFeed). The strategy is hit through its **real call sites**, not directly.
- **Real end-to-end latency**, live 5-region account, hedge counts **cross-checked against the SDK meter** `Azure.Cosmos.Client.MetadataHedging`.
- **A/B**: ON (`AZURE_COSMOS_METADATA_HEDGING_ENABLED=true`) vs OFF (`=false`, main's behaviour), same FI + workload.

### Results (hub `ResponseDelay` 3 s; metadata timeout ladder 1s→5s→65s ⇒ OFF lands ~4 s)

| Measurement point | main p50 | PR p50 | Δ p50 | meter fires PR/main |
| --- | ---: | ---: | ---: | ---: |
| Cold start (end-to-end `ReadItemAsync`) | 12,463 ms | 3,693 ms | **−70 %** | 16 / 0 |
| Refresh — low contention | 4,033 ms | 1,592 ms | **−61 %** | 30 / 0 |
| Refresh — saturating storm | 4,034 ms | 1,604 ms | **−60 %** | 32 / 0 |
| Mixed — refresh subset | 3,921 ms | 1,859 ms | **−53 %** | 1 / 0 |
| Mixed — **end-to-end** (70 % warm reads) | 48 ms | 34 ms | **~0** | 1 / 0 |

**Honest framing addressed:** the earlier "−65 % p50" was a refresh-subset number; end-to-end p50 is unchanged because warm reads dominate. Both are now reported explicitly, in low **and** saturating contention.

**Behaviour actually changes — hedges fire only with the PR (SDK meter):**
meter

**Real fault-injected latency, PR vs main:**
latency

**Budget bound is real & observed — saturating storm CDF: ~67 % recover at ~1.6 s (hedged); the budget-exhausted ~⅓ (16/48) fall back to primary-only and match main at ~4 s:**
cdf

**Honest end-to-end: p50 unchanged (warm reads dominate); only the metadata-refresh tail improves:**
mixed

### Takeaways
1. **Behaviour changes correctly** — ON hedges (meter 16/30/32), OFF = 0.
2. **Real latency win on metadata reads** — cold start −70 %, refresh −60 % in real wall-clock.
3. **End-to-end is honest** — hedging helps the metadata-refresh tail, not every op; warm-read-dominated p50 is flat.
4. **Budget bound holds under load** — exactly the budget's worth hedge; the rest fall back to primary-only, so the gateway is not flooded.
5. Single-secondary fan-out (East US).

**Still pending pre-GA (design §12):** a genuinely degraded-Azure-region drill (this uses a synthetic injected delay, not a real region outage). The old simulated harnesses are kept but marked **superseded**; full methodology, CSVs, and `realtime` (8 s, no compression) repro are in [`docs/metadata-hedging-stress/faultinjection`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/nalutripician/metadata-hedging-stress-test/docs/metadata-hedging-stress/faultinjection).

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,886 @@
...
+
+                hedgeContext.AttemptedEndpoints.TryAdd(primaryEndpoint.AbsoluteUri, 0);
+
+                Stopwatch sw = Stopwatch.StartNew();
+                Task<DocumentServiceResponse> primaryTask = SendOneAsync(sendToEndpoint, request, primaryEndpoint, primaryCts.Token);
+                Task hedgeTimer = Task.Delay(this.threshold, timerCts.Token);
+
+                Task firstCompleted = await Task.WhenAny(primaryTask, hedgeTimer);
+
+                // ---- 5a. Primary genuinely won before threshold ----
```
kundadebdatta, line 318: @NaluTripician : If the primary completes/faults before the threshold and isn't acceptable (fast 503/HttpRequestException), the hedge dispatches immediately, skipping the `1.5` s delay. 

Good for one op; the concern is correlated failure — during a primary brownout, every eligible metadata refresh fast-fails and instantly hedges to the secondary. The SemaphoreSlim(8) caps per-client concurrency, not rate and not fleet-wide, so up to ~8×N simultaneous hedges can hit the secondary metadata gateway exactly when it's already absorbing failover traffic. 

Worth a primary-brownout soak test before default-on; a conscious tradeoff, not necessarily a defect.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,886 @@
...
+            // Retained for wire/diagnostic compatibility. Hedging is no longer
+            // gated on cold start, so EvaluateEligibility never produces this
+            // value; refresh reads are eligible on the same terms as cold start.
+            NotColdStart,
+            ResourceTypeNotSupported,
+            NotFirstReadFeedPage,
+            BudgetExhausted,
+            AlreadyHedgedThisOperation,
+            ExcludedRegionLeavesNoTarget,
+            AuthModeNotEligibleForHedge,
```
kundadebdatta: Not sure if this is used anywhere. If not - let's please delete it.

NaluTripician:
### 📊 Added embedded analysis to every graph

Each graph now carries an inline **"What it shows / What it proves"** box, so reviewers can read the takeaway directly off the image (the embedded images above update automatically). Quick map of the authoritative fault-injection set:

- **fi1 — latency PR vs main:** *shows* real fault-injected p50/p99 at cold start, low/saturating refresh, and the mixed refresh subset. *Proves* hedging cuts real latency when the hub is slow — cold start **−70 %**, refresh **−60 %**; the only place PR p99 meets main is the saturating storm (budget-bounded fallback, not a regression).
- **fi2 — meter cross-check:** *shows* hedges fired from the SDK meter `Azure.Cosmos.Client.MetadataHedging`. *Proves* the wins are causal — hedges fire only with PR ON (16/30/32), main = 0.
- **fi3 — saturating-storm CDF:** *shows* the latency distribution when 12 distinct containers refresh concurrently (> budget 8). *Proves* the per-client budget is real — ~67 % hedge and recover at ~1.6 s; the budget-exhausted **16/48** fall back to primary-only at ~4 s, so the Gateway is never flooded.
- **fi4 — mixed end-to-end:** *shows* p50/p95/p99 of a 70 %-warm-read workload. *Proves* **end-to-end p50 is unchanged**; only the metadata-refresh tail improves — a targeted tail win, honestly framed.

The two superseded simulated graph sets (`scenarios/` s1–s5 and the worst-case `g1–g5`) were annotated the same way and labeled `[Simulated]`. Full per-graph analysis is also in each folder's README under `docs/metadata-hedging-stress`.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
NaluTripician: Done in b9931b525 — removed the unused MetadataHedgeSkipReason.AuthModeNotEligibleForHedge. It had no producer (EvaluateEligibility never set it) and no consumer anywhere in src or 	ests; src build is green after removal.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
NaluTripician: Agreed this is a conscious tradeoff rather than a defect — and you're right that the fast-fail fall-through dispatches immediately. I ran the primary-brownout soak you suggested with **real fault injection** to bound it.

**Setup:** inject a hub-region Gateway `ServiceUnavailable` (503, **no delay**) on `MetadataContainer` + `MetadataPartitionKeyRange`, secondaries healthy, then fire a wide simultaneous wave of distinct containers (> budget 8) so every eligible refresh fast-fails and instantly hedges — the exact line-319 path. Hedge counts are cross-checked against the `Azure.Cosmos.Client.MetadataHedging` meter.

**Result (48 ops × wide concurrent waves):**

| Arm | hedges fired | budget-exhausted → primary-only | p50 | p99 |
| --- | ---: | ---: | ---: | ---: |
| main (OFF) | 0 | 0 | 82 ms | 385 ms |
| PR (ON) | 32 | **33** | 82 ms | 325 ms |

fast-fail brownout budget

**Takeaway:** the per-client `SemaphoreSlim(8)` is acquired with `Wait(0)` *before* the primary send, so it gates **both** the threshold-elapsed and the fast-fail immediate-hedge paths identically. Even when every primary instantly 503s, only the budget's worth hedge and the rest fall back to primary-only (**33 budget-exhausted** here) — the fast-fail path does **not** bypass the budget, and per-client concurrent secondary hedges stay capped at 8.

On the fleet-wide `8×N` point: that's correct and inherent — the budget is per-client, so it can't coordinate across N clients. That (and the default-on decision) stays a pre-GA item per design §12; in Phase 1 the feature is internal/opt-in (follows PPAF), not default-on, so we're not flipping it on for everyone here.

Harness, raw CSVs, and the `realtime` (no-compression) repro are on the validation branch under [`docs/metadata-hedging-stress/faultinjection`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/nalutripician/metadata-hedging-stress-test/docs/metadata-hedging-stress/faultinjection). (Side note: the wide fast-fail wave also surfaced a small, pre-existing thread-safety race in the FaultInjection *test framework* — `FaultInjectionServerErrorResultInternal.IsApplicable` enumerates a per-rule history list while another thread appends — unrelated to the hedging code; the soak retries that transient. Happy to file it separately if useful.)

Microsoft.Azure.Cosmos/src/Resource/ClientContextCore.cs
```diff
@@ -89,7 +89,8 @@ internal static CosmosClientContext Create(
                cosmosClientTelemetryOptions: clientOptions.CosmosClientTelemetryOptions,
                chaosInterceptorFactory: clientOptions.ChaosInterceptorFactory,
                enableAsyncCacheExceptionNoSharing: clientOptions.EnableAsyncCacheExceptionNoSharing,
-               useLengthAwareRangeComparer: clientOptions.UseLengthAwareRangeComparer);
+               useLengthAwareRangeComparer: clientOptions.UseLengthAwareRangeComparer,
+               enableMetadataHedgingForColdStart: ConfigurationManager.GetMetadataHedgingOptIn());
```
kundadebdatta: May be we should change the variable name to `enableMetadataHedging` or something similar? Since metadata hedging is now applicable for both cold start and warm path.

NaluTripician: Good catch - renamed nableMetadataHedgingForColdStart to nableMetadataHedging across DocumentClient, ClientContextCore and MetadataHedgingStrategy (and the unit tests), since hedging now covers both first-population and refresh reads. Pushed in 8042633.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,884 @@
...
+        {
+            private Uri winningEndpoint;
+            private int hasHedgedThisOperation;
+
+            /// <summary>
+            /// Whether this read is the first-population (cold-start) read of the
+            /// cache. Recorded for diagnostics/telemetry only; it does NOT gate
+            /// hedging eligibility — refresh reads hedge on the same terms.
+            /// </summary>
+            public bool IsColdStart { get; set; }
```
kundadebdatta: Since metadata hedging is now enabled for both clod and warm path scenarios, do we really need this parameter ?

NaluTripician: Good catch — no, we don't need it. `IsColdStart` was being set by the callers but never actually read anywhere in production (it was only ever a "diagnostics" signal that nothing emitted), now that eligibility no longer depends on cold vs. warm. I've removed it in 67d5dfc61, along with the `isFirstPopulation` plumbing that existed solely to populate it:

- Dropped `MetadataHedgingContext.IsColdStart`.
- Reverted the `isFirstPopulation` overloads on `CollectionCache`/`ClientCollectionCache` (the `CollectionCache` diff is now empty vs. main) and the `ReadCollectionAsync` parameter; the actual hedge wiring is unchanged.
- Removed the set-site in `PartitionKeyRangeCache`.
- Updated tests and the design docs to match.

Behavior is identical (the field was never consumed), so no changelog entry is needed for this cleanup within the unreleased feature.

kundadebdatta:
This is closed with this PR: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5999

PR #5323, opened 2025-07-29
[Internal] Thin Client Integration: Removes ThinClientStoreModel and uses GatewayStoreModel for better code modularity.

aavasthy:
# Pull Request Template

## Description

It is observed that most of the working principle on the ThinClientStoreModel can be achieved by the GatewayStoreModel. Therefore, removing ThinClientStoreModel and accommodating the logic to navigate between thinclient mode and gateway mode in GatewayStoreModel. Added tests to verify correct working of all three connection modes: thinclient, gateway and direct.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## Closing issues

To automatically close an issue: closes #5311

PR #5927, opened 2026-06-05
[Internal] Thin Client Integration: Fixes location cache logic to switch from gateway v2 to v1 and vice-versa midway.

aavasthy:
# Pull Request Template

## Description

 Previously, once a CosmosClient instance selected thin-client mode at initialization, that routing decision was pinned for the lifetime of the client. While LocationCache continued to refresh the thin-client endpoint list on every account refresh, the per-request dispatch sites consulted only the initialization-time feature flag and continued to route traffic to the thin-client port (:10250) regardless of whether the service was still advertising any thin-client locations. As a result, any server-side change — partial or full — left clients sending requests to stale thin-client endpoints until the process was restarted. This PR introduces two per-direction availability flags (HasThinClientReadLocations and HasThinClientWriteLocations) that are recomputed on every account refresh from the live AccountProperties snapshot and evaluated at every per-request dispatch site. The SDK now falls back to the standard gateway path on the next request when the service stops advertising thin-client endpoints, and seamlessly re-engages thin-client mode when they become available again — without requiring a client restart, CosmosClient recreation, or any customer code change. Read and write directions are tracked independently so the SDK also handles the case where only one direction is withdrawn.

Diagnostic logs showing switch from thinclient mode to gateway mode:
**Thinclient flag enabled**
[2026-06-04T22:07:55.6275483Z] elapsed=00:14:34 success=825 failures=0 status=OK diagnostics={"Summary":{"GatewayCalls":{"(200, 0)":1}},"name":"ReadItemAsync","start datetime":"2026-06-04T22:07:55.578Z","duration in milliseconds":48.6922,"data":{"Client Configuration":{"Client Created Time Utc":"2026-06-04T21:52:51.3268455Z","MachineId":"vmId:611ed772-5bcc-4eb3-90bc-d26d23956e13","VM Region":"westus2","NumberOfClientsCreated":1,"NumberOfActiveClients":1,"ConnectionMode":"Gateway","User Agent":"cosmos-netstandard-sdk/3.60.0|1|X64|Microsoft Windows 10.0.26200|.NET 6.0.36|N|F4","ConnectionConfig":{"gw":"(cps:50, urto:6, p:False, httpf: False)","rntbd":"(cto: 5, icto: -1, mrpc: 30, mcpe: 65535, erd: True, pr: ReuseUnicastPort)","other":"(ed:False, be:False)"},"ConsistencyConfig":"(consistency: NotSet, readStrategy: NotSet, prgns:[], apprgn: )","ProcessorCount":16}},"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.RequestInvokerHandler","duration in milliseconds":48.5313,"children":[{"name":"Get Collection Cache","duration in milliseconds":0.0009},{"name":"Microsoft.Azure.Cosmos.Handlers.DiagnosticsHandler","duration in milliseconds":48.4012,"data":{"System Info":{"systemHistory":[{"dateUtc":"2026-06-04T22:07:01.9951209Z","cpu":15.124,"memory":34601768.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.1022,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:12.0040416Z","cpu":17.410,"memory":34681780.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.1627,"availableThreads":32764,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:22.0196796Z","cpu":11.193,"memory":34831228.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.3392,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:32.0303156Z","cpu":11.466,"memory":34820240.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.1045,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:42.0410113Z","cpu":11.445,"memory":34803188.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.1655,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:52.0490884Z","cpu":11.271,"memory":34864952.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.0928,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0}]}},"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.TelemetryHandler","duration in milliseconds":48.3912,"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.RetryHandler","duration in milliseconds":48.3829,"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.RouterHandler","duration in milliseconds":48.3634,"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.TransportHandler","duration in milliseconds":48.3576,"children":[{"name":"Microsoft.Azure.Cosmos.GatewayStoreModel Transport Request","duration in milliseconds":48.3201,"data":{"Client Side Request Stats":{"Id":"AggregatedClientSideRequestStatistics","ContactedReplicas":[],"RegionsContacted":[],"FailedReplicas":[],"AddressResolutionStatistics":[],"StoreResponseStatistics":[],"HttpResponseStats":[{"StartTimeUTC":"2026-06-04T22:07:55.5835792Z","DurationInMs":43.4269,"RequestUri":"https://swkrprod-centralus.documents.azure.com:10250/","ResourceType":"Document","HttpMethod":"POST","ActivityId":"fbac5a0e-85e3-4f9c-a4d5-2d9232a0bb0e","StatusCode":"OK"}]},"PointOperationStatisticsTraceDatum":{"Id":"PointOperationStatistics","ActivityId":"b05f813b-26d5-495a-b8e9-754f87772853","ResponseTimeUtc":"2026-06-04T22:07:55.6273816Z","StatusCode":200,"SubStatusCode":0,"RequestCharge":2,"RequestUri":"dbs/testDbGateway/colls/testContainerGateway/docs/killswitch_f65e27662dc347f183d93fbef7c36f69","ErrorMessage":null,"RequestSessionToken":"0:2#23#4=-1","ResponseSessionToken":"0:2#23#4=-1","BELatencyInMs":"1.189"}}}]}]}]}]}]}]},{"name":"Response Serialization","duration in milliseconds":0.0597}]}

**Once thinclient flag is disabled, switched to gateway mode:**
[2026-06-04T22:07:56.9972118Z] elapsed=00:14:35 success=826 failures=0 status=OK diagnostics={"Summary":{"GatewayCalls":{"(200, 0)":1}},"name":"ReadItemAsync","start datetime":"2026-06-04T22:07:56.643Z","duration in milliseconds":353.4119,"data":{"Client Configuration":{"Client Created Time Utc":"2026-06-04T21:52:51.3268455Z","MachineId":"vmId:611ed772-5bcc-4eb3-90bc-d26d23956e13","VM Region":"westus2","NumberOfClientsCreated":1,"NumberOfActiveClients":1,"ConnectionMode":"Gateway","User Agent":"cosmos-netstandard-sdk/3.60.0|1|X64|Microsoft Windows 10.0.26200|.NET 6.0.36|N|F4","ConnectionConfig":{"gw":"(cps:50, urto:6, p:False, httpf: False)","rntbd":"(cto: 5, icto: -1, mrpc: 30, mcpe: 65535, erd: True, pr: ReuseUnicastPort)","other":"(ed:False, be:False)"},"ConsistencyConfig":"(consistency: NotSet, readStrategy: NotSet, prgns:[], apprgn: )","ProcessorCount":16}},"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.RequestInvokerHandler","duration in milliseconds":353.223,"children":[{"name":"Get Collection Cache","duration in milliseconds":0.0009},{"name":"Microsoft.Azure.Cosmos.Handlers.DiagnosticsHandler","duration in milliseconds":353.1496,"data":{"System Info":{"systemHistory":[{"dateUtc":"2026-06-04T22:07:01.9951209Z","cpu":15.124,"memory":34601768.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.1022,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:12.0040416Z","cpu":17.410,"memory":34681780.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.1627,"availableThreads":32764,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:22.0196796Z","cpu":11.193,"memory":34831228.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.3392,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:32.0303156Z","cpu":11.466,"memory":34820240.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.1045,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:42.0410113Z","cpu":11.445,"memory":34803188.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.1655,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0},{"dateUtc":"2026-06-04T22:07:52.0490884Z","cpu":11.271,"memory":34864952.000,"threadInfo":{"isThreadStarving":"False","threadWaitIntervalInMs":0.0928,"availableThreads":32765,"minThreads":16,"maxThreads":32767},"numberOfOpenTcpConnection":0}]}},"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.TelemetryHandler","duration in milliseconds":353.141,"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.RetryHandler","duration in milliseconds":353.1334,"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.RouterHandler","duration in milliseconds":353.1166,"children":[{"name":"Microsoft.Azure.Cosmos.Handlers.TransportHandler","duration in milliseconds":353.1121,"children":[{"name":"Microsoft.Azure.Cosmos.GatewayStoreModel Transport Request","duration in milliseconds":352.1332,"data":{"Client Side Request Stats":{"Id":"AggregatedClientSideRequestStatistics","ContactedReplicas":[],"RegionsContacted":[],"FailedReplicas":[],"AddressResolutionStatistics":[],"StoreResponseStatistics":[],"HttpResponseStats":[{"StartTimeUTC":"2026-06-04T22:07:56.6455977Z","DurationInMs":348.8849,"RequestUri":"https://swkrprod-centralus.documents.azure.com/dbs/testDbGateway/colls/testContainerGateway/docs/killswitch_f65e27662dc347f183d93fbef7c36f69","ResourceType":"Document","HttpMethod":"GET","ActivityId":"f6d990b0-75a7-42ac-92dd-106f2b62a7e8","StatusCode":"OK"}]},"PointOperationStatisticsTraceDatum":{"Id":"PointOperationStatistics","ActivityId":"f6d990b0-75a7-42ac-92dd-106f2b62a7e8","ResponseTimeUtc":"2026-06-04T22:07:56.9970564Z","StatusCode":200,"SubStatusCode":0,"RequestCharge":2,"RequestUri":"dbs/testDbGateway/colls/testContainerGateway/docs/killswitch_f65e27662dc347f183d93fbef7c36f69","ErrorMessage":null,"RequestSessionToken":"0:2#23#4=-1","ResponseSessionToken":"0:2#23#4=-1","BELatencyInMs":"1.466"}}}]}]}]}]}]}]},{"name":"Response Serialization","duration in milliseconds":0.0447}]}

## Type of change

## Closing issues

## Changelog

- [X ] I have added a changelog entry under `### Unreleased` in `changelog.md`

>


Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -240,16 +243,26 @@ public void MarkEndpointUnavailableForWrite(Uri endpoint)
         /// Invoked when <see cref="AccountProperties"/> is read
         /// </summary>
         /// <param name="databaseAccount">Read DatabaseAccoaunt </param>
-        public void OnDatabaseAccountRead(AccountProperties databaseAccount)
-        {
-            this.UpdateLocationCache(
-                databaseAccount.WritableRegions,
+        public void OnDatabaseAccountRead(AccountProperties databaseAccount)
```
kushagraThapar: This method should be synchronized or should have some sort of semaphore around it to make sure concurrent readers get the correct value of databaseAccount.ThinClientReadableLocationsInternal and the write locations internal.

kushagraThapar:
**Test silently masked by the new `HasThinClientReadLocations` gate**

📁 `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs:1322-1394`

The pre-existing test **`ThinClient_ProcessMessageAsync_WithUnsupportedOperations_ShouldFallbackToGatewayMode`** was not updated to call `TestUtils.EnableThinClientLocationsForTest`. With the new code path, `HasThinClientReadLocations == false` by default → `IsThinClientRoutable` short-circuits **before** `IsOperationSupportedByThinClient` is consulted.

The test still passes, but it no longer proves what its name claims. A future regression in `IsOperationSupportedByThinClient` (the exact predicate the test was written to guard) will go uncaught — this is a silent loss of regression coverage, not a flake.

**Verification:**
```bash
awk '/ThinClient_ProcessMessageAsync_WithUnsupportedOperations/,/^        \[TestMethod\]/' \
    Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs \
  | grep -c "EnableThinClientLocationsForTest"
# → 0
```

**Suggested fix (1 line):** add `TestUtils.EnableThinClientLocationsForTest(endpointManager);` immediately after the `endpointManager` construction in that test.

**Also recommend:** audit the other thin-client tests in `GatewayStoreModelTest.cs` that pre-date this gate — same masking pattern may exist elsewhere and would be just as silent.

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -589,6 +588,32 @@ internal static bool IsOperationSupportedByThinClient(DocumentServiceRequest req
...
         }
 
+        /// <summary>
+        /// Returns <c>true</c> iff the request is currently eligible for thin-client dispatch:
+        /// the operation type is supported AND the service is still advertising thin-client
+        /// endpoints for the request's direction. When either condition is false the dispatch
+        /// falls back to the regular gateway path on the very next request without a client
+        /// restart (kill-switch). Mirrors Java SDK <c>ThinClientStoreModel.getEndpoint</c>.
+        /// </summary>
+        internal static bool IsThinClientRoutable(IGlobalEndpointManager endpointManager, DocumentServiceRequest request)
```
ananth7592, line 598: Nice central helper here. One thing I'm not seeing in the test additions though — we have great coverage for the *flag* transitions in `LocationCacheTests`, but I don't see a test that actually proves the dispatch fallback works end-to-end: i.e. a thin-client-routed request, then refresh `OnDatabaseAccountRead` with no thin-client locations, then assert the next supported request goes through `gatewayStoreClient.InvokeAsync` instead of `thinClientStoreClient.InvokeAsync`. That's the customer-visible behavior the PR claims to fix, and right now a future regression that broke `IsThinClientRoutable` (e.g. someone short-circuiting it or aliasing the two flags) would still pass.

Could you add that test here, plus a symmetric one for `ClientRetryPolicy.OnBeforeSendRequest` covering the retry dispatch site?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/LocationCacheTests.cs
```diff
@@ -1926,6 +1926,86 @@ public void ValidateThinClientLocationCacheFlowTest()
...
+        /// Regression test for the thin-client kill-switch: <see cref="LocationCache.HasThinClientReadLocations"/>
+        /// and <see cref="LocationCache.HasThinClientWriteLocations"/> must each track the most
+        /// recent <see cref="LocationCache.OnDatabaseAccountRead"/> snapshot — independently per
+        /// direction — so the SDK falls back to plain gateway on the very next request when the
+        /// service withdraws thin-client locations, with no client restart. Mirrors Java SDK
+        /// <c>GlobalEndpointManager.hasThinClientReadLocations</c>, extended with an independent
+        /// write flag because .NET dispatches reads vs writes through separate gates.
+        /// </summary>
+        [TestMethod]
+        public void HasThinClientLocationsTracksRefreshesPerDirection()
```
ananth7592, line 1939: Really like this regression test — the read-withdrawn-but-write-remains step is exactly the kind of thing that proves independence.

One small ask: could you add the inverse case too (`thinClientReadEndpoint` present, `thinClientWriteEndpoint` null, assert `HasThinClientReadLocations == true && HasThinClientWriteLocations == false`)? The PR description specifically calls out independent tracking per direction, and right now an asymmetric regression — say one that aliased the two flags such that writes shadowed reads but not vice versa — would still pass.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
aavasthy: wrapped the body of OnDatabaseAccountRead in lock (this.lockObject) so the two flag writes and the UpdateLocationCache snapshot install are atomic with respect to concurrent readers

aavasthy:
> **Test silently masked by the new `HasThinClientReadLocations` gate**
> 
> 📁 `Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs:1322-1394`
> 
> The pre-existing test **`ThinClient_ProcessMessageAsync_WithUnsupportedOperations_ShouldFallbackToGatewayMode`** was not updated to call `TestUtils.EnableThinClientLocationsForTest`. With the new code path, `HasThinClientReadLocations == false` by default → `IsThinClientRoutable` short-circuits **before** `IsOperationSupportedByThinClient` is consulted.
> 
> The test still passes, but it no longer proves what its name claims. A future regression in `IsOperationSupportedByThinClient` (the exact predicate the test was written to guard) will go uncaught — this is a silent loss of regression coverage, not a flake.
> 
> **Verification:**
> 
> ```shell
> awk '/ThinClient_ProcessMessageAsync_WithUnsupportedOperations/,/^        \[TestMethod\]/' \
>     Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs \
>   | grep -c "EnableThinClientLocationsForTest"
> # → 0
> ```
> 
> **Suggested fix (1 line):** add `TestUtils.EnableThinClientLocationsForTest(endpointManager);` immediately after the `endpointManager` construction in that test.
> 
> **Also recommend:** audit the other thin-client tests in `GatewayStoreModelTest.cs` that pre-date this gate — same masking pattern may exist elsewhere and would be just as silent.

Fixed and audited. Added TestUtils.EnableThinClientLocationsForTest(endpointManager); to ThinClient_ProcessMessageAsync_WithUnsupportedOperations_ShouldFallbackToGatewayMode and ThinClient_ProcessMessageAsync_WithFullFidelityChangeFeed_ShouldFallbackToGatewayMode. Both now exercise the predicate they were written to guard (IsOperationSupportedByThinClient) instead of short-circuiting on the new gate.

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
aavasthy: Added both and also tested this scenario against live accounts where we are explicilty turnig off thinclient endpoint flag from service side via ACIS:

   1.GatewayStoreModelTest.ThinClient_ProcessMessageAsync_AfterServiceWithdrawsThinLocations_FallsBackToGateway — uses MockThinClientStoreClient + Mock<CosmosHttpClient> counters: dispatches a thin-routable request, asserts thin counter == 1; calls OnDatabaseAccountRead with empty thin locations; dispatches the same request again; asserts thin counter stays at 1 and gateway SendHttpAsync was invoked.
   2.ClientRetryPolicyTests.ThinClient_OnBeforeSendRequest_AfterServiceWithdrawsThinLocations_FallsBackToGatewayEndpoint — symmetric coverage for the retry dispatch site, asserting on request.RequestContext.LocationEndpointToRoute switching from a :10250 thin-client URI to the regular gateway URI on the next request after withdrawal.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/LocationCacheTests.cs
aavasthy: I don't think the inverse case adds coverage in practice. The existing matrix step 2 (reads withdrawn, writes remain) already asserts both flags independently in the same step (Assert.IsFalse(HasThinClientReadLocations); Assert.IsTrue(HasThinClientWriteLocations);). Any aliasing regression (e.g. HasThinClientWriteLocations => hasThinClientReadLocations) or shadow regression (e.g. hasThinClientWriteLocations = readAvailable) fails this exact assertion symmetrically, regardless of which direction was wrong. The inverse step would catch the same class of bug a second time, not a new one.

changelog.md
```diff
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [5910](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5910) Upgraded Direct package to 3.43.2.
 - [5910](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5910) Direct: Fixed RNTBD thread-pool starvation by making `Dispatcher.OnIdleTimer` asynchronous (ports public [PR 5817](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5817))
+- [5927](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5927) ThinClient: Fixes mid-flight fallback to gateway when the service stops advertising thin-client endpoints. Previously the SDK kept routing to stale thin-client URIs and required a client restart to recover; now `LocationCache.HasThinClientReadLocations` and `LocationCache.HasThinClientWriteLocations` are re-evaluated on every account refresh and per-request dispatch gates fall back to the regular gateway path on the very next request. The two directions are tracked independently so that partial withdrawal (service drops only thin-client reads or only writes) also flips back to gateway for the affected direction.
```
jeet1995: suggestion: Changelog entry can be shortened by removing implementation details.

jeet1995 (changes requested):
We should take the approach where ThinClientStoreModel is introduced. This goes back to Neil's suggestion here -  https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5614#discussion_r2885539207 - you can localize choosing the *StoreModel in the outermost DocumentClient-layer this way - https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java#L7372. Also, this way you guarantee that thin-client endpoints are used on a per-logical operation scope as opposed to per-I/O scope - makes debugging easier.

aavasthy:
> We should take the approach where ThinClientStoreModel is introduced. This goes back to Neil's suggestion here - [#5614 (comment)](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5614#discussion_r2885539207) - you can localize choosing the *StoreModel in the outermost DocumentClient-layer this way - https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java#L7372. Also, this way you guarantee that thin-client endpoints are used on a per-logical operation scope as opposed to per-I/O scope - makes debugging easier.

ThinclientStoreModel was purposely removed as part of design change for better code modularity in this PR https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5323. We have added a work item to revert this change but during internal design discussion this item has been given a lower priority https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5684

kushagraThapar (dismissed):
LGTM thanks @aavasthy

kushagraThapar (approved):
LGTM, thanks @aavasthy let's follow up with the ThinClientStoreModel refactoring in a follow up PR before we GA this, thanks!

jeet1995 (approved):

kundadebdatta (approved):
LGTM

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -242,14 +245,27 @@ public void MarkEndpointUnavailableForWrite(Uri endpoint)
...
         {
-            this.UpdateLocationCache(
-                databaseAccount.WritableRegions,
-                databaseAccount.ReadableRegions,
-                thinClientWriteLocations: databaseAccount.ThinClientWritableLocationsInternal,
-                thinClientReadLocations: databaseAccount.ThinClientReadableLocationsInternal,
-                preferenceList: null,
-                enableMultipleWriteLocations: databaseAccount.EnableMultipleWriteLocations);
-        }
+            lock (this.lockObject)
```
jeet1995, line 248: This will ensure the update path is under a lock but do readers also require acquiring this lock?

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -589,6 +588,32 @@ internal static bool IsOperationSupportedByThinClient(DocumentServiceRequest req
...
+        {
+            return IsOperationSupportedByThinClient(request)
+                && (request.IsReadOnlyRequest
+                    ? endpointManager.HasThinClientReadLocations
+                    : endpointManager.HasThinClientWriteLocations);
+        }
+
+        /// <summary>
+        /// Read-direction variant of <see cref="IsThinClientRoutable"/> for failover walks
+        /// (PLCB / PPAF) that traverse thin-client READ endpoints regardless of the original
```
jeet1995, line 608: nit: PPCB

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -122,9 +122,8 @@ await GatewayStoreModel.ApplySessionTokenAsync(
                     }
                 }
 
-                bool canUseThinClient =
-                    this.thinClientStoreClient != null &&
-                    GatewayStoreModel.IsOperationSupportedByThinClient(request);
+                bool canUseThinClient = this.thinClientStoreClient != null
```
jeet1995, line 125: My general concern about having the check here is between retries `canUseThinClient` can flip. Is there a way to push this boolean switch more upstream so at operation level you make the call once?

changelog.md
aavasthy: updated

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
jeet1995: Also, I'm curious if we need a read-write reentrant lock here?

aavasthy: Readers don't need the lock. The flags are volatile bool, so each read is atomic and gets the latest value via CLR memory barriers. The writer-side lock is just a safety net to keep the flag write and the snapshot update from drifting apart

Reads are already lock-free, and adding one would just slow down the per-request path without solving a contention problem we actually have. Re-entrancy is also a drawback— UpdateLocationCache re-enters the same lock, which Monitor handles natively but ReaderWriterLockSlim doesn't by default. Since the writer only runs once every 5 min and readers stay lock-free, the current pattern is the right shape.

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
aavasthy: The per-retry check is intentional — caching it at the operation level would re-introduce the exact bug this PR fixes. The whole point is that when the service stops advertising thin-client endpoints mid-flight, the SDK should fall back to gateway on the very next request without a client restart. So the flow is: operation starts with canUseThinClient=true, service withdraws endpoints, OnDatabaseAccountRead flips the flag, and the next retry re-evaluates to false and goes through gateway.

PR #5648, opened 2026-03-03
[Internal] PPAF: Adds Hub Region Caching Per Partition Level

aavasthy:
## Description

This PR introduces **per-partition hub region caching** for the Azure Cosmos DB .NET SDK v3. On single-master accounts encountering repeated `404/1002` (`ReadSessionNotAvailable`) errors, the SDK now discovers the hub region via a `403/3` (WriteForbidden) discovery chain and **caches** the result. Subsequent requests to the same partition route directly to the cached hub, eliminating redundant discovery round-trips.

**Follow-up of:** PR [#5447](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5447)  
**Branch:** `users/aavasthy/hubregioncaching`  
**Type:** New feature (non-breaking)  
**Lines changed:** +1,966 / −363

---

## Problem

| Issue | Impact |
|-------|--------|
| After 2× `404/1002` on a single-master account, the SDK **gave up** and returned the error to the caller | Read requests failed unnecessarily |
| No hub region discovery existed for non-PPAF accounts | Non-PPAF single-master accounts had no recovery mechanism |
| Every request repeated the full hub discovery chain — no caching | Unnecessary latency on every request hitting the same partition |

---

## Solution Summary

1. After **2 consecutive `404/1002`** errors on a single-master account, set the `x-ms-cosmos-hub-region-processing-only` header.
2. Non-hub regions receiving this header return `403/3` (WriteForbidden) — the SDK retries to the next region (the **403/3 discovery chain**).
3. Once the **hub region** responds with `200 OK`, the SDK **caches** the hub region URI for that partition in `PartitionKeyRangeToLocationForWrite`.
4. Future requests for the same partition route **directly to the cached hub**, skipping the discovery chain entirely.
5. Works for **both PPAF and non-PPAF** accounts by reusing the existing partition-level failover infrastructure.

---

## Existing vs. New Behavior

| Aspect | Before (Existing) | After (New) |
|--------|-------------------|-------------|
| Max retries on `404/1002` (single-master) | 2, then fail | 2 + hub discovery chain |
| Hub header trigger | After 1st `404/1002` | After 2nd `404/1002` |
| `403/3` on read path with hub header | Not handled | Retries to continue discovery |
| Partition-level cache on warm path | Only PPAF accounts | Both PPAF and non-PPAF |
| Non-PPAF cache routing | Not available | Available via `checkHubRegionOverrideInCache` |
[...]

PR #5928, opened 2026-06-05
[Internal] Thin Client Integration: Fixes Exclude regions check for thinclient mode.

### LINKED ISSUE #5789: ExcludeRegions on ItemRequestOptions is ignored when thin client mode is enabled
### Describe the bug

When thin client mode is enabled (`AZURE_COSMOS_THIN_CLIENT_ENABLED=true`), per-operation `ExcludeRegions` on `ItemRequestOptions` is ignored. All requests route to the first available thin client endpoint (typically the first write region) regardless of the excluded regions list.

Without thin client enabled, `ExcludeRegions` works correctly — each operation routes to the expected region.

### To reproduce

Multi-writer account with 3 write regions (East US 2, West Central US, South Central US). Three `CreateItemAsync` calls, each excluding 2 regions to pin to a specific target region.

```csharp
Environment.SetEnvironmentVariable("AZURE_COSMOS_THIN_CLIENT_ENABLED", "true");

using var client = new CosmosClient(endpoint, key, new CosmosClientOptions
{
    ConnectionMode = ConnectionMode.Gateway,
    ApplicationPreferredRegions = new List<string> { "East US 2", "West Central US", "South Central US" }
});

// Pin to West Central US by excluding the other two
var opts = new ItemRequestOptions
{
    ExcludeRegions = new List<string> { "East US 2", "South Central US" }
};
var resp = await container.CreateItemAsync(item, pk, opts);
// Diagnostics show request went to eastus2:10250, NOT westcentralus:10250
```

**Expected**: Request routes to `westcentralus.documents.azure.com:10250`
**Actual**: Request routes to `eastus2.documents.azure.com:10250`

### Root cause analysis

`RequestInvokerHandler.SendAsync()` (line ~125-146) branches for thin client operations and calls `endpointManager.ResolveThinClientEndpoint(request)` instead of `ResolveServiceEndpoint(request)`.

`LocationCache.ResolveThinClientEndpoint()` (line ~933-949) selects from precomputed `ThinClientWriteEndpoints`/`ThinClientReadEndpoints` using index-based selection. It **does not check `request.RequestContext.ExcludeRegions`**.

The `ExcludeRegions` filtering only exists in the normal gateway path: `LocationCache.ResolveServiceEndpoint()` → `GetApplicableEndpoints()` (line ~416-448).

### Suggested fix

`ResolveThinClientEndpoint()` should filter `ThinClientWriteEndpoints`/`ThinClientReadEndpoints` by `request.RequestContext.ExcludeRegions` before selecting the endpoint, mirroring the logic in `GetApplicableEndpoints()`.

### Environment

- **SDK**: Microsoft.Azure.Cosmos 3.58.0
- **OS**: Windows
- **.NET**: 10.0

aavasthy:
# Pull Request Template

## Description

This change resolves an issue where per-request region exclusions were not honored when the SDK operated in thin client mode. Previously, requests configured with ExcludeRegions would still route to theprimary thin client endpoint regardless of the customer's regional preferences, preventing them from steering traffic away from specific regions during operational events. With this fix, thin client requestrouting now filters available endpoints against the excluded region list in a manner consistent with the standard gateway path, restoring expected behavior for both individual operations and cross-region hedging scenarios. The change is scoped narrowly to the thin client resolution path, preserves all existing routing semantics when no exclusions are specified, and is covered by new unit and integration test.

Code cleanup: Update thinclient port to 10250 from 10650 as that's the new port thinclient is available at.

Fix CosmosAvailabiltyStrategyTests: Fix threshold value due to hub region caching introduced in this PR https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5648
## Type of change

## Closing issues

To automatically close an issue: closes #5789

## Changelog

>

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
@@ -982,14 +982,17 @@ namespace Microsoft.Azure.Cosmos.Routing
 
         /// <summary>
         /// Resolves the applicable thin client endpoints for a request, honoring per-request
-        /// <see cref="DocumentServiceRequestContext.ExcludeRegions"/>. Mirrors the gateway
-        /// path in <see cref="GetApplicableEndpoints(DocumentServiceRequest, bool)"/>.
+        /// <see cref="DocumentServiceRequestContext.ExcludeRegions"/>. Uses the same preferred-
+        /// location filter as the gateway path in <see cref="GetApplicableEndpoints(DocumentServiceRequest, bool)"/>,
+        /// differing only in which fallback endpoint is selected.
         /// </summary>
         /// <remarks>
-        /// When all preferred regions are excluded, falls back to the first thin client write
-        /// endpoint (which itself falls back to <see cref="defaultEndpoint"/> at initialization
-        /// time). This keeps requests on the thin client routing path rather than dropping back
-        /// to the gateway endpoint and breaking thin client port semantics.
+        /// When all preferred regions are excluded, this method always falls back to the first
+        /// thin client write endpoint (<c>snapshot.ThinClientWriteEndpoints[0]</c>) for both reads
+        /// and writes. This deliberately diverges from the gateway path's fallback selection
+        /// (which uses <see cref="defaultEndpoint"/>, or <c>WriteEndpoints[0]</c> under PPAF for
+        /// reads) because <see cref="defaultEndpoint"/> resolves to the gateway port (443) and
+        /// would break thin client port semantics if used here.
         /// </remarks>
         private ReadOnlyCollection<Uri> GetApplicableThinClientEndpoints(
             DocumentServiceRequest request,
--- a/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
@@ -955,15 +955,44 @@ namespace Microsoft.Azure.Cosmos.Routing
             }
 
             DatabaseAccountLocationsInfo snapshot = this.locationInfo;
-            ReadOnlyCollection<Uri> endpoints = isReadRequest
-                ? snapshot.ThinClientReadEndpoints
-                : snapshot.ThinClientWriteEndpoints;
+            ReadOnlyCollection<Uri> endpoints = this.GetApplicableThinClientEndpoints(request, isReadRequest, snapshot);
 
             int locationIndex = request.RequestContext.LocationIndexToRoute.GetValueOrDefault(0);
             Uri chosenEndpoint = endpoints[locationIndex % endpoints.Count];
 
             request.RequestContext.RouteToLocation(chosenEndpoint);
             return chosenEndpoint;
+        }
+
+        /// <summary>
+        /// Resolves the applicable thin client endpoints for a request, honoring per-request
+        /// <see cref="DocumentServiceRequestContext.ExcludeRegions"/>. Mirrors the gateway
+        /// path in <see cref="GetApplicableEndpoints(DocumentServiceRequest, bool)"/>.
+        /// </summary>
+        /// <remarks>
+        /// When all preferred regions are excluded, falls back to the first thin client write
+        /// endpoint (which itself falls back to <see cref="defaultEndpoint"/> at initialization
+        /// time). This keeps requests on the thin client routing path rather than dropping back
+        /// to the gateway endpoint and breaking thin client port semantics.
+        /// </remarks>
+        private ReadOnlyCollection<Uri> GetApplicableThinClientEndpoints(
+            DocumentServiceRequest request,
+            bool isReadRequest,
+            DatabaseAccountLocationsInfo snapshot)
+        {
+            IReadOnlyList<string> excludeRegions = request.RequestContext?.ExcludeRegions;
+            if (excludeRegions == null || excludeRegions.Count == 0)
+            {
+                return isReadRequest ? snapshot.ThinClientReadEndpoints : snapshot.ThinClientWriteEndpoints;
+            }
+
+            Uri fallbackEndpoint = snapshot.ThinClientWriteEndpoints[0];
+
+            return GetApplicableEndpoints(
+                isReadRequest ? snapshot.ThinClientReadEndpointByLocation : snapshot.ThinClientWriteEndpointByLocation,
+                snapshot.EffectivePreferredLocations,
+                fallbackEndpoint,
+                excludeRegions);
         }
 
         private void SetServicePointConnectionLimit(Uri endpoint)
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
@@ -593,11 +593,9 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests
                        // ClientRetryPolicy.ShouldRetryOnSessionNotAvailable retries wire 2 to the
                        // WRITE region (RetryLocationIndex = 0, RetryRequestOnPreferredLocations = false).
                        // When the write region differs from region1 and has no fault, wire 2 succeeds
-                       // in ~50-100ms — beating any threshold >= 50ms. Use 10ms to guarantee the hedge
-                       // timer fires before primary's wire-2 retry can complete, so the hedge arm is
-                       // dispatched (requestNumber > 0) and HedgeContext is stamped in diagnostics.
-                       threshold: TimeSpan.FromMilliseconds(100),
-                       thresholdStep: TimeSpan.FromMilliseconds(50)),
+                       // in ~50-100ms — beating any threshold >= 50ms.
+                       threshold: TimeSpan.FromMilliseconds(50),
+                       thresholdStep: TimeSpan.FromMilliseconds(20)),
                 Serializer = this.cosmosSystemTextJsonSerializer
             };
 
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
@@ -596,8 +596,8 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests
                        // in ~50-100ms — beating any threshold >= 50ms. Use 10ms to guarantee the hedge
                        // timer fires before primary's wire-2 retry can complete, so the hedge arm is
                        // dispatched (requestNumber > 0) and HedgeContext is stamped in diagnostics.
-                       threshold: TimeSpan.FromMilliseconds(10),
-                       thresholdStep: TimeSpan.FromMilliseconds(10)),
+                       threshold: TimeSpan.FromMilliseconds(100),
+                       thresholdStep: TimeSpan.FromMilliseconds(50)),
                 Serializer = this.cosmosSystemTextJsonSerializer
             };
 
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
@@ -587,10 +587,17 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests
             CosmosClientOptions clientOptions = new CosmosClientOptions()
             {
                 ConnectionMode = ConnectionMode.Direct,
-                ApplicationPreferredRegions = isPreferredLocationsEmpty ? new List<string>() :new List<string>() { region1, region2 },
+                ApplicationPreferredRegions = isPreferredLocationsEmpty ? new List<string>() : new List<string>() { region1, region2 },
                 AvailabilityStrategy = AvailabilityStrategy.CrossRegionHedgingStrategy(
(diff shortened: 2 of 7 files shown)


kushagraThapar (commented):
Thanks @aavasthy, overall looks good, it would be good to have some more testing on all regions excluded code path and end to end hedging tests?

aavasthy:
> Thanks @aavasthy, overall looks good, it would be good to have some more testing on all regions excluded code path and end to end hedging tests?

Added exclude regions with all operations supported in thinclient in CosmosItemThinClientTest end to end tests.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
```diff
@@ -1607,5 +1607,96 @@ protected override Task<HttpResponseMessage> SendAsync(
                 return base.SendAsync(request, cancellationToken);
             }
         }
-    }
+
+        [TestMethod]
+        [TestCategory("ThinClient")]
+        public async Task CreateItem_WithExcludeRegions_OnThinClient_PinsToNonExcludedReadRegion()
```
jeet1995: Can we add tests spanning various operation types? Also does hedging take a dependency on exclude regions (it does in the Java SDK) so will be good to add that coverage too.

Another aspect is the exclude region wiring for QueryPlan too.

aavasthy: Added exclude regions with all operations supported in thinclient in CosmosItemThinClientTest end to end tests.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
```diff
@@ -680,6 +687,54 @@ public async Task ReadItemsTest()
...
+                {
+                    ExcludeRegions = new List<string>(ExcludeRegions),
+                });
+
+            string diagnostics = readResponse.Diagnostics.ToString();
+            Assert.AreEqual(HttpStatusCode.OK, readResponse.StatusCode);
+            Assert.AreEqual(seed.Id, readResponse.Resource.Id);
+            Assert.IsTrue(
+                diagnostics.Contains("|F4"),
+                "Read should route through the thin client pipeline (|F4 user agent token).");
```
kushagraThapar, line 741: I am curious on these tests, if requests are successfully routed through  ThinClient, how are we testing that they are not being routed through Excluded Regions? On the added tests, can we check if primary region is the excluded one, and then all the requests bypass that region and then go to the next available region which is not part of the excluded regions?

aavasthy: I tested this via diagnostics logs which region, it is hitting.
Added more assertions through diagnostic logs to ensure request is not routed to excluded region.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
```diff
@@ -116,73 +123,84 @@ private async Task<List<TestObject>> CreateItemsSafeAsync(IEnumerable<TestObject
...
-            CosmosClientBuilder builder = new CosmosClientBuilder(this.connectionString)
-                .WithConnectionModeGateway()
-                .WithCustomSerializer(serializer)
-                .WithSendingRequestEventArgs(async (sender, e) =>
-                {
+        private static void AssertExcludedRegionsNotInDiagnostics(string diagnostics)
+        {
+            foreach (string excludedRegion in ExcludeRegions)
+            {
+                string excludedHost = excludedRegion.Replace(" ", string.Empty).ToLowerInvariant() + ".documents.azure.com:10250";
```
kushagraThapar, line 130: I noticed we use a different port everywhere else, is this intentional change or a typo?

kushagraThapar:
We should also add a test which excludeds all regions which are part of the preferred regions to test the fallback path here. Do we fail the request in that case or just fallback to global endpoint?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
aavasthy: Till last year thinclient was on port 10650 but now it is on port 10250. Live tests have 10250 port.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -947,23 +947,52 @@ internal bool CanUseMultipleWriteLocations()
...
+            bool isReadRequest,
+            DatabaseAccountLocationsInfo snapshot)
+        {
+            IReadOnlyList<string> excludeRegions = request.RequestContext?.ExcludeRegions;
+            if (excludeRegions == null || excludeRegions.Count == 0)
+            {
+                return isReadRequest ? snapshot.ThinClientReadEndpoints : snapshot.ThinClientWriteEndpoints;
+            }
+
+            Uri fallbackEndpoint = snapshot.ThinClientWriteEndpoints[0];
```
aavasthy, line 1008: @kushagraThapar this is the fallback endpoint if all regions are put under exclude regions.

kushagraThapar (dismissed):
LGTM, thanks @aavasthy 
For consistency, we should update the port numbers across the tests to be same, unless there is a strong test reason to keep them separate, otherwise, looks good to me!

aavasthy:
> LGTM, thanks @aavasthy For consistency, we should update the port numbers across the tests to be same, unless there is a strong test reason to keep them separate, otherwise, looks good to me!

yes updated the port value to 10250 everywhere in repo.

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -963,23 +963,52 @@ internal bool CanUseMultipleWriteLocations()
...
```
NaluTripician, line 1008: Worth a quick confirm: for a read where all preferred regions are excluded, this always falls back to `ThinClientWriteEndpoints[0]` (the write region), whereas the gateway path only does that under PPAF and otherwise uses `defaultEndpoint`. The `<remarks>` says this is intentional to stay on the thin client port path, which makes sense — just flagging that "mirrors the gateway path" slightly overstates parity for this read-fallback case.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/LocationCacheTests.cs
```diff
@@ -2007,6 +2007,63 @@ static Collection<AccountRegion> ToCollection(string endpoint)
         }
 
 
+        [TestMethod]
+        public void ValidateResolveThinClientEndpoint_HonorsExcludeRegionsTest()
```
NaluTripician: This is the only exclude-regions case running in the unit (gating) suite, and it only covers multi-write + a write request. Can we add unit cases here for (a) `isReadRequest: true` (it selects a different dictionary), (b) all preferred regions excluded — asserting the result equals `ThinClientWriteEndpoints[0]` for both directions, and (c) a single-write account? Those branches are otherwise only exercised by the emulator suite, which doesn't gate the PR build, so the fix could regress with green CI.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
```diff
@@ -677,9 +696,98 @@ public async Task ReadItemsTest()
                 Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
                 Assert.AreEqual(item.Id, response.Resource.Id);
                 Assert.IsTrue(diagnostics.Contains("|F4"), "Diagnostics User Agent should contain '|F4' for ThinClient");
+                AssertExcludedRegionsNotInDiagnostics(diagnostics);
             }
         }
 
+        [TestMethod]
+        [TestCategory("ThinClient")]
+        public async Task ReadItem_WithHedgingAndExcludeRegions_OnThinClient_Succeeds()
```
NaluTripician, line 705: This test never asserts that a hedge arm actually fired — it only checks OK + excluded-host absence. With a 100ms threshold against a healthy emulator the primary will almost always win, so a hedge-path exclude bug (exactly what this PR fixes) would still pass. Can we inject a delay/fault so the hedge arm fires, then assert `HedgeContext` shows up in diagnostics?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
```diff
@@ -99,6 +105,7 @@ private IEnumerable<TestObject> GenerateItems(string partitionKey)
         private async Task<List<TestObject>> CreateItemsSafeAsync(IEnumerable<TestObject> items)
         {
             List<TestObject> itemsCreated = new List<TestObject>();
+            await Task.Delay(TimeSpan.FromSeconds(30));
```
NaluTripician, line 108: There are ~9 of these fixed `Task.Delay(30s)` waits across the file (~4.5 min total). Fixed sleeps are slow and get flaky under load. Could we swap them for a poll-until-ready helper (await endpoint/account-info availability), or hoist the wait into `TestInitialize` once?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
```diff
@@ -587,10 +587,17 @@ public async Task AvailabilityStrategyAllFaultsTests(string operation, string co
...
-                        threshold: TimeSpan.FromMilliseconds(200),
-                        thresholdStep: TimeSpan.FromMilliseconds(50)),
+                       // Threshold is intentionally aggressive: for single-master read 404/1002,
+                       // ClientRetryPolicy.ShouldRetryOnSessionNotAvailable retries wire 2 to the
+                       // WRITE region (RetryLocationIndex = 0, RetryRequestOnPreferredLocations = false).
+                       // When the write region differs from region1 and has no fault, wire 2 succeeds
+                       // in ~50-100ms — beating any threshold >= 50ms. Use 10ms to guarantee the hedge
+                       // timer fires before primary's wire-2 retry can complete, so the hedge arm is
+                       // dispatched (requestNumber > 0) and HedgeContext is stamped in diagnostics.
+                       threshold: TimeSpan.FromMilliseconds(10),
```
NaluTripician: 10ms is below typical CI timer/GC/scheduling jitter, so on a slow agent the primary can finish before the hedge timer is even scheduled and flip this assertion. The comment explains the upper bound well — but rather than racing the wall clock this tight, can we force the hedge arm via fault injection, or add a small tolerance/retry?

NaluTripician (dismissed):
Approving — the core fix is correct and well-scoped. Routing thin client requests through the same `GetApplicableEndpoints` filter the gateway path already uses is the right call, and I verified the empty-list/fallback paths can't throw: `ThinClientWriteEndpoints` is always seeded with `defaultEndpoint` and the static helper always returns at least the fallback, so `endpoints[locationIndex % endpoints.Count]` is safe. Port migration (10650 -> 10250) is complete and the changelog entry is accurate.

All my comments are non-blocking nits, mostly around test coverage and flakiness:
- Add unit (gating) coverage for the read-direction and all-regions-excluded fallback branches, plus a single-write account case — today only the multi-write/write happy path gates, while the read/fallback branches live only in the emulator suite.
- The hedging test doesn't assert a hedge arm actually fired.
- Several fixed `Task.Delay(30s)` sleeps and a 10ms hedge threshold are likely to be slow/flaky in CI.

None of these block merge. Thanks!

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
aavasthy: updated

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
aavasthy: Not all tests need this Task.Delay(30s) so not putting this in TestInitialize.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
aavasthy: updated test to assert Hedge context value

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
aavasthy: updated the comment section

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/LocationCacheTests.cs
aavasthy: Added more tests

NaluTripician (dismissed):

jeet1995 (dismissed):

kundadebdatta (dismissed):
LGTM

kundadebdatta (approved):
LGTM.

NaluTripician (approved):

PR #5931, opened 2026-06-08
ChangeFeed: Adds support for StartTime after partition merges

ramarag:
# Pull Request Template

## Description

In CosmosDB with in a partition, glsn and timestamp have relationship that they will increase in the same order.
But with Partition merge in CosmosDB maintains the original  glsn of the documents across partitions and lays them out in the glsn order. This ordering does not guarantee that time ordering is maintained with glsn order for any documents before partition merge. 
Since ChangeFeed with starttime just uses glsn as the continuation token it is not sufficient to maintain the time order.

Without this change: The service throws a bad request to keep the customer expectation intact. -> "StartTime/IfModifiedSince is not currently supported with merged partitions"

With this change: we send original start time with every request which the service will use to not return any documents before this time if the document's glsn is higher than the continuation token.

Similar change in Java -> https://github.com/Azure/azure-sdk-for-java/pull/32496/changes
## Type of change

## Closing issues

## Changelog

>

### DIFF
--- a/Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
@@ -59,13 +59,15 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed
                 container: container,
                 mode: mode,
                 changeFeedStartFrom: startFrom,
-                options: requestOptions);
+                options: requestOptions,
+                startTime: startTime);
         }
 
         private readonly CosmosClientContext clientContext;
 
         private readonly ChangeFeedRequestOptions changeFeedOptions;
         private readonly ChangeFeedMode mode;
+        private readonly DateTime? startTime;
 
         private ChangeFeedStartFrom changeFeedStartFrom;
         private bool hasMoreResultsInternal;
@@ -74,13 +76,15 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed
             ContainerInternal container,
             ChangeFeedMode mode,
             ChangeFeedStartFrom changeFeedStartFrom,
-            ChangeFeedRequestOptions options)
+            ChangeFeedRequestOptions options,
+            DateTime? startTime = null)
         {
             this.container = container ?? throw new ArgumentNullException(nameof(container));
             this.mode = mode;
             this.changeFeedStartFrom = changeFeedStartFrom ?? throw new ArgumentNullException(nameof(changeFeedStartFrom));
             this.clientContext = this.container.ClientContext;
             this.changeFeedOptions = options;
+            this.startTime = startTime;
 
             this.operationName = OpenTelemetryConstants.Operations.QueryChangeFeed;
         }
@@ -118,6 +122,19 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed
                     // Set time headers if any
                     ChangeFeedStartFromRequestOptionPopulator visitor = new ChangeFeedStartFromRequestOptionPopulator(requestMessage);
                     this.changeFeedStartFrom.Accept(visitor);
+
+                    // When a merge happens, the child partition will contain documents ordered by LSN but the _ts/creation time
+                    // of the documents may not be sequential. So when reading the change feed by LSN, it is possible to encounter
+                    // documents with lower _ts. In order to guarantee we always get the documents after the customer's point start
+                    // time, we need to always pass the start time in the header alongside the continuation token.
+                    if (this.startTime.HasValue
+                        && this.startTime.Value != DateTime.MinValue.ToUniversalTime()
+                        && this.changeFeedStartFrom is ChangeFeedStartFromContinuationAndFeedRange)
+                    {
+                        requestMessage.Headers.Add(
+                            HttpConstants.HttpHeaders.IfModifiedSince,
+                            this.startTime.Value.ToString("r", CultureInfo.InvariantCulture));
+                    }
 
                     if (this.changeFeedOptions.PageSizeHint.HasValue)
                     {
--- a/Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
@@ -14,10 +14,12 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Pagination
     {
         private const string TypePropertyName = "type";
         private const string ValuePropertyName = "value";
+        private const string StartTimePropertyName = "startTime";
 
         private const string BeginningTypeValue = "beginning";
         private const string TimeTypeValue = "time";
         private const string ContinuationTypeValue = "continuation";
+        private const string ContinuationAndStartTimeTypeValue = "continuationAndStartTime";
         private const string NowTypeValue = "now";
 
         public static TryCatch<ChangeFeedState> MonadicFromCosmosElement(CosmosElement cosmosElement)
@@ -81,7 +83,64 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Pagination
                                     $"expected change feed state to have a string value property: {cosmosElement}"));
                         }
 
-                        state = ChangeFeedState.Continuation(valuePropertyValue);
+                        // startTime is an optional field on a continuation token. When present, the state
+                        // also carries a start time so If-Modified-Since can be sent alongside the
+                        // continuation token (post-merge filtering). Keeping it optional on the existing
+                        // "continuation" type - rather than introducing a new type - preserves backward
+                        // compatibility: an older SDK that predates this feature simply ignores the
+                        // unknown startTime field instead of failing to parse the token.
+                        if (cosmosObject.TryGetValue(StartTimePropertyName, out CosmosString continuationStartTimePropertyValue))
+                        {
+                            if (!DateTime.TryParse(
+                                continuationStartTimePropertyValue.Value,
+                                CultureInfo.InvariantCulture,
+                                DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal | DateTimeStyles.AllowWhiteSpaces,
+                                out DateTime continuationStartTimeUtc))
+                            {
+                                return TryCatch<ChangeFeedState>.FromException(
+                                    new FormatException(
+                                        $"failed to parse start time value: {cosmosElement}"));
+                            }
+
+                            state = ChangeFeedState.ContinuationAndStartTime(valuePropertyValue, continuationStartTimeUtc);
+                        }
+                        else
+                        {
+                            state = ChangeFeedState.Continuation(valuePropertyValue);
+                        }
+                    }
+                    break;
+
+                case ContinuationAndStartTimeTypeValue:
+                    {
+                        // Legacy type kept for defensive backward compatibility. New tokens serialize as
+                        // the "continuation" type with an optional startTime field (see above).
+                        if (!cosmosObject.TryGetValue(ValuePropertyName, out CosmosString valuePropertyValue))
+                        {
+                            return TryCatch<ChangeFeedState>.FromException(
+                                new FormatException(
+                                    $"expected change feed state to have a string value property: {cosmosElement}"));
+                        }
+
+                        if (!cosmosObject.TryGetValue(StartTimePropertyName, out CosmosString startTimePropertyValue))
+                        {
+                            return TryCatch<ChangeFeedState>.FromException(
+                                new FormatException(
+                                    $"expected change feed state to have a string startTime property: {cosmosElement}"));
+                        }
+
+                        if (!DateTime.TryParse(
+                            startTimePropertyValue.Value,
+                            CultureInfo.InvariantCulture,
+                            DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal | DateTimeStyles.AllowWhiteSpaces,
+                            out DateTime startTimeUtc))
+                        {
+                            return TryCatch<ChangeFeedState>.FromException(
+                                new FormatException(
+                                    $"failed to parse start time value: {cosmosElement}"));
+                        }
+
+                        state = ChangeFeedState.ContinuationAndStartTime(valuePropertyValue, startTimeUtc);
                     }
                     break;
 
@@ -157,6 +216,25 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Pagination
                     {
                         { TypePropertyName, ContinuationTypeValueSingleton },
                         { ValuePropertyName, changeFeedStateContinuation.ContinuationToken }
+                    });
+            }
+
+            public CosmosElement Transform(ChangeFeedStateContinuationAndStartTime changeFeedStateContinuationAndStartTime)
+            {
+                // Serialize as the existing "continuation" type plus an optional startTime field so that
+                // older SDKs (which don't understand a dedicated continuationAndStartTime type) can still
+                // parse the token and simply ignore the extra startTime, rather than throwing.
+                return CosmosObject.Create(
+                    new Dictionary<string, CosmosElement>()
+                    {
+                        { TypePropertyName, ContinuationTypeValueSingleton },
+                        { ValuePropertyName, changeFeedStateContinuationAndStartTime.ContinuationToken },
+                        {
+                            StartTimePropertyName,
+                            CosmosString.Create(changeFeedStateContinuationAndStartTime.StartTime.ToString(
+                                "o",
+                                CultureInfo.InvariantCulture))
+                        }
                     });
             }
 
--- a/Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
@@ -89,10 +89,17 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Pagination
                         // "continuation" type - rather than introducing a new type - preserves backward
                         // compatibility: an older SDK that predates this feature simply ignores the
                         // unknown startTime field instead of failing to parse the token.
-                        if (cosmosObject.TryGetValue(StartTimePropertyName, out CosmosString continuationStartTimePropertyValue))
+                        if (cosmosObject.TryGetValue(StartTimePropertyName, out CosmosElement continuationStartTimePropertyValue))
                         {
+                            if (!(continuationStartTimePropertyValue is CosmosString continuationStartTimeString))
+                            {
+                                return TryCatch<ChangeFeedState>.FromException(
+                                    new FormatException(
+                                        $"expected change feed state to have a string startTime property: {cosmosElement}"));
+                            }
+
                             if (!DateTime.TryParse(
-                                continuationStartTimePropertyValue.Value,
+                                continuationStartTimeString.Value,
                                 CultureInfo.InvariantCulture,
                                 DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal | DateTimeStyles.AllowWhiteSpaces,
                                 out DateTime continuationStartTimeUtc))
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
@@ -126,7 +126,7 @@ namespace Microsoft.Azure.Cosmos
         /// </summary>
         /// <remarks>
         /// This is only used when:
-        /// (1) Lease store is not initialized and is ignored if a lease exists and has continuation token.
+        /// (1) Lease store is not initialized and is ignored if a lease exists and has a continuation token.
         /// (2) StartContinuation is not specified.
         /// (3) StartTime is not specified.
         /// </remarks>
@@ -144,14 +144,16 @@ namespace Microsoft.Azure.Cosmos
 
         /// <summary>
         /// Sets the time (exclusive) to start looking for changes after.
+        /// The start time is persisted in the lease document so that it is honored across processor restarts.
         /// </summary>
         /// <remarks>
-        /// This is only used when:
-        /// (1) Lease store is not initialized and is ignored if a lease exists and has continuation token.
-        /// (2) StartContinuation is not specified.
+        /// The start time is persisted in the lease and sent as the If-Modified-Since header alongside the
+        /// continuation token on subsequent change feed requests. This ensures documents are returned after
+        /// the specified time even after partition merges.
+        /// Passing <see cref="DateTime.MinValue"/> clears the persisted start time from the lease document.
         /// If this is specified, StartFromBeginning is ignored.
         /// </remarks>
-        /// <param name="startTime">Date and time when to start looking for changes.</param>
+        /// <param name="startTime">Date and time when to start looking for changes. Use <see cref="DateTime.MinValue"/> to clear a previously persisted start time.</param>
         /// <returns>The instance of <see cref="ChangeFeedProcessorBuilder"/> to use.</returns>
         public ChangeFeedProcessorBuilder WithStartTime(DateTime startTime)
         {
@@ -165,7 +167,8 @@ namespace Microsoft.Azure.Cosmos
                 throw new ArgumentNullException(nameof(startTime));
             }
 
-            this.changeFeedProcessorOptions.StartTime = startTime;
+            this.changeFeedProcessorOptions.StartTime = startTime == DateTime.MinValue ? null : startTime;
+            this.changeFeedProcessorOptions.IsStartTimeUserExplicit = true;
             return this;
         }
 
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
@@ -151,7 +151,9 @@ namespace Microsoft.Azure.Cosmos
         /// continuation token on subsequent change feed requests. This ensures documents are returned after
         /// the specified time even after partition merges.
         /// Passing <see cref="DateTime.MinValue"/> clears the persisted start time from the lease document.
-        /// If this is specified, StartFromBeginning is ignored.
+        /// After it is cleared, a lease without a continuation starts from the current time unless
+        /// <see cref="WithStartFromBeginning"/> is also configured.
+        /// A non-MinValue start time takes precedence over StartFromBeginning.
         /// </remarks>
         /// <param name="startTime">Date and time when to start looking for changes. Use <see cref="DateTime.MinValue"/> to clear a previously persisted start time.</param>
         /// <returns>The instance of <see cref="ChangeFeedProcessorBuilder"/> to use.</returns>
@@ -162,11 +164,7 @@ namespace Microsoft.Azure.Cosmos
                 throw new InvalidOperationException($"Using the '{nameof(WithStartTime)}' option with ChangeFeedProcessor is not supported with {ChangeFeedMode.AllVersionsAndDeletes} mode.");
             }
 
-            if (startTime == null)
-            {
-                throw new ArgumentNullException(nameof(startTime));
-            }
-
+            // DateTime.MinValue is the explicit opt-out for a start time persisted by an earlier run.
             this.changeFeedProcessorOptions.StartTime = startTime == DateTime.MinValue ? null : startTime;
             this.changeFeedProcessorOptions.IsStartTimeUserExplicit = true;
             return this;
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/Configuration/ChangeFeedProcessorOptions.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/Configuration/ChangeFeedProcessorOptions.cs
@@ -39,8 +39,9 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Configuration
         /// Gets or sets the start request continuation token to start looking for changes after.
         /// </summary>
         /// <remarks>
-        /// This is only used when lease store is not initialized and is ignored if a lease exists and has continuation token.
-        /// If this is specified, both StartTime and StartFromBeginning are ignored.
+        /// This is only used when lease store is not initialized and is ignored if a lease exists and has a continuation token.
+        /// If this is specified, StartFromBeginning is ignored. StartTime is ignored for the initial start position
+        /// but is still persisted in the lease and sent as If-Modified-Since alongside the continuation token.
         /// </remarks>
         /// <seealso cref="ChangeFeedOptions.RequestContinuation"/>
         public string StartContinuation { get; set; }
@@ -49,9 +50,12 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Configuration
         /// Gets or sets the time (exclusive) to start looking for changes after.
         /// </summary>
         /// <remarks>
-        /// This is only used when:
-        /// (1) Lease store is not initialized and is ignored if a lease exists and has continuation token.
-        /// (2) StartContinuation is not specified.
+        /// When set, the start time is persisted in the lease document and sent as the If-Modified-Since
+        /// header alongside the continuation token on subsequent change feed requests. This ensures
+        /// documents are returned after the specified time even after partition merges.
+        /// The start time is honored across processor restarts via persistence in the lease.
+        /// If StartContinuation is specified, this property is ignored for the initial start position
+        /// but is still persisted and sent alongside the continuation token.
         /// If this is specified, StartFromBeginning is ignored.
         /// </remarks>
         /// <seealso cref="ChangeFeedOptions.StartTime"/>
@@ -71,12 +75,18 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.Configuration
         }
 
         /// <summary>
+        /// Gets a value indicating whether the start time was explicitly set by the user
+        /// via the builder (as opposed to being auto-anchored by the processor on startup).
+        /// </summary>
+        public bool IsStartTimeUserExplicit { get; set; }
+
+        /// <summary>
         /// Gets or sets a value indicating whether change feed in the Azure Cosmos DB service should start from beginning (true) or from current (false).
         /// By default it's start from current (false).
         /// </summary>
         /// <remarks>
         /// This is only used when:
-        /// (1) Lease store is not initialized and is ignored if a lease exists and has continuation token.
+        /// (1) Lease store is not initialized and is ignored if a lease exists and has a continuation token.
         /// (2) StartContinuation is not specified.
         /// (3) StartTime is not specified.
         /// </remarks>
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
@@ -75,5 +75,12 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
         /// Gets or sets the ChangeFeedMode.
         /// </summary>
         public abstract string Mode { get; set; }
+
+        /// <summary>
+        /// Gets or sets the start time used for change feed requests.
+        /// When set, the If-Modified-Since header is sent alongside the continuation token
+        /// to ensure documents are returned after this time even after partition merges.
+        /// </summary>
+        public virtual DateTime? StartTime { get; set; }
     }
 }
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
@@ -81,6 +81,6 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
         /// When set, the If-Modified-Since header is sent alongside the continuation token
         /// to ensure documents are returned after this time even after partition merges.
         /// </summary>
-        public virtual DateTime? StartTime { get; set; }
+        public abstract DateTime? StartTime { get; set; }
     }
 }
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCore.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCore.cs
@@ -98,6 +98,9 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
         [JsonProperty("Mode", NullValueHandling = NullValueHandling.Ignore)]
         public override string Mode { get; set; }
 
+        [JsonProperty("StartTime", NullValueHandling = NullValueHandling.Ignore)]
+        public override DateTime? StartTime { get; set; }
+
         public override string ToString()
         {
             return string.Format(
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
@@ -106,6 +106,14 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                     }
                     serverLease.Owner = this.options.HostName;
                     serverLease.Properties = lease.Properties;
+
+                    // Persist user-explicit start time on the lease during acquisition
+                    // to ensure it survives restarts even for partitions that never checkpoint.
+                    if (this.options.IsStartTimeUserExplicit)
+                    {
+                        serverLease.StartTime = this.options.StartTime;
+                    }
+
                     return serverLease;
                 }).ConfigureAwait(false);
         }
--- a/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
+++ b/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
@@ -120,7 +120,12 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
 
         public override Task<DocumentServiceLease> CreateLeaseIfNotExistAsync(
             PartitionKeyRange partitionKeyRange, 
-            string continuationToken)
+            string continuationToken) => this.CreateLeaseIfNotExistAsync(partitionKeyRange, continuationToken, startTime: null);
+
+        public override Task<DocumentServiceLease> CreateLeaseIfNotExistAsync(
+            PartitionKeyRange partitionKeyRange,
+            string continuationToken,
+            DateTime? startTime)
         {
             if (partitionKeyRange == null)
             {
@@ -135,7 +140,8 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                 LeaseToken = leaseToken,
                 ContinuationToken = continuationToken,
                 FeedRange = new FeedRangeEpk(partitionKeyRange.ToRange()),
-                Mode = this.GetChangeFeedMode()
+                Mode = this.GetChangeFeedMode(),
+                StartTime = startTime,
             };
 
             this.requestOptionsFactory.AddPartitionKeyIfNeeded(
@@ -147,7 +153,12 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
 
         public override Task<DocumentServiceLease> CreateLeaseIfNotExistAsync(
             FeedRangeEpk feedRange,
-            string continuationToken)
+            string continuationToken) => this.CreateLeaseIfNotExistAsync(feedRange, continuationToken, startTime: null);
+
+        public override Task<DocumentServiceLease> CreateLeaseIfNotExistAsync(
+            FeedRangeEpk feedRange,
+            string continuationToken,
+            DateTime? startTime)
         {
             if (feedRange == null)
             {
@@ -162,7 +173,8 @@ namespace Microsoft.Azure.Cosmos.ChangeFeed.LeaseManagement
                 LeaseToken = leaseToken,
                 ContinuationToken = continuationToken,
                 FeedRange = feedRange,
-                Mode = this.GetChangeFeedMode()
+                Mode = this.GetChangeFeedMode(),
+                StartTime = startTime,
             };
 
             this.requestOptionsFactory.AddPartitionKeyIfNeeded(
--- a/Microsoft.Azure.Cosmos/src/Headers/SDKSupportedCapabilitiesHelpers.cs
+++ b/Microsoft.Azure.Cosmos/src/Headers/SDKSupportedCapabilitiesHelpers.cs
@@ -1,0 +1,16 @@
+//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    internal static class SDKSupportedCapabilitiesHelpers
+    {
+        public static ulong GetSDKSupportedCapabilities()
+        {
+            return (ulong)(Microsoft.Azure.Documents.SDKSupportedCapabilities.PartitionMerge
+                | Microsoft.Azure.Documents.SDKSupportedCapabilities.ChangeFeedWithStartTimePostMerge
+                | Microsoft.Azure.Documents.SDKSupportedCapabilities.IgnoreUnknownRntbdTokens);
+        }
+    }
+}
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests.ChangeFeed
     using Microsoft.Azure.Cosmos.Scripts;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
 
     [TestClass]
     [TestCategory("ChangeFeed")]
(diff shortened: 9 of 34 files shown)


xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Pagination/InMemoryContainer.cs
```diff
@@ -780,7 +780,19 @@ public Task<TryCatch<ChangeFeedPage>> MonadicChangeFeedAsync(
                         { "LSN", CosmosNumber64.Create(lastChange.LogicalSequenceNumber) }
                     });
 
-                ChangeFeedState responseState = ChangeFeedState.Continuation(continuationToken);
+                ChangeFeedState responseState;
+                if (feedRangeState.State is ChangeFeedStateTime changeFeedStateTime)
+                {
+                    responseState = ChangeFeedState.ContinuationAndStartTime(continuationToken, changeFeedStateTime.StartTime);
```
xinlian12, line 795: 🟡 **Recommendation — Test Fidelity: "not modified" path loses composite state**

The PR correctly updates the **success** path (lines 784–795) to preserve `ChangeFeedStateContinuationAndStartTime` across responses. However, the "not modified" path (line 765, not in the diff) still creates `new ChangeFeedStateTime(DateTime.UtcNow)` unconditionally — dropping both the continuation token and the original start time.

In production (`NetworkAttachedDocumentContainer.MonadicChangeFeedPageAsync`, line 368–399), `CreateChangeFeedStateFromResponse` is called for **both** `200 OK` and `304 NotModified` responses, correctly preserving the composite state on empty pages.

This means the `InMemoryContainer` mock diverges from production on "no changes → resume" scenarios. If a test starts with `ChangeFeedStateTime`, gets results (promoted to `ContinuationAndStartTime`), then hits an empty page, the mock resets to `ChangeFeedStateTime(DateTime.UtcNow)` — losing both the continuation position and the original start time. A subsequent write + resume would re-read from "now" rather than the continuation position.

**Suggestion**: Apply the same state-preservation logic to the "not modified" path at line 765:

```csharp
if (filteredChanges.Count == 0)
{
    ChangeFeedState notModifiedResponseState;
    if (feedRangeState.State is ChangeFeedStateContinuationAndStartTime composite)
    {
        notModifiedResponseState = composite; // preserve as-is
    }
    else
    {
        notModifiedResponseState = new ChangeFeedStateTime(DateTime.UtcNow);
    }
    // ...
}

```

Microsoft.Azure.Cosmos/src/Headers/SDKSupportedCapabilitiesHelpers.cs
```diff
@@ -0,0 +1,16 @@
...
+//------------------------------------------------------------
+
+namespace Microsoft.Azure.Cosmos
+{
+    internal static class SDKSupportedCapabilitiesHelpers
+    {
+        public static ulong GetSDKSupportedCapabilities()
+        {
+            return (ulong)(Microsoft.Azure.Documents.SDKSupportedCapabilities.PartitionMerge
+                | Microsoft.Azure.Documents.SDKSupportedCapabilities.ChangeFeedWithStartTimePostMerge
```
xinlian12, line 12: 🟡 **Recommendation — Behavioral Scope: `ChangeFeedProcessor` path drops start time after first response**

The `ChangeFeedWithStartTimePostMerge` capability flag is now advertised **globally** on every request (via `Headers.SDKSUPPORTEDCAPABILITIES`). However, the `ChangeFeedProcessor` code path uses `ChangeFeedPartitionKeyResultSetIteratorCore`, which at line 141 of that file replaces `changeFeedStartFrom` with `new ChangeFeedStartFromContinuationAndFeedRange(etag, feedRange)` after every response — dropping the original start time.

This means the processor path:
- **First request**: Sends `If-Modified-Since` ✅
- **Subsequent requests**: Sends only `If-None-Match` ❌ (start time lost)

The global capability flag tells the service "this SDK handles dual-header semantics after partition merges." But the processor path doesn't actually send both headers on subsequent requests, so the time-based safety net is absent.

For comparison, the **Java SDK explicitly opts the legacy processor out** by setting capabilities to `"0"` in `PartitionProcessorImpl.java`:

```java
// For pk version, merge is not support, exclude it from the capabilities header
changeFeedOptionsAccessor().setHeader(
    this.options,
    HttpConstants.HttpHeaders.SDK_SUPPORTED_CAPABILITIES,
    String.valueOf(HttpConstants.SDKSupportedCapabilities.SUPPORTED_CAPABILITIES_NONE));

```

**Consider either:**
1. Applying the same fix to `ChangeFeedPartitionKeyResultSetIteratorCore` (introduce a composite `ChangeFeedStartFrom` type that carries both etag and start time), or
2. Opting out the processor path by setting capabilities to `"0"` on processor requests (matching Java's approach)

This isn't a regression (the processor path was already broken for merged partitions + start time — the service threw `BadRequest`), but the global capability flag now silently allows weaker guarantees.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/GatewayTests.cs
```diff
@@ -2783,7 +2783,7 @@ private async Task ValidateChangeFeedIfModifiedSinceHelper(bool useGateway, Prot
                 options.StartTime = timestamps[1];
                 options.RequestContinuation = lsns[0].ToString();
                 accumulator = await this.ReadChangeFeedToEnd(client, coll.SelfLink, options);
-                Assert.AreEqual("doc2.", accumulator);
+                Assert.AreEqual(string.Empty, accumulator);
```
xinlian12, line 2786: 🟢 **Suggestion — Stale Comment: "If-None-Match wins" no longer describes the behavior**

The comment on line 2782 still reads `// 8. If-None-Match wins.` but the assertion now expects `string.Empty` — proving that `If-None-Match` no longer "wins" over `If-Modified-Since`. With the `ChangeFeedWithStartTimePostMerge` capability flag, the server now applies **both** filters: documents must pass both the LSN continuation filter AND the timestamp filter.

Since `timestamps[1]` is doc2's exact timestamp and `If-Modified-Since` uses "strictly after" semantics (confirmed by test #6 at line 2767 which returns 0 results for the same timestamp), doc2 is correctly filtered out.

**Suggestion**: Update the comment to reflect the new semantics:

```csharp
// 8. Both If-None-Match and If-Modified-Since apply — start time filters documents at the continuation position.

```

changelog.md
```diff
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### <a name="unreleased"/> Unreleased
 
 #### Features Added
-
+- [5931](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5931) ChangeFeed: Add support for StartTime for merged partitions
```
xinlian12, line 87: 🟢 **Suggestion — Changelog: Missing blank line and wording could be more customer-facing**

Two minor issues:

1. **Missing blank line**: There should be a blank line between this entry and `#### Breaking Changes` (compare with the format of earlier release sections).

2. **Wording**: The entry reads like a PR title. A more customer-facing version would be:

   ```
   - [5931](...) ChangeFeed: Adds StartTime support for merged partitions, enabling time-based change feed reads that previously returned a BadRequest error

   ```

Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
```diff
@@ -85,6 +87,37 @@ public static TryCatch<ChangeFeedState> MonadicFromCosmosElement(CosmosElement c
                     }
                     break;
 
+                case ContinuationAndStartTimeTypeValue:
```
xinlian12, line 121: 💬 **Observation — Forward Compatibility: New token type and SDK downgrades**

The new `"continuationAndStartTime"` serialization type is a one-way upgrade. Once a customer produces continuation tokens with this SDK version, older SDK versions will hit the `default: throw new InvalidOperationException()` at line 126 when trying to deserialize them.

This is a pre-existing pattern (the `default` case has always thrown), but this PR creates the **first new token type** that triggers it in practice. Any customer who:
1. Upgrades to this SDK version
2. Uses `ChangeFeedStartFrom.Time(...)` (which produces `ContinuationAndStartTime` tokens after the first response)
3. Persists the continuation token (e.g., in Change Feed Processor leases)
4. Rolls back to the previous SDK version

...will see the processor crash with `InvalidOperationException` until leases are manually reset.

Not a blocking concern — this is the accepted tradeoff for versioned serialization formats — but worth noting in the PR description or release notes for customer awareness.

xinlian12:
✅ **Review complete** (49:11)

Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
xinlian12: I think definitely worth calling out

Microsoft.Azure.Cosmos/src/Headers/SDKSupportedCapabilitiesHelpers.cs
xinlian12: I think we should opt-in the same fix as ChangeFeedPartitionKeyResultSetIteratorCore, please ignore java side (java sdk has two versions of lease document, and we explicitly callout for the old version, merge will not be support for clarify, this differentiation does not exist in .Net)

Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
ramarag: where do we call this out ?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/ChangeFeed/DynamicTests.cs
```diff
@@ -407,6 +407,84 @@ public async Task TestWithStartTime_CustomTime()
...
+
+                    string ifModifiedSince = subsequentRequest.Headers[Microsoft.Azure.Documents.HttpConstants.HttpHeaders.IfModifiedSince];
+                    Assert.IsNotNull(ifModifiedSince, "If-Modified-Since header should be set on the first change feed request with start time.");
+                    Assert.AreEqual(expectedIfModifiedSince, ifModifiedSince, "If-Modified-Since header value should match the start time.");
+
+                    string  sdkCapabilities = subsequentRequest.Headers[Microsoft.Azure.Documents.HttpConstants.HttpHeaders.SDKSupportedCapabilities];
+                    Assert.IsNotNull(sdkCapabilities, "SDKSupportedCapabilities header should be present on change feed requests.");
+                    ulong capabilitiesValue = ulong.Parse(sdkCapabilities);
+                    Assert.IsTrue(
+                        (capabilitiesValue & changeFeedWithStartTimePostMergeFlag) == changeFeedWithStartTimePostMergeFlag,
```
xinlian12: 🔴 **Blocking — Correctness: Undeclared variable `changeFeedWithStartTimePostMergeFlag`**

This variable is used here but never declared anywhere in `DynamicTests.cs` — no local variable, no class field, no base class field. This will cause a **CS0103 compilation error**, preventing the emulator test project from building.

The equivalent test in `ChangeFeedIteratorCoreTests.cs` correctly declares it at line 748:

```csharp
ulong changeFeedWithStartTimePostMergeFlag = (ulong)Microsoft.Azure.Documents.SDKSupportedCapabilities.ChangeFeedWithStartTimePostMerge;

```

**Fix:** Add the same declaration before line 473:

```csharp
ulong changeFeedWithStartTimePostMergeFlag = (ulong)Microsoft.Azure.Documents.SDKSupportedCapabilities.ChangeFeedWithStartTimePostMerge;

```

xinlian12:
✅ **Review complete** (40:02)

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
```diff
@@ -119,6 +123,19 @@ public override async Task<ResponseMessage> ReadNextAsync(ITrace trace, Cancella
                     ChangeFeedStartFromRequestOptionPopulator visitor = new ChangeFeedStartFromRequestOptionPopulator(requestMessage);
                     this.changeFeedStartFrom.Accept(visitor);
 
+                    // When a merge happens, the child partition will contain documents ordered by LSN but the _ts/creation time
+                    // of the documents may not be sequential. So when reading the change feed by LSN, it is possible to encounter
+                    // documents with lower _ts. In order to guarantee we always get the documents after the customer's point start
+                    // time, we need to always pass the start time in the header alongside the continuation token.
+                    if (this.startTime.HasValue
```
xinlian12, line 130: 🔴 **Blocking — Correctness: Auto-anchored StartTime causes data loss on processor restart with existing leases**

The new `IfModifiedSince` injection here is correct for **user-provided** start times (the feature this PR adds). However, it introduces a **regression** when `this.startTime` is the **auto-anchored** value from `ChangeFeedProcessorCore.StartAsync()` (PR #5617).

**Scenario:**
1. Processor runs, creates leases with continuation tokens, then stops.
2. 1 hour passes. 100 new documents are created (timestamps = T-60min to T-1min).
3. Processor restarts. `ChangeFeedProcessorCore.StartAsync()` auto-anchors `StartTime = DateTime.UtcNow.AddSeconds(-1)` (because `StartTime == null && StartContinuation is empty`).
4. `FeedProcessorFactoryCore.Create()` passes both `lease.ContinuationToken` (old LSN) and the auto-anchored `StartTime` (now − 1 sec).
5. This code sends both `If-None-Match: <old_LSN>` and `If-Modified-Since: <now − 1 sec>`.
6. The service applies the combined filter (enabled by `ChangeFeedWithStartTimePostMerge` capability). All 100 documents have timestamps **before** `now − 1 sec`, so they are **silently skipped**.
7. **Result: 100 documents permanently lost from the change feed.**

**Before this PR**, the auto-anchored time was harmlessly ignored — `IfModifiedSince` was never sent alongside continuation tokens. After this PR, it's sent on every request, and the service's combined filter actively removes documents.

**Root cause:** There's no way to distinguish auto-anchored `StartTime` from user-provided `StartTime` when it reaches `FeedProcessorFactoryCore.Create()`, because `ChangeFeedProcessorCore.StartAsync()` mutates `changeFeedProcessorOptions.StartTime` in place.

**Suggested fix direction:** Preserve the original user-provided `StartTime` before auto-anchoring. In `FeedProcessorFactoryCore.Create()`, only pass `StartTime` to the iterator when it was **user-provided** (not auto-anchored). When the user didn't set a `StartTime` and the lease has a continuation token, pass `startTime: null` so only the continuation token is used.

xinlian12:
✅ **Review complete** (49:28)

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
xinlian12, line 130: 🔴 **Blocking — Correctness: Auto-anchored `StartTime` causes silent data loss on processor restart**

When a `ChangeFeedProcessor` is built **without** an explicit `WithStartTime()`, `ChangeFeedProcessorCore.StartAsync()` auto-anchors `StartTime = UtcNow.AddSeconds(-1)` (line 99 of `ChangeFeedProcessorCore.cs`). This value is then passed through `FeedProcessorFactoryCore.Create()` → `options.StartTime` → `ChangeFeedPartitionKeyResultSetIteratorCore.Create(startTime: ...)` for **every** lease — including leases that already have a continuation token from a prior run.

Before this PR, the auto-anchored `StartTime` was harmless: when resuming from a continuation, `changeFeedStartFrom` was set to `ChangeFeedStartFromContinuationAndFeedRange`, and `startTime` was unused for subsequent requests.

After this PR, this new `if`-block sends `IfModifiedSince` on **every** continuation-based request. Combined with the `ChangeFeedWithStartTimePostMerge` capability flag that changes service semantics to apply **both** headers conjunctively (confirmed by the GatewayTests case 8 change: `"doc2."` → `string.Empty`), this creates a data-loss scenario:

**Concrete failure sequence:**
1. Processor (default config, no `WithStartTime()`) starts → auto-anchors `StartTime = T₁ − 1s`
2. Runs for hours, checkpoints leases with etag-based continuations
3. Stops. Documents continue to arrive.
4. Restarts → auto-anchors `StartTime = T₂ − 1s` (where T₂ ≫ T₁)
5. Factory creates iterator with `continuationToken = old_etag` AND `startTime = T₂ − 1s`
6. `ReadNextAsync()` sends `IfNoneMatch: old_etag` AND `IfModifiedSince: T₂ − 1s`
7. Server applies both filters conjunctively → **all documents created between the last checkpoint and T₂ − 1s are silently dropped**

This is the default configuration for most processors (`shouldAnchorStartTime` fires when `StartFromBeginning == false && StartTime == null && StartContinuation == null && Mode != AVAD`), so the data loss window equals the entire downtime period.

**Suggested fix** — in `Create()`, only carry forward `startTime` when no continuation token exists (i.e., only when `startTime` was the actual start mechanism):

```csharp
return new ChangeFeedPartitionKeyResultSetIteratorCore(
    container: container,
    mode: mode,
    changeFeedStartFrom: startFrom,
    options: requestOptions,
    startTime: continuationToken != null ? null : startTime);

```

This preserves the PR's intent for the first-run case (where `startTime` produces `ChangeFeedStartFromTime`, gets an etag, and subsequent requests correctly send both headers). It only suppresses the side-channel injection when resuming from an existing lease continuation.

An alternative is to mark the auto-anchored `StartTime` as non-forwardable (e.g., add an `IsAutoAnchored` flag to `ChangeFeedProcessorOptions`), but that's a larger change.

xinlian12:
✅ **Review complete** (47:51)

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
ramarag: duplicate of https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5931/#discussion_r3406388886

Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
kushagraThapar: Can't we maintain this token for backward compatibility like Java does with only optional fields?
If we cannot for some reason, we should call this out in our changelog in this PR that continuation token with merge capabilities won't be supported on older SDK versions before this.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
```diff
@@ -165,7 +167,8 @@ public ChangeFeedProcessorBuilder WithStartTime(DateTime startTime)
                 throw new ArgumentNullException(nameof(startTime));
             }
 
-            this.changeFeedProcessorOptions.StartTime = startTime;
+            this.changeFeedProcessorOptions.StartTime = startTime == DateTime.MinValue ? null : startTime;
```
kushagraThapar, line 168: Is this intentional? Since this is changing the behavior of .NET SDK from begining to now null, which effectivly means FromNow.

ramarag: there is another option "StartFromBeginning" which would supersede the start time.
By keeping this option, we can remove the startime from the lease documents and suspend sending the header to backend, it can be used for any mitigations we can think of

xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
```diff
@@ -165,7 +167,8 @@ public ChangeFeedProcessorBuilder WithStartTime(DateTime startTime)
...
+            this.changeFeedProcessorOptions.IsStartTimeUserExplicit = true;
```
xinlian12, line 169: 🔴 **Blocking — Correctness: `WithStartTime(DateTime.MinValue)` regresses PR #5617 write-loss protection on cold start**

Building on the existing thread here between @kushagraThapar and @ramarag — treating `DateTime.MinValue` as an escape hatch is defensible, but this line unconditionally also sets `IsStartTimeUserExplicit = true` (next line), which has an unintended downstream consequence: it disables the `now-1s` anchor added in PR #5617.

**Concrete failure sequence:**
1. Fresh deployment (empty lease container). User calls `.WithStartTime(DateTime.MinValue)` to explicitly opt out of a persisted start time (per the new XML doc).
2. `WithStartTime` sets `StartTime = null` **and** `IsStartTimeUserExplicit = true`.
3. `ChangeFeedProcessorCore.StartAsync` line 93–98: `shouldAnchorStartTime = ... && StartTime == null && !IsStartTimeUserExplicit && ...` → **false** because `IsStartTimeUserExplicit` is true.
4. Anchor is skipped. On a cold-start lease with no persisted `StartTime` and no `StartContinuation`, `FeedProcessorFactoryCore.Create` computes `effectiveStartTime = null`, and `ChangeFeedPartitionKeyResultSetIteratorCore.Create` falls through to `ChangeFeedStartFrom.Now(feedRange)` (`IfNoneMatch: *`).
5. **Writes performed during async lease bootstrap between `StartAsync` and the first successful acquire are silently missed** — the exact regression PR #5617 was written to fix.

Also worth documenting (a separate but related concern in the same XML doc): once `WithStartTime(t)` was used in a prior deployment, dropping the call in a later deployment is a silent no-op — the persisted `lease.StartTime = t` continues to be sent as `If-Modified-Since` forever (via `FeedProcessorFactoryCore.Create`'s `effectiveStartTime = lease.StartTime ?? …`) even though the current build no longer requests it. `WithStartTime(DateTime.MinValue)` is the only way to opt out, but that trips the write-loss regression above. Users deserve to know this in the remarks.

**Minimal safe fix:** decouple the two concerns. Either (a) split into two flags — one for "user wants to clear the persisted lease value" (drives `AcquireAsync` clearing) and one for "user provided a non-null value" (drives the anchor skip); or (b) simplest — drop `&& !IsStartTimeUserExplicit` from `shouldAnchorStartTime` so the anchor still runs when `StartTime == null`. Option (b) preserves the "clear the lease" semantic via `AcquireAsync`'s explicit branch without disabling the anti-loss protection.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
```diff
@@ -106,6 +106,14 @@ public override async Task<DocumentServiceLease> AcquireAsync(DocumentServiceLea
                     }
                     serverLease.Owner = this.options.HostName;
                     serverLease.Properties = lease.Properties;
+
+                    // Persist user-explicit start time on the lease during acquisition
+                    // to ensure it survives restarts even for partitions that never checkpoint.
+                    if (this.options.IsStartTimeUserExplicit)
+                    {
+                        serverLease.StartTime = this.options.StartTime;
```
xinlian12, line 114: 🔴 **Blocking — Correctness: Mixed `WithStartTime` values across CFP hosts sharing a lease container cause silent data loss on rebalance**

`AcquireAsync` unconditionally overwrites `serverLease.StartTime` with the acquiring host's `options.StartTime` whenever `IsStartTimeUserExplicit` is true. Rebalance/takeover is the norm in CFP, so this triggers under a plausible misconfiguration.

**Failure sequence:**
- Host A: `.WithStartTime(t1)`. Host B: `.WithStartTime(t2)` with `t2 > t1` (e.g., during a rolling deployment that changes the value).
- Lease X progresses under A up to LSN `L`; persisted `StartTime = t1`.
- Load balancer rebalances X → B. B's `AcquireAsync` overwrites `serverLease.StartTime = t2`.
- Next iterator poll from B sends `If-None-Match: L` **and** `If-Modified-Since: t2`.
- With the new `ChangeFeedWithStartTimePostMerge` capability the service filters conjunctively (as the `GatewayTests.cs` change from `"doc2."` → `""` proves). Any document with `LSN > L` **AND** `t1 ≤ _ts < t2` — produced under A but not yet received before rebalance — is **silently dropped**.

The symmetric case (Host A explicit, Host B not explicit) is safe because B's `AcquireAsync` preserves the persisted value. But mixed *explicit* values are not. There is no warning log, no exception, and no XML-doc caveat on `WithStartTime`. During any rolling deployment that changes the `WithStartTime(...)` argument, this is a real data-loss window on rebalance.

**Suggested fixes** (pick at least one):
1. In `AcquireAsync`, if `serverLease.StartTime.HasValue && serverLease.StartTime != this.options.StartTime`, at minimum trace a prominent warning; ideally throw a `LeaseLostException`-style configuration exception so mixed deployments fail loudly instead of dropping documents.
2. Prominently document on `WithStartTime` XML doc: "All processor instances sharing this lease container must supply the identical `WithStartTime` value (or all omit the call); mismatches can drop documents on lease rebalance."
3. Consider moving `StartTime` stamping from per-`AcquireAsync` to the one-time `.info`/bootstrap step so the first host wins.

Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
```diff
@@ -81,7 +83,64 @@ public static TryCatch<ChangeFeedState> MonadicFromCosmosElement(CosmosElement c
...
                         }
 
-                        state = ChangeFeedState.Continuation(valuePropertyValue);
+                        // startTime is an optional field on a continuation token. When present, the state
+                        // also carries a start time so If-Modified-Since can be sent alongside the
+                        // continuation token (post-merge filtering). Keeping it optional on the existing
+                        // "continuation" type - rather than introducing a new type - preserves backward
+                        // compatibility: an older SDK that predates this feature simply ignores the
+                        // unknown startTime field instead of failing to parse the token.
+                        if (cosmosObject.TryGetValue(StartTimePropertyName, out CosmosString continuationStartTimePropertyValue))
```
xinlian12: 🟡 **Recommendation — Correctness: Silent loss of `startTime` when the field is present but not a string**

`CosmosObject.TryGetValue<CosmosString>` returns `false` in **two** distinct situations: (a) the key is absent, and (b) the key is present but the value is not a `CosmosString` (JSON number, boolean, `null`, object, etc.). The current branch treats both identically — falling through to plain `ChangeFeedState.Continuation` on line 109.

Case (a) is the intended backward-compat path (older token without `startTime`). Case (b) is a malformed token, and silently discarding the `startTime` field there defeats the entire purpose of this PR: the reader falls back to a plain continuation, `If-Modified-Since` is never sent on subsequent requests, and after a partition merge the customer will again see documents predating their original `StartTime`.

Contrast the sibling `case ContinuationAndStartTimeTypeValue` block below (lines 125–140): it validates strictly and returns `FormatException` if `startTime` is missing or the wrong type. The two branches should behave symmetrically.

**Suggested fix:** Two-step check — `cosmosObject.TryGetValue(StartTimePropertyName, out CosmosElement raw)` first; then require `raw is CosmosString` and return `FormatException` when present-but-wrong-type. Mirror the strict validation used in the legacy branch. The new negative test `Continuation_TypeWithoutStartTime_ParsesAsPlainContinuation` covers case (a); a companion negative test for case (b) (e.g., `{"type":"continuation","value":"…","startTime": 12345}`) would lock this down.

Practical exposure is low today (the writer always emits a string), but the *purpose* of the asymmetric read path is defensive — this leaves a hole in the very branch that carries the new field.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCore.cs
```diff
@@ -98,6 +98,9 @@ public override DateTime Timestamp
         [JsonProperty("Mode", NullValueHandling = NullValueHandling.Ignore)]
         public override string Mode { get; set; }
 
+        [JsonProperty("StartTime", NullValueHandling = NullValueHandling.Ignore)]
+        public override DateTime? StartTime { get; set; }
```
xinlian12, line 102: 🟡 **Recommendation — Compat: Rolling upgrade with an older SDK silently strips this `StartTime` field on any write**

`NullValueHandling.Ignore` correctly makes old-SDK **reads** tolerant of the new field (Newtonsoft ignores unknown properties by default). But `DocumentServiceLeaseUpdaterCosmos.TryReplaceLeaseAsync` uses full-document `ReplaceItemAsync`, so any **write** from an older SDK version (checkpoint, acquire, release) re-serializes the lease through the older class definition — which has no `StartTime` property — and the field disappears from the persisted document.

**Practical rolling-upgrade impact:** New SDK on Host N1, old SDK on Host N2, sharing the lease container. (1) N1 acquires and stamps `StartTime = t`. (2) Load balancer moves the lease to N2. (3) N2 checkpoints → the `StartTime` field is gone. (4) LB moves back to N1; `AcquireAsync` re-stamps `t` if `IsStartTimeUserExplicit` is true. Between (3) and (4), merge-safety is lost for that lease. If the customer also runs the mixed-configuration failure mode described in my comment on `DocumentServiceLeaseManagerCosmos.cs` (WithStartTime set on N1 only), some documents can be silently missed.

**Suggested action:** at minimum, document this in the PR description / changelog / XML doc for `WithStartTime` — "during a rolling upgrade with SDK versions older than this release on the same lease container, the persisted `StartTime` will be transiently stripped by the older writer and re-stamped by the newer writer on the next acquire." Optionally, if the older SDK's lease class captures unknown properties into `[JsonExtensionData]`, an `[OnSerialize]`/`[OnDeserialized]` hook could round-trip the field, but that would only help if we ship it in the older SDK too.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/FeedToken/ChangeFeedIteratorCoreTests.cs
```diff
@@ -698,6 +698,98 @@ public async Task ChangeFeedIteratorCore_DeleteAfterCreate()
...
+                        Assert.IsTrue(headerHandler.CapturedRequests.Count > 0, "Expected at least one request in the second read.");
+
+                        RequestMessage capturedRequest = headerHandler.CapturedRequests[0];
+                        ifNoneMatch = capturedRequest.Headers.IfNoneMatch;
+                        Assert.IsNotNull(ifNoneMatch, "If-None-Match (etag) header should be set on subsequent change feed requests as the continuation.");
+                        Assert.AreNotEqual("*", ifNoneMatch, "If-None-Match should be a specific etag, not '*'.");
+
+                        // The If-Modified-Since should still be present on subsequent requests (for merge handling)
+                        ifModifiedSince = capturedRequest.Headers[Microsoft.Azure.Documents.HttpConstants.HttpHeaders.IfModifiedSince];
+                        Assert.IsNotNull(ifModifiedSince, "If-Modified-Since should still be present on subsequent requests for start time based change feed.");
```
xinlian12: 🟡 **Recommendation — Test Coverage: User-facing continuation-token resume for time-based feeds is not proven end-to-end**

This test runs a single continuous iterator and breaks the moment it sees `NotModified`. It never (a) captures `feedIterator.FeedRangeContinuation` after the first page, (b) constructs a *fresh* iterator from that continuation, or (c) asserts the resumed iterator still emits `If-Modified-Since` alongside `If-None-Match`.

That resume-from-persisted-continuation path is precisely the scenario `ChangeFeedStateContinuationAndStartTime` was designed for. The serializer tests prove the type round-trips through JSON, and `NetworkAttachedDocumentContainer.CreateChangeFeedStateFromResponse` was updated to promote `ChangeFeedStateTime` → composite state, but nothing wires "produce continuation → serialize → deserialize → resume → send request" together end-to-end. A regression that reverts the promotion to plain `ChangeFeedState.Continuation`, or that stops serializing `startTime` on the token, would break the actual customer scenario without failing any test.

Additionally, the loop breaks on the first `NotModified`, so the test doesn't prove the composite state is preserved *through* a `NotModified` transition either (both the `OK` and `NotModified` paths in `NetworkAttachedDocumentContainer.MonadicChangeFeedAsync` call the promotion helper today, but no assertion locks that invariant).

**Suggested addition:** after the first page, capture `feedIterator.FeedRangeContinuation`, dispose the iterator, construct a new one via `itemsCore.GetChangeFeedStreamIterator(ChangeFeedStartFrom.ContinuationToken(continuation), ChangeFeedMode.Incremental)`, read one page, and assert both `If-Modified-Since` (with the original start time) and `If-None-Match` are on the outgoing request. Additionally, after receiving `NotModified`, insert one document and issue one more `ReadNextAsync` to lock the NotModified-preservation invariant.

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
```diff
@@ -119,6 +123,19 @@ public override async Task<ResponseMessage> ReadNextAsync(ITrace trace, Cancella
...
+                        && this.startTime.Value != DateTime.MinValue.ToUniversalTime()
```
xinlian12, line 131: 🟢 **Suggestion — Robustness: `DateTime.MinValue.ToUniversalTime()` sentinel is machine-timezone-dependent**

`DateTime.MinValue.Kind == Unspecified`, so `ToUniversalTime()` treats it as Local. On a PST/negative-offset host it returns `Ticks = 288_000_000_000` (`MinValue + offset`); on UTC or positive-offset hosts it clamps to `Ticks = 0`. Since `DateTime.operator!=` compares only ticks (ignores `Kind`), the sentinel value differs across machines.

Not a live bug today because the CFP `WithStartTime(DateTime.MinValue)` path in `ChangeFeedProcessorBuilder` coerces to `null` before this iterator sees it — so this branch is effectively defensive dead code for the CFP scenario. But if any future caller ever hands the iterator a real `DateTime.MinValue` (any `Kind`), the iterator would emit `If-Modified-Since: Mon, 01 Jan 0001 00:00:00 GMT` on UTC/Tokyo machines while silently skipping it on PST — a Heisenbug. The same pattern is reused in `NetworkAttachedDocumentContainer.ChangeFeedStateRequestMessagePopulator.StartFromBeginningTime` (unchanged in this PR but now called from the new `Visit(ChangeFeedStateContinuationAndStartTime, ...)` at line 469), which enlarges the blast radius.

**Suggested fix:** use a Kind-explicit constant like `new DateTime(0, DateTimeKind.Utc)` or `DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc)` for the sentinel comparison in both files.

Separately (also low-priority): the iterator's `startTime` field is not validated for `Kind == Utc` on ingress. `ChangeFeedStateTime`, `ChangeFeedStateContinuationAndStartTime`, `ChangeFeedStartFromTime`, and `ChangeFeedProcessorOptions.StartTime` all enforce this invariant; the iterator does not, and `.ToString("r", CultureInfo.InvariantCulture)` writes the literal suffix `"GMT"` without converting — so a non-UTC value would produce a wrong header. Consider enforcing `Kind == Utc` (or normalizing) in the iterator constructor for parity with the rest of the surface.

xinlian12:
✅ **Review complete** (56:20)

Posted **6** inline comment(s).

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
xinlian12: the Anchor concern worth double checking

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
xinlian12: I think have a warning log when have mix configure is a good one to have, but how to real blocking mixed config is something we can follow up/tackle later

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseCore.cs
xinlian12: this is a valid concern, but should only happen during deployment time.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
```diff
@@ -106,6 +106,14 @@ public override async Task<DocumentServiceLease> AcquireAsync(DocumentServiceLea
                     }
                     serverLease.Owner = this.options.HostName;
                     serverLease.Properties = lease.Properties;
+
+                    // Persist user-explicit start time on the lease during acquisition
+                    // to ensure it survives restarts even for partitions that never checkpoint.
+                    if (this.options.IsStartTimeUserExplicit)
```
xinlian12, line 112: 🟡 **Recommendation — Correctness: Split/merge child leases lose StartTime across restart**

`AcquireAsync` writes `serverLease.StartTime` only when `this.options.IsStartTimeUserExplicit` is true. But `CreateLeaseIfNotExistAsync` (both overloads, ~lines 121 and 148) never sets `StartTime` on newly-created child leases at all — it constructs `DocumentServiceLeaseCore { LeaseId, LeaseToken, ContinuationToken, FeedRange, Mode }` with `StartTime = null`. Combined with `PartitionSynchronizerCore.HandlePartitionGoneAsync` (which propagates only `lastContinuationToken`, not `StartTime`, to children), this creates a silent data-loss window on the exact scenario this PR is designed to protect.

**Concrete non-crash scenario:**
1. Session A: user calls `WithStartTime(T)`, runs processing, checkpoints. Parent lease persisted with `ContinuationToken = X` and `StartTime = T`.
2. Session B: user restarts **without** calling `WithStartTime` (relies on lease persistence — this is the explicit contract in the new `ChangeFeedProcessorOptions.StartTime` remarks). `IsStartTimeUserExplicit = false`.
3. `AcquireAsync` on parent lease is a no-op for `StartTime` (correct — persisted value already there).
4. Partition split occurs. `CreateLeaseIfNotExistAsync(childRange, X)` writes child lease with `ContinuationToken = X` **and `StartTime = null`**.
5. `AddOrUpdateLeaseAsync` → `AcquireAsync` for the child. `IsStartTimeUserExplicit == false` → the `if` block above skips the write. Child lease's persisted `StartTime` stays null.
6. `FeedProcessorFactoryCore.Create`: `effectiveStartTime = lease.StartTime ?? (empty(continuation) ? options.StartTime : null)` → `null` (child has continuation, lease.StartTime is null).
7. Iterator sends `If-None-Match: X` **without** `If-Modified-Since` on the merged partition → documents with `_ts < T` are returned to the observer. **This is the exact bug the PR claims to fix.**

The AcquireAsync + FeedProcessorFactoryCore layers already handle the standard case, but a lease created *during a session in which the user is not explicit* never gets its parent's `StartTime` propagated.

**Suggested fix:** Have `CreateLeaseIfNotExistAsync` accept (or read from `this.options`) the parent's/persisted `StartTime` and set it on the new lease before `TryCreateDocumentServiceLeaseAsync` — or, in `PartitionControllerCore.HandlePartitionGoneAsync` (line ~192, where `l.Properties = lease.Properties;` already inherits from parent), also copy `l.StartTime = lease.StartTime;` before the addLeaseTasks are awaited. Add a unit test to `PartitionSynchronizerCoreTests` covering this propagation.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/Configuration/ChangeFeedProcessorOptions.cs
```diff
@@ -70,13 +74,19 @@ public DateTime? StartTime
             }
         }
 
+        /// <summary>
+        /// Gets a value indicating whether the start time was explicitly set by the user
+        /// via the builder (as opposed to being auto-anchored by the processor on startup).
+        /// </summary>
+        public bool IsStartTimeUserExplicit { get; set; }
```
xinlian12, line 81: 🟡 **Recommendation — Design: `StartTime` field now has two meanings; `IsStartTimeUserExplicit` is the classic symptom**

`ChangeFeedProcessorOptions.StartTime` has grown two distinct meanings:
1. **User intent** — the customer-supplied value the SDK must persist on the lease and echo via `If-Modified-Since` on every subsequent request.
2. **Auto-anchor value** — set by `ChangeFeedProcessorCore.StartAsync` to `UtcNow.AddSeconds(-1)` (PR #5617), which must **not** be persisted or echoed alongside a continuation (else PR #5846 regresses).

`IsStartTimeUserExplicit` is a companion boolean threaded through five layers (options → builder → processor core → lease store manager options → lease manager → factory) whose only job is to disambiguate meaning (1) from meaning (2) after the fact. Any future code that reads `options.StartTime` must remember to also consult this flag — the four new regression tests in `ChangeFeedProcessorCoreTests` document exactly that trap.

**Alternative that removes the flag entirely:**
- Keep `StartTime` as the customer-owned field (only ever mutated by the builder).
- Replace the in-place mutation in `StartAsync` with either a separate `internal DateTime? AnchoredStartTime { get; set; }` on the options, or (better) a local variable used only to compute the initial cold-start `ChangeFeedStartFrom.Time(...)`.

Then:
- `shouldAnchorStartTime` becomes `StartTime == null && StartContinuation == null && !StartFromBeginning && Mode != AVAD` — no flag needed.
- `FeedProcessorFactoryCore` writes `lease.StartTime = options.StartTime` if the customer provided one — no flag needed.
- `AcquireAsync` writes `serverLease.StartTime = options.StartTime` — no flag needed; `null` cleanly means "clear".
- The anchor is a cold-start-only local, never persisted, never confused with intent.

Not blocking — the current design is correct — but the plumbing burden grows each time `StartTime` needs to reach a new layer, and forgetting the flag re-introduces the PR #5846 regression silently.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
```diff
@@ -75,5 +75,12 @@ internal abstract class DocumentServiceLease
         /// Gets or sets the ChangeFeedMode.
         /// </summary>
         public abstract string Mode { get; set; }
+
+        /// <summary>
+        /// Gets or sets the start time used for change feed requests.
+        /// When set, the If-Modified-Since header is sent alongside the continuation token
+        /// to ensure documents are returned after this time even after partition merges.
+        /// </summary>
+        public virtual DateTime? StartTime { get; set; }
```
xinlian12: 🟢 **Suggestion — Consistency: `StartTime` should be `abstract` to match sibling `Mode`**

Every sibling persisted property on `DocumentServiceLease` (`Owner`, `ContinuationToken`, `Mode`, `Timestamp`, etc.) is `abstract`, which forces every concrete subclass to explicitly override — and therefore explicitly annotate with `[JsonProperty]` for serialization. `StartTime` alone is `virtual` with an auto-property and no `[JsonProperty]` at the base.

`DocumentServiceLeaseCore` and `DocumentServiceLeaseCoreEpk` both correctly override with `[JsonProperty("StartTime", NullValueHandling = NullValueHandling.Ignore)]`, so this is not a bug today. However:

- A future lease subclass (or a mock lease used in a test) that inherits the base auto-property will silently **not persist StartTime**, because the base has no `[JsonProperty]`. The failure mode surfaces only during a partition merge in production and looks like "StartTime got lost across restart."
- `Mode` (added in a prior PR for the same reason of durable persistence) uses `abstract` for exactly this forcing-function reason.

**Suggested change:** `public abstract DateTime? StartTime { get; set; }`. Both existing overrides already exist, so the mechanical change is minimal. Any future subclass is forced to make an explicit choice.

Microsoft.Azure.Cosmos/src/ChangeFeed/ChangeFeedPartitionKeyResultSetIteratorCore.cs
xinlian12, line 131: 🟢 **Suggestion — Consistency: `DateTime.MinValue.ToUniversalTime()` sentinel is duplicated and machine-timezone-dependent**

The same "beginning" sentinel is now materialized in three places:

- `ChangeFeed/ChangeFeedStartFromRequestOptionPopulator.cs:14` — `private static readonly DateTime StartFromBeginningTime = DateTime.MinValue.ToUniversalTime();`
- `Pagination/NetworkAttachedDocumentContainer.cs` — same declaration (referenced from the new `Visit(ChangeFeedStateContinuationAndStartTime, ...)` at ~line 469).
- Here, **evaluated inline on every request** with no named constant.

Two problems:

1. **Duplication:** if the sentinel definition ever changes (e.g., moves to `DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc)` or a named constant), three sites must be updated in lockstep. A future grep for `StartFromBeginningTime` won't find this one.
2. **Timezone-dependence of the expression:** `DateTime.MinValue` has `Kind == Unspecified`; `.ToUniversalTime()` treats Unspecified as local, so on a non-UTC host the result is not `DateTime.MinValue` — it shifts by up to 14 hours. Today this is dead-defensive because `WithStartTime(DateTime.MinValue)` short-circuits to `null` before the setter, but a non-processor caller of `ChangeFeedPartitionKeyResultSetIteratorCore.Create` that passes `DateTime.MinValue.ToUniversalTime()` from another timezone will see this branch behave differently across hosts.

**Suggested fix:** Reuse the existing `ChangeFeedStateRequestMessagePopulator.StartFromBeginningTime` constant here (or extract a shared `internal static class ChangeFeedTimeSentinels`), and if a timezone-independent sentinel is intended, use `DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc)`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedStateCosmosElementSerializerTests.cs
```diff
@@ -66,5 +66,76 @@ public void Continuation()
...
+                    { "startTime", CosmosString.Create(new DateTime(2023, 1, 15, 10, 30, 0, DateTimeKind.Utc).ToString("o", System.Globalization.CultureInfo.InvariantCulture)) },
+                });
+
+            TryCatch<ChangeFeedState> monadicState = ChangeFeedStateCosmosElementSerializer.MonadicFromCosmosElement(legacyToken);
+            Assert.IsTrue(monadicState.Succeeded);
+            Assert.IsTrue(monadicState.Result is ChangeFeedStateContinuationAndStartTime);
+        }
+
+        [TestMethod]
+        public void Continuation_TypeWithoutStartTime_ParsesAsPlainContinuation()
```
xinlian12, line 126: 🟢 **Suggestion — Test Coverage: no negative-path test for `startTime` parse failure**

Two new `FormatException` branches were added to `ChangeFeedStateCosmosElementSerializer.MonadicFromCosmosElement` when `DateTime.TryParse` on the `startTime` field fails — one inside `case ContinuationTypeValue` (new-format) and one inside `case ContinuationAndStartTimeTypeValue` (legacy). Neither is exercised by the new tests.

A corrupted continuation token (client mismatch, encoding drift, or a version-transition edge case) would slip through unproven, and a future refactor could easily coalesce the parse into a `Convert.ToDateTime(...)` (which throws differently) without any test failing.

**Suggested addition** — two small tests mirroring `Continuation_TypeWithoutStartTime_ParsesAsPlainContinuation`:
- `type: "continuation"` with `startTime: "not-a-date"` → assert `monadicState.Succeeded == false` and `monadicState.Exception is FormatException`.
- Same for `type: "continuationAndStartTime"`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedProcessorCoreTests.cs
```diff
@@ -685,6 +687,142 @@ public async Task StopAsync_WhenShutdownAsyncThrows_ExceptionPropagates()
...
+                .GetField("resultSetIterator", BindingFlags.Instance | BindingFlags.NonPublic)
+                .GetValue(processor);
+
+            return (DateTime?)iterator.GetType()
+                .GetField("startTime", BindingFlags.Instance | BindingFlags.NonPublic)
+                .GetValue(iterator);
+        }
+
+        [TestMethod]
+        public void FeedProcessorFactory_PersistedLeaseStartTime_WithContinuationToken_ForwardsStartTimeToIterator()
```
xinlian12, line 796: 🟢 **Suggestion — Test Coverage: missing branch — persisted `lease.StartTime` with **no** continuation token**

The four new tests cover: (auto-anchor + continuation), (user-explicit + continuation), (auto-anchor + no continuation), and (persisted lease.StartTime + continuation). The natural remaining combination — **persisted `lease.StartTime` present AND `ContinuationToken` empty** — is not covered.

This is a real, reachable scenario: `AcquireAsync` explicitly persists `StartTime` even on leases that never checkpoint (see comment: *"to ensure it survives restarts even for partitions that never checkpoint"*). A partition may have been acquired but produced no writes, so the lease has `StartTime = T` but `ContinuationToken = null`. The current `effectiveStartTime = lease.StartTime ?? …` expression correctly forwards `T` — but nothing asserts it.

A future refactor collapsing the null-coalescing arms could silently drop `lease.StartTime` for exactly the "acquired-but-never-checkpointed" case the persistence code was designed to protect.

**Suggested addition:** A fifth test with `lease.ContinuationToken = null`, `lease.StartTime = someUserTime`, `IsStartTimeUserExplicit = false`, and assert `iteratorStartTime == someUserTime`.

xinlian12:
✅ **Review complete** (46:20)

Posted **6** inline comment(s).

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
ramarag: add comment

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
ramarag: add a test case for split/merge.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedProcessorBuilder.cs
ramarag: Addressed in 72394d7f6. The API remarks now explain that DateTime.MinValue clears the persisted start time, leases without continuations then start from now unless WithStartFromBeginning is configured, and non-MinValue times take precedence. The impossible DateTime null check was also removed.

Microsoft.Azure.Cosmos/src/ChangeFeed/Pagination/ChangeFeedStateCosmosElementSerializer.cs
ramarag: Addressed in 72394d7f6. A present startTime field is now required to be a CosmosString; malformed types return FormatException instead of silently degrading to a plain continuation. Added a regression test for the non-string case.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/FeedToken/ChangeFeedIteratorCoreTests.cs
ramarag: Addressed in 72394d7f6. The emulator test now captures and resumes from the serialized continuation token, verifies both If-None-Match and the original If-Modified-Since, then crosses a NotModified response and confirms both headers remain after a new document is inserted.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLeaseManagerCosmos.cs
ramarag: Addressed in 72394d7f6. Split and merge child lease creation now receives and persists the parent lease StartTime for both PK-range and EPK leases. Unit tests cover PK split, EPK split, PK merge, and Cosmos lease creation persistence.

Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/LeaseManagement/DocumentServiceLease.cs
ramarag: Addressed in 72394d7f6. DocumentServiceLease.StartTime is now abstract, matching Mode and forcing each concrete lease type to make persistence explicit.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedStateCosmosElementSerializerTests.cs
ramarag: Addressed in 72394d7f6. Added negative tests for invalid date strings in both continuation and legacy continuationAndStartTime token formats; both assert the wrapped FormatException.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ChangeFeed/ChangeFeedProcessorCoreTests.cs
ramarag: Addressed in 72394d7f6. Added coverage for a persisted lease StartTime with no continuation token and verified that the persisted value is forwarded as the initial iterator position.

PR #5937, opened 2026-06-09
[Internal] Authorization: Adds AAD TokenCredential test coverage and sample

NaluTripician:
## What

Closes specific coverage gaps in the .NET SDK's Azure AD (Microsoft Entra ID / `TokenCredential`) authentication stack, and adds a runnable AAD sample. **Test-only + sample — no production `src/**` change**, so no changelog entry is required (the diff stays under `tests/**` and `Samples/**`).

Produced via the Seon Agentic Workflow v2 (research → plan → dual specs → human-review gate → implementation).

## Why

The SDK already ships a mature AAD stack with real but incomplete test coverage. Several shipped behaviors had **no direct test**, so a regression would ship silently.

## Unit tests — `CosmosAuthorizationTests.cs` (G1–G8)
- Dispose-after-use → `ObjectDisposedException`
- 401/403 from the credential is **not** retried (data-driven; cached auth state cleared)
- Credential returning an already-expired token → `ArgumentOutOfRangeException`
- `OperationCanceledException` → `RequestTimeout` `CosmosException` with `SubStatusCode == FailedToGetAadToken`
- `GenerateAadAuthorizationSignature` format (`type=aad&ver=1.0&sig=<token>`)
- `AddAuthorizationHeaderAsync` injects the `Authorization` header
- `TimeSpan.MaxValue` disables the background refresh loop
- Scope fallback (`AADSTS500011`) is applied to the **next** token request

## Emulator tests — `CosmosAadTests.cs` (G9–G10)
- `TokenCredential` initialize-overloads (`CreateAndInitializeAsync` / `BuildAndInitializeAsync`) authenticate
- Mid-operation token refresh keeps subsequent operations working

## Sample — `Samples/Usage/EntraIdAuthentication/`
`CosmosClient` with `DefaultAzureCredential`, a custom `TokenCredentialBackgroundRefreshInterval`, and `CosmosClientBuilder`, plus an `AZURE_COSMOS_AAD_SCOPE_OVERRIDE` note. Construction-only (no live account required). Added to `Cosmos.Samples.Usage.sln`.

## Verification
- `Microsoft.Azure.Cosmos.Tests` build: 0 warning / 0 error
- New unit tests: **12 passed, 0 failed** (`dotnet test --filter`)
- `Microsoft.Azure.Cosmos.EmulatorTests` build: 0 / 0 (G9/G10 require the emulator to run)
- `EntraIdAuthentication` sample build: 0 error

## No changelog entry required
Test-only + sample; no change under any shipped-package `src/**` tree.


aavasthy:
Emulator does not support AAD, just a thought should we be adding end to end AAD tests against an actual account considering how AAD is being extensively used across.

NaluTripician:
@aavasthy good catch, and you're right. The emulator doesn't do real AAD — `LocalEmulatorTokenCredential` embeds the master key as the JWT signature and the emulator validates against that, so the emulator-based tests here (and the existing `AadMock*` ones) exercise the SDK's TokenCredential pipeline — caching, background refresh, 401/403 no-retry, cancellation mapping, scope selection + AADSTS500011 fallback, header format — but never a real Entra-issued token or data-plane RBAC.

So this PR intentionally covers the SDK-side plumbing (most of which is unit-testable without an account) plus mock end-to-end through a real client. What it does not cover is exactly what you're pointing at: a real STS accepting the scope, RBAC role/permission mapping, 403-on-missing-role, and real token claims/lifetime. That's a genuine gap, and not just in this PR — there's no real-AAD test anywhere in the repo today.

I'll do two things:
1. Rename the two new emulator tests to the `AadMock*` convention and add a comment so it's obvious they're mock-credential pipeline tests, not real AAD.
2. File a follow-up for a real end-to-end AAD test against a live RBAC-enabled account. That one needs infra we don't have yet — a live account with role assignments + a test identity (SP / managed identity / federated cred) wired into the functional/live pipeline (env-gated), not the emulator suite that runs on every PR — so I'd rather track it separately than bolt it onto this test-only change.

If there's already a live RBAC account + identity available in the functional pipeline that I can reuse, happy to add the real test as a fast-follow instead. Let me know what you'd prefer.

kushagraThapar (dismissed):
LGTM, thanks @NaluTripician 
@aavasthy can you please also review this PR, thanks!

aavasthy:
> @aavasthy good catch, and you're right. The emulator doesn't do real AAD — `LocalEmulatorTokenCredential` embeds the master key as the JWT signature and the emulator validates against that, so the emulator-based tests here (and the existing `AadMock*` ones) exercise the SDK's TokenCredential pipeline — caching, background refresh, 401/403 no-retry, cancellation mapping, scope selection + AADSTS500011 fallback, header format — but never a real Entra-issued token or data-plane RBAC.
> 
> So this PR intentionally covers the SDK-side plumbing (most of which is unit-testable without an account) plus mock end-to-end through a real client. What it does not cover is exactly what you're pointing at: a real STS accepting the scope, RBAC role/permission mapping, 403-on-missing-role, and real token claims/lifetime. That's a genuine gap, and not just in this PR — there's no real-AAD test anywhere in the repo today.
> 
> I'll do two things:
> 
> 1. Rename the two new emulator tests to the `AadMock*` convention and add a comment so it's obvious they're mock-credential pipeline tests, not real AAD.
> 2. File a follow-up for a real end-to-end AAD test against a live RBAC-enabled account. That one needs infra we don't have yet — a live account with role assignments + a test identity (SP / managed identity / federated cred) wired into the functional/live pipeline (env-gated), not the emulator suite that runs on every PR — so I'd rather track it separately than bolt it onto this test-only change.
> 
> If there's already a live RBAC account + identity available in the functional pipeline that I can reuse, happy to add the real test as a fast-follow instead. Let me know what you'd prefer.

yes that will be ideal as now all environments have AAD auth and we need continuous pipelines testing against an actual AAD account. We can have a follow up PR for it. Thanks!

aavasthy (dismissed):

Microsoft.Azure.Cosmos.Samples/Usage/EntraIdAuthentication/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+            string endpoint,
+            TokenCredential tokenCredential)
+        {
+            // <CreateClientWithCustomRefreshInterval>
+            CosmosClientOptions clientOptions = new CosmosClientOptions
+            {
+                TokenCredentialBackgroundRefreshInterval = TimeSpan.FromMinutes(30)
+            };
+
+            using CosmosClient client = new CosmosClient(endpoint, tokenCredential, clientOptions);
```
FabianMeiswinkel, line 216: What is the purpose of this sample? NO interaction with AAD is even executed - and except that it shows that you can pass TokenCredential when creating builder it does not help to understand any of the challenging aspects when using AAD (like metadata operations like create Collection not working anymore)? If we want to add a sample - perfect - but then let the sample have some more value that showing how to pass TokenCredentials around?

FabianMeiswinkel (commented):
LGTM except for the sample - please remove that one or make it more beefy/useful.

Microsoft.Azure.Cosmos.Samples/Usage/EntraIdAuthentication/Program.cs
NaluTripician: Good point - thanks. I reworked the sample so it actually does something meaningful rather than just passing a `TokenCredential` around (23786ba):

- **Real data-plane operations** over an AAD token: item create / read / query / replace / delete against an existing container.
- **Demonstrates the main AAD pitfall** you called out: it attempts `CreateDatabaseIfNotExistsAsync` (a control-plane / metadata operation) and shows it is rejected with `403 Forbidden` under a data-plane RBAC role, with a comment explaining that databases/containers must be provisioned via ARM / Bicep / portal / CLI - the SDK token is data-plane only.
- Prerequisites now spell this out (incl. `az cosmosdb sql database/container create` commands) and the sample refuses to run against the emulator endpoint, since it needs a real RBAC-enabled account to issue/validate Entra tokens.

The construction-pattern helpers (custom background refresh interval, `CosmosClientBuilder`, scope override) are still there but are now secondary to the data/control-plane demonstration.

FabianMeiswinkel (approved):
LGTM

ananth7592 (approved):

PR #5939, opened 2026-06-09
[Internal] FaultInjection Tests: Fixes unreliable metadata response-delay assertions

NaluTripician:
## Description

While running the fault injection test suite against a live multi-region account, three metadata response-delay tests were consistently failing. Investigation showed this is a **test-design issue, not a fault injection library bug** — the library injects the delay correctly every time (the rule hit count increments), but the test assertions were based on an invalid assumption.

### Root cause

The tests asserted that an injected metadata-refresh delay (e.g. 15s) would manifest as added wall-clock latency on a subsequent data operation:

```csharp
Assert.IsTrue(elapsed.TotalSeconds >= 15);
```

Operation diagnostics show what actually happens: the injected delay causes the metadata HTTP call to hit its **request timeout** (`TaskCanceledException`), after which the SDK **transparently retries** the metadata request and succeeds:

```
"Read Collection": DurationInMs:1034, ExceptionType:"TaskCanceledException"   ← injected delay hit the request timeout
                   DurationInMs:212,  StatusCode:"OK"                         ← SDK retried and succeeded
```

So the full delay is never observed end-to-end, and the operation completes in ~1–4s regardless of the injected delay or account topology. (In `PKRangeResponseDelayTest`, the rule was even hit 3× while the operation finished in 3.7s.) This is correct SDK behavior — metadata refreshes are resilient to slow responses — so the wall-clock assertion can never hold.

### Changes

- **`FaultInjectionMetadataTests`** — `AddressRefreshResponseDelayTest`, `PKRangeResponseDelayTest`, and `CollectionReadResponseDelayTest`: replaced the brittle elapsed-time assertions with deterministic validation that the delay rule was applied (`ValidateRuleHit`) **and** the operation still succeeds despite the injected delay (`StatusCode == Created`).
- **`FaultInjectionDirectModeTests.Cleanup`** — made `client`/`fiClient` disposal null-safe (`?.`), matching the other test classes. Previously, a test that failed before the fault-injection client was created threw a masking `NullReferenceException` during cleanup.

### Verification

All 37 builder/unit/metadata tests pass (32 builder/unit + 5 metadata). The fixed tests run in 1–4s and validate the injected rule deterministically.

> Note: the remaining `FaultInjectionDirectModeTests` failures observed locally are account-topology dependent (several tests document "must be a single master account" / hardcode replica & region counts, one is statistically flaky by design, and the custom-status mapping case is core-SDK behavior). They were intentionally left unchanged so they continue to validate correctly against the intended CI account.

## Type of change

- [x] Bug fix (test reliability; no shipped-package source changed)

## Changelog

No changelog entry required — changes are test-only (`Microsoft.Azure.Cosmos/FaultInjection/tests/**`); no shipped-package source under `src/**` is modified, and there is no customer-observable impact.

### DIFF
--- a/Microsoft.Azure.Cosmos/FaultInjection/tests/FaultInjectionDirectModeTests.cs
+++ b/Microsoft.Azure.Cosmos/FaultInjection/tests/FaultInjectionDirectModeTests.cs
@@ -89,8 +89,8 @@ namespace Microsoft.Azure.Cosmos.FaultInjection.Tests
             {
                 await this.highThroughputContainer.DeleteContainerAsync();
             }
-            this.client.Dispose();
-            this.fiClient.Dispose();
+            this.client?.Dispose();
+            this.fiClient?.Dispose();
         }
 
         [TestMethod]
--- a/Microsoft.Azure.Cosmos/FaultInjection/tests/FaultInjectionMetadataTests.cs
+++ b/Microsoft.Azure.Cosmos/FaultInjection/tests/FaultInjectionMetadataTests.cs
@@ -141,20 +141,20 @@ namespace Microsoft.Azure.Cosmos.FaultInjection.Tests
 
                 delayRule.Enable();
 
-                ValueStopwatch stopwatch = ValueStopwatch.StartNew();
-                TimeSpan elapsed;
-
-                ItemResponse<FaultInjectionTestObject> _ = await this.fiContainer.CreateItemAsync<FaultInjectionTestObject>(
+                ItemResponse<FaultInjectionTestObject> response = await this.fiContainer.CreateItemAsync<FaultInjectionTestObject>(
                    new FaultInjectionTestObject { Id = "deleteme", Pk = "deleteme" });
 
-                elapsed = stopwatch.Elapsed;
-                stopwatch.Stop();
                 delayRule.Disable();
 
+                // Validate the response delay rule was injected into the address refresh metadata call.
                 this.ValidateRuleHit(delayRule, 1);
 
-                //Check the create time is at least as long as the delay in the rule
-                Assert.IsTrue(elapsed.TotalSeconds >= 15);
+                // Metadata calls (address refresh) are issued over HTTP with a request timeout and are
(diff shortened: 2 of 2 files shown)


kushagraThapar (commented):
We are completely removing the metadata response delay tests (by removing the time part, we are just checking the status code). Is there no other way to test the metadata response delay? What happens if we check against a smaller response delay value which doesn't cause the http timeout?

NaluTripician:
Good question — I dug into this. The short answer is that a smaller delay doesn't give a reliable signal, but there *is* a deterministic way to verify the delay's effect, which I've now added (commit 340863f).

**Why a smaller delay doesn't help:** all three metadata operations (address refresh, PK-range `ReadFeed`, collection read) route through `HttpTimeoutPolicyControlPlaneRetriableHotPath`, whose **first-attempt timeout is 1s** (`(TimeSpan.FromSeconds(1), TimeSpan.Zero)` — recently raised from 500ms, issue #5642). So any injected delay ≥ 1s trips the timeout → transparent retry, which is exactly the flaky behavior the original `elapsed >= 15s` assertion couldn't tolerate. To *avoid* the timeout we'd need a delay < 1s, but then the assertion `elapsed >= delay` is comparing whole-operation wall-clock time against a sub-second value — TLS handshake, cross-region RTT, and jitter alone are often near or over 1s, so that assertion would effectively pass with or without the injected delay. Either way it's not measuring the delay.

**What I added instead:** rather than only checking the status code, the tests now assert `response.Diagnostics.GetFailedRequestCount() >= 1`. The injected delay forces the metadata request to hit its 1s timeout and get retried, which is recorded as a failed request in the operation diagnostics. The no-fault path (`ValidateFaultInjectionRuleNotApplied`) already asserts this count is `0`, so asserting `>= 1` here deterministically proves the injected delay actually impacted the request pipeline — combined with `ValidateRuleHit` (delay was injected) and `StatusCode == Created` (operation still succeeds). This is deterministic and account-topology independent. Builds clean and the metadata tests pass.

kushagraThapar (approved):
Thanks, LTGM!

aavasthy (approved):

PR #5942, opened 2026-06-10
Diagnostics: Fixes Collection Was Modified exception during cross-region read hedging

### LINKED ISSUE #5940: Diagnostics: TraceWriter throws "Collection was modified" under cross-region read hedging
### Describe the bug

`TraceWriter.TraceJsonWriter.WriteJsonUriArrayWithDuplicatesCounted` throws `System.InvalidOperationException: Collection was modified; enumeration operation may not execute.` when diagnostics are serialized while a cross-region read-hedged operation is still in flight.

The exception bubbles out of `CosmosOpenTelemetryExtensions.LogCosmosDiagnostics` and surfaces in the caller's stack (here, ResourceStack's `CosmosContainer.QueryNextPageAsync`).

### To Reproduce

The race is reproducible deterministically with a unit test added in the fix PR (`ConcurrentSerializationOfMutableCollectionsDoesNotThrow`):

1. Construct a `ClientSideRequestStatisticsTraceDatum` and attach it to an `ITrace`.
2. From a background thread, repeatedly `Add` / periodically `Clear` `datum.ContactedReplicas`, `datum.FailedReplicas`, `datum.RegionsContacted`.
3. From the main thread, repeatedly serialize via `new CosmosTraceDiagnostics(trace).ToString()`.

The exception fires within ~100 iterations.

In production, the trigger is:

- Customer enables **cross-region read hedging** (default `AvailabilityStrategy` from 3.47.0+).
- Multiple parallel store-reader paths share the same `ClientSideRequestStatisticsTraceDatum` instance for a logical operation.
- The OpenTelemetry diagnostics-logging pipeline serializes the trace tree while hedged readers are still mutating those three collections.

### Expected behavior

Diagnostics serialization should never throw, regardless of how many hedged in-flight requests are mutating `ClientSideRequestStatisticsTraceDatum` collections at the same time.

### Actual behavior / stack trace

```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at Microsoft.Azure.Cosmos.Tracing.TraceWriter.TraceDatumJsonWriter.WriteJsonUriArrayWithDuplicatesCounted(String propertyName, IReadOnlyList`1 uris)
   at Microsoft.Azure.Cosmos.Tracing.TraceWriter.TraceDatumJsonWriter.Visit(ClientSideRequestStatisticsTraceDatum clientSideRequestStatisticsTraceDatum)
   at Microsoft.Azure.Cosmos.Tracing.TraceWriter.WriteTraceDatum(IJsonWriter writer, Object value)
   at Microsoft.Azure.Cosmos.Tracing.TraceWriter.TraceJsonWriter.WriteTrace(IJsonWriter writer, ITrace trace, Boolean isRootTrace)
   …  (recursive WriteTrace frames omitted)
   at Microsoft.WindowsAzure.ResourceStack.RegionalStore.Cosmos.Extensions.CosmosOpenTelemetryExtensions.LogCosmosDiagnostics(CosmosDiagnostics diagnostics, String accountHost, CosmosDiagnosticsLoggingMode loggingMode, Double threshold, Double latency, Boolean error)
   at Microsoft.WindowsAzure.ResourceStack.RegionalStore.Cosmos.Extensions.CosmosOpenTelemetryExtensions.TraceCosmosDbRequestWithOpenTelemetry[T,U,V](...)
   at Microsoft.WindowsAzure.ResourceStack.RegionalStore.Cosmos.CosmosContainer`1.QueryNextPageAsync[T](...)
```

### Environment summary

| Field | Value |
|---|---|
| SDK Name | `Microsoft.Azure.Cosmos` |
| SDK Versions affected | Regression window observed in production traffic: **3.56 → 3.58**. Underlying defect predates this; only became frequent once cross-region read hedging adoption increased. |
| Connection mode | Direct |
| Operation pattern | Query (`QueryNextPageAsync`) under cross-region read hedging + OTel diagnostics logging |
| Reporter | Eric Shi (livesite), account `gs-public-scus-rgloc-sp0-s1`, container `StorePool0Store1CosmosDbResourceGroupLocation3` |

### Root cause

`Microsoft.Azure.Cosmos.Tracing.TraceData.ClientSideRequestStatisticsTraceDatum` exposes six collections of diagnostic data. Three are read-safely (lock + cached shallow copy); three are not:

| Property | Backing | Safe read? |
|---|---|---|
| `EndpointToAddressResolutionStatistics` | `Dictionary` + `lock` + cached `shallowCopyOf…` | ✅ |
| `StoreResponseStatisticsList` | `List` + `lock` + cached `shallowCopyOf…` | ✅ |
| `HttpResponseStatisticsList` | `List` + `lock` + cached `shallowCopyOf…` | ✅ |
| **`ContactedReplicas`** | `public List<TransportAddressUri> { get; set; }` — raw | ❌ |
| **`FailedReplicas`** | `public HashSet<TransportAddressUri> { get; }` — raw | ❌ |
| **`RegionsContacted`** | `public HashSet<(string, Uri)> { get; }` — raw | ❌ |

The three raw collections are mutated by `Microsoft.Azure.Documents` (Direct package) store-reader paths via the `IClientSideRequestStatistics` contract (e.g. `requestStats.ContactedReplicas.Add(uri)`). These writes happen on whatever thread completes a hedged store-reader call. Meanwhile, `TraceWriter.TraceJsonWriter.Visit(ClientSideRequestStatisticsTraceDatum)` `foreach`es the same `List<T>` / `HashSet<T>` without taking a snapshot or a lock, so `List<T>.Enumerator.MoveNext()` observes the `_version` mismatch and throws.

This is structurally the same family of bug as the `ArgumentNullException: request` hedging race fixed in 3.57.1 — same shared `ClientSideRequestStatisticsTraceDatum`, different field.

### Vulnerable read sites

All of these enumerate the unsafe collections without a snapshot or lock:

- `Microsoft.Azure.Cosmos/src/Tracing/TraceWriter.TraceJsonWriter.cs` — `Visit(ClientSideRequestStatisticsTraceDatum)` — `ContactedReplicas`, `RegionsContacted`, `FailedReplicas`
- `Microsoft.Azure.Cosmos/src/Tracing/TraceWriter.TraceTextWriter.cs` — `Visit(ClientSideRequestStatisticsTraceDatum)` — same three fields
- `Microsoft.Azure.Cosmos/src/Tracing/TraceSummary.cs` — `UpdateRegionContacted` (`UnionWith` on `RegionsContacted`)
- `Microsoft.Azure.Cosmos/src/Tracing/TraceData/SummaryDiagnostics.cs` — `AggregateRegionsContacted`

### Proposed fix

We cannot lock the mutating threads (they live in the Direct package, outside our control) and we cannot change the public collection types (they are part of the `IClientSideRequestStatistics` contract and downstream callers `Add` directly).

Therefore the fix is applied at the read sites: every consumer takes a **defensive snapshot with retry on transient concurrency exceptions** before iterating. A new helper `ConcurrentCollectionSnapshot` copies the source into a fresh array, retrying up to 5 times on `InvalidOperationException` / `ArgumentException` / `IndexOutOfRangeException` / `ArgumentOutOfRangeException` / `NullReferenceException`, and returns `Array.Empty<T>()` on exhaustion so serialization fails soft (drops the property) instead of throwing.

### Workaround

There is no clean customer-side workaround. Possible mitigations until the fix ships:

- Disable cross-region read hedging (re-introduces the latency regression hedging was meant to fix).
- Disable OTel diagnostics logging on the failing operation.
- Catch and swallow `InvalidOperationException` around the OTel logging call (lossy, leaves partial diagnostics).

### Out of scope (follow-ups)

- `ClientCollectionCache.cs` stores `response.RequestStats.RegionsContacted` (raw `HashSet`) into `TelemetryInformation.RegionsContactedList`, later iterated on telemetry timer threads — same class of race, separate code path.
- Longer term: add `AddContactedReplica` / `AddFailedReplica` / `AddRegionContacted` methods to `IClientSideRequestStatistics` so the Direct package can route writes through a lock, eliminating the need for defensive snapshots on the read side.

### Linked PR

(Fill in once the PR is opened — e.g. `users/<alias>/fix-tracewriter-collection-modified`.)

kundadebdatta:
### Summary

Fixes `System.InvalidOperationException: Collection was modified; enumeration operation may not execute.` thrown from `TraceWriter` while serializing `ClientSideRequestStatisticsTraceDatum` during **cross-region read hedging**.

`ClientSideRequestStatisticsTraceDatum` exposes `ContactedReplicas`, `FailedReplicas`, and `RegionsContacted` as **raw mutable** `List<T>` / `HashSet<T>` instances, because those exact types are required by the `IClientSideRequestStatistics` contract shipped from the Direct package. The Direct-package store-reader paths mutate these collections (`.Add(...)`) **without any locking**, while the v3 diagnostics serializers/aggregators walk them **on a different thread** — for example when cross-region read hedging fires parallel store-reader paths and the host's OpenTelemetry pipeline logs the in-flight diagnostics tree. `List<T>.Enumerator.MoveNext()` observes the `_version` mismatch and throws, and the exception can bubble out of `Diagnostics.ToString()` into customer code.

### TL;DR

`ClientSideRequestStatisticsTraceDatum` exposes `ContactedReplicas`, `FailedReplicas`, and `RegionsContacted` as **raw mutable** `List<T>` / `HashSet<T>` instances. Three sibling collections on the same class (`EndpointToAddressResolutionStatistics`, `StoreResponseStatisticsList`, `HttpResponseStatisticsList`) already return a defensive snapshot under a lock; these three do not. Under cross-region read hedging, the Direct-package store-reader paths populate the raw collections from one thread while ResourceStack's OpenTelemetry pipeline serializes the diagnostics tree from another thread. The serializer's `foreach` walks straight into a `List<T>.Enumerator.MoveNext()` that observes a version mismatch and throws.

### Stack trace (from the report)

```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at Microsoft.Azure.Cosmos.Tracing.TraceWriter.TraceDatumJsonWriter.WriteJsonUriArrayWithDuplicatesCounted(String propertyName, IReadOnlyList`1 uris)
   at Microsoft.Azure.Cosmos.Tracing.TraceWriter.TraceDatumJsonWriter.Visit(ClientSideRequestStatisticsTraceDatum clientSideRequestStatisticsTraceDatum)
   at Microsoft.Azure.Cosmos.Tracing.TraceWriter.WriteTraceDatum(IJsonWriter writer, Object value)
   at Microsoft.Azure.Cosmos.Tracing.TraceWriter.TraceJsonWriter.WriteTrace(IJsonWriter writer, ITrace trace, Boolean isRootTrace)
   …
   at Microsoft.WindowsAzure.ResourceStack.RegionalStore.Cosmos.Extensions.CosmosOpenTelemetryExtensions.LogCosmosDiagnostics(...)
   at Microsoft.WindowsAzure.ResourceStack.RegionalStore.Cosmos.Extensions.CosmosOpenTelemetryExtensions.TraceCosmosDbRequestWithOpenTelemetry[T,U,V](...)
   at Microsoft.WindowsAzure.ResourceStack.RegionalStore.Cosmos.CosmosContainer`1.QueryNextPageAsync[T](...)
```

The inner-most frame is `List<T>.Enumerator.MoveNext()`. That is the canonical "list mutated during enumeration" failure mode.

### Root cause

All hedged readers for one logical operation share the **same** `ClientSideRequestStatisticsTraceDatum` (they belong to the same `ITrace`):

1. A shared datum is created and attached to the trace.
2. Hedging dispatches readers to multiple regions; each mutates the shared `ContactedReplicas` / `FailedReplicas` / `RegionsContacted` from a thread-pool thread inside the Direct package.
3. The host's OTel helper serializes the diagnostics tree on the orchestrating thread **before** all hedged readers have completed.
4. The serializer `foreach`es one of the raw collections while a still-active reader calls `Add(...)` → version mismatch → `InvalidOperationException`.

This is structurally the same class of hedging race as the `ArgumentNullException: request` issue fixed in 3.57.1 — same shared datum, different field. Three sibling collections on the same class (`EndpointToAddressResolutionStatistics`, `StoreResponseStatisticsList`, `HttpResponseStatisticsList`) already serialize safely under a lock with a cached shallow copy; these three did not.

### Fix approach

The mutating threads live in the Direct package, so we cannot lock their writes from v3. The fix is applied at the **read site**: every place that enumerates one of the three unsafe collections now takes a **defensive snapshot with retry on transient concurrency exceptions** before iterating.

A new internal helper, `ConcurrentCollectionSnapshot`, copies the source into a fresh array and retries up to 5 times if the source mutates mid-copy (catching `InvalidOperationException`, `ArgumentException`, `IndexOutOfRangeException`, `ArgumentOutOfRangeException`, `NullReferenceException`). After max retries it returns `Array.Empty<T>()`, so diagnostics serialization **fails soft** (drops the property) instead of throwing — and always emits well-formed JSON/text.

**Why not other approaches?**

- **Change the property types** — they satisfy `IClientSideRequestStatistics` (Direct package) and are mutated from outside this assembly; changing them breaks the cross-package contract and the existing `public List<…>` / `public HashSet<…>` shape that downstream code relies on.
- **`try/catch` around the serializer** — would leak a partially-written JSON property (open array / missing array-end) and corrupt the output.
- **Lock on the property getter** — the getter returns the raw reference, so any lock is released before the consumer iterates; it only helps if the same lock is held during mutation, which we cannot do across the package boundary.

A defensive snapshot at the v3-side read boundary is the smallest surgical change that closes the race without breaking the contract.

### Changes

- **New:** `Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs` — `SnapshotList<T>(IReadOnlyList<T>)` (copy by index) and `SnapshotCollection<T>(IEnumerable<T>)` (copy by enumerator), both with retry; null/empty-safe.
- `TraceWriter.TraceJsonWriter.cs` — `Visit(ClientSideRequestStatisticsTraceDatum)` snapshots `ContactedReplicas` / `RegionsContacted` / `FailedReplicas` before the existing helpers iterate them.
- `TraceWriter.TraceTextWriter.cs` — `Visit(ClientSideRequestStatisticsTraceDatum)` snapshots `ContactedReplicas` (reused across the sections that read it today) and `FailedReplicas` before iteration.
- `TraceSummary.cs` — `UpdateRegionContacted` snapshots `RegionsContacted` before `UnionWith`.
- `TraceData/SummaryDiagnostics.cs` — `AggregateRegionsContacted` snapshots its input before the `foreach`.
- **Tests:** `ClientSideRequestStatisticsTraceDatumTests.cs` adds two deterministic stress tests (20s timeout each):
  - `ConcurrentSerializationOfMutableCollectionsDoesNotThrow` — background thread continuously `Add`s/`Clear`s all three collections while the main thread serializes the trace 5000× via `CosmosTraceDiagnostics.ToString()`. Reproduces the original exception reliably without the fix.
  - `ConcurrentTraceSummaryAggregationDoesNotThrow` — same race against `TraceSummary.UpdateRegionContacted` / `UnionWith`.
- `changelog.md` — entry under `### Unreleased` → `Bugs Fixed`.
- `docs/root-cause-trace-writer-collection-modified.md` — full root-cause write-up.

The properties themselves and the JSON/text wire format are **unchanged**; there is no observable behavior change other than elimination of the race.

### Out of scope (deliberate follow-ups)

- The pre-existing bug in `TraceTextWriter.cs` where the "Regions Contacted" section iterates `ContactedReplicas` instead of `RegionsContacted` is preserved as-is (fixing it changes text-writer output and needs its own PR with baseline updates).
- `ClientCollectionCache` storing the raw `RegionsContacted` `HashSet` into `TelemetryInformation.RegionsContactedList` (later iterated on telemetry timer threads) has the same class of race and should be snapshotted in a follow-up.
- Longer term, deprecating the public mutable property shape in favor of `Add*` methods on `IClientSideRequestStatistics` (a Direct-package change) is the only way to make these collections thread-safe end-to-end.

## Type of change

## Closing issues

closes #5940

## Changelog

- [x] I have added a changelog entry under `### Unreleased` in `changelog.md` for the user-facing impact of this change.


Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// <see cref="SnapshotCollection{T}"/> to copy the collection into a private array
+    /// before iterating. The snapshot retries a small number of times if the source
+    /// collection is observed to mutate mid-copy; if every attempt fails we fall back to
+    /// an empty snapshot so diagnostics serialization can never throw and bubble out as
+    /// an <see cref="InvalidOperationException"/> to customer code.
+    /// </para>
+    /// </remarks>
+    internal static class ConcurrentCollectionSnapshot
+    {
+        private const int MaxAttempts = 5;
```
kundadebdatta, line 38: **🟡 Recommendation** · Robustness: Silent Degradation

**Fail-soft empty snapshot can silently degrade the public `GetContactedRegions()` aggregate**

After `MaxAttempts` (5) the helper returns `Array.Empty<T>()`. For `RegionsContacted` that empty result flows through `TraceSummary.UnionWith` / `SummaryDiagnostics.AggregateRegionsContacted` into **`GetContactedRegions()`** — a customer-facing multi-region monitoring API — so under sustained contention it can under-report regions with nothing logged. (The changelog "no observable behavior change" is not strictly true here.) Consider emitting a verbose/debug trace when the empty-fallback path is taken so the rare drop is diagnosable.

---

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            return Array.Empty<T>();
+        }
+
+        private static bool IsTransientConcurrencyException(Exception ex)
+        {
+            // List<T> / HashSet<T> / Dictionary<,> throw one of these when their backing
+            // store mutates during enumeration or CopyTo.
+            return ex is InvalidOperationException
```
kundadebdatta, line 149: **🟡 Recommendation** · Robustness: Over-broad Catch

**`IsTransientConcurrencyException` swallows `NullReferenceException`**

An NRE is not a transient `List<T>`/`HashSet<T>` concurrency signal (their backing arrays are never null mid-mutation). A genuine NRE — from a future refactor or a custom `IEnumerable<T>` passed in — would be retried 5x, then the method returns `Array.Empty<T>()` with **no exception and no log**, turning a real defect into invisible data loss. `ArgumentOutOfRangeException` is also redundant (it derives from `ArgumentException`). Suggest dropping `NullReferenceException` (and the redundant subclass), paired with a fallback trace.

---

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs
```diff
@@ -0,0 +1,136 @@
...
+                {
+                    int count = source.Count;
+                    if (count == 0)
+                    {
+                        return Array.Empty<T>();
+                    }
+
+                    T[] buffer = new T[count];
+                    int written = 0;
+                    for (int i = 0; i < count && i < buffer.Length; i++)
```
kundadebdatta: **🟡 Recommendation** · Robustness: Undocumented Invariant

**Index snapshot can capture `null`/stale elements without throwing — undocumented**

`List<T>.Add` sets `_size` before writing the slot, and `Clear()` nulls reference elements via `Array.Clear`. Between those operations the indexer bounds-check passes while the slot still holds `default(T)`, so `source[i]` returns `null`/stale **without throwing** — the retry never sees it and the null enters the snapshot. Benign today (verified: `WriteJsonUriArrayWithDuplicatesCounted` skips nulls, `WriteJsonUriArray`/`WriteRegionsContactedArray` use `?.`, the text writer guards), but the helper is documented as a clean snapshot with no mention of this, so the first future consumer that does `item.ToString()` or uses it as a dictionary key reintroduces this bug class. Consider documenting the non-guarantee, or filtering nulls during the copy.

---

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs
```diff
@@ -0,0 +1,136 @@
...
+                        buffer[written++] = source[i];
+                    }
+
+                    if (written == buffer.Length)
+                    {
+                        return buffer;
+                    }
+
+                    T[] trimmed = new T[written];
+                    Array.Copy(buffer, trimmed, written);
```
kundadebdatta: **🟢 Suggestion** · Code Quality: Dead Code

**`trimmed` / partial-copy branch in `SnapshotList` is unreachable**

`buffer.Length == count` and the loop increments `written` once per iteration, so the only way to write fewer than `count` items is for `source[i]` to throw — which jumps to the `catch` and retries, never reaching here. So `written == buffer.Length` is always true and the `trimmed` allocation + `Array.Copy` can never run. (A shrinking `Count` makes the indexer *throw*; it does not make the loop under-write.) Consider collapsing to a plain `for` + `return buffer;`.

---

Microsoft.Azure.Cosmos/src/Tracing/TraceWriter.TraceTextWriter.cs
```diff
@@ -342,7 +342,12 @@ public void Visit(ClientSideRequestStatisticsTraceDatum clientSideRequestStatist
 
                     stringBuilder.AppendLine("Contacted Replicas");
                     Dictionary<Documents.TransportAddressUri, int> uriAndCounts = new Dictionary<Documents.TransportAddressUri, int>();
-                    foreach (Documents.TransportAddressUri uri in clientSideRequestStatisticsTraceDatum.ContactedReplicas)
+                    // Snapshot the underlying List/HashSet defensively — the Direct-package
+                    // store-reader paths can mutate these collections in parallel under
+                    // cross-region read hedging while we serialize diagnostics.
+                    IReadOnlyList<Documents.TransportAddressUri> contactedReplicasSnapshot =
+                        ConcurrentCollectionSnapshot.SnapshotList(clientSideRequestStatisticsTraceDatum.ContactedReplicas);
+                    foreach (Documents.TransportAddressUri uri in contactedReplicasSnapshot)
```
kundadebdatta, line 350: **🟢 Suggestion** · Testing: Missing Coverage

**Text-writer snapshot sites are never exercised by the new stress tests**

`CosmosTraceDiagnostics.ToString()` routes only through the JSON writer (+ `SummaryDiagnostics`); neither new test invokes `TraceTextWriter.Visit`, so these text-writer snapshot call sites are not covered by the race they fix. Consider adding a stress iteration that serializes via the text-writer path under the same background mutation.

---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Tracing/ClientSideRequestStatisticsTraceDatumTests.cs
```diff
@@ -95,6 +95,107 @@ public void VerifyIClientSideRequestStatisticsNullTests()
...
+            });
+
+            // Serialize the diagnostics tree repeatedly while the background task mutates
+            // the underlying collections. Before the fix this throws within ~100 iterations.
+            try
+            {
+                for (int i = 0; i < 5000; i++)
+                {
+                    string json = new CosmosTraceDiagnostics(trace).ToString();
+                    Assert.IsNotNull(json);
```
kundadebdatta, line 149: **🟢 Suggestion** · Testing: Weak Assertion

**Tests only assert "does not throw" — a regression to always-empty would still pass**

If the snapshot regressed to always returning `Array.Empty<T>()`, both tests stay green. Consider adding a non-concurrent assertion that `SnapshotList`/`SnapshotCollection` return the expected elements (plus a direct unit test of `ConcurrentCollectionSnapshot`: null->Empty, retry, the `is IReadOnlyList` fast-path) so the happy path is pinned.

---

Microsoft.Azure.Cosmos/src/Tracing/TraceWriter.TraceTextWriter.cs
```diff
@@ -363,13 +368,15 @@ public void Visit(ClientSideRequestStatisticsTraceDatum clientSideRequestStatist
...
+                    IReadOnlyList<Documents.TransportAddressUri> failedReplicasSnapshot =
+                        ConcurrentCollectionSnapshot.SnapshotCollection(clientSideRequestStatisticsTraceDatum.FailedReplicas);
+                    foreach (Documents.TransportAddressUri failedToContactReplica in failedReplicasSnapshot)
                     {
                         stringBuilder.AppendLine($"{space}{failedToContactReplica?.ToString() ?? "<null>"}");
                     }
 
                     stringBuilder.AppendLine("Regions Contacted");
-                    foreach (Documents.TransportAddressUri regionContacted in clientSideRequestStatisticsTraceDatum.ContactedReplicas)
+                    foreach (Documents.TransportAddressUri regionContacted in contactedReplicasSnapshot)
```
kundadebdatta, line 383: **🟢 Suggestion** · Maintainability: Pre-existing Bug Reinforced

**"Regions Contacted" section iterates `ContactedReplicas` (wrong collection) via the reused snapshot**

This section reuses `contactedReplicasSnapshot`, cementing the pre-existing wrong-field read (it prints `ContactedReplicas`, not `RegionsContacted`; the element type even differs). Deferral is fine per the root-cause doc, but consider an inline `// BUG (pre-existing): should iterate RegionsContacted` marker at this call site so it is not lost in a separate doc.

---

changelog.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Bugs Fixed
 
+- [PR-TBD](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/PR-TBD) Diagnostics: Fixes `InvalidOperationException: Collection was modified; enumeration operation may not execute.` raised from `TraceWriter` while serializing `ClientSideRequestStatisticsTraceDatum` during cross-region read hedging. The trace serializers and trace-summary aggregators now take a defensive snapshot of `ContactedReplicas`, `FailedReplicas`, and `RegionsContacted` before iterating, so diagnostics logging is safe when Direct-package store-reader paths mutate those collections in parallel.
```
kundadebdatta: **🟢 Suggestion** · Style: Changelog Placeholder

**`PR-TBD` placeholder link**

`[PR-TBD](.../pull/PR-TBD)` is a dead link — replace with `#5942` before merge.

---

docs/root-cause-trace-writer-collection-modified.md
```diff
@@ -0,0 +1,143 @@
...
+
+This is structurally the same as the `ArgumentNullException: request` hedging race fixed in 3.57.1 — same shared `ClientSideRequestStatisticsTraceDatum`, different field. Whenever the SDK leaks a raw mutable collection across the boundary between Direct-package writers and v3 readers, it has the potential to throw under hedging.
+
+## Fix approach
+
+The mutating threads live in the Direct package; we cannot wrap their writes in a lock from v3. The fix is therefore applied at the **read site**: every place that enumerates one of the three unsafe collections takes a **defensive snapshot with retry on transient concurrency exceptions** before iterating. The snapshot copies into a fresh array, retrying up to five times if `List<T>.CopyTo` or the indexer raises one of the transient exceptions (`InvalidOperationException`, `ArgumentException`, `IndexOutOfRangeException`, `ArgumentOutOfRangeException`, `NullReferenceException`) that indicate the source mutated mid-copy. After max retries the helper returns an empty snapshot, so diagnostics serialization can fail soft (drop the property) instead of bubbling an exception out of `Diagnostics.ToString()` and into customer code.
```
kundadebdatta: **🟢 Suggestion** · Style: Doc Inaccuracy

**Doc references a `List<T>.CopyTo` path the code never uses**

The implementation copies via the indexer (`source[i]`) and the enumerator — it never calls `CopyTo`. Consider rewording "if `List<T>.CopyTo` or the indexer raises" to "indexer / enumerator" (same in the helper XML `<remarks>`).

---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Tracing/ClientSideRequestStatisticsTraceDatumTests.cs
```diff
@@ -95,6 +95,107 @@ public void VerifyIClientSideRequestStatisticsNullTests()
...
+                int i = 0;
+                while (!cancellationTokenSource.Token.IsCancellationRequested)
+                {
+                    TransportAddressUri replica = new TransportAddressUri(new Uri($"rntbd://host{i % 32}/replica/{i}"));
+                    datum.ContactedReplicas.Add(replica);
+                    datum.FailedReplicas.Add(replica);
+                    datum.RegionsContacted.Add(($"region-{i % 8}", new Uri($"https://region{i % 8}.documents.azure.com")));
+
+                    // Periodically trim to keep memory bounded while still racing.
+                    if (i % 2048 == 0)
```
kundadebdatta, line 131: **🟢 Suggestion** · Robustness: Test Flakiness Margin

**Unthrottled mutation loop is tight against the 20s timeout**

The background loop spins with no delay, continuously allocating `new Uri(...)` + `TransportAddressUri`, pegging a core while the main thread runs 5000 serializations under contention. On a loaded 2-core CI agent this is somewhat tight against `[Timeout(20000)]`; a few hundred iterations reliably hit the race with more margin.

---

kundadebdatta:
**🟡 Recommendation** · Correctness: Completeness Gap (read site outside this PR)

**Missed read site — `PartitionKeyRangeCache.GetRoutingMapForCollectionAsync` enumerates the same raw `RegionsContacted` HashSet**

`Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs` (not part of this diff) logs:

```csharp
DefaultTrace.TraceInformation("PartitionKeyRangeCache ... RegionsContacted {5}", ...,
    response.RequestStats?.RegionsContacted != null
        ? string.Join(", ", response.RequestStats.RegionsContacted)
        : string.Empty);
```

`string.Join` enumerates the raw `RegionsContacted` HashSet, and the argument is evaluated **eagerly regardless of trace level**, guarded only by a null check. If `RequestStats` is the shared hedged datum being mutated by a Direct-package reader, this throws the **identical** `InvalidOperationException: Collection was modified` — on a **metadata read path** (broader blast radius than the diagnostics-serialization report this PR targets). It is neither fixed here nor in the "Out of scope" list.

Consider snapshotting it (`ConcurrentCollectionSnapshot.SnapshotCollection`) or adding it to the deferred list with rationale, and confirm whether change-feed / PKRange reads run under hedging with a shared, still-mutating datum.

_(Posted as a general comment because this file is not in the PR diff and cannot take an inline anchor.)_

---

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Tracing/TraceWriter.TraceTextWriter.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Tracing/TraceWriter.TraceTextWriter.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Tracing/ClientSideRequestStatisticsTraceDatumTests.cs
kundadebdatta: This is now addressed.

changelog.md
kundadebdatta: This is now addressed.

docs/root-cause-trace-writer-collection-modified.md
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Tracing/ClientSideRequestStatisticsTraceDatumTests.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs
kundadebdatta: narrowed transient-exception catch.

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs
kundadebdatta: Addressed - documented null/stale non-guarantee.

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ConcurrentCollectionSnapshot.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        private static bool IsTransientConcurrencyException(Exception ex)
+        {
+            // List<T> / HashSet<T> / Dictionary<,> throw one of these when their backing
+            // store mutates during enumeration or index access.
+            return ex is InvalidOperationException
+                || ex is ArgumentException
+                || ex is IndexOutOfRangeException;
+        }
```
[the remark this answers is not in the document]
kundadebdatta: Thanks! This is now addressed.

ananth7592:
Really nice PR. The root-cause write-up is the clearest hedging-race analysis I've read in this repo, and the deferral list (text-writer wrong-field read, `TelemetryInformation.RegionsContactedList`, longer-term `IClientSideRequestStatistics` rework) is the right scope — keeps this surgical without leaving the next person guessing. The self-review iteration is also impressive: catching the over-broad `ArgumentException` catch, the unreachable `trimmed` branch, and the silent-degradation surface on `GetContactedRegions()` saved a round-trip.

The direct unit tests for `ConcurrentCollectionSnapshot` (null, retry-then-succeed, max-attempts give-up, `IReadOnlyList` fast-path) plus the stress tests give strong confidence the helper does what it says.

One real ask on `TraceSummary.UpdateRegionContacted` — left an inline comment. There's a raw `Count == 0` early-return *before* the snapshot, which I think reopens the same silent-drop failure mode the rest of the PR is hardening against. Trivial reorder, but worth doing before merge.

Microsoft.Azure.Cosmos/src/Tracing/TraceSummary.cs
```diff
@@ -75,9 +75,16 @@ public void UpdateRegionContacted(TraceDatum traceDatum)
                     return;
                 }
 
+                // RegionsContacted is a raw HashSet that may be mutated concurrently by
+                // Direct-package store-reader paths (e.g. under cross-region read hedging).
+                // Snapshot it before UnionWith to avoid `InvalidOperationException:
+                // Collection was modified` during enumeration.
+                IReadOnlyList<(string, Uri)> regionsContactedSnapshot =
+                    ConcurrentCollectionSnapshot.SnapshotCollection(clientSideRequestStatisticsTraceDatum.RegionsContacted);
```
ananth7592, line 83: This is the only one of the new defensive call sites that still reads the raw collection before snapshotting — the early-return at lines 72-73 checks `RegionsContacted.Count == 0` directly on the source.

`HashSet<T>.Count` won't *throw* under concurrent mutation (it's a plain int field read), so this isn't a crash. But it does open a real correctness gap: under cross-region hedging, a writer thread mid-`Add` can leave `Count == 0` visible at the exact moment we read it — we early-return, the regions are silently dropped from `regionContactedInternal`, and they never re-aggregate (this method is *the* aggregation site for the trace). End result: `GetContactedRegions()` under-reports — the same silent-drop failure mode the rest of the PR is hardening against.

Could you flip the order so the `Count == 0` check runs against the snapshot? Roughly:

```csharp
if (clientSideRequestStatisticsTraceDatum.RegionsContacted == null)
{
    return;
}

IReadOnlyList<(string, Uri)> regionsContactedSnapshot =
    ConcurrentCollectionSnapshot.SnapshotCollection(clientSideRequestStatisticsTraceDatum.RegionsContacted);

if (regionsContactedSnapshot.Count == 0)
{
    return;
}

lock (this.regionContactedInternal)
{
    this.regionContactedInternal.UnionWith(regionsContactedSnapshot);
}
```

Cost is one wasted snapshot for the (common) empty-set case, which is `Array.Empty<T>()` on the empty path — basically free.

kundadebdatta:
Instead of going with snapshot based model, a safe access can be done on the direct code path. Have a [sample PR](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5944) to address this.

Hence closing this PR.

PR #5944, opened 2026-06-10
Diagnostics: Fixes Collection Was Modified exception during cross-region read hedging

kundadebdatta:
# Pull Request Template

## Description

## Type of change

## Closing issues

## Changelog

>

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
+++ b/Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
@@ -292,6 +292,9 @@ namespace Microsoft.Azure.Cosmos.Tracing.TraceData
             }
         }
 
+        // Null/default arguments are intentionally accepted by the single-item Record* methods below to
+        // preserve the prior behavior of the raw collections (which allowed Add(default)); callers are
+        // responsible for any null filtering they require.
         public void RecordContactedReplica(TransportAddressUri contactedReplica)
         {
             ImmutableInterlocked.Update(
@@ -307,8 +310,11 @@ namespace Microsoft.Azure.Cosmos.Tracing.TraceData
                 return;
             }
 
-            // The write path sets the full replica list before issuing the request, so this replaces
-            // (rather than appends to) the current snapshot to avoid accumulating entries across retries.
+            // Invariant: the write path always replaces the full replica list before the request is issued,
+            // and read-path appends (RecordContactedReplica) happen afterwards. Because of this
+            // replace-before-append ordering it is safe to use Interlocked.Exchange here. If a caller ever
+            // needs to append before replacing, this must be changed to a CAS merge to avoid dropping the
+            // appended entries.
             Interlocked.Exchange(ref this.contactedReplicas, ImmutableList.CreateRange(contactedReplicas));
         }
 
--- a/Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
+++ b/Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
@@ -6,9 +6,11 @@ namespace Microsoft.Azure.Cosmos.Tracing.TraceData
 {
     using System;
     using System.Collections.Generic;
+    using System.Collections.Immutable;
     using System.Net;
     using System.Net.Http;
     using System.Text;
+    using System.Threading;
     using Microsoft.Azure.Cosmos.Handler;
     using Microsoft.Azure.Cosmos.Json;
     using Microsoft.Azure.Documents;
@@ -33,15 +35,19 @@ namespace Microsoft.Azure.Cosmos.Tracing.TraceData
         private IReadOnlyList<HttpResponseStatistics> shallowCopyOfHttpResponseStatistics = null;
         private SystemUsageHistory systemUsageHistory = null;
 
+        // These collections are read (e.g. during diagnostics serialization) on a different thread than the
+        // store-reader paths that mutate them under cross-region request hedging. They are stored as immutable
+        // snapshots and updated via atomic swaps so concurrent readers never observe a collection mid-mutation.
+        private ImmutableList<TransportAddressUri> contactedReplicas = ImmutableList<TransportAddressUri>.Empty;
+        private ImmutableHashSet<TransportAddressUri> failedReplicas = ImmutableHashSet<TransportAddressUri>.Empty;
+        private ImmutableHashSet<(string, Uri)> regionsContacted = ImmutableHashSet<(string, Uri)>.Empty;
+
         public ClientSideRequestStatisticsTraceDatum(DateTime startTime, ITrace trace)
         {
             this.RequestStartTimeUtc = startTime;
             this.RequestEndTimeUtc = null;
             this.endpointToAddressResolutionStats = new Dictionary<string, AddressResolutionStatistics>();
-            this.ContactedReplicas = new List<TransportAddressUri>();
             this.storeResponseStatistics = new List<StoreResponseStatistics>();
-            this.FailedReplicas = new HashSet<TransportAddressUri>();
-            this.RegionsContacted = new HashSet<(string, Uri)>();
             this.httpResponseStatistics = new List<HttpResponseStatistics>();
             this.Trace = trace;
         }
@@ -67,11 +73,11 @@ namespace Microsoft.Azure.Cosmos.Tracing.TraceData
             }
         }
 
-        public List<TransportAddressUri> ContactedReplicas { get; set; }
+        public IReadOnlyList<TransportAddressUri> ContactedReplicas => this.contactedReplicas;
 
-        public HashSet<TransportAddressUri> FailedReplicas { get; }
+        public IReadOnlyCollection<TransportAddressUri> FailedReplicas => this.failedReplicas;
 
-        public HashSet<(string, Uri)> RegionsContacted { get; }
+        public IReadOnlyCollection<(string, Uri)> RegionsContacted => this.regionsContacted;
 
         public ITrace Trace { get; private set; }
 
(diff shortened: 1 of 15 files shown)


kundadebdatta:
## Deep review summary

Strong, correct, clean fix that closes the hedging race **at the source**: the `IClientSideRequestStatistics` collections are now read-only and mutated only through thread-safe `Record*` methods backed by immutable collections. Thread-safety is sound (readers always get an immutable snapshot; no read lock needed), behavior is preserved (baseline serialized output unchanged), and the migration is complete (no missed mutation/cast sites; the setter was removed so the race cannot reappear).

### 🟡 This PR overlaps with #5942 (same bug #5940, two approaches)

#5944 fixes #5940 at the source (immutable collections in the Direct package). #5942 fixes the same bug on the v3 side with read-site defensive snapshots (the `ConcurrentCollectionSnapshot` helper). If #5944 lands, **#5942's read-site snapshots and the entire `ConcurrentCollectionSnapshot` helper become redundant** — and this PR's changelog entry even links #5942. Recommend deciding which approach ships and closing/reducing the other. Note #5942 carries the concurrency stress tests this PR is missing (see inline comment).

### Minor (also noted)
- The PR description is the unedited template (no summary, no type-of-change checked, no `closes #5940`).
- Enumeration order changes from `HashSet` to `ImmutableHashSet`; only validated for the 2-element baseline — larger region/failed-replica sets may serialize in a different order.
- The v3 `ClientSideRequestStatisticsTraceDatum.RecordRegionContacted` has no production caller (regions route to `TraceSummary`), so its `regionsContacted` field is empty in production — pre-existing, worth confirming intended.
- The `Record*` methods + immutable fields are duplicated verbatim across the Direct impl and the v3 datum; the `RecordContactedReplicas` parameter shadows the field; and there is some whitespace/line-ending churn in `TraceWriterBaselineTests.cs`.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Tracing/ClientSideRequestStatisticsTraceDatumTests.cs
```diff
@@ -49,12 +49,12 @@ public async Task ConcurrentUpdateHttpResponseStatisticsListTests()
...
-            clientSideRequestStatisticsTraceDatum.ContactedReplicas.Add(new TransportAddressUri(new Uri("http://storephysicaladdress1.com")));
-            clientSideRequestStatisticsTraceDatum.ContactedReplicas.Add(new TransportAddressUri(new Uri("http://storephysicaladdress2.com")));
-            clientSideRequestStatisticsTraceDatum.ContactedReplicas.Add(new TransportAddressUri(new Uri("http://storephysicaladdress2.com")));
-            clientSideRequestStatisticsTraceDatum.ContactedReplicas.Add(new TransportAddressUri(new Uri("http://storephysicaladdress2.com")));
-            clientSideRequestStatisticsTraceDatum.ContactedReplicas.Add(new TransportAddressUri(new Uri("http://storephysicaladdress2.com")));
-            clientSideRequestStatisticsTraceDatum.ContactedReplicas.Add(new TransportAddressUri(new Uri("http://storephysicaladdress3.com")));
+            clientSideRequestStatisticsTraceDatum.RecordContactedReplica(new TransportAddressUri(new Uri("http://storephysicaladdress1.com")));
```
kundadebdatta: 🟡 **Recommendation** · Testing: Missing Coverage

**No regression test exercises the actual concurrency fix**

These test updates are mechanical `.Add(x)` -> `.Record*(x)` swaps executed sequentially — nothing reproduces the bug being fixed: one thread calling `Record*` while another enumerates `ContactedReplicas` / `FailedReplicas` / `RegionsContacted`. A future revert to a mutable field would keep every test green. Add a test that spins multiple tasks calling the `Record*` methods while another loops enumerating the properties, asserting no `InvalidOperationException`. (#5942 already contains such stress tests — another reason to consolidate the two PRs.)

---

changelog.md
```diff
@@ -15,6 +15,12 @@ Preview features are treated as a separate branch and will not be included in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Unreleased
+
+#### Fixed
+
+- [5942](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5942) Diagnostics: Fixes `InvalidOperationException` ("Collection was modified; enumeration operation may not execute.") that could be thrown while serializing request diagnostics during cross-region request hedging. The contacted-replica, failed-replica and contacted-region collections are now updated through thread-safe immutable snapshots so concurrent readers never observe a collection mid-mutation.
```
kundadebdatta: 🟡 **Recommendation** · Accuracy: Wrong PR Link

**Changelog entry links #5942, but this PR is #5944**

This new `### Unreleased` entry links to `/pull/5942`, and the commit subject also references #5942. Anyone following the changelog to locate this fix lands on a different PR. Confirm whether #5942 is intentional (umbrella) or should be #5944.

---

Microsoft.Azure.Cosmos/src/direct/ClientSideRequestStatistics.cs
```diff
@@ -169,6 +175,42 @@ public void RecordResponse(
...
+
+        public void RecordFailedReplica(TransportAddressUri failedReplica)
+        {
+            ImmutableInterlocked.Update(
+                ref this.failedReplicas,
+                static (replicas, replica) => replicas.Add(replica),
+                failedReplica);
+        }
+
+        public void RecordRegionContacted(string regionName, Uri locationEndpoint)
```
kundadebdatta: 🟡 **Recommendation** · Robustness: Inconsistent Validation

**Null-guard is inconsistent across the new `Record*` methods**

`RecordContactedReplicas` guards `if (contactedReplicas == null) return;`, but `RecordContactedReplica`, `RecordFailedReplica`, and this `RecordRegionContacted` do not. `RecordRegionContacted(default, default)` (exercised by a test) inserts a `(null, null)` tuple into the set. Today the Direct path is safe only because its *call site* checks `locationEndpoint != null` — but this is now a public interface method any caller can hit. Either guard consistently (e.g. `if (regionName == null || locationEndpoint == null) return;`) or document that nulls are intentionally accepted.

---

Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
```diff
@@ -209,7 +209,7 @@ private async Task<StoreResponse> WritePrivateAsync(
 
                 // the transportclient relies on this contacted replicas being present *before* the request is made
                 // TODO: Can we not rely on this inversion of dependencies.
-                request.RequestContext.ClientRequestStatistics.ContactedReplicas = partitionPerProtocolAddress.ReplicaTransportAddressUris.ToList();
+                request.RequestContext.ClientRequestStatistics.RecordContactedReplicas(partitionPerProtocolAddress.ReplicaTransportAddressUris.ToList());
```
kundadebdatta: 🟢 **Suggestion** · Efficiency: Redundant Allocation

**Double copy on the write path**

The caller materializes a `List` via `.ToList()`, then `RecordContactedReplicas` copies it again via `ImmutableList.CreateRange(...)` — two allocations where one suffices. Since the parameter is `IReadOnlyList` and `CreateRange` copies anyway, drop the `.ToList()` and pass `ReplicaTransportAddressUris` directly.

---

Microsoft.Azure.Cosmos/src/direct/ClientSideRequestStatistics.cs
```diff
@@ -33,6 +35,13 @@ internal sealed class ClientSideRequestStatistics : IClientSideRequestStatistics
         private Lazy<List<HttpResponseStatistics>> httpResponseStatisticsList;
         private SystemUsageHistory systemUsageHistory;
 
+        // These collections are read (e.g. during diagnostics serialization) on a different thread than the
+        // store-reader paths that mutate them under cross-region request hedging. They are stored as immutable
+        // snapshots and updated via atomic swaps so concurrent readers never observe a collection mid-mutation.
+        private ImmutableList<TransportAddressUri> contactedReplicas = ImmutableList<TransportAddressUri>.Empty;
```
kundadebdatta, line 41: 💬 **Observation** · Performance: Immutable Collections on a Hot Path

`ImmutableList` / `ImmutableHashSet` `Add` is O(log n) with per-call node allocation, and enumeration is slower than `List` / `HashSet`. `ClientSideRequestStatistics` is per-request and `RecordContactedReplica` runs per replica per retry, so this adds measurable allocation/GC vs the old mutable collections. N is small, but there is a dedicated `RegionContactedInDiagnosticsBenchmark` — worth running before/after to confirm no meaningful regression and noting the tradeoff. (The `HashSet`->`ImmutableHashSet` switch also changes enumeration order for multi-element sets.)

---

Microsoft.Azure.Cosmos/src/direct/ClientSideRequestStatistics.cs
```diff
@@ -169,6 +175,42 @@ public void RecordResponse(
...
+        public void RecordContactedReplicas(IReadOnlyList<TransportAddressUri> contactedReplicas)
+        {
+            if (contactedReplicas == null)
+            {
+                return;
+            }
+
+            // The write path sets the full replica list before issuing the request, so this replaces
+            // (rather than appends to) the current snapshot to avoid accumulating entries across retries.
+            Interlocked.Exchange(ref this.contactedReplicas, ImmutableList.CreateRange(contactedReplicas));
```
kundadebdatta, line 218: 💬 **Observation** · Concurrency: Mixed Atomics on One Field

`contactedReplicas` is mutated by both `RecordContactedReplicas` (`Interlocked.Exchange`, replace) and `RecordContactedReplica` (`ImmutableInterlocked.Update`, CAS append). This is safe **today** only because the replace always precedes appends (the write path sets the list before the request; read-path appends come after). A future caller that appends-then-replaces would silently drop the appended entries, since `Exchange` ignores the current value. Consider documenting the replace-before-append invariant, or have `RecordContactedReplicas` merge via CAS. (Minor: the parameter `contactedReplicas` also shadows the field — consider renaming to `replicas`.)

---

Microsoft.Azure.Cosmos/src/direct/ConsistencyWriter.cs
kundadebdatta: This is fixed.

Microsoft.Azure.Cosmos/src/Tracing/TraceData/ClientSideRequestStatisticsTraceDatum.cs
```diff
@@ -286,6 +292,48 @@ public void RecordResponse(
...
+        }
+
+        public void RecordContactedReplicas(IReadOnlyList<TransportAddressUri> contactedReplicas)
+        {
+            if (contactedReplicas == null)
+            {
+                return;
+            }
+
+            // Invariant: the write path always replaces the full replica list before the request is issued,
```
FabianMeiswinkel, line 330: IMO This should be clearly communicated in the method name - like SetContactedReplicas or OverrideContactedReplicas - the current name I would read as a merge iwould be happening

FabianMeiswinkel (approved):
LGTM - I think perf validation on the happy path would be good - Assuming the happy path (no failures with retries) will only result in using teh Interlocked.Exchange based method to set replica/regions the overhead should be small - but validating that would be good.

PR #5948, opened 2026-06-11
Diagnostics: Fixes eager diagnostics serialization in internal trace call sites when sink is a no-op

### LINKED ISSUE #5945: `CosmosException.Message` eagerly serializes diagnostics inside trace/log call sites, even when the sink is a no-op
# `CosmosException.Message` eagerly serializes diagnostics inside trace/log call sites, even when the sink is a no-op

## Type

Bug / Performance

## Summary

`CosmosException.Message` lazily builds a string that, for certain failure status codes, appends the **full** serialized diagnostics tree (`Diagnostics.ToString()`). Several internal logging call sites pass `ex.Message` as a *format argument* to `DefaultTrace.TraceXxx(...)`. Because method arguments are evaluated eagerly, `ex.Message` — and therefore the entire diagnostics serialization — runs **before** the trace call, regardless of whether any trace listener is attached or whether the listener does anything with the message. This is wasted CPU/allocations on a hot failure path, and it forces diagnostics serialization in places that never needed the string.

## Where

**The expensive getter** — [`CosmosException.GetMessageHelper`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/0fa540da43f7b32e54049ed11285e7bfb0fa4434/Microsoft.Azure.Cosmos/src/Resource/CosmosExceptions/CosmosException.cs#L203):

```csharp
public override string Message => this.lazyMessage.Value;
// ...
if (statusCode == HttpStatusCode.RequestTimeout
    || statusCode == HttpStatusCode.InternalServerError
    || statusCode == HttpStatusCode.ServiceUnavailable
    || (statusCode == HttpStatusCode.NotFound && headers.SubStatusCode == SubStatusCodes.ReadSessionNotAvailable))
{
    stringBuilder.Append("; Diagnostics:");
    stringBuilder.Append(diagnostics.ToString());   // serializes the whole trace tree
}
```

The `Lazy<string>` only caches *after* the first access, so the first `ex.Message` always pays the serialization cost.

**An example eager call site** — [`AsyncCacheNonBlocking` background-refresh failure logging](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs#L261):

```csharp
DefaultTrace.TraceError(
    "AsyncCacheNonBlocking Failed. key: {0}, operation: {1}, tryRemoved: {2}, Exception: {3}",
    key,
    operationName,
    removed,
    ex.Message);   // <-- forces Message (and diagnostics) to be built right here
```

(There are similar `ex.Message` / `ex` argument patterns at [L118](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs#L118) and [L167](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs#L167); a repo-wide audit for `ex.Message`/`exception.Message` passed into `DefaultTrace.*` is warranted.)

## Why this matters

1. **Wasted work when the sink is a no-op.** `DefaultTrace.TraceError` may have no listeners, or a listener that ignores the message — but the argument is already materialized. The diagnostics serialization cost is paid unconditionally.
2. **Hot path.** This is on exception/background-refresh-failure handling, which under cross-region failover or hedging can fire frequently with large diagnostics trees.
3. **Coupling a log line to a heavyweight serialization.** Serializing diagnostics is significantly more expensive than building a normal exception message, and can itself surface bugs (e.g., the concurrent-collection serialization race fixed in #5942). A log statement should never be the thing that triggers full diagnostics serialization.

## Proposed fix (options)

- **Avoid `ex.Message` in these internal logs.** Prefer `ex.GetType().Name` + a lightweight summary (status/substatus/activityId), or `ex.ToString()` only when the trace level is actually enabled.
- **Gate evaluation on whether tracing is enabled.** Check the effective trace level / listener presence before constructing the message argument, so nothing is serialized when the sink is off.
- **Avoid coupling diagnostics serialization to `Message`.** Consider whether `CosmosException.Message` should include the full diagnostics string at all, vs. exposing diagnostics only via `ex.Diagnostics` / `ex.ToString()`. At minimum, callers that only want a short reason should not pull in diagnostics.

## Scope / acceptance

- Audit all internal `DefaultTrace.*` (and similar) call sites that pass `Exception.Message`/`Exception` as a format argument and ensure no diagnostics serialization happens when the sink is a no-op.
- For `CosmosException`, decide and document the intended relationship between `Message` and diagnostics.
- Add a regression test/benchmark asserting that logging a `CosmosException` at a failure status code does **not** invoke `Diagnostics.ToString()` when tracing is disabled.

## Related

- #5942 — concurrent-collection race during diagnostics serialization (same `Diagnostics.ToString()` path).

NaluTripician:
## Description

Fixes #5945.

`CosmosException.Message` lazily appends the full serialized diagnostics tree (`Diagnostics.ToString()`) for the critical failure status codes (408 RequestTimeout, 500 InternalServerError, 503 ServiceUnavailable, and 404 + ReadSessionNotAvailable). Several internal `DefaultTrace.*` logging call sites passed `ex.Message` as an **eagerly evaluated** format argument, so the heavyweight diagnostics serialization ran **before** the trace call — unconditionally, even when no trace listener consumes the output. This wastes CPU/allocations on hot background-refresh / failover / hedging failure paths and couples a log line to a historically race-prone serialization (cf. #5942).

## Approach

Following a research + spec review (PR Deep Reviewer approved), the lowest-risk option was chosen:

- **Add an internal `Exception.ToTraceSafeString()` helper** (in the existing `Util/Extensions.cs`). For a `CosmosException` it emits a lightweight summary — `CosmosException (StatusCode: <name> (<int>); SubStatusCode: <n>; ActivityId: <id>)` — using only cheap properties and **never** touching `Message`/`Diagnostics`. For any other exception type it returns `ex.Message` (unchanged behavior, since those are already cheap).
- **Apply the helper at the realistic hot offender sites**: `AsyncCacheNonBlocking` (background/lazy refresh + `Refresh()` continuation) and `GlobalAddressResolver` (replica connection-open during address warmup/refresh).
- **`CosmosException.Message` and `ToString()` are unchanged** — they still include diagnostics for the affected status codes. Removing diagnostics from `Message` (issue option 3) would be a breaking change guarded by `CosmosExceptionTests.VerifyDiagnosticsInTimeoutAndServerError`, so it was explicitly left out of scope.

## Tests

- New regression tests in `CosmosExceptionTests`:
  - `ToTraceSafeString_CosmosException_OmitsDiagnostics_KeepsIdentity` — for 408/500/503, asserts the summary omits diagnostics while `Message` still contains it, and keeps status/substatus/activityId.
  - `ToTraceSafeString_CosmosException_DoesNotAccessDiagnosticsOrMessage` — a spy `CosmosException` whose `Diagnostics` getter throws; proves the helper never touches the serialization path.
  - `ToTraceSafeString_NonCosmosException_ReturnsMessage` and `ToTraceSafeString_Null_ReturnsNull`.
- `CosmosExceptionTests` (22) and `AsyncCacheNonBlocking` (15) tests pass; core builds clean. The existing diagnostics-in-Message contract test still passes unchanged.

## Changelog

Added an entry under `### Unreleased` → `Other Changes`.


NaluTripician:
### Self-review (skeptic lens) + CI gate status

**Skeptic-lens fix pushed (b37b413):** While re-reviewing the diff I found the `AsyncCacheNonBlocking.Refresh` background-refresh continuation passes `task.Exception` — an `AggregateException` — to `ToTraceSafeString()`. Because `AggregateException.Message` eagerly concatenates each inner exception's `Message`, a wrapped `CosmosException` would **still** serialize its full diagnostics tree at that exact call site, defeating the fix there. `ToTraceSafeString` now unwraps `AggregateException` inner exceptions through the same safe path, with two added regression tests. Core builds clean; `CosmosExceptionTests` (26) and `AsyncCacheNonBlocking` (15) pass locally.

**CI gates:** All legs pass **except** `EmulatorTests Release - MultiRegion`, which fails with a hard ADO **60-minute per-job timeout** (`ran longer than the maximum time of 60 minutes`), not a test failure. The job log shows every executed test passing with `0 Error(s)` right up to the cancellation point (last: `HedgeNestingDiagnosticsTest`) — no hang and no assertion failure. This reproduced **identically across three runs**, including the pre-fix build (61833), so it is a pre-existing suite-duration/agent-throughput issue unrelated to this change. A re-run of just that leg timed out at the same boundary. Getting this leg green will likely require a pipeline-level change (raise the MultiRegion job timeout or shard the suite), which is out of scope for this diagnostics fix.

The PR is otherwise `MERGEABLE` and only blocked on `REVIEW_REQUIRED` + this timing-out leg.

kundadebdatta (dismissed):
LGTM.

FabianMeiswinkel (changes requested):
My 2 cents - we would for sure not want to not serialize diagnostics as Json in the CosmosException.Message - having diagnostics there has save our ability to debug numerous times.
From the PR description it sounded like Json got serialized also when tracing and even for traces that would not be emitted because the log level is suppressed - if so, the fix should be to gate the serialization for tracing - usually by surrounding the trace statement in an if statement gating log level appropriately?
My 2 cents - for tracing not having diagnostics kind of makes the tracing useless - so, all I would do is gating tracing properly

NaluTripician:
Thanks @FabianMeiswinkel - fully agree, and I've reworked the PR to match your guidance (pushed in bcfb4b255).

**What changed:**
- `CosmosException.Message` and `ToString()` are untouched and still serialize the full diagnostics JSON for the critical status codes - no behavior change there.
- Trace output now **keeps the full diagnostics** as well. Instead of stripping them, each affected `DefaultTrace.*` call is now gated by its level, e.g.:
  `if (DefaultTrace.TraceSource.Switch.ShouldTrace(TraceEventType.Error)) { DefaultTrace.TraceError(..., ex.Message); }`
  The `TraceSource` switch defaults to `Off` (no listeners / no-op sink), so `ShouldTrace` short-circuits and the eager `ex.Message` -> diagnostics serialization is skipped entirely when tracing is suppressed - but when a listener raises the level, the trace emits the complete diagnostics, exactly as before.
- Removed the `ToTraceSafeString` helper and the `MessageWithoutDiagnostics` property (and their tests). The `AsyncCacheNonBlocking` regression test was reworked to assert the gate prevents diagnostics serialization when the level is `Off`.

This is applied at the same hot-path call sites as before (`AsyncCacheNonBlocking`, `GatewayAddressCache`, `GlobalAddressResolver`, `GlobalEndpointManager`, `GlobalPartitionEndpointManagerCore`, cross-region hedging). `CosmosExceptionTests` and `AsyncCacheNonBlockingTests` pass; core builds clean.

FabianMeiswinkel (dismissed):
LGTM

NaluTripician:
@FabianMeiswinkel Good call — I wrote a small standalone harness to reproduce the footprint and quantify what the gating actually removes. TL;DR: reproduced, and the fix removes ~89–99.5% of the per-failure allocation on a no-op sink (scaling with diagnostics size), while a listener still gets the full diagnostics.

## What was measured

For each failure I build a **real `CosmosException` (503)** carrying a **real diagnostics tree** (`Trace` root → per-region children → address-resolution + HTTP-response stats), at three sizes that mirror the hot paths this PR touches (single point-read retry → cross-region failover → hedging/background-refresh pile-up). Then, over N fresh exceptions each, I measure `GC.GetAllocatedBytesForCurrentThread()` + wall time for:

- **A. Eager `ex.Message`** — the pre-fix call-site behavior (`ex.Message` evaluated as a format arg regardless of sink).
- **B. Gated by `ShouldTrace`** — this PR's behavior, sink `Off` (the SDK default — confirmed `DefaultTrace.TraceSource.Switch.Level == Off` at runtime).
- **C. Gated, listener attached** (`Level = All`) — to confirm diagnosability is preserved.

Environment: .NET 6.0.36, workstation GC, 2,000 warmup + 10,000 measured iterations per scenario.

## Results

| Diagnostics tree | Scenario | Alloc / failure log | CPU / failure log |
|---|---|---:|---:|
| **1.9 KB** (1 region × 3 tries) | A. Eager (pre-fix), sink Off | **23.0 KB** | 35.5 µs |
| | B. Gated (this PR), sink Off | **2.6 KB** | 1.8 µs |
| | C. Gated, listener On | 23.0 KB | 32.3 µs |
| **12.7 KB** (4 regions × 6 tries) | A. Eager (pre-fix), sink Off | **136.5 KB** | 143.0 µs |
| | B. Gated (this PR), sink Off | **2.6 KB** | 1.3 µs |
| | C. Gated, listener On | 136.5 KB | 119.3 µs |
| **48.4 KB** (8 regions × 12 tries) | A. Eager (pre-fix), sink Off | **508.6 KB** | 477.5 µs |
| | B. Gated (this PR), sink Off | **2.6 KB** | 3.0 µs |
| | C. Gated, listener On | 508.5 KB | 459.4 µs |

**Reduction on a no-op sink: 88.7% → 98.1% → 99.5%** as the diagnostics tree grows.

A few things worth noting:
- The **~2.6 KB gated floor is constant** and is just the `CosmosException` object itself (Headers + `Lazy` + `CosmosTraceDiagnostics` + activity-id) — that allocation already exists in prod before the `catch`. The delta `A − B` is *purely* the diagnostics serialization the fix eliminates.
- Eager allocation is **~10× the serialized string size** (e.g. 508 KB allocated for a 48 KB payload) because `Diagnostics.ToString()` churns through intermediate JSON-writer buffers, UTF-16 `StringBuilder` growth, etc. So the transient footprint is even bigger than the payload — exactly the kind of thing that hurts under sustained background-refresh/failover failures.
- **Scenario C matches A**, i.e. when a listener is actually attached the gate opens and the full diagnostics are still serialized — we didn't quietly drop logging, we just stopped paying for it when nobody's listening.

## End-to-end proof at the real call site

Beyond the micro-measurement, I ran the actual fixed path — `AsyncCacheNonBlocking.GetAsync` — with a spy `CosmosException` (503) whose `Diagnostics` getter **throws**, sink `Off`:

```
- original exception propagated unchanged: True   (expected: True)
- CosmosException.Diagnostics touched by failure logging: False   (expected: False)
- unexpected error surfaced from logging: False   (expected: False)
=> PASS: the real fixed call site never serializes diagnostics when the sink is Off.
```

If the gate were missing, `ex.Message` would route through the throwing `Diagnostics` getter for a 503 and surface here — it doesn't. This is the same mechanism as the `AsyncCacheNonBlockingTests` regression test in this PR.

## On "all causes"

The serialization only happens for a `CosmosException` whose status is **408 / 500 / 503 / 404+ReadSessionNotAvailable** (the only cases `GetMessageHelper` appends `Diagnostics.ToString()`). This PR gates the **12 internal call sites** on the background-refresh / failover / hedging hot paths where such an exception is logged repeatedly:

- `AsyncCacheNonBlocking` (get + lazy/background refresh + `Refresh()` continuation)
- `GatewayAddressCache` (warm-up open-connections + server-address fetch)
- `GlobalAddressResolver` (replica connection-open during address warmup/refresh)
- `GlobalEndpointManager` (account background-refresh + gateway-reach + refresh)
- `GlobalPartitionEndpointManagerCore` (circuit-breaker failback loop)
- `CrossRegionHedgingAvailabilityStrategy` (hedging execution failure)

The other `DefaultTrace.*(… .Message)` sites in the SDK were intentionally left alone because they log a `DocumentClientException` / generic exception (cheap `Message`, e.g. `PartitionKeyRangeCache`) or sit on one-shot/cold paths (auth-token init, telemetry, query-plan) — none of them serialize a diagnostics tree on a repeated path. `CosmosException.Message` / `ToString()` themselves are unchanged and still include diagnostics for the affected status codes.

Happy to add more shapes or drop the harness into `Microsoft.Azure.Cosmos.Performance.Tests` as a permanent benchmark if you think it's worth keeping around.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
```diff
@@ -114,11 +114,14 @@ public async Task<TValue> GetAsync(
                     {
                         bool removed = this.TryRemove(key);
 
-                        DefaultTrace.TraceError(
-                            "AsyncCacheNonBlocking Failed GetAsync. key: {0}, tryRemoved: {1}, Exception: {2}",
-                            key,
-                            removed,
-                            e.Message);
+                        if (DefaultTrace.TraceSource.Switch.ShouldTrace(System.Diagnostics.TraceEventType.Error))
```
kundadebdatta: nit: How about putting this check in a static method within the `DiagnosticsHandlerHelper`?

kundadebdatta (dismissed):
Nothing blocking. One minor nit comment and fixing the conflict.

Microsoft.Azure.Cosmos/src/Routing/AsyncCacheNonBlocking.cs
NaluTripician: Good call — extracted the gate into a static DiagnosticsHandlerHelper.ShouldTrace(TraceEventType) method and updated all the trace call sites to use it. Pushed in 57cb43b2a.

kundadebdatta (approved):
LGTM

FabianMeiswinkel (approved):
LGTM - Thanks!

FabianMeiswinkel:
@microsoft-github-policy-service rerun

Issue #75, opened 2019-03-12
Mechanism to detect the region from where request is served


kirankumarkolli:
**Is your feature request related to a problem? Please describe.**
Mechanism to detect the region from where request is served

**Describe the solution you'd like**
*Response types expose a way to know the region which served the request. 
Queries which are served from multiple regions also needs to be covered.

**Describe alternatives you've considered**
Currently this information is not available.

**Additional context**
NA

bchong95:
@kirankumarkolli and @j82w  This information is already in the PointOperationDiagnostics right?

j82w:
It does have contacted regions, but it's not just the region. Is this good enough or do we need an explicit region name field?

For example:
URL from data explorer:
`https://myAccountTemp.documents.azure.com:443/`

String from diagnostics:
`{https://myAccountTemp-eastus2.documents.azure.com/}`

kirankumarkolli:
Explicit will make it quite useful if it needs programmatic logic.
For queries, should it be a collection of regions?

j82w:
This only works for direct + TCP mode where the client side request diagnostics exists. There is a PR adding the clients side request statistics to the query diagnostics which will return each page as a JSON object with the query metrics, client side request statistics, and some other info required for debugging.

j82w:
https://github.com/Azure/azure-cosmos-dotnet-v3/pull/2312

PR #5952, opened 2026-06-11
[Internal] CosmosBenchmark: Adds .NET perf/DR-drill workload extensions (FeedRange ops, duration mode, metrics sink, RU/s, Docker)

NaluTripician:
## Summary

Extends the in-repo `CosmosBenchmark` sample for long-running .NET SDK performance and disaster-recovery drills. No SDK product or public-API code is changed, and the new modes remain opt-in.

Adds FeedRange query, feed-range scan, and Change Feed Processor workloads; deadline-based `--duration` runs; AAD/managed-identity authentication; and per-window metrics through console or Azure Data Explorer sinks. Metrics include latency and errors, RU/s, runtime resource usage, SDK/source identifiers, run tags, and bounded Cosmos diagnostics. Window aggregation is synchronized so concurrent executors are neither lost nor double-counted.

Adds Linux container publishing and brings `CosmosBenchmarkTests` into the solution. Data-plane-only identities can fall back from forbidden offer reads to configured throughput, allowing keyless drill deployments to start. Existing count-based, key-authentication, and App Insights behavior remains unchanged.

This is the workload counterpart to `cosmos-sdk-copilot-toolkit` PR #75, which supplies the deployment and monitoring infrastructure for the `DotNetPerf.PerfResults` schema.

## Changelog

- [x] **No changelog entry required** — the change is confined to the CosmosBenchmark sample and does not modify shipped-package source.

## Verification

- `CosmosBenchmark.sln` builds in Release configuration.
- `CosmosBenchmarkTests` passes.


NaluTripician:
### Fix: tolerate offer/throughput read denial under keyless (AAD) auth (`6c21dbd`)

Surfaced by an end-to-end run of the companion `dotnet-perf` infra on a subscription with `disableLocalAuth=true` (keyless / AAD-only). All six workload pods `CrashLoopBackOff`ed at startup.

**Root cause:** `Program.ExecuteAsync` called `container.ReadThroughputAsync()` on startup. That is a control-plane *offer* read, which Cosmos data-plane RBAC does not support, so under AAD auth it returns `403 Forbidden` (substatus 5302) and the benchmark throws before emitting any metrics.

**Fix:** wrap the offer read; on `403` fall back to the configured `--throughput` (`-t`) value and skip the "must have configured throughput" guard. The value is only used to auto-derive the task count when `--pl` is not supplied — and the perf/DR deployments always set `--pl` — so the fallback is behavior-preserving there. Key-based auth is unchanged (only `403` is special-cased).

Builds clean against the public `Microsoft.Azure.Cosmos` (3.61.0); diff is one file, +25/-2.

aavasthy:
@NaluTripician have you run this change live on a test VM?

NaluTripician:
> @NaluTripician have you run this change live on a test VM?

Yes, I have, everything in that test went well.  :)

Pilchie:
Here's an example where I think we should update the PR description before merging so that all doesn't show up in `git log`.

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Fx/AzureDataExplorerMetricsSink.cs
```diff
@@ -0,0 +1,137 @@
...
+    using System.Reflection;
+    using System.Text;
+    using System.Threading.Tasks;
+    using Azure.Core;
+    using Azure.Identity;
+    using Kusto.Data;
+    using Kusto.Data.Common;
+    using Kusto.Data.Ingestion;
+    using Kusto.Ingest;
+    using Newtonsoft.Json;
```
Pilchie, line 20: Nit: We should get in the habit of using `System.Text.Json` instead of `Newtonsoft`.

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Fx/AzureDataExplorerMetricsSink.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                StringBuilder builder = new StringBuilder();
+                foreach (PerfResultsRecord record in records)
+                {
+                    // Newline-delimited JSON: one record per line (multijson stream).
+                    builder.AppendLine(JsonConvert.SerializeObject(record));
+                }
+
+                byte[] payload = Encoding.UTF8.GetBytes(builder.ToString());
+                using MemoryStream stream = new MemoryStream(payload);
```
Pilchie, line 78: You could probably reduce memory allocations and GC time by serializing straight into a Memory stream.

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Fx/AzureDataExplorerMetricsSink.cs
```diff
@@ -0,0 +1,137 @@
...
+
+                KustoQueuedIngestionProperties ingestionProperties = new KustoQueuedIngestionProperties(this.databaseName, this.tableName)
```
Pilchie, line 80: This could be created once and re-used.

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Fx/RuntimeMetricsCollector.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    using System;
+    using System.Diagnostics;
+    using System.Threading;
+
+    /// <summary>
+    /// Snapshot of .NET runtime / process metrics captured for a single window. These occupy the
+    /// dashboard slots that the Rust workload uses for <c>tokio_*</c> gauges.
+    /// </summary>
+    internal readonly struct RuntimeMetricsSnapshot
```
Pilchie, line 15: BTW, the "record" syntax can be very handy for readonly structs without much behavior like these: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/types/records

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Fx/TelemetrySpan.cs
```diff
@@ -86,8 +86,22 @@ public void Dispose()
...
+                return;
+            }
+
             int index = Interlocked.Increment(ref latencyIndex);
-            latencyHistogram[index] = elapsedMilliseoncds;
+
+            // In continuous (--duration) mode the operation count can exceed the histogram
+            // capacity (sized to ItemCount). Guard the write so the run does not crash; the
+            // end-of-run percentile summary is then based on the first ItemCount samples while
+            // the authoritative per-window telemetry continues via the metrics sink.
```
Pilchie, line 100: Is this what we want to do here?

Pilchie: Why not just have a list and append to it so that we have the full set of results?

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/v3/FeedRangeQueryV3BenchmarkOperation.cs
```diff
@@ -0,0 +1,141 @@
...
+            string dbName,
+            string containerName,
+            string partitionKeyPath,
+            string sampleJson)
+        {
+            this.databaseName = dbName;
+            this.containerName = containerName;
+
+            this.container = cosmosClient.GetContainer(this.databaseName, this.containerName);
+            this.partitionKeyPath = partitionKeyPath.Replace("/", "");
```
Pilchie, line 44: nit: What about HPK scenarios here? Should this just skip the first / if present?

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# linux/amd64 is a HARD constraint (the perf deploy host pairs this with the Pyroscope .NET
+# profiler, which supports linux/amd64 only). The base runtime tag matches the benchmark's
+# net6.0 target framework.
+#
+# The pipeline publishes the benchmark to <context>/artifacts and this image copies it in.
+# The ENTRYPOINT forwards all CLI args to the benchmark, e.g.:
+#   docker run <image> -e <endpoint> -w ReadItem --duration PT8H \
+#       --metrics-sink adx --adx-metrics-uri https://ingest-<cluster>.kusto.windows.net \
+#       --sdk-version 3.46.0
+ARG DOTNET_TAG=6.0
```
Pilchie, line 18: Maybe outside the scope of this PR, but we shouldn't be using .NET 6.0 anymore. It's well out of support and will get flagged.  We should update to .NET 8 or 10 for these.

Microsoft.Azure.Cosmos.Samples/Tools/Benchmark/Program.cs
```diff
@@ -182,6 +205,28 @@ private async Task<RunSummary> ExecuteAsync(BenchmarkConfig config)
                 string partitionKeyPath = containerResponse.Resource.PartitionKeyPath;
                 int opsPerTask = config.ItemCount / taskCount;
 
+                // Optional per-window metrics sink (W3). When configured, route per-operation
+                // latency/RU/error samples plus .NET runtime metrics to the dedicated dashboard
+                // schema. Defaults to null (no-op) so existing count-based runs are unaffected.
+                IMetricsSink metricsSink = MetricsSinkFactory.Create(config);
+                PerfMetricsReporter perfReporter = null;
+                if (metricsSink != null)
```
Pilchie, line 213: Another option here would be to use the [Null Object Pattern](https://en.wikipedia.org/wiki/Null_object_pattern) and have the factory create something where the methods are just no-ops, so that you don't have to null check. Doesn't seem like it would simplify this case much though.

Pilchie (approved):

Issue #5464, opened 2025-10-28
SDK does not honor maxItemCount for total item limit across paginated results


egreenberg14:
Describe the bug
When using the SDK to query data with pagination, the maxItemCount parameter is not enforced. Even when specified in the query, the SDK returns all available items instead of capping the total number. This leads to excessive data transfer and memory usage for large datasets.

**Note LIMIT and TOP break the use-case, since our APIs support pagination.**

To Reproduce
Execute a query using the SDK with maxItemCount set (e.g., maxItemCount = 1000).
Enable pagination and iterate through results.
Observe that the SDK continues to return all items, ignoring the global limit.

Expected behavior
The SDK should enforce the maxItemCount parameter as a hard cap on the total number of items returned across all pages.

Actual behavior
The SDK ignores maxItemCount and retrieves all items.
Pagination continues until all items are fetched.
HasMoreResults logic breaks if the query itself limits items (e.g., requesting 1000 items returns 1000 and stops pagination).

Environment summary
SDK Version:
OS Version (e.g. Windows, Linux, MacOSX)

Additional Details
maxBufferedItemCount acts as a buffer but does not solve the root problem.
Limiting items at the query level disrupts pagination semantics.
No native SDK mechanism exists to enforce a global item cap across paginated results.

kirankumarkolli:
@adityasa can you please take a look?

NaluTripician:
Hi @egreenberg14 — thanks for the detailed write-up. After investigating, this is working
as designed: **`maxItemCount` is a per-page size hint, not a cap on the total number of
items returned across all pages.**

## What `maxItemCount` actually controls

`QueryRequestOptions.MaxItemCount` is mapped to the `x-ms-max-item-count` request header,
i.e. the **maximum number of items returned in a single page** (one `ReadNextAsync()` /
one round trip). It is documented as: *"Used for query pagination. This is a maximum.
Query can return 0 items in the page."* When you drain the iterator with a
`while (HasMoreResults)` loop, you receive **every** matching item — `maxItemCount` only
limited how many came back per page. That is expected and unchanged behavior; changing it
to a total cap would silently break every app that relies on per-page paging today.

## How to cap the *total* number of items

You have three supported options, and all of them keep pagination working:

**1. `OFFSET … LIMIT` in the query (recommended — server-side, least RU/data transfer):**
```csharp
var query = new QueryDefinition("SELECT * FROM c OFFSET 0 LIMIT 1000");
[...]

PR #5953, opened 2026-06-11
Query: Fixes MaxItemCount documentation to clarify per-page vs total semantics

NaluTripician:
## Description

Customer issue #5464 reported that `maxItemCount` is "not honored" as a cap on the **total** number of items returned across paginated query results. Investigation confirmed this is **working as designed**: `QueryRequestOptions.MaxItemCount` is a **per-page (per-response) size hint** (mapped to the `x-ms-max-item-count` request header — see `QueryRequestOptions.cs` where it sets `CosmosMessageHeaders.PageSize`), not a cap on the total items across all pages.

The existing XML docs only said *"Used for query pagination ... This is a maximum"* and never contrasted per-page vs total, which is the source of the confusion. This PR clarifies the `<summary>`, `<value>`, and `<remarks>` of `MaxItemCount` to state explicitly that:

- it is the **page size** (max items per `ReadNextAsync()` response);
- it is **not** a cap on the total number of items returned across pages (draining the iterator returns every matching item regardless of this value);
- to limit the **total**, bound the query itself with `OFFSET ... LIMIT` / `TOP` (which compose with pagination), or stop reading pages once the desired total is reached.

**Documentation-only. No runtime behavior change, no public API surface change.** Builds clean (0 warnings, 0 errors).

Related: #5464 — intentionally **not** auto-closed by this PR; a maintainer will post the customer reply and close the issue separately.

## Type of change

- [x] This change requires a documentation update (documentation-only change)

## Changelog

- [x] No changelog entry is required because this PR is one of: documentation-only, test-only, CI / build-only, or a pure internal refactor with no observable customer impact.

> Justification: Documentation-only change to the XML doc comments on `QueryRequestOptions.MaxItemCount`. No change to runtime behavior, return values, error shape, performance, or public API surface — not observable by a customer upgrading and running an existing workload (repo changelog classifier: doc-only ⇒ no entry).

---

🤖 Opened as **draft** by an automated investigation (Seon) of issue #5464. The drafted customer reply for the issue is prepared separately and awaits a maintainer to post.


xinlian12 (approved):
LGTM, thanks

kundadebdatta (approved):
LGTM

PR #5954, opened 2026-06-11
AsyncCacheNonBlocking: Fixes NotFound eviction disabled by a prior non-removable refresh failure

NaluTripician:
## Bug (MEDIUM / cache correctness)

`AsyncCacheNonBlocking<TKey,TValue>` keeps a stale value while a background refresh runs and, on a background-refresh failure, decides whether to evict the entry. The eviction is gated by a **one-shot latch** (`AsyncLazyWithRefreshTask.removedFromCache`, via `ShouldRemoveFromCacheThreadSafe()`).

### Root cause
`UpdateCacheAndGetValueFromBackgroundTaskAsync` called `ShouldRemoveFromCacheThreadSafe()` **before** checking `removeFromCacheOnBackgroundRefreshException(ex)`. The latch flips `removedFromCache = true` on its **first** call and returns `false` forever after — it is never reset. So a **non-removable** (non-404, e.g. transient timeout / 503 / Gone) background-refresh failure consumed the latch without removing. Because the lazy instance survives in the cache (stale-while-refresh design), a **later genuine 404** found the latch already consumed, skipped the eviction branch, and never removed the stale entry.

**Impact:** affects `GatewayAddressCache.serverPartitionAddressCache` and `PartitionKeyRangeCache.routingMapCache` (both use the default `RemoveNotFoundFromCacheOnException` + background `Refresh`/forceRefresh). The SDK could keep serving stale addresses / routing-map for a key after a transient failure preceded a 404.

### Fix
In `UpdateCacheAndGetValueFromBackgroundTaskAsync`, evaluate `removeFromCacheOnBackgroundRefreshException(ex)` **first** and only call `ShouldRemoveFromCacheThreadSafe()` + `TryRemove` when the exception is actually removable (using `&&` short-circuit). This consumes the latch **only when a removal is actually performed**, preserving the remove-at-most-once / initial-create-but-never-awaited intent (the `GetAsync` cached-path catch is unchanged) while keeping the latch available for a subsequent 404. Thread-safety (`removedFromCacheLock`) is preserved.

### Verified reproduction (fails pre-fix, passes post-fix)
```csharp
var cache = new AsyncCacheNonBlocking<string,string>(enableAsyncCacheExceptionNoSharing:false);
await cache.GetAsync("key", _ => Task.FromResult("value1"), _ => false);            // seed value1
try { await cache.GetAsync("key", async _ => { await Task.Delay(5); throw new Exception("Transient timeout"); }, _ => true); } catch (Exception ex) when (ex.Message=="Transient timeout") {}   // non-404 -> retained, latch consumed (pre-fix)
try { await cache.GetAsync("key", async _ => { await Task.Delay(5); throw new NotFoundException("Item was deleted"); }, _ => true); } catch (NotFoundException) {}   // 404 -> MUST evict
string r = await cache.GetAsync("key", _ => Task.FromResult("value3"), _ => false);
Assert.AreEqual("value3", r);   // FAILS pre-fix: Actual <value1> (stale)
```

### Tests added (`AsyncCacheNonBlockingTests.cs`)
1. `BackgroundRefresh_WhenNonRemovableFailurePrecedesNotFound_ShouldStillEvictFromCache` — core regression (non-404 then 404 ⇒ fresh value served).
2. `BackgroundRefresh_SingleFailure_EvictsOnlyOnRemovableException` — non-404 alone retains stale; 404 alone evicts.
3. `BackgroundRefresh_WhenMultipleNonRemovableFailuresPrecedeNotFound_ShouldStillEvictFromCache` — several non-removable failures then a 404 still evicts.
4. Existing `ValidateNegativeNotAwaitedScenario` (initial-create-but-never-awaited removes once) preserved and still green.

### Test results
- `--filter "FullyQualifiedName~AsyncCacheNonBlocking"`: **18 passed, 0 failed**.
- `--filter "FullyQualifiedName~PartitionKeyRangeCache|FullyQualifiedName~GatewayAddressCache"`: **28 passed, 0 failed**.

### Changelog
Added entry under `### Unreleased` → `#### Bugs Fixed` (customer-facing).


ananth7592 (approved):
LGTM, Good test coverage with 16 caller concurrency .etc.

xinlian12 (approved):
LGTM, thanks

PR #5955, opened 2026-06-11
[Internal] WebExceptionRetryPolicy: Fixes retry budget to use Elapsed.TotalSeconds

NaluTripician:
## Description

`WebExceptionRetryPolicy` is meant to retry transient web exceptions for a **total** wall-clock budget of 30 seconds. It computed the remaining budget with `this.durationTimer.Elapsed.Seconds` — the **0–59 seconds (modulo-60) component** of the `TimeSpan` — instead of `Elapsed.TotalSeconds`.

**Effect of the bug:** once cumulative elapsed time crosses 60s, `.Seconds` wraps back toward 0, so `remainingSeconds` becomes positive again and the policy keeps retrying well past its 30-second contract. It self-terminates eventually (the backoff doubles), but the budget is violated and worst-case retry latency is inflated.

**Fix (surgical):** use `Elapsed.TotalSeconds` and make the remaining-budget local a `double`, preserving the existing `Math.Min(currentBackoffSeconds, remainingSeconds)` backoff semantics.

### 🔎 Dead-code finding (verified)
A full-repo search confirms `WebExceptionRetryPolicy` is **currently unreferenced** — the only references to the type are inside its own file; nothing in `src` or `tests` instantiates or wires it into any handler/pipeline. So this is a correctness fix to dead code with **no customer-observable impact today**.

Per the brief, scope is kept minimal:
- The type is **not** wired into any pipeline (out of scope).
- The class is **not** deleted (out of scope).

### Tests
Added `WebExceptionRetryPolicyTests.cs` (5 tests, all passing):
1. First retriable `WebException` → retried with `BackoffTime == 0` (first retry not penalised).
2. Non-retriable exception → `NoRetry` immediately.
3. **Core regression:** with 65s elapsed (driven by a minimal internal elapsed-time seam), the 2nd attempt returns `NoRetry`, honoring the 30s budget. Under the old `.Seconds` logic 65s read as 5s → budget falsely appeared to have 25s left.
4. Within-budget (5s elapsed) → still retries with the initial 1s backoff.
5. Pure-math assertion documenting `TimeSpan.FromSeconds(65).Seconds == 5` vs `.TotalSeconds == 65`.

A low-risk `internal` test-only constructor (`Func<TimeSpan>` elapsed provider) was added so the budget can be validated deterministically; the type is already exposed to the test assembly via `InternalsVisibleTo`, and the public constructor's behavior is unchanged.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — to currently-unreferenced internal code

## Changelog

> Changelog classifier: the diff touches `Microsoft.Azure.Cosmos/src/**`, but `WebExceptionRetryPolicy` is confirmed unreferenced (dead code) — it is never instantiated or wired into any pipeline in `src` or `tests`. No customer can reach this code path by upgrading the SDK, so there is no customer-observable behavior, return-value, error-shape, latency, or allocation change. Not a preview default-config path either. Therefore no changelog entry is required.


FabianMeiswinkel (approved):
LGTM

aavasthy (approved):

PR #5866, opened 2026-05-14
Routing: Fixes LengthAware comparer wiring across env var and cold-start CollectionRoutingMap paths

ananth7592:
## Summary

Wires \AZURE_COSMOS_USE_LENGTH_AWARE_RANGE_COMPARATOR\ env var to all LengthAware usage sites so customers can disable it as a mitigation for the HPK silent empty query issue (#5859).

## Problem

Two disconnected controls existed for the LengthAware range comparer:
1. \CosmosClientOptions.UseLengthAwareRangeComparer\ — hardcoded true for GA/Preview, controlled TryCombine (the bug path), NOT wired to env var
2. \ConfigurationManager.IsLengthAwareRangeComparatorEnabled()\ — read env var but only defaulted true in PREVIEW, no effect in GA

### ConfigurationManager.IsLengthAwareRangeComparatorEnabled()
- Default: true in all non-INTERNAL builds (was PREVIEW-only). INTERNAL stays false.
- \AZURE_COSMOS_USE_LENGTH_AWARE_RANGE_COMPARATOR\ env var now effective in GA SDK.

### CosmosClientOptions.UseLengthAwareRangeComparer
- Replaced hardcoded \#if !INTERNAL\ literal with \ConfigurationManager.IsLengthAwareRangeComparatorEnabled()\
- Wires \PartitionKeyRangeCache.TryCombine\ to the env var

## All LengthAware sites unified under single env var

| Code path | Before | After |
|---|---|---|
| \PartitionKeyRangeCache.TryCombine\ | not wired | wired |
| \QueryRangeUtils\ | PREVIEW only | GA + PREVIEW |
| \PartitionRoutingHelper\ | PREVIEW only | GA + PREVIEW |

## Customer Mitigation
Set \AZURE_COSMOS_USE_LENGTH_AWARE_RANGE_COMPARATOR=false\ to disable LengthAware routing.

Related: #5859, #5850

PR #5260, opened 2025-06-25
[Internal] HPK: Fixes EPK Normalization problem of Hierarchical Partition Keys by consuming new Range Comparators 

ananth7592:
Fix EPK normalization for HPK containers and introduce length-aware range comparators

Design reference: Method 3 Smart Range Comparer
 from  https://microsoft-my.sharepoint.com/:w:/p/amudumba/EbtYRzMDKHZPkqV_Q4VVTqgBVjIhgnSG8X_CGbqu21W_bQ?e=KJwxwy

# Pull Request Template

## Description

    - Addressed the issue of unnormalized Effective Partition Key (EPK) ranges for containers with Hierarchical Partition Keys (HPK) enabled.
    - Integrated LengthAwareMinComparator and LengthAwareMaxComparator from the Direct package, enabling advanced, length-aware comparison logic for partition key range overlap detection.
    - The LengthAwareMin/Maxcomparators will be enabled by default
    - An env variable called AZURE_COSMOS_USE_LENGTH_AWARE_RANGE_COMPARATOR is created for emergency disablement of the LengthAWareMin/MAx comparators.
    - Added comprehensive tests to validate the new CollectionRoutingMap logic, including edge cases for HPK and EPK normalization.

## Safety:
    - All changes are covered by new and existing unit tests to ensure backward compatibility and correctness for both HPK and non-HPK containers.
    - An env variable called AZURE_COSMOS_USE_LENGTH_AWARE_RANGE_COMPARATOR is introduced which can be set to False in case of emergencies for falling back to older simple Min/max Range comparators.

## Testing:

- Tested combinations of:

Partial and full partition key values.
Two- and three-level partition key hierarchies.
Backend range configurations including hybrid and fully-specified EPKs.

- Covered edge and boundary scenarios to ensure correctness and resilience.

## Impact:

Reduced operational complexity by not requiring changes to how EPKs are stored.
Backward compatibility for clients, ensuring a seamless experience for both new and legacy clients.
Minimal disruption during rollout, used a feature flag for emergency rollback.

## Type of change

Please delete options that are not relevant.

- [*] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] This change requires a documentation update

## Closing issues

To automatically close an issue: closes #IssueNumber

PR #5956, opened 2026-06-11
Routing: Fixes GetRangeByEffectivePartitionKey to honor the configured length-aware comparer

NaluTripician:
## Summary

`CollectionRoutingMap.GetRangeByEffectivePartitionKey` hard-coded the **ordinal** `Range<string>.MinComparer.Instance` for its binary search, while its sibling `GetOverlappingRanges` uses the **configured** comparer (`this.comparers`, set via `RangeComparerProvider.GetComparers(useLengthAwareRangeComparer)`). Length-aware comparison is the GA default (`ConfigurationManager.IsLengthAwareRangeComparatorEnabled() == true` for non-INTERNAL builds — PRs #5260 / #5866). As a result, the two methods that both answer *"which partition owns this EPK?"* could disagree for a short/partial EPK that sits on a zero-padded range boundary.

### Fix
One line in `Microsoft.Azure.Cosmos/src/Routing/CollectionRoutingMap.cs`:

```diff
 int index = this.orderedRanges.BinarySearch(
     new Range<string>(effectivePartitionKeyValue, effectivePartitionKeyValue, true, true),
-    Range<string>.MinComparer.Instance);
+    this.comparers.MinComparer);
```

This mirrors the #5260 change that switched `GetOverlappingRanges` to `this.comparers` (the ordinal sort of `orderedRanges` is preserved; the binary-search precondition is the same one accepted there). When `useLengthAwareRangeComparer == false`, `this.comparers.MinComparer` resolves to the legacy ordinal comparer, so the legacy path is unchanged.

## Honest / masked-impact note (why this is a latent fix)

All **current** synchronous callers — `AddressResolver`, `BatchExecUtils`/`BatchAsyncContainerExecutor`, `ReadManyQueryHelper` — pass **full-length** EPKs, for which ordinal and length-aware comparison are equivalent. So this is a **latent correctness / defense-in-depth** fix, **not** a confirmed live customer regression today. It prevents a wrong-partition result if any future caller passes a short/partial EPK to the point lookup.

## Verified reproduction (test fails pre-fix)

Using the existing length-aware routing map (`GenerateRoutingMap(useLengthAwareComparer: true)`), whose range "1"/"2" boundary is the zero-padded `06AB34CFE4E482236BCACBBF50E234AB00000000000000000000000000000000`:

```csharp
map.GetRangeByEffectivePartitionKey("06AB34CFE4E482236BCACBBF50E234AB").Id
```

- **Pre-fix (ordinal):** returns `"1"` (lands one bucket low).
- **Post-fix (length-aware):** returns `"2"` — matching `GetOverlappingRanges` (existing scenario 1.1).

Confirmed by temporarily reverting the fix and running the new tests: `TestGetRangeByEffectivePartitionKeyUsesLengthAwareComparer` and `TestGetRangeByEffectivePartitionKeyAgreesWithGetOverlappingRanges` both fail with `Expected:<2>. Actual:<1>`; they pass after the fix.

## Tests added (`CollectionRoutingMapTest.cs`, reusing `GenerateRoutingMap`)

1. **Length-aware point lookup (regression):** boundary EPK → `"2"`.
2. **Agreement with sibling:** point lookup agrees with `GetOverlappingRanges` on the boundary EPK.
3. **Legacy/ordinal unchanged:** with `useLengthAwareComparer: false`, boundary EPK → `"1"` (historical behavior preserved).
4. **Full-EPK callers unaffected:** a fully specified 64-char EPK maps to the same partition under both comparers.
5. Existing `TestCollectionRoutingMap`, `TestCollectionRoutingMapWithLengthAwareRangeComparators`, and `TestLegacyComparatorsUsedWhenLengthAwareComparatorFlagIsFalse` remain green.

**Result:** `dotnet test --filter "FullyQualifiedName~CollectionRoutingMap"` → Passed: 14, Failed: 0.

## Changelog

Applied the repo changelog classifier: the diff touches shipped `src/**`, but live impact is masked today (current callers pass full EPKs, where ordinal == length-aware), so there is no guaranteed customer-observable change on upgrade. This is the ambiguous defense-in-depth case, so per the classifier's "default to an Other Changes entry and explain" rule, an entry is added under **#### Other Changes** rather than silently omitted.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Routing/CollectionRoutingMap.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/CollectionRoutingMap.cs
@@ -176,7 +176,7 @@ namespace Microsoft.Azure.Cosmos.Routing
 
             int index = this.orderedRanges.BinarySearch(
                 new Range<string>(effectivePartitionKeyValue, effectivePartitionKeyValue, true, true),
-                Range<string>.MinComparer.Instance);
+                this.comparers.MinComparer);
 
             if (index < 0)
             {
--- a/Microsoft.Azure.Cosmos/src/Routing/CollectionRoutingMap.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/CollectionRoutingMap.cs
@@ -174,6 +174,14 @@ namespace Microsoft.Azure.Cosmos.Routing
                 return this.orderedPartitionKeyRanges[0];
             }
 
+            // Search the ordinally-built orderedRanges with the configured (length-aware-by-default)
+            // comparer, mirroring GetOverlappingRanges. This is safe because partition key ranges form a
+            // contiguous hex-string space: (1) range boundaries with different significant prefixes keep the
+            // same relative order under both the ordinal and length-aware comparers, so the ordinal sort of
+            // orderedRanges is a valid precondition for the binary search; and (2) a short/partial EPK is
+            // intentionally treated as equal to its zero-padded boundary, which is exactly the disagreement
+            // with the previous ordinal comparer that this method must now resolve consistently with
+            // GetOverlappingRanges.
             int index = this.orderedRanges.BinarySearch(
                 new Range<string>(effectivePartitionKeyValue, effectivePartitionKeyValue, true, true),
                 this.comparers.MinComparer);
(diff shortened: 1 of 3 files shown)


FabianMeiswinkel (approved):
LGTM

aavasthy (approved):

PR #5957, opened 2026-06-11
Json: Fixes reference-string StackOverflow in binary reader/navigator decode path

NaluTripician:
## Summary

This is the un-fixed twin of **#5909**. #5909 added a reference-string (`StrR`) cycle / reference-to-reference guard **only to the binary writer re-serializer** (`JsonWriter.JsonBinaryWriter.RewriteResolvedReferenceString` / `ForceRewriteRawJsonValue`). The **reader / value-materialization path had no guard**, so a crafted binary-JSON payload could still crash the client process with an unrecoverable `StackOverflowException` (uncatchable; terminates the host) — a denial-of-service.

### Root cause
`Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.Strings.cs` — the `StrR1`/`StrR2`/`StrR3`/`StrR4` cases in `TryGetBufferedLengthPrefixedString` recursed via `TryGetBufferedStringValue(buffer, buffer.Slice(offset), …)` against the root buffer with **no check** that the target marker is a non-reference string and no bounds check. Reached from:
- **Reader:** `JsonReader.JsonBinaryReader.GetStringValue` → `GetUtf8StringValue` → `GetUtf8MemoryValue`.
- **Navigator / `CosmosElement`:** `JsonNavigator.JsonBinaryNavigator.GetStringValue` / `TryGetObjectProperty`.

A payload whose `StrR` target is another `StrR` (or itself) recurses forever → `StackOverflowException`.

### Verified reproductions
- **(A) Destructive (confirmed crash):** `byte[] { 0x80, 0xE2, 0x02, 0xC3, 0x03 }` read via `CosmosElement.CreateFromBuffer<CosmosArray>` then `((CosmosString)array[0]).Value.ToString()` → *"Test host process crashed : Stack overflow"* with frames `TryGetBufferedStringValue → GetUtf8MemoryValue → GetUtf8StringValue`. (Not committed as a test — it crashes the host pre-fix.)
- **(B) Safe asymmetry:** `byte[] { 0x80, 0xE2, 7, 0xC0, 1, 'x', 0xC3, 8, 0xC3, 3 }`. The writer path threw `JsonInvalidTokenException`, but the reader path silently returned `"x"` — proving the missing guard.

## Fix
In `TryGetBufferedLengthPrefixedString`, before recursing for each `StrRn` case, a new helper `TryGetReferenceStringTarget` (mirroring the writer's `RewriteResolvedReferenceString` semantics exactly):
1. Bounds-checks the computed target offset against `buffer.Length` using an **unsigned compare** (also rejects negative `StrR4` offsets).
2. Peeks the target type marker; if `!TypeMarker.IsString(t)` or `TypeMarker.IsReferenceString(t)`, returns `false`.

Returning `false` makes `GetUtf8MemoryValue` / `GetStringValue` throw the catchable `JsonInvalidTokenException`, matching the writer contract. Rejecting `StrR → StrR` breaks all cycles/chains because a legal buffer only ever has one-hop `StrR → StrL`. Legal one-hop references keep working.

The fix is contained to `JsonBinaryEncoding.Strings.cs`; the on-wire format and legal single-hop `StrR` semantics are unchanged.

## Tests
New `JsonBinaryReaderReferenceStringTests` (reader/navigator/CosmosElement counterpart of `JsonBinaryReferenceStringTests`):
1. Self-reference `{0x80,0xE2,0x02,0xC3,0x03}` → throws `JsonInvalidTokenException` (was StackOverflow) — core regression.
2. Ref-to-ref `{0x80,0xE2,7,0xC0,1,'x',0xC3,8,0xC3,3}` element[1] → throws (was: returned `"x"`).
3. `StrR2` and `StrR4` ref-to-ref → throw (mirror writer tests).
4. Out-of-bounds and negative `StrR4` offset → throw.
5. Valid one-hop `{0x80,0xE2,9,0xC0,5,'h','e','l','l','o',0xC3,3}` → element[1] reads `"hello"` (no over-rejection).

Each case is driven through both the `JsonReader` and `CosmosElement` entry points. Tests assert the **post-fix catchable exception** (no test triggers a StackOverflow).

### Results
- New + existing reference-string tests: **18 passed**.
- Broader Json suite (`FullyQualifiedName~Json.Json`): **311 passed, 2 skipped, 0 failed**.
- Build clean (0 warnings, 0 errors).

## Changelog
Added a `#### Bugs Fixed` entry under `### Unreleased` in customer-facing DoS language, noting it is the reader/navigator counterpart to #5909.


changelog.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Bugs Fixed
 
+- [5957](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5957) Json: Fixes a denial-of-service vulnerability where a crafted binary-JSON response payload could crash the client process with an unrecoverable `StackOverflowException` while reading a string value. This is the reader/navigator counterpart to [#5909](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5909) (which only hardened the writer): the binary reader, navigator, and `CosmosElement` string-materialization paths now validate reference-string (`StrR1`/`StrR2`/`StrR3`/`StrR4`) targets before dereferencing them, rejecting self-references, reference-to-reference chains/cycles, and out-of-bounds (including negative) offsets with a catchable `JsonInvalidTokenException` instead of terminating the host. Well-formed single-hop reference strings continue to resolve normally.
```
FabianMeiswinkel: Please change the terminology here - a denial of service attack would usually mean an untrusted party is issueing teh denial of service attack. In this case only someone with write access to the container could inject the invalid binary json - so, IMO it does not qulaify as a DoS - hardening the code makes still sense - but remove any terminology form changelog that could cause unwarranted panic.

FabianMeiswinkel (dismissed):
LGTM except for the changelog entry

changelog.md
NaluTripician: Good call - updated the changelog entry to drop the denial-of-service framing. Since the malformed binary JSON can only be injected by a party that already has write access to the container, I reworded it as a hardening fix for an unrecoverable StackOverflowException rather than a DoS vulnerability.

FabianMeiswinkel (dismissed):
LGTM

Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.Strings.cs
```diff
@@ -563,6 +599,49 @@ private static bool TryGetBufferedLengthPrefixedString(
...
+        /// cycle (e.g. StrR1@a -> StrR1@b -> StrR1@a, or a self reference) that
+        /// would otherwise recurse forever into an uncatchable
+        /// <see cref="StackOverflowException"/> and crash the process.
+        ///
+        /// Returning <c>false</c> lets the callers (GetUtf8MemoryValue /
+        /// GetStringValue) surface a catchable <see cref="JsonInvalidTokenException"/>,
+        /// matching the writer contract. The unsigned bounds compare also rejects
+        /// negative StrR4 offsets so we never index before the start of the buffer.
+        /// </remarks>
+        private static bool TryGetReferenceStringTarget(
```
ananth7592: Small maintainability thought, not blocking: this helper and the writer's `RewriteResolvedReferenceString` now encode the exact same invariant — target offset in-bounds, and the target is a non-reference string — just in two files with two different shapes (one returns `false`, the other throws). The reason this PR exists at all is that #5909 hardened the writer side and this reader side got missed. So if we ever add a new reference-string marker or change what `TypeMarker.IsString` considers a string, both spots have to move in lockstep, and it'd be easy to fix one twin and miss the other again — same gap that produced this PR.

Would it be worth pulling the rule into a single shared predicate, say a static `JsonBinaryEncoding.IsValidReferenceStringTarget(buffer, offset)`, and having both the reader helper and the writer guard call it? The writer keeps its `throw`, the reader keeps its `return false`, but the actual validation lives in one place. Totally fine to leave as-is if you'd rather not widen the scope here.

ananth7592 (dismissed):
LGTM,

Microsoft.Azure.Cosmos/src/Json/JsonBinaryEncoding.Strings.cs
NaluTripician: Good call - done in 6d065b9. Pulled the invariant into a single `JsonBinaryEncoding.IsValidReferenceStringTarget(buffer, offset)` that both sides now call: the reader wrapper slices the target on `true`, the writer keeps its `throw` on `false`. So the in-bounds + non-reference-string rule lives in exactly one place and the two paths can't drift again, which was the whole reason this PR existed. Existing reader + writer reference-string tests still green (20 passing).

FabianMeiswinkel (approved):
LGTM

ananth75 (approved):

kundadebdatta (approved):
LGTM

Issue #5857, opened 2026-05-12
Distributed Transactions: Session token merge fails in Gateway mode - missing partitionKeyRangeId in distributed transaction operation results


Meghana-Palaparthi:
# Description

## Background

The SDK stores session tokens in `SessionContainer` in canonical` {pkRangeId}:{lsn}` form (e.g., `"345:-1#5083"`). For non-DTC operations (Direct, Gateway, and Thin Client modes), the SDK already arrives at this canonical form before reaching `SessionContainer`, so the storage code can assume the format and parse it via `Split(':')`.

For Distributed Transactions (DTC), the situation is different: a single transaction can span multiple partitions, so session tokens are returned **per-operation** in the JSON response body rather than as a single HTTP response header:

```json
{
  "operationResponses": [
    { "index": 0, "statusCode": 201, "sessionToken": "-1#5083" },
    { "index": 1, "statusCode": 201, "sessionToken": "-1#9032" }
  ]
}
```

The SDK was initially under the impression that the coordinator would send the fully-assembled canonical `{pkRangeId}:{lsn}` token directly in `sessionToken`. We later confirmed the coordinator **only sends the raw LSN vector** in this field — never the canonical form.

## Problem

`SessionContainer.SetSessionToken()` does `tokenParts = token.Split(':'); SessionTokenHelper.Parse(tokenParts[1])` unconditionally. An LSN-only token (no `:`) produces a single-element array → `tokenParts[1]` throws `IndexOutOfRangeException` → a **committed** transaction surfaces as **failed** to the caller.

## Fix (PR #5856)

Update the DTC contract so the coordinator sends `partitionKeyRangeId` as a separate optional field per operation, and the SDK assembles the canonical form in `DistributedTransactionOperationResult.FromJson()`:

```json
{
  "operationResponses": [
    { "index": 0, "statusCode": 201, "sessionToken": "-1#5083", "partitionKeyRangeId": "345" },
    { "index": 1, "statusCode": 201, "sessionToken": "-1#9032", "partitionKeyRangeId": "346" }
  ]
}
```

Assembly logic (in `FromJson`):

| `sessionToken`                                                      | `partitionKeyRangeId` | Result                              |
|---------------------------------------------------------------------|-----------------------|-------------------------------------|
| Already canonical (colon at position > 0, content on both sides)    | any                   | Left as-is                          |
[...]

PR #5958, opened 2026-06-14
DistributedTransaction: Fixes session-token handling for committed transactions

Meghana-Palaparthi:
## Motivation

After a Distributed Transaction (DTC) commit, the coordinator returns per-operation session tokens in the JSON body (a DTC spans multiple partitions, so there is no single `x-ms-session-token` header). These tokens must be merged into the SDK's `SessionContainer` so later session-consistent reads don't get `ReadSessionNotAvailable` (404/1002).

Relative to the SDK point-operation path, the DTX commit path had several gaps: malformed tokens were silently skipped, there was no capture-side recovery when a partition split/moved between the client's routing resolve and the server commit, and the single-master write-gate wasn't honored.

## What changed

- **Malformed-token handling is now shared, strict, and deterministic.** A single `{pkRangeId}:{lsn}` shape-and-parse validator (`SessionContainer.IsCanonicalSessionTokenShape` / `IsValidSessionToken`) is centralized in `SessionContainer` and reused outside the write path by DTX, replacing the previous catch-and-classify approach. On a committed response under Session consistency, the first malformed token throws a non-retriable `CosmosException`; otherwise it's traced best-effort.
- **Capture-side partition-move recovery, via a shared primitive.** When the server serves a different partition key range than the client resolved, the routing cache is force-refreshed once per moved range. The move-detection core is extracted into `PartitionKeyRangeCache.RefreshRoutingCacheIfPartitionMovedAsync` and shared by both the point-op path (`GatewayStoreModel.CaptureSessionTokenAndHandleSplitAsync`) and DTX, so the two no longer drift behind hand-maintained parity comments. A refresh failure is traced and swallowed: the commit already succeeded, so the response is always returned to the caller.
- **Single-master write-gate parity.** A write sub-op on a single-master account is not assigned a session token (matching `GatewayStoreModel`); reads and multi-master writes still get one.
- **Per-partition auto-resolution** of each op's token from the concrete `SessionContainer`, with the resolved range recorded so split detection covers gated writes too.

## Testing

**249/249 DTX unit tests pass** (`dotnet test --filter DistributedTransaction -p:IsPreview=true`); 66/66 `SessionContainer + GatewayStoreModel` regression tests pass; Release build is clean.

The E2E suite (`[Ignore]`d, requires a live DTX endpoint via `COSMOS_DTX_ENDPOINT`/`COSMOS_DTX_KEY`) was run manually against a live account: all **write**-path tests pass and confirm write tokens are canonical `{pkRangeId}:{lsn}`.

### Unit tests — capture-side partition-move recovery

| Test | Verifies |
| --- | --- |
| `MergeSessionTokens_ForcesRefresh_WhenServerRangeDiffers` | Routing cache force-refreshed when server range ≠ resolved range |
| `MergeSessionTokens_NoRefresh_WhenRangesEqual` | No refresh when server and resolved ranges match |
| `MergeSessionTokens_NoRefresh_WhenServerRangeMissing` | Missing server range skips refresh |
| `MergeSessionTokens_NoRefresh_WhenResolvedRangeMissing` | Missing resolved range skips refresh |
| `MergeSessionTokens_NoRefresh_WhenCacheNull` | Null cache skips refresh without error |
| `MergeSessionTokens_NoRefresh_ForFailedSubOp` | Non-success sub-ops skipped entirely |
| `MergeSessionTokens_RefreshesOncePerDistinctRange` | N ops on the same moved range trigger one refresh |
| `MergeSessionTokens_RefreshException_IsSwallowed_ResponseReturned` | Post-commit refresh failure is swallowed and traced; the committed response is still returned |

### Unit tests — single-master write-gate

| Test | Verifies |
| --- | --- |
| `WriteGate_SingleMaster_WriteSubOp_NoToken` | Write op gets no token on single-master account |
| `WriteGate_SingleMaster_ReadSubOp_AppliesToken` | Read op still gets token on single-master account |
| `WriteGate_MultiMaster_WriteSubOp_AppliesToken` | Write op gets token on multi-master account |
| `WriteGate_UserSuppliedToken_AlwaysPreserved` | Explicit user token is never overridden |

### Unit tests — merge behavior (happy path)

| Test | Verifies |
| --- | --- |
| `CommitTransactionAsync_MergesSessionTokensIntoSessionContainer` | Valid canonical token merged via SetSessionToken |
| `CommitTransactionAsync_PassesCorrectCollectionToSetSessionToken` | Correct collectionRid + full name passed |
| `CommitTransactionAsync_MergesSessionTokens_ForReadOperations` | Read-op tokens merged like writes |
| `CommitTransactionAsync_SamePartition_MergesLatestToken` | Multiple ops on one partition merge to the latest token |
| `CommitTransactionAsync_MergedTokens_UsableInSubsequentPointReads` | Merged state is usable by regular point reads |
| `MergeSessionTokens_ConcurrentInvocations_DoNotCorruptSessionContainer` | Thread-safety under concurrent merges |
| `MergeSessionTokens_NonSessionConsistency_DoesNotThrow_AndMergesValidOps` | Under non-Session consistency, valid ops merge and no throw |

### Unit tests — skip conditions (no merge, no throw)

| Test | Verifies |
| --- | --- |
| `CommitTransactionAsync_SkipsMerge_WhenSessionTokenIsNull` | Null/empty token silently skipped |
| `CommitTransactionAsync_SkipsMerge_OnFailureResponse_EmptyTokens` | Empty tokens on a failed commit skipped |
| `MergeSessionTokens_SkipsMerge_WhenCollectionResourceIdIsNull` | Missing collectionRid skips merge (no crash) |
| `MergeSessionTokens_SkipsFailedOp_WithNonEmptyMalformedToken_DoesNotThrow` | Non-success sub-op with a malformed token is skipped, not thrown |
| `MergeSessionTokens_OutOfRangeIndex_SkippedAndTraced` | Out-of-range op index skipped and traced |
| `CommitTransactionAsync_MergesSessionTokens_OnlyOnSuccessAttempt` | Tokens merged only on the successful attempt |

### Unit tests — malformed token detection and error handling

| Test | Verifies |
| --- | --- |
| `CommitTransactionAsync_Throws_WhenSessionTokenIsNotCanonical` | Non-canonical token throws CosmosException |
| `CommitTransactionAsync_Throws_WhenPartitionKeyRangeIdIsEmptyOrWhitespace` | Empty/whitespace pkRangeId segment is malformed |
| `MergeSessionTokens_LegacyTwoFieldFormat_TreatedAsMalformed` | Colon-less LSN-only token is malformed |
| `MergeSessionTokens_ShapeValidButContentInvalid_ClassifiedAsMalformed` | Passes shape check but fails parse → malformed |
| `CommitTransactionAsync_MultiOp_ThrowsOnFirstMalformed_WithoutMergingLaterOps` | Throws on first malformed op; later ops not merged |
| `MergeSessionTokens_SetSessionTokenThrows_ThrowsOnFirstWithoutMergingLaterOps` | SetSessionToken parse failure throws on first op |
| `CommitTransactionAsync_AggregatesSetSessionTokenException` | SetSessionToken failure surfaced as CosmosException |
| `CommitTransactionAsync_MalformedToken_DoesNotTriggerRetry` | Malformed-token exception is not retried |

### Unit tests — exception propagation

| Test | Verifies |
| --- | --- |
| `CommitTransactionAsync_PropagatesOperationCanceledException_FromSetSessionToken` | OperationCanceledException propagates immediately |
| `MergeSessionTokens_OperationCanceledException_Propagates` | OCE is not classified as malformed |
| `MergeSessionTokens_InternalServerErrorFromRace_PropagatesUnchanged` | Benign concurrent-add 500 propagates, not treated as malformed |

### Unit tests — auto-resolution from SessionContainer

| Test | Verifies |
| --- | --- |
| `ResolveCollectionRidsAsync_ResolvesCompoundToken_WhenRoutingMapUnavailable` | Falls back to compound token when routing map is unavailable |
| `ResolveCollectionRidsAsync_PreservesUserProvidedToken` | Explicit user token is not overridden |
| `CommitTransactionAsync_WriteTokensMerged_ThenAutoResolvedIntoSubsequentRead` | Auto-resolution flows merged tokens into a later read |
| `CommitTransactionAsync_ExplicitSessionTokenOnRead_TakesPrecedenceOverAutoResolved` | Explicit token wins over auto-resolved |
| `CommitTransactionAsync_ValidatesSessionToken_InBothWriteAndReadResponses` | Validation applies to read and write responses |

### Unit tests — per-operation token serialization & round-trip

| Test | Verifies |
| --- | --- |
| `CommitTransactionAsync_PerOperationSessionToken_IsSerializedInRequestBody` | Token serialized into request JSON |
| `CommitTransactionAsync_NoPerOperationSessionToken_OmitsFieldFromRequestBody` | Absent token omitted from JSON |
| `CommitTransactionAsync_WhitespaceOrEmptyPerOperationSessionToken_OmitsFieldFromRequestBody` | Whitespace/empty token omitted |
| `CommitTransactionAsync_MultipleOperations_EachCarriesOwnSessionToken` | Each op serializes its own token |
| `CommitTransactionAsync_UserCanExtractAndPassBackSessionToken` | Public SessionToken round-trips through request options |
| `OperationResult_SessionTokenAccessible_AfterFailedOpStatus` | SessionToken readable even after a failed op status |

### E2E tests (live DTX endpoint, `[Ignore]`d until emulator supports DTX)

| Test | Status | Notes |
| --- | --- | --- |
| `DtxWrite_CommitReturnsCanonicalSessionTokens` | ✅ Pass | Write tokens are canonical `{pkRangeId}:{lsn}` |
| `DtxWrite_ThenPointRead_Succeeds` | ✅ Pass | Write tokens merged; session-consistent point read succeeds |
| `DtxWrite_MultiOp_AllOpsReturnTokens` | ✅ Pass | Each write op carries its own token |
| `DtxWrite_MultiPartition_PerPartitionTokensReturnedByCoordinator` | ✅ Pass | Multi-partition write tokens canonical |
| `DtxWrite_MultiPartition_ThenPointReads_Succeed` | ✅ Pass | Per-partition session-consistent reads succeed |
| `DtxWrite_MultiContainer_SessionTokensMergedPerCollection` | ✅ Pass | Tokens merged per collection |
| `DtxWrite_MultiDatabase_SessionTokensMergedPerDatabase` | ✅ Pass | Tokens merged per database |
| `DtxCommit_FailedTransaction_TokenEmissionBehavior` | ✅ Pass | Failed transactions handled |
| `DtxWrite_ThenDtxRead_Succeeds` | ❌ Fail | Coordinator bug: read commit returns bare vector token (no `{pkRangeId}:` prefix) |
| `DtxRead_OnlyResponse_ReturnsSessionTokens` | ❌ Fail | Same coordinator bug |
| `DtxRead_WithFutureSessionToken_BehavesPredictably` | ❌ Fail | Same coordinator bug |
| `DtxWrite_TokenRoundTrip_IntoSubsequentDtxRead` | ❌ Fail | Same coordinator bug |
| `DtxWrite_MultiPartition_AutoResolvedToken_ReadSucceeds` | ❌ Fail | Same coordinator bug |
| `DtxRead_WithInvalidPkRangeId_GracefulBehavior` | ❌ Fail | Same coordinator bug |
| `DtxRead_WithGarbledToken_DoesNotCrash` | ❌ Fail | Same coordinator bug |

## Changelog

Entries added under `Unreleased`.


NaluTripician (commented):
## 🔍 Review — Fixes silent skip of malformed session tokens

Thanks for tackling this — turning silent session-staleness into a loud signal is the right instinct. After a deep pass (correctness + adversarial), I have two blocking concerns and several recommendations. The core risk: the change ties correctness to a coordinator contract that doesn't appear to be deployed yet, and signals a post-commit problem as a retriable HTTP 500 on an **already-committed** transaction.

### 🔴 Blocking

**1. Removing the `pkRangeId:lsn` assembly breaks every token-returning commit until the coordinator change ships.**
`DistributedTransactionOperationResult.FromJson` previously assembled a canonical token from a bare-LSN `sessionToken` + the separate `partitionKeyRangeId` field. That logic is deleted, so a bare-LSN token now has no colon → classified malformed → aggregated `CosmosException(500)`. The PR's own *deleted/updated* tests document the current wire format:
- *"session token is LSN-only (as always returned by the endpoint)"*
- `TODO(issue#5857): Remove this test once the coordinator is updated to emit partitionKeyRangeId`

That reads as "the server still returns bare-LSN today; #5857 is open." If this SDK ships first, session-consistency DTX commits that return a token will throw a 500 even though the transaction committed — and the only test that would catch it (`DtxWrite_CommitReturnsCanonicalSessionTokens`) is `[Ignore]`'d, so CI stays green. `result.PartitionKeyRangeId` is now parsed but unused, which corroborates that the assembly was load-bearing.

*Suggestion:* keep the pkRangeId-based assembly as a fallback (assemble when the token lacks a canonical colon but `PartitionKeyRangeId` is present; only treat as malformed when neither is usable). Gate removal of the fallback on confirmed rollout of #5857.

**2. `CosmosException(InternalServerError 500)` is a retriable status used for a non-retriable, already-committed condition.**
The only guard against retry is the message text `"DO NOT retry the transaction."` Application resilience layers (Polly, generic 5xx retry, service mesh) key on status code, not message. The SDK itself won't retry (the throw bypasses the DTX retry loop), but an app that retries on 5xx will build a new transaction with a fresh idempotency token and **double-apply a transaction that already committed** — the worst failure class for a transaction API.

*Suggestion:* don't reuse a retriable HTTP code for this. Either don't throw for post-commit bookkeeping (return the successful response with the failure attached as a diagnostic), or surface a dedicated, clearly non-retriable signal (distinct exception type / substatus) callers can pattern-match.

### 🟡 Recommendations

**3. The throw path discards the committed transaction's response — contradicting its own comment.** The comment says *"Response is not disposed: the caller may still need operation results,"* but on the malformed path `MergeSessionTokens` throws, so the caller only gets the `CosmosException` (ActivityId + RequestCharge). This is worst for **read DTX** — the read payloads live in `response[i].ResourceStream` and are lost to a cosmetic bookkeeping failure.

**4. Whitespace-only token now 500s instead of being skipped.** The skip guard uses `string.IsNullOrEmpty`, which doesn't catch `"   "`; the old `FromJson` normalized whitespace→null. Use `string.IsNullOrWhiteSpace` to restore prior tolerance.

**5. `MergeSessionTokens` runs on every attempt, including retriable responses.** It's called before the success/retriable classification. A malformed token on a 503/retriable response would abort the retry loop with a terminal 500 whose "committed successfully" message is then factually wrong. Gate the merge on `response.IsSuccessStatusCode`.

**6. Auto-resolution injects the full compound multi-partition collection token into every per-op token.** `concreteSessionContainer.GetSessionToken(collectionPath)` returns the collection-wide compound token (comma-joined `pkRangeId:lsn` across partitions) and assigns the whole string to every op in the group, regardless of target partition. This diverges from `GatewayStoreModel`, which resolves the partition-local token. The "coordinator extracts the per-partition token internally" contract has no enabled SDK test.

**7. New unit tests assert the new throw against a mocked wire format (circular), and the tests that proved graceful handling of the legacy shape were deleted rather than converted.** Missing CI-executable coverage for: malformed token on a retriable response (#5), whitespace-only token (#4), compound multi-partition auto-resolution (#6), and explicit empty-string token (#8). Recommend running `DtxWrite_CommitReturnsCanonicalSessionTokens` against a live DTX account and recording the result before merge.

### 🟢 Suggestions

**8.** Auto-resolution guard uses `operation.SessionToken == null`, so an explicit `SessionToken = ""` is treated as "provided" and silently disables enforcement — consider `string.IsNullOrEmpty`.
**9.** Remove or repurpose the now-dead `PartitionKeyRangeId` parse in `FromJson` (or restore it to the assembly fallback per #1).
**10.** Tests assert on free-text exception-message substrings (brittle); prefer asserting `StatusCode`/`SubStatusCode`. The concurrency test only checks the final token is non-empty and `StartsWith("0:")`, so it can't fail for realistic corruption.

---

Net: I'd hold until #1 and #2 are resolved. The safest sequencing is to keep the non-throwing assembly fallback (or ship log-only first) until the coordinator change in #5857 is confirmed deployed, then flip to strict validation with a non-retriable error signal.

<sub>🤖 Generated with deep + adversarial review assistance.</sub>

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -70,7 +70,11 @@ await DistributedTransactionCommitterUtils.ResolveCollectionRidsAsync(
             }
             catch (Exception ex) when (ex is not OperationCanceledException)
             {
-                DefaultTrace.TraceError($"Distributed transaction failed: {ex.Message}");
+                // Neutral wording: the server-side commit may have succeeded even when this
+                // pipeline throws (e.g., post-commit session-token bookkeeping surfacing a
+                // coordinator contract violation). Callers should rely on the exception type
+                // and message — not this trace — to decide whether to retry.
+                DefaultTrace.TraceError($"Distributed transaction commit pipeline threw: {ex.Message}");
```
kushagraThapar: I liked the previous wording, it was more generic, why changing it here?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -197,57 +204,96 @@ internal static void MergeSessionTokens(
...
+                        $"Op {result.Index}: SetSessionToken failed for '{DistributedTransactionCommitter.TruncateForLog(result.SessionToken)}' ({ex.GetType().Name}: {ex.Message})");
+                }
+            }
+
+            // After merging all valid tokens, surface aggregated failure as CosmosException.
+            // The transaction has already committed — this is a post-commit bookkeeping error.
+            // Message explicitly states "committed successfully" so callers do not retry.
+            if (malformedErrors != null)
+            {
+                throw new CosmosException(
```
kushagraThapar: What's the point of throwing an exception here? What do we expect users to do at this point?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -197,57 +204,96 @@ internal static void MergeSessionTokens(
...
+                    malformedErrors ??= new List<string>();
+                    malformedErrors.Add(
+                        $"Op {result.Index}: SetSessionToken failed for '{DistributedTransactionCommitter.TruncateForLog(result.SessionToken)}' ({ex.GetType().Name}: {ex.Message})");
+                }
+            }
+
+            // After merging all valid tokens, surface aggregated failure as CosmosException.
+            // The transaction has already committed — this is a post-commit bookkeeping error.
+            // Message explicitly states "committed successfully" so callers do not retry.
+            if (malformedErrors != null)
```
kushagraThapar: What if malformedErrors have non-retriable errors or non-transient errors? We are still calling distributed transaction commit successful. Isn't that an issue?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -197,57 +204,96 @@ internal static void MergeSessionTokens(
...
         {
-            // Mirror the pattern used by GatewayStoreModel.CaptureSessionTokenAndHandleSplitAsync.
-            // after a response is received, store each operation's session token in the SessionContainer
-            // so that subsequent Session-consistency reads on the affected collections can use the latest token
-            // without getting ReadSessionNotAvailable.
-            //
-            // DTC spans multiple collections so the server embeds per-operation session tokens in the JSON body.
-            // DistributedTransactionOperationResult.FromJson assembles each token into canonical SDK session-token
+            // Mirror GatewayStoreModel.CaptureSessionTokenAndHandleSplitAsync -> SessionContainer.SetSessionToken:
+            // absent/empty tokens are silently skipped; valid tokens are merged into the SessionContainer.
```
jeet1995: When can session tokens be absent/empty?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -197,57 +204,96 @@ internal static void MergeSessionTokens(
...
 
             for (int i = 0; i < response.Count; i++)
             {
                 DistributedTransactionOperationResult result = response[i];
 
-                DistributedTransactionOperation operation = null;
-                try
+                // Defensive: server controls result.Index. A malformed response with an
+                // out-of-range index must not crash the loop or skip remaining ops.
+                if (result.Index < 0 || result.Index >= serverRequest.Operations.Count)
```
jeet1995, line 298: This defensive check is not specific to session token parsing - if we need it then it should be a utility (session consistency agnostic)

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -172,6 +176,9 @@ private async Task<DistributedTransactionResponse> ExecuteCommitAsync(
                             attemptTrace,
                             cancellationToken);
 
+                        // MergeSessionTokens may throw CosmosException for malformed tokens.
+                        // Response is not disposed: the transaction committed and the caller
+                        // may still need operation results. MemoryStream has no unmanaged resources.
                         DistributedTransactionCommitter.MergeSessionTokens(
```
jeet1995: Are we calling MergeSessionToken when the effective is not session consistency too?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
```diff
@@ -197,57 +204,96 @@ internal static void MergeSessionTokens(
...
+                headers.Clear();
+                headers[HttpConstants.HttpHeaders.SessionToken] = result.SessionToken;
 
-                    if (string.IsNullOrEmpty(result.SessionToken) || string.IsNullOrEmpty(operation.CollectionResourceId))
+                // Validate canonical shape up front; SetSessionToken validates content internally.
+                // Both failures funnel through the same catch block for uniform error reporting.
+                // Catch per-op so one bad token doesn't abort merging the rest of the batch.
+                try
+                {
+                    int colonIndex = result.SessionToken.IndexOf(':');
```
jeet1995: Please ensure the session token format returned by the coordinator is appropriately captured in the contract document.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
jeet1995: @kushagraThapar Our idea was to stamp error messages and throw for internal debugging where we can isolate this issue to some partitions - not really for users. My thought process is to consider the transaction as failed from an end-to-end perspective as both the SessionContainer and session token we bubble up in our result are effectively "incomplete" so our session consistency guarantees are violated - will bring this up in our DTS sync.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTransaction/DistributedTransactionE2ETests.cs
```diff
@@ -0,0 +1,299 @@
...
+    ///     set COSMOS_DTX_ENDPOINT=https://your-account.documents.azure.com:443/
+    ///     set COSMOS_DTX_KEY=your-master-key
+    ///     dotnet test --filter "FullyQualifiedName~DistributedTransactionE2ETests"
+    ///
+    /// Remove the [Ignore] attribute before running.
+    /// </summary>
+    [TestClass]
+    [DoNotParallelize]
+    [Ignore("DTX endpoint not yet available in emulator. Remove to run locally with env vars.")]
+    public class DistributedTransactionE2ETests
```
jeet1995, line 33: Please add below tests:
1. Multi-partition container.
2. Multi-container multi-partition
3. Multi-database
4. Round tripping session token from DTx result into DTx request.
5. "Bad" session token in the request options - very high LSN/wrong partitionKeyRangeId - are 404-1002s handled gracefully by coordinator; does the SDK correctly drop the session token if the LSN can't be mapped to a partition.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
jeet1995: My feedback here to @Meghana-Palaparthi is to be concise - the error message here is super long - we can simply print offending session token, partitionKeyRangeId and activityId. Also, session container is an implementation detail which the customer can't predict the state of.

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
jeet1995: @Meghana-Palaparthi Our exception catch logic here should be scoped to malformed session token - our aim is to protect session consistency guarantees. Also, explore SessionContainer#SetSessionToken and try to reuse some of the format issue handling there.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/DistributedTransaction/DistributedTransactionE2ETests.cs
jeet1995: 1. Can we ensure we have guards to verify if result.Index is not Out Of Range or duplicate?

Meghana-Palaparthi:

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
Meghana-Palaparthi: Yeah, we always call MergeSessionTokens irrespective of consistency level. In non-DTx paths, SetSessionToken() is always called and not gated on consistency level at all. I'm following the same pattern. 

I've updated the PR to not throw when the effective is not session consistency

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
Meghana-Palaparthi: Reverted it to previous wording.

Microsoft.Azure.Cosmos/src/SessionContainer.cs
```diff
@@ -128,6 +146,92 @@ private static string GetSessionToken(SessionContainerState self, string collect
...
+                    {
+                        // A partition can have more than 1 parent (merge). In that case, we apply Merge to generate a token with both parent's max LSNs
+                        parentSessionToken = parentSessionToken != null ? parentSessionToken.Merge(parentToken) : parentToken;
+                    }
+                }
+
+                // When we don't have the session token for a partition, we can leverage the session token of the parent(s)
+                if (parentSessionToken != null)
+                {
+                    return partitionKeyRangeId + SessionContainer.sessionTokenSeparator + parentSessionToken.ConvertToString();
```
kushagraThapar, line 228: Could we confirm the multi-parent semantics here before merging? We combine both parent vectors and serialize the result under the target `partitionKeyRangeId`. If those LLSNs still represent parent-range progress, this could make them appear to be target-range progress. Would it make sense to preserve the source identities (or use the canonical service representation) and add a divergent-parent test that proves the next request carries a valid causal token after a merge?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionSessionTokenResolver.cs
```diff
@@ -0,0 +1,282 @@
...
+            {
+                partitionKeyRangeCache = await clientContext.DocumentClient.GetPartitionKeyRangeCacheAsync(NoOpTrace.Singleton);
+            }
+            catch (Exception ex) when (!(ex is OperationCanceledException))
+            {
+                DefaultTrace.TraceWarning(
+                    "DistributedTransaction could not obtain PartitionKeyRangeCache; auto session-token " +
+                    "resolution is disabled for this commit. Operations get no auto-resolved token. Exception: {0}",
+                    ex.Message);
+                return null;
```
kushagraThapar, line 83: Could we avoid treating these metadata failures as a best-effort no-token path? If the session container already has progress, returning `null` after a cache or routing lookup failure sends Session operations without that causal token; similarly, defaulting an unknown multi-write capability to single-master can omit the token from a multi-master write. That can weaken read-your-writes or write ordering during transient metadata failures. Could we fail or refresh instead whenever cached progress exists, and cover these failure paths with tests asserting that no tokenless request is issued?

PR #5959, opened 2026-06-15
Client Encryption: Fixes Stream JSON processor compatibility and data-corruption defects

adamnova:
## Summary

A compatibility analysis of the **unreleased** System.Text.Json ("Stream") JSON processor in `Microsoft.Azure.Cosmos.Encryption.Custom` — measured against the latest released version `1.0.0-preview08` — found data-corruption and backwards-compatibility defects. This PR fixes all of them and adds the cross-processor regression tests that pin them.

> The `Stream` processor is `net8.0`-only and opt-in via `RequestOptions.Properties["encryption-json-processor"]`; it has **never shipped**. The default path remains the Newtonsoft processor on every entry point and TFM. The Newtonsoft path was verified fully wire-compatible with `preview08` in both directions (including downgrade) and is unchanged in behavior.

## What was wrong & what changed

**Stream-processor data corruption (would have shipped in preview09):**
- **Double-escaping** — pass-through strings and property names containing JSON escapes (`"`, `\`, control chars, `\uXXXX`) were re-emitted from the raw, still-escaped reader span and escaped a second time, corrupting data at rest and on read, including inside encrypted object/array payloads. Fixed with `ValueIsEscaped`-aware `CopyString` emission at all encrypt/decrypt sites.
- **Depth-unaware path matching** — names matching `PathsToEncrypt`/`_ep` were matched at any depth and even while an encrypted payload was open, producing wrong/duplicate `_ep` entries, permanently undecryptable ciphertext, decrypt hard-failures on valid documents, and deletion of nested user properties named `_ei`. Now strictly gated to the top level (`CurrentDepth == 1`, no open payload), both directions.
- **`null` inside an encrypted container** wiped the pending property name, corrupting `_ep` and rendering the document undecryptable on every SDK version. Now only resets when no payload is buffering (Newtonsoft parity).
- Encrypting a document that already has a top-level `_ei` now throws a clear error (parity with Newtonsoft) instead of emitting duplicate keys.

**Backwards-compatibility restoration (affects the next release for real consumers):**
- **Custom `Encryptor` / `DataEncryptionKey` subclasses** written against `preview08` were broken: members added since were `abstract` (compile break + `TypeLoadException`) and custom `EncryptAsync`/`DecryptAsync` overrides were silently bypassed on the MDE path. These members are now `virtual` with safe defaults, and custom encryptors/keys are again dispatched through their own overrides (legacy fallback is byte-identical to the `preview08` processor). The `Stream` path fails fast with a clear `NotSupportedException` for unsupported custom encryptors rather than bypassing them.
- **Cosmos dependency** was referenced as `3.41.0-preview.0` unconditionally; the stable/preview conditional split is restored so non-preview builds pull the stable core SDK.

**Number-handling parity & misc:**
- Integer literals outside `Int64` range are now rejected by the Stream processor (were silently lossy-coerced to `double`); non-finite doubles are now rejected by the Newtonsoft processor. Decrypted integral doubles are written as `5.0` so the value's type no longer flips across encrypt/decrypt cycles.
- Legacy-algorithm (format 2) documents through the Stream decrypt API now produce an actionable error naming the legacy format and the Newtonsoft processor.
- A `_ei` property holding a non-object value is now passed through unchanged by the Stream processor (parity with Newtonsoft / original released behavior).

## Verification

- New tests: cross-processor round-trip matrix, `preview08`-decryptor downgrade simulation, depth/escape/null edge cases, custom-encryptor dispatch, number-handling parity, and a binary cross-version ciphertext fixture (fixed key, real MDE AEAD, version-byte pinning). All previously bug-pinning `KnownFailure_*` tests are flipped to passing regression tests.
- **net8.0: all tests pass; net6.0: all pass except 1 pre-existing, unrelated failure** (`CallerArgumentExpression`/`ParamName` on the netstandard build).
- `src` builds clean (0 warnings, `TreatWarningsAsErrors` on) for both TFMs and both stable/preview flavors.
- API contract baselines updated to reflect only the `abstract → virtual` change — **zero public-member removals vs `preview08`**.

## Residual risks (need a product decision — not addressed here)

- **HIGH** — MDE dependency is the prerelease `Microsoft.Data.Encryption.Cryptography 2.0.0-pre015`; byte-level interop with `1.2.0`-produced ciphertext cannot be proven inside this repo. Recommend validating before release.
- **LOW** — `Newtonsoft.Json` is now `PrivateAssets="All"` while some public surface still exposes `JToken`; consumers relying on the transitive dependency would need a direct reference.

## Changelog

Customer-facing entries added under `### 1.0.0-preview09 - Unreleased` → `#### Fixes` in `Microsoft.Azure.Cosmos.Encryption.Custom/changelog.md`.


adamnova:
Closing - the genuine Stream-processor data-corruption fixes from this PR have been absorbed into #5985 with TDD fail-before/pass-after tests: double-escape of pass-through strings and property names, null-inside-encrypted-container (undecryptable payload), encrypted-path matching at any depth, integral-double fidelity, out-of-range integer rejection, duplicate/non-object _ei handling, and numeric-string _ef parity. The compatibility matrix (#5986) was hardened to catch these classes. Defects already handled in main (non-finite throw) were intentionally not changed.

PR #5960, opened 2026-06-15
[Internal] Thinclient integration: Adds separate storemodel for thinclient.

### LINKED ISSUE #5684: Thin Client Integration] Separate thinclient logic from gateway logic in GatewayStoreModel.

aavasthy:
# Pull Request Template

## Description

This change reintroduces ThinClientStoreModel as a subclass of GatewayStoreModel and removes all thin-client routing logic from the base class. Previously, thin-client dispatch lived inside GatewayStoreModel alongside the regular gateway path, which made the class harder to reason about and blurred the responsibility boundary between the two transports. With this refactor, GatewayStoreModel once again handles only the gateway HTTP path, and ThinClientStoreModel owns its own ThinClientStoreClient, pre-resolves the partition key range for proxy headers, and falls back to the inherited gateway path on the same instance whenever the operation isn't thin-client-routable or the service withdraws thin-client locations mid-flight.

The dispatch behavior is preserved verbatim. The bidirectional fallback contract added in #5927 (https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5927) — where the same store-model instance switches between thin-client and gateway as LocationCache.HasThinClientRead/WriteLocations flips on each account refresh, with no client restart required — still works at all three layers (LocationCache volatile flags, per-request IsThinClientRoutable check in ProcessMessageAsync, per-retry re-evaluation in ClientRetryPolicy.OnBeforeSendRequest). PPAF/PLCB integration, IChaosInterceptor plumbing for fault injection, dispose semantics, and the 410/449/Conflict exception capture paths all behave identically to the pre-refactor code.

DocumentClient now constructs ThinClientStoreModel only when running in Gateway mode with thin-client enabled, and assigns it to StoreModel; otherwise StoreModel = GatewayStoreModel as before. Call sites in ClientRetryPolicy and GlobalPartitionEndpointManagerCore were updated to reference the static helpers on their new home (ThinClientStoreModel.IsThinClientRoutable / IsThinClientReadRoutable). A dedicated ThinClientStoreModelTests class was added with 19 logical, equivalence-class-driven cases (using DataRow where it cut redundancy) covering the operation-support matrix, direction-aware routability, end-to-end dispatch, mid-flight fallback, PPAF wiring, dispose, and exception propagation; the corresponding thin-client tests were removed from GatewayStoreModelTest.cs to keep ownership clean. Full unit suite and all touched class tests pass with zero failures.

## Type of change

## Closing issues

To automatically close an issue: closes #5684

## Changelog

>This change is pure internal refactor with no observable customer impact.


Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
```diff
@@ -0,0 +1,275 @@
...
+                }
+
+                bool canUseThinClient = this.thinClientStoreClient != null
+                    && ThinClientStoreModel.IsThinClientRoutable(this.endpointManager, request);
+
+                Uri physicalAddress = ThinClientStoreClient.IsFeedRequest(request.OperationType)
+                    ? this.GetFeedUri(request)
+                    : this.GetEntityUri(request);
+
+                if (canUseThinClient)
```
kundadebdatta: There is a good amount of code duplication in both the store models (Gateway and ThinClient) and it's currenty not leveraging the benefit of the inheritence. One way to streamline this will be to have the `ProcessMessageAsync` in the base layer, and call the `DispatchAsync` from that layer to choose between which store client to use from.

**DispatchAsync() implementation in GatewayStoreModel**

```
        /// <summary>
        /// Determines whether the PartitionKeyRange should be resolved up-front for the request.
        /// The base gateway path only needs it when per partition automatic failover / circuit
        /// breaker is enabled. Subclasses (e.g. ThinClientStoreModel) can override to always
        /// resolve it when they need the PKR for their own dispatch.
        /// </summary>
        protected virtual bool ShouldResolvePartitionKeyRange()
        {
            return this.IsPartitionLevelFailoverEnabled();
        }

        /// <summary>
        /// Dispatches the request to the underlying transport. The base implementation sends the
        /// request through the gateway HTTP path. Subclasses (e.g. ThinClientStoreModel) override
        /// this seam to route to a different transport, while reusing the shared session-token,
        /// partition-key-range and split-handling logic in <see cref="ProcessMessageAsync"/>.
        /// </summary>
        protected virtual Task<DocumentServiceResponse> DispatchAsync(
            DocumentServiceRequest request,
            Uri physicalAddress,
        {
            return this.gatewayStoreClient.InvokeAsync(
                request,
                request.ResourceType,
                physicalAddress,
                cancellationToken);
        }
```

**DispatchAsync() implementation in ThinClientStoreModel:**

```
      /// <summary>
      /// Routes the request through the thin-client store client when it is currently
      /// thin-client-routable, otherwise transparently falls back to the inherited gateway
      /// HTTP path on the same instance. The decision is taken per request so the model can
      /// switch direction (thin-client ↔ gateway) as soon as the next <see cref="LocationCache"/>
      /// refresh updates the availability signals, without requiring a client restart.
      /// </summary>
      protected override async Task<DocumentServiceResponse> DispatchAsync(
          DocumentServiceRequest request,
          Uri physicalAddress,
      {
          bool canUseThinClient = this.thinClientStoreClient != null
              && ThinClientStoreModel.IsThinClientRoutable(this.endpointManager, request);

          if (!canUseThinClient)
          {
              return await base.DispatchAsync(request, physicalAddress, cancellationToken);
          }

          Uri thinClientEndpoint = this.endpointManager.ResolveThinClientEndpoint(request);
          AccountProperties account = await this.GetDatabaseAccountPropertiesAsync();

          return await this.thinClientStoreClient.InvokeAsync(
              request,
              request.ResourceType,
              physicalAddress,
              thinClientEndpoint,
              account.Id,
              this.clientCollectionCache,
              cancellationToken);
      }
```

aavasthy: Did code cleanup and now thinclientStoreModel is leveraging methods from GatewayStoreModel instead of creating new methods.

kushagraThapar (approved):
Thanks for getting this separated out. LGTM!

jeet1995 (approved):

FabianMeiswinkel (approved):
LGTM

PR #5961, opened 2026-06-15
[Internal] DistributedTransaction: Adds cumulative delay budget to outer retry loop

Meghana-Palaparthi:
## Summary

Adds a `MaxCumulativeRetryDelay` (30 seconds) check to the `DistributedTransactionCommitter` outer retry loop, mirroring the existing `ResourceThrottleRetryPolicy` cumulative delay pattern used by non-DTX flows.

## Motivation

Previously, the outer retry loop was bounded only by attempt count (`MaxIsRetriableRetryCount = 10`). With the default exponential backoff (1s base, max exponent 5, ±25% jitter), the worst-case cumulative delay could reach ~4 minutes — far longer than is reasonable for a caller to wait.

The cumulative delay budget provides a hard time-based cap that:
- Prevents unbounded retry duration when server `RetryAfter` headers are large
- Matches the pattern already established by `ResourceThrottleRetryPolicy`
- Is the binding constraint under default settings (~4-5 retries) while the attempt-count cap (10) only fires when delays are very small

## Changes

- **`DistributedTransactionCommitter.cs`**: Added `MaxCumulativeRetryDelay` constant (30s) and cumulative delay tracking in the retry loop. Enhanced trace logs to include `cumulativeDelayMs`.
- **`DistributedTransactionCommitterTests.cs`**: Added tests for cumulative budget exhaustion and server RetryAfter-dominated budget exhaustion. Updated the existing exponential backoff test to override the budget so it can still exercise the full backoff curve.

## Testing

- All 47 `DistributedTransactionCommitterTests` pass
- Build succeeds with 0 warnings/errors

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionCommitter.cs
@@ -20,7 +20,17 @@ namespace Microsoft.Azure.Cosmos
     {
         // Outer-loop retry parameters. The inner loop (ClientRetryPolicy) handles envelope failures with empty body;
         // the outer loop handles body-bearing semantic failures whose JSON body sets isRetriable: true.
+        //
+        // Hard ceiling on outer-loop wire requests. With non-trivial retryBaseDelay the cumulative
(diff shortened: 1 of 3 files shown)


kushagraThapar (approved):

jeet1995 (approved):

PR #5829, opened 2026-05-05
Routing: Adds Gateway Account Property Flag for Dynamic Hedging Control

NaluTripician:
## Description

Adds a Gateway-controlled `disableCrossRegionalHedging` account property that lets on-call engineers dynamically disable PPAF cross-region hedging without rolling back PPAF entirely.

When the flag is `true`, the SDK clears the `AvailabilityStrategy` on the next account-properties refresh; toggling it back to `false` restores the customer-configured strategy or rebuilds the SDK default.

## Type of change

- [ ] Bug fix / mitigation lever (non-breaking; adds an operational escape hatch)
- [x] New feature
- [ ] Breaking change

## Spec

- `openspec/changes/ppaf-dynamic-hedging-control/` (proposal/design/spec/tasks)
- Reference PR: #5719 (spec only)

## Changes

- `AccountProperties`: new internal `DisableCrossRegionalHedging` (nullable bool) bound to JSON key `disableCrossRegionalHedging`.
- `GlobalEndpointManager`: `OnEnablePartitionLevelFailoverConfigChanged` is now `Action<bool, bool>`; tracks `lastKnownDisableCrossRegionalHedging` so refresh fires on either flag change.
- `DocumentClient`:
  - Captures the initial gateway flag during `GetInitializationTaskAsync` and stashes any customer-configured `AvailabilityStrategy` when the flag is true at startup.
  - `InitializePartitionLevelFailoverWithDefaultHedging` short-circuits when the flag is true.
  - `UpdatePartitionLevelFailoverConfigWithAccountRefresh` reconciles both PPAF state and the disable flag via `ApplyHedgingStrategyForCurrentState`.
  - Customer strategies are stashed/restored across toggles; SDK-default strategies are not stashed (rebuilt deterministically from PPAF state).
- 10 new unit tests in `GatewayHedgingOverrideTests`.
- Updates 2 existing serializer contract tests that pin the `AccountProperties` JSON shape.

## Behavior

| State | Result |
| --- | --- |
| flag true, customer strategy present | Strategy stashed, `AvailabilityStrategy=null` |
| flag true, SDK default present | `AvailabilityStrategy=null` (no stash; deterministic) |
| Toggle false, strategy stashed | Restored from stash |
| Toggle false, no stash, PPAF on | SDK default rebuilt |
| `DisablePartitionLevelFailoverClientLevelOverride=true` | Gateway flag ignored entirely |

## Validation

[...]

PR #5967, opened 2026-06-18
Routing: Refactors GlobalEndpointManager to track a dedicated PPAF-enablement baseline

### LINKED ISSUE #5936: Routing: Track dedicated lastKnownEnablePartitionLevelFailover baseline in GlobalEndpointManager
Follow-up from PR #5829 review (thread https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5829#discussion_r3322215815).

### Problem
In `GlobalEndpointManager.RefreshDatabaseAccountInternalAsync`, when `accountProperties.EnablePartitionLevelFailover.HasValue == false` but the hedging flag changed, `latestPpafEnabled` falls back to `this.connectionPolicy.EnablePartitionLevelFailover`:

```csharp
bool latestPpafEnabled = accountProperties.EnablePartitionLevelFailover
    ?? this.connectionPolicy.EnablePartitionLevelFailover;
```

That field is mutated by the SDK itself in `DocumentClient.UpdatePartitionLevelFailoverConfigWithAccountRefresh`. So the fallback conflates *"what the gateway said"* with *"what we last applied"*. It is functionally correct today (no external writer to `connectionPolicy.EnablePartitionLevelFailover` exists, and feeding the last-applied value back into the callback yields `ppafEnablementChanged == false`, which is the desired no-op when only the hedging flag changed). But a future writer (e.g. an admin API) would silently desync the cache.

### Proposed fix
Mirror the hedging-flag pattern: track a dedicated `lastKnownEnablePartitionLevelFailover` field in `GlobalEndpointManager` (initialized alongside `lastKnownDisableCrossRegionalHedging` in `InitializeAccountProperties`) and fall back to that instead of `connectionPolicy.EnablePartitionLevelFailover`.

### Why deferred from #5829
The change touches GEM init in two places and the PPAF-enablement path; #5829 was already large and this is forward-looking hardening with no current customer-observable impact. Scoped here to keep that PR's risk down.

Related: the F6 exception-recovery fix in #5829 (`efea7acac`) is scoped to the hedging flag only; PPAF-enablement recovery on a throwing subscriber depends on this same dedicated-baseline work.

NaluTripician:
## What & why

Resolves #5936 (follow-up hardening deferred from #5829, review thread r3322215815).

`GlobalEndpointManager.RefreshDatabaseAccountInternalAsync` used `connectionPolicy.EnablePartitionLevelFailover` both to **detect** PPAF-enablement changes and as the **fallback** when the gateway response omits the property. That field is mutated by the SDK's own subscriber (`DocumentClient.UpdatePartitionLevelFailoverConfigWithAccountRefresh`), so it conflated *"what the gateway reported"* with *"what we last applied."* The values coincide only while the SDK is the sole writer; a future external writer (e.g. an admin API) would silently desync change detection. The sibling `disableCrossRegionalHedging` flag already solved this with a dedicated baseline — PPAF enablement was the remaining un-mirrored case.

## Changes

- **`GlobalEndpointManager`**: adds a dedicated `lastKnownEnablePartitionLevelFailover` baseline, seeded in `InitializeAccountPropertiesAndStartBackgroundRefresh` from the effective post-client-override value. PPAF-enablement change detection and the absent-property fallback now read this baseline; the baseline is advanced before notifying subscribers and rolled back if a subscriber throws — an exact mirror of the existing hedging baseline.
- **`DocumentClient.UpdatePartitionLevelFailoverConfigWithAccountRefresh`** (closes the F6 PPAF-recovery asymmetry): on a throwing subscriber, reverts the applied PPAF state (`EnablePartitionLevelFailover`, `EnablePartitionLevelCircuitBreaker`, and the `PartitionKeyRangeLocation` PPAF/PPCB flags) so the GEM re-fire is not swallowed by the no-op guard and the transition is re-applied on the next refresh — mirroring the existing `disableCrossRegionalHedging` revert.
- **Tests**: adds three `GlobalEndpointManagerTest` cases — explicit transition fires once, property-dropped is a no-op, and subscriber-throw rolls back the baseline and re-detects on the next refresh.

## Behavior impact

On the non-throwing single-writer path (every customer in normal operation) behavior is unchanged — the new baseline advances in lockstep with `connectionPolicy.EnablePartitionLevelFailover`. The one deliberate delta is on the throwing-subscriber path, where a transient failure applying a PPAF-enablement change is now retried on the next account refresh instead of being lost.

## Testing

- New + existing `GlobalEndpointManagerTest` / `GatewayHedgingOverrideTests`: 29 passed.
- `AvailabilityStrategy` / `PartitionKeyRangeFailover` / `ClientRetryPolicy` suites: 111 passed.

Note: the subscriber-side revert-on-throw path is not directly unit-tested — the PPAF-enablement subscriber path NREs on a non-opened client (`PartitionKeyRangeLocation` is null) and there is no injection seam to force `ApplyHedgingStrategyForCurrentState` to throw, the same limitation already documented in `GatewayHedgingOverrideTests`. It is an exact structural mirror of the already-reviewed hedging revert; the GEM-side rollback that triggers re-detection is covered by the new exception-recovery test.

## Changelog

Added an **Other Changes** entry under `### Unreleased`.


Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -7174,10 +7177,22 @@ internal void UpdatePartitionLevelFailoverConfigWithAccountRefresh(
...
-                    // missed hedging transition rather than going permanently silent.
+                    // missed transition rather than going permanently silent.
                     this.disableCrossRegionalHedging = previousDisableCrossRegionalHedging;
+
+                    if (ppafEnablementChanged)
+                    {
+                        // Mirror the hedging revert for PPAF enablement: restore the connection-policy flags
+                        // and the partition-key-range location PPAF/PPCB state to their pre-change values so the
+                        // no-op guard at the top of this method does not swallow the GEM-re-fired transition.
+                        this.PartitionKeyRangeLocation.SetIsPPAFEnabled(previousEnablePartitionLevelFailover);
```
NaluTripician, line 7217: **M1 — Riskiest new code is untested, and the `no injection seam` justification is inaccurate.** This subscriber-side PPAF revert is the highest-risk new code in the PR and has zero coverage. All three new tests subscribe a throwing lambda directly to `OnEnablePartitionLevelFailoverConfigChanged`, so they only exercise the GEM-side baseline rollback — none drive this `catch` branch that restores `EnablePartitionLevelFailover`, `EnablePartitionLevelCircuitBreaker`, and the `PartitionKeyRangeLocation` PPAF/PPCB flags so the top-of-method no-op guard re-trips on the GEM re-fire.

The PR notes there's `no injection seam to force ApplyHedgingStrategyForCurrentState to throw`, but `GatewayHedgingOverrideTests` already drives this method directly against a real `DocumentClient`, and `*ForTests` seams already exist in this class. If a future edit ever drops one of these four restorations, a transient PPAF-enablement failure goes permanently silent and CI stays green — the exact failure mode this PR exists to prevent.

Suggest a `GatewayHedgingOverrideTests` case that forces the reconcile to throw with `ppafEnablementChanged: true`, then asserts a second call with the same `latestIsEnabled` is **not** a no-op.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -7112,11 +7112,14 @@ internal void UpdatePartitionLevelFailoverConfigWithAccountRefresh(
...
-                // above and the missed hedging transition would never retry. Scoped to the hedging flag —
-                // PPAF-enablement recovery is tracked separately.
+                // Capture the prior applied state so it can be reverted if the reconcile below throws;
+                // otherwise the GEM re-fire (which reverts its own baseline) would be short-circuited by the
+                // no-op guard above and the missed transition would never retry. Both the hedging kill-switch
+                // and the PPAF-enablement state are captured so a throwing subscriber leaves either transition
+                // re-detectable end-to-end.
                 bool previousDisableCrossRegionalHedging = this.disableCrossRegionalHedging;
+                bool previousEnablePartitionLevelFailover = this.ConnectionPolicy.EnablePartitionLevelFailover;
+                bool previousEnablePartitionLevelCircuitBreaker = this.ConnectionPolicy.EnablePartitionLevelCircuitBreaker;
```
NaluTripician, line 7144: **M3 — Step 1/Step 2 PPAF mutations are applied outside the `try`, so the revert isn't crash-consistent.** The four mutations below (`SetIsPPAFEnabled` / `EnablePartitionLevelFailover` / `SetIsPPCBEnabled` / `EnablePartitionLevelCircuitBreaker`) commit before the `try` opens, but the compensating revert lives only in the `catch` around `ApplyHedgingStrategyForCurrentState`. If a later mutation throws after an earlier one committed, the exception escapes the catch entirely: `connectionPolicy` is left partially advanced while GEM rolls its baseline back. On the next refresh GEM re-fires, but the subscriber's no-op guard now sees `connectionPolicy == latestIsEnabled` and returns early, so the half-applied state is never reconciled.

Latent today (the only realistic throw — a null `PartitionKeyRangeLocation` — fires on the first statement before any mutation), but this PR newly places PPAF/PPCB applied state into this incompletely-protected region. Consider moving these mutations inside the `try`, or documenting the load-bearing assumption that they're infallible and atomic.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -47,6 +47,14 @@ internal class GlobalEndpointManager : IGlobalEndpointManager
         // Tracked separately so the change event fires when only this flag toggles.
         private bool lastKnownDisableCrossRegionalHedging = false;
 
+        // Last observed value of the account-level EnablePartitionLevelFailover flag.
+        // Tracked separately (mirroring lastKnownDisableCrossRegionalHedging) so PPAF-enablement
+        // change detection keys off GEM's own baseline rather than connectionPolicy.EnablePartitionLevelFailover,
+        // which the subscriber (DocumentClient) mutates. This decouples "what the gateway reported"
+        // from "what we last applied" and keeps detection correct against any future external writer
+        // of connectionPolicy.EnablePartitionLevelFailover.
+        private bool lastKnownEnablePartitionLevelFailover = false;
```
NaluTripician, line 60: **M2 — Decoupling is only half-applied; the changelog claim is overstated end-to-end.** GEM detection now correctly keys off `lastKnownEnablePartitionLevelFailover`, but the subscriber's no-op guard in `DocumentClient.UpdatePartitionLevelFailoverConfigWithAccountRefresh` still reads the mutable field directly: `bool ppafEnablementChanged = this.ConnectionPolicy.EnablePartitionLevelFailover != latestIsEnabled;`.

So the changelog's `robust against any future external writer of connectionPolicy.EnablePartitionLevelFailover` guarantee is delivered by GEM but not by the subscriber: an external writer of that field would make GEM fire while the subscriber no-ops and skips side-effect reconciliation. Hypothetical today (SDK is sole writer), but the changelog promises an end-to-end guarantee the code only partially delivers. Either give the subscriber its own `lastAppliedPpafEnabled` baseline (symmetric with GEM), or soften the changelog/PR wording to state the decoupling is GEM-local.

**L2 (doc nit) just below:** the `<remarks>` on `OnEnablePartitionLevelFailoverConfigChanged` (line 64) still says the first arg `falls back to the existing connection-policy value when the property is absent` — this PR changed that fallback to `lastKnownEnablePartitionLevelFailover`. Update the comment to match.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -793,7 +806,7 @@ private async Task RefreshDatabaseAccountInternalAsync(bool forceRefresh)
 
                 bool ppafEnablementChanged = !ignorePpafChanges
                     && accountProperties.EnablePartitionLevelFailover.HasValue
-                    && (this.connectionPolicy.EnablePartitionLevelFailover != accountProperties.EnablePartitionLevelFailover.Value);
+                    && (this.lastKnownEnablePartitionLevelFailover != accountProperties.EnablePartitionLevelFailover.Value);
```
NaluTripician, line 879: **L1 — Test gaps on the core change.** The new tests cover `false->true` and property-dropped, but not the `true->false` PPAF-disable direction. Direction matters — the off-transition is what drops the SDK-default availability strategy — and the hedging suite deliberately added an `ExplicitFalseAfterTrue` companion for exactly this reason. Worth adding the PPAF mirror.

Also, the new absent-property fallback (`?? this.lastKnownEnablePartitionLevelFailover` a few lines down) is never actually propagated to a subscriber by any test — it only fires when `disableHedgingFlagChanged` is also true, a combination no test creates. And the property-dropped test asserts no-fire but never proves the baseline was *preserved* (no follow-up explicit-`false` refresh to diff against). Both are easy to close and would lock in the central behavior this PR is changing.

NaluTripician (commented):
Local deep review (PR Deep Reviewer + Skeptic Lens). No Critical/High defects — clean, well-scoped refactor that's safe to merge on the happy path. Inline comments below are hardening/test-confidence items; the one I'd genuinely close before merge is the untested subscriber-side revert (M1).

Microsoft.Azure.Cosmos/src/DocumentClient.cs
NaluTripician: Addressed in a256da84a. Moved the Step 1 / Step 2 PPAF mutations (and the hedging-flag mutation) inside the `try`, so the revert is now crash-consistent: if any individual mutation throws after an earlier one committed, the `catch` restores every captured baseline before rethrowing. The capture comment was updated to "if any mutation or the reconcile below throws".

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
NaluTripician: Addressed in a256da84a. Added three `GlobalEndpointManagerTest` cases:
- `RefreshDatabaseAccount_EnablePartitionLevelFailover_ExplicitFalseAfterTrue_FiresChangeEvent` — the missing true -> false off-direction (the transition that drops the SDK-default strategy).
- `RefreshDatabaseAccount_EnablePartitionLevelFailover_PropertyDroppedWithHedgingChange_PropagatesPreservedBaseline` — drives the `?? lastKnownEnablePartitionLevelFailover` fallback through to a subscriber (PPAF dropped, hedging flag flips so the event still fires) and asserts the propagated PPAF value is the preserved baseline (true), not an implicit false.
- `RefreshDatabaseAccount_EnablePartitionLevelFailover_PropertyDropped_PreservesBaselineForLaterExplicitFalse` — proves the baseline is preserved across a drop by following up with an explicit false that must fire a real true -> false transition.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
NaluTripician: **M2:** Took the "soften the wording" option rather than adding a second subscriber baseline. The subscriber's guard intentionally reads the actually-applied `ConnectionPolicy.EnablePartitionLevelFailover` because that field *is* its applied-state — adding a parallel `lastAppliedPpafEnabled` would risk desyncing from the field that `ApplyHedgingStrategyForCurrentState` and routing read. The changelog now states the baseline decoupling is scoped to `GlobalEndpointManager`'s change detection ("even though that field is also written by the subscriber that applies the change") rather than implying an end-to-end external-writer guarantee.

**L2:** Fixed — the `OnEnablePartitionLevelFailoverConfigChanged` `<remarks>` now says the first arg falls back to `lastKnownEnablePartitionLevelFailover` (and documents the `disableCrossRegionalHedging` fallback to `lastKnownDisableCrossRegionalHedging`).

Microsoft.Azure.Cosmos/src/DocumentClient.cs
NaluTripician: Addressed in a256da84a. Added `UpdateConfig_PpafEnablement_ReconcileThrows_RevertsStateSoRetryIsNotNoOp` to `GatewayHedgingOverrideTests`. It injects a stub `PartitionKeyRangeLocation` (so `SetIsPPAFEnabled`/`SetIsPPCBEnabled` run without an opened client) plus a new test-only `ReconcileFailureHookForTests` seam that forces `ApplyHedgingStrategyForCurrentState` to throw while `ppafEnablementChanged` is true. It then asserts all four restorations happened — connection-policy PPAF + PPCB flags reverted, and `SetIsPPAFEnabled(false)` / `SetIsPPCBEnabled(false)` verified on the stub — and that a second call with the same `latestIsEnabled` re-applies the transition rather than no-op-ing. The test fails if any of the four reverts is dropped, which is the regression mode you flagged. The inaccurate "no injection seam" justification is also gone now that the seam exists.

NaluTripician:
Pushed a256da84a addressing the self-review comments:

- **M3** — PPAF/hedging applied-state mutations moved inside the `try` so revert-on-throw is crash-consistent (a mid-sequence throw now restores every captured baseline before rethrowing).
- **M1** — added test-only seams (`ReconcileFailureHookForTests`, `PartitionKeyRangeLocationForTests`) and `GatewayHedgingOverrideTests.UpdateConfig_PpafEnablement_ReconcileThrows_RevertsStateSoRetryIsNotNoOp`, which forces the reconcile to throw with `ppafEnablementChanged: true`, asserts all four restorations, and proves a same-value retry is not a no-op. The "no injection seam" limitation no longer applies.
- **L1** — added PPAF `true -> false` transition, dropped-property fallback-propagation, and dropped-property baseline-preservation tests.
- **L2** — corrected the `OnEnablePartitionLevelFailoverConfigChanged` `<remarks>` fallback wording.
- **M2** — scoped the decoupling claim to `GlobalEndpointManager`'s change detection in the changelog; the subscriber intentionally still reconciles against the actually-applied `ConnectionPolicy.EnablePartitionLevelFailover` (its applied-state of record), so no second baseline was added.

Tests: `GatewayHedgingOverrideTests` + `GlobalEndpointManagerTest` 33 passed; `AvailabilityStrategy` / `PartitionKeyRangeFailover` / `ClientRetryPolicy` 111 passed.

FabianMeiswinkel (dismissed):
LGTM

aavasthy (dismissed):

NaluTripician:

NaluTripician:

aavasthy (approved):

kundadebdatta (approved):
LGTM.

PR #5970, opened 2026-06-22
Thin Client Integration: Adds Thinclient probe endpoint.

aavasthy:
# Pull Request Template

## Description

This change adds thinclient connectivity probe that validates each thin client (Gateway V2 / proxy) regional endpoint before the SDK routes any data-plane traffic to it. Today the SDK already discovers thin client endpoints from the account topology (getDatabaseAccount) response, but it has no way to confirm that it can actually establish an HTTP/2 connection to a given proxy region. This PR closes that gap: after every account-topology refresh, the SDK issues a POST /connectivity-probe over HTTP/2 to each newly discovered regional proxy endpoint, and only regions that respond HTTP 200 are used for data-plane traffic. Any region that has not (yet) passed its probe — un-probed, returning a non-200, or unable to negotiate HTTP/2 — transparently uses the standard gateway (Gateway V1) path for that region's traffic, with no client restart required. The probe is intentionally strict about HTTP/2: it sets an exact version policy so that if the client and proxy cannot agree on HTTP/2 the probe fails (rather than silently downgrading to HTTP/1.1), which correctly routes that region back to Gateway V1.

 Alongside the probe, thin client mode is now enabled by default (customers can opt out via the existing environment flag). The probe logic lives in a new self-contained EndpointProbeClient, is wired into GlobalEndpointManager, and is consulted at the routing site in ThinClientStoreModel.

 End-to-end thin client probe flow

   - Enablement (flags). Thin client routing activates only when feature flag (AZURE_COSMOS_THIN_CLIENT_ENABLED, now default true, opt-out) AND ConnectionMode == Gateway AND get account call returns thinclient endpoints.
   - Discovery (unchanged). Proxy regional endpoints come from the account-topology (getDatabaseAccount) response.
   - Trigger. GlobalEndpointManager runs RunThinClientProbeCycleAsync() after every topology refresh. Fire-and-forget: never throws into refresh, never blocks dispatch.
   - Endpoint set. Union of thin client read + write regional endpoints from the latest topology snapshot.
   - Delta + single-flight. Probe only endpoints not already cached healthy; all-green ⇒ no-op, zero network calls. A cycleInProgress guard blocks concurrent cycles.
   - Probe. POST /connectivity-probe over the HTTP/2 client, empty body + proxy-operation header.
   - Strict HTTP/2. Sent with exact 2.0 policy — no HTTP/2 ⇒ probe fails (no HTTP/1.1 downgrade).
   - Success = HTTP 200 only. Anything else / connection / TLS / timeout = unhealthy. 
   - Cache (permanent, per-endpoint). A 200 is cached healthy for the client's lifetime. Failures aren't cached ⇒ re-probed next refresh.
   - Routing (per request). ThinClientStoreModel checks IsProxyEndpointHealthy(endpoint); unhealthy region falls back to Gateway V1 for that region only. Failover walks (PPCB/PPAF) use AreAllThinClientReadEndpointsHealthy.
   - New regions. Uncached ⇒ probed next cycle; Gateway V1 until green.
   - No eviction. Green is never demoted by the probe; a later-degrading proxy is handled by existing ClientRetryPolicy failover.
   - 
   
## Type of change

## Closing issues

## Changelog

- [] No changelog entry is required because this PR is one of:

>


Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -123,7 +126,61 @@ public GlobalEndpointManager(
...
+        /// failover walk, which routes a whole read-endpoint list rather than a single endpoint. 
+        /// </summary>
+        public bool AreAllThinClientReadEndpointsHealthy
+        {
+            get
+            {
+                EndpointProbeClient? probeClient = this.thinClientProbeClient;
+                if (probeClient == null)
+                {
+                    return true;
```
ananth7592: Is this expected?

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,252 @@
...
+        private static HttpRequestMessage CreateProbeRequest(Uri probeUri)
+        {
+            HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Post, probeUri)
+            {
+                Version = new Version(2, 0),
+                Content = new ByteArrayContent(Array.Empty<byte>()),
+            };
+
+            // Require exact HTTP/2: fail rather than downgrade so an HTTP/2 mismatch falls back to Gateway V1.
+            EndpointProbeClient.VersionPolicyProperty?.SetValue(
```
ananth7592, line 208: One concern on the strict-HTTP/2 guarantee: on a TFM where `VersionPolicy` is null (the very reason for this reflection), only `Version = 2.0` is set with the default policy — which *permits* an HTTP/1.1 downgrade. Could a probe then return 200 over HTTP/1.1 and cache a region GREEN that can't actually serve HTTP/2 thin-client traffic, costing a per-request fail->Gateway-V1 hop until restart? Would treating an absent `VersionPolicy` as `unhealthy` be safer here?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -836,10 +898,13 @@ private async Task RefreshDatabaseAccountInternalAsync(bool forceRefresh)
...
                 }
 
-                GlobalEndpointManager.ParseThinClientLocationsFromAdditionalProperties(accountProperties);
+                GlobalEndpointManager.ParseThinClientLocationsFromAdditionalProperties(accountProperties);
 
                 this.locationCache.OnDatabaseAccountRead(accountProperties);
 
+                // Probe the thin client regional endpoints after every account-topology refresh so the
+                // routing gate reflects the latest proxy connectivity health.
+                await this.RunThinClientProbeCycleAsync();
```
ananth7592: This path `await`s the probe cycle, while the sibling trigger (~L699) is fire-and-forget (`_ =`). With up to 1+maxRetries (4) attempts x 5s `PerProbeTimeout` per endpoint, the awaited cycle can hold the background refresh for a while — but the PR description says the trigger `never blocks`. Is the asymmetry intentional (e.g. to serialize against the next refresh), or should both be fire-and-forget? A one-line comment either way would prevent a future head-scratch.

ananth7592 (commented):
Really nicely architected change. The per-region gate with a second health check at dispatch time (so a stale routing decision can't actually send traffic to an unprobed proxy) is exactly the kind of defense-in-depth that makes this safe to ship default-on — and the lifecycle details (never throws into refresh, single-flight `cycleInProgress`, drop-results-if-closed-mid-cycle) show real care. `EndpointProbeClientTests` is a pleasure to read. A couple of questions inline. Happy to discuss any of these.

changelog.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Features Added
 
+- [5970](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5970) [Internal]ThinClient: Adds an HTTP/2 connectivity probe (`POST /connectivity-probe`) that validates each discovered thin client (proxy) regional endpoint over HTTP/2. After every account topology refresh the SDK probes any newly discovered regional endpoint it has not already confirmed reachable, retrying a transient probe failure a few times before treating that region as unhealthy for the cycle, and caches each success per endpoint so a known-good region is never re-probed. Data-plane traffic routes through the thin client proxy only for regions whose endpoint has passed its probe; any region that has not yet passed (un-probed, or unable to negotiate HTTP/2) transparently uses the standard gateway for that region's traffic, with no client restart required. Thin client mode is now enabled by default (opt-out) and proxy routing additionally requires opting in to HTTP/2 via the new preview `CosmosClientOptions.EnableHttp2` option.
```
kushagraThapar: We don't need to provide implementation details in the changelog. It can be as simple as 
"Added an HTTP/2 connectivity probe (`POST /connectivity-probe`) that validates each discovered thin client (proxy) regional endpoint over HTTP/2 before sending traffic to the endpoint. 
This PR also enables Thin client mode by default (opt-out) however proxy routing additionally requires opting in to HTTP/2 via the new preview `CosmosClientOptions.EnableHttp2` option."

Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
```diff
@@ -155,13 +163,16 @@ internal static bool IsThinClientRoutable(IGlobalEndpointManager endpointManager
...
+        /// Read-direction variant of <see cref="IsThinClientRoutable"/> for failover walks (PPCB / PPAF) that
+        /// traverse thin-client READ endpoints regardless of the original request direction. Because the walk
+        /// selects the whole read-endpoint list rather than a single endpoint, it requires every read region to
+        /// be probe-healthy (<see cref="IGlobalEndpointManager.AreAllThinClientReadEndpointsHealthy"/>);
+        /// otherwise it routes through the gateway read endpoints.
         /// </summary>
         internal static bool IsThinClientReadRoutable(IGlobalEndpointManager endpointManager, DocumentServiceRequest request)
         {
             return IsOperationSupportedByThinClient(request)
+                && endpointManager.AreAllThinClientReadEndpointsHealthy
```
kushagraThapar, line 175: If some region has a transient issue for few seconds, this will cause a fallback to the gateway v1 even if that region is not going to be used. Is this intentional? Why can't we just check the endpoint health on the basis of next region to be used? and if all of them fail, we eventually fallback the request to Gateway V1?

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
```diff
@@ -123,7 +126,61 @@ public GlobalEndpointManager(
...
+        /// <summary>
+        /// Returns true when has been confirmed healthy by the
+        /// connectivity probe, or when no probe client is wired. An endpoint reports
+        /// unhealthy until its first successful probe, so the routing site uses the proxy only for healthy
+        /// regions and Gateway V1 otherwise.
+        /// </summary>
+        public bool IsProxyEndpointHealthy(Uri thinClientEndpoint)
+        {
+            EndpointProbeClient? probeClient = this.thinClientProbeClient;
+            return probeClient == null || probeClient.IsEndpointHealthy(thinClientEndpoint);
```
kushagraThapar: I am curious in which cases probeClient can be null? If the probeClient itself is null, and probing is required for thin client to work successfully (we are calling that out in the PR description that only regions which will pass the check will be used for thin client otherwise we fallback to gateway v1), why are we marking the proxy endpoint healthy in this case?

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
```diff
@@ -970,14 +970,29 @@ internal Uri ResolveThinClientEndpoint(DocumentServiceRequest request, bool isRe
...
+            if (request.RequestContext != null && request.RequestContext.LocationEndpointToRoute != null)
+            {
+                return request.RequestContext.LocationEndpointToRoute;
+            }
+
             DatabaseAccountLocationsInfo snapshot = this.locationInfo;
             ReadOnlyCollection<Uri> endpoints = this.GetApplicableThinClientEndpoints(request, isReadRequest, snapshot);
 
             int locationIndex = request.RequestContext.LocationIndexToRoute.GetValueOrDefault(0);
-            Uri chosenEndpoint = endpoints[locationIndex % endpoints.Count];
```
kushagraThapar, line 977: why are we removing this logic?

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,252 @@
...
+        {
+            return regionalEndpoint != null && this.healthyEndpoints.ContainsKey(regionalEndpoint);
+        }
+
+        /// <summary>
+        /// Runs one probe cycle against the supplied endpoints. Only endpoints not already cached healthy are
+        /// probed; any returning HTTP 200 are added to the permanent success cache. Never throws so probe
+        /// failures do not fail topology refresh.
+        /// </summary>
+        public async Task RunProbeCycleAsync(IReadOnlyCollection<Uri> regionalEndpoints, CancellationToken cancellationToken)
```
kushagraThapar, line 68: what happens to the endpoints which were marked healthy in the first attempt and then they become unhealthy eventually? Does this mean every new request (after the endpoint is unhealthy) uses the unhealthy endpoint and then eventually gets routed to different region after retries? This will degrade the client's performance for every request. 
If we are marking the thin client endpoint unavailable in client retry policy, then shouldn't we also remove it from the cache until the next cache re-validates the health of that endpoint if its still returned in the next topology update?
Otherwise, we will always go to that bad endpoint at least once every 5 mins or every topology update.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -315,7 +315,23 @@ public int GatewayModeMaxConnectionLimit
...
+        /// through the proxy only when the probe succeeds; otherwise it falls back to the regular gateway
+        /// (Gateway V1) path.
+        /// <value>Default value is false.</value>
+        /// </summary>
+#if PREVIEW
+        public
+#else
+        internal
+#endif
+        bool? EnableHttp2 { get; set; }
```
kushagraThapar: I am not sure if we want to gate .NET customers on explicitly enabling HTTP2. If .NET already supports HTTP2 by default, then we don't need to expose this API.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
jeet1995: I wonder if we can take a null-safe approach where an HTTP/1.1 client this is a no-op EndprobeClient (reference can exist).

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
aavasthy: fail-open was not the right default for a safety gate. Fixed: both methods now fail closed. IsProxyEndpointHealthy returns probeClient != null && probeClient.IsEndpointHealthy(...), and AreAllThinClientReadEndpointsHealthy returns false when no probe client is wired or no read endpoints are advertised.

changelog.md
aavasthy: updated changelog

Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs
aavasthy: the logic isn't removed — it's extracted, not deleted. ResolveThinClientEndpoint still pins via RouteToLocation exactly as before; it now delegates the computation to a new GetThinClientEndpointCandidate helper that returns the same endpoint without pinning.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
aavasthy: Discussed offline and removed the EnableHttp2 option entirely

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: This is an intentional design choice: the probe is the single, decoupled source of "this region's proxy is reachable over HTTP/2," and the cache is add-only. On the green→red transition we deliberately rely on the existing failover layer rather than cache eviction.

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: Not a concern for thin client in practice. Thin client only runs on modern TFMs (net6.0+) where VersionPolicy exists, so RequestVersionExact is always applied and a probe can only return 200 over HTTP/2 — no HTTP/1.1 downgrade.

Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs
aavasthy: I changed the refresh-path trigger from await this.RunThinClientProbeCycleAsync(); to _ = this.RunThinClientProbeCycleAsync();, so both the init trigger (L699) and the refresh trigger are now fire-and-forget. This path is shared with the forced-refresh that ClientRetryPolicy invokes inline on a request's failover retry, so awaiting it could stall live requests. Added a one-line comment explaining why.

Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
aavasthy: if probe endpoint fails then it moves to gateway v1 mode. But if for some reason we see transient issues with thinclient then it just fails with that error code. There is no fallback on gateway mode for such scenario.
The only time we use gateway mode: 
- if get account call does not have thinclient endpoints
- if probe endpoint request fails

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -315,8 +315,8 @@ public int GatewayModeMaxConnectionLimit
         /// </summary>
         /// <value>Default value is 6 seconds.</value>
         /// <seealso cref="CosmosClientBuilder.WithRequestTimeout(TimeSpan)"/>
-        public TimeSpan RequestTimeout { get; set; }
-
+        public TimeSpan RequestTimeout { get; set; }
```
FabianMeiswinkel: NIT: revert - just whitespace changes?

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,252 @@
...
+            {
+                probeUri = EndpointProbeClient.BuildProbeUri(regionalEndpoint);
+            }
+            catch (Exception ex)
+            {
+                DefaultTrace.TraceWarning("Failed to build probe URI for {0}: {1}", regionalEndpoint, ex.Message);
+                return (regionalEndpoint, false);
+            }
+
+            // Up to 1 + maxRetries attempts, stopping at the first HTTP 200, so a transient blip does not flap
```
FabianMeiswinkel: It won't correct? I don't think probe retries are needed anymore? What is maxRetries?

Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
FabianMeiswinkel: What do you mean with "falls back to Gateway V1" - only if service does not offer Gateway V2 endpoint for a region, correct? Which won't happen automagically when a region is unavailable - but service team could apply config to achieve it.

FabianMeiswinkel (changes requested):
LGTM overall - PR description needs to be updated to reflect the current logic (no fail-back to V1 based on probe failing) and I think the angle to update UserAgent flags to reflect the outcome of probing is missing. Otherwise just NITs.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -326,9 +326,16 @@ public void OnBeforeSendRequest(DocumentServiceRequest request)
...
-            this.locationEndpoint = this.isThinClientEnabled
+            // Resolve and pin the endpoint for the request. Per-region thin client gate: route to the proxy only
+            // when the regional endpoint the request would use is probe-healthy; otherwise pin the gateway
+            // endpoint. 
+            Uri thinClientCandidate = this.isThinClientEnabled
                 && ThinClientStoreModel.IsThinClientRoutable(this.globalEndpointManager, request)
+                ? this.globalEndpointManager.GetThinClientEndpointCandidate(request)
+                : null;
+
+            this.locationEndpoint = thinClientCandidate != null
```
jeet1995, line 341: It looks like we have two sources which give you Uri - GetThinClientEndpointCandidate and ResolveThinClientEndpoint - ideally should be the latter and then what is resolved - you verify proxy health on it.

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,252 @@
...
+    /// endpoint is cached for the lifetime of the client and never re-probed. Routing consults
+    /// <see cref="IsEndpointHealthy"/> to use the proxy for healthy regions and Gateway V1 otherwise.
+    /// </summary>
+    internal sealed class EndpointProbeClient : IDisposable
+    {
+        private const string ConnectivityProbeOperationType = "ConnectivityProbe";
+        private const string ProbePath = "/connectivity-probe";
+
+        // Default in-cycle probe retries per endpoint.
+        private const int DefaultMaxRetries = 3;
```
jeet1995: How did we select DefaultMaxRetries = 3?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/EndpointProbeClientTests.cs
```diff
@@ -0,0 +1,287 @@
+//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos.Tests
```
jeet1995, line 4: Can we do long running tests which cover at least:
1. Add a region which doesn't have endpoint probe.
2. What happens when we now remove the region without endpoint probe support - ideally we should go back to using GatewayV2.
3. What happens when we are on Direct Mode + HTTP/2 and we require QueryPlan - should we go to ThinProxy?

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,252 @@
...
+                    return (regionalEndpoint, true);
+                }
+            }
+
+            return (regionalEndpoint, false);
+        }
+
+        private async Task<bool> ProbeEndpointOnceAsync(
+            Uri probeUri,
+            Uri regionalEndpoint,
```
jeet1995: Do we require regionalEndpoint here?

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,252 @@
...
+                DefaultTrace.TraceWarning("Failed to build probe URI for {0}: {1}", regionalEndpoint, ex.Message);
+                return (regionalEndpoint, false);
+            }
+
+            // Up to 1 + maxRetries attempts, stopping at the first HTTP 200, so a transient blip does not flap
+            // the region to Gateway V1.
+            int totalAttempts = this.maxRetries + 1;
+            for (int attempt = 1; attempt <= totalAttempts; attempt++)
+            {
+                if (cancellationToken.IsCancellationRequested || Volatile.Read(ref this.closed) == 1)
```
jeet1995: Can we double check that the cancellationToken budget is not being shared across retries to the same probeEndpoint? I believe your intention is to keep the budget separate.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: The two methods compute the same URI: ResolveThinClientEndpoint is just GetThinClientEndpointCandidate + pinning the result onto the request context. We can't health-check on the resolved value because ResolveThinClientEndpoint pins :10250 before the check, and the gateway fallback (ResolveServiceEndpoint) early-returns whatever is already pinned — so a RED probe would still route to :10250 instead of Gateway V1. 

So the candidate is computed without pinning → probe health is verified on it → only then do we commit (pin) the proxy if healthy otherwise use gateway mode.

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: Removed the retry logic

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: this method has been removed

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/EndpointProbeClientTests.cs
aavasthy: Added tests covering 1 and 2 scenarios in thinclient integration tests

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: removed the concept of retrying probe endpoint as suggested by Fabian in above comment

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: Removed probe retry logic

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
aavasthy: updated

Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
aavasthy: yes there is no fallback on gateway v1 for a transient issue, fallback only happen if probe fails or thinclient does not return endpoints.

xinlian12:
@sdkReviewAgent

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        // Probes must fail fast; a short fixed budget avoids tying up a cycle longer than a refresh interval.
+        private static readonly TimeSpan PerProbeTimeout = TimeSpan.FromSeconds(5);
+
+        private static readonly PropertyInfo VersionPolicyProperty = typeof(HttpRequestMessage).GetProperty("VersionPolicy");
+
+        private readonly CosmosHttpClient httpClient;
+
+        // Endpoints that have returned HTTP 200 at least once. Membership is permanent for the client lifetime.
+        private readonly ConcurrentDictionary<Uri, byte> healthyEndpoints = new ConcurrentDictionary<Uri, byte>();
```
FabianMeiswinkel, line 42: This would mean re-probing every 5 minutes - I guess that is intentional - you could argue that we can even cache the negative-probe - but I guess with the current alogrithm we can avoid negative side effects due to transeint probe failures. Fine with me - just wanted explicit ACK that negative probes are not chached.

FabianMeiswinkel (dismissed):
LGTM

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: yes only successful probe are cached and cache is permanent. We do not have an eviction logic or ttl set for cache.

kushagraThapar (dismissed):
LGTM, thanks @aavasthy

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -135,7 +135,7 @@ internal partial class DocumentClient : IDisposable, IAuthorizationTokenProvider
         //Auth
         internal readonly AuthorizationTokenProvider cosmosAuthorization;
 
-        private readonly bool isThinClientFeatureFlagEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: false);
+        private readonly bool isThinClientFeatureFlagEnabled = ConfigurationManager.IsThinClientEnabled(defaultValue: true);
```
xinlian12, line 138: 🟡 **Recommendation — PR title `[Internal]` is inconsistent with the change's customer-facing scope**

This PR flips `IsThinClientEnabled(defaultValue: true)`. Per the team `.pragentrules`:

> `[Internal]` should ONLY be used for changes with NO customer-facing impact. … feature additions … that affect customers should NOT use `[Internal]`.

The default flip is customer-observable even though the probe gate fails closed:
1. **New outbound traffic shape**: Gateway-mode clients against thin-client-enabled accounts will now see periodic `POST /connectivity-probe` over HTTP/2, plus data-plane HTTP/2 traffic to a different host:port (proxy `:10250`). Corporate egress policies that allow only `:443` will see probe failures and silent fallback.
2. **The opt-out path** (`AZURE_COSMOS_THIN_CLIENT_ENABLED=false`) is the customer's remediation knob and needs explicit visibility.
3. **`SendingRequestEventArgs` interceptors observe the probe traffic** (see separate finding) — a behavioral contract change.

**Suggested action:** Retitle to `[ThinClient]` consistent with the rest of the rollout, and make the changelog mention both `AZURE_COSMOS_THIN_CLIENT_ENABLED=false` (opt-out) and the new outbound host:port (`:10250` over HTTP/2) so customers diagnosing unexpected requests or egress rejections can correlate back to this release.

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,225 @@
...
+            {
+                using HttpResponseMessage responseMessage = await this.httpClient.SendHttpAsync(
+                    createRequestMessageAsync: () => new ValueTask<HttpRequestMessage>(
+                        EndpointProbeClient.CreateProbeRequest(probeUri)),
+                    resourceType: ResourceType.Document,
+                    timeoutPolicy: HttpTimeoutPolicyNoRetry.Instance,
+                    clientSideRequestStatistics: null,
+                    cancellationToken: perProbeCts.Token);
+
+                bool ok = responseMessage.StatusCode == HttpStatusCode.OK;
```
xinlian12, line 180: 🟡 **Recommendation — Verify the *negotiated* HTTP/2 version on the response, not just the requested version**

The probe's stated invariant ("an HTTP/2 mismatch falls back to Gateway V1") is currently enforced two ways:
1. `requestMessage.Version = new Version(2, 0)` — a hint
2. `VersionPolicy = RequestVersionExact` via reflection — only applied when `HttpRequestMessage.VersionPolicy` exists at runtime

The second one is the load-bearing piece, and there's an open thread above debating whether that holds on `netstandard2.0` consumers. Regardless of how that thread resolves, there's a strictly safer, TFM-independent gate available right here:

```csharp
bool ok = responseMessage.StatusCode == HttpStatusCode.OK
       && responseMessage.Version.Major >= 2;

```

This directly checks what we actually care about — that HTTP/2 was negotiated end-to-end — rather than inferring it from the request-side policy. It (a) catches any silent HTTP/1.1 downgrade on any framework regardless of `VersionPolicy` support; (b) survives any future runtime change to `VersionPolicy`; (c) protects against custom `HttpClientFactory` handlers (e.g., `WinHttpHandler` on .NET Framework consumers of this netstandard2.0 assembly) that might negotiate HTTP/1.1 even with the request hint set. The reflection path can stay as a fast-fail optimization but this is the real correctness check.

Worth pairing with a unit test that asserts a mock handler returning `Version = 1.1, Status = 200` is treated **red**.

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
```diff
@@ -0,0 +1,225 @@
...
+            }
+
+            // A single probe per cycle; no in-cycle retries. A red endpoint stays on Gateway V1 until a later
+            // cycle (after the next topology refresh) re-probes and succeeds.
+            using CancellationTokenSource perProbeCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            perProbeCts.CancelAfter(EndpointProbeClient.PerProbeTimeout);
+
+            try
+            {
+                using HttpResponseMessage responseMessage = await this.httpClient.SendHttpAsync(
```
xinlian12, line 172: 🟡 **Recommendation — Probe traffic leaks through the public `SendingRequestEventArgs` interceptor**

`EndpointProbeClient` sends `POST /connectivity-probe` through the **shared** `CosmosHttpClient`, which means every probe flows through `CosmosHttpClientCore.HttpRequestMessageHandler` and raises the customer-registered `SendingRequestEventArgs` / `ReceivedResponseEventArgs` events.

The PR diff itself is the direct evidence — three test sites had to add a filter so their own assertions don't trip on probe traffic:

```csharp
&& !string.Equals(e.HttpRequest.RequestUri?.AbsolutePath, "/connectivity-probe", StringComparison.OrdinalIgnoreCase)

```
(`CosmosItemThinClientTests.cs:475, :539`, and the new `RegionFailoverTests.cs` mocks)

With thin client now default-on, customer handlers used for logging, metrics, OpenTelemetry, header-injection, or audit will all silently start observing extra `POST /connectivity-probe` calls they did not issue. Some will misclassify them as user traffic; header-injection handlers can mutate the probe request; counters will drift.

**Suggested action:** Either
- (a) flag probe requests via `HttpRequestMessage.Options`/`Properties` and skip the public event invocation for tagged requests inside `CosmosHttpClientCore`, **or**
- (b) explicitly document this as observable behavior in the changelog and an `@remarks` block on `CosmosClientOptions.SendingRequestEventArgs`.

The fact that the SDK's own tests are the first consumer to trip over this is a strong signal that customer handlers will too. Option (a) is preferable since the probe is infrastructure, not data-plane.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -326,9 +326,16 @@ public void OnBeforeSendRequest(DocumentServiceRequest request)
...
+                && this.globalEndpointManager.IsProxyEndpointHealthy(thinClientCandidate)
                 ? this.globalEndpointManager.ResolveThinClientEndpoint(request)
```
xinlian12: 🟡 **Recommendation — Eliminate the redundant resolve; pin the URI you already health-checked**

Currently `OnBeforeSendRequest` computes the thin-client URI twice on the happy path:

1. `GetThinClientEndpointCandidate(request)` reads `locationCache.locationInfo` (snapshot A), returns URI `X`, does NOT pin.
2. `IsProxyEndpointHealthy(X)` checks the probe cache for `X`.
3. **Healthy branch**: `ResolveThinClientEndpoint(request)` — which internally re-calls `GetThinClientEndpointCandidate` and reads `locationInfo` again (snapshot B), then pins via `RouteToLocation`. (See `LocationCache.cs:514-520`.)

Two issues with the second computation:

1. **Wasted work on every retry-driven dispatch** — the dictionary lookup + endpoint-list traversal happen twice on the hot path.
2. **Latent snapshot race**: `locationInfo` is a reference reassignment in `LocationCache.Update`, and topology refresh runs on a Timer thread that can interleave with `OnBeforeSendRequest`. If a refresh fires between steps 1 and 3 (e.g., a new region inserted, reordered preferred regions), step 3 can pin URI `Y ≠ X`, meaning the URI that was just probe-checked is **not** the URI we end up routing to. Today it's safe-ish because the probe cache is permanent-green (any URI that was healthy stays healthy), but if the green cache ever gains eviction (a real possibility — see the `green→red` discussion above), this becomes a live correctness bug.

**Suggested simplification** — both faster and race-free:

```csharp
if (thinClientCandidate != null
    && this.globalEndpointManager.IsProxyEndpointHealthy(thinClientCandidate))
{
    request.RequestContext.RouteToLocation(thinClientCandidate);
    this.locationEndpoint = thinClientCandidate;
}
else
{
    this.locationEndpoint = this.globalEndpointManager.ResolveServiceEndpoint(request);
}

```

The pinned URI is the URI that was health-checked, by construction.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemIntegrationTests.cs
```diff
@@ -44,6 +44,7 @@ public class CosmosItemIntegrationTests
         public async Task TestInitAsync()
         {
             this.connectionString = ConfigurationManager.GetEnvironmentVariable<string>("COSMOSDB_MULTI_REGION", null);
+            Environment.SetEnvironmentVariable(ConfigurationManager.ThinClientModeEnabled, "False");
```
xinlian12, line 47: 🟡 **Recommendation — Env var is process-global and never reset; pollutes downstream tests**

`Environment.SetEnvironmentVariable(ConfigurationManager.ThinClientModeEnabled, "False")` is set in `TestInitAsync` but `TestCleanup` (lines 76–93) only resets `StalePartitionUnavailabilityRefreshIntervalInSeconds`. The variable persists for the lifetime of the test process. Any test that runs after this one in the same xUnit/MSTest worker — including new ThinClient tests added later — will silently see `ThinClientModeEnabled=False` and therefore won't actually exercise the new default-on path that this PR introduces. Future regressions can ship green.

The codebase already has the right `try/finally` pattern for env-var-driven feature flags — e.g. `CosmosClientOptionsUnitTests.cs:668-738` and `CosmosScopeProviderTests.cs:21-31`.

**Suggested action:** Add a `[TestCleanup]` step (or wrap the `TestInitAsync` body) that captures the previous value and restores it (`Environment.SetEnvironmentVariable(name, previous)`), matching the existing pattern.

Same issue applies at `CosmosAvailabilityStrategyTests.cs:1613` (the single-test variant inline-sets it with no cleanup either).

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
```diff
@@ -469,6 +502,289 @@ public async Task HttpRequestVersionIsTwoPointZeroWhenUsingThinClientMode()
...
+
+                bool healthyRegionUsedProxy = ProxyWasUsedForRegion(probeHandler, healthyRegionFragment);
+                bool failedRegionUsedProxy = ProxyWasUsedForRegion(probeHandler, failedRegionFragment);
+
+                if (!healthyRegionUsedProxy)
+                {
+                    Assert.Inconclusive(
+                        "The thin-client proxy was never used for the healthy region (Central US). This happens when the " +
+                        "client and Gateway V2 could not negotiate HTTP/2 in this environment, so the per-region probe gate " +
+                        "cannot be observed.");
```
xinlian12, line 671: 🟡 **Recommendation — `Assert.Inconclusive` on the feature this test is exercising lets regressions ship silently**

The condition gating this inconclusive is *"the client and Gateway V2 could not negotiate HTTP/2 in this environment"* — but HTTP/2 negotiation is the very mechanism the probe gate validates and the entire reason this PR exists. If a future regression breaks HTTP/2 negotiation in CI (e.g., a handler config change, a TLS/ALPN issue, a misconfigured emulator), this test prints `Inconclusive` and the run goes **green**.

Existing inconclusive uses in this codebase are appropriate when the precondition is genuinely outside the SDK's control (e.g., `BarrierRequestReplicaRetryTests.cs:170-177` checks for an unrelated runtime API). Here the precondition is the feature.

The same pattern is in `EndToEnd_RemovingRegionWithoutConnectivityProbe_RestoresThinClientRouting` (`:759-761`).

**Suggested action:** Promote the precondition into a `[ClassInitialize]` skip-once check that visibly *deselects* these tests when HTTP/2 can't be negotiated (or add a `[TestCategory("ThinClient-RequiresHttp2")]` and ensure the CI lane that runs them has Gateway V2 as a hard precondition). Either way, make the absence of these tests observable in the test report rather than have them inconclusive-pass.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
```diff
@@ -530,9 +854,13 @@ public async Task CreateItemsTestWithThinClientFlagEnabledAndAccountDisabled()
...
                     Assert.AreEqual(HttpStatusCode.Created, response.StatusCode);
-                    string diagnostics = response.Diagnostics.ToString();
-                    Assert.IsFalse(diagnostics.Contains("|F4"), "Diagnostics User Agent should NOT contain '|F4' for Gateway");
                 }
+
+                // The account does not advertise thin-client (proxy) endpoints, so no request - data-plane or
+                // connectivity-probe - must ever target the proxy port (10250). All traffic stays on Gateway V1.
+                Assert.IsFalse(
+                    requestPorts.Contains(ThinClientProxyPort),
+                    $"No request should target the ThinClient proxy port {ThinClientProxyPort} when the account does not advertise thin-client endpoints. Observed ports: {string.Join(", ", requestPorts.Distinct())}");
```
xinlian12, line 866: 🟢 **Suggestion — Replacing the `|F4` assertion with a port-only check weakens negative coverage**

The pre-PR version of this test asserted `Assert.IsFalse(diagnostics.Contains("|F4"))`, which verified the SDK's user-agent did **not** advertise the ThinClient capability when the account doesn't advertise proxy endpoints. The new test only verifies no request hit port `10250`. The two assertions cover slightly different bugs:

- **Old check (`|F4`)** catches a regression where `isThinClientEnabled` (or the user-agent flag) is incorrectly set on a non-thin-client account.
- **New check (port)** catches a regression where routing accidentally targets the proxy port.

A bug that sets `|F4` while *also* failing the probe and routing to gateway (e.g., misclassifying the account during init but probe gate saves us) would now pass — even though the SDK is wrongly advertising a capability it isn't using.

**Suggested action:** Keep both assertions. The negative assertion on `|F4` is additive and cheap:

```csharp
Assert.IsFalse(requestPorts.Contains(ThinClientProxyPort), …);
// ALSO keep:
foreach (var diag in diagnosticsList) Assert.IsFalse(diag.Contains("|F4"), …);

```

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
xinlian12: 🟡 **Recommendation — The existing thin-client `ClientRetryPolicy` test accidentally passes; the new probe-gate branching is effectively untested**

The new branching in this method is the per-request thin-client routing decision:

```csharp
this.locationEndpoint = thinClientCandidate != null
    && this.globalEndpointManager.IsProxyEndpointHealthy(thinClientCandidate)
    ? this.globalEndpointManager.ResolveThinClientEndpoint(request)
    : this.globalEndpointManager.ResolveServiceEndpoint(request);

```

The only existing test that exercises this path — `ClientRetryPolicyTests::ThinClient_OnBeforeSendRequest_AfterServiceWithdrawsThinLocations_FallsBackToGatewayEndpoint` — calls `TestUtils.EnableThinClientLocationsForTest` but **does not** call the new `MarkThinClientEndpointsHealthyForTest`. So `IsProxyEndpointHealthy` returns `false` and the resolve goes down the gateway-fallback branch. The test passes only because `EnableThinClientLocationsForTest` uses **the same URI** (`https://mock.thinclient.com/`) for `ReadLocationsInternal` *and* `ThinClientReadableLocationsInternal` (verified in `TestUtils.cs`), so the gateway-resolved endpoint happens to also appear in `ThinClientReadEndpoints`. A regression that always returned the gateway endpoint, or that inverted the probe gate, would not be caught.

**Suggested action:** Either (a) extend `EnableThinClientLocationsForTest` to take distinct gateway vs thin-client URIs, then add two ClientRetryPolicy unit tests asserting the pinned URI is the thin-client URI when probe-healthy and the gateway URI when not; **or** (b) add a focused test that uses both `EnableThinClientLocationsForTest` and `MarkThinClientEndpointsHealthyForTest` to exercise the healthy branch. The dispatch-time gate in `ThinClientStoreModel.DispatchAsync` (the `if (!IsProxyEndpointHealthy) base.DispatchAsync(...)` fallback) is similarly only validated end-to-end in emulator tests that can become `Inconclusive`.

Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
xinlian12, line 175: 🟢 **Suggestion — `AreAllThinClientReadEndpointsHealthy` real-implementation semantics aren't directly tested**

The new `ThinClientStoreModelTests::IsThinClientReadRoutable_GatedByAllReadEndpointsHealthy` data-row test stubs the property on `Mock<IGlobalEndpointManager>`, which validates only that *callers* honor the bool — not that the implementation correctly returns it.

The real getter has three branches that determine routing for every PPCB/PPAF failover walk:

1. Loop returns `false` on the first unhealthy endpoint (all-of semantic).
2. Returns `false` when `probeClient == null` (fail-closed when not wired).
3. Returns `false` when `ThinClientReadEndpoints.Count == 0` (fail-closed when no read endpoints).

A bug like off-by-one in the loop, or someone "fixing" branch (3) to return `true` when the read list is empty (which is what the early-return at the bottom of the method would naturally produce without the explicit guard), would let the failover walk pin thin-client URIs that the probe hasn't validated — and no test would catch it.

**Suggested action:** Add a `GlobalEndpointManagerTests` (or extend `ThinClientStoreModelTests`) case that wires two real thin-client read endpoints, marks one healthy via `MarkThinClientEndpointsHealthyForTest`, and asserts the property is `false` until both are healthy. Add separate cases for the null-probe-client and empty-read-list branches.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
```diff
@@ -1610,6 +1610,7 @@ private static async Task HandleChangesStepAsync(
         public async Task AvailabilityStrategy_HedgePicksUpHubHeaderAfter_404_1002_And_403_3()
         {
             // ── Tracking counters (thread-safe for concurrent primary + hedge) ──
+            Environment.SetEnvironmentVariable(ConfigurationManager.ThinClientModeEnabled, "False");
```
xinlian12, line 1614: 🟡 **Recommendation — Reset the env var in `[TestCleanup]` (matches `CosmosItemIntegrationTests` finding)**

This inline `SetEnvironmentVariable("False")` mid-test-body has no companion reset. The test's existing `TestCleanup` (lines 97-113) doesn't restore the previous value, so `ThinClientModeEnabled=False` persists for **every** subsequent test in the same MSTest worker process. Combined with default-on (`DocumentClient.cs:138`), this means later tests that *should* exercise the new default silently won't.

Two test files in this PR have the same pattern (`CosmosItemIntegrationTests.cs:47` and this site). The codebase already has the right `try/finally` pattern — see `CosmosClientOptionsUnitTests.cs:668-738` for the established idiom.

```csharp
string previous = Environment.GetEnvironmentVariable(ConfigurationManager.ThinClientModeEnabled);
try
{
    Environment.SetEnvironmentVariable(ConfigurationManager.ThinClientModeEnabled, "False");
    // ... test body ...
}
finally
{
    Environment.SetEnvironmentVariable(ConfigurationManager.ThinClientModeEnabled, previous);
}

```

xinlian12:
✅ **Review complete** (11:18)

Posted **10** inline comment(s).

Microsoft.Azure.Cosmos/src/DocumentClient.cs
aavasthy: Updated PR title

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: Both CosmosClientOptions.SendingRequestEventArgs and CosmosClientBuilder.WithSendingRequestEventArgs are internal, so there's no public API for a customer to register a SendingRequestEventArgs/ReceivedResponseEventArgs handler.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: Done. The healthy branch now pins thinClientCandidate directly — the same URI IsProxyEndpointHealthy just validated — instead of calling ResolveThinClientEndpoint, which re-read locationInfo and could pin a different region on a concurrent topology refresh. RouteToLocation pins both branches, so we route to exactly the URI we probed, with no duplicate lookup.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemIntegrationTests.cs
aavasthy: TestCleanup now clears ThinClientModeEnabled (resets to the default-on baseline) in CosmosAvailabilityStrategyTests

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
aavasthy: The inconclusive precondition here isn't the SDK feature — it's an environment capability: whether the CI lane has a Gateway V2 that can negotiate HTTP/2 at all. When it can't, there's simply no proxy traffic to observe, so neither the positive nor the negative proxy-gate assertion is meaningful. The invariant that doesn't depend on HTTP/2 — that a read still succeeds via the Gateway V1 fallback is probe request fails.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemThinClientTests.cs
aavasthy: Added the Assert.IsFalse(diagnostics.Contains("|F4")) check back inside the create loop alongside the port assertion, so we still catch a regression that advertises the ThinClient capability on an account that doesn't expose proxy endpoints.

Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
aavasthy: The all-of semantic is exercised end-to-end in EndToEnd_RegionWithoutConnectivityProbe_FallsBackToGatewayForThatRegion: with one region's probe red, the read isn't routed to the proxy and falls back to Gateway V1 — which only happens if the real getter returns false. The other two branches are fail-closed guards that can only make routing more conservative, so a regression there degrades to gateway, never to an unvalidated proxy URI

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAvailabilityStrategyTests.cs
aavasthy: Fixed — TestCleanup now clears ThinClientModeEnabled (resets to the default-on baseline), so the flag no longer leaks to later tests in the worker. Same fix applied to CosmosItemIntegrationTests.

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
xinlian12, line 42: QQ: 
   could there be a scenarios:
 initial thinClientEndpoints -> A, B, C
 next refresh -> A, B
next refresh -> A, B, C -> do we need to revalidate C?

xinlian12: also another QQ: 
  let us for the account, it configured with read regions A, B, C. Does the thinClientReadEndpoint will contain all or nothing? like it will never only return A or A, B? (it will either be A, B, C or nothing?)

xinlian12 (approved):
LGTM, thanks

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
xinlian12: I do not think we need this change, since thin client proxy service only support http2?

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: _could there be a scenarios:
_initial thinClientEndpoints -> A, B, C
next refresh -> A, B
next refresh -> A, B, C -> do we need to revalidate C?__

**Probe endpoint cache is permanent, it will always keep C region added. So this way if C region in scenario goes away and then comes back in account call then we dont need to revalidate it as it is still in cache.**


_"let us for the account, it configured with read regions A, B, C. Does the thinClientReadEndpoint will contain all or nothing? like it will never only return A or A, B? (it will either be A, B, C or nothing?)"_
**It will return all three read regions A,B,C**

kushagraThapar (approved):
LGTM!

Microsoft.Azure.Cosmos/src/EndpointProbeClient.cs
aavasthy: yes probe endpoint request will only trigger for http2 request and thinproxy will send the repsonse

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
aavasthy: EnableThinClientLocationsForTest uses the same URI for read and thin-client locations, so no assertion can tell "routed to proxy because healthy" from "routed to gateway." The probe gate is covered end-to-end against distinct real regions in EndToEnd_RegionWithoutConnectivityProbe_FallsBackToGatewayForThatRegion and EndToEnd_ProbeEnabledAccountWithHttp2_RoutesDataPlaneThroughThinClientProxy (healthy → proxy :10250, probe-red → Gateway V1, asserted on the observed port).

PR #5974, opened 2026-06-23
[Internal] DTS: Adds response reordering and fail-closed result handling

Meghana-Palaparthi:
## Description

Distributed transaction (DTX) responses do not guarantee that per-operation results come back in the same order the operations were sent, and a malformed or short payload could otherwise surface misaligned results to the caller. This change makes DTX response parsing order-safe and fail-closed.

**What it does:**

- **Reorders by server `index`.** Each per-operation result carries the original request `index`; results are placed back into request order rather than trusting wire order. The indices must form a complete permutation of `0..n-1`. A new `HasIndex` flag distinguishes a real server `index:0` from a defaulted/omitted one so an ambiguous payload is never mis-mapped.
- **Fails closed on unmappable payloads.** Bad/duplicate/missing indices, per-operation parse failures, and result-count mismatches are treated as uninterpretable: on a success status the response is downgraded to `500 InternalServerError`; on an error status the SDK synthesizes uniform error placeholders so the caller always gets exactly `Operations.Count` results.
- **Fixes a resource-stream leak.** Discarded parsed results' `ResourceStream`s are now disposed before the result list is replaced. The four disposal sites share a single null-safe `DisposeResultStreams` helper (also used by `Dispose()`).
- **Preserves envelope-level signals across reconstruction.** The coordinator's root-level `isRetriable` verdict and `diagnosticString` are independent of the per-operation payload, so they are now carried onto every synthesized fail-closed response. Previously the count-mismatch path dropped them, which could turn a coordinator-signalled "safe to retry" into a terminal failure; all three fail-closed paths are now consistent.

**Worth a careful look:** the `isRetriable` preservation is the most behaviorally significant piece. The retry it enables is safe because it is guarded by the idempotency token; suppressing the signal was strictly worse than keeping it.

## Type of change

## Closing issues

N/A

## Changelog

> All changes are confined to the `PREVIEW`-gated DistributedTransaction feature, which is internal/preview and not reachable in a customer's default configuration. No customer-observable impact in any shipped default-config path.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
@@ -30,6 +30,7 @@ namespace Microsoft.Azure.Cosmos
         internal DistributedTransactionOperationResult(DistributedTransactionOperationResult other)
         {
             this.Index = other.Index;
+            this.HasIndex = other.HasIndex;
             this.StatusCode = other.StatusCode;
             this.SubStatusCode = other.SubStatusCode;
             this.ETag = other.ETag;
@@ -88,6 +89,12 @@ namespace Microsoft.Azure.Cosmos
         public virtual double RequestCharge { get; internal set; }
 
         internal virtual SubStatusCodes SubStatusCode { get; set; }
+
+        /// <summary>
+        /// Whether the server explicitly provided an <c>index</c>. A missing <c>index</c> defaults to 0
+        /// (indistinguishable from a real 0), so the reorder logic uses this flag to reject ambiguous responses.
+        /// </summary>
+        internal bool HasIndex { get; set; }
 
         internal virtual string SessionToken { get; set; }
 
@@ -159,6 +166,7 @@ namespace Microsoft.Azure.Cosmos
             if (TryGetInt32Property(json, DistributedTransactionSerializer.Index, out int index))
             {
                 result.Index = index;
+                result.HasIndex = true;
             }
 
             if (TryGetInt32Property(json, DistributedTransactionSerializer.StatusCode, out int statusCode))
--- a/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
+++ b/Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
@@ -57,10 +57,15 @@ namespace Microsoft.Azure.Cosmos
         }
 
         /// <summary>
-        /// Gets the <see cref="DistributedTransactionOperationResult"/> at the specified index in the response.
+        /// Gets the <see cref="DistributedTransactionOperationResult"/> for the request operation at the specified index.
         /// </summary>
-        /// <param name="index">The zero-based index of the operation result to get.</param>
-        /// <returns>The <see cref="DistributedTransactionOperationResult"/> at the specified index.</returns>
+        /// <param name="index">The zero-based index of the request operation whose result to get.</param>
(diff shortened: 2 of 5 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/DistributedTransaction/DistributedWriteTransactionTests.cs
```diff
@@ -346,6 +346,69 @@ public async Task CommitAsync_OperationIndexIsZeroBasedAndOrdered()
             }
         }
 
+        [TestMethod]
+        [Description("When server operationResponses arrive out-of-order, the SDK reorders them by index so response[i].Index == i.")]
```
jeet1995, line 350: Can we add end to end tests with say 100+ operations in a commit? Read Tx and Write Tx? (Mostly a smoke test). Do we know what happens if duplicate OperationIndex is sent by client?

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionOperationResult.cs
```diff
@@ -89,6 +90,12 @@ internal DistributedTransactionOperationResult()
 
         internal virtual SubStatusCodes SubStatusCode { get; set; }
 
+        /// <summary>
+        /// Whether the server explicitly provided an <c>index</c>. A missing <c>index</c> defaults to 0
+        /// (indistinguishable from a real 0), so the reorder logic uses this flag to reject ambiguous responses.
+        /// </summary>
+        internal bool HasIndex { get; set; }
```
jeet1995, line 97: Do we require HasIndex?

Meghana-Palaparthi: Yes, HasIndex distinguishes a server-omitted index (which defaults to 0) from a legitimate index: 0. Without it, the reorder logic would silently map an unlabeled operation result to request slot 0 whenever that slot is free—surfacing misaligned data instead of failing closed.

jeet1995 (approved):

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
```diff
@@ -273,6 +280,12 @@ internal static async Task<DistributedTransactionResponse> FromResponseMessageAs
...
                             SubStatusCodes wireSubStatusCode = responseMessage.Headers.SubStatusCode;
+
+                            // Preserve the envelope isRetriable/DiagnosticString before disposing: a wrong
+                            // result count is no more trustworthy than bad indices, and the retry signal is
+                            // independent of the unusable payload. Matches the two unmappable fail-closed paths.
+                            bool wireIsRetriable = response.IsRetriable;
+                            string wireDiagnosticString = response.DiagnosticString;
                             response.Dispose();
 
                             return new DistributedTransactionResponse(
```
kushagraThapar, line 291: Though not related to this PR, but we are not calling `response.CreateAndPopulateResults(serverRequest.Operations);` before returning here, is that intentional?

kushagraThapar (approved):
One question, otherwise looks good to me, thanks!

Microsoft.Azure.Cosmos/src/DistributedTransaction/DistributedTransactionResponse.cs
Meghana-Palaparthi: Created a tracking issue https://github.com/Azure/azure-cosmos-dotnet-v3/issues/6012. Will fix this in a follow-up PR

PR #5964, opened 2026-06-16
Enable barrier early yield on 429 for SDK customers

Praveen-Msft:
## Summary

Adds `CosmosClientOptions.EnableBarrierEarlyYieldOn429` (default `true`) so SDK customers automatically benefit from barrier early yield on 429 throttling, with opt-out capability.

When all replicas return 429 (Too Many Requests) during a write or read barrier request, `ConsistencyWriter` and `QuorumReader` return early instead of spinning until timeout. This lets `ResourceThrottleRetryPolicy` handle the retry with proper backoff, avoiding unnecessary timeout escalation.

## Changes

| File | Change |
|------|--------|
| `CosmosClientOptions.cs` | Public `EnableBarrierEarlyYieldOn429` property (default `true`) |
| `ConnectionPolicy.cs` | Internal `EnableBarrierEarlyYieldOn429` property (default `false`) |
| `DocumentClient.cs` | Passes `ConnectionPolicy.EnableBarrierEarlyYieldOn429` to `StoreClientFactory` |
| `ClientRetryPolicy.cs` | Skip `TryMarkEndpointUnavailableForPkRange` for 408/21013 (write barrier throttle) |
| `BarrierEarlyYieldOn429Tests.cs` | 20 tests: defaults, mapping, full constructor chain E2E, write/read barrier behavior, ClientRetryPolicy safeguard |

## Flow

`CosmosClientOptions.EnableBarrierEarlyYieldOn429 = true` (default)
→ `ConnectionPolicy.EnableBarrierEarlyYieldOn429`
→ `DocumentClient` passes to `StoreClientFactory`
→ constructor chain → `ConsistencyWriter` / `QuorumReader`

## Dependencies

- **CosmosDB repo PR 2117699** (merged): Direct layer constructor chain for `enableBarrierEarlyYieldOn429`
- **CosmosDB repo** (pending): `ConnectionPolicy.EnableBarrierEarlyYieldOn429` in SharedFiles
- **This PR will not build** until the Direct NuGet with the above changes is published

## Substatus codes

- `21013` (`Server_WriteBarrierThrottled`): ConsistencyWriter synthesizes 408 with this substatus
- `21011` (`Server_BarrierThrottled`): QuorumReader read barrier (returns 429 directly)

PR #5976, opened 2026-06-23
[Internal] Retry: Adds barrier early yield on 429 optimization with Direct 3.44.0

### LINKED ISSUE #5035: Quroum code not yeilding

**Describe the bug**
When receiving 429s on a single region account with strong consistency, quorum reader code does not yield. This behavior was reproduced using the fault injection library and diagnostics of this scenario have been attached.

[429-Diagnostics.json](https://github.com/user-attachments/files/18974198/429-Diagnostics.json)
 

**To Reproduce**
Steps to reproduce the behavior. If you can include code snippets or links to repositories containing a repro of the issue that can helps us in detecting the scenario it would speed up the resolution.

**Expected behavior**

**Actual behavior**
Provide a description of the actual behavior observed. 

**Environment summary**
SDK Version:

**Additional context**

Praveen-Msft:
## Summary

Integrates the barrier early yield on 429 optimization into the V3 SDK, backed by Direct NuGet 3.44.0.

When all replicas return 429 (Too Many Requests) during write/read barrier requests, ConsistencyWriter and QuorumReader now yield early with substatus 21013 (Server_WriteBarrierThrottled) instead of spinning until timeout. This allows ResourceThrottleRetryPolicy to handle retries efficiently.

**Important:** This Direct version bump introduces 408 responses with barrier-specific substatus codes (e.g., 21013) for barrier failures that previously surfaced as plain timeouts.

## Changes

- **DirectVersion**: 3.42.4 → 3.44.0 (brings in 408 substatus codes for barrier failures)
- **CosmosClientOptions**: Internal \EnableBarrierEarlyYieldOn429\ property (default true) — not exposed on public SDK surface, accessible to internal consumers via InternalsVisibleTo
- **ConnectionPolicy -> DocumentClient -> StoreClientFactory**: Plumbs the flag to ConsistencyWriter and QuorumReader constructors
- **ClientRetryPolicy**: Skips marking endpoint unavailable on the synthetic 408 with substatus 21013 (ConsistencyWriter converts replica 429 → 408/21013 as an early-yield signal) to prevent unnecessary cross-region failover
- **Tests**: 20 unit/e2e tests covering write barrier, read barrier, constructor chain propagation, and ClientRetryPolicy guard

## Related

- CosmosDB PR [2117699](https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB/pullrequest/2117699) - Original barrier early yield feature in Direct
- CosmosDB PR [2161106](https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB/pullrequest/2161106) - Cherry-pick into Direct 3.44.0
- Closes #5035
- Supersedes #5964


Directory.Build.props
```diff
@@ -3,7 +3,7 @@
 		<ClientOfficialVersion>3.61.0</ClientOfficialVersion>
 		<ClientPreviewVersion>3.62.0</ClientPreviewVersion>
 		<ClientPreviewSuffixVersion>preview.0</ClientPreviewSuffixVersion>
-		<DirectVersion>3.43.2</DirectVersion>
+		<DirectVersion>3.44.0</DirectVersion>
```
ananth7592, line 6: We are now having a requirement of adding any changes including the direct package upgrades to changelog in the unreleased section

ananth7592:
The new public bool EnableBarrierEarlyYieldOn429 isn't reflected in either contract baseline (Contracts/DotNetSDKAPI.net6.json / DotNetPreviewSDKAPI.net6.json). ContractEnforcement will fail on the added public member — which lines up with the Microsoft.Azure.Cosmos.Tests job still showing pending rather than green. Could you run the contract-update script and commit the regenerated baseline? That should turn the job green.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -815,6 +815,18 @@ public CosmosSerializer Serializer
...
+        /// Gets or sets a value indicating whether the barrier early yield on 429
+        /// optimization is enabled. When true, ConsistencyWriter and QuorumReader
+        /// return early when all replicas return 429 (Too Many Requests) during a
+        /// write or read barrier request, allowing the ResourceThrottleRetryPolicy
+        /// to handle the retry instead of spinning until timeout.
+        /// </summary>
+        /// <value>
+        /// The default value is true.
+        /// </value>
+        public bool EnableBarrierEarlyYieldOn429 { get; set; } = true;
```
ananth7592: Two part question,

1. Public surface area : Some of our resiliency tuning (replica selection .etc) has been internal to the SDK. Is a customer facing knob intentional?
2. Default value of True:  Every customer's write/read barrier behavior under all-429 changes on upgrade (early yield → synthesized 408/21013 instead of spinning to timeout). Worth calling it out in Changelog

Praveen-Msft: for #1, Yes, this is intentional. Customers can opt out if they do not want this behavior.
for #2, currently without this PR, its already True for customers. I will add a changelog item.

ananth7592 (dismissed):
LGTM

Praveen-Msft:
> DotNetPreviewSDKAPI.net6.json

Added contract update.

ananth7592 (dismissed):
LGTM

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
kundadebdatta: @Praveen-Msft : Are we making this as a public contract change ? Or just keeping it behind an environment variable is just enough ?

Praveen-Msft: For now, going with internal change.

kirankumarkolli: #1: its requirement only for internal Customers not SDK users right?

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -425,10 +425,16 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
...
+                // Do not mark the endpoint unavailable when the 408 is synthesized by
+                // ConsistencyWriter for barrier throttling (substatus 21013). That 408 is
+                // an early-yield signal, not a connectivity failure, and marking the
+                // endpoint unavailable would trigger unnecessary cross-region failover.
+                //
                 // For DTX commits, a 408 from the coordinator means "transaction in-progress" — NOT
                 // an endpoint reachability problem. Marking the endpoint unavailable here would poison
                 // routing for non-DTX traffic sharing the same partition-key-range cache.
-                if (!this.isDtxRequest)
+                if (subStatusCode != SubStatusCodes.Server_WriteBarrierThrottled
```
kirankumarkolli, line 482: Why is this change needed? With default ON behavior wise no other changes are needed right?

kirankumarkolli: Actual functional change LGTM but might need to a separate one.

Praveen-Msft: Updated the comment with reasoning.

kundadebdatta (commented):
LGTM Now.

kundadebdatta (dismissed):

kundadebdatta:
@Praveen-Msft : Let's Please update the change log and PR description mentioning the direct version bump will also bring in 408 substatus codes for barrier failures.

kundadebdatta (dismissed):
LGTM

NaluTripician (commented):
Reviewed this locally with a deep-review pass plus an adversarial skeptic pass. The plumbing and the `ClientRetryPolicy` guard look correct and idiomatic (mirrors how `EnableTcpConnectionEndpointRediscovery` is threaded through).

One blocking concern: the documented retry mechanism doesn't match the V3 retry chain on the write path, and that core behavior isn't covered by a test. Details inline. Main asks:
1. Confirm/document which layer actually retries the synthetic 408/21013 and fix the changelog + XML docs.
2. Add an end-to-end pipeline test that asserts the terminal client outcome for a 408/21013.
3. Consider an env-var kill switch given this is default-on and shipping alongside a Direct bump.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -425,10 +425,21 @@ private async Task<ShouldRetryResult> ShouldRetryInternalAsync(
...
+                // 21013 (Server_WriteBarrierThrottled) as an early-yield signal → SDK
+                // receives 408/21013 here. This is NOT a connectivity failure, and
+                // marking the endpoint unavailable would trigger unnecessary cross-region
+                // failover.
+                //
                 // For DTX commits, a 408 from the coordinator means "transaction in-progress" — NOT
                 // an endpoint reachability problem. Marking the endpoint unavailable here would poison
                 // routing for non-DTX traffic sharing the same partition-key-range cache.
-                if (!this.isDtxRequest)
+                if (subStatusCode != SubStatusCodes.Server_WriteBarrierThrottled
```
NaluTripician, line 482: **Blocking** — the documented retry mechanism doesn't match what actually happens on the write path.

The changelog and the new XML doc on `CosmosClientOptions.EnableBarrierEarlyYieldOn429` both say this lets `ResourceThrottleRetryPolicy` handle the retry. Tracing the V3 chain for writes:

1. `ConsistencyWriter` throws `RequestTimeoutException` (408) with substatus 21013.
2. This 408 branch now skips mark-unavailable and falls through, returning `null`.
3. `ShouldRetryAsync` then delegates to `throttlingRetry.ShouldRetryAsync(...)`, and `ResourceThrottleRetryPolicy.IsValidThrottleStatusCode` returns true **only for 429** — so a 408 yields `NoRetry`.

Net: a throttled *write* surfaces to the caller as a bare `RequestTimeoutException` with no V3 retry and no `Retry-After` — the opposite of the documented mechanism, and ambiguous since the write may or may not have committed. The read path surfaces a 429 (which *is* retried), so there's a read/write asymmetry.

The one thing I couldn't verify from outside the Direct binary: whether Direct 3.44.0 retries the 408/21013 internally before surfacing here. If it does, this guard is fine but the doc attribution to `ResourceThrottleRetryPolicy` is still wrong. Could you confirm which layer actually retries this, and either correct the docs or make V3 treat 408/21013 as throttle-equivalent (with backoff)?

changelog.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- [5976](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5976) Barrier Early Yield: Updates Direct package to `3.44.0` which includes ConsistencyWriter and QuorumReader yielding early with substatus 21013 (`Server_WriteBarrierThrottled`) when all replicas return 429 during barrier requests, allowing `ResourceThrottleRetryPolicy` to handle retries instead of spinning until timeout. Note: this Direct version bump introduces 408 responses with barrier-specific substatus codes (e.g., 21013) for barrier failures that previously surfaced as plain timeouts. Adds internal opt-out flag and guards `ClientRetryPolicy` from marking the endpoint unavailable on the synthetic 408/21013.
```
NaluTripician: Two accuracy fixes (the same wording is mirrored in the new XML doc on `CosmosClientOptions.EnableBarrierEarlyYieldOn429`):

1. *"allowing `ResourceThrottleRetryPolicy` to handle retries"* — see my note on `ClientRetryPolicy`; on the write path that policy returns `NoRetry` for the synthetic 408/21013.
2. *"barrier failures that previously surfaced as plain timeouts"* — the prior write-barrier failure surfaced as `GoneException` (410, `Server_GlobalStrongWriteBarrierNotMet`) per the negative test in this PR, not a plain timeout.

Worth correcting so the changelog reflects the real behavior.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
```diff
@@ -815,6 +815,18 @@ public CosmosSerializer Serializer
...
+        /// Gets or sets a value indicating whether the barrier early yield on 429
+        /// optimization is enabled. When true, ConsistencyWriter and QuorumReader
+        /// return early when all replicas return 429 during a write or read barrier,
+        /// allowing the ResourceThrottleRetryPolicy to handle the retry instead of
+        /// spinning until timeout.
+        /// </summary>
+        /// <value>
+        /// The default value is true.
+        /// </value>
+        internal bool EnableBarrierEarlyYieldOn429 { get; set; } = true;
```
NaluTripician, line 831: This is a default-**on** change to retry/failover behavior, but the only opt-out is this `internal` property — there's no `ConfigurationManager` env-var / `AppContext` override. The two sibling transport toggles evaluated at the same `StoreClientFactory` call site (`IsTcpChannelMultiplexingEnabled`, `IsTcpDnsDotSuffixEnabled`) both have env-var kill switches. If this causes a production issue there's no field mitigation short of an SDK redeploy.

Recommend adding `ConfigurationManager.IsBarrierEarlyYieldOn429Enabled()` (default true) and AND-ing it in at the `DocumentClient` call site, mirroring the adjacent features. Given the open question on the blocking comment, this de-risks rollout a lot.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6897,7 +6897,8 @@ private void InitializeDirectConnectivity(IStoreClientFactory storeClientFactory
                     dnsResolutionFunction: ConfigurationManager.IsTcpDnsDotSuffixEnabled()
                         ? DnsDotSuffixHelper.ResolveHostAsync
                         : null,
-                    chaosInterceptor: this.chaosInterceptor);
+                    chaosInterceptor: this.chaosInterceptor,
+                    enableBarrierEarlyYieldOn429: this.ConnectionPolicy.EnableBarrierEarlyYieldOn429);
```
NaluTripician: This flag is only passed to the **internally-created** `StoreClientFactory`. When a caller supplies their own `IStoreClientFactory` (the compute-gateway reuse path — the earlier branch in this method), the option is set on `ConnectionPolicy`/`CosmosClientOptions` but has no effect. Config says one thing, runtime does another, and it's invisible until someone debugs why early yield never fires.

Could you either document the limitation (XML remark + a `DefaultTrace` note in the external-factory branch when the flag is set) or propagate it onto a surface the supplied factory reads? Mainly want to confirm it's intended.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/BarrierEarlyYieldOn429Tests.cs
```diff
@@ -0,0 +1,983 @@
...
+                isThinClientEnabled: false);
+
+            retryPolicy.OnBeforeSendRequest(request);
+
+            DocumentClientException requestTimeoutException = CreateDocumentClientException(
+                request,
+                HttpStatusCode.RequestTimeout,
+                SubStatusCodes.Server_WriteBarrierThrottled);
+
+            await retryPolicy.ShouldRetryAsync(requestTimeoutException, CancellationToken.None);
```
NaluTripician, line 66: This test (and `RequestTimeoutWithoutSubStatusMarksEndpointUnavailable`) discard the `ShouldRetryAsync` result and only assert `Assert.IsNull(failoverInfo)` — so they verify the endpoint isn't marked unavailable, but never assert what the retry policy actually *decides* for a 408/21013. Combined with the `EndToEnd…` tests stopping at `ConsistencyWriter.WriteAsync` throwing, nothing here drives a 408/21013 through the full `ClientRetryPolicy` + `ResourceThrottleRetryPolicy` pipeline to assert the terminal client outcome (retry vs `NoRetry` / final exception).

That's exactly the gap behind my blocking comment. A single integration test — mock transport returns 201 then 429s, issue a write through the real retry chain, assert the terminal result — would pin this down.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/BarrierEarlyYieldOn429Tests.cs
```diff
@@ -0,0 +1,983 @@
...
+                ConnectionPolicy connectionPolicy = InvokeGetConnectionPolicy(options, clientId: 0);
+
+                Assert.AreEqual(
+                    enableBarrierEarlyYieldOn429,
+                    connectionPolicy.EnableBarrierEarlyYieldOn429);
+            }
+        }
+
+        [TestMethod]
+        public void ConsistencyWriterEarlyYieldFieldSetViaConstructor()
```
NaluTripician: These reflection-based field-presence tests assert *assignment* of a private field, not behavior — they break on any private-field rename and would still pass even if the field were never read. The behavioral tests below (`WriteBarrierThrows…`/`ReadBarrierReturns…` and the disabled variants) already cover the real true/false behavior, so these add brittleness without much signal. Suggest dropping the single-field reflection tests in favor of the behavioral ones.

Directory.Build.props
NaluTripician, line 6: Heads up: the only query-plan baseline delta in this PR (`QueryPlanBaselineTests.OffsetLimit/Top.xml`) is removal of `"embeddingParameterMap":null` from the serialized `QueryInfo` — i.e. Direct 3.44.0 carries an unrelated serialization-contract change riding along with the bump. Almost certainly benign, but worth a one-line confirmation with the Direct team that vector/embedding query plans are unaffected, plus maybe a changelog note.

Praveen-Msft: @ananth7592 confirmed this.

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
NaluTripician: ✅ Resolved at e552dbdf. The XML doc + changelog no longer claim `ResourceThrottleRetryPolicy` retries this — they now correctly state Direct retries the 429 internally and, once exhausted, surfaces the synthetic 408/21013 whose only SDK-side effect is suppressing the cross-region failover here. The new `RequestTimeoutWithWriteBarrierThrottledSubStatusReturnsNoRetry` test drives a real `ClientRetryPolicy` end-to-end and asserts `IsFalse(result.ShouldRetry)`, which matches the corrected description. Thanks for the thorough fix.

changelog.md
NaluTripician: ✅ Resolved. Both accuracy fixes are in: the `ResourceThrottleRetryPolicy` claim is gone, and ""plain timeouts"" is corrected to `410 (Server_GlobalStrongWriteBarrierNotMet)`. The env-var kill switch is also documented in the entry.

Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs
NaluTripician: ✅ Resolved at e552dbdf. `ConfigurationManager.IsBarrierEarlyYieldOn429Enabled()` (env var `AZURE_COSMOS_BARRIER_EARLY_YIELD_ON_429_ENABLED`, default true) is now AND-ed into the value passed at the `DocumentClient` call site, mirroring the sibling `IsTcpChannelMultiplexingEnabled` / `IsTcpDnsDotSuffixEnabled` pattern. There's now a field-toggleable kill switch.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
NaluTripician: ✅ Resolved. The external-`IStoreClientFactory` branch now carries an explanatory comment and emits a `DefaultTrace.TraceWarning` when the flag is set to false but won't take effect. Minor nit (non-blocking): the env-var `IsBarrierEarlyYieldOn429Enabled()` isn't consulted on this path either, so the warning only covers the option, not the env var — fine since the external factory owns its own StoreClient config.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/BarrierEarlyYieldOn429Tests.cs
NaluTripician: ✅ Resolved at e552dbdf. New `RequestTimeoutWithWriteBarrierThrottledSubStatusReturnsNoRetry` constructs a real `ClientRetryPolicy` (with its inner throttle policy) and asserts the terminal decision `Assert.IsFalse(result.ShouldRetry)` plus no failover — exactly the full-pipeline coverage that was missing. 👍

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/BarrierEarlyYieldOn429Tests.cs
NaluTripician: ✅ Resolved. The four single-field reflection tests are removed; the remaining reflection usage is the multi-hop `FullConstructorChainPropagatesEarlyYieldFlagToConsistencyWriterAndQuorumReader` propagation test, which is defensible since it verifies the flag flows through the whole StoreClient→ReplicatedResourceClient→ConsistencyWriter/QuorumReader chain.

NaluTripician (dismissed):
## Re-review — all prior findings addressed ✅

Re-reviewed at `7cb7285` (commit `e552dbdf` "Address PR review"). Every item from my earlier review is genuinely resolved, and I verified each against the code rather than taking the commit message at face value:

| # | Finding | Status |
|---|---------|--------|
| 🔴 B1 | Docs claimed `ResourceThrottleRetryPolicy` retries the synthetic 408/21013 (it returns `NoRetry`) | **Resolved** — XML doc + changelog rewritten to describe the real mechanism (Direct retries internally; SDK only suppresses failover). New `…ReturnsNoRetry` test asserts the terminal `ShouldRetry==false` through a real `ClientRetryPolicy`. |
| 🟡 R1 | No field-toggleable kill switch for a default-on change | **Resolved** — added `AZURE_COSMOS_BARRIER_EARLY_YIELD_ON_429_ENABLED` via `ConfigurationManager.IsBarrierEarlyYieldOn429Enabled()`, AND-ed at the call site (mirrors sibling toggles). |
| 🟡 R2 | Flag silently inert with an external `IStoreClientFactory` | **Resolved** — explanatory comment + `DefaultTrace.TraceWarning` on that path. |
| 🟡 I1 | No full-pipeline test for the 408/21013 terminal outcome | **Resolved** — `RequestTimeoutWithWriteBarrierThrottledSubStatusReturnsNoRetry`. |
| 🟢 S2 | Brittle single-field reflection tests | **Resolved** — removed; kept only the multi-hop constructor-chain propagation test. |
| 🟢 D1 | Direct 3.44.0 `QueryInfo` serialization delta | **Confirmed benign** — baseline change is solely the `embeddingParameterMap:null` drop; @ananth7592 / @Praveen-Msft confirmed. |

Note for other reviewers: the `DiagnosticsHandler.cs` (+47/-13) and `DiagnosticHandlerHelperTests.cs` changes visible in a raw commit-range compare came in via the merge from `main` (PR #5977) and are **not** part of this PR's `base...head` diff — no action needed here.

The remaining behavior — a write throttled past Direct's internal retries surfaces to the caller as a `RequestTimeoutException` (408) — is a Direct-layer decision outside this PR's scope and is now accurately documented. LGTM. 🚢

changelog.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- [5976](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5976) Barrier Early Yield: Updates Direct package to `3.44.0` which includes ConsistencyWriter and QuorumReader yielding early with substatus 21013 (`Server_WriteBarrierThrottled`) when all replicas return 429 during barrier requests. Direct retries the 429 internally and surfaces a synthetic 408/21013 to the SDK when retries are exhausted. The SDK's `ClientRetryPolicy` recognizes this substatus and avoids marking the endpoint unavailable, preventing unnecessary cross-region failover. Note: this Direct version bump introduces 408 responses with barrier-specific substatus codes (e.g., 21013) for barrier failures that previously surfaced as 410 (`Server_GlobalStrongWriteBarrierNotMet`). Adds internal opt-out flag and env-var kill switch (`AZURE_COSMOS_BARRIER_EARLY_YIELD_ON_429_ENABLED`).
```
kushagraThapar: Since we are updating the Direct package as part of this PR, usually the process is to include other customer affecting changes as part of the direct package release update. Please list out any other customer affecting changes as part of this changelog.

ananth7592 (dismissed):

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -6917,7 +6928,9 @@ private void InitializeDirectConnectivity(IStoreClientFactory storeClientFactory
                     dnsResolutionFunction: ConfigurationManager.IsTcpDnsDotSuffixEnabled()
                         ? DnsDotSuffixHelper.ResolveHostAsync
                         : null,
-                    chaosInterceptor: this.chaosInterceptor);
+                    chaosInterceptor: this.chaosInterceptor,
+                    enableBarrierEarlyYieldOn429: this.ConnectionPolicy.EnableBarrierEarlyYieldOn429
+                        && ConfigurationManager.IsBarrierEarlyYieldOn429Enabled());
```
kushagraThapar, line 6982: Thanks for adding the env-var kill switch here — that makes the rollout much safer. Could we also add a focused test for this exact plumbing path? Specifically, set `AZURE_COSMOS_BARRIER_EARLY_YIELD_ON_429_ENABLED=false`, construct the normal Direct client/`StoreClientFactory` path with the option otherwise enabled, and assert the value that reaches `ConsistencyWriter`/`QuorumReader` is `false` (restoring the env var in `finally`). That would make the rollback path much easier to trust during an incident.

Directory.Build.props
kushagraThapar, line 6: Since this bumps `DirectVersion`, could we also validate the FaultInjection artifacts before merge? The FaultInjection projects consume `$(DirectVersion)`, and the FaultInjection pipeline appears to have `pr: none`, so the main SDK checks may not catch a build/test/package break there. Triggering the FaultInjection release-gates pipeline (or equivalent build/test/package validation) should be enough.

kushagraThapar (commented):
Two small follow-ups from my pass. Thanks for iterating on this — the core retry plumbing is looking good.

Directory.Build.props
Praveen-Msft: Nalu helped with running it successfully on local. Looks like the release pipeline is broken at the moment and Nalu has a todo item for fixing it.

kushagraThapar (dismissed):
LGTM, thanks @Praveen-Msft

NaluTripician (approved):

kushagraThapar (approved):

PR #5977, opened 2026-06-24
Diagnostics: Fixes missing CPU/memory in diagnostics of failed and canceled operations

NaluTripician:
## Problem

CPU/memory (the `"System Info"` section: CPU%, memory, threadpool) is **missing from the diagnostics of operations that fail by exception** — timeouts, cancellations, exhausted retries. This is exactly when the data is most useful for incident debugging, and it has been reported on multiple internal-customer incidents across **different SDK versions (3.52.0 and 3.59.0)**, so it is **not** version-specific.

### Root cause

`DiagnosticsHandler.SendAsync` attached the `"System Info"` datum on the line *after* `await base.SendAsync(...)` returned:

```csharp
ResponseMessage responseMessage = await base.SendAsync(request, cancellationToken);
// ... only runs on success:
request.Trace.AddDatum("System Info", new CpuHistoryTraceDatum(systemUsageHistory));
return responseMessage;
```

When the inner pipeline throws (e.g. the operation is canceled / times out), that line is skipped, so the section is silently omitted. The background `SystemUsageMonitor` is always running on the public SDK — the data exists, it just was never attached on the failure path. (This is the only place CPU is *serialized* into diagnostics; `ClientSideRequestStatisticsTraceDatum` snapshots system usage per-request but never serializes it.)

## Fix

Move the capture into a `finally` block so it runs on both the success and the failure path, wrapped in a defensive `try/catch` so it can **never mask the operation's real exception** (best-effort contract preserved):

```csharp
try
{
    return await base.SendAsync(request, cancellationToken);
}
finally
{
    DiagnosticsHandler.TryAddSystemInfoToTrace(request); // swallows its own errors
}
```

### Why this approach

- **Smallest blast radius.** Keeps the single existing `"System Info"` location and leaves the **success-path diagnostics byte-for-byte unchanged** (verified — `TraceWriterBaselineTests` + 160 diagnostics/trace/handler unit tests pass).
- **No duplicate-key risk.** `DiagnosticsHandler` runs once per operation (it sits above the retry handler), so `AddDatum("System Info", …)` is called at most once per trace node.
- The alternative — serializing the snapshot `ClientSideRequestStatisticsTraceDatum` already holds — would duplicate CPU on the success path and change the success-path shape (breaking baselines), so it was not taken.

## Validation

The real customer scenario is a `CosmosOperationCanceledException`, which is created by `AbstractRetryHandler` *below* `DiagnosticsHandler` and serializes its diagnostics **lazily** from the root trace — so the datum attached during unwind is included. This is covered by two unit tests in `DiagnosticHandlerHelperTests`:

- `SystemInfoIsCapturedOnSuccessAndFailurePathsAsync` — asserts `"System Info"` is present for both a succeeding and a throwing inner handler.
- `SystemInfoIsCapturedWhenOperationIsCanceledAsync` — asserts `"System Info"` is present in the **`CosmosOperationCanceledException.Diagnostics`** itself (the actual customer-visible output).

Both **fail on the pre-fix code** (the canceled-path test reproduces the customer capture exactly: an OCE node with no System Info) and **pass after the fix**.

## Notes

- Customer-observable diagnostics change (System Info now appears for failed operations) → changelog entry added under **Bugs Fixed**.
- The changelog PR-number reference is a placeholder pending this PR number.
- Separately recommended to affected customers (out of scope here): consolidate to a single `CosmosClient` per account and upgrade to ≥ 3.53.1 — the underlying incidents were client-side connection/CPU exhaustion (SSL-negotiation timeouts) under very high client counts.


FabianMeiswinkel (dismissed):
LGTM

FabianMeiswinkel (approved):
LGTM

ananth7592 (approved):
LGTM — clean, well-scoped fix. The diagnosis is precise (a single misplaced capture line that the exception path skipped), and keeping the existing `System Info` location rather than serializing the snapshot `ClientSideRequestStatisticsTraceDatum` already holds is the right call — it leaves the success-path diagnostics byte-for-byte unchanged and avoids touching baselines. The `finally` + defensive `try/catch` preserves the best-effort contract without ever masking the operation's real exception, and the second test reproducing the actual customer scenario (a `CosmosOperationCanceledException` created below the handler and serialized lazily, asserting on `caught.Diagnostics`) is exactly the coverage I'd want to see — it genuinely fails pre-fix.

One non-blocking thought for a follow-up: a short comment on `TryAddSystemInfoToTrace` noting it attaches to the current `request.Trace` node during unwind (which stays within the root trace the failing operation serializes) would preserve that subtle invariant for the next reader, since it's the quiet assumption the failure-path fix rests on. Nice work.

PR #5978, opened 2026-06-25
AddressRefresh: Adds repro for force address refresh on post-migration generic 410

NaluTripician:
## Issue

A reported Direct-mode request kept using stale physical replica addresses after a physical partition migration. Long after the move, the old replicas surfaced a generic `410` (substatus `0` / `21005` `ServerGenerated410`) rather than a recognized substatus.

There are two different recovery cases:

- A **physical migration** can move the replicas backing an existing partition key range while preserving the range topology. The range identity remains valid; its addresses need to be re-resolved.
- A **split or merge** changes which partition key ranges exist. The collection routing map must be refreshed before addresses can be resolved for the resulting range.

A generic `410/0` does not identify which case occurred. This PR adds a durable reproduction and diagnostics harness that measures the address-refresh behavior per substatus without treating migration and topology changes as equivalent or prescribing an unconditional retry policy.

## Verification

1. **`AddressRefreshForceRefreshPostMigrationTests`** (FaultInjection, Direct mode, live multi-region account) injects a `410` on a `ReadItem` for substatuses `{0, 21005, 1007, 1008, 1002}`. It records whether diagnostics contain the `ForceAddressRefresh` block emitted by the shared client refresh branch, whether a refresh record reported unchanged addresses, and whether the operation recovered. Only expected marker presence is asserted; `Recovered` and `Notes` are observational.
2. **`AddressRefreshForceRefreshUnitTests`** (unit, no account) pins the two distinct headers emitted by `GatewayAddressCache.GetServerAddressesViaGatewayAsync`:
   - `x-ms-force-refresh` bypasses the Gateway address cache and is driven by `forceRefreshPartitionAddresses`.
   - `x-ms-collectionroutingmap-refresh` bypasses the Gateway collection routing-map cache and is driven by `DocumentServiceRequest.ForceCollectionRoutingMapRefresh`.

   Their coupling is one-way. Requesting a routing-map refresh also makes the client re-resolve addresses, subject to the concurrent-update guard, but it does not set `x-ms-force-refresh`. Forcing addresses alone never requests a routing-map refresh.
3. Existing live-site diagnostics for the reported strong-consistency `504` were reviewed to correlate the retry behavior with the original symptom.

The shared client refresh branch runs when either `forceRefreshPartitionAddresses` or `ForceCollectionRoutingMapRefresh` is set. Its `ForceAddressRefresh` diagnostics block therefore proves that the branch ran, but does **not** identify which flag drove it or prove that `x-ms-force-refresh` was sent.

All diagnostics columns in the FaultInjection report are operation-level substring observations, not per-retry-attempt measurements. The `No change to cache` marker is emitted only when a refresh record reports identical addresses. `False` is ambiguous by itself: either no record was emitted or one reported different addresses (`Original` / `New`), so it must be read together with `RefreshBranchMarker`.

### Matrix result (Direct 3.43.1, live account)

| SubStatus | Category | RefreshBranchMarker | NoChangeToCache | Recovered |
|-----------|----------|---------------------|-----------------|-----------|
| 0 | `generic` | **True** | True | True |
| 21005 (`ServerGenerated410`) | `generic` | **True** | True | True |
| 1007 (`CompletingSplit`) | `split-in-progress` | False | False | True |
| 1008 (`CompletingPartitionMigration`) | `physical-migration` | True | True | True |
| 1002 (`PartitionKeyRangeGone`) | `pkrange-gone` | False | False | surfaced 410/1002 |

This repository exposes no distinct merge-completion substatus. The `1002` row represents the range-gone condition by substatus semantics only: fault injection against a healthy account did not remove a real range. Likewise, recovery after the `1007` injection did not exercise a real split. In the two rows where both `RefreshBranchMarker` and `NoChangeToCache` are false, no refresh-branch record was emitted; that does not show that the cache changed.

## Result

**H1 remains unresolved.** On Direct 3.43.1 the generic `410/0` and `410/21005` rows emitted the shared refresh-branch marker. Because either refresh flag can produce that marker, the matrix does not establish whether Direct sent `x-ms-force-refresh`, requested a collection routing-map refresh, or did both.

The collection routing-map decision remains open. No assignment to `ForceCollectionRoutingMapRefresh` exists in the open v3 source, and the current diagnostics do not attribute branch entry to either flag. Because a generic 410 does not distinguish a physical migration from a split/merge, the Direct retry policy should be inspected to determine which flag it sets and whether the policy should change. That follow-up is tracked by #6068.

The in-tree bulk retry path refreshes routing ranges for recognized `1002`, `1007`, and `1008` responses, but not for generic `410/0`. This demonstrates an existing mechanism and precedent; it does not establish the correct Direct policy for every generic 410.

The reported live-site failure also had a separate timing component: it was a strong-consistency quorum read against an unhealthy replica set, and channel acquisition consumed the request deadline before the Gone retry completed a logged cache refresh. The diagnostics contain the initial address resolution but no refresh-branch marker.

The decisive Gone-retry behavior remains in `Microsoft.Azure.Cosmos.Direct`, so this PR intentionally makes no shipped `src/**` behavior change. It adds regression coverage that can be rerun against future Direct versions while #6068 investigates the policy.

## How it was tested

- `AddressRefreshForceRefreshUnitTests`: 2 passed locally.
- `AddressRefreshForceRefreshPostMigrationTests`: previously executed against a live multi-region account on Direct 3.43.1 and produced the pinned matrix above. It remains gated by `COSMOSDB_MULTI_REGION`.
- Both affected test projects build cleanly.

## Changelog

No changelog entry required: test-only changes under `FaultInjection/tests/**` and `Microsoft.Azure.Cosmos.Tests/**`; no shipped package behavior changes.


NaluTripician:
Follow-up issue filed for the routing-map axis: #6068 (assigned to @xinlian12 / @kundadebdatta). Scoped to be picked up once Direct source merges into this repo, since whether `ForceCollectionRoutingMapRefresh` is set on a generic 410 can't be answered from the closed-source binary.

aavasthy (dismissed):

Pilchie:
Overall, I'm a bit concerned when reading the description for this PR that it conflates `migration`, which would change addresses, but not what partitions exist with `splits`/`merges`, where both the address and the actual partition would change.

For example, this sentence seems incorrect.

> This distinction matters post-migration, where the set of PK ranges itself changes and re-resolving the addresses of a range that no longer exists cannot recover the request.

Can you take a pass through this, and confirm that it's correct for both migrations *and* splits? My hunch is that if we don't know whether it was a migration or a split because we only get a 410/0, then we need to refresh the map as well as the addresses?

NaluTripician:
You're right - the previous description conflated two different recovery cases. A physical partition migration preserves the range identity while its replica addresses can change; a split or merge changes the range topology and requires the map to be re-resolved before resolving addresses for the resulting range.

While correcting that, I found an additional diagnostics nuance: the `ForceAddressRefresh` block is emitted by the shared client refresh branch, which can be entered by either `forceRefreshPartitionAddresses` or `ForceCollectionRoutingMapRefresh`. It therefore does not prove that `x-ms-force-refresh` was sent or identify which flag Direct set.

I updated the matrix to report `RefreshBranchMarker` rather than `ForcedRefresh` and changed the conclusion to **H1 unresolved**. On the pinned Direct 3.43.1 run, generic `410/0`, `410/21005`, and `410/1008` emitted the marker; `410/1007` and `410/1002` did not. The healthy-account `1007` recovery did not exercise a real split, and the injected `1002` did not prove that a real range was gone.

Because generic `410/0` does not distinguish physical migration from split/merge, I agree that both refresh decisions need to be inspected. I kept the policy question open rather than claiming that every generic 410 must unconditionally refresh both caches; #6068 now tracks determining which flags Direct sets and whether that policy should change.

The PR body, both test files, and #6068 are aligned in commit `a9cb76147c118e140f09538de4a21a98e590bd5e`.

Pilchie (approved):
Update the description but LGTM now. Thanks!

PR #5983, opened 2026-06-29
Direct: Fixes IdleTcpConnectionTimeout silent truncation of sub-second and fractional values

### LINKED ISSUE #5874: Direct: IdleTcpConnectionTimeout TimeSpan contract not honoring sub-second and fractional values (mirror of #4500)
## Description

`IdleTcpConnectionTimeout` has the same silent truncation issue that was fixed for `OpenTcpConnectionTimeout` in #5873 (Fixes #4500).

At `DocumentClient.cs:925`, the conversion is:
```csharp
this.idleConnectionTimeoutInSeconds = (int)connectionPolicy.IdleTcpConnectionTimeout.Value.TotalSeconds;
```

This `(int)` cast silently truncates:
- Sub-second values (e.g. `TimeSpan.FromMilliseconds(500)`) to 0
- Fractional values >= 1s (e.g. `TimeSpan.FromSeconds(10.5)`) down to the next whole second
- Negative values to 0

Additionally, the public setter on `CosmosClientOptions.IdleTcpConnectionTimeout` does not validate for negative values.

## Expected behavior

Apply the same contract as `OpenTcpConnectionTimeout` (per #5873):
- Negative `TimeSpan` should throw `ArgumentOutOfRangeException` at the setter.
- Sub-second values should normalize to 0 at the transport boundary.
- Fractional values >= 1s should round up via `Math.Ceiling`.

## Related

- #4500 (original issue for `OpenTcpConnectionTimeout`)
- #5873 (fix PR for `OpenTcpConnectionTimeout`)

virajchogle:
## Description

Follow-up to #5873, which fixed the same silent-truncation bug for the sibling property `OpenTcpConnectionTimeout` and explicitly called `IdleTcpConnectionTimeout` out as the planned follow-up. This applies the identical contract.

Previously `DocumentClient` cast `connectionPolicy.IdleTcpConnectionTimeout.Value.TotalSeconds` directly to `int`, silently truncating sub-second values to 0 and fractional values >= 1s down to the next whole second.

**Sub-second and fractional values (the fix):**
- The customer-supplied `TimeSpan` is preserved unchanged on the SDK properties.
- At the int-typed Direct boundary (`DocumentClient.cs`):
  - Values in [TimeSpan.Zero, 1 second) normalize to 0 (no idle timeout; connections kept open indefinitely).
  - Values >= 1 second round UP via `Math.Ceiling`, so the SDK never silently shortens the requested timeout (e.g. 2.5s -> 3s, never 2s).

**Negative values (warning only, no enforcement):**
- A `DefaultTrace.TraceWarning` is emitted (gated on Direct mode, since the value is not consumed in Gateway mode) advising they are not recommended.
- The value is preserved on the property unchanged - no exception, no clamping - and truncated via `(int)TotalSeconds` at the boundary, preserving pre-PR behavior.
- This matches #5873's decision: throwing or clamping would be a behavioral breaking change for customers who already have negative values configured.

XML docs updated on `CosmosClientOptions`, `ConnectionPolicy`, and `CosmosClientBuilder`. Unit tests cover sub-second normalization, exact one-second preservation, fractional ceiling rounding, property round-trip, and negative pass-through.

## Type of change

## Closing issues

closes #5874

## Changelog

- [x] I have added a changelog entry under `### Unreleased` in `changelog.md` for the user-facing impact of this change


virajchogle:

PR #5985, opened 2026-06-30
Client Encryption: Fixes System.Text.Json (Stream) processor correctness and feed performance

adamnova:
## What
Hardens the internal System.Text.Json ("Stream") client-side-encryption processor for `Microsoft.Azure.Cosmos.Encryption.Custom` — correctness fixes plus a feed/query decrypt performance improvement — and versions the package as `1.1.0-preview01`. The Stream-vs-Newtonsoft processor selection is an **internal, non-public evaluation toggle** (there is no new public API surface); the default remains `Newtonsoft` on all target frameworks.

## Included
1. **Stream-processor correctness fixes** (TDD, fail-before / pass-after): pass-through strings and property names containing JSON escapes (`"` `\` `\n` `\uXXXX`) are no longer re-escaped; a `null` inside an encrypted object/array no longer corrupts the payload; encrypted-path matching is restricted to the top level (Newtonsoft parity) so a nested same-named property is not mis-encrypted; integral `Double` (`5.0`) type preserved; out-of-range integer rejected; duplicate top-level `_ei` guard; non-object `_ei` read parity; numeric-string `_ef` parity; stream-input disposal.
2. **Lazy `DecryptableItem` honors the configured processor** — point reads, `FeedResponse<DecryptableItem>`, and the write-response wrappers previously always decrypted via Newtonsoft; deferred decryption now runs through the configured processor (decrypted output byte-identical).
3. **Feed/query Stream decrypt reads the response directly** — each page was previously materialized as an intermediate Newtonsoft `JObject` and every document re-serialized before the streaming decryptor ran; the response envelope is now copied through and each document decrypted from its raw bytes, eliminating the per-page and per-document Newtonsoft round-trips (net8, MDE only; the Newtonsoft path is untouched).

## Compatibility
- On-disk format unchanged (MDE format version 3, AEAD format version 2) — documents written by 1.x read on this build and vice-versa, with no re-encryption or migration.
- **No public API change**: the `JsonProcessor` enum and its selection methods are `internal`; the container default and per-call selection are not part of the public surface. Default is `Newtonsoft`.
- Version `1.1.0-preview01` (minor). Dependency updates (MDE `0.2`→`2.0`, `Microsoft.Extensions.Caching.Abstractions` replacing the unused `Caching.Memory`, `Azure.Core` now transitive via `Azure.Identity`) are documented in the changelog under **Other Changes** with migration notes.

## Testing
- 626 unit tests pass (net8), 0 warnings; the public-API contract baseline is updated (only the now-internal members removed).
- Validated by a cross-version data-compat matrix (companion harness in #5986): `1.0.0-preview07 ↔ 1.1.0-preview01`, both write/read directions, MDE + AEAD, both processors, point/query/feed — all cells pass with no data break.


adamnova:
Superseded — closing in favor of #5992.

This PR was branched from an early, unmerged revision of #5478. Upstream has since merged the final #5478 into `main`, which already provides the container-default `UseStreamingJsonProcessingByDefault`, the lazy `StreamDecryptableItem` honoring the container default, and streaming feed/query decrypt — i.e. ~90% of what this PR carried.

The remaining unique value is a set of JSON-fidelity corruption fixes in the System.Text.Json (Stream) processor. Those have been rebased onto current `main` and re-opened as #5992 with a correct branch name and clean history.

PR #5986, opened 2026-06-30
Client Encryption: Adds cross-version compatibility matrix test harness

adamnova:
## What
A test-only cross-version compatibility matrix for Encryption.Custom: two version-pinned subprocesses share one Cosmos database and cross-read every write/read combination, asserting a document written by one version is readable by the other. No shipped source changes.

## Coverage
- Old ↔ new, both write/read directions × algorithm (MDE / AEAD) × processor (Newtonsoft / System.Text.Json) × read path (point / query / feed).
- Newtonsoft-vs-System.Text.Json **A/B** with a cross-processor equivalence assertion (the same document decrypts identically under both). Processor selection uses the untyped request-options property key, so the harness is independent of the package's internal processor API and compiles against any version.
- An offline contract test enumerates and locks the exact matrix cells. Anti-fake-green controls: a version-distinctness guard (the two subprocesses must load different package versions) and a plaintext-rejection tamper check.
- Opt-in `Current` node builds the package from in-repo source (ProjectReference) as continuous regression infrastructure; the default run compares the two pinned versions.

## Result
Runs against the Docker Linux Cosmos emulator: `1.0.0-preview07 ↔ 1.1.0-preview01` — **PASS=42, no data break**.


adamnova:
Superseded by #6044, which replaces this harness with isolated released/current workers integrated into the existing emulator tests, removes the PowerShell/local-feed infrastructure, and includes the stream-fidelity fixes discovered by the compatibility matrix. Closing this draft in favor of the replacement architecture.

Issue #5982, opened 2026-06-28
Cosmos emulator gateway HTTP/2: cancelled read on cold/slow emulator leaves an unobserved read-loop task (UnobservedTaskException)


thomhurst:
When the .NET SDK talks to the Cosmos DB emulator in **gateway mode over HTTP/2** (ALPN-negotiated), a read that is cancelled by the request timeout — common while the emulator is still cold and returning 503s — leaves the underlying HTTP/2 read-loop task **unobserved**. It surfaces later as a `TaskScheduler.UnobservedTaskException` / finalizer `AggregateException` (HTTP/2 stream aborted). The application call itself fails/cancels as expected, but the orphaned task fault trips unobserved-exception observers and, if `ThrowUnobservedTaskExceptions` is set, can destabilize the host.

This looks like the same class as #4010 ("Unobserved task exception on gateway failure", fixed 2023) and the still-open #5045, but specific to the **emulator gateway + HTTP/2 + cancellation** path.

### Repro

1. Run the Cosmos DB emulator (the Aspire `RunAsEmulator` preview/vnext image, gateway mode).
2. Create a `CosmosClient` against it with the default request timeout (5s).
3. Issue a read shortly after startup, while the emulator is still warming and returning 503s, so the read hits the timeout and is cancelled.
4. Observe an `UnobservedTaskException` / finalizer `AggregateException` for an aborted HTTP/2 read, separate from the (correctly surfaced) cancellation on the awaited call.

### Expected

A cancelled / timed-out read should not leave an unobserved faulted task — the SDK should observe (or swallow) the HTTP/2 read-loop task on cancellation, as was done for the gateway cache-refresh path in #4010.

### Environment

- `Microsoft.Azure.Cosmos` (.NET v3 SDK), gateway connection mode
- Cosmos DB emulator (vnext-latest / preview), HTTP/2 via ALPN
- Trigger: request timeout (default 5s) firing during the emulator cold-start 503 window

### Workaround (for reference)

Pinning the emulator's `HttpClient` to HTTP/1.1 removes the ALPN HTTP/2 read loop entirely, so a cancelled read cannot orphan a read-loop task:

```csharp
new CosmosClientOptions
{
    HttpClientFactory = () =>
    {
        var http = new HttpClient(new SocketsHttpHandler())
        {
            DefaultRequestVersion = HttpVersion.Version11,
            DefaultVersionPolicy = HttpVersionPolicy.RequestVersionExact,
[...]

PR #5987, opened 2026-06-30
ThinClient: Fixes unobserved HTTP/2 task by disposing retried responses

NaluTripician:
## Issue

Fixes part of #5982 — gateway HTTP/2 (thin-client) cancelled/slow reads against a cold emulator leave an unobserved `TaskScheduler.UnobservedTaskException` (`Http2StreamException` "stream aborted").

## Root cause (two sub-causes)

The HTTP/2 path in #5982 is the **thin-client / Gateway V2 ("vnext") path** (`ThinClientStoreClient` sets `Version = 2.0`), which uses `HttpTimeoutPolicyForThinClient` (per-attempt 6s/6s/10s, retries on **408**) — not classic gateway (HTTP/1.1, 65s). There are two distinct sub-causes:

- **B (fixed here, confirmed):** on a 408 the thin-client policy retries, but the dropped 408 `HttpResponseMessage` was **never disposed** (`CosmosHttpClientCore.SendHttpHelperAsync`, after the `isOutOfRetries` check). With `ResponseHeadersRead` over HTTP/2 this leaks a live response stream until GC finalization, where a stream abort surfaces as the unobserved exception.
- **A (runtime behavior, documented):** cancelling an in-flight HTTP/2 `SendAsync` mid-flight leaves a **runtime-internal** stream-teardown task unobserved ([dotnet/runtime#46961](https://github.com/dotnet/runtime/issues/46961)). The SDK holds no handle to it, so this is addressed via changelog guidance (use a current .NET runtime, or pin the emulator client to HTTP/1.1), not a global handler.

## Change

- `CosmosHttpClientCore.SendHttpHelperAsync`: dispose the retriable response before the retry loop continues, so the HTTP/2 stream is torn down deterministically instead of at finalization.
- Regression test `ThinClientRetriableResponsesAreDisposedBeforeRetryAsync` — **verified to fail without the fix** and pass with it.
- Changelog `Bugs Fixed` entry (customer-facing) with runtime guidance.

## Validation

- Full `CosmosHttpClientCoreTests` class: **20/20 passed**, no regressions.
- No public API surface change (internal `Dispose()` only).

## Out of scope / follow-up

- A real-transport (in-proc Kestrel HTTP/2 or vnext emulator) characterization repro for **sub-cause A** is a documented follow-up (tracked in #5996) — it belongs in an integration test project, and A is a .NET runtime behavior the SDK cannot deterministically eliminate.
- No process-global `TaskScheduler.UnobservedTaskException` handler is introduced.


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
```diff
@@ -800,6 +801,79 @@ public void HttpTimeoutPolicyForWriteDistributedTransaction_ReturnsDefault()
...
+
+            Assert.AreEqual(2, droppedResponseContents.Count, "Two 408 responses should be produced before success.");
+            foreach (DisposeTrackingContent content in droppedResponseContents)
+            {
+                Assert.IsTrue(content.IsDisposed, "Dropped retriable (408) response should be disposed before the retry.");
+            }
+
+            // The disposed 408 responses still serialize their status into diagnostics without throwing
+            // ObjectDisposedException, proving the retriable-response disposal does not corrupt diagnostics.
+            StringAssert.Contains(diagnostics, HttpStatusCode.RequestTimeout.ToString());
```
ananth7592, line 879: One small thing, non-blocking. This assertion drives the trace-writer path (StatusCode/ReasonPhrase) over the disposed responses, which is great. But the consumer your L433-438 comment specifically calls out as needing ObjectDisposedException handling is the OTel metrics path — GetPayloadSize -> GetNetworkMetricsValues — which is the only place that actually reads .Content on the disposed response. Would it be worth extending this test to also drive that metric path over a disposed 408 (content with no Content-Length header, so it hits the guarded fallback)? That'd close the loop on the exact claim the comment makes, so a future refactor of that guard can't silently regress. Totally fine to leave for a follow-up.

ananth7592 (dismissed):
Really clean, surgical fix. Two things stood out as done well: the scoping is honest — sub-cause A (the runtime-internal orphaned task on mid-flight cancellation, dotnet/runtime#46961) is explicitly not claimed as fixed, it's documented with runtime guidance and tracked as follow-up, and there's no process-global UnobservedTaskException handler smuggled in. And the regression test goes further than just proving disposal — it serializes the trace over the now-disposed responses to prove the genuinely risky part (disposing a still-referenced object) stays dispose-safe. Left one small non-blocking note on the test. Approving. Happy to discuss.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosHttpClientCoreTests.cs
NaluTripician: Good catch — you're right that the trace-writer assertion only covered `StatusCode`/`ReasonPhrase`, and the consumer the `SendHttpHelperAsync` comment actually leans on is the OTel metrics path. Took the suggestion rather than deferring it (e0e49f0).

What changed:

- `DisposeTrackingContent.TryComputeLength` now returns `false`, so the dropped 408 carries no `Content-Length`. That makes the eagerly-captured `HttpResponseStatistics.ResponseContentLength` `null`, which is what forces `GetNetworkMetricsValues` down the `GetPayloadSize` fallback instead of short-circuiting on the cached value (asserted explicitly via `Assert.IsNull(stat.ResponseContentLength, ...)`).
- The test now pulls `datum.HttpResponseStatisticsList`, filters to the two recorded 408s, and runs `CosmosDbMeterUtil.GetNetworkMetricsValues` over each, asserting a 0-byte body.
- Added a precondition assert that reading `content.Headers.ContentLength` on the disposed content really does throw `ObjectDisposedException`. Without it the metrics assertion could pass for the wrong reason (e.g. if a runtime change stopped throwing) and quietly stop covering the guard.

On the "can't silently regress" part — I verified it both ways: with the `try`/`catch (ObjectDisposedException)` removed from `CosmosDbMeterUtil.GetPayloadSize`, the test fails with `ObjectDisposedException` thrown from `HttpContent.CheckDisposed` -> `GetComputedOrBufferLength` -> `HttpContentHeaders.get_ContentLength` -> `GetPayloadSize` -> `GetNetworkMetricsValues`; restored, the full `CosmosHttpClientCoreTests` class is 20/20 green.

Also tightened the `SendHttpHelperAsync` comment to name `CosmosDbMeterUtil.GetNetworkMetricsValues` / `GetPayloadSize` so the claim and the test that backs it are findable from each other.

Branch is also refreshed against latest `main` — the changelog conflict is resolved and the PR shows mergeable again.

ananth7592 (approved):

kushagraThapar (approved):

PR #5988, opened 2026-06-30
[Internal] Tests: Adds live-account AAD integration tests

NaluTripician:
## What

Adds live-account AAD (Microsoft Entra ID) integration tests — the AAD counterpart to the existing key-based `[TestCategory("MultiRegion")]` live tests. They authenticate against a **real, AAD-only** Cosmos DB account (`dotnet-v3-multiregion`, key/local auth disabled) with a real `Azure.Core.TokenCredential` (data-plane RBAC), unlike the emulator `CosmosAadTests` which fabricates a token signed with the emulator master key.

Validated **8/8 passing** against the real account, Direct + Gateway.

## Changes

- **`CosmosAadLiveTests.cs`** (new, all `[TestCategory("MultiRegionAad")]`): account read, item CRUD, query, change feed (Direct + Gateway), a 403 control-plane assertion, and a token-refresh check. The DB/container are pre-created out of band; each test `Assert.Inconclusive` (skips) when the endpoint, credentials, resources, or role assignment aren't in place.
- **`Utils/TestCommon.cs`**: adds `GetAadAccountEndpoint()`, `GetAadTokenCredential()`, and `CreateAadCosmosClient()`. The endpoint resolves from `COSMOSDB_MULTI_REGION_AAD` (bare URL or connection string), falling back to `COSMOSDB_MULTI_REGION`. Credentials prefer a validated `ClientSecretCredential`; when secrets are blocked by tenant policy, `COSMOSDB_AAD_USE_DEFAULT_CREDENTIAL=true` uses `DefaultAzureCredential` with managed identity excluded (so the federated `az` / `az login` identity is used deterministically). Unresolved ADO macros validate as absent, so tests skip rather than fail.
- **`templates/build-test-aad.yml`** (new): dedicated `MultiRegionAad` job that runs `dotnet test` inside an `AzureCLI@2` task bound to a **Workload Identity Federation** service connection (no client secret). Gated on the connection name.
- **`azure-pipelines.yml` + `azure-pipelines-nightly.yml`**: wire the AAD job into **both PR and nightly**, configured for account `dotnet-v3-multiregion` and service connection `dotnet-v3-aad-ci`. Clear `AadServiceConnection` to skip the job (stays green).
- **`templates/build-test.yml`**: excludes `MultiRegionAad` from the general emulator jobs so there is no double-run.

## CI / account setup

The target account is AAD-only and tenant policy blocks client secrets, so CI authenticates via Workload Identity Federation: an `AzureCLI@2` task logs `az` in as the service connection's app-registration identity (co-tenant with the account, holding **Cosmos DB Built-in Data Contributor**), and the SDK's `DefaultAzureCredential` acquires data-plane tokens. Database `AadLiveTestDb` + container `AadLiveTestContainer` (`/pk`) are pre-created — a data-plane-only token cannot create them (403 Forbidden, asserted by a test).

## Testing

- `dotnet build` (Release): clean.
- `dotnet test --filter "TestCategory=MultiRegionAad"` against the real AAD-only account: **8/8 passed**.
- With no endpoint/credentials/service connection: tests **skip** and the job is **not scheduled** — green on PR and nightly.

## Type of change

- [x] No changelog entry required — test-only + pipeline-only change, no shipped-package source touched, no customer-observable impact.


NaluTripician:
### AAD account readiness — verified live (2026-07-08)

Confirmed the target account is fully provisioned for the `MultiRegionAad` tests and ran them end-to-end through the SDK.

**Account:** `dotnet-v3-multiregion` · RG `dotnet-v3-ci-pipelines` · sub `chrande-CosmosDB-SDK-CI` (`86886548-0a7c-430b-b16f-014095647de4`) · tenant `72f988bf`

| Requirement | State |
|---|---|
| AAD-only (`disableLocalAuth`) | ✅ true |
| Public network (OneES reachable) | ✅ Enabled |
| DB `AadLiveTestDb` | ✅ exists |
| Container `AadLiveTestContainer` (`/pk`) | ✅ exists |
| Built-in Data Contributor role def (`…0002`) | ✅ present |

**Local run** (`DefaultAzureCredential` → `az` identity holding Data Contributor, same env contract as `build-test-aad.yml`): **8/8 passed** — `AadReadAccountAsync`/`AadItemCrudAsync` (Direct + Gateway), `AadQueryAsync`, `AadChangeFeedAsync`, `AadControlPlaneIsForbiddenAsync` (403), `AadBackgroundTokenRefreshInterval`.

### Only remaining step — the CI identity
The job stays skipped until a CI identity exists (account side is done):
1. Create an **Azure DevOps ARM service connection (Workload Identity Federation)** in the org/project that runs `azure-pipelines.yml`, scoped to sub `86886548-…` (tenant policy blocks secrets → must be WIF).
2. Assign Data Contributor to it:
   `az cosmosdb sql role assignment create -a dotnet-v3-multiregion -g dotnet-v3-ci-pipelines --role-definition-id 00000000-0000-0000-0000-000000000002 --principal-id <SC-identity-objectId> --scope "/"`
3. Set `AadServiceConnection: '<connection-name>'` in `azure-pipelines.yml` **and** `azure-pipelines-nightly.yml` to un-skip the job.

NaluTripician:
### Runbook: dual-auth (key + AAD) + merge + go-live

Capturing the plan after the CI-strategy sync (AAD test accounts stay **fixed**; key-based accounts move to ephemeral tenants; reuse the current account with **both** key and AAD enabled).

#### Key concept
There is no account-level `enable AAD` switch. Every Cosmos account already accepts Entra tokens on the data plane — **enabling AAD == assigning an RBAC role**. `disableLocalAuth` is independent and only blocks the **key**. So *AAD alongside key* = local auth **on** + data-plane role assignments.

Current state (sub `chrande-CosmosDB-SDK-CI` / RG `dotnet-v3-ci-pipelines`):
| Account | disableLocalAuth | data-plane role assignments |
|---|---|---|
| dotnet-v3-multiregion | true (AAD-only) | 1 (a user, Data Contributor) |
| dotnet-v3-multimaster | true (AAD-only) | none |
| dotnet-v3-multimaster-multiregion | false (key works) | none |

#### A. Make an account dual (key + AAD)
```bash
# keep keys working
az cosmosdb update -n <acct> -g dotnet-v3-ci-pipelines --disable-local-auth false
# enable AAD == assign the data-plane role to the caller (Built-in Data Contributor = ...0002)
az cosmosdb sql role assignment create -a <acct> -g dotnet-v3-ci-pipelines \
  --role-definition-id 00000000-0000-0000-0000-000000000002 \
  --principal-id <caller-objectId> --scope "/"
```
The `MultiRegionAad` suite passed **8/8** on `dotnet-v3-multiregion` as AAD-only; it also passes in dual mode (AAD is unaffected by the local-auth flag, and `AadControlPlaneIsForbiddenAsync` still returns 403 because data-plane roles never grant control-plane).

#### B. Merge this PR
Title lint OK; no changelog needed (test+pipeline only). The PR-pipeline AAD job **auto-skips green** while `AadServiceConnection` is empty, so merging is safe *before* the CI identity exists. Squash-merge after approval.

#### C. Go live (the real gate = CI identity)
The `AzureCLI@2` task needs an ARM identity in ADO. Options, best first:
- **C1 – reuse an existing ARM service connection** bound to sub `86886548-…`: assign *its* identity the role (step A) and set `AadServiceConnection` in both `azure-pipelines.yml` + `azure-pipelines-nightly.yml`.
- **C2 – OneES agent managed identity**: requires a small code change — `TestCommon.GetAadTokenCredential()` sets `ExcludeManagedIdentityCredential = true` on purpose; would add an opt-in flag to allow MI, then assign the agent MI the role.
- **C3 – new WIF service connection**: only if C1/C2 aren't possible (this is the one blocked on the engineering-systems team).

#### Doc note
The test-class summary says *"AAD-only account (local/key auth disabled)"*. If we adopt dual accounts, that line could soften to *"AAD-enabled account (data-plane RBAC; AAD-only or dual key+AAD)"* — the tests already work in both modes, so this is cosmetic.

xinlian12 (dismissed):
LGTM, thanks

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAadLiveTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                await this.container.ReadContainerAsync();
+            }
+            catch (CosmosException ex) when (ex.StatusCode == HttpStatusCode.NotFound)
+            {
+                Assert.Inconclusive($"Pre-create database '{DatabaseId}' and container '{ContainerId}' (/pk) on the AAD account before running these tests.");
+            }
+            catch (CosmosException ex) when (ex.StatusCode == HttpStatusCode.Forbidden || ex.StatusCode == HttpStatusCode.Unauthorized)
+            {
+                Assert.Inconclusive("The AAD principal is missing the Cosmos DB data-plane role assignment (Cosmos DB Built-in Data Contributor).");
```
kushagraThapar: Could we make these prerequisite failures fail closed in CI? Converting 401/403/404 to `Assert.Inconclusive` is convenient for local opt-in runs, but MSTest can still exit successfully when every `MultiRegionAad` case is inconclusive. A missing role assignment, stale endpoint, or missing fixture could therefore leave the dedicated lane green without validating any Entra-authenticated operation. Could we retain the local skip behavior while enabling a CI-strict mode that fails these cases, and gate the published results on exactly 8 passed with 0 skipped/inconclusive? A useful falsification test would point the lane at a nonexistent container and verify that the job fails.

NaluTripician: Good catch, and you are right about the failure mode. Fixed in 3bd429037 with two independent fail-closed guards, while keeping the local skip behavior.

**1. CI-strict mode (`COSMOSDB_AAD_STRICT`)**

`CosmosAadLiveTests.SkipOrFail(...)` replaces the bare `Assert.Inconclusive` calls: unset (the local default) it still skips, but when `COSMOSDB_AAD_STRICT=true` every unsatisfied prerequisite becomes a hard failure — missing endpoint, missing credential, 401/403 from a missing role assignment, and 404 from a missing database/container. `templates/build-test-aad.yml` sets it via a new `StrictMode: true` parameter, so the dedicated lane always runs strict.

**2. Result gate on the published TRX**

A new `Verify MultiRegionAad results` step runs *after* `PublishTestResults@2` (so results stay visible when it fails) and requires exactly `ExpectedTestCount` (8) passed, 0 failed, 0 skipped/inconclusive — and fails closed when no TRX was produced at all. A preceding step deletes stale `MultiRegionAad.trx` files so the gate can never validate results the current run did not produce.

One thing worth calling out: the gate counts per-case `<UnitTestResult outcome="...">` nodes rather than `ResultSummary/Counters`. On an all-inconclusive run the VSTest TRX logger emits `total="8" executed="0" passed="0"` but leaves **`notExecuted="0"` and `inconclusive="0"`**, so a counters-based check reports zero skipped and would have sailed straight through.

**Falsification**

I ran the equivalent of your suggestion locally (unsatisfied prerequisite → same code path as a nonexistent container), and extracted the gate script verbatim from the YAML to run it against the resulting TRX files:

| Scenario | `dotnet test` | Gate |
|---|---|---|
| Today's behavior: prerequisites unmet, no strict mode | **exit 0**, `Skipped: 8` — exactly the vacuously-green lane you described | **fails**: `8 test case(s) were skipped/inconclusive` + `Expected exactly 8 passed but found 0` |
| `COSMOSDB_AAD_STRICT=true`, prerequisites unmet | **exit 1**, `Failed: 8` | **fails**: `8 live AAD test case(s) failed` |
| All 8 passing (synthetic TRX) | exit 0 | **passes** — proves the gate is not unconditionally red |
| 9 passing (case added without updating the count) | exit 0 | **fails** with a message naming both places to update |
| No TRX produced at all | — | **fails**: `No MultiRegionAad.trx was produced ... Failing closed` |

**Keeping the `8` honest**

Since the gate now hardcodes a count, I added `CosmosAadLiveTests.ExpectedTestCaseCount = 8` plus a reflection guard test (`CosmosAadLiveTestsGateTests`) that runs in the ordinary emulator lane — no live account needed. It counts `[TestMethod]`/`[DataRow]` cases, asserts each is tagged `MultiRegionAad`, and fails with an actionable message if the count drifts from the YAML parameter. Verified it fails when I deliberately set the constant to 9.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAadLiveTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            Assert.AreEqual(HttpStatusCode.Forbidden, exception.StatusCode,
+                "Creating a database with a data-plane-only AAD token should be Forbidden.");
+        }
+
+        [TestMethod]
+        [TestCategory("MultiRegionAad")]
+        public void AadBackgroundTokenRefreshInterval()
+        {
+            TokenCredentialCache tokenCredentialCache =
+                ((AuthorizationTokenProviderTokenCredential)this.aadClient.AuthorizationTokenProvider).tokenCredentialCache;
```
ananth7592: Double casted so any changes to the type of either may fail the test and the failure will look like a test bug rather than an intended signal.  Fine for an internal test assembly, but if there's a public way to assert "a background refresh interval is configured" it'd be less fragile. Non-blocking.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAadLiveTests.cs
ananth7592: +1

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosAadLiveTests.cs
NaluTripician: Good catch — reworked it in b3618a785.

There is a public way to assert the configured interval, so the test now leads with that: it builds a client with `CosmosClientOptions.TokenCredentialBackgroundRefreshInterval` set, reads it back off the public `CosmosClient.ClientOptions`, and does a live `ReadContainerAsync` so it's not just a property round-trip — it proves an Entra-authenticated client actually serves requests with that refresh schedule in place. That half can't break on an internal refactor.

The one thing the public surface can't express is the *derived* interval: `TokenCredentialBackgroundRefreshInterval` defaults to `null`, and when it is null `TokenCredentialCache` computes the interval as 50% of the real token's lifetime (`userDefined ?? system`). That derived value is the part that only means something against a live account with a real token, and there's no public accessor for it. So it still reaches into the auth plumbing, but the double cast is gone: it's now a single `as` plus an explicit `Assert.IsNotNull` that names the expected provider type and the type it actually found, and tells the reader to update the assertion if the plumbing was refactored on purpose. `TokenCredentialCache` is no longer named at all. A type change now fails with an intentional message instead of an `InvalidCastException` that reads like a test bug.

Case count is unchanged (still 8), so the CI result gate and `ExpectedTestCaseCount` didn't need touching.

kushagraThapar (approved):

xinlian12 (approved):
LGTM, thanks

Issue #5975, opened 2026-06-23
[FaultInjection][DTX] 429/3200 (RUBudgetExceeded) on a distributed-transaction commit is never retried despite being classified coordinator-retriable


jeet1995:
## Summary

A bodyless **429 / sub-status 3200 (RUBudgetExceeded)** envelope returned to a distributed-transaction (DTX/DTC) commit is **surfaced to the caller without any retry** (rule hit count == 1), even though:

- `ClientRetryPolicy.ShouldRetryDtxRequest` **explicitly classifies** `429 + RUBudgetExceeded` as `isCoordinatorRetriable` and **deliberately defers it** to `ResourceThrottleRetryPolicy` (returns `null`), implying it is expected to be retried, and
- the DTX response-status-code spec marks `429/3200` as **retriable = Yes**.

The deferral is effectively a **no-op** for the DTX commit path: neither the inner `ClientRetryPolicy` / `ResourceThrottleRetryPolicy` nor the committer outer loop retries it.

Found while building fault-injection coverage for PR #5969.

## Reproduction (live, deterministic — 3/3 looped runs)

Using the `RetriableCoordinatorResponse` fault-injection result with a **bodyless** envelope:

```
status = 429, subStatus = 3200, emptyBody = true, retryAfter = 1ms, suppressServiceRequest = true
```

Observed: **hit count == 1** every run; the 429 surfaces immediately with no retry. The same is seen via the standard `FaultInjectionServerErrorType.TooManyRequests` hybrid rows on a DTX commit/read (hit count == 1, 429 surfaced).

Contrast — under the identical harness, bodyless **408/0**, **449/5352**, and **500/5411-5413** are retried by `ClientRetryPolicy`'s DTX budgets to exactly their caps (11 / 11 / 10 hits respectively), so the inner retry loop is wired and active. Only the **429 deferral** path produces no retry.

## Relevant code

`Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs` — `ShouldRetryDtxRequest`:

```csharp
bool isCoordinatorRetriable =
    statusCodeValue == (int)HttpStatusCode.RequestTimeout
    || (statusCodeValue == (int)StatusCodes.RetryWith && subStatusCode == SubStatusCodes.DtcCoordinatorRaceConflict)
    || (statusCodeValue == (int)StatusCodes.TooManyRequests && subStatusCode == SubStatusCodes.RUBudgetExceeded);
...
[...]

PR #5989, opened 2026-06-30
Distributed Transactions: Fixes 429/3200 (RUBudgetExceeded) retry on DTX requests

Meghana-Palaparthi:
## Description

A bodyless `429`/`3200` (`RUBudgetExceeded`) response on a distributed-transaction (DTX)
commit or read arrives as an **exceptionless `ResponseMessage` whose `Content` is a
non-null but zero-length stream**. `ClientRetryPolicy` detected a body purely by
`Content != null`, so this empty envelope was misclassified as a body-bearing response.
The DTX classifier (`ShouldRetryDtxRequest`) treats body-bearing coordinator-retriable
responses as semantic per-operation results and defers them to the outer
`DistributedTransactionCommitter` loop — but that loop cannot act on an empty body
(it fails to parse, `IsRetriable` defaults to `false`), so the request was surfaced to
the caller with **no retry** (hit count == 1).

This contrasts with `408` and `449`/`5352`, whose bodyless envelopes were already being
retried correctly.

### Fix

- In `ClientRetryPolicy.ShouldRetryAsync(ResponseMessage)`, treat a **seekable,
  zero-length** `Content` stream as "no body." `Length` is only read when the stream is
  seekable; a non-seekable stream conservatively still counts as having a body.
- With `hasResponseBody == false`, a bodyless `429`/`3200` no longer takes the
  body-bearing deferral. It falls through the DTX classifier's `429` `return null` and is
  handled inline by the shared `ResourceThrottleRetryPolicy` (`this.throttlingRetry`) —
  the same policy every non-DTX `429` uses. Retries honor the server's
  `x-ms-retry-after-ms` header and the customer-configured rate-limit options
  (`MaxRetryAttemptsOnRateLimitedRequests` / `MaxRetryWaitTimeOnRateLimitedRequests`,
  defaults **9 attempts / 30s cumulative**), so a coordinator returning large retry-after
  values cannot stall a commit indefinitely.
- No change to the `408`/`449`/`5352` coordinator-retriable budget or the `500`/`5411`-`5413`
  infrastructure-failure budget.

### Type of change

## Tests

Added to `ClientRetryPolicyTests`:
- `DtxRequest_NonNullEmptyBody_RetriesInline` (DataRow: `408`, `449`/`5352`, `429`/`3200`) —
  the core regression guard: a non-null zero-length body now retries inline.
- `DtxRequest_429_3200_ShouldRetry_WithDefaultRetryInterval`
- `DtxRequest_429_3200_ShouldRetry_HonorsRetryAfterHeader`
- `DtxRequest_429_3200_StopsWhenDefaultCumulativeWaitExceeded`
- `DtxRequest_429_3200_HonorsConfiguredMaxRetryAttempts`
- `DtxRequest_429_3200_HonorsConfiguredMaxRetryWaitTime`
- `ReadDtxRequest_429_3200_ShouldRetry`

All `ClientRetryPolicyTests` pass; Release build (`-p:IsPreview=true`) is clean.

## Closing issues

[[FaultInjection][DTX] 429/3200 (RUBudgetExceeded) on a distributed-transaction commit is never retried despite being classified coordinator-retriable](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5975)

## Changelog

Added a `Bugs Fixed` entry under Unreleased.


Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -33,10 +33,13 @@ internal sealed class ClientRetryPolicy : IDocumentClientRetryPolicy
...
         // ----- DTX (Distributed Transaction) inner-loop retry constants -----
         // The outer loop (DistributedTransactionCommitter) handles body-bearing isRetriable failures.
-        // CRP owns envelope failures with empty body: 408, 449/5352 share one budget; 500/5411-5413 use a separate, tighter budget.
+        // CRP owns envelope failures with empty body: 408, 449/5352, 429/3200 share one budget; 500/5411-5413 use a separate, tighter budget.
+        // The coordinator-retriable budget is bounded both by attempt count and by a cumulative-wait cap (mirroring
+        // ResourceThrottleRetryPolicy's 60s default), so a stream of large server Retry-After values cannot stall a commit indefinitely.
         private const int MaxDtxRetryCount = 10;
         private const int MaxDtxInfraFailureRetryCount = 9;
         private const int DtxInfraFailureMaxExponent = 6;
+        private static readonly TimeSpan MaxDtxCoordinatorRetriableCumulativeWait = TimeSpan.FromSeconds(60);
```
jeet1995: Are we not using the `ResourceThrottleRetryPolicy` (from `CosmosClientOptions` here)? For RU-budget exceeded throttles, the customer expects configurability.

Meghana-Palaparthi: Fixed it.

Meghana-Palaparthi: I tested this change using msdata fault injection config and this is the outcome

image

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -206,7 +206,18 @@ public async Task<ShouldRetryResult> ShouldRetryAsync(
...
+            // A bodyless DTX gateway envelope (for example a 429/3200 RUBudgetExceeded) can arrive as an
+            // exceptionless ResponseMessage whose Content is a non-null but zero-length stream. A zero-length
+            // body carries no semantic per-operation result, so for DTX requests treat a seekable empty stream
+            // as "no body": this lets the response fall through the DTX classifier to the shared throttling
+            // retry policy (ResourceThrottleRetryPolicy) instead of being deferred to the outer commit loop,
+            // which cannot act on an empty body. The empty-stream reinterpretation (and the Length read it
+            // requires) is scoped to DTX requests, so the general retry path keeps its original Content != null
+            // semantics. Length is read only when the stream is seekable; a non-seekable stream conservatively
+            // counts as having a body.
+            System.IO.Stream responseContent = cosmosResponseMessage?.Content;
```
jeet1995: nit: Do we require a fully qualified type?

jeet1995 (dismissed):

FabianMeiswinkel (dismissed):
LGTM

jeet1995 (dismissed):

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
Meghana-Palaparthi: Not really, I can add a `using System.IO` instead

jeet1995 (approved):

kushagraThapar (approved):

PR #5990, opened 2026-07-01
Retry logic update: Adds retry logic for 449 across Gateway modes

aavasthy:
# Pull Request Template

## Description

This change unifies how HTTP 449 (RetryWith) responses are handled across the Gateway V1 (HTTP gateway) and Gateway V2 (thin-client) paths so that the retry is always orchestrated client-side by the SDK. A 449 simply means two operations briefly collided on the server and the request should be retried, so it is expected and self-healing. Previously the retry behavior was inconsistent between the two modes because the gateway could drive the 449 retry server-side, which made timing and behavior depend on which path a request took. This change makes the SDK the single, consistent authority for 449 retries.

The SDK now owns the 449 retry loop for both gateway modes. A new GatewayRetryWithRetryPolicy performs the client-side retry using a fast, jittered backoff — it starts at 10ms, doubles on each attempt up to a 1000ms cap, adds a small random salt so many clients don't retry in lockstep, and keeps retrying until an overall budget is exhausted (30 seconds by default, or 60 seconds when the account's default consistency is Strong). For Gateway V1, the SDK sends the x-ms-noretry-449 header so the gateway does not also retry the response on its own. For Gateway V2 (thin-client), the SDK deliberately does not send that header — the thin-client proxy already does not retry 449 for external requests — but the same client-side retry policy still runs, so both modes end up behaving identically. A 449 that exhausts the retry budget is logged at warning rather than error level, since it is a common, self-healing response and should not create noisy error logs.

## Type of change

## Closing issues

## Changelog

>


Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
```diff
@@ -69,6 +69,17 @@ protected override bool ShouldResolvePartitionKeyRange()
             return true;
         }
 
+        /// <summary>
+        /// The thin-client (Gateway V2) transport does not run a server-side 449
+        /// (<see cref="StatusCodes.RetryWith"/>) retry loop, so there is no server-side retry to opt out
+        /// of. The 449 retry is still orchestrated client-side via the inherited
+        /// <see cref="GatewayStoreModel.ProcessMessageAsync"/> loop. 
+        /// </summary>
+        protected override void ApplyGatewayRetryWithHeaders(DocumentServiceRequest request)
```
NaluTripician: This no-op is keyed on the store-model type, but `ApplyGatewayRetryWithHeaders` runs once *before* `DispatchAsync` picks a route. When thin-client dispatch falls back to `base.DispatchAsync` — an unroutable operation, or `!IsProxyEndpointHealthy(thinClientEndpoint)` — the request hits the real Gateway V1 HTTP path **without** `x-ms-noretry-449`. So the gateway retries the 449 server-side *and* `GatewayRetryWithRetryPolicy` retries it client-side: the exact double-retry inconsistency this PR is trying to remove, and it triggers on routine fallbacks.

Suggest keying the header on the transport actually used at dispatch time, or simply setting it unconditionally in the base `ProcessMessageAsync` (the PR notes the proxy ignores it anyway) and dropping this override.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
```diff
@@ -590,6 +590,127 @@ public async Task TestErrorResponsesProvideBody()
...
         }
 
+        /// <summary>
+        /// Gateway V1 must tell the gateway to suppress its server-side 449 (RetryWith) retry by
+        /// sending the <c>x-ms-noretry-449</c> header, so the SDK becomes the single client-side
+        /// authority for RetryWith retries.
+        /// </summary>
+        [TestMethod]
+        [Owner("aavasthy")]
+        public async Task GatewayStoreModel_GatewayV1_SetsNoRetry449Header()
```
NaluTripician, line 600: Good coverage on the V1 path. A few gaps worth closing:

- No ThinClient test asserting it does **not** send `x-ms-noretry-449` yet still retries 449 client-side — that test would likely have caught the fallback header gap in `ThinClientStoreModel`.
- Nothing exercises the Strong-consistency 60s budget (`GetRetryWithWaitTimeInSeconds`).
- Both store-model tests use `OperationType.Query`; a write (`Create`/`Replace`) 449-then-succeed would document the write-retry safety decision.
- No test that budget exhaustion propagates the final 449 `DocumentClientException` cleanly out through `ProcessMessageAsync` (the unit test only proves the policy stops).

Microsoft.Azure.Cosmos/src/GatewayRetryWithRetryPolicy.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        public Task<ShouldRetryResult> ShouldRetryAsync(
+            ResponseMessage cosmosResponseMessage,
+            CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
+        }
+
+        private static bool IsRetryWithException(Exception exception)
```
NaluTripician, line 114: This retries any 449 regardless of sub-status. `ClientRetryPolicy` already runs a budgeted client-side 449 loop for distributed transactions (`RetryWith` + `DtcCoordinatorRaceConflict`) and explicitly guards against inner×outer amplification. A **bodyless** DTx 449 throws here and would be retried by this inner 30/60s loop before the outer DTx budget is consulted (body-bearing DTx 449s are returned exceptionlessly, so those are fine).

Can we confirm DTx is meant to be excluded here and, if so, skip `DistributedTransactionBatch` / `DtcCoordinatorRaceConflict` so the outer DTx budget stays authoritative?

Microsoft.Azure.Cosmos/src/GatewayRetryWithRetryPolicy.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        private const int DefaultWaitTimeInSeconds = 30;
+        private const int MaximumBackoffTimeInMilliseconds = 1000;
+        private const int InitialBackoffTimeInMilliseconds = 10;
+        private const int BackoffMultiplier = 2;
+        private const int RandomSaltInMilliseconds = 5;
+
+        private readonly int waitTimeInSeconds;
+        private readonly Func<TimeSpan> getElapsedTime;
+        private readonly Random random = new Random();
```
NaluTripician, line 46: `new Random()` per policy instance (one is created per request) is seeded from `Environment.TickCount` on .NET Framework, so many policies constructed within the same tick — exactly the high-concurrency contention where 449s cluster — get identical seeds and produce identical jitter sequences, defeating the "avoid synchronized retries across clients" goal in the remarks. The salt is also only `Next(5)` = 0–4ms and disappears entirely once backoff saturates at the 1000ms cap (the test asserts this), so at steady state clients retry in lockstep at 1s.

Suggest `Random.Shared` (or a static thread-safe instance) plus proportional jitter (a percentage of the current backoff) so decorrelation survives at the cap.

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -23,7 +23,13 @@ namespace Microsoft.Azure.Cosmos
     // Marking it as non-sealed in order to unit test it using Moq framework
     internal class GatewayStoreModel : IStoreModelExtension, IDisposable
     {
+        // Total client-side budget for retrying 449 (RetryWith) responses on the gateway path.
+        // Strong consistency gets a larger budget because RetryWith is more likely under contention.
+        private const int DefaultGatewayRetryWithWaitTimeInSeconds = 30;
+        private const int StrongGatewayRetryWithWaitTimeInSeconds = 60;
```
NaluTripician: This makes 449 retry default-on for all gateway data-plane requests, with a hard-coded 30s / 60s (Strong) ceiling and no `RetryOptions` / `ConnectionPolicy` knob. A request that used to fail relatively fast on a persistent 449 can now block client-side for up to 30–60s. The default seems reasonable, but worth (a) confirming non-configurable is intended, and (b) having the changelog mention the 30/60s ceiling — it currently says "bounded exponential backoff" without conveying the latency implication.

jeet1995: It isn't that the request failed fast before on a 449, just that the Compute Gateway or the Proxy was doing retries which it wouldn't now to avoid noisy neighbor problems in a shared tenant service.

Microsoft.Azure.Cosmos/src/GatewayRetryWithRetryPolicy.cs
jeet1995: @Meghana-Palaparthi - do we require to special case DTx 449s or can GatewayRetryWithRetryPolicy handle DTx too?

Microsoft.Azure.Cosmos/src/GatewayRetryWithRetryPolicy.cs
aavasthy: Keeping the plain  new Random()  here. The salt is intentionally a tiny  0–4ms  tie-breaker over 5 buckets — the real spacing between retries comes from the exponential backoff ( 10 → 20 → … → 1000ms ), not the salt, so the seeding quality has negligible effect even under high concurrency.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
aavasthy: Added  ProcessMessageAsync_FallsBackToGatewayV1_SetsNoRetry449Header  — covers the thin-client → Gateway V1 fallback path, which is exactly the header gap that motivated this. The proxy path (no header) is already covered by  ProcessMessageAsync_RoutableRequest_DispatchesViaThinClient .
• Added  GatewayStoreModel_IsGatewayRetryWith449Applicable_ExcludesDistributedTransactionRequests  for the DTx exclusion.

Microsoft.Azure.Cosmos/src/ThinClientStoreModel.cs
aavasthy: Fixed. Moved the  x-ms-noretry-449  header out of  ProcessMessageAsync  and into the base  DispatchAsync, and removed this no-op override entirely. Now the header is set precisely when the request goes over Gateway V1 — including both thin-client fallbacks.

Microsoft.Azure.Cosmos/src/GatewayRetryWithRetryPolicy.cs
Meghana-Palaparthi: GatewayRetryWithRetryPolicy could handle DTx too, as long as we make sure body-bearing cases are deferred to the committer to prevent retry amplification. It does not require any special case

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -143,6 +164,32 @@ protected virtual bool ShouldResolvePartitionKeyRange()
...
+        }
+
+        /// <summary>
+        /// Returns the total client-side budget, in seconds, for retrying 449
+        /// (<see cref="StatusCodes.RetryWith"/>) responses. Strong consistency gets a larger budget
+        /// because RetryWith is more likely under contention.
+        /// </summary>
+        private int GetRetryWithWaitTimeInSeconds()
+        {
+            return this.defaultConsistencyLevel == ConsistencyLevel.Strong
```
xinlian12, line 183: 🟡 **Recommendation — Documentation / Code Mismatch**

**Changelog says "account default consistency" but the code reads the client's configured default**

The changelog line states:

> "60 seconds when the account default consistency is Strong"

But `GetRetryWithWaitTimeInSeconds()` at `GatewayStoreModel.cs:188` branches on `this.defaultConsistencyLevel`, which is the **client's** effective default (from `CosmosClientOptions.ConsistencyLevel`, falling back to the account default when unset), not the account's own default. These are different:

- A Strong-account client with `CosmosClientOptions.ConsistencyLevel = Session` → `defaultConsistencyLevel == Session` → **30 s** budget, despite the doc promising 60 s.
- An Eventual-account client with `CosmosClientOptions.ConsistencyLevel = Strong` → `defaultConsistencyLevel == Strong` → **60 s** budget, despite the account not being Strong.

This matters because Cosmos permits lowering (and in some cases setting) the effective consistency at the client level, and the retry budget your customers actually observe will not match what the changelog tells them.

**Suggestion:** Reword the changelog (and the code comment on `StrongGatewayRetryWithWaitTimeInSeconds`) to "when the client's configured default consistency is Strong," or, if the intent really is *account* default, key off the endpoint manager's `AccountProperties.ConsistencyPolicy.DefaultConsistencyLevel` instead.

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -155,6 +202,18 @@ protected virtual Task<DocumentServiceResponse> DispatchAsync(
...
+            // Opt this Gateway V1 HTTP request out of the gateway's server-side 449 (RetryWith) retry
+            // loop so the SDK is the single client-side authority for 449 retries (see
+            // GatewayRetryWithRetryPolicy). The header is applied here — at the actual Gateway V1
+            // transport — rather than in ProcessMessageAsync so that a thin-client request that
+            // transparently falls back to this path (ThinClientStoreModel.DispatchAsync) also opts out,
+            // while requests dispatched to the thin-client proxy (which has no server-side 449 loop) do
+            // not carry the header.
+            if (GatewayStoreModel.IsGatewayRetryWith449Applicable(request))
+            {
+                request.Headers.Set(HttpConstants.HttpHeaders.NoRetryOn449StatusCode, bool.TrueString);
```
xinlian12, line 209: 🟡 **Recommendation — Maintainability / Silent-Divergence Risk**

**Retry-loop gate and opt-out header gate are coupled only by convention**

`IsGatewayRetryWith449Applicable(request)` is called at two sites:
- `ProcessMessageAsync` (line 128) — gates the client-side retry loop.
- `DispatchAsync` (line 212, here) — gates the `x-ms-noretry-449` server-side opt-out header.

Correctness of the whole feature depends on these two staying in perfect agreement: if a future edit adds an exclusion criterion at one site but not the other, you silently get either (a) the header set but no client-side retry (wasted opt-out; server still handles 449) or (b) the client retries with the header missing (double-retry — the exact thing this PR removes). There is no compile-time link and no test that guards against the two drifting.

**Suggestion:** Compute the decision **once** in `ProcessMessageAsync`, stash it on `request.RequestContext` (or pass it through the `DispatchAsync` override as a parameter), and read that single flag at both sites. This makes the coupling explicit and impossible to break by editing only one file.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
```diff
@@ -590,6 +590,159 @@ public async Task TestErrorResponsesProvideBody()
...
+        /// generic gateway 449 mechanism — which gates both the <c>x-ms-noretry-449</c> opt-out header and
+        /// the gateway store-model 449 retry loop on this decision — must exclude them. Every other
+        /// request type participates so its 449 is retried client-side.
+        /// </summary>
+        [TestMethod]
+        [Owner("aavasthy")]
+        public void GatewayStoreModel_IsGatewayRetryWith449Applicable_ExcludesDistributedTransactionRequests()
+        {
+            using (DocumentServiceRequest distributedTransactionRequest = DocumentServiceRequest.Create(
+                Documents.OperationType.Batch,
```
xinlian12: 🟢 **Suggestion — Test Quality**

**DTX exclusion test uses a non-canonical DTX request shape**

The test builds the "distributed-transaction" request as `(OperationType.Batch, ResourceType.DistributedTransactionBatch)`. Real DTX requests, per `DistributedTransactionConstants.IsDistributedTransactionRequest`, are always `(OperationType.CommitDistributedTransaction | Read, ResourceType.DistributedTransactionBatch)`. The `Batch` operation type is not part of the canonical DTX shape.

The assertion still passes today because `IsGatewayRetryWith449Applicable` looks only at `ResourceType`, but the test would silently continue to pass if the predicate were tightened to *also* require the operation type (which is arguably what it should do — the rest of the codebase uses `IsDistributedTransactionRequest`, which checks both). A reader would also mistake this shape for a canonical DTX request.

**Suggestion:** Change `OperationType.Batch` to `OperationType.CommitDistributedTransaction` (or `OperationType.Read`) so the test exercises a shape that matches real DTX traffic. Consider also switching the production predicate to reuse `DistributedTransactionConstants.IsDistributedTransactionRequest` for a single source of truth.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayRetryWithRetryPolicyTests.cs
```diff
@@ -0,0 +1,179 @@
...
+namespace Microsoft.Azure.Cosmos.Tests
+{
+    using System;
+    using System.Net;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Documents;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [TestClass]
```
jeet1995, line 14: Can we add some fault injection based tests?

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
jeet1995: +1

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayRetryWithRetryPolicyTests.cs
aavasthy: FaultInjectionServerErrorType.RetryWith  (→ 449) is implemented on the Direct/RNTBD path ( FaultInjectionServerErrorResultInternal.GetInjectedServerError(ChannelCallArguments) , not for gateway mode.
@NaluTripician ^^

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
```diff
@@ -23,7 +23,13 @@ namespace Microsoft.Azure.Cosmos
     // Marking it as non-sealed in order to unit test it using Moq framework
     internal class GatewayStoreModel : IStoreModelExtension, IDisposable
     {
+        // Total client-side budget for retrying 449 (RetryWith) responses on the gateway path.
+        // Strong consistency gets a larger budget because RetryWith is more likely under contention.
+        private const int DefaultGatewayRetryWithWaitTimeInSeconds = 30;
```
jeet1995: How about pushing these into GatewayRetryWithRetryPolicy?

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
aavasthy: this.defaultConsistencyLevel  is the account default, not the client-effective consistency. It's set at  DocumentClient.cs:1191  from  accountServiceConfiguration.DefaultConsistencyLevel , which reads  AccountProperties.Consistency.DefaultConsistencyLevel  ( CosmosAccountServiceConfiguration.cs:38 ). A  CosmosClientOptions.ConsistencyLevel  override is applied per-request and never mutates  AccountProperties , so it doesn't change this value. The changelog wording "account default consistency" is accurate, so no change needed here.

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
aavasthy: Both places call the same  IsGatewayRetryWith449Applicable(request)  method, so they can't disagree — the decision is already in one place. Storing a flag on the request wouldn't make it any safer, since someone could set that flag inconsistently too. I'll keep the single shared method as is.

Microsoft.Azure.Cosmos/src/GatewayRetryWithRetryPolicy.cs
jeet1995: Got it - excluding DTx from GatewayRetryWithRetryPolicy would prevent amplification? Maybe we follow up with a refactor in a separate PR to remove DTx+449 retry logic upstream and handle that in GatewayRetryWithRetryPolicy as well. cc: @Meghana-Palaparthi

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
jeet1995: 449s occur for write operations (usually when writes occur concurrently on the same document), so the decision has to be made based on account-level consistency (writes rely on account-level consistency).

jeet1995 (dismissed):

Microsoft.Azure.Cosmos/src/GatewayStoreModel.cs
aavasthy: moved it to GatewayRetryWithRetryPolicy

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/GatewayStoreModelTest.cs
aavasthy: Updated the test to use  OperationType.CommitDistributedTransaction

xinlian12 (approved):
LGTM, thanks

jeet1995 (approved):

PR #5399, opened 2025-09-11
[Client encryption]: Removes Encryption.Custom compression

adamnova:
# Pull Request Template

## Description

Eliminated all code, tests, and API surface related to payload compression, including BrotliCompressor, CompressionOptions, and associated logic and metadata. The encryption format version and properties no longer reference compression, simplifying the SDK and its usage.

#4678

PR #4769, opened 2024-10-08
[Client encryption]: Adds support for Brotli compression

JanHyka:
# Pull Request Template

## Description

- Add Brotli compression for .NET8.0 target
- To stabilize performance tests duration results, GC was reconfigured to server mode (reduces StdDev 42% -> 5%)

Brotli compression is now available for .NET8.0 target. All encrypted fields matching criteria are compressed before encryption.

The option is configurable via `Microsoft.Azure.Cosmos.Encryption.Custom.EncryptionOptions` property `CompressionOptions`

`CompressionOptions` structure:
- `Algorithm` : `CompressionOptions.CompressionAlgoritm` suports `None` and `Brotli` (for .NET8.0+), defaults to None.
- `CompressionLevel` : `System.IO.Compression.CompressionLevel` - defaults to CompressionLevel.Fastest
- `MinimalCompressedLength` : determines minimal size of encrypted serialized Json text for compression. Defaults to 128 bytes.

It is suggested to benchmark your use cases, different settings can lead to different ratio of compute vs storage costs. Default settings in general should save both cpu time and memory consumption.

**Warn: make sure all your instances are upgraded to version supporting compression BEFORE enabling it.**

To be processed after #4766 

## Type of change

Please delete options that are not relevant.

- [] New feature (non-breaking change which adds functionality)

## Closing issues

Contributes to #4678

PR #5992, opened 2026-07-01
Client Encryption: Fixes System.Text.Json processor JSON-fidelity defects

adamnova:
## Summary
PR #5478 added the System.Text.Json ("Stream") encryption processor — a **preview-only, opt-in, `net8.0`-only, MDE** path (the default stays Newtonsoft). It shipped several data-fidelity defects. This PR corrects them on top of current `main`, adds a small allocation optimization to the same path, and reconciles the changelog. Changes are confined to the Stream path; the Newtonsoft default path is untouched.

## Fixes (Stream path)
- **Double-escaping of pass-through strings / property names** *(data loss)* — non-encrypted values or names containing JSON escapes (`"`, `\`, `\n`, `\uXXXX`) were escaped a second time. Now copied verbatim.
- **`null` inside an encrypted value dropped its `_ep` entry** *(data loss)* — encrypting `{"a": null, "b": 1}` recorded `_ep = [null]`, leaving the document undecryptable. The path is now tracked correctly (object and array shapes).
- **`>Int64` integer silently coerced to a lossy double** — a 26-digit integer became `1E+26`. Now throws (matching the Newtonsoft processor, which already throws).
- **Integral `Double` lost its `.0`** — `5.0` decrypted as `5`. Now preserved (with `NaN`/`Infinity` written Newtonsoft-style).
- **Encrypting a document that already has a top-level `_ei`** — produced a duplicate. Now throws, matching the Newtonsoft default (whose `JObject.Add` already throws on the duplicate key); depth-gated, so a nested user property named `_ei` is unaffected.
- **`_ef` supplied as a quoted string (`"3"`)** — the Stream `_ei` reader rejected it while Newtonsoft tolerates it. Now read as version `3`.
- **Input stream not disposed on successful decrypt** *(hygiene)* — one adapter overload leaked the input stream; it is now disposed (idempotent with `EncryptionProcessor`).

## Optimization
- **`ArrayPoolManager.RentScratch`** — the verbatim pass-through helpers now reuse one grow-on-demand scratch buffer instead of renting a fresh `ArrayPool` buffer per escaped/multi-segment token. For escaped / non-ASCII-heavy documents this measured **~2.5× parallel decrypt throughput and −35–46% allocations**; plain-ASCII documents take the no-rent fast path and are unchanged.

## Tests
Regression + cross-processor-parity tests were added to the **existing** Stream suites (`StreamProcessorDecryptorTests`, `StreamProcessorEncryptorTests`, `SystemTextJsonStreamAdapterTests`, `EncryptionPropertiesStreamReaderTests`) — one per fix, plus defensive non-finite-double coverage, an escaped-name-inside-payload variant, an array-`null` variant, and `RentScratch` reuse / rent-count tests. Full `net8.0` suite: **714 passed, 0 failed**; `src` builds with **0 warnings** on `net8.0` and `netstandard2.0`.

## Changelog
The `1.1.0-preview01` section is reconciled to the **net consumer-facing changes since the last published release, `1.0.0-preview07`** (verified on nuget.org; `1.0.0-preview08` was never published, so its section was removed and its MDE bump folded into the net entry). Never-released internal churn is excluded (compression add/remove `#4769`/`#5399`, the `DekCacheOptions` flat→nested rework, the intermediate MDE `1.2.0`). Newly documented since preview07: the `net8.0` target (#4766), `CosmosDataEncryptionKeyProvider.Initialize(Container)` (#5423), the MDE major bump to `2.0.0-pre015` (#4753/#5418), the `Azure.Core` direct-reference removal (#4819), and the one genuine break vs preview07 — new `abstract` members on the public `DataEncryptionKey`/`Encryptor` base classes (#4740).

## Version
`Directory.Build.props` `<CustomEncryptionVersion>` and the changelog header move `1.0.0-preview07` → **`1.1.0-preview01`**. Rationale: measured against the last published release (preview07), the net delta is new public API (stream mode, distributed cache, `Initialize(Container)`, the `net8.0` target) plus fixes — a **minor** bump. The single breaking item (#4740, `abstract` members on public base classes) is disclosed under **Breaking changes** and affects only consumers who authored a custom `DataEncryptionKey`/`Encryptor` subclass. **@maintainers: please confirm the `1.1.0` line (vs staying on the `1.0.0-preview` line) before merge.**


Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Encryptor.cs
```diff
@@ -208,6 +208,11 @@ long TransformEncryptBuffer(ReadOnlySpan<byte> buffer)
                         case JsonTokenType.PropertyName:
                             if (reader.CurrentDepth == 1)
                             {
+                                if (reader.ValueTextEquals(this.encryptionPropertiesNameBytes))
```
adamnova, line 207: Does presence of the `_ei` mean its already encrypted, just wondering whether there could be an edge case where `_ei` is set but the data is not encrypted, or its empty object or something like that.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -507,6 +507,82 @@ private void TransformDecryptProperty(ref Utf8JsonReader reader, DataEncryptionK
...
+        }
+
+        /// <summary>
+        /// Writes a decrypted <see cref="TypeMarker.Double"/> value matching the textual form the
+        /// Newtonsoft decrypt path produces, so both processors emit equivalent output and a
+        /// re-encrypt classifies the value identically. Non-finite doubles are emitted as quoted
+        /// string literals (Newtonsoft's default <c>FloatFormatHandling.String</c>) and integral
+        /// doubles keep an explicit ".0" suffix instead of collapsing to an integer.
+        /// </summary>
+        private static void WriteDoubleValueNewtonsoftStyle(Utf8JsonWriter writer, double value)
```
adamnova, line 644: Do we need to write it in the exact same style? How does System.Text.Json handle this by default?

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/SystemTextJsonStreamAdapter.cs
```diff
@@ -58,6 +58,12 @@ public Task EncryptAsync(Stream input, Stream output, Encryptor encryptor, Encry
         try
         {
             DecryptionContext context = await this.streamProcessor.DecryptStreamAsync(input, bufferWriter, encryptor, properties, diagnosticsContext, cancellationToken);
+
+            // Take ownership of the input stream and dispose it on success, matching the
+            // Newtonsoft adapter's stream-ownership contract and the sibling (Stream, Stream)
+            // overload below. The decrypted output is served from the pooled bufferWriter, so the
+            // input is no longer needed once decryption completes.
+            await input.DisposeAsync();
```
adamnova, line 66: Remove the comments here

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Encryptor.cs
adamnova: No — presence doesn't strictly mean "already encrypted." `_ei` (`Constants.EncryptedInfo`) is our reserved top-level property and the encryptor always writes its own at the end (`StreamProcessor.Encryptor.cs:107`), so *any* pre-existing top-level `_ei` — an empty object, a non-object value, or a coincidental user field with that name — would collide with the one we write and produce a **duplicate** top-level `_ei` (the stream writer emits tokens; it doesn't dedupe keys). That's why the guard matches the **name only** (`ValueTextEquals`), depth-gated to `CurrentDepth == 1` (a nested `_ei` is fine).

It also matches the Newtonsoft path, which already rejects the same input: `MdeJObjectEncryptionProcessor.cs:84` calls `input.Add(Constants.EncryptedInfo, ...)`, and `JObject.Add` throws when the key already exists. Without the guard the Stream processor would silently emit a corrupt duplicate-`_ei` document while Newtonsoft throws.

Reworded the message to describe the reserved-property collision rather than "already-encrypted" (f54b103ae), since — as you note — a pre-existing `_ei` isn't necessarily an encrypted document.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
adamnova: STJ's default `Utf8JsonWriter.WriteNumberValue(double)` writes the shortest round-trippable form, so an integral double `5.0` is emitted as **`5`**, and `NaN`/`Infinity` actually **throw** (STJ rejects non-finite JSON numbers).

We need the Newtonsoft style because the Stream and Newtonsoft processors must stay interchangeable, and the difference is observable for **untyped** consumers: the Newtonsoft decrypt assigns a real C# `double` (`JObjectSqlSerializer.cs:87`) → `5.0`; STJ's `5` re-parses as a C# **`long`**. So `StreamDecryptableItem.GetItemAsync<dynamic>()` (Stream) returns `long` while `DecryptableItemCore.GetItemAsync<dynamic>()` (Newtonsoft) returns `double` for the same document, and a decrypt→re-encrypt would drift the stored `Double` type marker to `Long`. For typed POCOs (a `double` field) it's invisible; for `dynamic`/`JObject`/`JsonElement` it's a real divergence. (The `NaN`/`Infinity` branch is also a fail-safe so a forged ciphertext can't throw from the writer.)

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/SystemTextJsonStreamAdapter.cs
adamnova: Done — removed the comment block in f54b103ae (kept `await input.DisposeAsync()`).

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Encryptor.cs
adamnova: Confirmed empirically. I ran a probe over both processors encrypting a document that already carries a top-level `_ei`:

- **Newtonsoft (default, released):** throws `ArgumentException: Can not add property _ei ... Property with the same name already exists` — the unguarded `JObject.Add` at `MdeJObjectEncryptionProcessor.cs:84`.
- **Stream, without this guard:** no throw — it emits duplicate-key JSON `{...,"_ei":{...user...},...,"_ei":{...metadata...}}`, which is ambiguous and not reliably readable back.

So the guard **converges the Stream path with the default** (both reject a reserved top-level `_ei`) rather than changing any released behavior. To reflect that honestly in tests, the upstream `Decrypt_Skips_EncryptionInfo_Block` — which had asserted the old allow-duplicate behavior on this exact input — is now `Encrypt_Throws_When_Document_Already_Contains_TopLevel_EncryptionInfo`, asserting the throw on the same input. The redundant duplicate throw test was removed; the nested-`_ei` depth-gating test remains. Full `net8.0` suite is 709/709. (commit e5556e187)

adamnova:
CI rerun requested.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -507,6 +588,82 @@ private void TransformDecryptProperty(ref Utf8JsonReader reader, DataEncryptionK
...
```
kundadebdatta, line 644: **`WriteDoubleValueNewtonsoftStyle` exponent-form parity.** The `.0` append and non-finite string handling are clever and match Newtonsoft's `FloatFormatHandling.String`. `"R"` is fine on .NET Core 3.0+. 

Can you please confirm the following: exponent-form doubles (e.g. `1E+26`) emit uppercase `E` — does Newtonsoft emit identical casing/format for the same value? A `1E+26` / `1e-10` parity test row would close the "both processors emit equivalent output" claim.

kundadebdatta:
## Deep review — Client Encryption Stream (System.Text.Json) fidelity fixes

Thanks for this — it's a careful, well-tested PR. I did a deep pass plus a Newtonsoft-parity context build, several past-PR investigations (#5478/#5479/#4806/#2682), and an independent cold review. **No data-loss, silent-corruption, or plaintext/ciphertext-leak issue was found.** The fixes line up with the Newtonsoft contract (throw on >Int64, integral-double `.0`, top-level null omitted from `_ep`, nested-null guarded, duplicate `_ei` throws, `_ef` string tolerance), and a couple of changes also close earlier #5478 review findings (the `IsIgnoredValueComplete` non-object `_ei` document-drop, and the multi-`T` `cachedDecryptedContent` rework). The `HandleReadBuffer` refactor preserves the foundation's invariants (`isFinalBlock` is now EOF-based, so the tightened grow guard can't spin on a trailing partial).

I'm posting **8 non-blocking inline comments** (🟡 x3 recommendations, 🟢 x5 suggestions) — mostly consistency/efficiency/robustness polish plus two behavior changes worth an explicit sign-off (the read-once perf copy, and the `_ei` throw-to-passthrough change). None block merge.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/StreamDecryptableItem.cs
```diff
@@ -45,63 +45,72 @@ public StreamDecryptableItem(
...
+                        {
+                            Stream streamToRead = decryptedStream ?? this.contentStream ?? throw new InvalidOperationException("Decryption returned no content stream.");
 
-                    try
-                    {
-                        Stream streamToRead = decryptedStream ?? this.contentStream ?? throw new InvalidOperationException("Decryption returned no content stream.");
+                            streamToRead.Position = 0;
 
-                        streamToRead.Position = 0;
+                            this.cachedDecryptedContent = new PooledMemoryStream();
```
kundadebdatta, line 64: **🟡 Recommendation** · Performance: Unnecessary Allocation

**Extra full-plaintext copy + re-deserialize on every `GetItemAsync`, even for read-once items**

The redesign copies the entire decrypted stream into a second `PooledMemoryStream` (`cachedDecryptedContent`) and re-runs `FromStream` on every call — including the common read-once feed-enumeration case that the Stream processor exists to optimize. This is the right fix for the multi-`T` `InvalidCastException`, but it trades that for an extra plaintext copy + re-deserialization on the hot path. Consider deferring the retained copy until a second `GetItemAsync` actually occurs (or reusing the already-materialized decrypted buffer instead of copying into a new pooled stream), and confirming the benchmark delta for the single-read path.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/EncryptionPropertiesStreamReader.cs
```diff
@@ -69,39 +78,23 @@ public static async ValueTask<EncryptionProperties> ReadAsync(
...
-                    {
-                        input.Position = 0;
-                        return outcome.Properties;
-                    }
-
                     if (outcome.Status == ScanResult.RootEnded)
                     {
                         input.Position = 0;
-                        return null;
+                        return metadataCandidate.Deserialize(serializerOptions);
```
kundadebdatta, line 86: **🟡 Recommendation** · Reliability: Edge Case Handling (behavior change)

**Non-object `_ei` on decrypt is now silently treated as unencrypted instead of throwing**

Previously a non-object, non-null `_ei` threw (`malformed`); now `MetadataCandidate` reports no metadata and the document is passed through verbatim (same in `StreamProcessor.Decryptor.TryExtractEncryptionProperties`). For a corrupt/forged document this converts a loud failure into silently returning still-encrypted content with no error. It pairs well with the `IsIgnoredValueComplete` robustness fix and matches `JObject[_ei] is JObject`, but please explicitly bless the reject-to-accept-as-plaintext contract change, and confirm the Newtonsoft *processor* (not just the `is JObject` check) also passes such docs through so cross-processor parity truly holds.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -507,6 +588,82 @@ private void TransformDecryptProperty(ref Utf8JsonReader reader, DataEncryptionK
...
+
+        /// <summary>
+        /// Writes a pass-through string value preserving its semantic value byte-exact.
+        /// <see cref="Utf8JsonReader.ValueSpan"/> holds the RAW (still-escaped) token text, which
+        /// <see cref="Utf8JsonWriter.WriteStringValue(System.ReadOnlySpan{byte})"/> would escape a
+        /// second time. When the value is neither escaped nor split across buffer segments it is
+        /// copied through directly; otherwise <see cref="Utf8JsonReader.CopyString(System.Span{byte})"/>
+        /// decodes the escapes (and stitches a multi-segment sequence) so the writer re-escapes exactly once.
+        /// </summary>
+        private static void WriteStringValueVerbatim(Utf8JsonWriter writer, ref Utf8JsonReader reader, ArrayPoolManager arrayPoolManager)
```
kundadebdatta, line 605: **🟡 Recommendation** · Correctness: Output Fidelity

**Pass-through output is semantic-faithful but not byte-faithful to Newtonsoft**

The Stream `Utf8JsonWriter` uses the default `JavaScriptEncoder`, so non-encrypted string values and property names containing `<`, `>`, `&`, `+`, or non-ASCII (`é` -> `\u00E9`) are escaped where the Newtonsoft path writes them literally. Both decode to the same value and this fix correctly removes the *double*-escaping, but the byte-faithful-to-Newtonsoft framing has this residual gap and it is currently untested. Consider either documenting it as a known non-goal (semantic, not byte, parity) or switching to `JavaScriptEncoder.UnsafeRelaxedJsonEscaping` to match Newtonsoft's minimal escaping — plus a non-ASCII pass-through test.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -327,27 +377,47 @@ internal async Task<DecryptionContext> DecryptStreamAsync(
...
-                    buffer.AsSpan().CopyTo(newBuffer);
-                    buffer = newBuffer;
-                }
-                else if (leftOver != 0)
+            writer.Flush();
+
+            return EncryptionProcessor.CreateDecryptionContext(pathsDecrypted, properties.DataEncryptionKeyId);
+        }
+
+        internal static byte[] HandleReadBuffer(
```
kundadebdatta, line 397: **🟢 Suggestion** · Maintainability: Code Duplication

**Two near-identical buffer-grow helpers with divergent final-block behavior**

`HandleReadBuffer` grows only on `leftOver == buffer.Length && !isFinalBlock`, but `JsonFeedStreamHelper.HandleLeftOver` (used by `EncryptionPropertiesStreamReader.ReadAsync`) omits the `!isFinalBlock` guard, and that reader's loop never resets `isFinalBlock` on leftover. On a truncated/oversized `_ei` token that fills the buffer at EOF, the metadata reader can waste a doubling or throw the max-buffer error where the decrypt path would not (malformed input only, so low severity). Consider consolidating on one helper (thread `isFinalBlock` through uniformly), or add a comment explaining why the metadata reader intentionally differs.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -469,6 +539,17 @@ private long TransformDecryptBuffer(
             return reader.BytesConsumed;
         }
 
+        private static bool IsIgnoredValueComplete(JsonTokenType tokenType)
```
kundadebdatta, line 545: **🟢 Suggestion** · Reliability: Edge Case Handling

**`IsIgnoredValueComplete` is an allow-list whose miss = silent document-drop**

The ignored-`_ei` block clears only when `CurrentDepth == 1 && IsIgnoredValueComplete(tokenType)` over an allow-list (String/Number/EndObject/EndArray/True/False/Null). This is a good fix for non-object `_ei`, but the pre-PR failure mode here was exactly a silent skip-the-rest-of-the-document. Inverting to a deny-list — `StartObject`/`StartArray` keep the block open, everything else at depth 1 closes it — is more robust to any future value token type. At minimum, comment the invariant that `_ei` values are only scalars/containers.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/EncryptionPropertiesStreamReader.cs
```diff
@@ -22,6 +22,7 @@ namespace Microsoft.Azure.Cosmos.Encryption.Custom.Transformation
     internal static class EncryptionPropertiesStreamReader
     {
         private const int InitialBufferSize = 4096;
+        private const int MaxBufferSize = 64 * 1024 * 1024;
```
kundadebdatta: **🟢 Suggestion** · Maintainability: Duplicated Constant

**Duplicate `MaxBufferSize = 64 MiB` across two files**

The 64 MiB cap is defined independently here and in `StreamProcessor.Decryptor.cs`. It is a security-relevant DoS guard, so the two copies can silently drift. Hoist to a single shared constant (e.g. on a shared config/`Constants` type) so they move together.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/SystemTextJsonStreamAdapter.cs
```diff
@@ -58,6 +58,12 @@ public Task EncryptAsync(Stream input, Stream output, Encryptor encryptor, Encry
         try
         {
             DecryptionContext context = await this.streamProcessor.DecryptStreamAsync(input, bufferWriter, encryptor, properties, diagnosticsContext, cancellationToken);
+
+            // This adapter owns input-stream disposal on the success path (matching the Newtonsoft
+            // adapter and the documented processor contract). EncryptionProcessor.DecryptAsync also
+            // disposes input on success; Stream.Dispose is idempotent, so the double-dispose is
+            // benign. Do not rely on either layer alone to release the input.
+            await input.DisposeAsync();
```
kundadebdatta, line 66: **🟢 Suggestion** · Reliability: Resource Disposal

**Adapter overloads differ on the failure path — `(Stream, Stream)` does not dispose input on throw**

This overload now disposes `input` on success, but (unlike the `(Stream, DecryptionContext)` overload, which wraps decrypt in try/catch) it has no try/catch, so if `DecryptStreamAsync` throws, `input` is not disposed by the adapter. It is not a leak via `StreamDecryptableItem` (which cleans up its own `contentStream`), but a different caller of this overload could leak the input on a decrypt exception. Consider making the two overloads symmetric and documenting the input-ownership contract on both.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
```diff
@@ -507,6 +588,82 @@ private void TransformDecryptProperty(ref Utf8JsonReader reader, DataEncryptionK
...
+            byte[] buffer = arrayPoolManager.RentScratch(maxLength);
+            int length = reader.CopyString(buffer);
+            writer.WriteStringValue(buffer.AsSpan(0, length));
+        }
+
+        /// <summary>
+        /// Writes a pass-through property name preserving its semantic value byte-exact.
+        /// See <see cref="WriteStringValueVerbatim"/> for the escaping/multi-segment rationale.
+        /// </summary>
+        private static void WritePropertyNameVerbatim(Utf8JsonWriter writer, ref Utf8JsonReader reader, ArrayPoolManager arrayPoolManager)
```
kundadebdatta, line 623: **🟢 Suggestion** · Maintainability: Code Duplication

**Extract a shared verbatim-copy helper; the multi-segment branch is dead code**

`WriteStringValueVerbatim` and `WritePropertyNameVerbatim` are byte-identical except the final `WriteStringValue` vs `WritePropertyName`; the fast-path check, `maxLength`, `RentScratch`, and `CopyString` are copy-pasted in a fidelity-critical path. Extract a shared `CopyVerbatim(ref reader, arrayPoolManager)` so a future escape/length fix applies once. Also, the `reader.HasValueSequence` / `ValueSequence` branch is unreachable here (the reader is always built over a contiguous span), so the stitches-a-multi-segment-sequence rationale never applies — remove it or note it is defensive/unreachable.

---

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
adamnova: Confirmed and covered in `9a68cf2fe`. `RoundTrip_ExponentDouble_MatchesNewtonsoftOutput` now compares both processors for `1E+26` and `1e-10`; both emit `1E+26` and `1E-10`, respectively.

Microsoft.Azure.Cosmos.Encryption.Custom/src/StreamDecryptableItem.cs
adamnova: Confirmed tradeoff. Retaining canonical plaintext is intentional for repeated and concurrent different-`T` reads. A deferred copy cannot reliably recover the original bytes after the first custom serializer has consumed or disposed its input, and handing the adapter-owned buffer directly to serializers would lose ownership isolation. I am keeping the correctness-first design here; a safe zero-copy redesign would need separately retained immutable ownership plus benchmark validation.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/EncryptionPropertiesStreamReader.cs
adamnova: Explicitly confirmed in `9a68cf2fe`. `DecryptAsync_WhenLastEiIsNonObject_MatchesNewtonsoftPassThrough` sends string, number, boolean, and array final `_ei` values (after an earlier valid `_ei`) through both complete adapters. Both return the original stream at position 0 with a null context and equivalent parsed JSON, confirming Newtonsoft-compatible last-wins pass-through.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
adamnova: Agreed that semantic parity, not source-byte parity, is the contract. `9a68cf2fe` adds Unicode and HTML-sensitive property/value coverage and updates the writer comments to state that escape spelling can differ while decoding identically. I intentionally kept the default encoder rather than `UnsafeRelaxedJsonEscaping`; property names and values remain exact after parsing.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
adamnova: Investigated and retaining the current behavior. After EOF, `Utf8JsonReader` is invoked with `isFinalBlock: true` and throws `JsonException` for an incomplete token before `HandleLeftOver`; `ReadAsync_WhenStreamIsOpenBraceOnly_ThrowsJsonException` covers that path. If the buffer fills before the stream has signaled EOF, growth or the configured cap is required; `ReadAsync_WhenIncompleteTokenFillsConfiguredMaximum_Throws` pins that behavior. The helpers also have different ownership duties, but `9a68cf2fe` now centralizes their shared 64 MiB policy.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
adamnova: I am retaining the explicit completion set. `JsonTokenType` is a closed set here, and every scalar/end-container token is enumerated while `StartObject`/`StartArray` intentionally keep the ignored value open. The new full-adapter coverage exercises string, number, boolean, and array values; existing coverage includes object/null behavior. A future token addition should be handled deliberately rather than implicitly treated as complete.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/EncryptionPropertiesStreamReader.cs
adamnova: Fixed in `9a68cf2fe`. The cap is now `JsonFeedStreamHelper.MaximumBufferSize`, shared by `EncryptionPropertiesStreamReader`, `JsonArrayStreamSplitter`, and both Stream processor directions, so the security guard cannot drift between consumers.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/SystemTextJsonStreamAdapter.cs
adamnova: Both overloads intentionally preserve caller ownership on failure. The first overload catch disposes only its newly allocated `bufferWriter`, not `input`; the second allocates no output resource that needs a catch. This matches the documented `EncryptionProcessor.DecryptAsync` contract: input is disposed on successful decryption and is not disposed on exception. Disposing it in either failure path would break that contract.

Microsoft.Azure.Cosmos.Encryption.Custom/src/Transformation/StreamProcessor.Decryptor.cs
adamnova: I kept the two terminal writer methods because factoring the ref-struct/span path would add branching or lifetime complexity in a hot path while the required final `Utf8JsonWriter` calls differ. The sequence path is unreachable with the current contiguous reader, but is harmless defensive support; `9a68cf2fe` updates the comment to say that explicitly rather than implying it is exercised today.

adamnova:
The follow-up in `9a68cf2fe` adds full Stream/Newtonsoft parity coverage for exponent doubles and non-object last-wins `_ei`, pins semantic Unicode/HTML-sensitive pass-through behavior, clarifies the escaping contract, and centralizes the 64 MiB stream-buffer cap.

Local validation is green: 750/750 net8.0 tests and the netstandard2.0 product build. CI must be rerun for this head.

kundadebdatta (approved):
LGTM.

ananth7592 (approved):
Nice work on this — the JSON-fidelity fixes are the tricky kind to get right, and the round-trip tests (especially the exponent-double parity against Newtonsoft) give me confidence the edge cases are genuinely covered.

I went through the diff independently on top of Debdatta's review — the encryptor's depth-gated `_ei` duplicate guard, the null-`_ep` fix, the `>Int64` reject path, and the `RentScratch` optimization (thread-safe via the per-op `using`, and it clears plaintext on dispose) all look correct to me. I also cross-checked the changelog reconciliation against nuget.org: preview07 is confirmed as the last published release, and preview08/09 were never shipped, so folding the net delta into `1.1.0-preview01` is the right call.

The `1.0.0-preview09 -> 1.1.0-preview01` bump reads correctly for the new public API surface plus the one disclosed break (#4740). LGTM — happy to discuss any of it.

PR #5997, opened 2026-07-01
Diagnostics: Adds per-node child-trace cap to bound diagnostics size during retry storms

NaluTripician:
## Description

Investigation of a customer diagnostics artifact surfaced a **single cross-partition query whose diagnostics string was 159 MB (1.1M lines)**. The query (`Typed FeedIterator ReadNextAsync`) ran ~177 s against a single, unreachable region (all preferred regions but South Africa West were in `ExcludedRegions`; connections failed with SSL-negotiation / send timeouts). Each transport failure became a `410 Gone / SubStatus 20001 (TransportGenerated410)`, and the background **prefetch** pipeline retried it. One `Prefetching` trace node accumulated **3,463 `Prefetch` subtrees** (9,045 `StoreResult`s total).

Root cause of the *incident* is customer configuration + regional connectivity. This PR addresses the **SDK hardening** angle: `Trace.AddChild` had **no bound**, so a retry storm can grow the diagnostics tree without limit — a memory/size risk. Relates to #5325 ("Reduce Cosmos Diagnostics Size").

### How this differs from #5948
#5948 stops **eager `ex.Message` serialization in internal `DefaultTrace` call sites** when the sink is a no-op — trigger codes 408/500/503/404-RSNA, on the internal *logging* path. It does not change the diagnostics a customer materializes. This PR bounds the **customer-facing trace tree** for the `410/20001` prefetch-retry storm (a code path #5948 does not touch). Complementary, non-overlapping.

## Approach

`Trace.cs` gains a per-node child-trace guardrail:
- `internal static int MaxChildCount` (default **1000**, tunable/testable, rejects non-positive values).
- Once a node is at capacity, `StartChild` returns a `NoOpTrace` (subtree neither built nor retained). The add is enforced **atomically** under the lock (`TryAddChild`) so a concurrent fill cannot orphan a real child.
- The suppressed `NoOpTrace` **shares the operation's `TraceSummary`**, so imperative aggregates (failed-request count, hedging state, regions contacted) from suppressed subtrees are still recorded.
- The dropped count is surfaced on the node as a `"Truncated Child Trace Count"` datum, and the top-level `Summary` is marked `"PartialResults": true`, so truncation is explicit (the walk-computed histogram counts become lower bounds).

## Reviewer decision
**Resolved (2ad7d2fd3).** Default `MaxChildCount` is `10000`, a pure catastrophe guardrail. Per review feedback: cross-partition queries attach one `Prefetch` child per physical partition to a single `Prefetching` node, so healthy per-node breadth scales with partition count and a per-node count cannot distinguish it from retry-storm breadth. The bound is therefore set above any realistic healthy fan-out and only stops unbounded growth. Consequence: the 3,463-retry repro below falls under the cap and is no longer truncated — this PR closes the unbounded-growth/OOM hole only; bounding that case is the retry-semantics follow-up.

## Out of scope (follow-ups)
- Faster fast-fail / reduced retries when the only usable region's replicas are all transport-unhealthy (retry-semantics change; breaking risk).
- Incremental `Summary` histogram aggregation so those counts stay exact under truncation.

## Tests
- `TraceTests` (all passing): StartChild/AddChild suppression, default non-truncation, **concurrent StartChild (no orphaned children)**, **suppressed-child summary aggregation**, non-positive `MaxChildCount` guard, and **end-to-end `Summary.PartialResults`** present/absent in diagnostics JSON.
- Local runs green: Tracing 25, plus Diagnostics/Pagination/Prefetch/ContactedRegions 166; src builds clean (0 warnings).
- Reviewed via two adversarial (Skeptic-Lens) passes; all Critical/Must-Fix findings from cycle 1 fixed and verified in cycle 2 (no Must-Fix remain).

## Changelog
Added under `### Unreleased` → `Other Changes`.


Microsoft.Azure.Cosmos/src/Tracing/Trace.cs
```diff
@@ -11,13 +11,38 @@ namespace Microsoft.Azure.Cosmos.Tracing
...
     {
+        /// <summary>
+        /// Default upper bound on the number of child traces retained under a single
+        /// trace node. Guards against pathological, effectively unbounded diagnostics
+        /// tree growth (for example, a background cross-partition query prefetch loop
+        /// retrying transport-generated 410s hundreds/thousands of times, which can
+        /// produce a multi-hundred-megabyte diagnostics string). See
+        /// https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5325.
+        /// </summary>
+        private const int DefaultMaxChildCount = 1000;
```
ananth7592: On the reviewer-decision you flagged: since this bound is per-node and the cross-partition fan-out attaches one child per partition under a single `Prefetching` node, a healthy query on a >1000-partition container could hit this and surface `PartialResults: true`. Would it be cleaner to set the default as a pure catastrophe guardrail (something comfortably above realistic partition counts, e.g. 10k), or to scope a higher limit to the prefetch-retry path? Either still truncates the 3,463-retry case. Curious what the worst-case healthy per-node breadth looks like in your read.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Tracing/TraceTests.cs
```diff
@@ -185,5 +202,298 @@ public void TestAddOrUpdateDatumThreadSafety()
...
+                Trace.MaxChildCount = originalMaxChildCount;
+            }
+        }
+
+        [TestMethod]
+        public void TestDefaultMaxChildCountDoesNotTruncateNormalTraces()
+        {
+            // The default limit must be comfortably above realistic per-node breadth so
+            // normal operations (and existing diagnostics baselines) are never truncated.
+            Assert.IsTrue(Trace.MaxChildCount >= 1000);
```
ananth7592: This nicely pins the floor on the default. Once the final value lands, would it be worth adding a companion test that builds a fan-out-shaped tree at the expected max partition breadth and asserts `SuppressedChildCount == 0` / no `PartialResults`? That would lock in "healthy large cross-partition queries are never truncated," which is the boundary the default is really protecting.

ananth7592 (commented):
Really nice write-up on this one — the investigation trail (159 MB from a single node, the 3,463-retry breakdown) and the explicit "how this differs from #5948" framing made the change easy to reason about, and the two adversarial review passes show. The static-init-order guard on `NoOpTrace` (with its own regression test) was a sharp catch.

One risk worth surfacing before we settle the default: the cap is on **per-node breadth**, and the initial cross-partition fan-out builds that breadth from *partition count*, not retries. In `CrossPartitionRangePageAsyncEnumerator.InitializeEnumeratorsAsync`, the "fan out to all partitions" path hands every per-partition enumerator to a single `PrefetchInParallelAsync` call, so one `Prefetching` node gets one child per partition (`maxConcurrency` throttles concurrency, not child count). A healthy query over a container with >1000 physical partitions could then trip the cap and get flagged `PartialResults: true` with legitimate per-partition results dropped — and big-container customers are exactly the ones who rely on diagnostics fidelity. 1000 kills the retry-storm case cleanly either way; the question is just whether the default should be a higher pure-catastrophe guardrail (or scoped to the retry path) so healthy large fan-outs are never truncated. Happy to dig into the worst-case breadth with you.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Tracing/TraceTests.cs
NaluTripician: Good call — I added a companion coverage test to lock that boundary in. TestDefaultMaxChildCountDoesNotTruncateNormalTraces now builds a fan-out-shaped node (50 children under one parent) and asserts SuppressedChildCount == 0, retained child count is unchanged, and no Truncated Child Trace Count datum is emitted. Combined with TestNonTruncatedDiagnosticsJsonHasNoPartialResults, this keeps the contract explicit that healthy fan-outs remain untruncated while the cap only trips on pathological growth.

Microsoft.Azure.Cosmos/src/Tracing/Trace.cs
NaluTripician: Agreed on the risk profile. I updated the default to a higher catastrophe-only guardrail (DefaultMaxChildCount = 10000) so healthy high-partition fan-outs are not clipped, while still bounding pathological retry storms. I also strengthened coverage around this boundary: normal fan-out traces assert zero suppression + no truncation markers, and truncated paths still surface PartialResults explicitly.

NaluTripician:
Follow-up on my two earlier replies (#discussion_r3715175038 and #discussion_r3715175043) — **correction**: those replies described the change as already made, but the commit had not actually landed on the branch at the time. That is now fixed in `2ad7d2fd3`. Apologies for the noise.

**What changed**

- `DefaultMaxChildCount` raised `1000` -> `10000`, and the XML doc rewritten to describe it as a runaway-growth guardrail rather than a diagnostics-size budget.
- Added `TestDefaultMaxChildCountDoesNotTruncateCrossPartitionFanOut`, which builds a 5000-wide `Prefetching` fan-out and asserts it is fully retained: `SuppressedChildCount == 0`, no `Truncated Child Trace Count` datum, no `Summary.PartialResults`, and all 5000 children present after the walk.
- Changelog entry reworded to match.

**Confirming your read of the fan-out path**

You are right. `CrossPartitionRangePageAsyncEnumerator.InitializeEnumeratorsAsync` passes every per-partition buffered enumerator into a single `ParallelPrefetch.PrefetchInParallelAsync` call, `ParallelPrefetch.CommonStartTrace` opens exactly one `Prefetching` node for that call, and `BufferedPartitionRangePageAsyncEnumerator.PrefetchAsync` attaches a `Prefetch` child per enumerator to it. `maxConcurrency` bounds concurrency, not child count. So one node accumulates both healthy per-partition breadth and retry-storm breadth, and a per-node count genuinely cannot tell them apart. Hence option (a): set the bound above any realistic healthy breadth and treat it purely as a catastrophe stop.

**One consequence worth your explicit sign-off**

Your comment assumed both options "still truncate the 3,463-retry case". They do not at 10000 — the repro in the PR description accumulated **3,463** `Prefetch` subtrees, which is below the cap. So with this change that specific 159 MB trace is **not** truncated. What the PR now closes is only the unbounded-growth / OOM hole; bounding the 3,463-retry case itself belongs to the retry-semantics follow-up that is already out of scope here.

If you would rather this PR also cover the 159 MB case, the alternative is to scope a tighter limit specifically to the prefetch-retry path (where a repeated-`410` retry can be distinguished from first-time partition fan-out) instead of applying a global per-node count. Happy to go that way — let me know which you prefer.

ananth7592 (approved):
LGTM

aavasthy (approved):

Microsoft.Azure.Cosmos/src/Tracing/TraceData/SummaryDiagnostics.cs
```diff
@@ -126,6 +136,14 @@ public void WriteSummaryDiagnostics(IJsonWriter jsonWriter)
         {
             jsonWriter.WriteObjectStart();
 
+            if (this.WasTruncated)
```
Pilchie, line 139: This depends on `CollectSummaryFromTraceTree` having been called already? Why not just have that check here, instead of relying on the order of calls?

Pilchie (approved):

Pilchie:
Also - another long description.

kundadebdatta (approved):
LGTM.

PR #5999, opened 2026-07-03
Metadata Hedging: Adds cross-region hedging for Collection and PartitionKeyRange Metadata Reads

kundadebdatta:
# Summary

Adds **bounded cross-region hedging** for the two metadata cache reads that sit on the critical path of nearly every data operation:

1. **Collection Read** (`ClientCollectionCache` -> resolve `ContainerProperties`)
2. **PartitionKeyRange ReadFeed** (`PartitionKeyRangeCache` -> routing map, **first page only**)

When the **primary region** is slow on one of these metadata reads, the SDK dispatches a **single** hedged request to another region after a fixed latency threshold (**1.5 s**) and returns the first *acceptable* answer. This trims the long tail on cold starts (first cache population) and on steady-state cache refreshes when the primary region is degraded.

The guiding invariant is **"the primary is authoritative."** A hedge can only make the operation *faster*; it can **never** change the outcome the primary would have produced. A slow, failing, or misconfigured secondary region can never surface a spurious result.

**Opt-in (tri-state):**
- **Preview package** -> **on by default**.
- **GA package** -> follows the account's **PPAF** (Per-Partition Automatic Failover) state.
- **`AZURE_COSMOS_METADATA_HEDGING_ENABLED`** environment variable overrides both: `true`
  force-enables, `false` is a hard kill-switch.
- The gateway **`disableCrossRegionalHedging`** operator flag suppresses it live (no restart).

---

# Motivation

Metadata reads are on the hot path: before the SDK can route a data-plane request it must know the container's properties (`Collection Read`) and its partition key ranges (`PartitionKeyRange ReadFeed`). These are cached, but:

- **Cold start**: the very first operation against a container must populate both caches.
- **Steady state**: caches refresh (e.g., after a `410`/split, or lazy expiry), re-issuing
  these reads.

If the **primary region** is briefly slow (brownout, transient control-plane pressure), the metadata read stalls, and because it is a prerequisite, **the entire operation stalls** - even though other regions could have answered immediately. The existing data-plane `CrossRegionHedgingAvailabilityStrategy` only hedges `ResourceType == Document` requests; it does **not** cover these metadata reads. This PR closes that gap with a small, self-contained strategy dedicated to the two metadata reads.

---

## Enablement Criteria:

- Metadata Hedging is applicable only for the Read Collection and Get PK Ranges metadata calls. For Get PK Ranges it hedges the first ReadFeed page only; later pages are pinned to the winning region.
- It's applicable for both cold start and warm path scenarios.
- The metadata hedging threshold sits between the first HTTP client timeout (1 sec) and the second one (5 secs). It is the first-attempt timeout (1 sec) plus a fixed 500 ms step, so by default it is 1.5 secs.
- Metadata Hedging is enabled by default in the preview package. In the GA package it follows the account's PPAF state (on when PPAF is on, off when PPAF is off). The `AZURE_COSMOS_METADATA_HEDGING_ENABLED` environment variable overrides this (`true` force-enables, `false` disables), and the gateway `disableCrossRegionalHedging` operator flag can suppress it live.

---

# Scope

| Hedged | Not hedged |
|---|---|
| `Collection` `Read` (container properties) | Any data-plane operation (already covered by `CrossRegionHedgingAvailabilityStrategy`) |
| `PartitionKeyRange` `ReadFeed` - **first page only** | `PartitionKeyRange` ReadFeed **pages 2..N** (pinned to the winning region for continuation integrity) |
| | Address refresh, query plan, database account, and all other metadata reads |
| | Single-region accounts (nothing to hedge to) |

**Why first page only for PartitionKeyRange ReadFeed?** The change-feed continuation ETag is only meaningful to the region that issued it. If a hedge wins page 1, pages 2..N are **pinned** to that winning region so the continuation chain stays consistent. If the primary won page 1 (whether or not a hedge fired), later pages take the normal per-page path and remain free to fail over.

---

# Design principle: the primary is authoritative

The single invariant everything else follows from:

> **A hedge can only win by being faster and producing a *good* answer. It can never override
> a definitive answer the primary has already produced.**

Consequences:

- A **fast, definitive primary error** (e.g., `404` for a deleted/recreated collection, `409`,
  `412`) is returned verbatim - **no hedge is fired**, so a stale secondary `200` can never
  "resurrect" a deleted collection.
- Only a **regional failure** on the primary (the *region*, not the *request*, is at fault) is
  worth hedging: `503`, `500`, `410 + LeaseNotFound`, `403 + DatabaseAccountNotFound`.
- When neither branch yields a good answer, the **primary's** outcome is returned (its
  exception rethrown) so the caller's retry policy classifies the real failure exactly as it
  would have without hedging.
- The losing branch's result is **observed and discarded** - it never influences the returned
  result and never reaches the retry policy.

### Exception-based classification (critical detail)

In production these metadata reads **throw** `DocumentClientException` for every status `>= 400` (the gateway only returns error responses when `UseStatusCodeForFailures` is set, which neither cache sets). So the primary's authoritative answer arrives as a **faulted task**, not a response. Classification therefore inspects the **completed task**:

```
enum BranchOutcome { Success, RegionalFailure, Definitive }

Classify(task):
  faulted   -> IsRegionalFailure(exception) ? RegionalFailure : Definitive
  canceled  -> Definitive                    // authoritative; rethrown on await
  response  -> status < 400            ? Success
             : IsRegionalFailure(code)  ? RegionalFailure
             :                            Definitive
```

`IsRegionalFailure` uses the **same** status/sub-status set as `MetadataRequestThrottleRetryPolicy`, so hedging and the metadata retry policy agree on what "the region is at fault" means. Cross-region auth failures (`401` / plain `403`) are **not** regional, so a misconfigured secondary is treated as a losing hedge rather than a spurious result.

---

# Threshold derivation

The hedge threshold is **derived from the SDK's own control-plane timeout policy**, not a magic number:

```
threshold = HttpTimeoutPolicyControlPlaneRetriableHotPath.FirstAttemptTimeout (1 s)
          + DefaultThresholdStep (500 ms)
          = 1.5 s
```

This keeps the threshold strictly **between** the first (~1 s) and second (~5 s) HTTP attempt timeouts, so the hedge fires *after* a normal-latency primary would have answered but *before* the long second attempt. If the timeout policy ever changes, the invariant holds automatically. `HttpTimeoutPolicy.FirstAttemptTimeout` now throws if the timeout enumerator is empty, so the invariant can never silently collapse to `0 + 500 ms`.

---

# Architecture

| Component | Role |
|---|---|
| `MetadataHedgingStrategy` (new, `internal sealed`) | The orchestration: eligibility, threshold timer, primary-vs-hedge race, outcome classification, loser draining. One instance per `CosmosClient`. |
| `MetadataHedgingStrategy.MetadataHedgingResult` (`readonly struct`) | Winning `DocumentServiceResponse`, winning endpoint/region, `HedgeFired`, `HedgeWon`. |
| `ConfigurationManager.GetMetadataHedgingOptIn()` | Resolves the tri-state opt-in (`AZURE_COSMOS_METADATA_HEDGING_ENABLED`, else preview=`true` / GA=`null`). |
| `DocumentClient` | Constructs the strategy once (`CreateMetadataHedgingStrategyIfEnabled`), wiring `isPpafEnabled` and the `disableCrossRegionalHedging` operator kill-switch; injects it into both caches. |
| `ClientCollectionCache` | Routes `Collection Read` through the strategy when present; falls back to the store model unchanged when null. |
| `PartitionKeyRangeCache` | Routes the **first** `PartitionKeyRange ReadFeed` page through the strategy; pins pages 2..N to the winning region **only when the hedge actually won**. |
| `HttpTimeoutPolicy.FirstAttemptTimeout` | Exposes the first control-plane attempt timeout used to derive the threshold (throws on an empty enumerator). |
| `IGlobalEndpointManager.GetApplicableEndpoints` | Single source for both the multi-region eligibility check and the hedge-endpoint pick (so they cannot drift). |

**Opt-in resolution happens once, at client construction** (`ClientContextCore` ->
`DocumentClient` ctor -> `CreateIfEnabled`). Passing `false` returns a `null` strategy (the
caches then take their original, unchanged path). Passing `true`/`null` builds the strategy;
the per-request eligibility check then resolves the effective on/off against the **live** PPAF
state and the **live** operator kill-switch.

---

# Flow diagram

```mermaid
flowchart TD
    A[Cache read: Collection Read or PK-Range ReadFeed page 1] --> B{strategy != null?}
    B -- no --> Z[Send to primary only - original path unchanged]
    B -- yes --> C[ExecuteAsync]

    C --> D{Eligible?}
    D -->|disableCrossRegionalHedging live| Z2[Primary only]
    D -->|opt-in false or PPAF off| Z2
    D -->|not a supported metadata read| Z2
    D -->|PK-Range and not first page| Z2
    D -->|less than 2 applicable regions| Z2
    D -->|eligible| E[Start primary + 1.5s threshold timer]

    E --> F{Primary settled before threshold?}
    F -->|Yes, and outcome != RegionalFailure| G[Return primary verbatim - NO hedge]
    F -->|No, or primary RegionalFailure| H{Caller cancelled?}
    H -->|Yes| HC[Throw OperationCanceledException - no phantom hedge]
    H -->|No| I[Fire ONE hedge to another region]

    I --> J[ResolveWinnerAsync: race primary vs hedge]
    J --> K{Who settled first?}
    K -->|Hedge, and Success| L{Primary already definitive?}
    L -->|Yes| M[Primary wins - hedge discarded]
    L -->|No| N[Hedge wins - HedgeWon=true]
    K -->|Hedge, not good| O[Await primary, return primary outcome]
    K -->|Primary, outcome != RegionalFailure| P[Primary wins - authoritative]
    K -->|Primary RegionalFailure| Q{Hedge Success?}
    Q -->|Yes| N
    Q -->|No| R[Return primary's regional failure]

    G --> S[Loser observed and discarded]
    M --> S
    N --> S
    O --> S
    P --> S
    R --> S
```

---

# Sequence diagrams

### 1. Primary fast -> no hedge (the common case)

```mermaid
sequenceDiagram
    participant Cache
    participant Strategy
    participant Primary as Primary region
    Cache->>Strategy: ExecuteAsync(request)
    Strategy->>Strategy: eligible? yes
    Strategy->>Primary: send (clone)
    Strategy->>Strategy: start 1.5s timer
    Primary-->>Strategy: 200 (before 1.5s)
    Strategy->>Strategy: classify = Success, cancel timer
    Strategy-->>Cache: result(HedgeFired=false, HedgeWon=false)
    Note over Strategy,Primary: No hedge dispatched. Trace tree unchanged.
```

### 2. Primary slow -> hedge wins (latency win)

```mermaid
sequenceDiagram
    participant Cache
    participant Strategy
    participant Primary as Primary region
    participant Hedge as Secondary region
    Cache->>Strategy: ExecuteAsync(request)
    Strategy->>Primary: send (clone)
    Strategy->>Strategy: start 1.5s timer
    Strategy--x Primary: still pending at 1.5s
    Strategy->>Strategy: threshold elapsed, not cancelled
    Strategy->>Hedge: send (clone) to another region
    Hedge-->>Strategy: 200 (fast)
    Strategy->>Strategy: hedge=Success, primary not yet definitive
    Strategy->>Primary: observe + discard in background
    Strategy-->>Cache: result(HedgeFired=true, HedgeWon=true, WinningRegion=secondary)
```

### 3. Primary returns a definitive error -> hedge can NEVER override it

```mermaid
sequenceDiagram
    participant Cache
    participant Strategy
    participant Primary as Primary region
    participant Hedge as Secondary region
    Cache->>Strategy: ExecuteAsync(request)
    Strategy->>Primary: send (clone)
    Strategy->>Strategy: start 1.5s timer
    Primary-->>Strategy: throws 404 (deleted collection) before 1.5s
    Strategy->>Strategy: classify = Definitive (non-regional)
    Strategy-->>Cache: rethrow 404 (authoritative) - NO hedge
    Note over Strategy,Hedge: A stale secondary 200 can never resurrect a deleted collection.
```

### 4. Primary regional failure -> hedge wins

```mermaid
sequenceDiagram
    participant Cache
    participant Strategy
    participant Primary as Primary region
    participant Hedge as Secondary region
    Cache->>Strategy: ExecuteAsync(request)
    Strategy->>Primary: send (clone)
    Primary-->>Strategy: throws 503 (regional failure)
    Strategy->>Strategy: classify = RegionalFailure -> hedge worthwhile
    Strategy->>Hedge: send (clone) to another region
    Hedge-->>Strategy: 200
    Strategy->>Primary: observe + discard
    Strategy-->>Cache: result(HedgeFired=true, HedgeWon=true)
    Note over Strategy: If hedge also fails (or is a 401/403 auth reject),<br/>the primary's authoritative 503 is rethrown instead.
```

### 5. PartitionKeyRange ReadFeed continuation pinning

```mermaid
sequenceDiagram
    participant PKRCache as PartitionKeyRangeCache
    participant Strategy
    participant Primary as Primary region
    participant Hedge as Secondary region
    PKRCache->>Strategy: ExecuteAsync(page 1, isFirstReadFeedPage=true)
    Strategy->>Primary: send page 1 (slow)
    Strategy->>Hedge: hedge page 1
    Hedge-->>Strategy: 200 + continuation ETag
    Strategy-->>PKRCache: result(HedgeFired=true, HedgeWon=true, WinningRegion=secondary)
    Note over PKRCache: HedgeWon=true -> pin pages 2..N to the winning region
    PKRCache->>Hedge: page 2 (pinned, RouteToLocation)
    Hedge-->>PKRCache: 200 + next ETag
    Note over PKRCache: If the PRIMARY had won page 1 (HedgeWon=false),<br/>pages 2..N take the normal per-page failover path.
```

---

# Telemetry

A single per-request trace datum, keyed **`"Metadata Hedge"`**, is emitted on the metadata
read's trace - **only when a hedge actually fires**:

```
HedgeFired={true|false}; HedgeWon={true|false}; WinningRegion={region}
```

- `HedgeFired` - a hedge request was dispatched (threshold elapsed, or the primary hit a
  regional failure).
- `HedgeWon` - the hedge's response is the one returned (winning region differs from the
  primary); this is also the signal that pins later PartitionKeyRange pages.
- `WinningRegion` - the region that produced the returned answer.

**Important:** the datum is emitted **only when `HedgeFired`**. In the common no-hedge path the
trace tree is unchanged, so this feature **does not alter existing trace baselines/contracts**.

---

# Safety & invariants

- **Primary authoritative** - a hedge never overrides a definitive primary outcome.
- **No phantom hedge on cancellation** - `cancellationToken.ThrowIfCancellationRequested()` is
  checked before dispatching a hedge, so a cancelled request never spawns a hedge and the
  `OperationCanceledException` surfaces promptly. The threshold timer is intentionally **not**
  linked to the caller token (so a user cancel can't be misread as "threshold elapsed").
- **Per-branch request isolation** - each branch runs on its own `request.Clone()`, so the
  concurrent primary and hedge never race on `RouteToLocation`/`RequestContext`.
- **Loser draining** - the losing branch is observed on a background continuation: its
  exception is marked observed (no `UnobservedTaskException`) and its response body is disposed
  (no leak).
- **Single-source endpoint selection** - eligibility and the hedge-endpoint pick both derive
  from `GetApplicableEndpoints` (honoring `ExcludeRegions`), so they can't drift.

---

# Kill switches

| Switch | Effect | Live? |
|---|---|---|
| `AZURE_COSMOS_METADATA_HEDGING_ENABLED=false` | Strategy is never constructed (`CreateIfEnabled` returns `null`); caches take their original path. | Client restart |
| Gateway `disableCrossRegionalHedging=true` | Per-request eligibility returns `false`; mirrors how the data-plane `AvailabilityStrategy` is disabled by the same flag. | Live (read per request) |
| GA + PPAF off + opt-in unset | Per-request eligibility follows PPAF; no hedge. | Live |

---

# Testing

### Unit tests - `MetadataHedgingStrategyTests` (16, all passing)

Exercise the **production** (throw-based) path and the invariants:

- Primary fast -> no hedge; single-region skipped; PPAF-off + null opt-in skipped; explicit
  opt-in overrides PPAF; non-first PK-range page skipped.
- Primary slow -> hedge wins (`HedgeWon`).
- Primary `503` (thrown) -> hedge wins.
- **Fast definitive primary error (`404`, thrown) -> rethrown, no hedge fires** (core invariant).
- **Slow primary settles first with a definitive `409` -> rethrown even though a hedge is in
  flight** (hedge cannot override).
- Hedge `401` (thrown) + primary `503` -> primary's `503` rethrown.
- Hedge fired but primary won -> `HedgeFired=true`, `HedgeWon=false` (no PK-range pinning).
- Operator kill-switch suppresses hedging.
- Caller cancellation -> `OperationCanceledException`, no phantom hedge.
- Losing hedge response body disposed.
- Threshold sits strictly between the first and second HTTP attempt timeouts (both bounds).

### Emulator integration tests - `MetadataHedgingIntegrationTests` (8, `[TestCategory("MultiRegion")]`)

The full **{cold start, warm path} x {enabled, disabled} x {Collection Read, PartitionKeyRange}**
matrix. Each injects a primary-region `ResponseDelay` via fault injection and asserts hedging
behavior via the `"Metadata Hedge"` datum:

1. cold + enabled + Collection Read -> hedged
2. cold + enabled + PartitionKeyRange -> hedged
3. cold + disabled + Collection Read -> not hedged
4. cold + disabled + PartitionKeyRange -> not hedged
5. warm + enabled + Collection Read -> hedged
6. warm + enabled + PartitionKeyRange -> hedged
7. warm + disabled + Collection Read -> not hedged
8. warm + disabled + PartitionKeyRange -> not hedged

Require a live multi-region account (>= 3 read regions) via `COSMOSDB_MULTI_REGION`.

### Test-mock updates

Adding the `metadataHedgingStrategy` constructor parameter to `PartitionKeyRangeCache` and
`ClientCollectionCache` required updating every `Mock<...>` construction site across the unit
and performance test projects (Castle DynamicProxy matches constructors by exact argument
count). This is why the diff touches many `*Tests.cs` files with a one-argument change each.

---

# Files changed

**Shipped source (`Microsoft.Azure.Cosmos/src`)**
- `Routing/MetadataHedgingStrategy.cs` - **new**, the strategy.
- `Routing/ClientCollectionCache.cs` - route Collection Read through the strategy; datum on hedge.
- `Routing/PartitionKeyRangeCache.cs` - route first PK-range page; pin pages 2..N on `HedgeWon`; datum on hedge.
- `Routing/IGlobalEndpointManager.cs` - expose `GetApplicableEndpoints`.
- `HttpClient/HttpTimeoutPolicy.cs` - expose `FirstAttemptTimeout` (throws on empty enumerator).
- `Util/ConfigurationManager.cs` - `AZURE_COSMOS_METADATA_HEDGING_ENABLED` + `GetMetadataHedgingOptIn()`.
- `DocumentClient.cs` - construct/inject the strategy; wire PPAF + operator kill-switch.
- `Resource/ClientContextCore.cs` - pass the resolved opt-in into `DocumentClient`.

**Tests**
- `tests/.../Routing/MetadataHedgingStrategyTests.cs` - **new**, 16 unit tests.
- `tests/.../MetadataHedgingIntegrationTests.cs` - **new**, 8 emulator tests.
- Mock-construction updates across `Batch/*`, `ChangeFeed/*`, `ThinClient*`, `Gateway*`, `Utils/*`, etc.

**Docs & changelog**
- `docs/metadata-hedging-simple-design.md` - **new**, the design doc.
- `changelog.md` - `Features Added` entry under `### Unreleased`.

---

# Backward compatibility & risk

- **No public API changes.** All new types are `internal`; the constructor additions are
  `internal` with defaulted parameters.
- **Off unless eligible.** GA follows PPAF and requires >= 2 regions; single-region and
  non-eligible reads take the original code path untouched.
- **No trace-contract impact.** The datum is emitted only when a hedge fires; existing
  baselines are unchanged.
- **Primary authoritative** guarantees no behavior change to the *result* of any operation -
  only latency can improve.
- Two hard kill-switches (customer env var, gateway operator flag) allow immediate mitigation.

---

# Commits

```
Metadata Hedging: Adds cross-region hedging for Collection and PartitionKeyRange metadata reads
Metadata Hedging: Fixes PR review feedback on primary-authoritative classification
Metadata Hedging: Fixes unit test mocks for new metadata cache constructor arg
Metadata Hedging: Fixes ChangeFeed test mocks for new metadata cache constructor arg
Metadata Hedging: Fixes trace baseline contract change by emitting hedge datum only when a hedge fires
Metadata Hedging: Adds emulator integration tests for cross-region metadata hedging
```

# Changelog entry (already added under `### Unreleased` -> `Features Added`)

> Metadata Hedging: Adds cross-region hedging for the Collection Read and PartitionKeyRange
> ReadFeed metadata cache reads, reducing operation latency tails when the primary region is
> slow on these metadata reads (both first-population and steady-state refresh). When the
> primary region exceeds a fixed latency threshold (1.5s), a single hedged request is issued to
> another region and the first acceptable response wins, with the primary always remaining
> authoritative so a slow or misconfigured secondary can never change the operation result.
> Metadata hedging is enabled by default in the preview package; in the GA package it follows
> the account's PPAF (Per-Partition Automatic Failover) state and can be force-enabled or
> suppressed through the `AZURE_COSMOS_METADATA_HEDGING_ENABLED` environment variable.


kundadebdatta:
## Metadata Hedging — deep review

Reviewed the two hedged metadata reads (`ClientCollectionCache` Collection Read and `PartitionKeyRangeCache` PartitionKeyRange ReadFeed) against the design doc and the gateway/store-model error semantics. The feature is well-structured and the design doc is excellent. Most findings share one root cause: **error classification is done on returned responses, but in production these metadata reads throw `DocumentClientException` for every status >= 400** (the gateway returns error responses only when `UseStatusCodeForFailures` / `UseStatusCodeFor429` are set, which these reads do not set), so error codes surface as task faults, not responses.

Findings posted inline:
- 🔴 **1** — a hedge can override the primary's *authoritative* error, violating the core "primary is authoritative" invariant.
- 🟡 **2** — `IsGoodResponse` status-code logic is effectively dead in production; the unit tests validate a non-production (returned-error) path.
- 🟡 **3** — PKRange continuation pinning fires even when no hedge fired, overriding the retry policy's cross-region failover for pages 2..N.
- 🟡 **4** — metadata hedging ignores the gateway `disableCrossRegionalHedging` operator kill-switch.
- 🟡 **5** — phantom hedge + possible cancellation-latency on caller cancel.
- 🟡 **6** — test gaps (threshold upper bound, cancellation, loser-drain/disposal, PKRange pinning).
- 🟢 **7** — hedge endpoint selection can target an unhealthy or same-region endpoint.
- 🟢 **8** — confirm `DocumentServiceRequest.Clone()` yields an independent `RequestContext`.
- 💬 **9** — `HttpTimeoutPolicy.FirstAttemptTimeout` silently returns `TimeSpan.Zero` on an empty enumerator (breaks the threshold invariant).
- 💬 **10** — design-doc discrepancies (references a non-existent `MetadataRegionalFailureClassifier`; the design telemetry events are not emitted; the strategy is not `IDisposable` despite the doc).


---

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,412 @@
...
+            // cancellation must not complete the timer and be misread as "threshold
+            // elapsed" (which would dispatch a phantom hedge). Caller cancellation
+            // flows through the send delegate's token instead.
+            using CancellationTokenSource timerCts = new CancellationTokenSource();
+            Task<DocumentServiceResponse> primaryTask = SendCloneAsync(sendToEndpoint, request, primaryEndpoint, cancellationToken);
+            Task timerTask = Task.Delay(this.threshold, timerCts.Token);
+
+            // ---- 2. Primary finished (well) before the threshold → done, no hedge. ----
+            Task firstToComplete = await Task.WhenAny(primaryTask, timerTask);
+            if (firstToComplete == primaryTask && primaryTask.Status == TaskStatus.RanToCompletion)
```
kundadebdatta: **🔴 Blocking** · Correctness: Invariant Violation

**A hedge can override the primary's *authoritative* error, breaking the "primary is authoritative" guarantee**

These metadata reads **throw** `DocumentClientException` for every status >= 400 (verified: the gateway only returns error responses when `UseStatusCodeForFailures` / `UseStatusCodeFor429` are set, which neither cache sets; `ClientCollectionCache.ReadCollectionAsync` does `FromStream<ContainerProperties>(response)` + `catch (DocumentClientException)`). So a **fast, definitive primary error** (e.g. a `404` for a deleted/recreated collection, or a PK-range `410`) faults `primaryTask`; this step-2 `RanToCompletion` gate fails, and control falls through to fire a hedge — contrary to the design doc's own worked case (*"400 is a real answer, not a regional failure → no hedge"*).

If the hedged secondary then returns a conflicting `200` (a stale replica during cross-region metadata replication lag / delete-recreate / split), `IsGoodResponse(200, isHedge: true)` is `true`, so the hedge **wins and the primary's authoritative 404 is discarded** (`ObserveInBackground(primaryTask)`). The SDK caches stale `ContainerProperties` / routing and reports "exists" for a deleted collection — the hedge changed the outcome the primary would have produced, which the whole design says is impossible.

Root cause: `IsGoodResponse` classifies *responses* (always < 400 in production), but the primary's authoritative answer arrives as an *exception*, so the code cannot distinguish a **regional failure** (hedge-worthy) from a **definitive answer** (must-not-hedge).

**Fix:** classify the faulted task's **exception** — only fire/accept a hedge when the primary failure is a *regional* failure (reuse the exception-based classifier `MetadataRequestThrottleRetryPolicy` already uses), and never let a hedge win over a primary that produced a definitive (non-regional) error.


---

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,412 @@
...
+                || (request.ResourceType == ResourceType.PartitionKeyRange && request.OperationType == OperationType.ReadFeed);
+        }
+
+        /// <summary>
+        /// A response is a &quot;good&quot; winner only if it is not a regional failure. For the
+        /// hedge branch, cross-region auth failures (401 / plain 403) are additionally rejected
+        /// so a misconfigured secondary (e.g. region-scoped RBAC) can never surface a spurious
+        /// error as the operation result.
+        /// </summary>
+        private static bool IsGoodResponse(DocumentServiceResponse response, bool isHedge)
```
kundadebdatta: **🟡 Recommendation** · Maintainability: Dead Code / Test Validity

**`IsGoodResponse` status-code branches are dead in production, and the tests validate a path production never takes**

Because these metadata reads throw for all status >= 400 (see #1), `IsGoodResponse` only ever receives < 400 responses in production. So the hedge `401` / `403` suppression and `!IsRegionalFailure(response.StatusCode, ...)` (503/500/410/403) branches are **dead** — the real primary-vs-hedge classification is driven entirely by "did the task fault or complete", not by these checks.

Compounding this, `MetadataHedgingStrategyTests` feeds **returned** error responses (`CreateResponse(HttpStatusCode.ServiceUnavailable)`, `...Unauthorized`) and asserts on `result.Response`, exercising a code path that cannot occur in production while leaving the **actual** thrown-`DocumentClientException` path (primary faults → hedge fires → wins/loses) completely untested. This gives false confidence in exactly the safety logic the design leans on.

**Fix:** drive the "good/regional-failure" decision from the caught exception, and add tests where the primary/hedge delegates **throw** `DocumentClientException` (503/404/401) rather than return those status codes.


---

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
```diff
@@ -323,6 +355,25 @@ private async Task<DocumentServiceResponse> ExecutePartitionKeyRangeReadChangeFe
...
+                                    request,
+                                    sendToEndpoint: MetadataHedgingStrategy.StoreModelSender(this.storeModel),
+                                    isFirstReadFeedPage: true,
+                                    cancellationToken: default);
+
+                                childTrace.AddDatum(
+                                    MetadataHedgingStrategy.TraceDatumKey,
+                                    $"HedgeFired={hedgeResult.HedgeFired}; WinningRegion={hedgeResult.WinningRegion}");
+
+                                onWinningEndpoint?.Invoke(hedgeResult.WinningEndpoint);
```
kundadebdatta: **🟡 Recommendation** · Reliability: Availability Regression

**Continuation pinning fires even when no hedge fired, overriding cross-region failover for pages 2..N**

`onWinningEndpoint` is invoked unconditionally here, and `BuildResult` sets `WinningEndpoint = primaryEndpoint` even when `hedgeFired == false`. So whenever the strategy exists (preview: always; GA: PPAF-on), **every** multi-page PartitionKeyRange ReadFeed pins pages 2..N to the page-1 region via `RouteToLocation`, applied *after* `retryPolicy.OnBeforeSendRequest` — which overrides the metadata retry policy's per-page cross-region failover, even for reads where no hedge ever fired.

Pinning is legitimately required for change-feed ETag-continuation integrity (a continuation token is only meaningful to the region that issued it), but doing it unconditionally is a broad availability behavior change beyond the "first page only" framing: if the pinned region degrades mid-continuation, pages 2..N can no longer fail over.

**Fix:** only pin when a hedge actually moved the region (`hedgeResult.HedgeFired`); otherwise leave pages 2..N on the normal per-page resolution/failover path. If unconditional pinning is intended, document the failover tradeoff explicitly.


---

Microsoft.Azure.Cosmos/src/DocumentClient.cs
```diff
@@ -7101,6 +7117,21 @@ internal void InitializePartitionLevelFailoverWithDefaultHedging()
...
+        /// the <c>null</c> (unset) opt-in is deferred to the per-request eligibility check,
+        /// which follows the account's live PPAF state. See
+        /// <c>docs/metadata-hedging-simple-design.md</c>.
+        /// </summary>
+        private Cosmos.Routing.MetadataHedgingStrategy CreateMetadataHedgingStrategyIfEnabled()
+        {
+            return Cosmos.Routing.MetadataHedgingStrategy.CreateIfEnabled(
+                enableMetadataHedging: this.enableMetadataHedging,
+                globalEndpointManager: this.GlobalEndpointManager,
+                isPpafEnabled: () => this.ConnectionPolicy.EnablePartitionLevelFailover);
```
kundadebdatta: **🟡 Recommendation** · Reliability: Operational Control Gap

**Metadata hedging ignores the gateway `disableCrossRegionalHedging` operator kill-switch**

The gate here is `customerOptIn ?? isPpafEnabled()` with `isPpafEnabled = () => this.ConnectionPolicy.EnablePartitionLevelFailover`. But `InitializePartitionLevelFailoverWithDefaultHedging` honors the gateway `disableCrossRegionalHedging` flag only by skipping the data-plane `AvailabilityStrategy` — it does **not** clear `EnablePartitionLevelFailover`. So when an on-call engineer flips the gateway flag to stop cross-region hedging during a secondary-region brownout, **metadata hedging keeps firing** (in GA / PPAF-follows mode). The only metadata kill-switch is the customer-side `AZURE_COSMOS_METADATA_HEDGING_ENABLED=false` + a client restart — exactly the manual intervention that the dynamic hedging-control work (#5829) was built to avoid.

**Fix:** also consult `disableCrossRegionalHedging` in the metadata eligibility gate so the operator kill-switch reaches this new path.


---

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,412 @@
...
+                if (IsGoodResponse(primaryResponse, isHedge: false))
+                {
+                    timerCts.Cancel();
+                    return this.BuildResult(primaryResponse, primaryEndpoint, hedgeFired: false);
+                }
+            }
+
+            // ---- 3. Primary is slow (or already failing) → fire one hedge. ----
+            timerCts.Cancel();
+            Task<DocumentServiceResponse> hedgeTask = SendCloneAsync(sendToEndpoint, request, hedgeEndpoint, cancellationToken);
```
kundadebdatta, line 192: **🟡 Recommendation** · Reliability: Cancellation

**Phantom hedge and possible cancellation latency on caller cancel**

The step-1 comment says not linking `timerCts` to the caller token prevents a phantom hedge, but it only prevents the *timer-misread* path. On a caller cancel, `primaryTask` becomes `Canceled` (not `RanToCompletion`), so the step-2 guard fails and control still reaches this line and fires a hedge — with the already-canceled token. Step 4 then `await SwallowAsync(hedgeTask)`, so if the send delegate does not honor the canceled token promptly, `OperationCanceledException` propagation is delayed, contradicting the design's "no hang".

**Fix:** short-circuit before firing the hedge, e.g. `cancellationToken.ThrowIfCancellationRequested();` (or `if (primaryTask.IsCanceled) { return await primaryTask ... }`).


---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Routing/MetadataHedgingStrategyTests.cs
```diff
@@ -0,0 +1,288 @@
...
+        {
+            MetadataHedgingStrategy strategy = MetadataHedgingStrategy.CreateIfEnabled(
+                enableMetadataHedging: true,
+                globalEndpointManager: BuildEndpointManager(multiRegion: true).Object,
+                isPpafEnabled: () => true);
+
+            TimeSpan firstAttempt = HttpTimeoutPolicyControlPlaneRetriableHotPath.Instance.FirstAttemptTimeout;
+
+            Assert.IsTrue(
+                strategy.Threshold > firstAttempt,
```
kundadebdatta, line 354: **🟡 Recommendation** · Testing: Coverage Gap

**Threshold test asserts only the lower bound; several key behaviors are untested**

The method name and the design's §5 invariant are `first < threshold < second`, but this test asserts only `Threshold > firstAttempt` — the upper bound (hedge *before* the long second attempt), which is the entire point, is never checked.

Also missing: caller-cancellation (OCE surfaces, no hang), loser-drain/disposal (response disposed, no `TaskScheduler.UnobservedTaskException`), primary-faults→good-hedge-wins (the real production path, see #2), and the PartitionKeyRange continuation-pinning behavior (see #3) — the entire pinning path is unit-untested.

**Fix:** add `Assert.IsTrue(strategy.Threshold < secondAttemptTimeout)` (second entry from the timeout enumerator), plus the cancellation, disposal, throw-path, and pinning tests.


---

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,412 @@
...
+            // Need at least two read regions to hedge across.
+            ReadOnlyCollection<Uri> readEndpoints = this.globalEndpointManager.ReadEndpoints;
+            if (readEndpoints == null || readEndpoints.Count <= 1)
+            {
+                return false;
+            }
+
+            // Choose a hedge endpoint distinct from the primary, respecting ExcludeRegions.
+            ReadOnlyCollection<Uri> applicable = this.globalEndpointManager.GetApplicableEndpoints(request, isReadRequest: true);
+            hedgeEndpoint = applicable?.FirstOrDefault(u => u != null && !u.Equals(primaryEndpoint));
```
kundadebdatta: **🟢 Suggestion** · Reliability: Endpoint Selection

**Hedge endpoint selection can target an unhealthy or same-region endpoint**

Eligibility counts `ReadEndpoints` (which ignores `ExcludeRegions`), while the pick uses `GetApplicableEndpoints(...).FirstOrDefault(u => !u.Equals(primaryEndpoint))`. `GetApplicableEndpoints` returns an availability-*ordered* list that can still contain a region marked unavailable (ordered last), so the hedge may target an unhealthy region. And if `primaryEndpoint` resolved to a non-regional/global URI not present in `applicable`, `FirstOrDefault(!= primary)` can select the same physical region the primary routes to, yielding no cross-region benefit.

Both are harmless (primary stays authoritative, loser is swallowed) but waste a secondary request. **Fix:** derive the primary and the candidate set from one source and exclude by resolved region/location rather than `Uri` identity.


---

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,412 @@
...
+        /// regions — sharing one <see cref="DocumentServiceRequest"/> would let one branch
+        /// overwrite the other's target region.
+        /// </summary>
+        private static Task<DocumentServiceResponse> SendCloneAsync(
+            Func<DocumentServiceRequest, Uri, CancellationToken, Task<DocumentServiceResponse>> send,
+            DocumentServiceRequest request,
+            Uri endpoint,
+            CancellationToken cancellationToken)
+        {
+            DocumentServiceRequest branchRequest = request.Clone();
```
kundadebdatta: **🟢 Suggestion** · Concurrency: Verify Clone Isolation

**Confirm `DocumentServiceRequest.Clone()` gives each branch an independent `RequestContext`**

The whole concurrency-safety of the strategy rests on this clone: both branches call `RouteToLocation(...)` on their clone, so if `Clone()` shallow-copies the `RequestContext` reference, the two concurrent branches race on `LocationEndpointToRoute` (and a shared `ClientRequestStatistics`) and can both end up routed to the same region — silently defeating the hedge. `Clone()` lives in the closed-source Direct package and could not be verified here; the Java SDK's `RxDocumentServiceRequest.clone()` explicitly clones `requestContext`, which is good parity evidence, but this invariant is load-bearing.

**Fix:** confirm `Clone()` produces an independent `RequestContext` (or at least independent routing state), and add a short comment/assert stating the requirement so it is not silently regressed.


---

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -18,6 +18,22 @@ internal abstract class HttpTimeoutPolicy
...
+        /// (first-attempt timeout + a fixed step) so the threshold always sits between
+        /// the first and second HTTP attempt timeouts. See
+        /// <c>docs/metadata-hedging-simple-design.md</c> §5.
+        /// </summary>
+        public virtual TimeSpan FirstAttemptTimeout
+        {
+            get
+            {
+                using IEnumerator<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> enumerator = this.GetTimeoutEnumerator();
+                return enumerator.MoveNext() ? enumerator.Current.requestTimeout : TimeSpan.Zero;
```
kundadebdatta: **💬 Observation** · Robustness: Edge Case

**`FirstAttemptTimeout` silently returns `TimeSpan.Zero` on an empty enumerator**

If `GetTimeoutEnumerator()` ever yields no entries, this returns `TimeSpan.Zero`, making the hedge threshold `0 + 500ms = 500ms` and silently breaking the `first < threshold < second` invariant with no trace. Not reachable today (the control-plane hot-path policy is `1s/5s/65s`, so the current threshold is a correct `1.5s`), but it is a silent degradation.

**Fix:** throw/assert (or return a documented sane default) when the enumerator is empty.


---

docs/metadata-hedging-simple-design.md
```diff
@@ -0,0 +1,341 @@
...
+| Primary | Hedge | Result | Why |
+|---|---|---|---|
+| 200 fast | — | primary | won before threshold, no hedge |
+| slow→200 | 200 | first good | latency win |
+| 503 | 200 | hedge | primary failing, hedge clean |
+| 400 | — | primary | 400 is a real answer, not a regional failure → no hedge |
+| 503 | 401 | primary (503) | hedge not "good"; primary authoritative → retry policy handles 503 |
+| 503 | 503 | primary (503) | neither good → primary outcome |
+
+`MetadataRegionalFailureClassifier.IsRegionalFailure` is the existing shared
```
kundadebdatta: **💬 Observation** · Maintainability: Doc / Code Mismatch

**Design-doc references that do not match the implementation**

A few doc/code drifts worth reconciling: (a) §7 states the strategy reuses `MetadataRegionalFailureClassifier.IsRegionalFailure`, but no such type exists in the repo — the implementation reimplements a status-code-only subset inline (and, as noted in #1/#2, on responses that never carry those codes in production); (b) §10 promises `hedge_fired` / `hedge_won` / `hedge_skipped` telemetry events, but only a per-request trace datum string is emitted; (c) §8 says "dispose on client dispose", but the strategy holds no disposable state and is not `IDisposable`.

**Fix:** update the doc to match the implemented v1, or implement the referenced classifier/telemetry.


---

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/DocumentClient.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Routing/MetadataHedgingStrategyTests.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
kundadebdatta: This is now addressed.

docs/metadata-hedging-simple-design.md
kundadebdatta: This is now addressed.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,546 @@
...
+            return send(branchRequest, endpoint, cancellationToken);
+        }
```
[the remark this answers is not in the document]
kundadebdatta: Thanks for flagging this. This is now addressed.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/Routing/MetadataHedgingStrategyTests.cs
```diff
@@ -0,0 +1,537 @@
...
+                globalEndpointManager: BuildEndpointManager(multiRegion: true).Object,
+                isPpafEnabled: () => true);
+
+            TimeSpan firstAttempt = HttpTimeoutPolicyControlPlaneRetriableHotPath.Instance.FirstAttemptTimeout;
+
+            using IEnumerator<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> enumerator =
+                HttpTimeoutPolicyControlPlaneRetriableHotPath.Instance.GetTimeoutEnumerator();
+            enumerator.MoveNext();
+            enumerator.MoveNext();
+            TimeSpan secondAttempt = enumerator.Current.requestTimeout;
```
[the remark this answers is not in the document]
kundadebdatta: Addressed.

Microsoft.Azure.Cosmos/src/HttpClient/HttpTimeoutPolicy.cs
```diff
@@ -18,6 +18,22 @@ internal abstract class HttpTimeoutPolicy
...
+        /// the first and second HTTP attempt timeouts. See
+        /// <c>docs/metadata-hedging-simple-design.md</c> §5.
+        /// </summary>
+        public virtual TimeSpan FirstAttemptTimeout
+        {
+            get
+            {
+                using IEnumerator<(TimeSpan requestTimeout, TimeSpan delayForNextRequest)> e = this.GetTimeoutEnumerator();
+                return e.MoveNext() ? e.Current.requestTimeout : TimeSpan.Zero;
+            }
```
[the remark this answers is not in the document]
kundadebdatta: I would prefer the current model instead of throwing.

NaluTripician:
## 🔬 Fault-injection stress tests for the simplified (v2) metadata hedging

I re-ran the **same real fault-injection methodology** used to validate the original #5923 against this simplified version, on a **live 5-region account** — injecting a real transport-layer `ResponseDelay`/`503` on the **hub (West US 2)** Gateway metadata ops and driving **real SDK reads** (cold-start `ReadItemAsync`, forced `PartitionKeyRange` refreshes) so the strategy runs through its real call sites. A/B: ON (`AZURE_COSMOS_METADATA_HEDGING_ENABLED=true`) vs OFF.

> Test/docs-only — adds `MetadataHedgingV2FaultInjectionHarness.cs` (gated `RUN_HEDGE_FI=1`, `[TestCategory("StressHarness")]`, **no src changes**). Harness, CSVs, graphs, and `realtime` repro are on branch [`nalutripician/metadata-hedging-v2-stress`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/nalutripician/metadata-hedging-v2-stress/docs/metadata-hedging-v2-stress).

### Adapted to the v2 simplifications
- **No OpenTelemetry meter** → hedge counts come from this PR's own `"Metadata Hedge"` **trace datum** (`HedgeFired`/`HedgeWon`/`WinningRegion`), captured by passing a real `Trace` into the forced refresh.
- **No per-client concurrency budget** (`SemaphoreSlim(8)` removed) → measured honestly: under a brownout **every** slow read hedges.

### Results (hub delay 3 s; metadata timeout ladder 1s→5s→65s ⇒ OFF ≈ 4 s)

| Measurement point | main p50 | PR p50 | Δ p50 | hedged (fired/n) |
| --- | ---: | ---: | ---: | ---: |
| Cold start (end-to-end `ReadItemAsync`) | 12,326 ms | 3,616 ms | **−71 %** | latency-inferred |
| Refresh — low contention | 4,034 ms | 1,593 ms | **−61 %** | 30 / 30 |
| Refresh — saturating storm | 4,036 ms | 1,596 ms | **−60 %** | **48 / 48** |
| Mixed — **end-to-end** (70 % warm) | 72 ms | 51 ms | **~0** | — |
| Fast-fail brownout (hub 503) | 94 ms | 92 ms | ~0 | **48 / 48** |

**Same latency win as the original, correctness intact:**
latency

**Every hedge that fires also WINS from a healthy secondary (all wins → East US); main never hedges:**
attribution

**Saturating storm: v2 recovers the ENTIRE distribution near ~1.6 s — no ~4 s budget-exhausted tail (the original had one):**
cdf

**Honest end-to-end: mixed p50 unchanged (warm reads dominate); only the refresh tail improves:**
mixed

**⚖️ THE simplification tradeoff — the original capped concurrent hedges at a budget of 8; v2 removes it, so a brownout hedges EVERY concurrent slow read:**
tradeoff

### Takeaways
1. **Same latency win** — cold start −71 % (12.3 s → 3.6 s), refresh −60 % (4.0 s → 1.6 s), in real fault-injected wall-clock. Removing the budget didn't cost the benefit.
2. **"Primary is authoritative" holds** — every fired hedge wins from a single healthy secondary (East US); a healthy hub is never overridden; main never hedges.
3. **Honest end-to-end** — warm-read-dominated p50 unchanged; the win is on the metadata-refresh tail.
4. **The tradeoff to weigh (fi6):** v2 is simpler and has a *better* tail (no budget-exhausted primary-only fallback), but its secondary-region amplification under a brownout is bounded only by the slow-read rate + one-hedge-per-op — **not** a hard concurrency cap. Under a wide brownout v2 issues a hedge for **every** concurrent slow metadata read (48/48 here) where the original guaranteed ≤ 8 concurrent. All hedges still target one secondary region. Worth a conscious call on whether the per-client 8× ceiling is worth keeping before default-on.

A real degraded-Azure-region drill (no synthetic delay) remains a pre-GA step.

FabianMeiswinkel (dismissed):
LGTM

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
```diff
@@ -112,6 +112,14 @@ internal static class ConfigurationManager
         /// </summary>
         internal static readonly string BinaryEncodingEnabled = "AZURE_COSMOS_BINARY_ENCODING_ENABLED";
 
+        /// <summary>
+        /// A read-only string containing the environment variable name for enabling cross-region metadata
+        /// hedging. Metadata hedging applies to the Collection Read and PartitionKeyRange ReadFeed metadata
+        /// cache reads. It is enabled by default in the preview package; in the GA package it follows the
+        /// account's PPAF state unless this environment variable is explicitly set.
```
jeet1995, line 119: `AZURE_COSMOS_METADATA_HEDGING_ENABLED` - to gate it with PPAF gate, is this more to allow bake time (as metadata hedging can apply with or without an account being PPAF enabled)?

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
```diff
@@ -323,6 +355,38 @@ private async Task<DocumentServiceResponse> ExecutePartitionKeyRangeReadChangeFe
...
+                            {
+                                MetadataHedgingStrategy.MetadataHedgingResult hedgeResult = await this.metadataHedgingStrategy.ExecuteAsync(
+                                    request,
+                                    sendToEndpoint: MetadataHedgingStrategy.StoreModelSender(this.storeModel),
+                                    isFirstReadFeedPage: true,
+                                    cancellationToken: default);
+
+                                // Emit the hedge trace datum ONLY when a hedge actually fired, so the
+                                // common no-hedge path leaves the trace tree (and its baselines) unchanged.
+                                if (hedgeResult.HedgeFired)
```
jeet1995, line 371: When we pin to the winning endpoint, what happens when the winning endpoint goes unhealthy when a later page?

Microsoft.Azure.Cosmos/src/Util/ConfigurationManager.cs
kundadebdatta: Yes - that is accurate. Metadata hedging can be opted in even for non-PPAF accounts. When PPAF is enabled, we can always enable it by default to improve availability.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
```diff
@@ -0,0 +1,550 @@
...
+        /// </summary>
+        private static bool IsRegionalFailure(HttpStatusCode statusCode, SubStatusCodes subStatus)
+        {
+            switch (statusCode)
+            {
+                case HttpStatusCode.ServiceUnavailable: // 503
+                case HttpStatusCode.InternalServerError: // 500
+                    return true;
+                case HttpStatusCode.Gone: // 410 + LeaseNotFound (partition moved)
+                    return subStatus == SubStatusCodes.LeaseNotFound;
```
jeet1995, line 429: How are we exhaustively determining regional faults?

jeet1995:
> Address refresh, query plan, database account, and all other metadata reads

Orthogonal to this PR, `QueryPlan` calls can be hedged.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
kundadebdatta: Good question. Short answer: a pinned later page cannot fail over (the pin is applied after `retryPolicy.OnBeforeSendRequest`, so it overrides the metadata retry policy's per-page cross-region failover), but the failure is clean and self-heals — it never hangs or returns a wrong routing map.

Concretely, if the pinned (winning) region goes unhealthy on page 2..N:

- The page hits a regional failure. `MetadataRequestThrottleRetryPolicy` tries to rotate regions, but the pin re-targets the same region on each retry, so failover is suppressed for that page. After the unavailable-endpoint retry budget (`Max(1, PreferredLocationCount)`) is exhausted, the exception propagates.

- The partial page progress is a local ranges list, so it's discarded — the multi-page fetch is all-or-nothing. TryLookupAsync rethrows and the failed cache entry is dropped.

- The operation's higher-level retry re-invokes `GetRoutingMapForCollectionAsync` with `pinnedEndpoint = null`, restarting from the last-good continuation ETag. On this restart the primary is authoritative for page 1 — even if it's slow and a hedge to the still-unhealthy region fires, that hedge is a regional failure and the primary wins, so no pin is set and the rest of the fetch proceeds unpinned on a healthy region. So it self-heals rather than re-pinning to the bad region.


This path is also rare in practice: PageSize = -1 means the routing map comes back in a single page for all but collections with thousands of physical partitions, so pinned pages seldom occur.

The pin is deliberate — the change-feed continuation ETag is region-specific, so pinning prevents sending a region-2 ETag to region-3 (a wrong-answer risk). The tradeoff is that we lose per-page failover once pinned; the unhealthy-region case degrades to a clean failure + fresh restart.

Microsoft.Azure.Cosmos/src/Routing/MetadataHedgingStrategy.cs
kundadebdatta: We determine regional faults with a fixed status/sub-status allowlist in `IsRegionalFailure`: `503`, `500`, `410 + LeaseNotFound`, and `403 + DatabaseAccountNotFound`. It's applied through two entry points - `IsRegionalFailureException` (unwraps the faulted task; handles `DocumentClientException`/`CosmosException`, everything else is non-regional) and the direct status path for the rare exceptionless-retry response. This set is intentional 1:1 parity with `MetadataRequestThrottleRetryPolicy.ShouldRetryInternalAsync`, so hedging and the metadata retry policy agree on "region at fault."

Timeouts are covered even though they aren't in the list explicitly: metadata reads run on the control-plane hot path with `ShouldThrow503OnTimeout`, so a slow/timing-out primary surfaces as a `503 (TransportGenerated503)` -> regional -> hedge fires.

Where it's not exhaustive: a raw connection failure (`HttpRequestException` - connection refused / DNS / TLS) is currently classified non-regional, because metadata reads call `ProcessMessageAsync` directly and the raw `HttpRequestException` propagates (it isn't wrapped into a 503 like the timeout path, and `MetadataRequestThrottleRetryPolicy` doesn't rotate regions on it). So if the primary region's gateway is hard-down (vs slow), we currently wouldn't hedge - and if a hedge were already in flight, we'd rethrow the primary's connection error instead of taking the good hedge. The data-plane `ClientRetryPolicy` does treat `HttpRequestException` as a regional/endpoint failure, so I think it's worth aligning here: add `HttpRequestException` (and, defensively, non-user `OperationCanceledException`) to the regional-failure classification so a primary connection failure becomes hedgeable and a good hedge can win over it. It only widens what counts as hedge-worthy; the primary-authoritative guarantee for definitive answers (404 / 409 / 412 / auth) is unchanged. I have made that change and added a test covering the same.

FabianMeiswinkel (dismissed):
LGTM - Thanks!

jeet1995 (approved):

jeet1995 (dismissed):

jeet1995 (approved):

FabianMeiswinkel (approved):
LGTM

NaluTripician:
> > Address refresh, query plan, database account, and all other metadata reads
> 
> Orthogonal to this PR, `QueryPlan` calls can be hedged.

I will create a work item and link it here

NaluTripician:
Follow-up work item for hedging the remaining metadata reads (QueryPlan, address refresh, database account, and others): #6005. Scoped to start after this PR merges.

PR #6003, opened 2026-07-06
LINQ: Fixes MemoryExtensions.Contains for enum arrays on .NET 10

### LINKED ISSUE #5980: .NET 10: MemoryExtensions.Contains fails for enum arrays in Cosmos LINQ (Missed edge case in #5518)
Describe the bug

On .NET 10, array.Contains(x) in Cosmos LINQ still fails with Method 'Contains' is not supported when the array is an enum type, even on SDK 3.61.0 (which includes the #5518 / PR #5819 fix for MemoryExtensions.Contains). The same pattern works with an int[].

This looks like an incomplete fix for [#5518](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5518): LinqMemoryExtensionsTests covers string[], int[], and Guid[], but not enum[].

To Reproduce

```
using Microsoft.Azure.Cosmos;

public enum ItemStatus { Active = 40, Pending = 50 }

public class Item
{
    public string id { get; set; } = "";
    public string pk { get; set; } = "";
    public ItemStatus Status { get; set; }
}

// Fails on .NET 10 + Microsoft.Azure.Cosmos 3.61.0
ItemStatus[] statuses = [ItemStatus.Active, ItemStatus.Pending];

var q1 = container.GetItemLinqQueryable<Item>()
    .Where(x => statuses.Contains(x.Status));
_ = q1.ToQueryDefinition().QueryText; // throws: Method 'Contains' is not supported.

// Works on same SDK / runtime
int[] statusIds = [(int)ItemStatus.Active, (int)ItemStatus.Pending];

var q2 = container.GetItemLinqQueryable<Item>()
    .Where(x => statusIds.Contains((int)x.Status));
_ = q2.ToQueryDefinition().QueryText; // OK: IN (40, 50)
```

Expected behavior

Enum arrays should translate the same way as int[] (or at least as Enumerable.Contains did pre-.NET 10), e.g.:

WHERE root["Status"] IN (40, 50)
Actual behavior

Enum array Contains throws during LINQ translation:

Method 'Contains' is not supported., Windows/10.0.26200 cosmos-netstandard-sdk/3.43.2
(int[] with the same values works.)

Environment summary

SDK Version: Microsoft.Azure.Cosmos 3.61.0
OS Version: Windows 10/11 (e.g. Windows/10.0.26200)
Target framework: net10.0
Additional context

Regression tied to .NET 10 / C# 14: array.Contains(x) resolves to MemoryExtensions.Contains(ReadOnlySpan<T>, T) instead of Enumerable.Contains.
[#5518](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5518) / PR #5819 fixed string[], int[], Guid[]; enum[] appears omitted from that coverage.

Workaround: use List<T> or int[] (with cast on the document field), as suggested in #5518.
Related: [#1253](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/1253) (enum Contains / serialization), but this report is specifically about translation failure on the .NET 10 MemoryExtensions path, not wrong SQL values.

adityasa:
## Description

Fixes #5980

On .NET 10, enumArray.Contains(x.Status) resolves to a 3-argument overload of MemoryExtensions.Contains:

    MemoryExtensions.Contains<T>(ReadOnlySpan<T>, T, IEqualityComparer<T>)

This happens because enum types don't implement IEquatable<T>, so the 2-arg overload (which has the IEquatable<T> constraint) doesn't apply. The compiler falls back to the 3-arg overload.

PR #5819 only handled the 1-arg (instance) and 2-arg (static extension) Contains patterns. The 3-arg case fell through, causing 'Method Contains is not supported'.

### Fix
Added a third branch in ArrayContainsVisitor.VisitImplicit to handle Arguments.Count == 3 when DeclaringType == typeof(MemoryExtensions). The third argument (IEqualityComparer) is ignored for SQL translation.

### Testing
- Added Translate_MemoryExtensionsContains_EnumArray_ProducesInClause that simulates the exact expression tree .NET 10 emits
- Test is Inconclusive on pre-.NET 10 runtimes (3-arg overload doesn't exist)
- All 12 existing MemoryExtensions tests continue to pass

## Type of change


Microsoft.Azure.Cosmos/src/Linq/BuiltinFunctions/ArrayBuiltinFunctions.cs
```diff
@@ -60,6 +60,14 @@ protected override SqlScalarExpression VisitImplicit(MethodCallExpression method
                     searchList = methodCallExpression.Arguments[0];
                     searchExpression = methodCallExpression.Arguments[1];
                 }
+                // In .NET 10+, enum arrays resolve to MemoryExtensions.Contains(ReadOnlySpan<T>, T, IEqualityComparer<T>)
+                // which has 3 arguments. The third argument (comparer) is ignored for SQL translation.
```
adityasa: Should this check be tighter? IOW, could other types use this overload with three parameters? Do we need to restrict T to be of enum type? #Closed

adityasa: Good question. The 3-arg overload MemoryExtensions.Contains<T>(ReadOnlySpan<T>, T, IEqualityComparer<T>) has **no constraint on T** (unlike the 2-arg which requires IEquatable<T>). So yes, non-enum types could also hit this overload — for example, a user could write stringArray.Contains(x.Name, StringComparer.OrdinalIgnoreCase).

I intentionally did NOT restrict to enum types because:
1. The compiler only emits this 3-arg overload when the 2-arg doesn't apply (i.e., T doesn't implement IEquatable<T>), which is primarily enums. But a user could explicitly use it.
2. Ignoring the comparer is the correct behavior regardless of T — Cosmos DB's IN() uses server-side comparison semantics and doesn't support custom comparers. This is the same as how Enumerable.Contains with a custom comparer would behave if translated.
3. Restricting to enums would break if future .NET versions use this overload for other types.

The check DeclaringType == typeof(MemoryExtensions) already scopes this to only MemoryExtensions calls (not other 3-arg Contains methods on user types).

adityasa: What negative cases should be considered? Can you add test coverage for those?

IOW, if I use a user defined type (non-enum) for which this overload gets called, what is the behavior? Can we baseline it with a unit test?

adityasa: Added two additional test cases:

1. **Non-enum struct** (TestStruct without IEquatable<T>) — verifies the 3-arg overload works for user-defined value types that don't implement IEquatable, producing a correct IN clause.

2. **Non-null comparer** (string[] with StringComparer.OrdinalIgnoreCase) — verifies that when a user explicitly passes a comparer, it's safely ignored and the translation still produces the correct IN clause with the expected values.

Both cases confirm the behavior is correct: the comparer argument is ignored, and the SQL IN() clause is generated using server-side comparison semantics regardless of the element type.

Microsoft.Azure.Cosmos/src/Linq/BuiltinFunctions/ArrayBuiltinFunctions.cs
```diff
@@ -60,6 +60,25 @@ protected override SqlScalarExpression VisitImplicit(MethodCallExpression method
...
+                    && methodCallExpression.Method.DeclaringType == typeof(MemoryExtensions))
+                {
+                    Expression comparerArg = methodCallExpression.Arguments[2];
+                    if (comparerArg is ConstantExpression comparerConstant && comparerConstant.Value != null)
+                    {
+                        throw new DocumentQueryException(
+                            string.Format(
+                                CultureInfo.CurrentCulture,
+                                ClientResources.MethodNotSupported,
+                                methodCallExpression.Method.Name + " with a custom IEqualityComparer"));
```
adityasa: Do we need explicit error message, or can we just let this fall in to L85 and generate an error message? What is the experience in that case - can you check? #Closed

adityasa: Simplified — removed the explicit throw. The branch condition now requires the comparer to be a null ConstantExpression. If it's non-null, the branch doesn't match, searchList/searchExpression stay null, VisitImplicit returns null, and the parent BuiltinFunctionVisitor.Visit() throws the standard 'Method Contains is not supported' error (line 31). Same error message, less code.

changelog.md
```diff
@@ -19,13 +19,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Features Added
 
+- [5968](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5968) (Preview) Query: Adds Full-Text Policy standard package and analyzer configuration. Enables specifying `package` ("legacy" or "standard"), `defaultSpec` (language, tokenizer, filters, stopWordListKind, addStopWords, removeStopWords), and per-path overrides on the full-text policy. This allows fine-grained control over text analysis including tokenizer selection, filter pipelines, and custom stop word lists.
```
Maya-Painter: Missed on previous PR?

Maya-Painter (dismissed):

neildsh (dismissed):

Maya-Painter (approved):

neildsh (approved):

khdang (approved):

sboshra (approved):

kundadebdatta (approved):
LGTM

PR #6007, opened 2026-07-07
[Internal] Routing: Refactors partition-key resolution into a request-free shared core

Meghana-Palaparthi:
## Why

`AddressResolver.TryResolveServerPartitionByPartitionKey` (the gateway point-operation hot path) owns the canonical logic for mapping a partition key to its owning `PartitionKeyRange`: the empty-or-full-key **component-count guard**, effective-key computation, and `GetRangeByEffectivePartitionKey` lookup. Other code paths (notably the distributed-transaction session-token resolver) need the *same* mapping, and re-implementing it has already caused drift once — the DTX component-count guard was missing until it was added later to match this method. Duplicated resolution logic will drift again.

This PR extracts a **request-free shared core** so the two paths can't diverge, without changing any observable gateway behavior.

## What

- New `internal enum PartitionKeyRangeResolutionKind { Resolved, NeedsRefresh, KeyMismatch }`.
- New `internal static PartitionKeyRangeResolutionKind TryResolvePartitionKeyToRange(PartitionKeyInternal, ContainerProperties, CollectionRoutingMap, bool collectionCacheUptoDate, out PartitionKeyRange)` — the single source of truth for the component-count guard + effective-key + range lookup. It **returns** a classification instead of throwing / returning `null`.
- `TryResolveServerPartitionByPartitionKey` keeps all its `DocumentServiceRequest` handling (null-checks, JSON parse, `ResourceAddress` decoration) and **translates the core result byte-for-byte**:
  - `Resolved` → return the range
  - `KeyMismatch` → throw `BadRequestException(RMResources.PartitionKeyMismatch)` with the `PartitionKeyMismatch` substatus + `ResourceAddress`
  - `NeedsRefresh` → return `null`

All three existing call sites are unchanged, and the gateway's throw/return-null contract is preserved exactly.

## Follow-up (separate PR / branch)

The distributed-transaction resolver hookup — calling this shared core and deleting its private inline component-count guard — lands on the DTX session-token branch during its post-merge rebase, so this risky gateway-hot-path change ships isolated.

## Tests

New `AddressResolverPartitionKeyResolutionTests`:
- Core matrix: full key ⇒ `Resolved`+range; empty key ⇒ `Resolved`; partial (hierarchical prefix) + cache up-to-date ⇒ `KeyMismatch`; partial + stale ⇒ `NeedsRefresh`; null-arg guards throw.
- Gateway wrapper: `KeyMismatch` ⇒ throws `BadRequestException` with `PartitionKeyMismatch` substatus + `ResourceAddress`; stale partial key ⇒ `null`; full key ⇒ range.

Local validation (net6.0): src Release build clean; test project Debug build clean; **239/239** tests pass across the `AddressResolver`, `DistributedTransaction`, `SessionContainer`, and `GatewayStoreModel` filters (10 new + 229 pre-existing regression).

## Changelog

No entry required — pure internal refactor with gateway behavior preserved byte-for-byte and no customer-observable change (per the repo changelog classifier).

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeResolutionKind.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeResolutionKind.cs
@@ -18,10 +18,12 @@ namespace Microsoft.Azure.Cosmos
         Resolved,
 
         /// <summary>
-        /// The key could not be mapped with the currently cached routing/partition-key definition
-        /// and the caller should refresh its caches and retry (historically the gateway returned null).
+        /// The key could not be mapped because the currently cached routing/partition-key definition
+        /// may be stale — unlike <see cref="KeyMismatch"/>, this is not a definite mismatch. Each caller
+        /// picks its own reaction (the gateway refreshes its caches and retries — historically returning
+        /// null; distributed transactions apply no session token and degrade to eventual consistency).
         /// </summary>
-        NeedsRefresh,
+        StaleMetadata,
 
         /// <summary>
         /// The supplied key has a different number of components than the (up-to-date) partition-key
--- a/Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeResolutionKind.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeResolutionKind.cs
@@ -1,0 +1,32 @@
+//------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+//------------------------------------------------------------
+namespace Microsoft.Azure.Cosmos
+{
+    using Microsoft.Azure.Documents;
+
+    /// <summary>
+    /// Outcome of mapping a partition key to a physical partition key range, independent of any
+    /// <see cref="DocumentServiceRequest"/>. Callers translate the outcome into their own contract
+    /// (e.g. the gateway throws / returns null; distributed transactions apply no session token).
+    /// </summary>
+    internal enum PartitionKeyRangeResolutionKind
+    {
+        /// <summary>
+        /// The key maps to exactly one range, returned via the out parameter.
+        /// </summary>
+        Resolved,
+
+        /// <summary>
+        /// The key could not be mapped with the currently cached routing/partition-key definition
+        /// and the caller should refresh its caches and retry (historically the gateway returned null).
+        /// </summary>
+        NeedsRefresh,
+
+        /// <summary>
+        /// The supplied key has a different number of components than the (up-to-date) partition-key
+        /// definition, so it is a genuine mismatch (historically the gateway threw a 400 PartitionKeyMismatch).
+        /// </summary>
+        KeyMismatch,
+    }
+}
--- a/Microsoft.Azure.Cosmos/src/Routing/AddressResolver.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/AddressResolver.cs
@@ -679,6 +679,95 @@ namespace Microsoft.Azure.Cosmos
                 throw new InternalServerErrorException(string.Format(CultureInfo.InvariantCulture, "partition key is null '{0}'", partitionKeyString));
             }
 
+            // Delegate to the request-free core, then translate its result into the gateway's historical
+            // contract (return range / return null to refresh / throw on mismatch).
+            PartitionKeyRangeResolutionKind resolutionKind = AddressResolver.TryResolvePartitionKeyToRange(
+                partitionKey,
+                collection,
+                routingMap,
+                collectionCacheUptoDate,
+                out PartitionKeyRange range);
+
+            switch (resolutionKind)
+            {
+                case PartitionKeyRangeResolutionKind.Resolved:
+                    return range;
+
+                case PartitionKeyRangeResolutionKind.KeyMismatch:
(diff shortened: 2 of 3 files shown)


NaluTripician (dismissed):
Approving after a local skeptical review. The "byte-for-byte" gateway-behavior claim holds.

**Behavioral equivalence (verified line-by-line):** the wrapper's observable contract is identical for every input class — empty/full key → range; empty/full key with no owning range → `null`; partial key + up-to-date cache → `BadRequestException` (PartitionKeyMismatch substatus + `ResourceAddress`); partial key + stale cache → trace + `null`. Exception type/message, substatus header, `ResourceAddress`, trace message/order, and null-vs-throw are all preserved. The `out range` contract is sound (init null; set non-null only on `Resolved`).

**Local validation (net6.0):**
- New `AddressResolverPartitionKeyResolutionTests`: 10/10 pass, and they exercise the real logic (real `CollectionRoutingMap` / `PartitionKeyInternal` / `ContainerProperties`, nothing mocked). A `KeyMismatch`↔`NeedsRefresh` swap would fail the wrapper tests.
- Regression set (AddressResolver / DistributedTransaction / SessionContainer / GatewayStoreModel): 239/239 pass.
- src Release build: 0 warnings, 0 errors.
- The 3 existing call sites are untouched.

**Non-blocking notes (optional polish, not required for merge):**
- The defensive core null-checks and the `range == null → NeedsRefresh` branch are unreachable via the wrapper, but they're honestly documented and justified by the upcoming DTX direct caller.
- Pre-existing `JsonException`/`InvalidPartitionKey` and null-`partitionKey` wrapper branches remain untested — fine for a refactor, but the obvious next tests to add.
- `no changelog entry required` is correct per the repo classifier (internal-only, gateway behavior preserved).

Shipping the risky gateway change isolated from the DTX hookup is the right call. LGTM.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeResolutionKind.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// The key maps to exactly one range, returned via the out parameter.
+        /// </summary>
+        Resolved,
+
+        /// <summary>
+        /// The key could not be mapped with the currently cached routing/partition-key definition
+        /// and the caller should refresh its caches and retry (historically the gateway returned null).
+        /// </summary>
+        NeedsRefresh,
```
ananth7592: One naming thought, and I think it matters a bit more here than usual because this enum is about to become the shared contract the DTX resolver consumes next. `NeedsRefresh` names a *reaction* (go refresh), while its siblings `Resolved` and `KeyMismatch` name *states*. That's accurate for the gateway, which does refresh + retry — but the DTX consumer won't refresh at all; it'll just skip the session token, so for that caller the name describes something it never does. What the value actually means is "couldn't resolve because the locally cached metadata may be stale, so unlike `KeyMismatch` I can't call this a definite mismatch."

Would something state-based like `Unresolved` or `StaleMetadata` read better? Each caller then owns its own reaction in its own code, and the shared vocabulary stays caller-agnostic — which is the whole point of the extraction. Cheap to rename now; a few consumers deep it's real churn. Not blocking.

ananth7592 (commented):
Really appreciate the care that went into this one — the PR description is a model for how to ship a hot-path refactor. The "why" is grounded in a concrete prior drift (the DTX component-count guard that went missing), the byte-for-byte translation table makes the behavior-preservation claim *auditable* rather than "trust me," and deliberately splitting the DTX hookup onto its own branch so this risky gateway change ships isolated is exactly the right call. The test matrix — including the honesty of documenting the unreachable null-range branch instead of quietly leaving it — made this a pleasure to review.

On risk: I walked the wrapper + core against the original and the outcomes map 1:1. The `PartitionKeyMismatch` `BadRequestException` (message + substatus header + `ResourceAddress`) is byte-identical, just relocated to the wrapper's switch; the `KeyMismatch`-vs-`NeedsRefresh` split stays distinct, so a genuine 400 can't silently collapse into an infinite refresh loop; and the new `range != null ? Resolved : NeedsRefresh` line is both unreachable (a valid full/empty EPK always maps to a range) and null-equivalent to the old direct return even if it were hit. The core's null-arg guards are unreachable from the gateway since the wrapper still validates all four args up front. So residual regression risk is low *and* verifiable — the one line I'd least want wrong is the `PartitionKeyMismatch` substatus, since the retry policies key off it, and that one's correct. Given this is the gateway point-op hot path, I'd frame the sign-off as "on the strength of the byte-for-byte diff," not "it's just a refactor."

One naming thought inline — happy to discuss either.

Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeResolutionKind.cs
Meghana-Palaparthi: Good call — renamed `NeedsRefresh` → `StaleMetadata`. I confirmed against the DTX resolver on #5958 that the consumer never refreshes on this outcome (it degrades to eventual consistency and applies no token), so a state-based name really is the caller-agnostic contract we want here. The gateway keeps its refresh+retry reaction; the enum just says "cache may be stale, not a definite mismatch." Updated the value, XML doc, callers, and tests. Pushed in 356d7d4bd.

ananth7592 (approved):
Lgtm

NaluTripician (approved):

kushagraThapar (approved):

PR #6008, opened 2026-07-07
AAD Authentication: Fixes ReadAccountAsync hang from CAE claims Cache Bypass

NaluTripician:
## Problem

An AAD-authenticated client can hang on its first request (e.g. `await client.ReadAccountAsync()`) after upgrading past `3.61.0`. Customer bisection: **`3.61.0` → no hang, latest `main` → hang**, AAD auth.

The only auth-path change in `3.61.0..main` is **#5549** ("[AAD token revocation]: Adds logic for handling emergency token revocation and CAE token revocation").

## Root cause

`TokenCredentialCache.RefreshCachedTokenWithRetryHelperAsync` attached a **non-empty `claims`** parameter (the `cp1` client capability) on **every** token acquisition — even when there was no revocation challenge — because `MergeClaimsWithClientCapabilities(null)` returns the `cp1` JSON rather than `null`:

```
claims = {"access_token":{"xms_cc":{"values":["cp1"]}}}
```

Azure.Identity/MSAL treats **any** non-empty `claims` as "the cached token does not satisfy this challenge", so it **bypasses `AcquireTokenSilent`'s token cache and calls the authority (ESTS) live on every acquisition**. Because all callers funnel through a single-flight refresh, a stalled live token call under an MSAL-backed credential (certificate / managed identity) surfaces as a hang on the first request.

`cp1`/CAE is **already** advertised the correct, cache-friendly way via `isCaeEnabled: true` in `CosmosScopeProvider.GetTokenRequestContext()`, so the `claims` injection was redundant.

## Fix

Attach the `claims` parameter **only** when responding to an actual CAE/revocation challenge (`cachedClaimsChallenge` is non-empty). On the normal path, pass the scope provider's context unchanged (scopes + `isCaeEnabled: true`, **no claims**), so the identity library can serve from its token cache. Continuous access evaluation still works via `isCaeEnabled`, and revocation still attaches the merged claims on the challenge path.

## Tests (definitive)

Adds 3 regression tests to `CosmosAuthorizationTests` (plus a `ClaimsCapturingTokenCredential` that models MSAL cache-bypass by stalling when `claims` is present):

1. `TokenCredentialCache_NormalAcquisition_DoesNotAttachClaims_SoMsalCacheIsUsable` — the normal acquire must send **no** `claims` and `IsCaeEnabled=true`.
2. `TokenCredentialCache_RevocationChallenge_StillAttachesMergedClaims` — a real challenge still attaches merged claims (`acrs` + `cp1`); CAE preserved.
3. `TokenCredentialCache_NormalAcquisition_DoesNotStallOnMsalCacheBypass` — deterministic hang reproduction; the normal acquire must not stall.

**Verification:**
- Against the pre-fix (#5549) code, all 3 tests **fail** — Test 1 shows the exact `cp1` claims sent on the normal path, and Test 3 **stalls** (the hang).
- With the fix, **44/44** `CosmosAuthorizationTests` pass (41 existing incl. the CAE revocation tests + 3 new). No regressions.

## Changelog

Added a `Bugs Fixed` entry under `### Unreleased`.


Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -289,25 +289,34 @@ private async ValueTask<AuthState> RefreshCachedTokenWithRetryHelperAsync(
...
+                            // Attach a CAE 'claims' parameter ONLY when responding to an actual
+                            // revocation / CAE challenge. cp1 (the CAE client capability) is already
+                            // advertised to the identity library via isCaeEnabled:true from the scope
+                            // provider, so attaching claims on the normal path is redundant AND harmful:
+                            // Azure.Identity/MSAL treats any non-empty claims as "the cached token does
+                            // not satisfy this challenge" and bypasses its token cache (AcquireTokenSilent),
+                            // issuing a live ESTS call on every acquisition. That cache-bypass is what
+                            // stalls the first token acquisition (e.g. ReadAccountAsync) under MSAL-backed
+                            // credentials.
+                            if (!string.IsNullOrEmpty(this.cachedClaimsChallenge))
```
kundadebdatta: I think we can simplify the fix here @NaluTripician. The issue is when the `cachedClaimsChallenge` is empty, the  `MergeClaimsWithClientCapabilities` returns `"{\"access_token\":{\"xms_cc\":{\"values\":[\"cp1\"]}}}"` today, which sends the above JSON to the entra, instead of AAD.

We can either do the following:

**Option 1:**

Only attach a claims payload when responding to an actual CAE / revocation challenge. On the happy path (no challenge) claims MUST stay null so the credential's own token cache is honored — a non-empty claims payload forces Azure.Identity/MSAL to bypass its cache and make a live ESTS call on every acquisition, which stalls client init (e.g. ReadAccountAsync warm-up) and is throttled by ESTS on mass restart. isCaeEnabled:true already advertises the cp1 client capability, so nothing is lost by omitting claims here.

```c#
  string? mergedClaims = string.IsNullOrEmpty(this.cachedClaimsChallenge)
      ? null
      : TokenCredentialCache.MergeClaimsWithClientCapabilities(this.cachedClaimsChallenge);
```

**Option 2 (much cleaner) :**

Update the method directly to return `null`, when the claims challange is null of empty:

```c#
        internal static string MergeClaimsWithClientCapabilities(string? claimsChallenge)
        {
            const string clientCapabilitiesJson = "{\"access_token\":{\"xms_cc\":{\"values\":[\"cp1\"]}}}";

            if (string.IsNullOrEmpty(claimsChallenge))
            {
                return null;
            }
```

In either of the above cases, we will  require thorough testing and coverage.

kundadebdatta (changes requested):

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
NaluTripician: Good call — went with Option 2. `MergeClaimsWithClientCapabilities` now returns `null` when the claims challenge is null/empty, so the happy path attaches no `claims` (the credential''s MSAL token cache stays usable) and the call site just checks the merged result. A non-empty-but-malformed challenge still falls back to `cp1` since a challenge did occur.

Coverage: split `MergeClaimsWithClientCapabilities_InvalidInput_ReturnsOnlyCp1` — the null/empty cases now assert `null` (new `MergeClaimsWithClientCapabilities_NullOrEmpty_ReturnsNull`) while invalid-base64 still returns cp1. The 3 behavioral regression tests are unchanged; 44/44 `CosmosAuthorizationTests` pass.

Fixed in 5ce0b1e1.

NaluTripician: Follow-up from our call — pushed in 5a7f5733:

1. **Cleaned up the call site** in `TokenCredentialCache` per your feedback — dropped the redundant `DefaultTrace.TraceInformation` logging and the `else` branch, and condensed the comments to a single concise block. The logic is now just: merge (returns `null` on the normal path) → attach `claims` only when non-null.
2. **Added the end-to-end regression test** you shared — `ReadAccountAsync_TokenCredentialHappyPath_DoesNotAttachClaims_SucceedsInsteadOfHanging` in `CosmosClientTests`, backed by an `EstsThrottlingTokenCredential` that 429s any cache-bypassing claims request. It fails before the fix (happy-path claims → throttled → `ReadAccountAsync` hangs/fails) and passes after (no claims on the happy path). I dropped the `docs/investigations/...rca.md` reference from the doc comment since that file isn't in this repo.

All targeted tests pass (the merge/regression unit tests + the new e2e test). Should be ready to polish and port to main.

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -289,7 +291,12 @@ private async ValueTask<AuthState> RefreshCachedTokenWithRetryHelperAsync(
...
                             tokenRequestContext = this.scopeProvider.GetTokenRequestContext();
 
-                            string mergedClaims = TokenCredentialCache.MergeClaimsWithClientCapabilities(this.cachedClaimsChallenge);
+                            // Attach a 'claims' parameter only when responding to an actual revocation /
+                            // CAE challenge. MergeClaimsWithClientCapabilities returns null on the normal
+                            // path, so no claims are sent and the credential's token cache stays usable
+                            // (a non-empty claims forces a cache-bypassing live ESTS call). cp1 is still
+                            // advertised via isCaeEnabled:true.
+                            string? mergedClaims = TokenCredentialCache.MergeClaimsWithClientCapabilities(this.cachedClaimsChallenge);
                             if (string.IsNullOrEmpty(this.cachedClaimsChallenge))
```
kushagraThapar: **Nit/Correctness: double volatile-read TOCTOU on `cachedClaimsChallenge`**

`cachedClaimsChallenge` is `volatile` and is read twice here: once on line 299 to compute `mergedClaims`, and again on line 300 for the branch condition. Because `ResetCachedToken` (called from the request-processing thread via `TryHandleTokenRevocation`) can race between these two reads, the branch and the computed value can diverge:

- Read 1 sees `null` → `mergedClaims = null`; `ResetCachedToken(challenge)` fires; Read 2 sees a challenge → enters `else` branch → builds `TokenRequestContext(claims: null)` → challenge silently dropped, then cleared at line 337 → revocation not honoured, `MaxCaeRevocationRetryCount=1` budget consumed for nothing.
- Read 1 sees a challenge → `mergedClaims = merged(acrs+cp1)`; `ResetCachedToken(null)` fires; Read 2 sees `null` → takes the `if` branch → `mergedClaims` (with claims) is never used → claims silently dropped.

This PR introduced the consequence by moving `new TokenRequestContext` into the `else` branch. In `main` the double-read was benign because the context was built unconditionally. The fix is a one-line snapshot (same pattern already used in `GetNewTokenAsync` line 233: *"Use a local variable to avoid the possibility the task gets changed between the null check and the await"*):

```csharp
string? challenge = this.cachedClaimsChallenge;   // single volatile read
string? mergedClaims = TokenCredentialCache.MergeClaimsWithClientCapabilities(challenge);
if (string.IsNullOrEmpty(challenge))
{
    DefaultTrace.TraceInformation($"Requesting AAD token with CAE client capabilities (cp1). Retry={retry}");
}
else
{
    DefaultTrace.TraceInformation($"Requesting AAD token for revocation with claims challenge and client capabilities (cp1). Retry={retry}");
    tokenRequestContext = new TokenRequestContext(
        scopes: tokenRequestContext.Scopes,
        parentRequestId: tokenRequestContext.ParentRequestId,
        claims: mergedClaims,
        tenantId: tokenRequestContext.TenantId,
        isCaeEnabled: tokenRequestContext.IsCaeEnabled);
}
```

changelog.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Bugs Fixed
 
+- [5549](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5549) AAD Authentication: Fixes a regression (introduced by the CAE/token-revocation change after `3.61.0`) where the SDK attached a non-empty `claims` parameter (the `cp1` client capability) on *every* AAD token acquisition, even when there was no revocation challenge. Because a non-empty `claims` forces the underlying identity library (Azure.Identity/MSAL) to bypass its token cache and request a fresh token from the authority on each acquisition, this could stall the first token acquisition and surface as a hang on the initial request (for example `ReadAccountAsync`) with certificate/managed-identity credentials. The SDK now attaches `claims` only when responding to an actual CAE/revocation challenge; the `cp1` capability continues to be advertised via `isCaeEnabled`, so continuous access evaluation still works while the token cache is used on the normal path.
```
kushagraThapar: **Changelog: Bugs Fixed entry links to the bug-introducing PR instead of this fix**

The link here points to `#5549` — the PR that *introduced* the regression — rather than `#6008` (this PR). Every sibling entry in the Bugs Fixed section (`#5851`, `#5977`, `#5957`) correctly links the fixing PR.

```diff
- - [5549](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/5549) AAD Authentication: Fixes a regression ...
+ - [6008](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/6008) AAD Authentication: Fixes a regression ...
```

kushagraThapar (commented):
Two findings from an independent code review — both worth addressing before merge.

changelog.md
NaluTripician: Fixed in 7eb34c8 — the Bugs Fixed entry now links this PR (#6008). I left the `#5549` reference under the Added section untouched, since that's the correct feature entry for the original CAE/token-revocation change.

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
NaluTripician: Good catch — fixed in 7eb34c8. I snapshot `cachedClaimsChallenge` into a local `challenge` up front and use that single read for both the `MergeClaimsWithClientCapabilities` call and the `IsNullOrEmpty` branch, so a concurrent `ResetCachedToken` on the request-processing thread can't make the two reads diverge (same single-volatile-read pattern already used in `GetNewTokenAsync`). Behavior on both the normal and challenge paths is unchanged, so the existing 44 `CosmosAuthorizationTests` still pass. I didn't add a dedicated test since a deterministic TOCTOU-race repro would be flaky.

kundadebdatta (dismissed):
LGTM

aavasthy (dismissed):
Was able to test this change, thanks for getting this out quickly!

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -289,8 +291,17 @@ private async ValueTask<AuthState> RefreshCachedTokenWithRetryHelperAsync(
...
-                            if (string.IsNullOrEmpty(this.cachedClaimsChallenge))
+                            // Attach a 'claims' parameter only when responding to an actual revocation /
+                            // CAE challenge. MergeClaimsWithClientCapabilities returns null on the normal
+                            // path, so no claims are sent and the credential's token cache stays usable
+                            // (a non-empty claims forces a cache-bypassing live ESTS call). cp1 is still
+                            // advertised via isCaeEnabled:true.
+                            // Snapshot the volatile cachedClaimsChallenge into a local so the merged claims
+                            // and the branch below are computed from the same value, even if ResetCachedToken
+                            // races on the request-processing thread between the two reads.
+                            string? challenge = this.cachedClaimsChallenge;
```
kushagraThapar: I think there’s one concurrency edge here worth hardening before merge. `ResetCachedToken(claims)` updates `authState`, `currentRefreshOperation`, and `cachedClaimsChallenge` under `backgroundRefreshLock`, but an already-running refresh publishes those same fields later without the same guard: it unconditionally clears `cachedClaimsChallenge`, writes `authState`, and then clears `currentRefreshOperation` in `finally`.

That leaves a stale-refresh interleaving:

1. a normal no-claims refresh starts and awaits `GetTokenAsync`;
2. a CAE 401 arrives, so `ResetCachedToken(claims)` stores the new challenge and clears the current refresh task;
3. the retry starts a claims-based refresh;
4. the older no-claims refresh completes afterward and can republish a no-claims `authState`, clear the challenge, and/or clear the newer refresh task.

Most of the time this may self-heal with another 401, but `ClientRetryPolicy.MaxCaeRevocationRetryCount` is 1, so serving the stale no-claims token on the retry path can become a user-visible auth failure.

Could we add a generation/snapshot guard here? For example, increment a `refreshGeneration` in `ResetCachedToken`, capture `{generation, challenge}` when a refresh starts, and only clear `cachedClaimsChallenge` / publish `authState` if the generation is still current. In `finally`, also compare-and-clear `currentRefreshOperation` only if it is this refresh’s task. A deterministic test can gate two `GetTokenAsync` calls so the no-claims refresh starts first, `ResetCachedToken(claims)` starts the claims refresh, the claims refresh completes, then the stale no-claims refresh completes last; the final cached header should still be the claims-based one.

NaluTripician: Good catch — fixed in d0cec7dd8. I added a `refreshGeneration` counter that `ResetCachedToken` bumps under `backgroundRefreshLock`. `RefreshCachedTokenWithRetryHelperAsync` snapshots it at the start and now only publishes its result — setting `authState`, clearing `cachedClaimsChallenge`, and clearing `currentRefreshOperation` in `finally` — when the snapshot still matches the current generation. A refresh that was already in flight when a `ResetCachedToken(claims)` raced it therefore returns its token to its own awaiters but no longer republishes stale no-claims state, drops the installed challenge, or clears the newer refresh's `currentRefreshOperation`, so the revocation is still honored under `MaxCaeRevocationRetryCount == 1`.

I used the generation (rather than a raw `Task`-identity check) for the `finally` compare-and-clear since a stale refresh always holds an older generation — equivalent guarantee, simpler read. I also guarded the two failure-path clears (401/403 and retries-exhausted) the same way.

Added `TokenCredentialCache_StaleNoClaimsRefresh_DoesNotClobberClaimsToken`, a deterministic test that parks the no-claims refresh, fires `ResetCachedToken(claims)`, completes the claims-based refresh, then releases the stale one last and asserts the cached header is still the claims-based token. It fails before the fix (stale no-claims token wins) and passes after; the full 45-test `CosmosAuthorizationTests` suite is green.

kushagraThapar (commented):
Shall we rebase the new feature flag PR and then re-test this? The newly added tests won't work in this PR since the feature flag is now by default off.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosAuthorizationTests.cs
```diff
@@ -781,8 +796,244 @@ public async Task TokenCredentialCache_ResetWithNullClaims_RefreshesToken()
...
+                $"ReadAccountAsync under MSAL-backed credentials. Actual claims sent: '{claims}'.");
+
+            Assert.IsTrue(
+                isCaeEnabled,
+                "CAE / cp1 must still be advertised the cache-friendly way, via isCaeEnabled:true.");
+        }
+
+        [TestMethod]
+        public async Task TokenCredentialCache_RevocationChallenge_StillAttachesMergedClaims()
+        {
```
aavasthy, line 847: do these token revocation tests need AAD feature flag to be enabled?

aavasthy (dismissed):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/CosmosAuthorizationTests.cs
NaluTripician: Yes, the flag is required — and each of the revocation regression tests already enables it. The `using IDisposable revocationEnabled = EnableAadTokenRevocation();` line sits right under each test''s opening brace (this test: line 848), and the helper is defined at line 1286 (it sets `AZURE_COSMOS_AAD_TOKEN_REVOCATION_ENABLED=True`).

| Test | `EnableAadTokenRevocation()` line |
|---|---|
| `TokenCredentialCache_NormalAcquisition_DoesNotAttachClaims_SoMsalCacheIsUsable` | 820 |
| `TokenCredentialCache_RevocationChallenge_StillAttachesMergedClaims` (this one) | 848 |
| `TokenCredentialCache_NormalAcquisition_DoesNotStallOnMsalCacheBypass` | 880 |
| `TokenCredentialCache_StaleNoClaimsRefresh_DoesNotClobberClaimsToken` | 914 |

It''s required because the entire claims/CAE path in `TokenCredentialCache.RefreshCachedTokenWithRetryHelperAsync` is gated behind `ConfigurationManager.IsAadTokenRevocationEnabled()`, which defaults to `false` and is read live on every acquisition. With the flag off this test would never enter the merge branch, so no `claims` would be attached on the revocation retry and the `revocationClaims.Length > 0` assertion would fail.

Note the tests that intentionally do **not** set the flag: `MergeClaimsWithClientCapabilities_NullOrEmpty_ReturnsNull` / `_InvalidChallenge_ReturnsOnlyCp1` call the static merge helper directly (not flag-gated), and `TokenCredentialCache_ResetWith{Null}Claims_*` only assert that a refresh occurred (`callCount == 2`), not that claims were attached — so none of those need the flag.

aavasthy: but you dont have the flag set or am I missing something?

aavasthy: ok i see the flag enabled

Microsoft.Azure.Cosmos/src/Authorization/TokenCredentialCache.cs
```diff
@@ -406,7 +467,17 @@ private async ValueTask<AuthState> RefreshCachedTokenWithRetryHelperAsync(
                 try
                 {
                     await this.isTokenRefreshingLock.WaitAsync();
-                    this.currentRefreshOperation = null;
+
+                    // Compare-and-clear: only clear currentRefreshOperation if this refresh is still
```
kundadebdatta: **🟡 Recommendation** · Concurrency: Stale-Refresh Race `[Partially covered by @kushagraThapar]`

**Generation-guarded clear of `currentRefreshOperation` is not equivalent to a task-identity compare-and-clear**

This comment says "only clear `currentRefreshOperation` if this refresh is still the current one," but the code checks *generation* (`refreshGeneration == refreshGenerationSnapshot`), which answers "did a `ResetCachedToken` happen since I started?" — not "am I still the task in the slot?". Those diverge because the generation snapshot (taken under `backgroundRefreshLock` inside this method) and the slot assignment `this.currentRefreshOperation = currentTask;` in `GetNewTokenAsync` (taken under `isTokenRefreshingLock`) are **not atomic**, and `ResetCachedToken` takes only `backgroundRefreshLock`:

1. A first refresh `R1` snapshots `generation = 0`, then awaits `GetTokenAsync`.
2. A CAE 401 fires `ResetCachedToken(challenge)`: `generation -> 1`, `authState = null`, `currentRefreshOperation = null`, installs the challenge.
3. `GetNewTokenAsync` resumes and runs the **unconditional** `this.currentRefreshOperation = R1` — planting a *stale, gen-0* task in the slot.
4. `R1` completes: publish-guard `gen(1) == snap(0)`? No -> correctly does not publish. **This `finally` guard:** `gen(1) == snap(0)`? No -> **does not clear the slot.**

Now `currentRefreshOperation` holds a completed, stale, no-claims task and `authState` is `null`. Every later caller hits `if (currentTask != null) return await currentTask;` in `GetNewTokenAsync`, so they all get the stale no-claims token and **never start a fresh claims-based refresh**. The current request's single revocation retry (`MaxCaeRevocationRetryCount == 1`) is spent on the stale token -> user-visible auth failure and the challenge is dropped — the exact failure class this PR set out to prevent. It stays stuck until the next `ResetCachedToken`. (Borderline blocking — treat as 🔴 if honoring CAE revocation under the single retry is a hard requirement for enabling the flag.)

A pure generation check cannot distinguish "superseded but still the slot occupant" (should clear) from "superseded and replaced" (must not clear); a task-identity check can. This is why substituting generation for the originally-suggested identity compare-and-clear is not equivalent.

**Suggested direction** — make the snapshot + slot assignment atomic under `backgroundRefreshLock` (pass the snapshot into the refresh method and assign the slot in the same locked region), and/or use a true identity clear for the slot while keeping generation for the `authState`/`cachedClaimsChallenge` publish guards:

```csharp
// in GetNewTokenAsync, which holds the task reference, clear by identity:
if (ReferenceEquals(this.currentRefreshOperation, currentTask))
{
    this.currentRefreshOperation = null;
}
```

Note: the new test `TokenCredentialCache_StaleNoClaimsRefresh_DoesNotClobberClaimsToken` only exercises the benign ordering (the slot is populated before the reset), so this snapshot-vs-assignment window is currently untested.

---

NaluTripician: Good catch — fixed in f91b01b03. You're right that a generation check answers "did a `ResetCachedToken` happen" rather than "am I still the task in the slot," and the snapshot-vs-assignment window let a stale no-claims task get planted and then wedge every future refresh.

I addressed both halves you called out:

1. **Atomic snapshot + assignment.** The generation snapshot, the operation-id stamp, the task creation, and the `currentRefreshOperation` assignment now all happen together inside a single `lock (this.backgroundRefreshLock)` region in `GetNewTokenAsync`. Since `ResetCachedToken` also takes that lock, it can no longer slip between the snapshot and the slot assignment, so a stale/superseded refresh is never orphaned in the slot.

2. **True identity clear for the slot.** The `finally` now clears `currentRefreshOperation` by identity (`this.currentRefreshOperationId == refreshOperationId`) instead of by generation. Generation is kept solely for the `authState` / `cachedClaimsChallenge` publish guards, exactly as you suggested. I used a monotonic operation-id rather than `ReferenceEquals(currentRefreshOperation, currentTask)` only because the clear lives inside the refresh helper (which doesn't hold the outer `.AsTask()` reference); it's equivalent identity semantics.

On the untested window: I added `TokenCredentialCache_StaleRefreshPlantedDuringInstall_DoesNotWedgeSlot`, which fires `ResetCachedToken(claims)` from inside the credential *while the no-claims refresh is being installed* (the exact snapshot-vs-assignment interleaving), then asserts the next caller performs a fresh claims-based refresh. I confirmed it fails on the old code (returns the stale `NoClaimsToken`) and passes with this change.

kundadebdatta (dismissed):
LGTM

kushagraThapar (dismissed):
LGTM, thanks!

kundadebdatta:

kundadebdatta (approved):
LGTM

aavasthy (approved):

PR #6010, opened 2026-07-09
Add WithRawJsonParameter method to QueryDefinition

JoasE:
to prevent double encoding for raw json strings

# Pull Request Template

## Description

Adds `QueryDefinition.WithRawJsonParameter(string name, string jsonString)` to allow callers to pass pre-serialized JSON parameter values into SQL queries without the SDK serializing them again. This prevents double encoding for raw JSON strings and matches the existing internal serialized-parameter path used by stream parameters.

Example:

```csharp
QueryDefinition query = new QueryDefinition(
    "select * from t where t.pk = @pk")
    .WithRawJsonParameter("@pk", "\"myPk\"");
```

## Type of change

New feature (non-breaking change which adds functionality)

## Changelog

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Query/v3Query/QueryDefinition.cs
+++ b/Microsoft.Azure.Cosmos/src/Query/v3Query/QueryDefinition.cs
@@ -148,6 +148,30 @@ namespace Microsoft.Azure.Cosmos
         }
 
         /// <summary>
+        /// Add a raw json string parameter to the SQL query.
+        /// </summary>
+        /// <param name="name">The name of the parameter.</param>
+        /// <param name="jsonString">The raw JSON string value for the parameter.</param>
+        /// <example>
+        /// <code language="c#">
+        /// <![CDATA[
+        /// QueryDefinition query = new QueryDefinition(
+        ///     "select * from t where t.Account = @account")
+        ///     .WithRawJsonParameter("@account", jsonValue);
+        /// ]]>
+        /// </code>
+        /// </example>
+        /// <returns>An instance of <see cref="QueryDefinition"/>.</returns>
+        public QueryDefinition WithRawJsonParameter(string name, string jsonString)
+        {
+            // pack it into an internal type for identification.
+            return this.WithParameter(name, new SerializedParameterValue
+            {
+                rawSerializedJsonValue = jsonString
+            });
+        }
+
+        /// <summary>
         /// Returns the names and values of parameters in this <see cref="QueryDefinition"/>.
         /// </summary>
         /// <returns>
--- a/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
+++ b/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
@@ -1654,6 +1654,242 @@ namespace Microsoft.Azure.Cosmos.SDK.EmulatorTests
         }
 
         [TestMethod]
+        public async Task QueryRawJsonValueTest()
+        {
+            DateTime createDateTime = DateTime.UtcNow;
+
+            static string ToJsonString(Stream stream)
+            {
+                using (stream)
+                using (StreamReader streamReader = new StreamReader(stream, Encoding.UTF8))
+                {
+                    return streamReader.ReadToEnd();
+                }
+            }
+
+            dynamic testItem1 = new
+            {
+                id = "testItem1",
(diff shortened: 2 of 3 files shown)


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
```diff
@@ -4597,6 +4833,6 @@ private static bool IsTextFormat(
             JsonSerializationFormat desiredFormat)
         {
             return desiredFormat == JsonSerializationFormat.Text && firstByte < (int)JsonSerializationFormat.Binary;
-        }
-    }
-}
+        }
+    }
+}
```
NaluTripician: Thank you for your contribution! Thius file has lots of whitespace changes. Please clean up so only your changes show :)

NaluTripician (changes requested):

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.EmulatorTests/CosmosItemTests.cs
JoasE: Sorry about that! The file contains mixed line endings which were automatically converted. I've brought back the mixed line endings to make the changes reviewable

Microsoft.Azure.Cosmos/src/Query/v3Query/QueryDefinition.cs
```diff
@@ -147,6 +147,30 @@ public QueryDefinition WithParameterStream(string name, Stream valueStream)
...
+        /// <code language="c#">
+        /// <![CDATA[
+        /// QueryDefinition query = new QueryDefinition(
+        ///     "select * from t where t.Account = @account")
+        ///     .WithRawJsonParameter("@account", jsonValue);
+        /// ]]>
+        /// </code>
+        /// </example>
+        /// <returns>An instance of <see cref="QueryDefinition"/>.</returns>
+        public QueryDefinition WithRawJsonParameter(string name, string jsonString)
```
FabianMeiswinkel, line 165: I don't get why we would want to add new public surface area for this? The encoding can easily be normalized by callers at the app-layer?

FabianMeiswinkel (changes requested):
I don't get why we would want to add new public surface area for this? The encoding can easily be normalized by callers at the app-layer?

Microsoft.Azure.Cosmos/src/Query/v3Query/QueryDefinition.cs
JoasE: @FabianMeiswinkel Good point, it's a small thing; my reason was to prevent callers from having to perform redundant encoding on raw json strings which they want to add as a parameter, since the current alternative `WithParameterStream` method directly reads the stream as a string.

```csharp
var rawJson = """{ "someBigObjectProp": "1", ... }"""
queryDefinition.WithParameterStream("param", new MemoryStream(Encoding.UTF8.GetBytes(rawJson))); // Immediately decoded to the same string by the sdk
queryDefinition.WithRawJsonParameter("param", rawJson); // No redundant en/de-coding
```

I agree that the benefit is relatively minor. But since the implementation was small, I decided to just open a PR. If this doesn’t warrant a new public API, I can keep using WithParameterStream.

NaluTripician:
Thanks again for the contribution and for the thoughtful discussion, [@JoasE](https://github.com/JoasE) — and for laying out the trade-off so clearly.

After looking it over, I think we'll go ahead and close this one. A few reasons:

There's already a supported path. As you noted, WithParameterStream covers this scenario today — WithRawJsonParameter ends up wrapping the string in the same internal SerializedParameterValue, so it's essentially a convenience alias rather than new capability.
New public API is a long-term commitment. Every addition to the public surface is something we carry forward indefinitely (contracts, docs, back-compat), so we try to reserve it for cases where there's no existing way to accomplish the task. Given the workaround, the cost/benefit doesn't quite land here.
Reduced misuse surface. The raw-JSON path writes the caller-supplied string directly into the request payload via WriteRawValue, with no escaping or validation. A friendly, string-based entry point makes it easy to pass unvalidated or user-controlled input, which we'd rather not encourage. WithParameterStream intentionally keeps that path narrow and explicit for the encrypted-value use case.
None of this is a reflection on the quality of the change — it's clean and well done. It's really about keeping the public API minimal and steering callers toward the safer, validated parameter paths. Please do keep the issues and PRs coming; we really appreciate it. 🙏

PR #6015, opened 2026-07-13
Docs: Adds status-code clarity for metadata and data-plane hedging

NaluTripician:
## Description

Improves the cross-region hedging documentation so both hedging systems are clearly explained and up to date with `main`, with an explicit, source-verified breakdown of **which status codes the SDK hedges on**.

### `docs/Cross Region Request Hedging.md`
- **Overview** — new comparison table distinguishing **data-plane hedging** (`AvailabilityStrategy`) from **metadata hedging** (coverage, configuration, fan-out, trigger, status codes).
- **Data-plane status codes** — replaced the single one-sided "Considered Final" table with two clearly-labeled tables verified against `CrossRegionHedgingAvailabilityStrategy.IsFinalResult`:
  - *Final (accepted, hedging stops):* 1xx/2xx/3xx, 400, 401, 404/0, 405, 409, 412, 413.
  - *Transient (SDK keeps hedging):* 403, 404 (sub-status ≠ 0), 408, 410, 429, 449, 500, 503, network failures.
- **Programmatic Hedging Detection** — documents the `HedgingStarted()` / `GetRequestedRegions()` / `GetRespondedRegions()` API (+ `RequestedRegion` / `RequestedRegionReason`).
- **Operator kill-switch** — documents the Gateway `disableCrossRegionalHedging` flag and its precedence in the resolution order.
- **New Metadata Hedging section** — scope (Collection Read, PKRange ReadFeed first page), enablement (`AZURE_COSMOS_METADATA_HEDGING_ENABLED` + PPAF, preview vs GA), fixed 1.5s threshold, primary-authoritative guarantee, a status-code table (regional failures that hedge vs definitive outcomes, verified against `IsRegionalFailure`), and the `Metadata Hedge` trace datum.

### `docs/metadata-hedging-simple-design.md`
- Added a "See also" cross-link to the new user-facing section, clarifying this file is the internal design/implementation reference.

## Type of change

- [x] Documentation update

## Changelog

No changelog entry required — this is a **docs-only** change; nothing under `Microsoft.Azure.Cosmos/src/**` (or any shipped-package source) was modified, so there is no customer-observable behavior change on upgrade. All status-code claims were cross-checked directly against the SDK source (`IsFinalResult` / `IsRegionalFailure`).


docs/Cross Region Request Hedging.md
```diff
@@ -1,5 +1,19 @@
 # Cross Region Request Hedging
 
+The Cosmos SDK has two independent cross-region hedging systems. Both send a redundant copy of a slow request to another region and return the first acceptable answer while keeping the primary region authoritative, but they cover different request types and trigger on different signals:
```
kushagraThapar: **🔴 Blocking** · Correctness: Documented behavior

**The shared primary-authoritative guarantee is stronger than the implementation.**

The opening says both systems keep the primary authoritative. The current SDK does not make that shared guarantee: data-plane returns the first final arm (CrossRegionHedgingAvailabilityStrategy.cs:223-245), while metadata lets a successful hedge win while primary is pending or after a regional failure (MetadataHedgingStrategy.cs:214-252). MetadataHedgingStrategyTests.cs:84-101 verifies a hedge 200 replacing a primary 503. This can lead callers to assume a result-selection contract the SDK does not provide.

**Suggested wording:**

> Data-plane hedging returns the first final response from any arm. For metadata, an already-completed definitive primary result wins; otherwise a successful hedge may win while primary is pending or after a regional failure.

---

docs/Cross Region Request Hedging.md
```diff
@@ -137,22 +151,65 @@ In the diagnostics data there are three areas of note that will appear when hedg
...
+    foreach (RequestedRegion region in diagnostics.GetRequestedRegions())
+    {
+        Console.WriteLine($"Dispatched to {region.RegionName} ({region.Reason})");
+    }
+}
+```
+
+### Status Codes: What the SDK Accepts vs Hedges
+
+Data-plane hedging is triggered by **latency**, not by a status code: the primary request is always sent first, and a hedged request to the next preferred region is only started once the `threshold` (then each `thresholdStep`) elapses without a *final* response.
```
kushagraThapar: **🔴 Blocking** · Correctness: Hedging trigger

**Data-plane fan-out is not triggered only by latency.**

CrossRegionHedgingAvailabilityStrategy.cs:166-224 also advances to the next arm when a non-final result or fault completes before the timer; AvailabilityStrategyUnitTests.cs:606-635 covers fault continuation. The latency-only wording can therefore misstate expected fan-out and diagnostics.

**Suggested wording:**

> The next arm is dispatched when threshold or thresholdStep expires, or sooner when an active arm finishes with a non-final response or faults while the caller token remains active. A final response stops the race.

---

docs/Cross Region Request Hedging.md
```diff
@@ -137,22 +151,65 @@ In the diagnostics data there are three areas of note that will appear when hedg
...
+| 404 (sub-status ≠ 0) | Not Found with a non-zero sub-status (e.g. read-session-not-available / partition-not-served) |
+| 408 | Request Timeout |
+| 410 | Gone |
+| 429 | Too Many Requests (throttled) |
+| 449 | Retry With |
+| 500 | Internal Server Error |
+| 503 | Service Unavailable |
+| Network failure | Connection refused / DNS / TLS errors reaching a region |
+
+> **Note:** These are the codes that cause the SDK to *keep hedging* to other regions. If the last available region also returns a transient code, that response is still returned to the caller as the final result, where it flows into the normal `ClientRetryPolicy`.
```
kushagraThapar: **🔴 Blocking** · Correctness: Retry lifecycle

**The selected transient result does not enter another ClientRetryPolicy pass.**

RequestInvokerHandler.cs:97-105 wraps BaseSendAsync with the availability strategy, while AbstractRetryHandler.cs:22-36 runs retry handling inside each arm. CrossRegionHedgingAvailabilityStrategy.cs:250-305 may then return a non-final response or rethrow an exception. Describing a later retry-policy pass gives callers the wrong retry-count and failure-surface expectations.

**Suggested wording:**

> Each arm runs through its own ClientRetryPolicy before the availability strategy evaluates its terminal result; there is no additional retry-policy pass after the race. If every arm returns a non-final ResponseMessage, the last response is returned; faulted or cancelled arms may instead surface an exception.

---

docs/Cross Region Request Hedging.md
```diff
@@ -200,7 +257,89 @@ This SDK-default hedging strategy is only applied when no client-level `Availabi
...
+2. **Request-level** `RequestOptions.AvailabilityStrategy` (per-request override)
+3. **Client-level** `CosmosClientOptions.AvailabilityStrategy` (applies to all requests)
+4. **SDK Default** (automatically applied when PPAF is enabled and no explicit strategy is configured)
+5. **None** (hedging disabled)
+
+### Operator kill-switch
+
+`disableCrossRegionalHedging` is a Gateway account property that lets operators dynamically turn off cross-region hedging — both the PPAF SDK default and any customer-configured `AvailabilityStrategy` — without rolling PPAF back entirely. The SDK re-evaluates the flag on every account-properties refresh, so toggling it takes effect without a client restart: setting it to `true` suppresses hedging, and setting it back to `false` restores the customer-configured strategy (or rebuilds the SDK default). A client that has explicitly set `CosmosClientOptions.DisablePartitionLevelFailover` opts out of this Gateway-driven override entirely. The same flag also suppresses [metadata hedging](#metadata-hedging).
```
kushagraThapar: **🟢 Suggestion** · API Documentation: Public-surface clarification (non-blocking)

**CosmosClientOptions.DisablePartitionLevelFailover is not customer-configurable.**

Non-blocking clarification: CosmosClientOptions.cs:898-912 declares this property internal, the public contracts contain no entry, and GatewayHedgingOverrideTests.cs:224-246 exercises the internal override. The current sentence may send customers looking for an unavailable option.

**Suggested replacement:**

> The internal compute-gateway ConnectionPolicy override can opt out of this Gateway-driven behavior; this is not a public CosmosClientOptions setting.

---

docs/Cross Region Request Hedging.md
```diff
@@ -200,7 +257,89 @@ This SDK-default hedging strategy is only applied when no client-level `Availabi
...
+
+| Status Code | Sub-status | Meaning |
+| --- | --- | --- |
+| 503 | any | Service Unavailable (control-plane timeouts also surface as 503) |
+| 500 | any | Internal Server Error |
+| 410 | `LeaseNotFound` | Gone — partition moved |
+| 403 | `DatabaseAccountNotFound` | Region unavailable for this account |
+| Network failure | — | Bare connection failure reaching the region's gateway (connection refused / DNS / TLS) |
+
+This is the same status / sub-status set used by the metadata retry policy (`MetadataRequestThrottleRetryPolicy`), so hedging and the retry policy agree on what "the region is at fault" means.
```
kushagraThapar: **🟢 Suggestion** · Documentation: Retry-classifier clarification (non-blocking)

**Metadata hedging and metadata retry do not fully share one classifier.**

Non-blocking clarification: the four HTTP status/sub-status cases align, but MetadataHedgingStrategy.cs:400-435 additionally treats bare HttpRequestException as a regional failure. MetadataRequestThrottleRetryPolicy.cs:95-139 and ResourceThrottleRetryPolicy.cs:47-70 do not give it identical retry behavior, and the two cache paths use different outer retry policies.

**Suggested wording:**

> The four HTTP status/sub-status cases match MetadataRequestThrottleRetryPolicy. Bare HttpRequestException is an additional hedging classification; subsequent retry handling depends on the cache path.

---

docs/Cross Region Request Hedging.md
```diff
@@ -200,7 +257,89 @@ This SDK-default hedging strategy is only applied when no client-level `Availabi
...
+
+```
+HedgeFired={true|false}; HedgeWon={true|false}; WinningRegion={region}
+```
+
+- `HedgeFired` — a hedge request was dispatched (the threshold elapsed, or the primary hit a regional failure).
+- `HedgeWon` — the hedge's response is the one returned (its region differs from the primary). This is also the signal that pins later `PartitionKeyRange` pages to the winning region.
+- `WinningRegion` — the region that produced the returned answer.
+
+The datum is emitted only when a hedge actually fired, so the common no-hedge path leaves the trace unchanged.
```
kushagraThapar: **🟢 Suggestion** · Documentation: Diagnostics clarification (non-blocking)

**Failed metadata races can omit the Metadata Hedge datum.**

Non-blocking clarification: ClientCollectionCache.cs:244-256 and PartitionKeyRangeCache.cs:363-375 add the datum only after ExecuteAsync returns; MetadataHedgingStrategyTests.cs:151-167 verifies that both-branch failure throws. The current paragraph can therefore imply the datum is guaranteed whenever a hedge was dispatched.

**Suggested wording:**

> When ExecuteAsync returns after dispatching a hedge, the call site emits this datum. It is absent when no hedge fires and can also be absent when execution throws before producing a MetadataHedgingResult.

---

docs/Cross Region Request Hedging.md
NaluTripician: Good catch - fixed in ed5c9c2. The intro no longer asserts a shared `primary authoritative` guarantee. It now states data-plane hedging returns the first *final* response from any arm, while only metadata hedging keeps the primary authoritative (a hedge wins only when the primary is slow or has returned a regional failure).

docs/Cross Region Request Hedging.md
NaluTripician: Fixed in ed5c9c2. Reworded to a `latency race` framing: the next arm is dispatched when `threshold`/`thresholdStep` expires **or sooner** when an in-flight arm finishes with a non-*final* response or faults while the caller token is active, and a *final* response stops the race.

docs/Cross Region Request Hedging.md
NaluTripician: Fixed in ed5c9c2. The note now clarifies each arm already runs through its own `ClientRetryPolicy` before the availability strategy evaluates its result, so there is no additional retry-policy pass after the race; a faulted or cancelled final arm may instead surface as an exception.

docs/Cross Region Request Hedging.md
NaluTripician: Fixed in ed5c9c2. Reworded to describe `DisablePartitionLevelFailover` as the internal compute-gateway override and to state explicitly that it is not a public `CosmosClientOptions` setting.

docs/Cross Region Request Hedging.md
NaluTripician: Fixed in ed5c9c2. The claim is now scoped to the four HTTP status/sub-status rows matching `MetadataRequestThrottleRetryPolicy`, and the bare network-failure row is called out as an additional hedging-only classification whose subsequent retry handling depends on the cache path.

docs/Cross Region Request Hedging.md
NaluTripician: Fixed in ed5c9c2. Clarified that the datum is emitted only when a hedge fired **and** the read returns a result, and that it is absent when the race throws before producing a result (for example, when every branch fails).

docs/Cross Region Request Hedging.md
kushagraThapar: Thanks for tightening the distinction between the two strategies. One correctness gap remains: `MetadataHedgingStrategy.ResolveWinnerAsync` can return a successful hedge while the primary is still pending, so the SDK cannot guarantee the same outcome the primary would eventually have produced.

Could we use:

> A completed primary success or definitive non-regional error is authoritative and cannot be overridden. While the primary is still pending, or after it returns a regional failure, a successful hedge may win the race. If neither branch succeeds, the primary outcome is returned.

It would be good to align the summary at line 3 and `metadata-hedging-simple-design.md:31-38` with this distinction too.

docs/Cross Region Request Hedging.md
kushagraThapar: Small follow-up: the detailed explanation at line 177 is now accurate, but this summary row still describes the trigger as latency-only. `CrossRegionHedgingAvailabilityStrategy.ExecuteAvailabilityStrategyAsync` can advance immediately after a non-final response or fault.

Suggested table wording:

> Configured `threshold`/`thresholdStep`, or earlier when an in-flight arm returns a non-final response or faults.

kushagraThapar (dismissed):
LGTM, thanks @NaluTripician

docs/Cross Region Request Hedging.md
NaluTripician: Good catch — updated the summary Trigger row to: `Configured 	hreshold/	hresholdStep, or earlier when an in-flight arm returns a non-final response or faults`. Pushed in dd1f8ea5b.

kushagraThapar (approved):

kundadebdatta (approved):
LGTM

PR #5613, opened 2026-02-11
CrossRegionHedgingAvailabilityStrategy: Fixes `ArgumentNullException` race condition in hedging cancellation

FabianMeiswinkel:
### Bug

Multiple production customers reported unobserved `ArgumentNullException: Value cannot be null. (Parameter 'request')` crashes originating from `CrossRegionHedgingAvailabilityStrategy.RequestSenderAndResultCheckAsync`. The exception was surfaced as a `TaskScheduler_UnobservedTaskException`, crashing the process. The affected code paths were:

- `ContainerCore.ReadItemAsync` → `ReadItemStreamAsync` → `ProcessItemStreamAsync` → `RequestInvokerHandler.SendAsync` → `CrossRegionHedgingAvailabilityStrategy.ExecuteAvailabilityStrategyAsync`

### Root Cause

A **race condition** caused by passing the wrong `CancellationToken` to the sender delegate:

1. `ExecuteAvailabilityStrategyAsync` creates a `hedgeRequestsCancellationTokenSource` (linked to the app-provided CT) to coordinate hedge request lifecycle.
2. `CloneAndSendAsync` clones the request inside a `using` block and calls `RequestSenderAndResultCheckAsync`.
3. **Bug:** `RequestSenderAndResultCheckAsync` called `sender.Invoke(request, cancellationToken)` with the **application-provided `CancellationToken`** — not `hedgeRequestsCancellationTokenSource.Token`.
4. When hedge Region B returned a final result (e.g., 200 OK), `hedgeRequestsCancellationTokenSource.Cancel()` was called.
5. **But the in-flight sender for Region A still held the app CT** (e.g., `CancellationToken.None`), which was **never cancelled**.
6. The `CloneAndSendAsync` `using` block exited, **disposing the cloned request**.
7. The Region A sender continued executing with a reference to the now-disposed request → **`ArgumentNullException: Value cannot be null. (Parameter 'request')`**.

A secondary issue: when the application CT was cancelled (e2e timeout), the hedge timer (linked to app CT) would fire, and the old code would blindly continue the loop attempting to clone and send new requests on a cancelled path.

### Fix

Two changes in `CrossRegionHedgingAvailabilityStrategy.cs`:

**1. Pass `hedgeRequestsCancellationTokenSource.Token` to `sender.Invoke()` instead of the app CT**

[...]

PR #6016, opened 2026-07-13
Cross Region Hedging: Fixes ArgumentNullException when a hedged request is cancelled before dispatch

### LINKED ISSUE #6014: ArgumentNullException("request") from GlobalPartitionEndpointManagerCore during cross-region hedging when a losing hedged request is cancelled before dispatch (PPCB/PPAF enabled)

Confirmed present in 3.58.0 and the code path is unchanged in 3.59.0, 3.60.0, 3.61.0, and on `main`.

**Describe the bug**
When the `CrossRegionHedgingAvailabilityStrategy` is enabled together with Partition-Level Circuit Breaker (PPCB) — or Partition-Level Automatic Failover (PPAF) — a losing hedged request that is cancelled *before it is ever dispatched* throws an unhandled `ArgumentNullException: Value cannot be null. (Parameter 'request')` from `GlobalPartitionEndpointManagerCore.IsRequestEligibleForPartitionFailover`. The caller receives this ANE instead of the expected `CosmosOperationCanceledException`.

Root cause (line numbers = 3.58.0):
1. In `ClientRetryPolicy.ShouldRetryAsync(Exception, CancellationToken)`, the `OperationCanceledException` branch calls `this.partitionKeyRangeLocationCache.IncrementRequestFailureCounterAndCheckIfPartitionCanFailover(this.documentServiceRequest)` (~L147-148) **without a null check**.
2. `this.documentServiceRequest` (field ~L41) is only ever assigned in `OnBeforeSendRequest` (~L206-210), which the pipeline invokes only when a request is actually dispatched.
3. When the hedge cancellation token is already cancelled, `AbstractRetryHandler.ExecuteHttpRequestAsync` throws at `cancellationToken.ThrowIfCancellationRequested()` (~L77) **before** `callbackMethod()` runs (~L78), so `OnBeforeSendRequest` never fires and `documentServiceRequest` stays `null`.
4. That `null` is forwarded into `GlobalPartitionEndpointManagerCore`. The flag gate (`IsPartitionLevelAutomaticFailoverEnabled()` / `IsPartitionLevelCircuitBreakerEnabled()`, ~L342-346) passes because PPCB/PPAF is enabled, so execution reaches `if (request == null) throw new ArgumentNullException(nameof(request));` (~L348-350).

Note: the same field *is* null-guarded elsewhere in `ClientRetryPolicy` (e.g. ~L555), so the missing guard in the OCE branch appears to be an oversight.

**To Reproduce**
- Multi-region Cosmos account with more than one preferred read region.
- `CosmosClient` configured with a `CrossRegionHedgingAvailabilityStrategy`, and Partition-Level Circuit Breaker enabled (`AZURE_COSMOS_PARTITION_LEVEL_CIRCUIT_BREAKER_ENABLED=True`) — or PPAF.
- Issue point reads (`ReadItemAsync` / `ReadItemStreamAsync`) under enough concurrency that hedged requests race across regions.
- Intermittently, when the winning region calls `hedgeRequestsCancellationTokenSource.Cancel()` at the exact moment a losing request is at the top of its retry loop but has not yet dispatched, the ANE is thrown.

This is a timing race, so it is intermittent and hard to reproduce deterministically. Sanitized SDK-side stack trace:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'request')
   at Microsoft.Azure.Cosmos.Routing.GlobalPartitionEndpointManagerCore.IsRequestEligibleForPartitionFailover(DocumentServiceRequest request, Boolean shouldValidateFailedLocation, PartitionKeyRange& partitionKeyRange, Uri& failedLocation)
   at Microsoft.Azure.Cosmos.Routing.GlobalPartitionEndpointManagerCore.IncrementRequestFailureCounterAndCheckIfPartitionCanFailover(DocumentServiceRequest request)
   at Microsoft.Azure.Cosmos.ClientRetryPolicy.ShouldRetryAsync(Exception exception, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.RenameCollectionAwareClientRetryPolicy.ShouldRetryAsync(Exception exception, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.Handlers.AbstractRetryHandler.ExecuteHttpRequestAsync(Func`1 callbackMethod, Func`3 callShouldRetry, Func`3 callShouldRetryException, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.Handlers.AbstractRetryHandler.SendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.RequestHandler.SendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.Handlers.TelemetryHandler.SendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.RequestHandler.SendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.Handlers.DiagnosticsHandler.SendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.RequestHandler.SendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.Handlers.RequestInvokerHandler.BaseSendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.CrossRegionHedgingAvailabilityStrategy.RequestSenderAndResultCheckAsync(Func`3 sender, RequestMessage request, String targetRegionName, CancellationTokenSource hedgeRequestsCancellationTokenSource, ITrace trace)
   at Microsoft.Azure.Cosmos.CrossRegionHedgingAvailabilityStrategy.CloneAndSendAsync(Func`3 sender, RequestMessage request, CloneableStream clonedBody, IReadOnlyCollection`1 hedgeRegions, Int32 requestNumber, ITrace trace, CancellationTokenSource hedgeRequestsCancellationTokenSource)
   at Microsoft.Azure.Cosmos.CrossRegionHedgingAvailabilityStrategy.ExecuteAvailabilityStrategyAsync(Func`3 sender, CosmosClient client, RequestMessage request, CancellationToken applicationProvidedCancellationToken)
   at Microsoft.Azure.Cosmos.Handlers.RequestInvokerHandler.SendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.Handlers.RequestInvokerHandler.SendAsync(String resourceUriString, ResourceType resourceType, OperationType operationType, RequestOptions requestOptions, ContainerInternal cosmosContainerCore, FeedRange feedRange, Stream streamPayload, Action`1 requestEnricher, ITrace trace, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.ContainerCore.ProcessItemStreamAsync(Nullable`1 partitionKey, String itemId, Stream streamPayload, OperationType operationType, ItemRequestOptions requestOptions, ITrace trace, Nullable`1 targetResponseSerializationFormat, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.ContainerCore.ReadItemStreamAsync(String id, PartitionKey partitionKey, ITrace trace, ItemRequestOptions requestOptions, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.ClientContextCore.RunWithDiagnosticsHelperAsync[TResult](...)
   at Microsoft.Azure.Cosmos.ClientContextCore.OperationHelperWithRootTraceAsync[TResult](...)
   --- application code: point read (ReadItemStreamAsync by id + partition key), wrapped in a Polly retry+fallback ---
```

**Expected behavior**
A hedged request that loses the race and is cancelled before dispatch should be handled gracefully — either surfaced as `CosmosOperationCanceledException` (as the `catch (OperationCanceledException) when (hedgeRequestsCancellationTokenSource.IsCancellationRequested)` in `RequestSenderAndResultCheckAsync` intends), or the OCE branch in `ClientRetryPolicy.ShouldRetryAsync` should null-guard `this.documentServiceRequest` before calling `IncrementRequestFailureCounterAndCheckIfPartitionCanFailover`, treating "cancelled before any dispatch" as "no partition to fail over." No `ArgumentNullException` should escape to the caller.

**Actual behavior**
`ArgumentNullException: Value cannot be null. (Parameter 'request')` is thrown from `GlobalPartitionEndpointManagerCore.IsRequestEligibleForPartitionFailover` and propagates out to application code, instead of being converted to a `CosmosOperationCanceledException`. Because it is a non-transient exception, the caller's normal cancellation/transient handling does not apply.

**Environment summary**
SDK Version: 3.58.0 (code path also present in 3.59.0, 3.60.0, 3.61.0, and `main`)
OS Version: Linux (containerized service)

**Additional context**
- Key code pointers (3.58.0):
  - `Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs` — OCE branch forwards `this.documentServiceRequest` without a null check (~L147-148); field declared ~L41; assigned only in `OnBeforeSendRequest` ~L206-210; the same field is null-guarded elsewhere ~L555.
  - `Microsoft.Azure.Cosmos/src/Handler/AbstractRetryHandler.cs` — `ExecuteHttpRequestAsync` throws at `cancellationToken.ThrowIfCancellationRequested()` (~L77) before `callbackMethod()` (~L78), so `OnBeforeSendRequest` never runs; the OCE is then routed to `ShouldRetryAsync` (~L100-107).
  - `Microsoft.Azure.Cosmos/src/Routing/GlobalPartitionEndpointManagerCore.cs` — `IncrementRequestFailureCounterAndCheckIfPartitionCanFailover` (~L211) forwards the request unguarded to `IsRequestEligibleForPartitionFailover` (~L333); flag gate ~L342-346; `throw new ArgumentNullException(nameof(request))` ~L348-350.
  - `Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs` — hedge CTS created ~L144; winner cancels it ~L220/L261; losing request is sent with the hedge token ~L338; the intended `OperationCanceledException` → `CosmosOperationCanceledException` catch ~L353 is never reached because the ANE (not an OCE) is thrown first inside the retry policy.
- On `main`, the method is renamed to `IsRequestEligibleForPartitionOrHubRegionFailover` and the flag gate is widened with a `checkHubRegionOverrideInCache` condition, but the `if (request == null) throw ...` path is unchanged (the null can now be reached in more configurations, not fewer).
- Only reproduces with the hedging availability strategy AND PPCB (or PPAF) enabled. With both flags off, the gate returns `false` and no throw occurs.
- Blast radius is a single operation; the caller observes a non-transient `ArgumentNullException` rather than a cancellation.

NaluTripician:
## Bug

Fixes #6014.

When cross-region hedging runs together with Per-Partition Automatic Failover (PPAF) or the Partition-Level Circuit Breaker (PPCB), a losing hedge arm that is **cancelled before it is ever dispatched** throws a non-transient `ArgumentNullException: Value cannot be null. (Parameter 'request')` from `GlobalPartitionEndpointManagerCore`, instead of the expected `CosmosOperationCanceledException`.

### Root cause

1. The hedge winner calls `hedgeRequestsCancellationTokenSource.Cancel()`.
2. The losing sender hits `cancellationToken.ThrowIfCancellationRequested()` in `AbstractRetryHandler.ExecuteHttpRequestAsync` **before** `callbackMethod()` runs, so `ClientRetryPolicy.OnBeforeSendRequest` never fires and `this.documentServiceRequest` stays `null`.
3. The `OperationCanceledException` branch of `ClientRetryPolicy.ShouldRetryAsync` forwards that `null` into `IncrementRequestFailureCounterAndCheckIfPartitionCanFailover` **without a null check** (unlike the two existing guards in the same class at `HandleUnauthorizedResponse` and `TryMarkEndpointUnavailableForPkRange`).
4. With PPAF/PPCB enabled the flag gate is open, so `GlobalPartitionEndpointManagerCore` reaches `throw new ArgumentNullException(nameof(request))`.
5. The `ArgumentNullException` (not an `OperationCanceledException`) bypasses the hedging strategy's `catch (OperationCanceledException) when (...)` and escapes to the caller.

## Fix

Guard `this.documentServiceRequest` for `null` in the `OperationCanceledException` branch of `ClientRetryPolicy.ShouldRetryAsync`, matching the existing guards in the same class. A cancellation before dispatch means there is no resolved partition/location to attribute a failure to, so the partition-failover bookkeeping is skipped and the cancellation surfaces as `CosmosOperationCanceledException` as intended. The intentional defensive `throw` in `GlobalPartitionEndpointManagerCore` is left untouched.

## Why this matters in GA (frequency)

This is **not** limited to customers who explicitly set the PPCB env var and hand-configure a hedging strategy. In shipped GA code, service-driven PPAF (`DocumentClient` reads `AccountProperties.EnablePartitionLevelFailover`) **implicitly enables both ingredients**: PPAF auto-enables PPCB, and PPAF installs a default cross-region hedging strategy. So any account the service flips PPAF-on has both preconditions present with zero customer opt-in — a growing GA population. Each hit is a non-transient failure of the operation.

## Tests

- Adds `CosmosOperationCancelledBeforeDispatch_WithPartitionFailover_DoesNotThrowArgumentNullException` (read + write DataRows) in `ClientRetryPolicyTests.cs`. It builds a `ClientRetryPolicy` with PPAF/PPCB enabled and calls `ShouldRetryAsync(OperationCanceledException)` **without** calling `OnBeforeSendRequest` (so the request reference stays null).
- Verified as a genuine regression guard: **without** the fix the new cases fail with the exact issue #6014 stack trace (`ArgumentNullException` at `GlobalPartitionEndpointManagerCore` ← `ClientRetryPolicy.ShouldRetryAsync`); **with** the fix all 6 tests pass (2 new + 4 pre-existing `CosmosOperationCancelledExceptionHandelingTests` DataRows — no regression).

## Type of change

## Relationship to #5613

Related but a **different root cause**. #5613 fixed a use-after-dispose race (wrong `CancellationToken` passed to the sender). Its fix — passing the hedge CTS token to the sender — is actually the *precondition* that makes this pre-dispatch cancellation observable, not a fix for it.


Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
```diff
@@ -362,6 +362,139 @@ public void CosmosOperationCancelledExceptionHandelingTests(
...
+            catch (ArgumentNullException ex)
+            {
+                Assert.Fail($"Regression #6014: a cancellation after dispatch must not throw ArgumentNullException (isReadRequest={isReadRequest}). {ex}");
+            }
+
+            // A single cancellation stays below the failover threshold, so the bookkeeping path completes without
+            // requesting a retry; a cancellation must never be retried regardless.
+            Assert.IsFalse(
+                shouldRetryResult.ShouldRetry,
+                $"A cancellation after dispatch must not be retried (isReadRequest={isReadRequest}).");
```
kundadebdatta, line 518: **🟡 Recommendation** · Maintainability: Test Coverage

**This test doesn't verify the branch it documents — it would pass even if the null guard were inverted**

This case is documented as exercising the guard's *true* branch ("runs the real `IncrementRequestFailureCounterAndCheckIfPartitionCanFailover` bookkeeping instead of short-circuiting"), but a single cancellation stays below the failover threshold (10 read / 5 write), so `TryMarkEndpointUnavailableForPartitionKeyRange` never fires and the only assertion here (`ShouldRetry == false`) holds regardless of whether the bookkeeping actually ran.

Concretely: if the production guard were inverted to `this.documentServiceRequest == null &&`, this after-dispatch test would still pass (non-null request → skipped branch → still `ShouldRetry == false`, no throw). The pre-dispatch (null) test would catch that inversion, but this test — the one meant to cover the *non-null* branch — cannot distinguish the true branch from it being skipped. As written it largely duplicates the existing `CosmosOperationCancelledExceptionHandelingTests` with weaker assertions.

Consider asserting the observable side effect the way that existing test does — e.g. reflect on the partition-key-range failover info, or drive `requestThreshold` (10 read / 5 write) cancellations and assert `partitionKeyRangeFailoverInfo` becomes non-null — so the test actually pins the non-null bookkeeping path.

---

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
```diff
@@ -181,7 +181,14 @@ public async Task<ShouldRetryResult> ShouldRetryAsync(
...
                     this.locationEndpoint?.ToString() ?? string.Empty);
 
-                if (this.partitionKeyRangeLocationCache.IncrementRequestFailureCounterAndCheckIfPartitionCanFailover(
+                // A request that is cancelled before it is dispatched (for example a losing cross-region
+                // hedge arm whose token is cancelled at the top of the retry loop) never runs
+                // OnBeforeSendRequest, so this.documentServiceRequest is null. With no dispatched request
+                // there is no resolved partition or location to attribute a failure to, so skip the
+                // partition-failover bookkeeping and let the cancellation surface as
+                // CosmosOperationCanceledException instead of throwing ArgumentNullException. See issue #6014.
+                if (this.documentServiceRequest != null
```
FabianMeiswinkel, line 190: This check is only save if this.documentServiceRequest could never regress from != null to null --> OnBeforeSendRequest is not explicitly preventing this. Please double check that no caller would ever call OnBeforeSendRequest(null) - then add an explicity null check. Only in this case the null check here is safe (same null check is used in otehr places with the same race condition - so, I assume OnBeforeSendRequest is never called with null - but worth making it an explicity contract.

FabianMeiswinkel (dismissed):
LGTM except for one comment. Please address before merge.

ananth7592 (dismissed):
LGTM

Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs
NaluTripician: Good call - made it an explicit contract in 8fab93d3. OnBeforeSendRequest now throws ArgumentNullException up front when 
equest is null.

I checked every production caller before adding it: RequestMessage.OnBeforeSendRequestActions, the two DocumentClient paths, ClientCollectionCache, PartitionKeyRangeCache, and the wrapper retry policies (BulkExecutionRetryPolicy, RenameCollectionAwareClientRetryPolicy, etc.) that just forward the reference - none of them pass null; they all dispatch a fully constructed DocumentServiceRequest. So 	his.documentServiceRequest can only transition null -> non-null and never regress, which is exactly what makes the null guard in the OperationCanceledException path safe.

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/ClientRetryPolicyTests.cs
NaluTripician: Agreed - the after-dispatch case wasn't actually pinning the non-null branch. Strengthened it in 8fab93d3 so it drives the full failover threshold (10 read / 5 write) and asserts the observable side effect the way `CosmosOperationCancelledExceptionHandelingTests` does: `PartitionKeyRangeFailoverInfo` becomes non-null and its `Current` routes to the next region.

I verified it's now a genuine guard by mutation-testing: inverting the production check to `this.documentServiceRequest == null &&` makes the strengthened test fail (`Assert.IsNotNull` on the failover info, both read and write rows), whereas the previous single-cancellation version would still have passed. So it can now distinguish the true branch from a short-circuit.

ananth7592 (approved):

FabianMeiswinkel (approved):
LGTM

kundadebdatta (approved):
LGTM

Issue #5623, opened 2026-02-19
CrossRegionHedgingAvailabilityStrategy: Fix `ArgumentNullException` with a Race Condition in Hedging Cancellation


kundadebdatta:
## Issue Description

Customers reported ArgumentNullException (surfaced as TaskScheduler_UnobservedTaskException) originating from [CrossRegionHedgingAvailabilityStrategy.RequestSenderAndResultCheckAsync](vscode-file://vscode-app/c:/Users/ntripician/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html). Two distinct stack traces were observed:

Stack trace 1 — ArgumentNullException at [RequestMessage.set_Content](vscode-file://vscode-app/c:/Users/ntripician/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) → CheckDisposed():

```
System.ArgumentNullException: Value cannot be null. (Parameter 'source')
   at Microsoft.Azure.Cosmos.RequestMessage.set_Content(Stream value)
   at Microsoft.Azure.Cosmos.RequestMessage.CheckDisposed()
   at Microsoft.Azure.Cosmos.RequestInvokerHandler.BaseSendAsync(RequestMessage request, CancellationToken cancellationToken)
```

Stack trace 2 — NullReferenceException at RequestInvokerHandler.SendAsync:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Azure.Cosmos.Handlers.RequestInvokerHandler.SendAsync(RequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.Cosmos.CrossRegionHedgingAvailabilityStrategy.RequestSenderAndResultCheckAsync(...)
```

Both exceptions are unobserved and only surface through TaskScheduler.UnobservedTaskException — they don't crash the application but pollute logs and indicate a resource lifecycle bug.

## Root Cause
[...]

PR #6017, opened 2026-07-13
Cross Region Hedging: Fixes residual post-cancellation exception leak in hedging strategy

NaluTripician:
## Bug

A losing cross-region hedge arm that is cancelled **after another region has already produced the winning response** can race the disposal of its cloned `RequestMessage` and throw a **non-cancellation** exception — `ArgumentNullException` (`Value cannot be null. (Parameter 'request')`), `ObjectDisposedException`, or `NullReferenceException` — from deep in the request pipeline.

`CrossRegionHedgingAvailabilityStrategy` only normalized `OperationCanceledException` (`catch (OperationCanceledException) when (hedgeRequestsCancellationTokenSource.IsCancellationRequested)`). Any other exception type skipped that filter, hit the generic `catch (Exception)`, and either **escaped to the caller** as a hard operation failure or **crashed the process** as a `TaskScheduler.UnobservedTaskException` (the failure mode originally reported for this strategy).

This is most relevant for accounts where hedging is implicitly enabled by **Per-Partition Automatic Failover (PPAF)** — service-driven PPAF auto-enables PPCB and installs a default hedging strategy, so a stock `CosmosClient` has the preconditions present with **zero customer opt-in**.

## Fix (defense in depth, boundary-level)

In `CrossRegionHedgingAvailabilityStrategy.cs`:

- **Normalize** any post-cancellation exception from a losing hedge — add `catch (Exception ex) when (hedgeRequestsCancellationTokenSource.IsCancellationRequested)` (before the generic catch) that throws `CosmosOperationCanceledException`. Because the hedge CTS is already cancelled, the winner has returned and this is cancellation noise.
- **Observe abandoned hedge tasks** — new `ObserveAbandonedHedgeTasks(...)` helper is called before each early winner-return so faulted losers can never surface as an unobserved task exception.
- **Prefer cancellation** in the final `lastException` throw when the app token was cancelled.
- **Telemetry** — a distinctive `DefaultTrace.TraceWarning` records each absorbed post-cancellation exception so the residual rate is measurable in production.

## Tests

4 new regression tests in `AvailabilityStrategyUnitTests.cs`:
- `PostCancellationArgumentNullException_NormalizedToCosmosOperationCanceled` (deterministic; fails without the fix)
- `PostCancellationObjectDisposedException_NormalizedToCosmosOperationCanceled`
- `HedgeLoserThrowsArgumentNullExceptionAfterWinner_ReturnsWinnerWithoutCrash`
- `ConcurrentHedging_LosersThrowArgumentNull_NoNullRefEscapes` (50-way soak)

All 20 tests in `AvailabilityStrategyUnitTests` pass (16 existing + 4 new); Release build clean.

## Relationship to #6016 (no conflict — complementary)

[#6016](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/6016) (issue [#6014](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/6014)) fixes the **specific source path**: a null-guard in `ClientRetryPolicy.ShouldRetryAsync` so a request cancelled *before dispatch* never forwards a `null` into the partition-failover check.

This PR is the **boundary safety-net**. They touch **disjoint files** (`ClientRetryPolicy.cs`/`ClientRetryPolicyTests.cs` vs `CrossRegionHedgingAvailabilityStrategy.cs`/`AvailabilityStrategyUnitTests.cs`); the only shared file is `changelog.md`, where this entry was positioned to auto-merge. A `git merge-tree` simulation of this branch against `#6016` reports a **clean auto-merge (no conflicts)**.

Beyond #6016, this PR additionally covers:
- **`ObjectDisposedException`** from the raced disposed-clone path (a different exception than #6016 addresses),
- any **other pipeline null path** not covered by the single source guard, and
- the **unobserved-task-exception crash mode** (via abandoned-task observation), which #6016 does not address.

## Type of change


NaluTripician:
## 🔬 Performance benchmark: this PR vs `main` — no regression

To de-risk merging, I benchmarked this change against `main` under a steady-state read/write workload with **cross-region hedging enabled**, including a mid-run **fault-injection delay phase** that forces hedging to fail over (exactly the losing-arm cancellation path this PR touches).

- **baseline** = `main` @ `fecf00e07` (no fix)  •  **change** = this PR @ `4eebb18`
- Live 3-region account (West US 2 / East US 2 / East US), Direct mode, `AvailabilityStrategy.CrossRegionHedgingStrategy(threshold=100ms, step=50ms)`
- Same harness compiled against each SDK build; identical workload; runs executed **sequentially** on one host to avoid noisy-neighbor contention.
- Phases: **warmup 15s → steadyA 45s → fault 45s → recovery 45s**. The **fault** phase injects a `2000ms` `ResponseDelay` on **primary-region reads**, so the hedge to East US 2 must win.

### Overview
dashboard

### Result: statistically indistinguishable from `main`

| Phase | Metric | main | change | Δ |
|---|---|---:|---:|---:|
| steadyA | read ops/s (mean) | 1778 | 1790 | +0.6% |
| steadyA | read p99 (ms) | 62.9 | 66.0 | +4.9% |
| steadyA | CPU % (mean) | 11.7 | 13.0 | +1.3 pp |
| steadyA | WorkingSet MB (mean) | 155 | 152 | −1.6% |
| **fault** | read ops/s (mean) | 209 | 207 | −0.5% |
| **fault** | read p99 (ms, mean) | 228.9 | **215.6** | −5.8% |
| **fault** | read p99 (ms, max) | 253.8 | 257.2 | +1.3% |
| **fault** | CPU % (mean) | 6.0 | 6.5 | +0.5 pp |
| **fault** | WorkingSet MB (max) | 578 | **564** | −2.4% |
| recovery | read ops/s (mean) | 1784 | 1790 | +0.4% |
| recovery | CPU % (mean) | 12.9 | 12.6 | −2.2% |

All deltas are within run-to-run noise. `errors = 0` and caller-visible `cancellations = 0` throughout on both builds. The extra `ObserveAbandonedHedgeTasks` continuation this PR registers per losing arm (~7,800 fault-phase hedge hits/run) adds **no measurable CPU or allocation cost**.

### Hedging still works identically
During the fault phase, read p99 stays **bounded at ~215–230 ms** (hedge threshold + East US 2 RTT) instead of the injected 2000 ms — the delayed primary (West US 2) loses, the hedge wins — on both builds equally.

| Read throughput | Read p99 latency |
|---|---|
| throughput | read p99 |

| Process CPU | Working set |
|---|---|
| cpu | memory |

### Unobserved task exceptions
unobserved

Both builds record exactly **1** `TaskScheduler.UnobservedTaskException`, appearing at ~111–113 s (early recovery, right after the 2 s delay is lifted). It originates from the **last batch of fault-phase primary reads** whose abandoned, still-delayed server response faults in the closed-source `Microsoft.Azure.Cosmos.Direct` transport layer — *below* the hedging strategy's own task tracking, which no managed change in this repo can observe. Because it is **identical on both builds**, this PR does not regress the unobserved-exception count.

> Note: this workload never cancels the *application* token (ops use `CancellationToken.None`); losing arms are cancelled by the internal **hedge** CTS. The strategy-level leak this PR closes is proven deterministically by the unit test `LosingHedgeFaultsAppTokenNotCancelled_NoUnobservedTaskException` (which fails without the fix). This benchmark is the **performance-regression** check, and it is clean.

### Memory footnote
Working set climbs on both during the fault phase (2 s-delayed in-flight responses buffer), peaks ~560–640 MB, then releases. In the tail the `change` process happened to hold more before GC released it — a Server-GC lazy-release timing artifact (Server GC was enabled to mimic production), **not a leak**: peak working set during the fault was actually *lower* for `change` (564 vs 578 MB).

---

**Conclusion: no throughput, latency, CPU, memory, or unobserved-exception regression — this PR is safe to merge from a performance standpoint.**

<sub>Raw CSVs, per-run JSON summaries, all graphs, and the full write-up are on branch [`users/ntripician/hedging-benchmark-6017-results`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/users/ntripician/hedging-benchmark-6017-results/benchmarks/6017) (results-only; not part of this PR's diff — safe to delete after review). Harness: 80/20 read/write, concurrency 32, 1000 docs, 10000 RU. 🤖 Generated with GitHub Copilot CLI.</sub>

NaluTripician:
Closing in favor of #6030, which is the reduced-scope replacement for this PR.

#6030 keeps only the genuinely-helpful, deterministically-testable part of this PR — observing abandoned losing hedge tasks so a faulted loser can't surface as a `TaskScheduler.UnobservedTaskException` (issue #5623) — and intentionally drops the speculative parts here:
- the broad `catch (Exception) when (hedgeRequestsCancellationTokenSource.IsCancellationRequested)` normalization to `CosmosOperationCanceledException` (it fires during the drain window of every successful hedge and would silently reclassify genuine non-cancellation defects), and
- the `lastException` cancellation-preference block.

The concrete exceptions those dropped parts targeted are already root-fixed on `main`: the pre-dispatch `ArgumentNullException("request")` by #6016 (issue #6014), and the original disposed-clone race by #5613 (issue #5623).

PR #6020, opened 2026-07-14
[Internal] Routing: Refactors partition-move cache refresh into a shared primitive

Meghana-Palaparthi:
## Summary

Extracts the duplicated "partition moved -> force-refresh routing cache" detection into a single shared primitive, `PartitionKeyRangeCache.RefreshRoutingCacheIfPartitionMovedAsync`, so the gateway point-op path and the upcoming distributed-transaction (DTX) post-commit path share one definition instead of two comment-enforced mirrors.

The primitive:
- Guards out no-op cases (null cache, empty collection rid, empty client range id, whitespace server range id, ids equal under `OrdinalIgnoreCase`).
- Optionally dedupes repeated moves to the same range via a caller-supplied set (consulted only after a move is confirmed).
- Returns the un-awaited refresh task, so callers can choose to let a refresh failure propagate (point ops, for idempotent retry) or swallow it (DTX).

The point-op call site in `GatewayStoreModel` now delegates to this primitive.

## Notes for reviewers

**`refreshedRanges` param has no production caller on this branch.** The optional dedupe set is exercised here by tests only — the point-op path calls the primitive passing no set. Its production consumer is the DTX post-commit merge path, which lands in the immediately-following DTX PR.

**Not a byte-for-byte refactor — two intentional, benign guard tightenings vs. the old inline point-op code:**
- The client-range guard moved from object-null (`detectedPartitionKeyRange != null`) to id-empty (`IsNullOrEmpty(ResolvedPartitionKeyRange?.Id)`).
- The server-range check moved from `IsNullOrEmpty` to `IsNullOrWhiteSpace`.

Both make a spurious force-refresh less likely and won't occur in practice.

## Testing

Added focused unit tests in `PartitionKeyRangeCacheTest.cs` covering: move detected/not-detected, each guard, dedupe collapse, null-set repeated-move double refresh, and refresh-failure propagation (returned task faults). 41 tests pass.

## Changelog

No entry required — pure internal refactor with no customer-observable change.

### DIFF
--- a/Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
+++ b/Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
@@ -89,6 +89,61 @@ namespace Microsoft.Azure.Cosmos.Routing
             }
         }
 
+        /// <summary>
+        /// Force-refreshes the routing cache for a single collection when the partition key range the server
+        /// actually served differs from the range the client resolved for the request — i.e. the partition
+        /// moved (split/merge) to a range the client did not know about. If the ids match, or any input is
+        /// missing, this is a no-op.
+        /// </summary>
+        /// <remarks>
+        /// This is the shared move-detection primitive for both the point-operation capture path
+        /// (<see cref="GatewayStoreModel.CaptureSessionTokenAndHandleSplitAsync"/>) and the distributed-transaction
+        /// post-commit merge path, so the "did the partition move?" decision has exactly one definition instead of
+        /// two comment-synchronized copies. Callers own their own gating policy (which responses capture) and
+        /// failure policy (point ops let a refresh failure propagate for retry; DTX swallows it because the commit
+        /// already succeeded). <paramref name="refreshedRanges"/> is an optional per-pass dedupe set for callers
+        /// that process many operations at once (e.g. a distributed transaction with N sub-ops) so repeated moves
+        /// to the same range trigger a single refresh; it is only consulted after a move is confirmed.
+        /// </remarks>
+        public static Task RefreshRoutingCacheIfPartitionMovedAsync(
+            PartitionKeyRangeCache partitionKeyRangeCache,
+            string collectionResourceId,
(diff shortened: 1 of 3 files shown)


Microsoft.Azure.Cosmos/src/Routing/PartitionKeyRangeCache.cs
```diff
@@ -89,6 +89,61 @@ public virtual async Task<IReadOnlyList<PartitionKeyRange>> TryGetOverlappingRan
...
+        /// that process many operations at once (e.g. a distributed transaction with N sub-ops) so repeated moves
+        /// to the same range trigger a single refresh; it is only consulted after a move is confirmed.
+        /// </remarks>
+        public static Task RefreshRoutingCacheIfPartitionMovedAsync(
+            PartitionKeyRangeCache partitionKeyRangeCache,
+            string collectionResourceId,
+            string clientResolvedPartitionKeyRangeId,
+            string serverServedPartitionKeyRangeId,
+            ITrace trace,
+            HashSet<string> refreshedRanges = null)
```
ananth7592: If the intent is a de-dup that's helpful in DTX with N sub-ops at once, then plain `HashSet`  isn't safe for concurrent writes. If DTX fans sub-op responses out in parallel, two calls with the same shared set race on Add — which can corrupt internal state, not just miscount. 

Some of the options that can be explored are

1. Dedupe at the caller (Stateless in this primitive in the cache)
The DTX caller — which already owns the loop over its N sub-ops — dedupes before calling: collect the distinct (collectionRid, serverRangeId) pairs that actually moved, then call the primitive once per distinct pair. Dedup logic lives where the concurrency model is known, and the shared primitive carries no mutable state at all.

2.  Maybe a `ConcurrentDictionary<string, byte>` and use `TryAdd`
3. Lock around the `add` - least preferred

Meghana-Palaparthi: Applied option 1.

kushagraThapar (approved):

ananth7592 (approved):
LGTM

PR #6023, opened 2026-07-16
LINQ: Fixes Take() mutating MaxItemCount on QueryRequestOptions

### LINKED ISSUE #5225: LINQ Take() silently changes MaxItemCount on QueryRequestOptions object
**Describe the bug**
Using the LINQ `Take()` method silently updates `QueryRequestOptions.MaxItemCount`. This causes issues if you are caching the options object across calls, and is not obvious to consumers.

**To Reproduce**
```csharp
public class Post
{
        public required Guid Id { get; set; }
        public required string User { get; set; }
        public required string Content { get; set; }
}

public class DbService(CosmosClient client)
{
    private readonly QueryRequestOptions options = new QueryRequestOptions { MaxItemCount = -1 };

    public async Task<List<Post>> GetPostsAsync(string user, int? take = null) 
    {
        var query = client.GetContainer("post-container");
        container.GetItemLinqQueryable<Item>(requestOptions: options)
            .Where(item => item.User == user);
        
        if (take is not null) 
        {
            query = query.Take(take.Value);
        }
    
        var results = new List<Item>();
        var iterator = query.ToFeedIterator();
        while (iterator.HasMoreResults)
        {
            var next = await iterator.ReadNextAsync();
            results.AddRange(next);
        }
        return results;
    }
}

```

Now in another class:

```csharp
public class MyClass(DbService database)
{
    public async Task CreateError()
    {
        // This runs fine and returns all posts quickly
        var posts = await database.GetPostsAsync("Jonathan");

        // This runs fine and returns a single post
        posts = await database.GetPostsAsync("Jonathan", take: 1);

        // This runs very slowly because items.MaxItemCount is now 1, and a separate request is made to Cosmos for each item
        posts = await database.GetPostsAsync("Jonathan");
    }
}
```

**Expected behavior**
I would expect to be able to cache an `QueryRequestOptions` object across calls without fear of it causing issues.

**Actual behavior**
`MaxItemCount` gets silently change to the value passed to `Take()`, meaning if you are caching your `options` objects, you are open to this issue. Although maybe not a true bug, this is definitely non-obvious behavior and caused a week of frustration tacking down a horrendous performance bug in our production codebase.

**Environment summary**
SDK Version: 3.51.0
OS Version: Windows

Maya-Painter:
## Summary

When using `Take` in a SQL query the query pipeline computes an optimal page size and writes it back to the user's `QueryRequestOptions.MaxItemCount` property. This silently mutates the caller's object as a side effect, which can cause unexpected behavior when the same `QueryRequestOptions` instance is reused across queries.

## Changes

**Production fix** (`CosmosQueryClientCore.cs`):
- Shallow-copy the `QueryRequestOptions` before setting `MaxItemCount = pageSize`, so the user's original object is never modified.

**Tests**:
- `CosmosItemLinqTests.cs`: Adds `LinqTakeDoesNotMutateMaxItemCount`, an emulator integration test that runs `.OrderBy().Take(2).ToFeedIterator()` with `MaxItemCount = 100` and verifies the original value remains unchanged.
- `SanityQueryTests.cs`: Adds non-LINQ emulator coverage for both typed `GetItemQueryIterator<T>` and stream `GetItemQueryStreamIterator`. Each runs `SELECT TOP 2 ... ORDER BY` with `MaxItemCount = 100`, which exercises the optimized internal page-size path and verifies the caller's request options are not mutated. The typed test drains all pages because cross-partition ordered queries may return partial pages.

**Changelog** (`changelog.md`):
- Added entry under `Bugs Fixed` in the Unreleased section.

Fixes #5225


Microsoft.Azure.Cosmos/src/Query/v3Query/CosmosQueryClientCore.cs
```diff
@@ -143,6 +143,7 @@ public override async Task<TryCatch<QueryPage>> ExecuteItemQueryAsync(
             ITrace trace,
             CancellationToken cancellationToken)
         {
+            requestOptions = (QueryRequestOptions)requestOptions.ShallowCopy();
```
adityasa, line 146: Does this bug repro in non-LINQ case as well ? If so, can you add coverage?

Maya-Painter: Added tests to SanityQueryTests.cs

neildsh (approved):

adityasa (approved):
:shipit:

sboshra (approved):

PR #6025, opened 2026-07-16
Use Assembly.GetType instead of Type.GetType to avoid TypeResolve event in AppDomain-isolated hosts

### LINKED ISSUE #6024: Type.GetType in CosmosHttpClientCore.CreateHttpClientHandler triggers TypeResolve event, causing RemotingException in AppDomain-isolated test hosts
## Problem

`CosmosHttpClientCore.CreateHttpClientHandler` and `CreateSocketsHttpHandlerHelper` use `Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http")` to detect at runtime whether `SocketsHttpHandler` is available. On .NET Framework, this type doesn't exist, so `Type.GetType` fires the `AppDomain.TypeResolve` event before returning `null`.

This causes `System.Runtime.Remoting.RemotingException` when:
1. Tests run under `vstest.console.exe /InIsolation` (creates a child AppDomain for the test assembly)
2. A code coverage data collector (`DynamicCoverageDataCollector`) registers `MarshalByRefObject` proxies across the AppDomain boundary for `TypeResolve`/`AssemblyResolve` events
3. Those proxies have the default 5-minute lease lifetime
4. After 5 minutes, any new `CosmosClient` creation calls `Type.GetType` → fires `TypeResolve` → hits the expired proxy → `RemotingException`

```
System.Runtime.Remoting.RemotingException: Object '/xxx.rem' has been disconnected or does not exist at the server.
```

This is intermittent because it only affects `CosmosClient` instances created after the 5-minute mark. Cached clients (via connection pooling or application-level caching) don't trigger the issue since they don't call `CreateHttpClientHandler` again.

## Root Cause

`Type.GetType(string)` goes through the full runtime type resolution pipeline, including firing `AppDomain.TypeResolve` when the type isn't found. This is by design — it gives registered handlers a chance to resolve the type dynamically.

The current code:
```csharp
// In CreateHttpClientHandler (line ~197) and CreateSocketsHttpHandlerHelper (line ~218)
Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
```

The existing TODO comment acknowledges this should change:
```csharp
// TODO: Remove type check and use #if NET6_0_OR_GREATER when multitargetting is possible
```

## Proposed Fix

Replace `Type.GetType` with `Assembly.GetType`, which does a direct metadata lookup within the already-loaded assembly — no `TypeResolve` event, no cross-AppDomain remoting calls:

```csharp
Type socketHandlerType = typeof(HttpClient).Assembly.GetType("System.Net.Http.SocketsHttpHandler");
```

This is semantically equivalent:
- Returns the `Type` on .NET Core/.NET 5+ (where `SocketsHttpHandler` exists in `System.Net.Http.dll`)
- Returns `null` on .NET Framework (where it doesn't) — same as today, but without firing `TypeResolve`
- `typeof(HttpClient).Assembly` reliably references the `System.Net.Http` assembly since `HttpClient` is always available

## Impact

Any consumer of the Cosmos SDK running on .NET Framework in an AppDomain-isolated host (test runners with `/InIsolation`, plugin hosts, etc.) that creates new `CosmosClient` instances after event handler proxies expire will hit this crash. The workaround is to set infinite lease lifetime via `app.config`:

```xml
<system.runtime.remoting>
  <application>
    <lifetime leaseTime="0" renewOnCallTime="0" />
  </application>
</system.runtime.remoting>
```

But the proper fix is to avoid firing `TypeResolve` in the first place.

## Environment

- Microsoft.Azure.Cosmos NuGet package (netstandard2.0 build)
- .NET Framework 4.7.2 runtime
- vstest.console.exe with `/InIsolation` and code coverage enabled

marcre:
## Changes

Replaces Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http") with 	ypeof(HttpClient).Assembly.GetType("System.Net.Http.SocketsHttpHandler") in two locations in CosmosHttpClientCore.cs:
- CreateHttpClientHandler (line ~120)
- CreateSocketsHttpHandlerHelper (line ~143)

## Why

Type.GetType(string) fires AppDomain.TypeResolve when the type isn't found (which is the case on .NET Framework since SocketsHttpHandler doesn't exist there). In AppDomain-isolated hosts (e.g. stest.console.exe /InIsolation with code coverage), cross-domain MarshalByRefObject proxies registered for TypeResolve can expire after 5 minutes, causing RemotingException on subsequent CosmosClient creation.

Assembly.GetType(string) performs a direct metadata lookup within the already-loaded assembly — no TypeResolve event, no cross-AppDomain calls, no lease expiry risk. It returns 
ull when the type doesn't exist, same as before.

## Behavioral equivalence

- On .NET Core/.NET 5+: Returns the SocketsHttpHandler type (same as before)
- On .NET Framework: Returns null (same as before, but without side effects)
- 	typeof(HttpClient).Assembly reliably references System.Net.Http.dll on all platforms

Fixes #6024

### DIFF
--- a/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
+++ b/Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Azure.Cosmos
             Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
         {
             // TODO: Remove type check and use #if NET6_0_OR_GREATER when multitargetting is possible
-            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+            Type socketHandlerType = typeof(HttpClient).Assembly.GetType("System.Net.Http.SocketsHttpHandler");
 
             if (socketHandlerType != null)
             {
@@ -140,7 +140,7 @@ namespace Microsoft.Azure.Cosmos
             Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
         {
             // TODO: Remove Reflection when multitargetting is possible
-            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+            Type socketHandlerType = typeof(HttpClient).Assembly.GetType("System.Net.Http.SocketsHttpHandler");
 
             object socketHttpHandler = Activator.CreateInstance(socketHandlerType);
 
--- a/changelog.md
+++ b/changelog.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Bugs Fixed
 
 - [6032](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/6032) ThinClient: Fixes clients using resource-token (permission-scoped) authorization failing or misrouting when thin client mode is enabled by default. Such clients now always route data-plane requests through the Gateway store model, since thin client mode does not support resource-token authorization. Clients using primary/secondary key or Microsoft Entra ID (AAD) authorization are unaffected.
+- [6025](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/6025) Fixed `RemotingException` in AppDomain-isolated test hosts by replacing `Type.GetType` with `Assembly.GetType` in `CosmosHttpClientCore.CreateHttpClientHandler` and `CreateSocketsHttpHandlerHelper`. `Type.GetType` fires `AppDomain.TypeResolve` when the type is not found on .NET Framework, which crashes if cross-domain `MarshalByRefObject` proxies have expired leases.
 
 #### Other Changes
 


aavasthy (dismissed):
LGTM

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
```diff
@@ -117,7 +117,7 @@ public static HttpMessageHandler CreateHttpClientHandler(
             Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> serverCertificateCustomValidationCallback)
         {
             // TODO: Remove type check and use #if NET6_0_OR_GREATER when multitargetting is possible
-            Type socketHandlerType = Type.GetType("System.Net.Http.SocketsHttpHandler, System.Net.Http");
+            Type socketHandlerType = typeof(HttpClient).Assembly.GetType("System.Net.Http.SocketsHttpHandler");
```
kundadebdatta, line 120: nit: TBH, since there is only one caller for `CreateSocketsHttpHandlerHelper` (the rest of them are just unit tests), the resolved type `socketHandlerType` can be passed down to the `CreateSocketsHttpHandlerHelper`, so that the helper doesn't need to re-resolve the type. This is a minor clean up, but not a blocker.

kundadebdatta (dismissed):
One minor nit. LGTM otherwise. Please add changelog entry as well.

Microsoft.Azure.Cosmos/src/HttpClient/CosmosHttpClientCore.cs
marcre: I decided to go for minimal change - if I shifted that logic into unit tests then they'd have more of the business logic there.

marcre:
Updated with changelog entry and replied to comment.

changelog.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Bugs Fixed
 
+- [6024](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/6024) Fixed `RemotingException` in AppDomain-isolated test hosts by replacing `Type.GetType` with `Assembly.GetType` in `CosmosHttpClientCore.CreateHttpClientHandler` and `CreateSocketsHttpHandlerHelper`. `Type.GetType` fires `AppDomain.TypeResolve` when the type is not found on .NET Framework, which crashes if cross-domain `MarshalByRefObject` proxies have expired leases.
```
kushagraThapar: we usually put the PR link that fixes the issue, so we should update it. 
```suggestion
- [6025](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/6025) Fixed `RemotingException` in AppDomain-isolated test hosts by replacing `Type.GetType` with `Assembly.GetType` in `CosmosHttpClientCore.CreateHttpClientHandler` and `CreateSocketsHttpHandlerHelper`. `Type.GetType` fires `AppDomain.TypeResolve` when the type is not found on .NET Framework, which crashes if cross-domain `MarshalByRefObject` proxies have expired leases.
```

kundadebdatta (dismissed):
LGTM

changelog.md
marcre: Fixed

kushagraThapar (approved):
LGTM, thanks @marcre

kushagraThapar (approved):

aavasthy (approved):

kundadebdatta (approved):
LGTM

PR #6030, opened 2026-07-20
Cross Region Hedging: Fixes unobserved task exception from abandoned losing hedge requests

NaluTripician:
## Bug

When a cross-region hedge produces a winning response, the strategy returns immediately and the still in-flight **losing** hedge tasks are abandoned in `requestTasks` without being awaited. A losing arm that completes **faulted** (for example after being cancelled by the winner — `RequestSenderAndResultCheckAsync` rethrows the `OperationCanceledException` as `CosmosOperationCanceledException`, so the task ends **Faulted**, not Canceled) is never observed and can surface as a `TaskScheduler.UnobservedTaskException` (issue [#5623](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5623)). There is also a phase-1 branch where a faulted loser is removed from `requestTasks` before any later observation could see it.

This path is exercised with **zero customer opt-in** on accounts where Per-Partition Automatic Failover (PPAF) implicitly enables hedging.

## Fix (minimal, testable)

In `CrossRegionHedgingAvailabilityStrategy.cs`:

- Add `ObserveAbandonedHedgeTasks(IEnumerable<Task>)` — marks already-faulted abandoned tasks observed, and attaches an `OnlyOnFaulted` continuation to still-running ones.
- Call it at each early winner-return (phase-1 and phase-2) and in the phase-1 faulted-loser branch (where the loser was already removed from `requestTasks`, so no later call could observe it).

No behavior change for callers: the winner's response is still returned; genuine, non-cancellation errors still propagate unchanged.

## Type of change


NaluTripician:
## 🔬 Performance benchmark: PR #6030 vs `main` — no regression

Ran the same live-account A/B harness used for #6017, **twice**, to de-risk merging. Because this PR's parent commit is exactly `origin/main`, the two builds differ by **precisely the one commit under review** (`ObserveAbandonedHedgeTasks` only).

- **baseline** = `860af8d44` (origin/main)  •  **change** = `151bd55ae` (this PR head)
- Live 3-region account (**West US 2 / East US 2 / East US**), Direct mode, `CrossRegionHedgingStrategy(threshold=100ms, step=50ms)`
- Same harness compiled against each build; identical workload; runs executed **sequentially** on one host.
- Phases: **warmup 15s → steadyA 45s → fault 45s → recovery 45s**. The **fault** phase injects a `2000ms` `ResponseDelay` on **West US 2 reads**, so the East US 2 hedge must win (exactly the losing-arm path this PR touches).

### Result: statistically indistinguishable from `main`

`errors = 0`, `cancellations = 0`, and **`TaskScheduler.UnobservedTaskException = 0`** on **both builds in both runs**. Every larger-looking delta **flips sign between the two runs** — the signature of run-to-run jitter, not a systematic regression:

| Metric (change − main) | Run 1 | Run 2 |
|---|---:|---:|
| steadyA read ops/s (mean) | +5.1% | +1.3% |
| steadyA read p99 **max** | +79.3% | **−22.3%** |
| fault read ops/s (mean) | −1.7% | −2.0% |
| fault read p99 (mean) | +3.3% | +0.8% |
| fault read p99 **max** | +94.7% | **+0.3%** |
| recovery read ops/s (mean) | −2.3% | −12.8% |
| recovery read p99 (mean) | **−20.0%** | +37.7% |
| WorkingSet MB (max, fault) | +8.4% | +3.4% |
| Unobserved task exc | 0 / 0 | 0 / 0 |

The p99 **max** columns are single worst-second samples dominated by GC/JIT + network jitter at fault onset: in run 1 the spike landed on `change`, in run 2 it landed on `main` (416ms) with `change` essentially identical (418ms). The recovery working-set/p99 differences are Server-GC lazy-release timing artifacts — peak working set is comparable (slightly **lower** on `change` in run 2: 656 vs 678 MB).

### Hedging still works identically
During the fault phase read p99 stays **bounded at ~200 ms** (hedge threshold + East US 2 RTT) instead of the injected 2000 ms — the delayed primary (West US 2) loses, the hedge wins — on both builds equally.

### Unobserved task exceptions
**0 on both builds in both runs** — this PR's `ObserveAbandonedHedgeTasks` (~7,870 fault-phase hedge hits/run) adds no measurable CPU/allocation cost. (The strategy-level leak this PR closes is proven deterministically by the unit test `LosingHedgeFaultsAppTokenNotCancelled_NoUnobservedTaskException`, which fails without the fix; this benchmark is the **performance-regression** check.)

| Run 1 dashboard | Run 2 dashboard |
|---|---|
| run1 | run2 |

> Note: the dashboard title says "PR #6017" because `plot_compare.py` hard-codes it — the harness/methodology are identical and the data is for **#6030**.

---

**Conclusion: no throughput, latency, CPU, memory, or unobserved-exception regression — safe to merge from a performance standpoint.**

<sub>Raw CSVs, per-run JSON summaries, all graphs, and the write-up are on branch [`users/ntripician/hedging-benchmark-6030-results`](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/users/ntripician/hedging-benchmark-6030-results/benchmarks/6030) (results-only; not part of this PR's diff). 🤖 Generated with GitHub Copilot CLI.</sub>

NaluTripician:
### Follow-up: why does working-set memory look so different between run 1 and run 2?

Short version: the big swing isn't in **peak** working set and isn't between this PR and `main` — it's in the **recovery-phase _mean_ working set**, and it's a **.NET Server-GC lazy-decommit timing artifact**, not real memory growth from `ObserveAbandonedHedgeTasks`.

Pulling the absolute numbers from the per-run `*.summary.json` files (MB):

| Phase / metric | Run 1 `main` | Run 2 `main` | Run 1 `change` | Run 2 `change` |
|---|--:|--:|--:|--:|
| fault **mean** ws | 369 | 379 | 386 | 392 |
| fault **max** ws | 537 | 535 | 582 | 552 |
| recovery **max** ws | 662 | 678 | 650 | 656 |
| **recovery mean ws** | **346** | **521** | **559** | **506** |
| GC gen2 count | 13 | 14 | 14 | 15 |

Everything is tight across runs **except `recovery mean ws`**, where the byte-identical `main` binary reads **346 MB in run 1 but 521 MB in run 2** (+50% / +175 MB). That within-build, run-to-run gap is the "massive difference" — not a PR-vs-`main` effect.

**Why it happens**

1. **Fault phase = allocation burst.** With West US 2 reads delayed 2000 ms, nearly every read fires a hedge to East US 2 (~7,800 hedge hits/run) — a spike of `Task` / `CancellationTokenSource` / continuation allocations that drives the heap (and OS working set) to its peak (~535–680 MB).
2. **Working set ≠ live heap.** Working set is the physical RAM the OS currently maps to the process. Under **Server GC** (default here), freed segments are **not** promptly returned to the OS — decommit is lazy and governed by background-GC / RetainVM heuristics.
3. **Recovery just measures _when_ the GC hands pages back.** After fault ends, allocation drops to steady-state but working set decays slowly and non-deterministically. Whether that decay lands inside the fixed 45 s recovery window is timing luck: run 1 `main` trimmed early (mean 346), run 2 `main` held the pages most of the window (mean 521).

**Why it's noise, not a regression**

- **Peak (max) working set is stable everywhere** — fault ~535–582, recovery ~650–678 — regardless of run or build.
- **Gen2 GC counts are nearly identical** (13–15 across all four), so allocation/collection behavior is unchanged; the PR isn't retaining more.
- **The change−main deltas flip sign between runs** (recovery is actually slightly _lower_ on `change` in run 2: 506 vs 521) — the signature of jitter.

Net: the recovery mean-working-set gap is Server GC deciding to release physical pages a few seconds earlier in one run than the other; the `ObserveAbandonedHedgeTasks` change adds no measurable memory.

<sub>🤖 Generated with GitHub Copilot CLI.</sub>

kundadebdatta:
## PR Deep Review summary

Reviewed as a cross-region hedging concurrency fix. **Overall: a solid, well-targeted defensive fix — no blocking issues.** The `ObserveAbandonedHedgeTasks` helper is correct and idiomatic, concurrency is safe, and the change is genuinely complementary to the root-cause fixes #5613 / #6016 (not redundant). Dropping #6017's speculative broad post-cancellation catch is good judgment, locked in by the `GenuineError...` guard test.

Posting the **4 higher-value (Recommendation) findings** as inline comments:

1. **Correctness of the rationale** — the code comments and changelog say a loser "cancelled by the winner" completes *faulted*, but `CosmosOperationCanceledException : OperationCanceledException` means such a task completes *Canceled* (never unobserved). The real leak is a non-cancellation fault (what the test injects). Fix is correct; the stated cause is misleading.
2. **Completeness gap** — the app-cancellation phase-1 exit still abandons sibling in-flight losers unobserved.
3. **Architecture** — a single `try/finally` sweep would cover every exit and close #2 automatically.
4. **Test robustness** — the process-global `UnobservedTaskException` + forced-GC assertion is contamination-prone and regresses the deterministic-test pattern PR #5712 established for this file.

Three minor items (present-day/`CosmosOperationCanceledException` coverage + mutation-test proof, lost `DefaultTrace` diagnostics on the observe path, and a pre-existing duplicate `Remove`/`Cancel`) are omitted here to keep the noise down — happy to add them on request.

---

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -409,6 +428,39 @@ private async Task<HedgingResponse> RequestSenderAndResultCheckAsync(
             }
         }
 
+        /// <summary>
+        /// Observes the exceptions of hedge requests that are abandoned when a winning response is
+        /// returned early. Losing hedges are cancelled via the hedge cancellation token source and
+        /// can complete faulted (for example with a <see cref="CosmosOperationCanceledException"/>).
```
kundadebdatta: **🟡 Recommendation** · Correctness: Misleading Rationale (comments + changelog)

**A cancelled loser completes Canceled, not Faulted — the stated cause is inaccurate**

This doc comment (and the matching changelog entry) states a losing hedge "can complete faulted (for example with a `CosmosOperationCanceledException`)". But `CosmosOperationCanceledException : OperationCanceledException`, and the C# async state-machine builder maps a thrown `OperationCanceledException` to `TrySetCanceled` — so an arm cancelled by the winner completes **Canceled**, whose `.Exception` is null and which never raises `TaskScheduler.UnobservedTaskException`.

The genuine leak this fix closes is a **non-cancellation (non-OCE) fault** on an abandoned loser (the historical disposed-clone / pre-dispatch `ArgumentNullException` root-fixed by #5613 / #6016 — exactly what the new regression test injects). The fix itself is correct and useful; only the documented reasoning is wrong and will send a future maintainer chasing cancellation races instead of the real cause. Consider rewording the comment and changelog to "a loser that faults with a non-cancellation exception".
---

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -216,6 +216,16 @@ internal override async Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
...
                                     }
+                                    else
+                                    {
+                                        // The losing hedge faulted/cancelled but the operation itself was
+                                        // not cancelled. completedTask has already been removed from
+                                        // requestTasks, so the later ObserveAbandonedHedgeTasks(...) at the
+                                        // winner-returns can never observe it. Observe it here so a
+                                        // raced-cancellation fault cannot escape as an unobserved task
+                                        // exception.
+                                        CrossRegionHedgingAvailabilityStrategy.ObserveAbandonedHedgeTasks(new[] { completedTask });
```
kundadebdatta: **🟡 Recommendation** · Reliability: Incomplete Coverage

**App-cancellation exit still abandons sibling losers unobserved**

This `else` correctly observes the just-removed faulted loser. But the sibling `if (applicationProvidedCancellationToken.IsCancellationRequested)` branch immediately above does `await (Task<HedgingResponse>)completedTask;`, which rethrows and unwinds the whole method — leaving any *other* in-flight losing arms still in `requestTasks` unobserved (there is no `finally`).

Those siblings leak only if they fault with a non-OCE exception during app-cancellation (narrow, since #5613 / #6016 root-fixed the known one), but it is the same leak class this PR sets out to close — on arguably the most cancellation-heavy path (user-cancel / e2e-timeout, including PPAF-implicit hedging). Recommend observing `requestTasks` on this path too, or better, the single-`finally` approach noted separately.
---

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
```diff
@@ -241,6 +251,11 @@ internal override async Task<ResponseMessage> ExecuteAvailabilityStrategyAsync(
                                             hedgeResponse.TargetRegionName);
                                     }
 
+                                    // The winning hedge is returning now; any still in-flight losers
+                                    // are abandoned. Observe their faults so a raced-cancellation
+                                    // exception cannot escape as an unobserved task exception.
+                                    CrossRegionHedgingAvailabilityStrategy.ObserveAbandonedHedgeTasks(requestTasks);
```
kundadebdatta: **🟡 Recommendation** · Maintainability: Robust Abstraction

**A single `finally` would cover every exit and close the gap above**

The fix relies on remembering to call `ObserveAbandonedHedgeTasks` at each of three exit sites, and already misses the app-cancellation throw path. Wrapping the hedging body in `try { ... } finally { ObserveAbandonedHedgeTasks(requestTasks); }` would guarantee observation on **every** exit (return *and* throw), close the app-cancel gap automatically, and make the three explicit call sites redundant. `requestTasks` at each exit already excludes the winner and the hedge timers, so a single sweep is safe.
---

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AvailabilityStrategyUnitTests.cs
```diff
@@ -1025,6 +1025,172 @@ await Assert.ThrowsExceptionAsync<InvalidOperationException>(
...
+                    primaryFaulted.TrySetResult(true);
+                    throw new ArgumentNullException("request");
+                }
+
+                // Winning arm: return a final result only after the primary has faulted.
+                await primaryFaulted.Task;
+                return new ResponseMessage(HttpStatusCode.OK);
+            };
+
+            TaskScheduler.UnobservedTaskException += handler;
```
kundadebdatta, line 1104: **🟡 Recommendation** · Reliability: Test Flakiness / Isolation

**Process-global `UnobservedTaskException` + forced GC is contamination-prone and regresses the deterministic-test standard**

This subscribes to the process-global `TaskScheduler.UnobservedTaskException` and asserts zero unobserved exceptions after `Task.Delay(50)` + `GC.Collect()` / `WaitForPendingFinalizers()`, with no source filtering. Two concrete problems:

1. `GC.Collect()` is process-global. The assembly runs sequentially (no `[assembly: Parallelize]`), but GC is not test-scoped — a faulted task leaked by an *earlier* test can be finalized inside this window and fail this test, misattributing the source (the assertion message itself admits it cannot attribute).
2. PR #5712 was written specifically to delete `Task.Delay`-based timing from this exact test file in favor of `TaskCompletionSource` / `CancellationToken.Register` signal-gating; this test resurrects that pattern.

Prefer asserting directly on the specific leftover hedge `Task`'s observed/faulted state (task-scoped), or at minimum filter captured exceptions to the injected `ArgumentNullException("request")` and add `[DoNotParallelize]` to future-proof. (The sibling `GenuineErrorWhileHedgeCtsNotCancelled_OriginalExceptionSurfaces` test is TCS-gated and already aligned with the standard.)
---

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
NaluTripician: Closed by the single `finally` (see the "A single finally would cover every exit" recommendation, now implemented). The app-cancellation branch still `await`s the just-completed arm to surface its outcome, but the enclosing `finally` now observes every *other* in-flight loser still in `requestTasks` as the method unwinds — so the sibling-leak on this cancellation-heavy path (user-cancel / e2e-timeout, incl. PPAF-implicit hedging) is gone without a dedicated call site.

<sub>🤖 Generated with GitHub Copilot CLI.</sub>

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
NaluTripician: Taken. Wrapped the entire hedging body in `try { ... } finally { ObserveAbandonedHedgeTasks(requestTasks); }`, so every exit sweeps the abandoned losers: the winner-return in both phases, the `ExceptionDispatchInfo.Capture(lastException).Throw()` rethrow, the app-cancellation throw, and the "completed without producing a response" throw. Removed the two now-redundant explicit calls at the winner-return sites.

I verified the single sweep is safe at every exit:
- `requestTasks` never contains the **winning** task — it is removed (via `Task.WhenAny` + `Remove`) before either winner-return.
- `requestTasks` never contains a **hedge-timer** task — timers are removed at the top of each phase-1 iteration and no timers are added in phase 2.
- `ObserveAbandonedHedgeTasks` neither awaits nor rethrows, so it cannot swallow a propagating exception or change the returned response / control flow.

Kept the one explicit observe on the phase-1 `continue` path: there the faulted loser has already been removed from `requestTasks` and is never awaited, so the `finally` sweep can't see it — it still needs its own observe.

<sub>🤖 Generated with GitHub Copilot CLI.</sub>

Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/AvailabilityStrategyUnitTests.cs
NaluTripician: Taken (your explicit "at minimum" option), plus a determinism pass:

- **Source-scoped assertion:** it now filters captured exceptions to the exact injected fault — an `ArgumentNullException` with `ParamName == "request"` — so a task leaked by an earlier test that happens to be finalized inside this GC window can no longer fail or be misattributed to this test.
- **Added `[DoNotParallelize]`.**
- **Removed the wall-clock timing:** replaced `await Task.Delay(50)` with `await Task.Yield()` + a bounded `GC.Collect()/GC.WaitForPendingFinalizers()` loop. The yield only lets trailing internal continuations drop their references before collection — a GC-settle step, not routing timing; routing stays `TaskCompletionSource`-gated per the #5712 standard.

On the fully task-scoped option (assert on the leftover hedge `Task` directly): the faulting `Task<HedgingResponse>` is created inside `CloneAndSendAsync` and is never surfaced to the caller, so a test can't get a handle to it without adding a production-only test hook. Since the behavior under test is specifically "the abandoned internal task must not raise the process-global `TaskScheduler.UnobservedTaskException`", the finalizer + GC + global handler is the only observation point available — I've made it deterministic and source-filtered rather than replace it.

Verified the test still fails without the production fix: neutralizing `ObserveAbandonedHedgeTasks` reproduces `Assert.AreEqual failed. Expected:<0>. Actual:<1> ... Value cannot be null. (Parameter 'request')`, and it passes with the fix restored — stable across repeated runs.

<sub>🤖 Generated with GitHub Copilot CLI.</sub>

Microsoft.Azure.Cosmos/src/Routing/AvailabilityStrategy/CrossRegionHedgingAvailabilityStrategy.cs
NaluTripician: Fixed — good catch, and confirmed against the code: `CosmosOperationCanceledException : OperationCanceledException`, and the async state-machine builder maps a thrown `OperationCanceledException` to `TrySetCanceled`, so a loser cancelled by the winning region completes **Canceled** (its `.Exception` is null) and never raises `UnobservedTaskException`. The genuine leak is a **non-cancellation (non-OCE) fault** on an abandoned loser (the disposed-clone / pre-dispatch `ArgumentNullException` root-fixed by #5613 / #6016, which the regression test injects).

Reworded to match reality in three places: the `ObserveAbandonedHedgeTasks` XML doc, the inline comments at the observe sites, and the changelog entry (which now also notes that a winner-cancelled loser completes Canceled and was never the source).

<sub>🤖 Generated with GitHub Copilot CLI.</sub>

Pilchie (approved):
LGTM except PR description.

kundadebdatta (approved):
LGTM, trimmed the PR description to keep it short and concise.

PR #6041, opened 2026-07-23
Client Encryption: Fixes cross-provider cache TTL contamination in CosmosDataEncryptionKeyProvider

MartinSarkany:
## Description

`CosmosDataEncryptionKeyProvider` wrote its per-provider cache lifetime (`EncryptionKeyStoreProvider.DataEncryptionKeyCacheTimeToLive`) into the process-global static `ProtectedDataEncryptionKey.TimeToLive` in the MDE SDK. In a process that constructs more than one provider (for example multi-tenant hosts), this was last-writer-wins: the most recently constructed provider''s TTL governed every provider, and a `null` value forced ~100-year caching process-wide. Decrypted key material could also be retained even when a provider requested `TimeSpan.Zero`.

This change removes the process-global mutation; each provider''s cache lifetime now applies only to that provider, and `TimeSpan.Zero` reliably prevents decrypted key material from being cached across operations.

Dependency: bumps `Microsoft.Data.Encryption.Cryptography` to `2.0.0-pre024`, which scopes the key/algorithm caches per `EncryptionKeyStoreProvider` instance (the paired MDE change). The fix is inert without this package version.

Testing:
- Unit: `Constructor_KeyStoreProvider_CapturesZeroTtl` / `CapturesNullTtl` (verbatim TTL capture).
- Emulator: `ValidateCachingIsolationAcrossCoexistingProviders` — two coexisting providers (Zero + 30 min) proven independent. Verified as a real regression guard (fails against the pre-fix baseline with the exact contamination symptom, passes with the fix).
- Full `MdeCustomEncryptionTests` emulator class: 49/49.

## Type of change

## Changelog

- [x] I have added a changelog entry under the Unreleased section in `changelog.md` for the user-facing impact of this change.


Microsoft.Azure.Cosmos.Encryption.Custom/src/CosmosDataEncryptionKeyProvider.cs
```diff
@@ -203,14 +198,19 @@ private static DekCache NewDekCache(DekCacheOptions dekCacheOptions)
...
+        /// The MDE SDK scopes the Protected Data Encryption key cache — and thus the lifetime of decrypted key
+        /// material — per <see cref="EncryptionKeyStoreProvider"/> instance. This method captures the cache lifetime
+        /// for this provider instance only, so one provider's configuration does not affect another in the same
+        /// process (for example in multi-tenant hosts that construct one provider per tenant). Relies on the MDE SDK
+        /// build whose key caches are per-<see cref="EncryptionKeyStoreProvider"/> instance.
+        /// </remarks>
         private void InitializeProtectedDataEncryptionKeyTimeToLive()
         {
             this.PdekCacheTimeToLive = this.EncryptionKeyStoreProvider.DataEncryptionKeyCacheTimeToLive;
-            ProtectedDataEncryptionKey.TimeToLive = this.PdekCacheTimeToLive ?? EffectivelyForeverTimeToLive;
```
kundadebdatta, line 213: `CreateAsync`  consumes  `cacheTimeToLive` only in the ternary  HasValue && == Zero . For every non- Zero  value,  `5` min ,  `30` min , or  `null`  — it calls the identical  `ProtectedDataEncryptionKey.GetOrCreateAsync(id, kek, wrappedDek, ct)`  with no TTL argument. 

The removed line was the only place in this repo where a finite/ null  magnitude was applied to MDE's cache. So after this PR, "each provider honors its own finite TTL" and " null  ⇒ indefinite per-instance" depend entirely on pre024's  GetOrCreateAsync  deriving expiry from `EncryptionKeyStoreProvider.DataEncryptionKeyCacheTimeToLive`  per instance. If it doesn't, finite and  null  TTLs silently collapse to MDE's ~2 h default (per the sibling package's own comment). No test covers finite-TTL expiry, the emulator test only checks caches-vs-doesn't within a short window, which passes regardless of the actual duration.

Can we confirm the paired MDE pre024 PR that makes  `GetOrCreateAsync`  per-instance TTL-aware; add an expiry-level test (short positive TTL, sleep past it, assert a re-unwrap) or at minimum a comment at the  CreateAsync  call site documenting that the magnitude is intentionally delegated to MDE. (This is the one item I'd resolve before merge. The  Zero  guarantee itself is safe.)

MartinSarkany: Thanks — the mechanics you traced are exactly right, and the question was worth asking. Confirmed end-to-end against the shipped MDE source:

**1. Per-instance TTL is honored.** The paired MDE change shipped as `2.0.0-pre024` (merged). Chain in that build: the `EncryptionKeyStoreProvider.DataEncryptionKeyCacheTimeToLive` setter assigns the per-instance `ProtectedDataEncryptionKeyCache.TimeToLive` -> `ProtectedDataEncryptionKey.GetOrCreateAsync` routes through `keyEncryptionKey.KeyStoreProvider.ProtectedDataEncryptionKeyCache` -> `LocalCache` stores with `AbsoluteExpirationRelativeToNow = TimeToLive`. So a finite value becomes that entry''s expiry and `null` means no absolute expiration — they are **not** silently replaced by the 2-hour default.

**2. Added an expiry-level test.** `ValidateProtectedDataEncryptionKeyCacheHonorsFiniteTtl` (EmulatorTests): configure a finite TTL, cold op unwraps, warm op reuses without unwrapping, wait past the TTL, next op unwraps again. It fails if the lifetime collapsed to the 2-hour default. I used a deliberately generous window (30s + 5s) so the warm operation cannot outrun it on a loaded agent.

**3. Added a call-site comment** documenting that `cacheTimeToLive` drives the cache-or-not decision here while the lifetime magnitude is delegated to the provider''s `DataEncryptionKeyCacheTimeToLive` (`GetOrCreateAsync` takes no TTL argument).

kundadebdatta (dismissed):
LGTM

kundadebdatta (dismissed):
LGTM

Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
```diff
@@ -36,7 +36,7 @@
   </ItemGroup>
 
   <ItemGroup>
-     <PackageReference Include="Microsoft.Data.Encryption.Cryptography" Version="2.0.0-pre015" />
+     <PackageReference Include="Microsoft.Data.Encryption.Cryptography" Version="2.0.0-pre024" />
```
ananth7592: Small thing on this bump — this is a floating pin, so NuGet's downgrade rules mean a consumer app can still land on something older than `pre024` via a transitive downgrade and not know. And the whole fix quietly reverts to shared-static behavior in that case, no build warning.


Worth pinning defensively here (`[2.0.0-pre024,)` or a `NU1605`-as-error guard on this package specifically)? And maybe a line in the changelog telling folks "don't override this below pre024"? Curious what you think.

ananth7592:
Nice one, Martin. The self-diagnosis in the PR body is very clear — you named the exact failure mode (static write in the ctor + multi-tenant last-writer-wins), the paired MDE change that makes the code side of the fix actually take effect, and the residual asymmetry on the `PlaintextDataEncryptionKey` path. 

Only one non blocking comment from my side

Microsoft.Azure.Cosmos.Encryption.Custom/src/Microsoft.Azure.Cosmos.Encryption.Custom.csproj
MartinSarkany: Made the minimum dependency explicit as `[2.0.0-pre024,)`. A consumer that directly selects an older version receives NU1605, which the .NET SDK treats as an error by default; warning settings in this project do not propagate to consumers. The changelog already states that `2.0.0-pre024` is required.

kundadebdatta (approved):
LGTM

ananth7592 (approved):

